PR #5263, opened 2016-12-29
Updated for Ubuntu 16

digitalcohesion:
I'm not entirely sure if I missed any other config steps for Ubuntu because I upgraded from 14 and may have had some things already in place, but this worked for me after the upgrade.

DO NOT DELETE THIS TEXT

#### Please note


laf:
Thanks for contributing :)

As an FYI, any future contributions will need you to sign the contributors agreement as per the checkbox at the top. We can let it slide for such a simple doc change this time :)

PR #5275, opened 2016-12-31
Add support for SVG images

corny:
DO NOT DELETE THIS TEXT

#### Please note


corny:
@laf the current PNG logos are looking aweful on high DPI displays.

I updated my pull request. `foreach (array("svg", "png") as $ext) {` should work with PHP 5.3? I replaced some logos by SVG versions. More work has to done, but PNG is still supported.

laf:
I think I'd be more comfortable personally if you could add a doc on how to generate svg images to be used otherwise people are going to continue to use png images.

You've got more shorthand array calls which need updating. https://travis-ci.org/librenms/librenms/jobs/187973765

Thanks for contributing btw. You do need to sign the contributors agreement - you removed all of that text above, I've added it back in, please confirm as per the instructions.

corny:
I added myself to the `CONTRIBUTING.md`, ok thanks for re-adding the text.

There is basically nothing to be aware of when creating SVG images. They are automatically scaled by stylesheets. I grabbed some images from wikipedia and I also cut some. Logos should just not be larger than ~ 20 kb. If the svg is too large, then paths should be simplified (e.g. with Inkscape). Where can I document it?

The SVG version of some logos is even smaller than the PNG.

corny:
I also changed the second shorthand array and set a height for logo inside of the `<span class="device_icon">`.

laf:
The contributors agreement needs to be signed as per the doc which means commiting a specific message.

As for docs, I'd say add something about the icon here:

http://docs.librenms.org/Developing/Support-New-OS/

and here:

http://docs.librenms.org/Support/FAQ/#how-do-i-add-support-for-a-new-os

corny:
Thanks, I added a dedicated commit for the contributors agreement.

laf:
Once you do some docs explaining to users creating new os' that svg is accepted and where to place the files then this will be good to merge imho.

corny:
I added the docs.

corny:
I used the [SVG optimizer](https://github.com/svg/svgo) shrink the images by more than 50 % in average.

includes/functions.php
```diff
@@ -146,9 +146,19 @@ function getImage($device)
 
 function getImageSrc($device)
 {
+    return 'images/os/' . getImageName($device);
+}
+
+function findImage($basename)
```
murrant: Can you rename this to findOsImage() or something?

murrant:
I would like to have SVGs for use in larger thatn 32x32 locations for v2, so +1.

The php code could use some refinement.  I think the findImage() function needs to be removed and folded in.  See code notes for more.

includes/functions.php
```diff
@@ -158,31 +168,34 @@ function getImageName($device, $use_database = true)
...
+        $distro,
+        $config['os'][$device['os']]['icon'],
+        $device['os'],
+    );
 
-    // guess the icon has the same name as the os
-    if (file_exists($config['html_dir'] . '/images/os/' . $device['os'] . '.png')) {
-        return $device['os'];
+    foreach ($possibilities as $icon) {
+        if (!empty($icon)) {
```
murrant: You don't need this check if you use is_file() instead of file_exists() (which returns true for directories)

includes/functions.php
```diff
@@ -158,31 +168,34 @@ function getImageName($device, $use_database = true)
...
+            $name = findImage($icon);
+            if (!empty($name)) {
```
murrant: If you bring the findImage code into this function, you won't need the !empty() check here...

corny:
@murrant thanks for the review. I pushed another commit.

murrant:
Looking good now.

I would prefer the red and blue for the Cisco logo however :)

corny:
@murrant The red/blue Cisco logo looks much better. I replaced it.

murrant:
@corny This needs a rebase (docs got moved around in another commit)

@librenms/reviewers Are we concerned about supporting IE8 and older?  SVG img tags is IE9+ (and any other browser)

laf:
IE 8 is completely end of life. 

I'd be happy to drop support.

corny:
I rebased my branch on top of master.

murrant:
Ok, I also reviewed the icons.  They all look great, a few deviate in a way that might be concerning.
apple - some might not like the change to the rainbow logo, I'm fine with it.
aruba and mimosa: these could be changed to just the first letter of the logo as they were before.

A good indicator to me is the favicon of a companies website.  Perhaps it would be a good idea to have both icon and logo (but it would be pointless if they are not used).

corny:
Yeah, I could not find a nice looking apple logo. Do you prefer another one?
https://commons.wikimedia.org/wiki/File:Apple_logo_black.svg (look at "Other versions")

I can cut the Aruba and Mimosa logos to the first letter. As a small square icon they would be definately easier to be recognized. Other opinions?

murrant:
I prefer the rainbow logo over the plain black.

corny:
I rebased my branch on top of master.

laf:
trigger ci

FTBZ:
IMO, I don't think that you can use the logo that you prefer. You need to use the correct corporate logo that match the actual design. The Apple rainbow logo isn't used since 1998...

laf:
@librenms/reviewers I'm ok with this. Just rebasing it now so it's ready for a merge.

murrant:
Did we want to change the Apple icon as FTBZ mentioned?  It would probably be to the monochrome grey one.

laf:
I'm not overly fussed but the black one is more modern imho.

corny:
I replaced by Apple logo by the grey one and rebased again.

Issue #5282, opened 2017-01-02
jpgraph is missing imageantialias on ubuntu 14.04


guggemand:
e20a242785de0acd849734c0fa9740179d2d30ac broke jpgraph on ubuntu 14.04

The following hides the error, and the graphs works again.

[13:56:53] <@laf> spec_: you could try editing ./vendor/amenadiel/jpgraph/src/image/Image.php
[13:57:13] <@laf> line 103
[13:57:18] <@laf> comment it out and see if it works

PR #5284, opened 2017-01-02
fix: Update jpgraph source file to remove check for imageantialias() …

laf:
…#5282

DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5282 

I've edited the source code directly for jpgraph here, basically implementing a change they have done themselves but in a version which drops support for php 5.3. https://github.com/HuasoFoundries/jpgraph/commit/d7c76bc0ecdcdc41affb41db75a797ac9400811e

@murrant - You ok with this? We won't be updating jpgraph until we drop php 5.3 support anyway at which point we can update and this fix will then be implemented anyway.


murrant:
Our Jpgraph is custom at this point anyway, until we don't support PHP 5.3...

PR #5290, opened 2017-01-03
Zombies arghhhhhhhhh

laf:
DO NOT DELETE THIS TEXT

#### Please note

No not

image

Rather

image

It looks like when a graph errors (for whatever reason), the graph_error() function is called which would run exit(); and not allow rrdtool_close() to run leaving zombiesssssss....

Sorry to tag you @murrant but you've looked at the graph code the most in recent times, does the above seem likely and therefore the fix?

### DIFF
--- a/html/includes/graphs/graph.inc.php
+++ b/html/includes/graphs/graph.inc.php
@@ -76,7 +76,6 @@ function graph_error($string)
             fpassthru($fd);
             fclose($fd);
             unlink($graphfile);
-            //exit();
         }
     } else {
         if (!$debug) {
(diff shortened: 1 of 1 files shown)


murrant:
Aaaaaaaaahhhhhhh!    Thanks for the fix @laf.  I would think rrdtool_close() is not required...  But maybe that is only on recent PHP versions...

laf:
I could 100% reliably reproduce it on centos7 with php7. Any graph which errored generated a defunct process, one page with 6 graphs that error = 6 defunct processes.

murrant:
Must be something wrong in \LibreNMS\Proc then...   It should close the process when it is destroyed (all objects are destroyed on exit).

PR #5315, opened 2017-01-05
feature: Add support for disabling Auto TLS in mail notifications

### LINKED ISSUE #5314: email_smtp_secure is not honored on TLS-enabled servers
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php`.
- [ ] If you would like us to add a new device or your device is listed as Generic then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)
````
==========================================================
Component | Version
--------- | -------
LibreNMS  | 036f411e138bff272a15e9b64b7df9f3e7bdedbb
DB Schema | 152
PHP       | 5.6.29
MySQL     | 5.6.34
RRDTool   | 1.6.0
SNMP      | NET-SNMP 5.7.3
==========================================================
````
The SMTPAutoTLS option needs to be set to `false` and passed to PHPMailer to prevent it from opportunistically using TLS even when 'email_smtp_secure' is not set.

Reference https://github.com/PHPMailer/PHPMailer/issues/550#issuecomment-153181003

### PR DESCRIPTION
**Pull request by rdnn**
DO NOT DELETE THIS TEXT

#### Please note

### CODE REVIEW

**Comment by rdnn** [comment]
The root cause is that we're using a self-signed certificate for our internal SMTP server which is used for a couple of other projects that require TLS. We'd rather not use TLS for alerting e-mails so we have email_smtp_secure set to ''. This causes an issue since PHPMailer is attempting to take advantage of the fact that our SMTP server advertises TLS support, but the connection fails due to the self-signed certificate.

There are other options available: we can spin up a new SMTP server with TLS implicitly disabled or a config option could be added to LibreNMS to permit PHPMailer to accept self-signed certificates.

**Comment by laf** [comment]
Does it not make more sense then to add his option to the webui so users can choose?

**Comment by rdnn** [comment]
Not necessarily. What we're ultimately talking about here is mitigating unexpected behavior. If one sets email_smtp_secure to '', the expected behavior is that TLS won't be used at all.

Adding a configuration option to accept self-signed certificates doesn't mitigate the unexpected behavior but definitely should be considered as a feature request. Adding a config option to disable auto TLS seems a bit unnecessary as I can't imagine an instance where one would want to implicitly not enable TLS but would want auto TLS to take place.

I'm open to ideas, but I believe the best, and simplest, course of action is to simply disable auto TLS. I modified the title of the issue to more appropriately reflect the real issue.

**Comment by murrant** [comment]
Isn't there a way to allow it to accept a self-signed cert?

**Comment by rdnn** [comment]
@murrant Absolutely; there's a way to configure PHPMailer to accept self-signed certificates. This pull request is designed to coerce PHPMailer into honoring the email_smtp_secure option by disabling the use of opportunistic TLS when people don't desire the use of TLS.

Another pull request can, and should be, created to create a configuration option to allow self-signed certificates. However like I mentioned above, creating a configuration option to allow self-signed certificates doesn't prevent PHPMailer from defaulting to using TLS on TLS-enabled SMTP servers even when email_smtp_secure is not configured (i.e., set to '').

Ultimately the goal of the pull request is this: if one doesn't want to use TLS, email_smtp_secure set to '' should disable TLS no questions asked. Right now, email_smtp_secure is not doing that (as a result of PHPMailer's SMTPAutoTLS configuration option not being set to false) which results in unexpected/undesired behavior.

**Comment by laf** [comment]
Would you mind testing this patch and sql code, it will add an option to the webui where you can select to disable this tls option:

```php
diff --git a/html/pages/settings/alerting.inc.php b/html/pages/settings/alerting.inc.php
index 24286df..14ae6fa 100644
--- a/html/pages/settings/alerting.inc.php
+++ b/html/pages/settings/alerting.inc.php
@@ -278,6 +278,11 @@ $mail_conf = array(
           'type'               => 'select',
           'options'            => $dyn_config['email_smtp_secure'],
     ),
+    array('name'               => 'email_auto_tls',
+        'descr'                => 'SMTP Auto TLS Support',
+        'type'                 => 'select',
+        'options'              => array('true', 'false'),
+    ),
     array('name'               => 'email_smtp_auth',
           'descr'              => 'SMTP Authentication',
           'type'               => 'checkbox',
diff --git a/includes/functions.php b/includes/functions.php
index 1954fdc..3676599 100644
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -797,6 +797,7 @@ function send_mail($emails, $subject, $message, $html = false)
                 $mail->Port       = $config['email_smtp_port'];
                 $mail->Username   = $config['email_smtp_username'];
                 $mail->Password   = $config['email_smtp_password'];
+                $mail->SMTPAutoTLS= $config['email_auto_tls'];
                 $mail->SMTPDebug  = false;
                 break;
             default:
```

sql code:

```sql
INSERT INTO `config` (`config_name`, `config_value`, `config_default`, `config_descr`, `config_group`, `config_group_order`, `config_sub_group`, `config_sub_group_order`, `config_hidden`, `config_disabled`) VALUES ('email_auto_tls','true','true','Enable / disable Auto TLS support','alerting',0,'general',0,'0','0');
```

rdnn:
@laf Sorry for the slow reply. That works!

laf:
@rdnn 

Ace, do you want just put that code into this PR then along with the schema file and we can merge.

rdnn:
@laf Should be good to go... squashed the commits.

Thanks!

PR #5316, opened 2017-01-05
Update Centos 7 nginx install steps

matthewaaronthacker:
Found some issues on a fresh install of Centos 7. PHP defaults to listening on a port, while LibreNMS expects a unix socket. Also, the listen owner has to match the account that nginx runs under.
This is my first proposed change on Github so if I'm doing this all wrong, feel free to smack me down.
Thanks.

DO NOT DELETE THIS TEXT

#### Please note


laf:
Thanks for this, we try and keep the docs standard so we can maintain them across all variations. If you can either remove the two new headers you've added `####` or add the same to the other three main install docs, Centos7 Apache, Ubuntu 16.04 Apache / Nginx then we can merge this.

Issue #5118, opened 2016-12-06
Using regexp in device groups bugged?


sliddjur:
- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
Yes

- [x] Please include all of the information between the 

```
==========================================================
Component | Version
--------- | -------
LibreNMS  | 09767f7e95d425772760481a1b36cae57d6dea4a
DB Schema | 151
PHP       | 5.5.14
MySQL     | 10.0.22-MariaDB
RRDTool   | 1.4.7
SNMP      | NET-SNMP 5.7.3
==========================================================

```
- [x] Please provide as much detail as possible.

When creating a device group with this pattern: 
`%devices.hostname ~ "[[:digit:]]{5}p-swi"
`

The first time I visit the group in web gui (/devices/group=84/) I see all devices. If I then press refresh within a couple of seconds, I see only half of the devices, if I then press refresh again I only see a couple of devices. In the end it ended up only showing one device.

To retrigger this bug I only changed the pattern a little bit like this `%devices.hostname ~ "[[:digit:]]{5}p-sw"
`

The manual says regexp can be used http://docs.librenms.org/Extensions/Device-Groups/
Regexp manual for mariaDB https://mariadb.com/kb/en/mariadb/regular-expressions-overview/

laf:
Hmm, seems to work for me.

What does running the following query multiple times result in for you: `select count(*) from devices where hostname REGEXP "[[:digit:]]{5}p-swi";`

sliddjur:
@laf  yes that gives me the right amount of devices from the sql query.
However, when using the group in librenms, I get no results.

This is the group I've created. ![capture](https://cloud.githubusercontent.com/assets/10002057/20958543/6f2bd5ee-bc56-11e6-919e-7bb3adeefd66.PNG)

I created a new group, with the same regexp. This successfully shows 22 out of 25 devices. 22 in web gui, and 25 in sql query.

sliddjur:
One thing I noticed when adding another pattern to the rule, it messed up the formatting. Not only in the GUI, but this gets bugged in the database as well. 
![capture](https://cloud.githubusercontent.com/assets/10002057/20961314/c26d0242-bc64-11e6-89b3-6c4e7b9f2ec2.PNG)

[...]

PR #5327, opened 2017-01-06
feature: Add mysql failed query logging + fixed queries that break ONLY_FULL_GROUP_BY

laf:
DO NOT DELETE THIS TEXT

#### Please note

Will inline comment on these. Genuinely shouldn't break anything but fixes a lot of broken queries or ones that don't work with ONLY_FULL_GROUP_BY enabled which is default.


build.sql
```diff
@@ -58,7 +58,6 @@ CREATE TABLE IF NOT EXISTS `ports_vlans` ( `port_vlan_id` int(11) NOT NULL AUTO_
...
 CREATE TABLE IF NOT EXISTS `cmpMemPool` (  `cmp_id` int(11) NOT NULL auto_increment,  `Index` varchar(8) NOT NULL,  `cmpName` varchar(32) NOT NULL,  `cmpValid` varchar(8) NOT NULL,  `device_id` int(11) NOT NULL,  `cmpUsed` int(11) NOT NULL,  `cmpFree` int(11) NOT NULL,  `cmpLargestFree` int(11) NOT NULL,  `cmpAlternate` tinyint(4) default NULL,  PRIMARY KEY  (`cmp_id`),  KEY `device_id` (`device_id`)) ENGINE=MyISAM  DEFAULT CHARSET=latin1;
 CREATE TABLE IF NOT EXISTS `hrDevice` ( `hrDevice_id` int(11) NOT NULL auto_increment,  `device_id` int(11) NOT NULL,  `hrDeviceIndex` int(11) NOT NULL,  `hrDeviceDescr` text NOT NULL,  `hrDeviceType` text NOT NULL,  `hrDeviceErrors` int(11) NOT NULL,  `hrDeviceStatus` text NOT NULL,  `hrProcessorLoad` tinyint(4) default NULL,  PRIMARY KEY  (`hrDevice_id`),  KEY `device_id` (`device_id`)) ENGINE=MyISAM  DEFAULT CHARSET=latin1;
-CREATE TABLE IF NOT EXISTS `dbSchema` ( `revision` int(11) NOT NULL default '0', PRIMARY KEY (`revision`)) ENGINE=MyISAM DEFAULT CHARSET=latin1;
```
laf: We have already created the table so no point trying again.

html/includes/functions.inc.php
```diff
@@ -788,9 +788,9 @@ function getlocations()
 
     // Fetch regular locations
     if ($_SESSION['userlevel'] >= '5') {
-        $rows = dbFetchRows('SELECT D.device_id,location FROM devices AS D GROUP BY location ORDER BY location');
```
laf: device_id is never used so get rid to fix query

html/includes/functions.inc.php
```diff
@@ -788,9 +788,9 @@ function getlocations()
...
+        $rows = dbFetchRows('SELECT location FROM devices AS D GROUP BY location ORDER BY location');
     } else {
-        $rows = dbFetchRows('SELECT D.device_id,location FROM devices AS D, devices_perms AS P WHERE D.device_id = P.device_id AND P.user_id = ? GROUP BY location ORDER BY location', array($_SESSION['user_id']));
```
laf: Same here

html/includes/print-alert-rules.php
```diff
@@ -143,7 +143,7 @@
 
 echo '</select></td>';
 
-$count_query = 'SELECT COUNT(id)';
```
laf: Can't use COUNT(id) without using GROUP BY which we don't need.

html/includes/print-alert-rules.php
```diff
@@ -152,7 +152,7 @@
     $param = array($device['device_id']);
 }
 
-$query       = " FROM alert_rules $sql ORDER BY id ASC";
```
laf: don't need ORDER BY for count query so moved to full query.

html/includes/print-map.inc.php
```diff
@@ -84,7 +84,7 @@
                              `D1`.`device_id` != `D2`.`device_id`
                              $where
                              $sql
-                      GROUP BY `P1`.`port_id`,`P2`.`port_id`
```
laf: GROUP BY must contain all the fields in the SELECT. Tested and # results are the same.

html/includes/print-map.inc.php
```diff
@@ -125,7 +125,7 @@
                              `remote_device_id` != 0
                              $where
                              $sql
-                      GROUP BY `P1`.`port_id`,`P2`.`port_id`
```
laf: As above.

html/includes/table/alert-schedule.inc.php
```diff
@@ -25,7 +25,7 @@
     $sql .= " AND (`S`.`title` LIKE '%$searchPhrase%' OR `S`.`start` LIKE '%$searchPhrase%' OR `S`.`end` LIKE '%$searchPhrase%')";
 }
 
-$count_sql = "SELECT COUNT(`id`) $sql";
```
laf: id doesn't exist, it's schedule_id

html/includes/table/customers.inc.php
```diff
@@ -38,7 +38,7 @@
     $sql .= " LIMIT $limit_low,$limit_high";
 }
 
-$sql = "SELECT * $sql";
```
laf: We only use port_descr_descr so change SELECT to only that.

html/includes/table/poll-log.inc.php
```diff
@@ -26,9 +26,12 @@
     $sort = 'last_polled_timetaken DESC';
 }
 
-$sql .= " AND D.status ='1' AND D.ignore='0' AND D.disabled='0' ORDER BY $sort";
```
laf: Same as other comment, don't need ORDER BY for count query.

html/pages/ports.inc.php
```diff
@@ -87,9 +87,9 @@
 <?php
 
 if ($_SESSION['userlevel'] >= 5) {
-    $results = dbFetchRows("SELECT `device_id`,`hostname` FROM `devices` GROUP BY `hostname` ORDER BY `hostname`");
```
laf: GROUP BY hostname is irrelevant here. hostnames are unique anyway. Same for other queries in this file.

html/pages/routing/bgp.inc.php
```diff
@@ -240,14 +240,6 @@
             }
         }
 
-        $peerhost = dbFetchRow('SELECT * FROM ipaddr AS A, ports AS I, devices AS D WHERE A.addr = ? AND I.port_id = A.port_id AND D.device_id = I.device_id', array($peer['bgpPeerIdentifier']));
```
laf: `$peername` set and never used so removed as query for `$peerhost` is totally wrong (non-existent table).

html/pages/search/arp.inc.php
```diff
@@ -38,7 +38,7 @@
     $param[] = $_SESSION['user_id'];
 }
 
-$sql .= " WHERE M.port_id = P.port_id AND P.device_id = D.device_id $where GROUP BY `device_id` ORDER BY `hostname`";
```
laf: No need for GROUP BY device_id as they are unique already.

html/pages/search/ipv4.inc.php
```diff
@@ -35,7 +35,7 @@
     $param[] = $_SESSION['user_id'];
 }
 
-$sql .= " $where GROUP BY `hostname` ORDER BY `hostname`";
```
laf: Same here.

html/pages/search/ipv6.inc.php
```diff
@@ -34,7 +34,7 @@
     $param[] = $_SESSION['user_id'];
 }
 
-$sql .= " $where GROUP BY `hostname` ORDER BY `hostname`";
```
laf: Same here.

html/pages/search/mac.inc.php
```diff
@@ -34,7 +34,7 @@
     $param[] = $_SESSION['user_id'];
 }
 
-$sql .= " $where GROUP BY `hostname` ORDER BY `hostname`";
```
laf: Same here

html/pages/services.inc.php
```diff
@@ -99,10 +99,10 @@
     </tr>
 <?php
 if ($_SESSION['userlevel'] >= '5') {
-    $host_sql = 'SELECT * FROM devices AS D, services AS S WHERE D.device_id = S.device_id GROUP BY D.hostname ORDER BY D.hostname';
```
laf: Same here

includes/dbFacile.php
```diff
@@ -39,23 +39,14 @@ function dbQuery($sql, $parameters = array())
         }
     }
 
-    /*
```
laf: Removed redundant code

includes/dbFacile.php
```diff
@@ -39,23 +39,14 @@ function dbQuery($sql, $parameters = array())
...
-        $time_start = microtime(true);
-     */
-
     $result = mysqli_query($database_link, $fullSql);
-    // sets $this->result
-    /*
-        if($this->logFile) {
-        $time_end = microtime(true);
-        fwrite($this->logFile, date('Y-m-d H:i:s') . "\n" . $fullSql . "\n" . number_format($time_end - $time_start, 8) . " seconds\n\n");
+    if (!$result) {
```
laf: So what we do here is check if the mysql log level is INFO or ERROR AND don't contain duplicate entry in the error (we have these and they need to be fixed but will come back to that) and print out any mysql errors OR if the log level is DEBUG then just print out all failed queries.

I've then used this to find all broken queries.

includes/dbFacile.php
```diff
@@ -420,7 +411,11 @@ function dbMakeQuery($sql, $parameters)
     krsort($namedParams);
 
     // split on question-mark and named placeholders
-    $result = preg_split('/(\?|:[a-zA-Z0-9_-]+)/', $sql, -1, (PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE));
+    if (preg_match('/(\[\[:[\w]+:\]\])/', $sql)) {
```
laf: This is a 'hack' I couldn't get a preg_split regex to work across all cases so this retains existing compatibility but checks for people using mysql regex with [[:SOMETHING:]].

If someone can find a good regex to cover all basis then go for it.

It currently splits on ? and : as they are both placeholders for variable substitution and causes problems with [[:digit:]] as per issue #5118 (so fixes that)

includes/defaults.inc.php
```diff
@@ -31,6 +31,8 @@
 
 // MySQL extension to use
 $config['db']['extension']       = 'mysqli';//mysql and mysqli available
+// MySQL Debug level
+$config['mysql_log_level']       = 'ERROR';
```
laf: Set this as the default so all installs will log errors. We could set it to NONE but imho this is now handy to find non-working things easily and won't spam the log file anyway.

laf: Well unless someone has an issue.

includes/discovery/cisco-pw.inc.php
```diff
@@ -53,7 +53,7 @@
     // Cycle the list of pseudowires we cached earlier and make sure we saw them again.
     foreach ($device['pws_db'] as $pw_id => $pseudowire_id) {
         if (empty($device['pws'][$pw_id])) {
-            dbDelete('vlans', '`pseudowire_id` = ?', array($pseudowire_id));
```
laf: Yeah, lol

includes/sql-schema/update.php
```diff
@@ -36,34 +36,6 @@
     $insert = 1;
 }
 
-// For transition from old system
```
laf: I literally can't find where we do anything with revision - I think it's a hang up from the fork so got rid of it.

sql-schema/002.sql
```diff
@@ -75,7 +75,6 @@ ALTER TABLE `temperature` ADD `temp_precision` INT(11) NULL DEFAULT '1';
 UPDATE temperature SET temp_precision=10 WHERE temp_tenths=1;
 ALTER TABLE `temperature` ADD  `temp_index` INT NOT NULL AFTER  `temp_host` , ADD  `temp_mibtype` VARCHAR( 32 ) NOT NULL AFTER  `temp_index`;
 ALTER TABLE `temperature` DROP `temp_tenths`;
-CREATE TABLE IF NOT EXISTS `dbSchema` ( `revision` int(11) NOT NULL default '0', PRIMARY KEY (`revision`)) ENGINE=MyISAM DEFAULT CHARSET=latin1;
```
laf: We've already created this table (then tried 3 more times!)

laf:
Pushed further updates to the base sql build which fixes invalid queries. I have a feeling build.sql contains a lot of what was in 001-006.sql as most of the tidying up was done there.

laf:
Exported the database using this new branch, the old branch with both strict disabled and some defaults in centos. A diff shows the difference below (comparing old with non-strict and new shows only differences are datetime)

```
[root@mysql01 ~]# diff /tmp/ltest_old_nonstrict.sql /tmp/ltest_new.sql
151,152c151,152
<   `start` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00',
<   `end` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00',
---
>   `start` timestamp NOT NULL DEFAULT '1970-01-01 23:00:01',
>   `end` timestamp NOT NULL DEFAULT '1970-01-01 23:00:01',
895c895,896
<   `version` int(11) NOT NULL
---
>   `version` int(11) NOT NULL DEFAULT '0',
>   PRIMARY KEY (`version`)
1275c1276
<   `datetime` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
---
>   `datetime` datetime NOT NULL DEFAULT '1000-01-01 00:00:00',
3391c3392
< -- Dump completed on 2017-01-06 15:30:25
---
> -- Dump completed on 2017-01-06 15:29:19
[root@mysql01 ~]# diff /tmp/ltest_old_strict.sql /tmp/ltest_new.sql
142a143,168
> -- Table structure for table `alert_schedule`
> --
>
> DROP TABLE IF EXISTS `alert_schedule`;
> /*!40101 SET @saved_cs_client     = @@character_set_client */;
> /*!40101 SET character_set_client = utf8 */;
> CREATE TABLE `alert_schedule` (
>   `schedule_id` int(11) NOT NULL AUTO_INCREMENT,
>   `start` timestamp NOT NULL DEFAULT '1970-01-01 23:00:01',
>   `end` timestamp NOT NULL DEFAULT '1970-01-01 23:00:01',
>   `title` varchar(255) NOT NULL,
>   `notes` text NOT NULL,
>   PRIMARY KEY (`schedule_id`)
> ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
> /*!40101 SET character_set_client = @saved_cs_client */;
>
> --
> -- Dumping data for table `alert_schedule`
> --
>
> LOCK TABLES `alert_schedule` WRITE;
> /*!40000 ALTER TABLE `alert_schedule` DISABLE KEYS */;
> /*!40000 ALTER TABLE `alert_schedule` ENABLE KEYS */;
> UNLOCK TABLES;
>
> --
869c895,896
<   `version` int(11) NOT NULL
---
>   `version` int(11) NOT NULL DEFAULT '0',
>   PRIMARY KEY (`version`)
1238a1266,1296
> -- Table structure for table `eventlog`
> --
>
> DROP TABLE IF EXISTS `eventlog`;
> /*!40101 SET @saved_cs_client     = @@character_set_client */;
> /*!40101 SET character_set_client = utf8 */;
> CREATE TABLE `eventlog` (
>   `event_id` int(11) NOT NULL AUTO_INCREMENT,
>   `host` int(11) NOT NULL DEFAULT '0',
>   `device_id` int(11) NOT NULL,
>   `datetime` datetime NOT NULL DEFAULT '1000-01-01 00:00:00',
>   `message` text CHARACTER SET latin1,
>   `type` varchar(64) CHARACTER SET latin1 DEFAULT NULL,
>   `reference` varchar(64) CHARACTER SET latin1 NOT NULL,
>   PRIMARY KEY (`event_id`),
>   KEY `host` (`host`),
>   KEY `datetime` (`datetime`),
>   KEY `device_id` (`device_id`)
> ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
> /*!40101 SET character_set_client = @saved_cs_client */;
>
> --
> -- Dumping data for table `eventlog`
> --
>
> LOCK TABLES `eventlog` WRITE;
> /*!40000 ALTER TABLE `eventlog` DISABLE KEYS */;
> /*!40000 ALTER TABLE `eventlog` ENABLE KEYS */;
> UNLOCK TABLES;
>
> --
3334c3392
< -- Dump completed on 2017-01-06 15:29:54
---
> -- Dump completed on 2017-01-06 15:29:19

```

Rosiak:
Doesn't seem to break my test install, haven't reviewed the actual code yet.

murrant:
Looks good, I didn't test and compare a fresh db run, been running it for a day without issue.   I have a few errors logged, but I'll fix those in a different PR.

laf:
We ok for a merge then?

PR #5333, opened 2017-01-06
feature: Oxidized basic config search

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

Hacked this together this eve.

Depending on the progress of https://github.com/ytti/oxidized-web/pull/111

This enables you to leverage the Oxidized's API Config search function, giving you the ability to search through configs stored in Oxidized. Example could be that you want to find all the Cisco devices with "no service password-encryption" command set.

This is VERY basic, code should be refactored and prettified some. 
Output should also be better as it only yields matches for now.
Feel free to take it from here, at least this is a start.

screen shot 2017-01-07 at 00 10 04

### DIFF
--- a/html/includes/forms/search-oxidized-config.inc.php
+++ b/html/includes/forms/search-oxidized-config.inc.php
@@ -2,7 +2,7 @@
 /*
  * LibreNMS
  *
- * Copyright (c) 2015 Søren Friis Rosiak <sorenrosiak@gmail.com>
+ * Copyright (c) 2017 Søren Friis Rosiak <sorenrosiak@gmail.com>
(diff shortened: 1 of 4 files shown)


laf:
@Rosiak can you give this small update a test, it's just to use the helper functions to clean and display code.

Rosiak:
@laf Works fine, thanks!

DonovanEsterhuizen:
Where in th ui can I find this?

DonovanEsterhuizen:
Has this been released yet?

laf:
It's already live in the dev / daily branch but relies on an update from oxidized which afaik isn't yet released.

PR #5338, opened 2017-01-07
refactor: MySQL strict and query fixes

laf:
DO NOT DELETE THIS TEXT

#### Please note

These now start to address STRICT_TRANS_TABLES


html/pages/deluser.inc.php
```diff
@@ -14,7 +14,7 @@
             $delete_username = dbFetchCell('SELECT username FROM users WHERE user_id = ?', array($vars['id']));
 
             if ($vars['confirm'] == 'yes') {
-                if (deluser($delete_username)) {
+                if (deluser($vars['id']) >= 0) {
```
murrant: Why not put the >= 0 check inside the function so it just returns a boolean?

laf: That function is in all of the authentication modules (well 3/4) so this seemed a bit more central.

murrant: I was thinking of the function.   What do you want to know upon return of the function.  Was it successful or not.  (rather than the id that you already sent or not if if it failed).

laf: Are you talking about the deluser() function?

murrant: yes.

laf: That function is in all of the auth modules so it would mean updating them all to do that same check and return true / false when we can just do it here.

murrant: Yes, because when you call deluser you will never want to know the user id you deleted.
Basically, there is a hidden rule that you need to add `>= 0` after this method everytime it is called.

I understand that it is only called once and there is three implementations...  Can't fix everything at once.

laf: I'm happy to update it but it just seems a bit ott for a function that we most likely won't ever call unless someone creates a deluser.php for cli.

laf:
Rebased

html/pages/deluser.inc.php
laf: Want me to update the deluser() functions to be able to merge this?

murrant: It's fine, I guess I'm just trying to express a general guideline...

Issue #5319, opened 2017-01-05
White border around LibreNMS logo in 'dark' theme


axemann:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.

Install is up-to-date as of 1/5/2017

- [x] Please include all of the information between the `====================================` section of `./validate.php`.

Component | Version
--------- | -------
LibreNMS  | 036f411e138bff272a15e9b64b7df9f3e7bdedbb
DB Schema | 153
PHP       | 7.0.8-0ubuntu0.16.04.3
MySQL     | 10.0.28-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3

- [x] If you would like us to add a new device or your device is listed as Generic then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)
- [x] Please provide as much detail as possible.

It appears that selecting the 'dark' theme/style in the config does not switch the logo from librenms_logo_light.png to librenms_logo_dark.png as expected any longer.  This was not an issue prior to 12/10/2016 from what I can tell/remember.

Gorian:
Adding a screenshot for reference.

![image](https://cloud.githubusercontent.com/assets/1012176/21700960/d1baa4fc-d357-11e6-9210-26692e10f7c2.png)

axemann:
Thanks Gorian, I totally meant to attach those and didn't.  😞

Gorian:
Hehe, no problem ^.^

InsaneSplash:
This is due to the order in which the title_image is defined. A quick work around until this is resolved is to place the following configuration line in to config.php

$config['title_image'] = 'images/librenms_logo_'.$config['site_style'].'.png';

axemann:
Thanks @InsaneSplash for that.  Works a treat. 😄

PR #5342, opened 2017-01-08
fix: dark/mono logo was incorrect #5319

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5319


murrant:
Probably more like this.  Ideally, we would only have include files that contain functions or definitions, no logic.

Issue #5307, opened 2017-01-04
pdf report error


kristoferus75:
Hi !

If i want to create an pdf report under alert log settings i get this error:

TCPDF ERROR: [Image] Unable to get the size of the image: /opt/librenms/vendor/tecnickcom/tcpdf/examples/images/../../images/librenms_logo_light.png

kind regards

kristoferus75

PR #5343, opened 2017-01-08
fix: hack but saves editing the tcppdf files in vendor folder #5307

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5307

### DIFF
--- a/html/pdf.php
+++ b/html/pdf.php
@@ -48,7 +48,7 @@ $pdf->setTextShadow(array('enabled' => false, 'depth_w' => 0.2, 'depth_h' => 0.2
 
 if (isset($_GET['report']) && !empty($_GET['report'])) {
     $report = mres($_GET['report']);
-    $pdf->SetHeaderData('../../'.$config['title_image'], 40, ucfirst($report), $config['project_name'], array(0, 0, 0), array(0, 64, 128));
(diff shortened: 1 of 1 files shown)


html/pdf.php
```diff
@@ -48,7 +48,7 @@
 
 if (isset($_GET['report']) && !empty($_GET['report'])) {
     $report = mres($_GET['report']);
-    $pdf->SetHeaderData('../../'.$config['title_image'], 40, ucfirst($report), $config['project_name'], array(0, 0, 0), array(0, 64, 128));
+    $pdf->SetHeaderData('../../../../../html/'.$config['title_image'], 40, ucfirst($report), $config['project_name'], array(0, 0, 0), array(0, 64, 128));
```
murrant: why not use $config['install_dir']?

laf: Because the pdf library already has it's path set so it ends up being /opt/librenms/vendor/phppdf/dasa/dasdas/dasdas/opt/librenms/html/

murrant: That's pretty annoying...

laf: Yeah, we can override it in the config file of the pdf library but if we update it will potentially revert that anyway

Issue #5140, opened 2016-12-09
APC Envirnomental  Monitoring


difi80211g:
Looking for device support on an APC Envirnomental box to report the temperature and humidity inside of LibreNMS.  

Here is the output of the commands:
http://pastebin.com/1Br8n6nP

at the time of testing the temperature was 77 F and the relative humidity was 7%

Thank you in advance

Rosiak:
Any way you can provide access to this device?

difi80211g:
If you can send me a private email with the address you are coming from I could.

FTBZ:
This is for NetBoz series?

difi80211g:
No it is not, Here is a document on it.
http://www.apc.com/salestools/ASTE-6Z5QAY/ASTE-6Z5QAY_R0_EN.pdf

Rosiak:
@difi80211g Which mail address shall send the info to?

Rosiak:
Any news @difi80211g ?

difi80211g:
gmdjunk (it's a gmail address)

laf:
Can you try adding `.1.3.6.1.4.1.318` to the snmpwalk command instead of just `.`

axemann:
I'd like to see support for these devices as well.  

@Rosiak, I can provide access to an [AP9320](http://www.apc.com/salestools/ASTE-6Z5QAX/ASTE-6Z5QAX_R0_EN.pdf) that I have, which is very similar to the AP9319 that @difi80211g has, if needed.

Here's a pastebin of the data @laf was asking for, but for the '9320: [http://pastebin.com/MSui5WT1](http://pastebin.com/MSui5WT1).  The '19 should be similar, but with slightly less output.

I'm actually working on adding support for these on my own system, but I suck at coding, so that banging sound everyone is hearing is me applying face to keyboard repeatedly.  🙁

axemann:
Guys,

I have some additional code I put together for pulling the Sensors and Contacts on my particular unit.  I don't want to mung up the PR that @Rosiak submitted for the initial bits, so I'll wait until after his is approved to see if I need to make any changes.

nerdalertdk:
If you need an test unit I have one I can set up?

PR #5356, opened 2017-01-09
newdevice: APC Environmental monitoring units #5140

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5140

Few things.

1. These devices allow users to specify the data is in Fahrenheit rather than Celsius so I've added a function to convert.

2. The includes/polling/sensors/X/ folders were named wrong so have been moved. They don't follow the plural like disco but use the name registered in discover_sensor() which is the singular!

3. Added a pre-cache for sensor polling.


includes/discovery/sensors/humidity/aos-emu2.inc.php
```diff
@@ -0,0 +1,40 @@
...
+
+if ($device['os'] === 'aos-emu2') {
+    foreach ($emu2_temp as $id => $temp) {
+        if (isset($temp['emsProbeStatusProbeHumidity']) && $temp['emsProbeStatusProbeHumidity'] > 0) {
+            $index           = $temp['emsProbeStatusProbeIndex'];
+            $oid             = '.1.3.6.1.4.1.318.1.1.10.3.13.1.1.3.' . $index;
+            $descr           = $temp['emsProbeStatusProbeName'];
+            $low_limit       = $temp['emsProbeStatusProbeMinHumidityThresh'];
+            $low_warn_limit  = $temp['emsProbeStatusProbeLowHumidityThresh'];
+            $high_limit      = $temp['emsProbeStatusProbeMaxHumidityThresh'];
```
Rosiak: The device I tested against doesn't send any info back for:
emsProbeStatusProbeMaxHumidityThresh
emsProbeStatusProbeMinHumidityThresh

includes/discovery/sensors/humidity/aos-emu2.inc.php
```diff
@@ -0,0 +1,40 @@
...
+ * @link       http://librenms.org
+ * @copyright  2016 Neil Lathwood
+ * @author     Neil Lathwood <neil@lathwood.co.uk>
+ */
+
+if ($device['os'] === 'aos-emu2') {
+    foreach ($emu2_temp as $id => $temp) {
+        if (isset($temp['emsProbeStatusProbeHumidity']) && $temp['emsProbeStatusProbeHumidity'] > 0) {
+            $index           = $temp['emsProbeStatusProbeIndex'];
+            $oid             = '.1.3.6.1.4.1.318.1.1.10.3.13.1.1.3.' . $index;
```
Rosiak: I belive this should be .1.3.6.1.4.1.318.1.1.10.3.13.1.1.6.

includes/discovery/sensors/temperatures/aos-emu2.inc.php
```diff
@@ -0,0 +1,40 @@
...
+ * @author     Neil Lathwood <neil@lathwood.co.uk>
+ */
+
+if ($device['os'] === 'aos-emu2') {
+    foreach ($emu2_temp as $id => $temp) {
+        if (isset($temp['emsProbeStatusProbeTemperature']) && $temp['emsProbeStatusProbeTemperature'] > 0) {
+            $index           = $temp['emsProbeStatusProbeIndex'];
+            $oid             = '.1.3.6.1.4.1.318.1.1.10.3.13.1.1.3.' . $index;
+            $descr           = $temp['emsProbeStatusProbeName'];
+            $low_limit       = fahrenheit_to_celsius($emu2_temp_scale, $temp['emsProbeStatusProbeMinTempThresh']);
```
Rosiak: The device I tested against doesn't send any info back for:
emsProbeStatusProbeMaxTempThresh
emsProbeStatusProbeMinTempThresh

includes/discovery/sensors/humidity/aos-emu2.inc.php
laf: well spotted, updated now

includes/discovery/sensors/humidity/aos-emu2.inc.php
laf: does it not then just auto calculate?

Rosiak:
Disco: http://pastebin.com/2hgQtvVA
Poll: http://pastebin.com/8QCDH7sX

laf:
Forgot to push this last night. Added a function to set the default value to null based on the minimum we expect so if none of the thresholds are valid (i.e 0 according to your output) then we default to null so they are excluded.

The Fahrenheit was fixed in another commit but you were on the older one hence it didn't work for you.

laf:
bump

laf:
fixed rebase.

Rosiak:
Last thing:
Device I have access to does not respond on emsProbeStatusProbeMaxHumidityThresh or  emsProbeStatusProbeMinHumidityThresh

laf:
Thanks, I've moved set_null to inside discover_sensor now so any invalid high/low limits will be nulled out.

Rosiak:
Still an issue, you are aware that it was Humidity that was wrong, right?
Temp polling/warnings is perfectly fine :)

laf:
Yes but the call is now in discover_sensor() so cleans anything. Up what's actually wrong and where's the debug on the latest commit?

laf:
Can we look at merging this. The issue for @Rosiak is not from this PR, defaults for humidity sensors with no thresholds are 70/70

PR #5368, opened 2017-01-09
remove references to php-snmp it's not needed

laf:
DO NOT DELETE THIS TEXT

#### Please note

We don't use php-snmp so removed references to install and validate for it.


murrant:
Hmm, I'm writing an experimental php-snmp backend for the snmp re-write...  Might want to leave this in for a bit.

PR #5381, opened 2017-01-10
Update Installation-Ubuntu-1604-Nginx.md

BurntTech:
Remove default site link to avoid conflicts
rm /etc/nginx/sites-enabled/default

DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/doc/Installation/Installation-Ubuntu-1604-Nginx.md
+++ b/doc/Installation/Installation-Ubuntu-1604-Nginx.md
@@ -98,6 +98,7 @@ server {
 ```
 
 ```bash
+rm /etc/nginx/sites-enabled/default
 systemctl restart nginx
 ```
 


laf:
Is this the same for centos? If so can you update that doc as well

Gorian:
@laf last version of CentOS I used, nginx didn't have `/sites-{available,enabled}` by default, and instead used `/etc/nginx/conf.d`, I've always had to add the directory and config lines manually.

laf:
Sorry I don't mean that path specifically more of a default exists like it does for apache.

Gorian:
ah, okay. Yeah, should be able to add `rm /etc/nginx/conf.d/default` for CentOS I think

laf:
@BurntTech bump

Issue #5271, opened 2016-12-31
Unifi AP's not showing wireless data


TyShawn76:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

- [] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ ] Please include all of the information between the `====================================` section of `./validate.php`.
- [ ] If you would like us to add a new device or your device is listed as Generic then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)
- [ ] Please provide as much detail as possible.

New Box
==========================================================
Component | Version
--------- | -------
LibreNMS  | 11345e1c73f0ec3bde2fd62b85228cd4fdd4ac5f
DB Schema | 153
PHP       | 5.6.29-0+deb8u1
MySQL     | 10.0.28-MariaDB-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1
==========================================================

Old Box
==========================================================
Component | Version
--------- | -------
LibreNMS  | 11345e1c73f0ec3bde2fd62b85228cd4fdd4ac5f
DB Schema | 153
PHP       | 5.6.29-0+deb8u1
MySQL     | 5.5.53-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1
==========================================================

I have multiple Gen 1 and Gen 2 Unifi Aps over multiple locations. When I opened a ticket to have the new OID's added to the APS everything was working perfectly APs I noticed the network section never added for the APs. I didn't pay it much mind as I thought I was doing something wrong. A couple of weeks ago I loaded up a new test box and found the same issue, none of the APs would show the network section that housed the information I was looking for. 

Later I loaded a fresh longterm box that I had to run to a site and found the same issue again. None of the APS Gen 1 or Gen 2 will show the network section. Yet I can still poll the test AP without fail. 

I am including two pictures one with the working section and one without.
![screen shot 2016-12-31 at 10 30 28 am](https://cloud.githubusercontent.com/assets/21324034/21578179/ca605704-cf44-11e6-9629-625873731033.jpg)
[...]

PR #5383, opened 2017-01-10
fix: don't support unifi clients that don't report data

murrant:
better messages from the poller (for unifi at least)
Some cleanups in wifi.inc.php

DO NOT DELETE THIS TEXT

#### Please note

Addresses: #5271

### DIFF
--- a/includes/polling/mib/ubnt-unifi-mib.inc.php
+++ b/includes/polling/mib/ubnt-unifi-mib.inc.php
@@ -3,7 +3,7 @@
 // based on Airfiber MIB work of Mark Gibbons
 
 // UBNT-UniFi-MIB
-echo ' UBNT-UniFi-MIB ';
+echo 'UBNT-UniFi-MIB ';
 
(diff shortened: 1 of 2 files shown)


jquagga:
I like this.  Much prettier than what I glued together!  :).  

I tested this on my UAPs and they work just as before.  I don't have any Gen1's to test the "not supported" functionality though.  

One thing to put out there on the radar is that the newest UAPs (for example the UAP-AC-HD which is beta) have 3 radios instead of the two.  They do 2.4, 5Ghz and have a separate management radio.  I don't own one so I'm not sure how that's going to show up in snmp.  If it's separate I believe the existing code would consider clients connected to the management radio as being part of "na" since they aren't "ng".  Again, that's speculation as I don't have the device to test.

murrant:
The OIDs used for this count the number of clients per SSID and group them based on frequency class (ng or ac).

So, I expect this will continue to function fine with the new radios.

Issue #5254, opened 2016-12-27
Wireless Access Point -> Wireless Type?


arrmo:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

- [X ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ X] Please include all of the information between the `====================================` section of `./validate.php`.
- [ X] If you would like us to add a new device or your device is listed as Generic then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)
- [ X] Please provide as much detail as possible.

Hi,

I know I can manually change this (1-by-1), but thinking there should be a better way (and there likely is). For WAP's running, for example, OpenWRT, DD-WRT, ASUSWRT-Merlin, Tomato, etc. ... because they run Linux they show up as a device type of "server" (based on Operating System) - but perhaps based on Platform they could / should be categorized as a device type of "Wireless"?

Thoughts?

I'm fine changing this in the code to help out, just can't figure out what / where to change it ... :(.

Thanks!

laf:
If we have a way to detect these devices then we should - we'd need the output of the info asked for in point 3 above.

arrmo:
Sure, NP! Files attached - for the three commands (discovery, poller and snmpbulkwalk), for both ASUSWRT-Merlin and Tomato. I don't have DD-WRT (that router died), and OpenWRT (not using it) - but start with these two at least? And - perhaps use the output of distro?

Thanks!





[discovery.ASUSWRT-Merlin.txt](https://github.com/librenms/librenms/files/674980/discovery.ASUSWRT-Merlin.txt)
[discovery.Tomato.txt](https://github.com/librenms/librenms/files/674981/discovery.Tomato.txt)
[poller.ASUSWRT-Merlin.txt](https://github.com/librenms/librenms/files/674982/poller.ASUSWRT-Merlin.txt)
[poller.Tomato.txt](https://github.com/librenms/librenms/files/674985/poller.Tomato.txt)
[snmpbulkwalk.ASUSWRT-Merlin.txt](https://github.com/librenms/librenms/files/674983/snmpbulkwalk.ASUSWRT-Merlin.txt)
[...]

PR #5398, opened 2017-01-11
newdevice: Added Tomato and AsusWRT-Merlin OS #5254

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5254 

These work with distro script as well but as it's possible to detect I've done just that


murrant:
I know how to move the discovery to yaml, but it requires another change first...

PR #5400, opened 2017-01-11
feature: HPE Comware Processor Discovery + New Logo

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note


murrant:
Probably would be a good idea to optimize that svg with something like this: https://jakearchibald.github.io/svgomg/

includes/discovery/processors/comware.inc.php
```diff
@@ -0,0 +1,32 @@
...
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+if ($device['os'] == 'comware') {
+    echo 'Comware ';
+
+    $x = 1;
+    $entphydata = dbFetchRows("SELECT `entPhysicalIndex` FROM `entPhysical` WHERE `device_id` = ? AND `entPhysicalName` LIKE '%Board%' ORDER BY `entPhysicalIndex`", array($device['device_id']));
```
laf: This relies on entity-physical discovery module to be enabled. We don't normally have any cross over between modules - anyway we can do this without that module?

Rosiak: We could just poll entPhysical directly, but seems a bit silly.
Only other way is to poll h3cEntityExtMemSize(Yeah I know..) as I see it.

laf: I'd personally like to avoid any reliance on other modules as you can't guarantee they have run or ever will run.

Rosiak: I fully agree, I'll convert it to poll entPhysicalName instead.

PR #5401, opened 2017-01-11
refactor: move include based discovery after yaml discovery

murrant:
Fix edgeswitch detection to include rather than exclude
Fix zynos detection it requires both sysOjbectId and sysDescr to match
move zywall detection to yaml
remove linux exceptions as they aren't needed

DO NOT DELETE THIS TEXT

#### Please note


includes/discovery/os/edgeswitch.inc.php
```diff
@@ -1,5 +0,0 @@
-<?php
-
-if (starts_with($sysObjectId, '.1.3.6.1.4.1.4413') && !str_contains($sysDescr, array('vxworks', 'Quanta', 'FASTPATH Switching'), true)) {
```
laf: Think this has broken the detection, #5404

murrant: Edgeswitch == UnifiSwitch?

Issue #5391, opened 2017-01-11
Ghost (empty) toner discovered and alerted


tuxi56:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [X] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.

```
Component | Version
--------- | -------
LibreNMS  | e95772e5b481c9bffc42922afa0b6cf909197463
DB Schema | 153
PHP       | 7.0.14-1~dotdeb+8.1
MySQL     | 5.5.53-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1
```
Hi

For the same device described at https://github.com/librenms/librenms/issues/4404, we have some "ghost" (i.e. empty) extra toner discovered (oid .1.3.6.1.4.1.367.3.2.1.2.24.1.1.5**.5** -> we have already .1 to .4 for our regular toner cartridges, no _fifth_ color added recently). I guess this could be the toner waste collector, but it's value is always at 0 (and our collector is nor empty nor full). 

Here's an extract from the `toner` table in mysql : 

```
+----------+-----------+-------------+------------+-------------------------------------+---------------+----------------+---------------+--------------------+
| toner_id | device_id | toner_index | toner_type | toner_oid                           | toner_descr   | toner_capacity | toner_current | toner_capacity_oid |
+----------+-----------+-------------+------------+-------------------------------------+---------------+----------------+---------------+--------------------+
|       27 |        15 |           1 | jetdirect  | .1.3.6.1.4.1.367.3.2.1.2.24.1.1.5.1 | Toner black   |            100 |           100 |                    |
|       28 |        15 |           2 | jetdirect  | .1.3.6.1.4.1.367.3.2.1.2.24.1.1.5.2 | Toner cyan    |            100 |            50 |                    |
|       29 |        15 |           3 | jetdirect  | .1.3.6.1.4.1.367.3.2.1.2.24.1.1.5.3 | Toner magenta |            100 |            30 |                    |
[...]

PR #5410, opened 2017-01-12
fix: Fix toner detection discovery #5391

laf:
DO NOT DELETE THIS TEXT

#### Please note

I think this is right.

Fixes: #5391 

Before hand duplicates were detected by capacity_oid which could be null, switched to oid as that will be unique (or should be).

The old if `($os` check was fundamentally wrong as $os is not defined.

The idea being now we use the standard Printer-MIB details and fall back if they are empty which in some situations they are. We also get the internationalised text to use for descr if it's available.

### DIFF
--- a/includes/discovery/toner.inc.php
+++ b/includes/discovery/toner.inc.php
@@ -18,19 +18,30 @@ if ($device['os_group'] == 'printer') {
 
     foreach ($oids as $index => $data) {
         $last_index = substr($index, strrpos($index, '.') + 1);
-        if ($device['os'] == 'ricoh' || $device['os'] == 'nrg' || $device['os'] == 'lanier') {
+        
+        $raw_toner     = $data['prtMarkerSuppliesLevel'];
+        $descr         = $data['prtMarkerSuppliesDescription'];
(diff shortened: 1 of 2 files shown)


includes/discovery/toner.inc.php
```diff
@@ -54,11 +66,12 @@
 d_echo("\n Checking valid toner ... \n");
 d_echo($valid_toner);
 
-$sql = "SELECT * FROM toner WHERE device_id = '" . $device['device_id'] . "'";
-foreach (dbFetchRows($sql) as $test_toner) {
-    $toner_index = $test_toner['toner_index'];
+$sql = dbFetchRows("SELECT * FROM toner WHERE device_id = '" . $device['device_id'] . "'");
```
murrant: $sql variable name is a bit odd

laf: Done

murrant:
Worked on every printer I could test.  But I don't have a wide selection.

jquagga:
This works on my troublesome Samsung ML series as well.  That's all I have to test.

PR #5411, opened 2017-01-12
newdevice: Add extra Sophos discovery

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

New firmware..


tests/snmpsim/sophos4.snmprec
```diff
@@ -0,0 +1 @@
+1.3.6.1.2.1.1.1.0|4|Linux g4dc0e52
```
murrant: Is the actually sysDescr from the device?

Rosiak: Nope, but don't have other than that right now :(

laf: We could look at using `&& preg_match('/Linux ([\S]+) ([0-9\-.]{1,11})(g[\w\d]{7})(.*)/', $sysDescr)`

That matches the test data I found from an old issue: https://gist.github.com/laf/67cee069eb3db4af26dccdb67a7da3f5 but we'd need to find others potentially.

murrant: Thanks for the regex @laf  I was thinking something like this but gave me a good head start :)

laf: Np :)

PR #5413, opened 2017-01-12
newdevice: Add support for KTI switches

corny:
DO NOT DELETE THIS TEXT

#### Please note


tests/snmpsim/kti.snmprec
```diff
@@ -0,0 +1 @@
+1.3.6.1.2.1.1.1.0|4|KGS-1060 GigaBit Ethernet Switch
```
murrant: We need both sysDescr and sysObjectId

includes/polling/os/kti.inc.php
```diff
@@ -0,0 +1,4 @@
+<?php
+
+$hardware     = trim(snmp_get($device, ".1.3.6.1.2.1.1.1.0", '-Ovq'), '"');
```
murrant: sysDescr is already fetched, you can use `$poll_device['sysDescr']`

tests/snmpsim/kti.snmprec
laf: It's also worth switching the yaml discover to use sysObjectId if it's useful. Relying on sysDescr matches should be the last resort.

corny: The sysObjectId is not useful:
```
$ snmpbulkwalk -OntQX -v2c -c public $host 1.3.6.1.2.1.1
.1.3.6.1.2.1.1.1.0 = "KGS-1060 GigaBit Ethernet Switch"
.1.3.6.1.2.1.1.2.0 = .0.0
```

laf: Shoot that vendor

includes/polling/os/kti.inc.php
```diff
@@ -0,0 +1,4 @@
+<?php
+
+$hardware     = trim(snmp_get($poll_device['sysDescr'], ".1.3.6.1.2.1.1.1.0", '-Ovq'), '"');
```
laf: What murrant meant here was that $poll_device['sysDescr'] already contains the output you want so you don't need another snmp call just $hardware = $poll_device['sysDescr']

laf:
I've pushed the update.

Issue #5412, opened 2017-01-12
device type not being set for any new devices


sysmonk:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.

==========================================================
Component | Version
--------- | -------
LibreNMS  | 78cdca022290f53fce4c2d541d11ecbc2e4c1bde
DB Schema | 153
PHP       | 7.0.14-1~dotdeb+8.1
MySQL     | 10.0.28-MariaDB-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1
==========================================================

Librenms stopped setting the device type for new devices added recently. This sounds related to https://github.com/librenms/librenms/blob/master/includes/common.php#L1529 - it's sets the type (and makes any later checks not see that it changed), but does not update the db.

laf:
The type setting isnt done there. It's done here. https://github.com/librenms/librenms/blob/master/includes/discovery/os.inc.php

Please post output onto pastebin for ./discovery.php -h HOSTNAME -d -m os

laf:
Actually don't need this, confirmed, fix being done.

PR #5414, opened 2017-01-12
fix: Detection blank or unknown device types and update #5412

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5412

### DIFF
--- a/includes/common.php
+++ b/includes/common.php
@@ -1527,9 +1527,12 @@ function load_os(&$device)
     );
 
     // Set type to a predefined type for the OS if it's not already set
-    if ($device['type'] == 'unknown' || $device['type'] == '') {
+    if ($device['type'] == 'unknown' || $device['type'] == '' || $config['os'][$device['os']]['type'] != $device['type']) {
         if ($config['os'][$device['os']]['type']) {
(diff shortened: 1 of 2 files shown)


includes/common.php
```diff
@@ -1527,9 +1527,12 @@ function load_os(&$device)
     );
 
     // Set type to a predefined type for the OS if it's not already set
-    if ($device['type'] == 'unknown' || $device['type'] == '') {
+    if ($device['type'] == 'unknown' || $device['type'] == '' || $config['os'][$device['os']]['type'] != $device['type']) {
```
murrant: It feels like this should be, I'm not somewhere where I can test it right now...
```php
    // Set type to a predefined type for the OS if it's not already set
    if ($config['os'][$device['os']]['type'] && $device['type'] != $config['os'][$device['os']]['type']) {
        $device['type'] = $config['os'][$device['os']]['type'];
        $sql = dbUpdate(array('type' => $device['type']), 'devices', 'device_id=?', array($device['device_id']));
        log_event('Device type changed '.$device['type'].' => '.$config['os'][$device['os']]['type'], $device, 'system');
        echo "Changed Type! : ".$device['type'].PHP_EOL;
    }
```

Issue #5405, opened 2017-01-12
sensors module not adding dbm for juniper


Izemaen:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ ] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.

==========================================================
Component | Version
--------- | -------
LibreNMS  | c5a3d82040083e2f366da29958ec8a6b62ba7bbb
DB Schema | 153
PHP       | 5.6.29-0+deb8u1
MySQL     | 5.5.53-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1
==========================================================

When the discovery module tries to discover sensors on juniper devices it fails to do some snmp walks/gets e.g. for precaching OID due to a problem with mibs and/or mib directories.

http://paste.debian.net/908395/

You can replicate that by running the snmpwalk manually

http://paste.debian.net/908400/

Once you switch the order of the mib directories used in the snmp command (/mibs first than /mibs/junos, it works and no longer gives snmp errors

http://paste.debian.net/908401/

NOTE: The actual query from the pastes above still fails because i did not change Hostname and Community into the correct values, but when i do it also returns the snmp values.

The issue seems to be there for other snmp queries in our installation as well, but for us it breaks juniper dbm sensor detection.

For our installation i made it work by changing stuff in snmp.inc.php to change the order of mib Dirs used in the queries:

http://paste.debian.net/908402/

Afterwards none of our systems complains about broken mibs anymore and more important dbm sensors got added for the juniper devices.

I have no idea wheter this is general problem or specific to our install. I was able to replicate the issue on two Debian 8.6 systems and the error is device independent as snmp itselfs fails to load the mibs before even contacting the device.

laf:
Thanks for all the info :)

[...]

PR #5415, opened 2017-01-12
fix: Re-order snmp mib dirs to load mibs correctly #5405

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5405

### DIFF
--- a/includes/snmp.inc.php
+++ b/includes/snmp.inc.php
@@ -85,7 +85,7 @@ function get_mib_dir($device)
 function mibdir($mibdir = null, $device = array())
 {
     global $config;
-    // FIXME: prepend + to allow system mibs?
+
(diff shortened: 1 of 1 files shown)


includes/snmp.inc.php
```diff
@@ -102,10 +101,10 @@ function mibdir($mibdir = null, $device = array())
 
     if (str_contains($mibdir, '/')) {
         // pass through mib dir (for legace compatability
-        return " -M $extra_dir$mibdir";
+        return " -M $mibdir$extra_dir";
```
murrant: This one seems backwards.

murrant: But I think this case can be removed now.

Issue #5387, opened 2017-01-10
Adding more Mikrotik environment OIDs for graphing


Cheeze-It:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [ x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ x] Please include all of the information between the 

Component | Version
--------- | -------
LibreNMS  | e95772e5b481c9bffc42922afa0b6cf909197463
DB Schema | 153
PHP       | 5.5.9-1ubuntu4.20
MySQL     | 5.5.53-0ubuntu0.14.04.1
RRDTool   | 1.4.7
SNMP      | NET-SNMP 5.7.2

- [ x] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ x] Please provide as much detail as possible.

I would like to put in a request for an update to Mikrotik graphs. Currently the system tempurature (which is usually a motherboard temp) is listed as .1.3.6.1.4.1.14988.1.1.3.10.0. Can we also please get another graph for the following temps as well?

[<user>@<device>] /system health> print 
fan-mode: auto
use-fan: main
active-fan: main
cpu-overtemp-check: yes
cpu-overtemp-threshold: 100C
cpu-overtemp-startup-delay: 1m
voltage: 24V
current: 1626mA
temperature: 40C
cpu-temperature: 53C
power-consumption: 39W
fan1-speed: 0RPM
fan2-speed: 4416RPM

[<user>@<device>] /system health> print oid 
active-fan: .1.3.6.1.4.1.14988.1.1.3.9.0
voltage: .1.3.6.1.4.1.14988.1.1.3.8.0   <--- Currently implemented
temperature: .1.3.6.1.4.1.14988.1.1.3.10.0   <--- Currently implemented
processor-temperature: .1.3.6.1.4.1.14988.1.1.3.11.0   <--- Can we please implement?
current: .1.3.6.1.4.1.14988.1.1.3.13.0   <--- Can we please implement?
[...]

PR #5419, opened 2017-01-13
newdevice: Add more sensor detection for Mikrotik devices #5387

geordish:
Added more sensors for Mikrotik. (#5387)

Could do with some testing, as my Mikrotik devices don't have these sensors!

DO NOT DELETE THIS TEXT

#### Please note


includes/discovery/sensors/fanspeeds/mikrotik.inc.php
```diff
@@ -1,28 +1,19 @@
...
-            $index            = $split_oid[(count($split_oid) - 1)];
-            $descr            = 'Fan '.$index;
-            $oid              = '.1.3.6.1.4.1.14988.1.1.3.9.'.$index;
-            $fanspeed         = (snmp_get($device, $oid, '-Oqv') / $divisor);
-            if ($fanspeed > 0) {
-                discover_sensor($valid['sensor'], 'fanspeed', $device, $oid, $index, $type, $descr, $divisor, '1', null, null, null, null, $fanspeed);
-            }
+        if (is_numeric($value)) {
+            $descr = $descr_prefix . ($index + 1);
+            discover_sensor($valid['sensor'], 'fanspeed', $device, $oid, $index, 'snmp', $descr, 1, 1, null, null, null, null, $value);
```
laf: The change in oid here will cause old sensors to be deleted and new added. What's the reason for the change?

geordish: The old OID that was being pulled doesn't return fan speed.
snmptranslate -On MIKROTIK-MIB::mtxrHlActiveFan
.1.3.6.1.4.1.14988.1.1.3.9

From the issue:
iso.3.6.1.4.1.14988.1.1.3.9.0 = STRING: "

On my Mikrotik:
.1.3.6.1.4.1.14988.1.1.3.9.0 = STRING: "n/a"

From the MIB (its a string):
mtxrHlActiveFan OBJECT-TYPE
    SYNTAX DisplayString
    MAX-ACCESS read-only
    STATUS current
    DESCRIPTION ""
    ::= { mtxrHealth 9 }

The 2 OIDs now polled are the correct ones.

I wonder if their MIB changed at some point.

laf: Thanks for the clarification :)

Cheeze-It: The old OID actually asks for which fan "stage" is actually being in use, whereas the new OIDs for fans actually give the RPM in an INTEGER format.

Old was a STRING format that just said "main" or "backup" in regards to which fan stage was actually being used. Useful as well but not as useful as the RPM ratings. I believe other devices also did the same as well that have supported this.

laf: Strings won't have been discovered so we should be good now.

PR #5433, opened 2017-01-14
Update samsungprinter icon to use svg from Wikimedia Commons

jquagga:
Good day,

With the move to svg icons, I've been trying to find a permissible, better icon for my Samsung Printer (which looks pretty bad).  I think I've found one.  It's in a CC BY-SA 4.0 licensed icon set here on github. https://github.com/alecive/FlatWoken/blob/master/FlatWoken/scalable/apps/samsung.svg is the actual file. 

I've noted the link and license in the commit message.  If I should update General/Credits.md I can do that as well.  Just let me know.  

DO NOT DELETE THIS TEXT

#### Please note


laf:
Why not https://commons.m.wikimedia.org/wiki/File:Samsung_Logo.svg

jquagga:
I think the only real negative there is that that icon isn't square so on the main page, it dislocates the title bar a bit.  The scribe.internal is pushed over a bit.  

And it's a bit smushed on the full devices page.  But it's still an improvement over the icon which is there (which is fine, just low resolution so it becomes pixelated on many screens).  

If you'd like me to swap it out in this PR, I can certainly do that.  

screenshot2-fs8
screenshot-fs8

jquagga:
I went ahead and swapped it out anyway.  It grew on me.

Issue #5181, opened 2016-12-16
Add support for temperature sensor graph Smart Edge 600


kyob:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

- [ ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ ] Please include all of the information between the `====================================` section of `./validate.php`.
- [ ] If you would like us to add a new device or your device is listed as Generic then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)
- [ ] Please provide as much detail as possible.

Component | Version
--------- | -------
LibreNMS  | 969b963180ea42272c5b9c40851c683daf1c6337
DB Schema | 152
PHP       | 7.0.13
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2

---
**Temperature sensor graph for Smart Edge 600 (Ericsson)**

OID: 1.3.6.1.4.1.2352.2.4.1.4.1

```bash
snmpbulkwalk -OUneb -v2c -c COMMUNITY HOST .1.3.6.1.4.1.2352.2.4.1.4.1
.1.3.6.1.4.1.2352.2.4.1.4.1.2.13.1 = STRING: "Temperature sensor on slot 1"
.1.3.6.1.4.1.2352.2.4.1.4.1.2.18.1 = STRING: "Temperature sensor on slot 2"
.1.3.6.1.4.1.2352.2.4.1.4.1.2.29.1 = STRING: "Temperature sensor on slot 8 - Processor"
.1.3.6.1.4.1.2352.2.4.1.4.1.2.29.2 = STRING: "Temperature sensor on slot 8 - Exhaust Air"
.1.3.6.1.4.1.2352.2.4.1.4.1.2.29.3 = STRING: "Temperature sensor on slot 8 - Inlet Air"
.1.3.6.1.4.1.2352.2.4.1.4.1.3.13.1 = Gauge32: 0
.1.3.6.1.4.1.2352.2.4.1.4.1.3.18.1 = Gauge32: 38
[...]

PR #5441, opened 2017-01-15
Changed Redback to SEOS, and added logo and temperature discovery

geordish:
DO NOT DELETE THIS TEXT

Hopefully will address issue #5181

#### Please note


laf:
Couple of things with this.

1. Will need an announcement going out to let people know we're changing the os in case people use it for grouping / alerting: https://github.com/librenms/librenms.github.io/blob/master/notifications.rss
Pick a date a few days ahead.

2. Can you also add sysObjectId to the test file, even though it's not used for detection we should have it.

3. Can we not use sysObjectId for detection? It's more reliable if we can.

4. Can you put the mibs into a sub dir like (ericsson) and update the yaml file.

laf:
Notification has gone out, merge on the 19th of Jan.

PR #5444, opened 2017-01-15
Added check for if git exists

geordish:
Should hopefully not stop validate running if git is missing.

DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/validate.php
+++ b/validate.php
@@ -108,7 +108,7 @@ if (!($username === 'root' || (isset($config['user']) && $username === $config['
     print_fail('You need to run this script as root' . (isset($config['user']) ? ' or '.$config['user'] : ''));
 }
 
-if ($git_found == true) {
+if ($git_found === true) {
     if ($config['update_channel'] == 'master' && $cur_sha != $versions['github']['sha']) {
         $commit_date = new DateTime('@'.$versions['local_date'], new DateTimeZone(date_default_timezone_get()));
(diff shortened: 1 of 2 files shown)


validate.php
```diff
@@ -75,6 +75,11 @@
     exit;
 }
 
+$git_found = check_git_exists();
+if ($git_found !== true) {
```
Gorian: could you use `if (!$git_found){` instead?

validate.php
```diff
@@ -103,9 +108,11 @@
     print_fail('You need to run this script as root' . (isset($config['user']) ? ' or '.$config['user'] : ''));
 }
 
-if ($config['update_channel'] == 'master' && $cur_sha != $versions['github']['sha']) {
-    $commit_date = new DateTime('@'.$versions['local_date'], new DateTimeZone(date_default_timezone_get()));
-    print_warn("Your install is out of date, last update: " . $commit_date->format('r'));
+if ($git_found === true) {
```
Gorian: could you use `if ($git_found){` here?

Gorian:
I think that @mention-bot thing is really cool. Look at who last contributed to a file and then mention them to review new changes.

validate.php
laf: === is better, it's more explicit

Gorian: Alright, as long as we are doing it purposefully, you are the master of the php code ;)

Issue #5432, opened 2017-01-14
NWA3560-N & NWA3160-N is detected as firewall should be wireless.


msjohansen:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date?
- [x] Please include all of the information of `./validate.php`.
- [x] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.

[discovery.txt](https://github.com/librenms/librenms/files/706002/discovery.txt)
[poller.txt](https://github.com/librenms/librenms/files/706000/poller.txt)
[snmpbulkwalk.txt](https://github.com/librenms/librenms/files/706001/snmpbulkwalk.txt)

Rosiak:
Seems like this is detected as a zywall rather than zyxelnwa due to the sysObjectID.

PR #5450, opened 2017-01-15
update zywall and zyxelnwa detection #5343

laf:
DO NOT DELETE THIS TEXT

#### Please note

Got test data for a Zyxel NWA unit so updated to have this detected correctly.

Fixes: #5432


includes/definitions/zywall.yaml
```diff
@@ -6,7 +6,10 @@ over:
     - { graph: device_bits, text: Traffic }
 icon: zyxel
 discovery:
-    - sysObjectId:
-        - .1.3.6.1.4.1.890.1.6
-        - .1.3.6.1.4.1.890.1.15
+    -
+      sysObjectId:
+          - .1.3.6.1.4.1.890.1.6
+          - .1.3.6.1.4.1.890.1.15
```
murrant: My thought was this:
```
   -
      sysObjectId:
          - .1.3.6.1.4.1.890.1.6
          - .1.3.6.1.4.1.890.1.15
      sysDescr_regex:
          - '/^ZyWALL/
```

I'm not a fan of negative matches, they make a big mess.

includes/definitions/zyxelnwa.yaml
```diff
@@ -5,4 +5,4 @@ type: network
 icon: zyxel
 discovery:
     - sysDescr_regex:
-        - '/^NWA-/'
+        - '/^NWA/'
```
murrant: Can we combine this with a sysObjectId?

laf:
I've updated them but I honestly don't think we have enough test data to do it this way and could just lead to more broken discovery.

murrant:
@laf Possibly, any way you could put together that callback for sysDescr and sysObjectId?  I won't have any hacking time till maybe tonight and I'll have to familiarize myself with the callback stuff.

laf:
@murrant callback?

murrant:
@laf the data collection thingy in `includes/callback.php` :)

laf:
Is this waiting for us to get more test data?

murrant:
Yeah, but we can probably merge.  I'll go through them when we have data.

PR #5459, opened 2017-01-16
various small doc improvements

laf:
DO NOT DELETE THIS TEXT

#### Please note


murrant:
not sure why the -m os was in there...  messed me up on that other ticket I think.

laf:
Legacy, that faq entry is for adding OS support so all we really used to care about is sysObjectId and sysDescr which was enough

PR #5465, opened 2017-01-16
Updated icons to use Font Awesome

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

- [X ] Have you signed the [Contributors agreement](http://docs.librenms.org/General/Contributing/) - please do NOT submit a pull request unless you have (signing the agreement in the same pull request is fine). Your commit message for signing the agreement must appear as per the docs.

I have tried to match the current fa icons to the ones in place to continue with a standard, but maybe in the future these can be reviewed.


laf:
The code formatting needs some small updates: https://travis-ci.org/librenms/librenms/jobs/192382723 (you can run ./scripts/pre-commit.php in your install to do the same tests before submitting).

I'm not 100% sure the blue looks right, needs toning down a bit.

Aside from that, I like it :)

InsaneSplash:
No problem, let me work on that a bit more. This was the first stab and moving away from the png files so there is still work to be done :)..

PR #5468, opened 2017-01-16
Updated icons to use Font Awesome

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note


murrant:
Like this overall, but some of the replacements are a little too generic.  Like specifically memory->cog.  I think it is ok to keep the sensor icons as regular images for now.

html/includes/table/devices.inc.php
```diff
@@ -203,11 +203,11 @@
             $col_port = '';
         }
         if ($port_count) {
-            $col_port = '<img src="images/icons/port.png" align="absmiddle"> ' . $port_count . '<br>';
+            $col_port = '<i class="fa fa-link" style="color:#0080FF"></i> ' . $port_count . '<br>';
```
Rosiak: Not a fan of the color of this.

html/includes/table/devices.inc.php
```diff
@@ -203,11 +203,11 @@
...
         }
 
         if ($sensor_count) {
-            $col_port .= '<img src="images/icons/sensors.png" align="absmiddle"> ' . $sensor_count;
+            $col_port .= '<i class="fa fa-heartbeat" style="color:#0080FF"></i> ' . $sensor_count;
```
Rosiak: Same goes here.

html/includes/table/devices.inc.php
InsaneSplash: Its easy enough to change... I tried to match the blue icon that was originally used.

html/includes/table/devices.inc.php
InsaneSplash: Same as above.

InsaneSplash:
I tried to match the current icons that were used for fa in the menus etc, but I agree that it needs some refinement and standardisation. This is something I was going to look in to once I received feedback that this change was interesting enough to pursue.

murrant:
Honestly, we could do with some auditing of icons to make sure they are all consistent and meaningful.  It might even require some icon creation.

InsaneSplash:
Agreed. Would you like me to give it a go and see how it turns out?

murrant:
Sure, but if I were you I'd break it up into manageable steps.

InsaneSplash:
This commit is going to need the latest version of Font Awesome (4.7.0).

laf:
We used to use icons everywhere and have moved a lot to FA.

I'm not personally a fan of using images as icons, it's just extra bloat we don't need. I'm happy with a full switch to FA.

laf:
Looks good to me. @librenms/reviewers

html/pages/device/overview/generic/sensor.inc.php
```diff
@@ -6,12 +6,60 @@
 }
 
 if (count($sensors)) {
+    switch (strtolower($sensor_type)) {
```
murrant: Might want to put this into a function.  There are other places where sensor icons are used, such as the about page.

includes/defaults.inc.php
```diff
@@ -844,6 +844,7 @@
 
 // General GUI options
 $config['gui']['network-map']['style']                  = 'new';//old is also valid
+$config['theme_icon_colour']                            = 'black'; // Font Awesome icon colour customisation
```
murrant: This feels odd.  Shouldn't color be mostly dependent on the spot where they are used?
Also, shouldn't this be in CSS, no PHP?

murrant:
Overall, the I like the icon selection 👍 .

There are few spots where color could help out quite a bit.

- The color of the icons on the device overview tabs feels off, I'm not sure if it should match the text or somethings else or if it is fine as is.
- I would really like to seem some colors in the device actions (table view and device overview) to make it easier to identify individual actions.

Those things could be worked out later.  But I do think the default color setting should be moved to css.

InsaneSplash:
Thanks, it's a start in the right direction. I thought about the CSS being
the place to set the icon colour but as users don't generally edit anything
in there it would be easier for them to change the colour of their icon
pallate through a config file than try their hand at the css file which
would probably be replaced through updating.

I also set all the icons to the same colour to try set a standard look. One
has to set a colour as the link style makes them light blue.

I think it will evolve over time but ultimately you guys are in charge :)

InsaneSplash:
cool! 
I'll do a version with the larger icons for you and move the icon colour to the css :)

Lets see what you think?

murrant:
@InsaneSplash `grep -r png .` :-D

InsaneSplash:
eheh yeah, but some need little more investigating :)

_Im done with this PR_. Will review once this has been merged and create a new PR with any changes etc.

laf:
I've squashed this down and rebased for you, contains the agreement message still which is the most important. Just waiting for checks to pass and will merge.

PR #5476, opened 2017-01-16
webui: Update Font Awesome to 4.7.0

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note


laf:
Are we duplicating files here? Before FA was loaded into lib/ which seems to the case still but all the symlinks from html/* back to lib/fa/* are now replaced with the full files as well.

InsaneSplash:
Do we not just need the 2 CSS files, or do you require the whole package to be part of the repo?

Rosiak:
We briefly had a discussion on IRC regarding this earlier today.
Should we remove the lib completely and just keep the necessary files?

InsaneSplash:
I'd say why include stuff that's not required...it would de-bloat the installation and reduce its size a little if the whole package wasn't included. Also make for easier updating in the future.

laf:
Remove.

PR #5484, opened 2017-01-17
Modify xerox.yaml for DocuPrint M225

wackychocolatefactory:
DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/includes/definitions/xerox.yaml
+++ b/includes/definitions/xerox.yaml
@@ -10,3 +10,4 @@ discovery:
         - Xerox Phaser
         - Xerox WorkCentre
         - FUJI XEROX DocuPrint
+    - sysDescr_regex: '/^.*DocuPrint.*/'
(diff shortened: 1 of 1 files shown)


laf:
Thanks for this, we also need a unit test as well to cover the example.

https://github.com/librenms/librenms/blob/master/tests/snmpsim/xerox-docuprint.snmprec
https://github.com/librenms/librenms/blob/master/tests/OSDiscoveryTest.php#L1494

wackychocolatefactory:
Sure thing, I'll write them up tomorrow at work. :)

wackychocolatefactory:
Curious, does this one need a new unit test, wouldn't the existing unit test cover the new regex?

laf:
It possibly will but it's more the real world data we want. If detection was changed in the future we would only be able to ensure it matches what we have now and could break things.

includes/definitions/xerox.yaml
```diff
@@ -10,3 +10,4 @@ discovery:
         - Xerox Phaser
         - Xerox WorkCentre
         - FUJI XEROX DocuPrint
+    - sysDescr_regex: '/^.*DocuPrint.*/'
```
murrant: This is actually the same as `- sysDescr: DocuPrint`  the sysDescr section searches for the given string anywhere in the sysDescr

PR #5495, opened 2017-01-17
WebUI: Allows one to view a map of the SNMP location set for a device

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/html/includes/dev-overview-data.inc.php
+++ b/html/includes/dev-overview-data.inc.php
@@ -82,7 +82,7 @@ if ($device['sysContact']) {
 if ($device['location']) {
     echo '<tr>
         <td>Location</td>
-        <td>'.$device['location'].'&nbsp;&nbsp;<a href="http://maps.google.com/?q='.$device['location'].'" target="_blank"><button type="button" class="btn btn-success btn-xs"><i class="fa fa-map-marker" style="color:white" aria-hidden="true"></i> Map</button></a></td>
+        <td>'.$device['location'].'</td>
       </tr>';
     if (get_dev_attrib($device, 'override_sysLocation_bool') && !empty($device['real_location'])) {
         echo '<tr>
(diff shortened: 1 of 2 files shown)


laf:
We translate location to lat/lng, it's probably worth checking we have those values (locations table) and if so then display the link using those coords.

Is it also worth changing 'Map' for a globe icon?

InsaneSplash:
I did wonder about that but as google maps would be searching for the same location, as you have done to translate it to lat/lat, I thought that it would be less complex to merely use the same information provided. It would also help in situations where the user specified a location that is not that accurate, such as a city rather than an exact street address. 

I do have a question though, is the lat/long important if the user is not entering it themselves? Maybe in the future we could give the the user a way to specify the lat/long rather using the OpenStreetMap or GoogleMap API to obtain a more accurate value?

I am more than happy to change the way it works based if there is a lat/longresult from the db rather.

I decided on using the map-pin in the end as the globe is already used in the other proposed PR as a http link to the device and so tried to find something less confusing. Its also the same icon used for Locations in our main menu.

InsaneSplash:
I guess one could also move the link to the device settings drop down with SSH, Telnet, Web, but its current location makes it more accessible.

html/includes/dev-overview-data.inc.php
```diff
@@ -99,7 +99,7 @@
 if (is_array($loc)) {
     echo '<tr>
         <td>Lat / Lng</td>
-        <td>['.$loc['lat'].','.$loc['lng'].']</td>
+        <td>['.$loc['lat'].','.$loc['lng'].'] &nbsp;&nbsp;<a href="http://maps.google.com/?q='.$loc['lat'].'+'.$loc['lng'].'" target="_blank"><div class="pull-right"><button type="button" class="btn btn-success btn-xs"><i class="fa fa-map-marker" style="color:white" aria-hidden="true"></i> Map</button></div></a></td>
```
laf: You can merge the a href + button into one:

http://getbootstrap.com/css/#anchor-element

I'd also update the maps.google.com to be https.

laf:
This pr changes the default behaviour and enables lat/lng lookups by default. I'm 100% ok with this. @librenms/reviewers ? We probably should send a notification out.

laf:
Tagging to merge in 48 hours - @librenms/reviewers

spinza:
Not sure what this supposed to do?

Anyway we have locations as strings such as `something [38.123,17.12312]`.  Are these handled?

laf:
If you see "Lat / Lng" in your overview then yes a map icon will appear with this update.

Issue #3770, opened 2016-07-04
Nagios check_snmp and mikrotik signal graphs


wizzycom:
Hello.

I have configured some services for polling signal strength via snmp from a mikrotik device. Everything works as expected (command results and alerting) but I am facing some issues with graphing.

Here are the results of **check-services.php -d** : 

DEBUG!

SQL[SELECT \* FROM `devices` AS D, `services` AS S WHERE S.device_id = D.device_id ORDER by D.device_id DESC] 
Nagios Service - 1
Request:  /usr/lib64/nagios/plugins/check_snmp -H 10.70.25.234 -C public -o 1.3.6.1.4.1.14988.1.1.1.2.1.3.0.12.66.102.140.141.6 -c -80: 
Perf Data - None.
Perf Data - DS: iso.3.6.1.4.1.14988.1.1.1.2.1.3.0.12.66.102.140.141.6, Value: -65, UOM: 
Response: SNMP OK - -65 
Service DS: {
    "iso.3.6.1.4.1.14988.1.1.1.2.1.3.0.12.66.102.140.141.6": ""
}
RRD[create /opt/librenms/rrd/cronus.lan.local/services-1.rrd  DS:iso.3.6.1.4.1.14988.1.1.1.2.1.3.0.12.66.102.140.141.6:GAUGE:600:0:U RRA:AVERAGE:0.5:1:2016 RRA:AVERAGE:0.5:6:1440 RRA:AVERAGE:0.5:24:1440 RRA:AVERAGE:0.5:288:1440  RRA:MIN:0.5:1:720 RRA:MIN:0.5:6:1440     RRA:MIN:0.5:24:775     RRA:MIN:0.5:288:797  RRA:MAX:0.5:1:720 RRA:MAX:0.5:6:1440     RRA:MAX:0.5:24:775     RRA:MAX:0.5:288:797  RRA:LAST:0.5:1:1440 ] **ERROR: Invalid DS name**
RRD[update /opt/librenms/rrd/cronus.lan.local/services-1.rrd N:-65] 
Nagios Service - 2
Request:  /usr/lib64/nagios/plugins/check_snmp -H 10.70.25.234 -C public -o .1.3.6.1.4.1.14988.1.1.1.2.1.3.212.202.109.18.54.105.7 -c -80:
Perf Data - None.
Perf Data - DS: iso.3.6.1.4.1.14988.1.1.1.2.1.3.212.202.109.18.54.105.7, Value: -60, UOM: 
[...]

Issue #1232, opened 2015-06-10
Mikrotik Wireless Interface Data


spinza:
Looking to capture more detail on wireless info from Mikrotik SNMP.  I was hoping detailed CCQ is available in Mikrotik SNMP but I see just the overall CCQ is available on each wireless interface.  We really want the CCQ on each registration table entry but that's not available in Mikrotik SNMP info.

I dump the info below.  Happy to do it.  Would need some guidance on what I need to do.

I imagine first is to start with wireless interface info only.

Wireless interfaces

```
/interface wireless print                   
Flags: X - disabled, R - running 
 0  R name="wlan" mtu=1500 mac-address=XX:XX:XX:XX:XX:XX arp=enabled 
      interface-type=Atheros AR9300 mode=ap-bridge ssid="xxx" 
      frequency=2412 band=2ghz-b/g/n channel-width=20/40mhz-ht-above 
      scan-list=default wireless-protocol=any wds-mode=disabled 
      wds-default-bridge=none wds-ignore-ssid=no bridge-mode=enabled 
      default-authentication=yes default-forwarding=yes default-ap-tx-limit=0 
      default-client-tx-limit=0 hide-ssid=no security-profile=default 
      compression=no 
```

OID Information 

```
/interface wireless print oid
 0 tx-rate=.1.3.6.1.4.1.14988.1.1.1.3.1.2.12 
   rx-rate=.1.3.6.1.4.1.14988.1.1.1.3.1.3.12 
   ssid=.1.3.6.1.4.1.14988.1.1.1.3.1.4.12 
   bssid=.1.3.6.1.4.1.14988.1.1.1.3.1.5.12 
   client-count=.1.3.6.1.4.1.14988.1.1.1.3.1.6.12 
   frequency=.1.3.6.1.4.1.14988.1.1.1.3.1.7.12 
   band=.1.3.6.1.4.1.14988.1.1.1.3.1.8.12 
   noise-floor=.1.3.6.1.4.1.14988.1.1.1.3.1.9.12 
   overall-ccq=.1.3.6.1.4.1.14988.1.1.1.3.1.10.12
```

Wireless registration table (this is a list of clients). 

```
/interface wireless registration-table print
 # INTERFACE           RADIO-NAME       MAC-ADDRESS       AP  SIGNAL... TX-RATE
 0 wlan                                 xx:xx:xx:xx:xx:xx no  -51dBm... 48.0...
[...]

PR #5498, opened 2017-01-18
newdevice: Added signal support for RouterOS

laf:
DO NOT DELETE THIS TEXT

#### Please note

Renamed existing mikrotik files to routeros.

Fixes: #3770


geordish:
Not quite working. The sensor is detected correctly, but values are not being stored.

Could be due to the device returning -ve values? 

I would imagine either the RRD creation needs updating to accept a -ve number as a min, or setting the division to -1.

laf:
Odd as negative values are used in quite a few places, this is the test that's been done and works fine on negative values.:

image

spinza:
Will test on a large install of RouterOS devices.  Any chance of catching CCQ at the same time?

laf:
What's CCQ?

wizzycom:
It is a link quality counter...

OIDs for the interfaces we worked for the signaling
.1.3.6.1.4.1.14988.1.1.1.3.1.10.7
.1.3.6.1.4.1.14988.1.1.1.3.1.10.6

I am just trying to understand how you accomplished the signal sensor (not a programmer myself), because there are too many things that can be graphed and I need to take my chances to graph them

laf:
https://github.com/laf/librenms/blob/0c3e32d2d9f41f10d03f1bb6e86332562f359022/includes/discovery/sensors/signal/routeros.inc.php

I just walked the OID for the signal strength, the rest of the code is kind of standard and we just register the sensor which then takes the leg work out of it.

So for CCQ we don't have a sensory type that would match so we'd need to add something appropriate.

wizzycom:
Thank you very mush for the info. This will help me further. 
So my issue is solved. I think it will be merged soon.

Thank you very much @laf

geordish:
Been going a few hours now, and nothing:
image

I will take a proper look into why its not working for me later.

laf:
Chuck the debug output up when you can.

spinza:
>What's CCQ?

CCQ is a[ measure of quality](http://wiki.mikrotik.com/wiki/Manual:Wireless_FAQ#What_is_CCQ_and_how_are_the_values_determined.3F).  

It should be available on both Tx and Rx of wireless interfaces though I think Mikrotik mainly makes the Tx available on SNMP.

spinza:
I referenced the stuff of interest in #1232 before.  

I imagine it makes more sense to put the signal, CCQ and these numbers all on the relevant port page as you'd have it per wireless port.

geordish:
[debug.txt](https://github.com/librenms/librenms/files/714404/debug.txt)

geordish:
As @laf pointed out I'm an idiot.

I can confirm it works.

image

laf:
image

spinza:
I see the signal is coming through.  Seems to be working on my setup, but I have to point out again that the signal is a per interface item. Not really a device level sensor.

laf:
We have no support for adding this to interfaces so for now it's either this or nothing.

caspercat:
How does it works? mtxrWlRtabStrength shows average value from all connected radio users?
@spinza Do You think it should by shown per interface in port section or better in wireless section?
@laf Adding that sort of data (for example to interfaces) causes lot of fundamental changes?

wizzycom:
mtxrWlRtabStrength is grabbed from the wireless registration table of mikrotik and I think that it shows the average per user or connection. 
Signaling per interface will be trick, because if on a connection you are on the client side, then as far as I remember there are no signaling data on the interface.
Mikrotik is back ages on SNMP.

caspercat:
Thanks for explanation, in that case it works as I supposed.

spinza:
> We have no support for adding this to interfaces so for now it's either this or nothing.

Then this.  

I'd like to make it work on interfaces at some point as I'd love to add some of the other per interface values.  I just need a way to make it work?  Basically

spinza:
This is working on three different devices I tested:
image

caspercat:
I'm not sure if signal stuff is working correctly.
In miktotik device wlan interfaces are down if there is no connected clients.
In that case clients has disconnects and the sensor are gone:
[discovery-g6-rbswr.ap.txt](https://github.com/librenms/librenms/files/718719/discovery-g6-rbswr.ap.txt)
Iv'e got 4 wlan's, discovery found 2, with connected clients. While polling client has disconnected, one wlan left and there was some errors.

laf:
Your device isn't responding back to the snmp request with data so the sensor being removed is expected.

I don't think Mikrotik are known for there snmp implementation but this isn't something we can fix.

wizzycom:
In the case of point to point links, this is working like a charm. In the case where you have multiple clients connected on an AP, then there will be issues

wizzycom:
As @laf correctly said

spinza:
The bottom graph on my post above is my home wifi.  So it does work with multiple clients.  The others are point to point.

spinza:
I'm rolling this out today on a large mikrotik base.

caspercat:
@laf What a pity it's not fixable. But if it wasn’t polled as a sensor, but as a interface value it wouldn’t by vanished when wlan is down.
While device is being discovered ./discovery.php finds connected user OIDs, when ./poller.php works it's pooling those OIDs, when client is disconnected, OID can't be found and signal is 0dB. IMO - if it works that way there's no sense to polling that type of information (including @geordish work at #5509) except p2p configured interfaces. But still it should be assign to interface because it can be down, but not gone. I'm truly confused where to present rrd's, in ports section or wireless section. On the other hand it would be nice to know how many clients are connected to AP.

PR #5135, opened 2016-12-08
feature: RRD rename operations

murrant:
#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you signed the [Contributors agreement](http://docs.librenms.org/General/Contributing/)
- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

Template rename operations and only run them once

PR #5509, opened 2017-01-19
Initial Mikrotik/RouterOS WiFi Stuff

geordish:
I'm not convinced about any of this.
* Not a fan of the way this is done... it feels very cumbersome
* It WILL break with multiple radios. Unsure how to address this.
* Unsure about what RRD types the OIDs should be. The MIB was definitely wrong on Client Count, so can't blindly follow that.

Merge at your peril.

DO NOT DELETE THIS TEXT

#### Please note


caspercat:
Thanks for answer and your work.
Wireless section is done:
print screen 2017-01-19 10 25 01
polling wifi section: [walk-output-wifi-module.txt](https://github.com/librenms/librenms/files/716160/walk-output-wifi-module.txt)
and it brakes with multiple radios, just like You sad.
But I don't understand one thing. If we have polled several radios, what's the point to put noise floor to one rrd? Maybe it has point for client count on ap box, but nothing more. Is it possible to split those values to separate rrd graphs and present them in some order? Has it some connection with:

> Unable how to address this.

You've route, that

> MIB was definitely wrong on Client Count

what do You mean?
Global mikrotik wifi client count is still bugy, but @novafire99 was writing by that in #1232 maybe that's the solution. Haven't check.

wizzycom:
I just saw the wireless output posted and it seems very promising. But there is a case that breaks everything and that is the case of the access point where you have multiple clients or random clients if the AP is public. The number of wireless clients is a very useful graph but each time a client connects, some new RRDs will be created for CCQ, noise etc. This will end up with multiple RRDs that will not be in use after the client disconnects.

All these graphs would be useful only on point-2-point links where the client is one and unique.

I don't know if there are public access points on 5GHz out there but here we always use 2.4GHz. 5GHz are only used for point-2-point links. So the discovery process can first read the band used on the interface (OID .1.3.6.1.4.1.14988.1.1.1.3.1.8) and if the band is at 5GHz then the poller should generate wireless graphs.

Just a thought. I don't know if it is relavant but I hope it helps

wizzycom:
Hello guys.

Just a question. Is this information taken from the registration table ?? (table OID .1.3.6.1.4.1.14988.1.1.1)

caspercat:
Those values are taken from mtxrWlAptabTable.

.1.3.6.1.4.1.14988.1.1.1 - mtxrWireless
.1.3.6.1.4.1.14988.1.1.1.1 - mtxrWlStatTable
.1.3.6.1.4.1.14988.1.1.1.2 - mtxrWlRtabTable - this is registration table
.1.3.6.1.4.1.14988.1.1.1.3 - mtxrWlAptabTable - this is ap table

prin screen 2017-01-20 14 38 18

caspercat:
@wizzycom about 5GHz AP (I've got Your earlier post on email) I'm polling now ~40 RB800 AP with 4 wlan's interface. 3 wlan's are working in 2.4GHz and 1 in 5GHz. Generating graphs based on band isn't good idea.
At the time I don't have any Mikrotik bridge, I'll check next week if we posses some boxes to make little lab to check how it look like. 
I've got one p2p radio, but on Ubiquity I'll also check how It's discovering and polling.

geordish:
Librenms/reviewers, how should we progress this? Hard code the max number of radios to something like 4 is my best idea (but that is pretty crappy)

Any better ideas?

spinza:
I think things like ADSL specifc stuff goes onto the port page when it's present.  It really feels to me that the wireless specific stuff needs to go onto the port page. This should then also take care of multiple wireless interfaces.

laf:
I won't protest to knowing TIK's in any level of detail nor wireless in general so can't comment on how best to approach it at the moment.

However on the code aspect it seems like you could shrink this down to just poll ifType, this gives you the index you want and the type to check against.

ifDescr and ifName don't appear to be used other than debug output so I'd ditch them.

geordish:
I've come up with a solution for multiple radios. Any RRDs created with the previous code will need removing.

```bash
find -name mikrotik-wifi.rrd -exec rm {} \;
```

laf:
Is mikrotik-wifi.rrd from this PR or is it existing data people might have?

geordish:
It was created as part of this PR. Doesn't exist outside of this patch.

spinza:
Mikrotik wifi related snmp items are all tied to a particular port/interface.

geordish:
I've added a WiFi section to the ports page, however I'm not very happy with it. It is completely mikrotik specific.

Ideally, there would be a common set of statistics that are collected about radios, which can be presented, but I'm not sure what that would be, or if its even possible.

I'm tempted to revert this latest commit.

murrant:
@geordish would #5135 help.   I haven't worked on because I couldn't find any use cases.  It could also be used to change rrd files, not just rename.

geordish:
@murrant Do you mean that about my mention of deleting the existing RRD, and recreating it? If so, that isn't a problem as that RRD was only created for this PR.

Or are you suggesting it to try and normalize what statistics are currently being collected... That sounds challenging!

murrant:
Yes, that is exactly what I am suggesting, it would be challenging.  Mostly a base set of things we should be able to get from wireless devices (and ones people want).  I have this on my to-do list but I haven't even begun to think about it.

Also, it doesn't have to be done in this PR.

geordish:
Well if we can get a base set of things that are wanted from wireless graphing, I can massage this PR to store and present the data in a sensible format. Other devices can then be moved across as and when.

Only problem with this is my knowledge of useful WiFi information is pretty slim. Feedback about what useful metrics are (from anyone) would be appreciated. Not just from Mikrotik, but in general.

murrant:
I'm actually in the same boat with knowledge, I don't have a good understanding of the needs.  Perhaps a community post is in order?

geordish:
Created community post here: https://community.librenms.org/t/improve-wifi-collection-and-graphing-code/453

laf:
What's the score with this one?

geordish:
It does work, but is very mikrotik specific. I'm not really keen with how its implemented.

Could merge and fix later?

murrant:
I'd be worried about users data.  If we have more generic rrds and structure that will help with future work.

geordish:
OK, plan of attack:
* Trawl through existing wireless collection code to see what fun stats are collected there
* Change this to be generically named, and record NaN for any stats that mikrotik doesn't present

Sound good?

geordish:
Upon reflection, there is LOADS of stuff collected. I stopped half way through looking at the Canopy file.

Tx Power 
Rx Power 
Tx Modulation  
Rx Modulation 
Tx Data Rate 
Rx Data Rate 
Aggregate Data Rate  
Noise Floor 
Signal Strength Ratio  

// Cambium stuff...
TDD Sync Status  - Gauge (Seems to be for LTE?)
RSSI
SNR
GPS Visiable Satelites
GPS Tracked Satelites
Uplink MCS Mode
Downlink MCS Mode
Number of Connected STA
Number of Registered STA
Network Entry Attempts
Network Entry Successes
Effective Sync Source
STA Connected RF Frequency

//Canopy
FEC Errors In
FEC Errors Out
CRC Errors
Jitter
dBm Horizontal
dBm Vertical


I can't imagine how we would make this generic. They all seem to be fairly unique in terms of what stats they produce.

murrant:
I think the best approach is one rrd for each type of data.  Then for the UI, detect the available rrds and show graphs for them.

geordish:
I've added a commit experimenting with extending discovery for WiFi stuff.

Should give you an idea about what I'm thinking... there is a lot of junk there mind.

Thoughts?

geordish:
Bugger, something has gone really wrong with this....

I will try and split this out into a new PR :(

geordish:
Broke this branch, so moved to #5848

kalamchi75:
Hi Guys, 

i'm having some issue with graphing Mikrotik  RB912UAG-5HPnD Wireless stuff. 
Mainly The RxTx rate graphs show zero value all the time. Please see below. 
txrx-combined
txrx-seperate

I have those graphs in Cacti and they do produce values. 
Also, the Tx and Rx channels sensors are not being discovered. I also have those graphs in cacti. 

Part of the issue (i think) is that the way Mikrotik handles the OID's of those sensors, where every device has its own OID for a specific sensor rather than a unified OID in all wireless device of the same model. 

I would kindly like to ask if you could guide me on how to solve this issue. We have a bunch of those wireless devices and I would like to monitor them in LibreNMS. 

Thanks

PR #5516, opened 2017-01-19
Swap order of tests, and error our if one fails rather than continuing

geordish:
The pre-check often fails on style and lint, but I spend ages waiting for the unit tests to complete before getting to this.

I've swapped the default order of the tests, and made it bomb out if one of the tests fail so they can be corrected.

DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/scripts/pre-commit.php
+++ b/scripts/pre-commit.php
@@ -48,7 +48,7 @@ $completed_tests = array(
 $all = !check_opt($options, 'l', 'lint', 's', 'style', 'u', 'unit');
 if ($all) {
     // no test specified, run all tests in this order
-    $options += array('u' => false, 's' => false, 'l' => false);
+    $options += array('l' => false, 's' => false, 'u' => false);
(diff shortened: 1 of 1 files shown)


murrant:
The order change is fine, but I'm not a fan of the bailing out.  I'd like to fix all errors in one pass if possible rather than running the command over and over again.

Also, if you are going to exit on failure you might as well pass --stop-on-error and --stop-on-failure to phpunit.

laf:
Here's the timing for mine:

```
Running lint check... success

real	0m25.931s
user	0m19.092s
sys	0m19.779s
Running style check... success

real	0m17.056s
user	0m12.168s
sys	0m0.466s
Running unit tests... success

real	0m29.180s
user	0m25.174s
sys	0m2.913s
```

I do tend to cancel pre-commit when it shows issues so exiting for me is fine. Be interested in other peoples thoughts.

Issue #5514, opened 2017-01-19
Problem with non-standard mysql port


fenice2:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [ ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ ] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ ] Please provide as much detail as possible.
- [ ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

I'm having a problem setting a non-standard port for MySQL, I'm trying to connect to a Galera cluster through proxysql. My install is updated on a daily basis and after a question of the mailing list I tried the suggestion of using this format in the config file: $config['db_host'] = '192.168.1.221:6033';

Unfortunately this seems to cause a problem, when I run validate.php with the port set I get the following output:

./validate.php
<h2>MySQL Error</h2>Unknown MySQL server host '192.168.1.221:6033' (-2)[root@mistral

The output from a working set-up connected to the local mysql on the standard port gives this:

==========================================================
Component | Version
--------- | -------
LibreNMS  | 61b73e8a73e145823a3afff0c735e32f1a3c78c0
DB Schema | 155
PHP       | 5.4.16
MySQL     | 10.1.21-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
==========================================================

I think I've done that correctly but If there's anything I've missed then let me know.

Regards

Bill

geordish:
I've created a PR to add a $config['db_port'] option which might help you. Fancy giving it a go?

fenice2:
I initially assumed that was the correct format but no such luck. :)

Yes, I'd certainly like to give it a try, what do I need to do?

Regards


Bill

geordish:
wget https://patch-diff.githubusercontent.com/raw/librenms/librenms/pull/5517.diff 
git apply 5517.diff

[...]

PR #5517, opened 2017-01-19
Added config option for database port

geordish:
Fix for #5514

DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/poller-wrapper.py
+++ b/poller-wrapper.py
@@ -77,7 +77,7 @@ except:
 poller_path = config['install_dir'] + '/poller.php'
 db_username = config['db_user']
 db_password = config['db_pass']
-db_port     = config['db_port']
+db_port     = int(config['db_port'])
 
 if config['db_host'][:5].lower() == 'unix:':
     db_server = config['db_host']
--- a/poller-wrapper.py
+++ b/poller-wrapper.py
@@ -77,6 +77,7 @@ except:
 poller_path = config['install_dir'] + '/poller.php'
 db_username = config['db_user']
 db_password = config['db_pass']
+db_port     = config['db_port']
 
 if config['db_host'][:5].lower() == 'unix:':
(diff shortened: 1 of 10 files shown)


poller-wrapper.py
```diff
@@ -86,7 +87,6 @@ def get_config_data():
     db_port = int(config['db_host'].rsplit(':')[1])
 else:
```
geordish: I'm tempted to just remove this completely, as I don't see how it could work with the rest of librenms?

laf:
`./LibreNMS/IRCBot.php` needs updating

Would also be good to do `./html/plugins/Weathermap/data-pick.php` from the plugins repo. Happy to do that.

Looks ok to me, needs testing quite a bit though.

laf:
Also needs:

./poller-service.py
./discovery-wrapper.py

poller-wrapper.py
laf: Probably a hangup from when we used mysql_connect which I think did/does accept :port

laf:
Running it in test install now.

laf:
@librenms/reviewers Works for me. Anyone else want to test?

VimCommando:
If you run into an issue like I did, where `poller-wrapper.py` tells you it cannot connect to the database, but `validate.php` and `poller.php` say the DB is OK; double-check your `db_host` string. If you included the port previously (i.e. "localhost:3306") this update will break that.

The fix is to simply define `db_port` separately in config.php and `poller-wrapper.py` should work fine again.

laf:
Sorry that this might have broken something for you - out of interest how did you have poller.php working on the non-standard port or is it just that you'd specified :3306 all along?

VimCommando:
I just had the default port specified in my URL string, even though I didn't need to.

PR #5529, opened 2017-01-20
fix: Changing device type now is persistant

laf:
DO NOT DELETE THIS TEXT

#### Please note

A bit of a quick hash for now. Changing device types is broken at the moment as load_os detects it's different from the what it's supposed to be and changes it back as soon as the page loads.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -636,7 +636,7 @@ function createHost($host, $community, $snmpver, $port = 161, $transport = 'udp'
         if (host_exists($host, $snmphost) === false) {
             $device_id = dbInsert($device, 'devices');
             if ($device_id) {
-                oxidized_reload_nodes();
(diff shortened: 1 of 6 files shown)


spinza:
Tested on my demo install.  Was able to change the device type of a device I was not able to change before checking this out.

includes/functions.php
```diff
@@ -636,7 +636,7 @@ function createHost($host, $community, $snmpver, $port = 161, $transport = 'udp'
         if (host_exists($host, $snmphost) === false) {
             $device_id = dbInsert($device, 'devices');
             if ($device_id) {
-                oxidized_reload_nodes();
+                oxidized_reload_§nodes();
```
murrant: Looks like an accidental change here.

PR #5530, opened 2017-01-20
Add MIB for CoolIT's liquid cooling units

cubranic:
Source: https://www.coolitsystems.com/index.php/data-sheets/category/8-software.html?download=10:rack-dclc-mib


laf:
This mib won't be used, what's the reason for adding it in?

cubranic:
Given the large collection of MIBs in the repo, I thought you'd welcome having this one too. It's for CoolIT cooling units.

laf:
It's not about welcoming it or not, it's about that adding the mib has no benefit, code has to be written to use it so unless you've got some additional features you want adding which make use of the mib then their is no point in adding the file.

cubranic:
closing

PR #5550, opened 2017-01-21
refactor: Tidy up sensor discovery

laf:
DO NOT DELETE THIS TEXT

#### Please note

Initial start at cleaning up sensor discovery.

We now no longer cycle through all include files and instead only process files that exist for the device group, device os or rfc1628.

Some files have been moved around so they are named correctly after the OS, removed because the code was moved to another file or duplicated code because it needed to run on more than one OS but no shared group.

It may give a slight performance increase because it's not looping through all the files but that wasn't the aim of the PR. It's done so that the code is simplified and not including lots of files.


includes/discovery/functions.inc.php
```diff
@@ -953,3 +953,35 @@ function ignore_storage($descr)
...
+
+/**
+ * @param $types
+ * @param $device
+ * @param array $pre_cache
+ */
+function sensors($types, $device, $pre_cache = array())
+{
+    global $config;
+    if (is_array($types)) {
```
murrant: You can avoid the is_array check by casting $types to array `foreach ((array)$types as $type)`

laf:
Updated them

includes/discovery/functions.inc.php
```diff
@@ -953,3 +953,33 @@ function ignore_storage($descr)
...
+        $dir = $config['install_dir'] . '/includes/discovery/sensors/' . $type .'/';
+
+        if (is_file($dir . $device['os_group'] . '.inc.php')) {
+            include_once $dir . $device['os_group'] . '.inc.php';
+        }
+        if (is_file($dir . $device['os'] . '.inc.php')) {
+            include_once $dir . $device['os'] . '.inc.php';
+        }
+        if (isset($config['modules_compat']['rfc1628'][$device['os']]) && $config['modules_compat']['rfc1628'][$device['os']]) {
+            if (is_file($dir  . '/rfc1628.inc.php')) {
```
murrant: This check might not be needed.

Could also avoid the previous is_file() checks if you use `@include_once`, but I'm not sure which would be better/preferred.

includes/discovery/sensors/fanspeeds/linux.inc.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+$oids = snmp_walk($device, '.1.3.6.1.4.1.10876.2.1.1.1.1.3', '-OsqnU', 'SUPERMICRO-HEALTH-MIB');
```
murrant: Supermicro is a bit of an oddball because it should be supported on windows too, but it probably shouldn't be it's on os.  Perhaps move to supermicro.inc.php and include in windows and linux?

includes/discovery/sensors/temperatures/linux.inc.php
```diff
@@ -15,3 +15,59 @@
         discover_sensor($valid['sensor'], 'temperature', $device, $sensor_oid, 1, $sensor_type, $descr, 1, 1, null, null, null, null, $value);
     }
 }
+
+echo 'HP_ILO ';
```
murrant: It might be a good idea to retain the `if (startswith($sysObjectId, '.1.3.6.1.4.1.232.') ) {` check here.

includes/discovery/sensors/temperatures/linux.inc.php
```diff
@@ -15,3 +15,59 @@
...
+        $threshold     = snmp_get($device, $threshold_oid, '-Oqv', '');
+
+        if (!empty($temperature)) {
+            discover_sensor($valid['sensor'], 'temperature', $device, $temperature_oid, $oid, 'hpilo', $descr, '1', '1', null, null, null, $threshold, $temperature);
+        }
+    }
+}
+
+// Supermicro sensors
+$oids = snmp_walk($device, '.1.3.6.1.4.1.10876.2.1.1.1.1.3', '-Osqn', 'SUPERMICRO-HEALTH-MIB');
```
murrant: I know it is probably out of scope for this PR, but this should probably be in pre-cache.

includes/discovery/functions.inc.php
laf: Why can we ignore the checks? the rfc1628 aren't named after groups or os so need calling.

Not sure what you mean by avoid the is_file checks, do you mean just call include_once and not worry about no file existing?

laf:
Updated

includes/discovery/functions.inc.php
murrant: I mean is_file(rfc1628.inc.php) will always be true.

includes/discovery/sensors/temperatures/linux.inc.php
```diff
@@ -15,3 +15,30 @@
...
+            $temperature = snmp_get($device, $temperature_oid, '-Oqv', '');
+
+            $threshold_oid = ".1.3.6.1.4.1.232.6.2.6.8.1.5.$temperature_id";
+            $threshold = snmp_get($device, $threshold_oid, '-Oqv', '');
+
+            if (!empty($temperature)) {
+                discover_sensor($valid['sensor'], 'temperature', $device, $temperature_oid, $oid, 'hpilo', $descr, '1', '1', null, null, null, $threshold, $temperature);
+            }
+        }
+    }
```
murrant: I think you are missing the include supermicro.inc.php in this file.

includes/discovery/functions.inc.php
laf: Ah, well rfc1628 doesn't exist in all sensor types hence the checks still.

includes/discovery/sensors/temperatures/linux.inc.php
laf: Yeah put it in the supermicro one instead lol :(

includes/discovery/sensors/temperatures/linux.inc.php
```diff
@@ -15,3 +15,30 @@
         discover_sensor($valid['sensor'], 'temperature', $device, $sensor_oid, 1, $sensor_type, $descr, 1, 1, null, null, null, null, $value);
     }
 }
+
+if (startswith($sysObjectId, '.1.3.6.1.4.1.232.')) {
```
murrant: Should be starts_with(

laf:
Updated

laf:
Rebasing and pushing update now.

laf:
Whilst it's not big deal, rfc1628.inc.php doesn't exist for all sensor modules so for devices which support it they will try and include non-existent files.

murrant:
@laf oh, I had a slight misunderstanding on rfc1628 stuff.  Give me a bit to sort it out.

murrant:
Ok, let me do one more testing pass and this should be good to merge.

laf:
I'm rebasing this now, we ok to merge?

laf:
I've rebased this.

laf:
Thanks to @Rosiak for spotting the last issue, fix now pushed so should be good to go @librenms/reviewers

laf:
Anything further on this one?

PR #5557, opened 2017-01-23
Correcting some HP to HPE logo

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

- Edited some HP device to use the HPE logo without changing any text that can break other things.


laf:
imho this logo isn't as good as the old hp.svg one :(

FTBZ:
Yes, but it's no longer the same company...I used the logo from @murrant in the PR #5400. 

I noticed that HPE is using only the green rectangle in some infrastructure. Perhaps it's better?

murrant:
HPE has the option to have just the green rectangle, should we use that?

FTBZ:
I changed the logo to only the rectangle. It's not correct with the Terms of use, but this is the case for the majority of logos and HPE itself use this in infrastructure interface like vCenter.

I just added two devices on my dev VM to show it:

screen shot 2017-01-24 at 15 34 12

murrant:
@FTBZ Perhaps #5573 might interest you :)

laf:
My personal opinion is we stick with the current hp logo but I'm not a user of any hpe stuff in LibreNMS so take my answer loosely

murrant:
@laf actually, we should be using both as they are two separate things now.

laf:
Well yes because of your PR but that wasn't the case before.

murrant:
@laf I meant both HP and HPE logos.  HP is basically printers, HPE is enterprise stuff.

laf:
I know the companies are two separate things but that HPE logo was pretty poor and we have a choice on using one or the other.

html/images/os/hpe.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="300" height="300" viewBox="0 0 300 300"><path fill="#00B188" d="M277.3 173.6H22.7v-47.2h254.6v47.2zm18.2-65.4H4.5v83.6h291v-83.6z"/></svg>
```
murrant: I wonder if this should have all the padding removed...

PR #5562, opened 2017-01-23
removed references to librenms being a fork

laf:
DO NOT DELETE THIS TEXT

#### Please note

I think we can stand on our own two feet now. Doesn't remove copyright


InsaneSplash:
There is also reference in the About page under - Acknowledgements.

laf:
I've left that one, just wanted to remove references to us being a fork, no need to remove attribution

PR #5563, opened 2017-01-23
tidy up some file permissions and removed unused files

laf:
DO NOT DELETE THIS TEXT

#### Please note

Changed a few files to 644 perms. Removed some unused scripts and moved others.


murrant:
All look good, but I'm not sure why you kept makemin.pl  it seems to be a wierd js minifier that used to come with overlib...

laf:
Because we don't minify our js but should :)

murrant:
laf: we'd use a real one then :)

PR #5569, opened 2017-01-24
Graph for services would not work

nerdalertdk:
I agree to the conditions of the Contributor Agreement contained in doc/General/Contributing.md.

Just found an bug, when adding an new service is sets the server_ds to {} [See here, end of line L46](https://github.com/librenms/librenms/blob/d5296319fbadd41dc5a82cbe4bbc6e3fa82b1cb7/includes/services.inc.php#L46)
And here we check if service is equal to {} and not empty

DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/includes/services.inc.php
+++ b/includes/services.inc.php
@@ -152,7 +152,7 @@ function poll_service($service)
             $DS[$k] = $v['uom'];
         }
         d_echo("Service DS: "._json_encode($DS)."\n");
-        if ( ($service['service_ds'] == "{}") || ($service['service_ds'] == "") {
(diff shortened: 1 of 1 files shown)


murrant:
Are you sure you don't need to check for both "" and "{}"?

nerdalertdk:
Don't but all new services get the {} on insert

laf:
I have legacy services where service_ds is blank so yes, a check for both should be included.

nerdalertdk:
Changed it but blanks should be working in the current version as long as the nagios service is returning graph data

PR #5575, opened 2017-01-24
fix: Fix broken netstats ip forward polling

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/includes/polling/netstats/netstats-ip_forward.inc.php
+++ b/includes/polling/netstats/netstats-ip_forward.inc.php
@@ -1,39 +1,18 @@
 <?php
-
 if (!starts_with($device['os'], array('Snom', 'asa'))) {
     echo ' IP-FORWARD';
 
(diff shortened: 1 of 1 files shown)


includes/polling/netstats/netstats-ip_forward.inc.php
```diff
@@ -1,39 +1,18 @@
...
-
+    $oid = 'ipCidrRouteNumber';
     $fields = array();
-    foreach ($oids as $oid) {
-        if (is_numeric($data[0][$oid])) {
-            $value = $data[0][$oid];
-        } else {
-            $value = 'U';
-        }
+    $rrd_def = array();
```
murrant: For a single RRD def, you can just use
```
 $rrd_def = "DS:$oid:GAUGE:600:U:5000000";
```
Doesn't need to be an array.

PR #5584, opened 2017-01-24
Updating help text when validate cannot find a bin

andrewwippler:
Adding additional help text to help new users configure their librenms install. In my install, fping was not at the expected path. This help text would have helped me.

DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/validate.php
+++ b/validate.php
@@ -207,7 +207,7 @@ $space_check = (disk_free_space($config['install_dir']) / 1024 / 1024);
 if ($space_check < 512 && $space_check > 1) {
     print_warn('Disk space where '.$config['install_dir'].' is located is less than 512Mb');
 }
-d
+
 if ($space_check < 1) {
     print_fail('Disk space where '.$config['install_dir'].' is located is empty!!!');
 }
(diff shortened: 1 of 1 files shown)


validate.php
```diff
@@ -219,6 +219,8 @@
     $cmd = rtrim(shell_exec("which {$config[$bin]} 2>/dev/null"));
     if (!$cmd) {
         print_fail("$bin location is incorrect or bin not installed");
+        print_fail("You can also manually set the path to $bin by placing the following in config.php:");
+        print_fail("\$config['$bin'] = \"/path/to/$bin\";");
```
laf: This shouldn't be wrapped in a new print_fail as it show as a new line, you will need to include it in the line above.

andrewwippler: Ok, should I include line breaks?

laf: See how it's done here https://github.com/andrewwippler/librenms/blob/4d3120b74a682de7748bc06188e19e85aa79e4f5/validate.php#L147

andrewwippler: Disregard this next pull. I think you want me to add all missing $bin into an array and then format output.... it that correct?

validate.php
```diff
@@ -207,7 +207,7 @@
 if ($space_check < 512 && $space_check > 1) {
     print_warn('Disk space where '.$config['install_dir'].' is located is less than 512Mb');
 }
-
+d
```
laf: Think this was added by accident :)

andrewwippler: It was :)

Issue #5585, opened 2017-01-24
Dashboard | Widget Map - Engine Leaflet - Don't loading Up+Down


dneto82:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [X] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

My Version:
==========================================================
Component | Version
--------- | -------
LibreNMS  | 3c0a2109e592e501b9a1ae6746e46cbe132c776a
DB Schema | 157
PHP       | 7.0.14-pl1-gentoo
MySQL     | 5.7.12
RRDTool   | 1.6.01.6.0
SNMP      | NET-SNMP 5.7.3
==========================================================

The problem:

Overview page not loading map, when I have a lot of nodes. 

[Screenshot of Problem](https://ibin.co/3A1qbAU5Nv6K.png)
[Screenshot with only down Devices](https://ibin.co/3A1rF8FM4RjJ.png)

Developer Console showing errors:

[link:](https://ibin.co/3A27SA6EzDpA.png)

Maybe the problem is related with some data of my locations (sysLocation). Because before, everything was working, but when I add more nodes on last days. I'll check again all data of them.

dneto82:
Hi Guys,

As I suspected (and @laf guessed ) , it was really related to my location data. The problem is that despite all other parts of the system are capable of handling atypical characters, maps can't.

I identified two problems in my data, that when fixed in the device, the map came back perfectly:

1. [0,0] Coordinates like this are not interpreted by the map engine. Although they should, as they are point 0 (near to Guine Gulf). 

[...]

PR #5589, opened 2017-01-25
fix for issue #5585 sanitize location string

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/includes/polling/functions.inc.php
+++ b/includes/polling/functions.inc.php
@@ -489,6 +489,7 @@ function location_to_latlng($device)
     $device_location = $device['location'];
     if (!empty($device_location)) {
         $new_device_location = preg_replace("/ /", "+", $device_location);
+        $new_device_location = preg_replace('/[^A-Za-z0-9\-\+]/', '', $device_location); // Removes special chars.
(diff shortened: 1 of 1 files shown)


InsaneSplash:
I included the removal (by virtual or it not being allowed) the , in a name. I figured that its not required when searching google for a map location as long as each word is separated with a +

laf:
This will cause more location lookups to happen. What about users on non-english characters?

InsaneSplash:
Why would it cause any more than is currently, as its still within the same routine? I might not understand fully so please excuse me. I was looking at that, and have some conversion code we can use, but we would need to do some more testing I just don't have the "right" languages in my computer to do this?

Would it be worse than it is now?

laf:
so the lookups are done on the location string, if that changes another lookup has to be done.

InsaneSplash:
OK, do you think LibreNMS UTF-8 compliant?

laf:
I doubt it :/

InsaneSplash:
:) then we should be OK, and it gives us something to maybe look at moving towards :) ehhe

PR #5594, opened 2017-01-25
feature: Added phpunit db setup tests

laf:
DO NOT DELETE THIS TEXT

#### Please note

I'm sure @murrant will rewrite this but it gets us validation at the moment that build-base will work

Doesn't use mysql strict at the moment as it will fail with this.


laf:
Not sure why hhvm is failing here, especially seeing as it's set to allow failures.

laf:
@librenms/reviewers Rebased and it's now all working, think murrant fixed something :)

I have tested php build-base.php and web installs still work off the back of this.

includes/init.php
```diff
@@ -32,8 +32,10 @@
 chdir($install_dir);
 
 // Libraries
-require('Net/IPv4.php');
-require('Net/IPv6.php');
+if (!module_selected('phpunit', $init_modules)) {
```
laf: I couldn't install them into the travis env so the test would fail.

murrant: Odd, I don't remember running into that problem...

laf: How did you do it before? We ask users to install them via pear or apt-get normally.

laf:
Reverted back to not using net/ipvX modules.

laf:
Yes I put the do test first so it's ready for other tests. What's up with it being first?

murrant:
I changed around some things, so it wasn't done first.  I can put it back.  But right now, it creates a db, builds it up and then drops the database, so it is self-contained and no other tests rely on it.

If you want to build up a database for other tests to use, I don't think this is the right approach.

laf:
I know overall it's not the right approach but we don't have a proper dbal to make use of things like in v2 so at present this is just a make do with what we have.

murrant:
I'm happy with this PR as a way to test our sql schema files.  I think we should merge this as is and make changes later when we need the DB for another test.

laf:
I have started to look at using a dbal and writing wrappers around our current calls to the dbal. Probably the easiest way to deal with this for now.

murrant:
I looked at that awhile ago, then I ended up at a spot where I wanted full eloquent :(

Then, I decided it was better to focus energy elsewhere, like v2.

PR #5595, opened 2017-01-25
webui: rename Device Group -> Dynamic Device Group

murrant:
DO NOT DELETE THIS TEXT

#### Please note

Thoughts?


InsaneSplash:
Maybe a little long? What about just renaming it to Dynamic Groups?

5cents...

laf:
Dynamic groups is wrong though as it only deals with devices.

I don't care either way, I don't personally see the need for the change.

murrant:
I'm unsure about the change too.

Issue #5561, opened 2017-01-23
Apache shoots to ~100%; vendor-ezyang-No such file or directory, Permission denied 


ipguru99:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ ] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ x] Please provide as much detail as possible.
- [x ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

LibreNMS  | 3807ec9811abe98d5fbe5faea52cf915d28f6472
DB Schema | 157
PHP       | 7.0.13-0ubuntu0.16.04.1
MySQL     | 10.0.28-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3

I had an issue with a client on 14.04 and Apache just crawling. There were < 10 devices being monitored, all Cisco gear.. almost all modules turned off. It was a PC, not a server, so the first thing I did was add RRDCACHED, verified it was working. Still, just terribly slow. 

We started over with a real server (I have this running on servers for about 10 of my clients right now, so this was not my first time installing or tweaking). It is still terrible! Very slow to bring up some pages??

I am a network guy, not really a software guy, so I started searching for troubleshooting Apache issues. Then I remember I have NFSen also running on this box.. and it runs great??

So I watch 'top' and get the process ID of Apache skyrocketing. I do a "strace -p XXXX" and BOOM.. here is what I see:

stat("/opt/librenms/vendor/ezyang/htmlpurifier/library/HTMLPurifier/DefinitionCache/Serializer/HTML", 0x7fff27ef0820) = -1 ENOENT (No such file or directory)
[...]

PR #5596, opened 2017-01-25
fix: HTML Purifier would create tmp caches within the vendor folder, moved to users tmp dir #5561

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5561 

HTMLPurifier will create a file in /opt/librenms/vendor/ezyang/htmlpurifier/library/HTMLPurifier/DefinitionCache/Serializer/HTML/ which is owned by your web server so a php composer update stops.

### DIFF
--- a/includes/common.php
+++ b/includes/common.php
@@ -1503,9 +1503,10 @@ function clean($value)
  */
 function display($value)
 {
-    $purifier = new HTMLPurifier(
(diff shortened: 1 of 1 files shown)


murrant:
Looks good.

We should probably only initialize one of these rather than doing it every time right?

laf:
Probably makes more sense

PR #5617, opened 2017-01-26
Added support for Brocade 200E

aldemira:
DO NOT DELETE THIS TEXT

#### Please note


laf:
You need to also add the extra test call into tests/OSDiscoveryTest.php

laf:
Also, whilst you are at it, can you change your github name in AUTHORS.md to aldemira rather than aldemir_a.

It's also worth to mention that you should avoid working in your master branch. We have docs to help you with that :)

laf:
Sorry my head isn't with it :(

Can you add sysDescr to the test unit file, it's not used for discovery but helps with the whole unit testing piece in general.

aldemira:
I would, but sysDescr string says "Fibre Channel Switch."
I think it's a bit too generic, don't you think so?

laf:
Not really, yes it is generic but if someone adds a new device which only matches on Fibre Channel sysDescr and it runs before this one then it will break detection. The more test data we have the better :)

PR #5629, opened 2017-01-27
webui: New LibreNMS logo assets

murrant:
SVG images where we can.
Updated favicon and app icons.

DO NOT DELETE THIS TEXT

#### Please note


murrant:
An example.

Old:
librenms-header-mono-old

New:
librenms-header-mono

InsaneSplash:
Nice!

laf:
The only thing with this is the export to pdf (used once in alerts log), it doesn't render properly. Really not an issue but we should address.

laf:
btw looks freakin sweet :)

PR #5652, opened 2017-01-28
webui: Final Font Awesome conversion

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note


html/includes/functions.inc.php
```diff
@@ -710,32 +710,38 @@ function print_optionbar_end()
...
-    if ($message == 'Interface enabled') {
-        $icon = 'if-enable.png';
+    if ($message == 'Interface enabled' || $message == 'ifAdminStatus: down -> up') {
+        $icon = 'fa-bookmark';
+        $icon_colour = 'green';
     }
 
     if (isset($icon)) {
-        return $icon;
+        return array($icon,$icon_colour);
```
laf: Would it not make more sense to do `array('icon' => $icon, 'colour' => $icon_colour)` here then you can refernce `$icon_type = $icon_returned['icon']` later?

html/includes/print-menubar.php
```diff
@@ -422,8 +422,8 @@
 foreach ($app_list as $app) {
     if (isset($app['app_type'])) {
         $app_i_list = dbFetchRows("SELECT DISTINCT(`app_instance`) FROM `applications` WHERE `app_type` = ? ORDER BY `app_instance`", array($app['app_type']));
-        $image = $config['html_dir']."/images/icons/".$app['app_type'].".png";
-        $icon = (file_exists($image) ? $app['app_type'] : "apps");
+        //$image = $config['html_dir']."/images/icons/".$app['app_type'].".png";
```
laf: No point in commenting these out. Just delete them.

html/includes/functions.inc.php
InsaneSplash: Cool, either would work, but I'll update to use the same format used in other parts of LibreNMS.

Issue #5600, opened 2017-01-25
force_ip_to_sysname shows sysname twice in device list


towster:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x ] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x ] Please provide as much detail as possible.
- [x ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

==========================================================
Component | Version
--------- | -------
LibreNMS  | 3807ec9811abe98d5fbe5faea52cf915d28f6472
DB Schema | 157
PHP       | 7.0.14
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
==========================================================

All my devices are discovered by IP.
When listing out devices (devices - all devices) it previously showed under the "Device" column the IP (hotlinked to device details) and below that the sysname.
I turned on force_ip_to_sysname so that the sysname could be seen more easily.
After changing this setting it shows under the "Device" column the sysname (hotlinked to device details) and below that the sysname.

Any chance it could show the IP instead of the sysname twice?

PR #5656, opened 2017-01-29
webui: Only show sysName once if force_ip_to_sysname is enabled #5600

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5600


html/includes/table/devices.inc.php
```diff
@@ -197,7 +197,9 @@
         $platform = $device['hardware'] . '<br>' . $device['features'];
         $os = $device['os_text'] . '<br>' . $device['version'];
         $uptime = formatUptime($device['uptime'], 'short');
-        $hostname .= '<br>' . $device['sysName'];
+        if (ip_to_sysname($device, $device['hostname']) !== $device['sysName']) {
```
murrant: It seems like you don't need the if statement here.

towster:
instead of simply pulling the second sysName from the display could you make the second line display the IP?

laf:
Updated.

PR #5676, opened 2017-01-30
Documentation update for Oxidized integration.

jackson-tim:
DO NOT DELETE THIS TEXT

#### Please note


murrant:
trigger ci

murrant:
Looks good. Thanks!

murrant:
With nginx, it won't be the web server that is connecting, but php-fpm.  So, likely this won't be needed or a different permission would be needed.

laf:
Awesome :)

I guess it's a good idea to cover in the apache one anyway as that's not the only thing we do http lookups for.

PR #5679, opened 2017-01-30
webui: Remove ifIndex from ports list and add debug button for data access

laf:
DO NOT DELETE THIS TEXT

#### Please note

Removed the ifIndex as discussed here: https://community.librenms.org/t/benefit-of-listing-ifindex-number/467/3

The only way to get the ifIndex after that would be a db lookup so I've added a small tag icon which you can click to get more info.

image

### DIFF
--- a/html/includes/functions.inc.php
+++ b/html/includes/functions.inc.php
@@ -1396,3 +1396,17 @@ function search_oxidized_config($search_in_conf_textbox)
     $context  = stream_context_create($opts);
     return json_decode(file_get_contents($oxidized_search_url, false, $context), true);
 }
+
+/**
+ * @param $data
+ * @return bool|mixed
+ */
+function array_to_htmljson($data)
+{
+    if (is_array($data)) {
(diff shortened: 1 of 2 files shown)


adaniels21487:
Philosophical question:
Do we want to be displaying debugging output when debug is not enabled?

laf:
It doesn't bother me either way. Easy enough to change it to do that. 

By default it's not shown, you have to click the icon to get the info.

html/includes/functions.inc.php
```diff
@@ -1396,3 +1396,17 @@ function search_oxidized_config($search_in_conf_textbox)
...
+
+/**
+ * @param $data
+ * @return bool|mixed
+ */
+function array_to_htmljson($data)
+{
+    if (is_array($data)) {
+        $data = htmlentities(json_encode($data));
+        return str_replace(',', ',<br />', $data);
```
murrant: I think it would look cleaner without the `,` before the <br />

laf:
trigger ci

PR #5697, opened 2017-01-31
doc: added information about the icon theme colour

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note


laf:
That seems to indicate editing the style sheets we ship. We shouldn't recommend that and instead give instructions on how to add it to a custom css file.

InsaneSplash:
Does one create a new one, or is there a single file the user should be making changes in that will override the defaults?

is it styles.css?

laf:
Users would have to specify to use a new css file per the install docs. imho I'm not really sure this is worth documenting.

PR #5700, opened 2017-01-31
webui: Set the device logo and cell to have a max width

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

- Set a maximum width for the device logo.
- Set a maximum width for the <td> cell that the device image is displayed in.

This keeps the device name text in the same place where in the past it would be moving all over the place.

### DIFF
--- a/html/css/styles.css
+++ b/html/css/styles.css
@@ -2166,7 +2166,17 @@ label {
   max-width: 250px;
 }
 
-.device-title-bar {
-  width: 300px;
-  max-width: 300px;
+.device-title-bar-1 {
+  width: 25%;
+  max-width: 25%;
+}
+
+.device-title-bar-2 {
+  width: 25%;
+  max-width: 25%;
+ }
+
+.device-title-bar-3 {
+  width: 50%;
+  max-width: 50%;
 }
--- a/html/css/styles.css
+++ b/html/css/styles.css
@@ -2163,5 +2163,10 @@ label {
 
 .device-icon-48h img {
   height: 48px;
-  max-width: 220px
+  max-width: 250px;
+}
+
+.device-title-bar {
+  width: 300px;
+  max-width: 300px;
 }
--- a/html/css/styles.css
+++ b/html/css/styles.css
@@ -2163,4 +2163,5 @@ label {
 
 .device-icon-48h img {
   height: 48px;
+  max-width: 220px
 }
--- a/html/includes/device-header.inc.php
+++ b/html/includes/device-header.inc.php
@@ -21,7 +21,7 @@ $host_id = get_vm_parent_id($device);
 
 echo '
             <tr bgcolor="'.$device_colour.'" class="alert '.$class.'">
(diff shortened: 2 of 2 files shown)


laf:
I didn't think setting column width was good from a fluid perspective? However this just adds a lot of white space next to icons not 220 wide.

Imho looks better as is.

InsaneSplash:
I agree with your points, the problem is 2 fold.

1. The icon presented in the title bar changes in size depending on the width. For example CentOS/Fedora being a more square icon is one size, but when using a logo like Radwin/Lenovo it becomes a massive picture so the sizing needs to be kept in check no matter the original size of the icon.

2. The problem with the way the title bar is built with the graphs, is that as the page loads and the graphs start to be included, the title moves around in the title bar. This fixed cell for the picture stops the title moving around as the browser loads and will keep it aligned to the left next to a cell that does not change its width.

3. This also keeps the title in the same place in the title no matter the icon size.

Examples:
**Before:**
capture-mt

capture-radwin

capture-lenovo

**After:**
image

image

image

laf:
Can we do it without the td width and just rely on css and do min-width?

InsaneSplash:
I'll try move this in to the CSS :)

murrant:
I'm not sure what the goal of the colgroup is , but it forces the graphs onto a second line at certain sizes.

InsaneSplash:
Would you have an example? I've tested against different resolutions and haven't seen that. I didn't set any sizes for the remaining td cells.

InsaneSplash:
The colgroup in HTML5 helps to set the attributes for each cell in the table. The problem is there is a need to set both the image max width in line with also setting the cell's max width. Ultimately I think the title bar table should has a set width for each field to stop the browser from working it out as the data and graphs are loaded.

I am going to try change it a little bit to work with %'s and include all 3 cells. Its not an easy one to solve, but I feel its important to sort it out as from the included pics, its just weird looking :)

murrant:
Have you tried switching it to bootstrap?

InsaneSplash:
I think we would need to remove the table, and go with a grid to achieve the bootstrap conversion?

murrant:
Correct.

html/css/styles.css
```diff
@@ -2163,4 +2163,20 @@ label {
 
 .device-icon-48h img {
   height: 48px;
+  max-width: 250px;
+}
+
+.device-title-bar-1 {
+  width: 25%;
+  max-width: 25%;
```
murrant: isn't max-width redundant?

InsaneSplash: I read up somewhere that its advisable to include it for "older, non compliant browsers" so I put in there for safety. It didn't make any difference to me, but wanted to try cover all bases. Want me to remove it?

murrant: No, idea I'm not a web expert by any means.  I was mostly genuinely curious.

laf:
I'm ok with this now if someone else is?

laf:
trigger ci

PR #5701, opened 2017-01-31
Made the 2 tool icons more appropriate in the menu bar

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

- Changed RIPE icon to be a little more unique.
- Change the Oxidized Icon to be unique and not like the availability icon.

### DIFF
--- a/html/includes/print-menubar.php
+++ b/html/includes/print-menubar.php
@@ -102,11 +102,11 @@ if ($config['oxidized']['enabled'] === true && isset($config['oxidized']['url'])
 <?php
 
 if (isset($config['enable_syslog']) && $config['enable_syslog']) {
-    echo '              <li><a href="'.generate_url(array('page'=>'syslog')).'"><i class="fa fa-archive fa-fw fa-lg" aria-hidden="true"></i> Syslog</a></li>';
+    echo '              <li><a href="'.generate_url(array('page'=>'syslog')).'"><i class="fa fa-tag fa-fw fa-lg" aria-hidden="true"></i> Syslog</a></li>';
 }
 
 if (isset($config['graylog']['server']) && isset($config['graylog']['port'])) {
-    echo '              <li><a href="'.generate_url(array('page'=>'graylog')).'"><i class="fa fa-archive fa-fw fa-lg" aria-hidden="true"></i> Graylog</a></li>';
+    echo '              <li><a href="'.generate_url(array('page'=>'graylog')).'"><i class="fa fa-tag fa-fw fa-lg" aria-hidden="true"></i> Graylog</a></li>';
 }
 
 ?>
--- a/html/includes/print-menubar.php
+++ b/html/includes/print-menubar.php
@@ -88,10 +88,10 @@ if ($config['title_image']) {
           <li class="dropdown-submenu">
            <a href="<?php echo(generate_url(array('page'=>'overview'))); ?>"><i class="fa fa-wrench fa-fw fa-lg" aria-hidden="true"></i> Tools</a>
            <ul class="dropdown-menu scrollable-menu">
-           <li><a href="<?php echo(generate_url(array('page'=>'ripenccapi'))); ?>"><i class="fa fa-chevron-right fa-fw fa-lg" aria-hidden="true"></i> RIPE NCC API</a></li>
+           <li><a href="<?php echo(generate_url(array('page'=>'ripenccapi'))); ?>"><i class="fa fa-star fa-fw fa-lg" aria-hidden="true"></i> RIPE NCC API</a></li>
 <?php
 if ($config['oxidized']['enabled'] === true && isset($config['oxidized']['url'])) {
-    echo '<li><a href="'.generate_url(array('page'=>'oxidized')).'"><i class="fa fa-arrow-circle-up fa-fw fa-lg" aria-hidden="true"></i> Oxidized</a></li>';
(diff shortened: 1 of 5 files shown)


InsaneSplash:
I updated the menu Event Log and Syslog Icons so they are the same as the ones used within the device overview.

html/includes/print-menubar.php
```diff
@@ -88,25 +88,25 @@
           <li class="dropdown-submenu">
            <a href="<?php echo(generate_url(array('page'=>'overview'))); ?>"><i class="fa fa-wrench fa-fw fa-lg" aria-hidden="true"></i> Tools</a>
            <ul class="dropdown-menu scrollable-menu">
-           <li><a href="<?php echo(generate_url(array('page'=>'ripenccapi'))); ?>"><i class="fa fa-chevron-right fa-fw fa-lg" aria-hidden="true"></i> RIPE NCC API</a></li>
+           <li><a href="<?php echo(generate_url(array('page'=>'ripenccapi'))); ?>"><i class="fa fa-star fa-fw fa-lg" aria-hidden="true"></i> RIPE NCC API</a></li>
 <?php
 if ($config['oxidized']['enabled'] === true && isset($config['oxidized']['url'])) {
-    echo '<li><a href="'.generate_url(array('page'=>'oxidized')).'"><i class="fa fa-arrow-circle-up fa-fw fa-lg" aria-hidden="true"></i> Oxidized</a></li>';
+    echo '<li><a href="'.generate_url(array('page'=>'oxidized')).'"><i class="fa fa-stack-overflow fa-fw fa-lg" aria-hidden="true"></i> Oxidized</a></li>';
```
laf: maybe fa-archive?

html/includes/print-menubar.php
```diff
@@ -88,25 +88,25 @@
...
 }
 
 ?>
            </ul>
           </li>
             <li role="presentation" class="divider"></li>
-            <li><a href="<?php echo(generate_url(array('page'=>'eventlog'))); ?>"><i class="fa fa-book fa-fw fa-lg" aria-hidden="true"></i> Eventlog</a></li>
+            <li><a href="<?php echo(generate_url(array('page'=>'eventlog'))); ?>"><i class="fa fa-calendar fa-fw fa-lg" aria-hidden="true"></i> Eventlog</a></li>
```
laf: maybe one of the fa-list options? calendar seems inappropriate

InsaneSplash: Its really just using the same icon for syslog and eventlog which is used within the device overview pages. The view of the icon shows more of an "timed event" than a calendar entry. I could look for another icon, but I'd need to change it in the menu bar, device overview and about pages to keep things the same :)

html/includes/print-menubar.php
InsaneSplash: Archive is also an option. I thought the stack overflow icon worked with it "packing configuration in an inbox" type... but I can change it if you were prefer :)

html/includes/print-menubar.php
laf: Maybe the others need updating, just because something is today doesn't mean it's right.

html/includes/print-menubar.php
laf: Just making suggestions.

html/includes/print-menubar.php
```diff
@@ -88,25 +88,25 @@
...
            </ul>
           </li>
             <li role="presentation" class="divider"></li>
-            <li><a href="<?php echo(generate_url(array('page'=>'eventlog'))); ?>"><i class="fa fa-book fa-fw fa-lg" aria-hidden="true"></i> Eventlog</a></li>
+            <li><a href="<?php echo(generate_url(array('page'=>'eventlog'))); ?>"><i class="fa fa-bookmark fa-fw fa-lg" aria-hidden="true"></i> Eventlog</a></li>
 <?php
 
 if (isset($config['enable_syslog']) && $config['enable_syslog']) {
-    echo '              <li><a href="'.generate_url(array('page'=>'syslog')).'"><i class="fa fa-book fa-fw fa-lg" aria-hidden="true"></i> Syslog</a></li>';
+    echo '              <li><a href="'.generate_url(array('page'=>'syslog')).'"><i class="fa fa-tag fa-fw fa-lg" aria-hidden="true"></i> Syslog</a></li>';
```
laf: I'm not sure on fa-tag for syslog

laf:
@librenms/reviewers anyone else want to chime in?

benwa:
Not a reviewer (mention-bot likes me), but personally I don't like font icons.  They're "easy to use" but is using the wrong tool for the job.  Font Awesome is also great, but doesn't really gear towards the tech icons that Libre kind of needs.

I don't want to say that [Material](https://material.io/icons/) icons are a better fit, but if you had a list of every icon currently being used, we might be able to find something that works.  That may be a lot of extra work for no real gain, maybe for v2?


For your original question though, I'd say `fa-terminal` or `fa-file-text-o` is better for syslog

laf:
All opinions welcome.

We've actually just ditched an icon set in favour of fa :) I personally prefer them (we're fa in v2 as well).

InsaneSplash:
fa-terminal is used for telnet icon
fa-file-text-o is used for notes and MIB

murrant:
@benwa we are definitely open to improvements.  Switching to fa was an improvement over what had.  There can still be more improvement :-)

PR #5721, opened 2017-02-01
Revert "fix: make mono theme panel headers black"

laf:
Reverts librenms/librenms#5705

Not sure this was intentional:

image

@murrant I'd like to revert this if you don't mind

### DIFF
--- a/html/css/mono.css
+++ b/html/css/mono.css
@@ -22,16 +22,6 @@
     color: black;
 }
 
-.panel-default>.panel-heading {
-  background: black;
-  color: white;
-}
-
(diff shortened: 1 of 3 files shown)


murrant:
That was intended.  Set the mono theme and look at the logon screen :-(

laf:
I'm not a fan personally but I've switched to the default theme now instead so it's not an issue for me.

murrant:
That's fine.  I started just changing the header and then saw it changed those, so I went with it.  Felt ok to me.  I can change just the logon header...

PR #5722, opened 2017-02-02
fix: Ignore ports where we only have two entries in the array, this signals bad data #1366

### LINKED ISSUE #1366: ~250 empty ports created for D-Link DGS-3100-48
I have done some basic troubleshooting with @laf regarding interfaces without any data from this switch, couldn’t find any good solution so let's start again from here.
1. DGS-3100 creates 250 empty ports in LibreNMS with no description, name or traffic data. Sometimes they get deleted but comes back after a while.
2. CPU/Memory graphs doesn't work. Maybe not supported from SNMP on this switch?

If you need SNMP access for troubleshooting please let me know.

laf:
DO NOT DELETE THIS TEXT

#### Please note

It's a little bit of a hack but it works. Fixes #1366

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -955,7 +955,10 @@ function is_port_valid($port, $device)
 
     global $config;
 
-    if (strstr($port['ifDescr'], "irtual") && strpos($port['ifDescr'], "Virtual Services Platform") === false) {
+    if (count($port) === 2) {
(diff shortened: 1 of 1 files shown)


laf:
Updated this to ignore ports without ifDescr, ifAlias and ifName. Tested by @tobiasa1

PR #5724, opened 2017-02-02
feature: support non-standard unix socket

murrant:
DO NOT DELETE THIS TEXT

#### Please note


laf:
One issue with build-base which I'll push a change for.

What about *-wrapper.py scripts?

laf:
Oh and html/install.php

laf:
Actually build-base.php needs updating to take config options from install.

murrant:
Need to check the python scripts.
I'm not sure about build-base.php, I thought it was ok as it is in this PR, we do not pull in dbFacile.php in that file at this time.

laf:
I'll have to re-run it but it complains about needing an int not a string for the port.

murrant:
OK, that's easy, I'll  check it.

includes/dbFacile.php
```diff
@@ -17,6 +17,23 @@
  * 3. Oh, and dbFetchAll() is now dbFetchRows()
  */
 
+function dbConnect($host = null, $user = '', $password = '', $database = '', $port = null, $socket = null)
```
laf: Does it not make more sense to not pass all these params and just call $config[] from within the function?

f0o: the installer & co doesnt have $config present at the time of execution AFAIK

murrant: Also, you aren't using this function anywhere.

PR #5732, opened 2017-02-02
newdevice: Exterity AvediaPlayer support

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

---------

- Add support for the Exterity AvediaPlayer units.

This a video converter to IP, we have no category for this kind of stuff... I used Appliance.

screen shot 2017-02-02 at 17 09 22
screen shot 2017-02-02 at 17 09 33


laf:
Unit tests are failing at the moment.

Can you add sysObjectId output to the unit test file as well so it's more complete.

includes/definitions/avr-hd.yaml
```diff
@@ -0,0 +1,9 @@
+os: avr-hd
+text: 'avr-hd'
+type: appliance
+icon: exterity
+over:
+    - { graph: device_bits, text: Traffic }
+discovery:
+    - sysDescrId:
```
laf: We really should try and use sysObjectId for detection if possible. If not it's sysDescr anyway

FTBZ:
Can you double check the code? I've noticed that all my devices are recognized as avr-hd and I don't know if it's since this PR or not (I've made some try after your message with the sysObjectId).

includes/definitions/avr-hd.yaml
FTBZ: SNMPwalk here: 
http://pastebin.com/raw/Rem5nv9d

Don't think that's possible?

murrant:
```
Discovery:
    sysObectID:
        - .1.3.6.1.4.1.28194.
```

laf:
I've just pushed some small changes. Will merge once travis clears.

includes/definitions/avr-hd.yaml
FTBZ: Ok thx. During my test I used the latest digit, it was my error.

PR #5738, opened 2017-02-02
bug: fixed sanitizing of google map information

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/includes/polling/functions.inc.php
+++ b/includes/polling/functions.inc.php
@@ -489,7 +489,7 @@ function location_to_latlng($device)
     $device_location = $device['location'];
     if (!empty($device_location)) {
         $new_device_location = preg_replace("/ /", "+", $device_location);
-        $new_device_location = preg_replace('/[^A-Za-z0-9\-\+]/', '', $device_location); // Removes special chars.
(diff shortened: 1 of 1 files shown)


includes/polling/functions.inc.php
```diff
@@ -489,7 +489,7 @@ function location_to_latlng($device)
     $device_location = $device['location'];
     if (!empty($device_location)) {
         $new_device_location = preg_replace("/ /", "+", $device_location);
-        $new_device_location = preg_replace('/[^A-Za-z0-9\-\+]/', '', $device_location); // Removes special chars.
+        $new_device_location = preg_replace('/[^A-Za-z0-9\-\+]/', '', $new_device_location); // Removes special chars.
```
laf: Is this right still? I'm not sure we should have merged the last one as we're basically replacing spaces with + and then getting rid of the + again.

InsaneSplash: That should replace everything that is not specified between the []'s? If one doesnt include the - or + it would remove it right?

laf: Ah sorry yes, forgot the ^ negative match

PR #5740, opened 2017-02-02
feature: Added Cisco Spark Transport

### LINKED ISSUE #3182: Alerting Settings - Cisco Spark
With the release of Cisco Spark APIs it would be great if we could see the type of intgration within librenms for slack and hipchat also for Cisco Spark?

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

Fix #3182


laf:
Updated the schema file as it fails with the default mysql config options now. Tested after that (other than sending the transport).

PR #5760, opened 2017-02-04
Improve netgear tests

murrant:
Currently this detects Switches, APs, Routers and ReadyNAS devices.
netgear os has little more than an icon and a poor attempt at platform info.
I'm not 100% sure what to do with this, but I do know they have different MIBS.

DO NOT DELETE THIS TEXT

#### Please note


laf:
Why not just leave it alone until we have the data to add the other specific OS' either from stats or a request from a user?

laf:
bump @murrant

murrant:
We can leave it alone.  I didn't add those other os's because I didn't take the time to research what MIBs, etc Netgear has... and what devices support them.

laf:
Mind if we close then?

PR #5767, opened 2017-02-04
bug: Updated NFSv3 Stats Application OID

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note


InsaneSplash:
One could also have used OID : .1.3.6.1.4.1.8072.1.3.2.4.1.2.9.110 but the one I included is more specific in case there is anything else living under there....

crcro:
i think that there is related to initial usage of nfs-stats (includes/polling/applications/nfs-stats.inc.php)

i've recommended sometime ago that initial version should be removed as it's obsolete, as far as i can see the original script was removed from librenms-agent but the webui files are still here:

- includes/polling/applications/nfs-stats.inc.php
- html/includes/graphs/application/nfs-stats.inc.php
some entries in: - html/includes/functions.inc.php

InsaneSplash:
Yeah I could remove the original nfs-stats... @laf ? what do you think?

laf:
I don't use this so don't really know but we merged in https://github.com/librenms/librenms-agent/pull/38/files which was for nfs v3. Does this PR make use of that now?

crcro:
yes, this one uses that.
the files i've posted previously are no longer used (they where for an older nfsstats, which is removed from agent repo).

laf:
So should this be merged?

crcro:
after @InsaneSplash removes those webui files that are not used

laf:
I'm not a user of apps to be able to test this. Code seems ok although we are changing the oid so do we know for sure this won't break things.

@librenms/reviewers thoughts?

murrant:
I would assume it will break existing users, we don't automatically update scripts on devices.

murrant:
No idea, why not check for both oids?

crcro:
the original version is not compatible with v3 due to different data processed by the snmp extend script.

i find it hard to believe someone is having a working nfs-stats apps (due to the fact that the snmp extend script is not available in librenms-agent and also original version would have some bugs in webui - due to various patches to the apps page).

InsaneSplash:
When I tried walking the original OID it displayed an error saying this OID was invalid...

laf:
Does anyone have a link to the origin nfs-stats agent file?

InsaneSplash:
```
#!/bin/bash
############################################################
# copy this file somewhere like /opt and chmod +x it       #
# edit your snmpd.conf and add the below line and restart: #
# extend nfs-stats /opt/nfs-stats.sh                       #
############################################################
CFG_NFSFILE='/proc/net/rpc/nfsd'
BIN_CAT='/usr/bin/cat'
BIN_SED='/usr/bin/sed'
BIN_AWK='/usr/bin/awk'
BIN_TR='/usr/bin/tr'
BIN_PASTE='/usr/bin/paste'
BIN_RM='/usr/bin/rm'
BIN_MV='/usr/bin/mv'
LOG_OLD='/tmp/nfsio_old'
LOG_NEW='/tmp/nfsio_new'
LOG_FIX='/tmp/nfsio_fix'

#get reply cache (rc - values: hits, misses, nocache)
$BIN_CAT $CFG_NFSFILE | $BIN_SED -n 1p | $BIN_AWK '{print $2,$3,$4}' | $BIN_TR " " "\n" > $LOG_NEW

#get server file handle (fh - values: lookup, anon, ncachedir, ncachenondir, stale)
$BIN_CAT $CFG_NFSFILE | $BIN_SED -n 2p | $BIN_AWK '{print $2,$3,$4,$5,$6}' | $BIN_TR " " "\n" >> $LOG_NEW

#get io bytes (io - values: read, write)
$BIN_CAT $CFG_NFSFILE | $BIN_SED -n 3p | $BIN_AWK '{print $2,$3}' | $BIN_TR " " "\n" >> $LOG_NEW

#get read ahead cache (ra - values: cache_size, 0-10%, 10-20%, 20-30%, 30-40%, 40-50%, 50-60%, 60-70%, 70-80%, 80-90%, 90-100%, not-found)
$BIN_CAT $CFG_NFSFILE | $BIN_SED -n 5p | $BIN_AWK '{print $3,$4,$5,$6,$7,$8,$9,$10,$11,$12,$13}' | $BIN_TR " " "\n" >> $LOG_NEW
$BIN_CAT $CFG_NFSFILE | $BIN_SED -n 5p | $BIN_AWK '{print $2}' > $LOG_FIX

#get server packet stats (net - values: all reads, udp packets, tcp packets, tcp conn)
$BIN_CAT $CFG_NFSFILE | $BIN_SED -n 6p | $BIN_AWK '{print $2,$3,$4,$5}' | $BIN_TR " " "\n" >> $LOG_NEW

#get server rpc operations (rpc - values: calls, badcalls, badfmt, badauth, badclnt)
$BIN_CAT $CFG_NFSFILE | $BIN_SED -n 7p | $BIN_AWK '{print $2,$3,$4,$5,$6}' | $BIN_TR " " "\n" >> $LOG_NEW

#get nfs v3 stats (proc3 - values: null, getattr, setattr, lookup, access, readlink, read, write, create, mkdir, symlink, mknod, remove, rmdir, rename, link, readdir, readdirplus, fsstat, fsinfo, pathconf, commit)
$BIN_CAT $CFG_NFSFILE | $BIN_SED -n 8p | $BIN_AWK '{print $3,$4,$5,$6,$7,$8,$9,$10,$11,$12,$13,$14,$15,$16,$17,$18,$19,$20,$21,$22,$23,$24}' | $BIN_TR " " "\n" >> $LOG_NEW

$BIN_PASTE $LOG_FIX
$BIN_PASTE $LOG_NEW $LOG_OLD | while read a b ; do
  echo $(($a-$b))
done

$BIN_RM $LOG_OLD 2>&1
$BIN_MV $LOG_NEW $LOG_OLD 2>&1

```

laf:
Isn't that just: https://raw.githubusercontent.com/librenms/librenms-agent/master/snmp/nfs-stats.sh which is the current one? I'm referring to the one that we no longer have apparently.

InsaneSplash:
That's the one i used when i updated the oid.

crcro:
thats the v3, the one which is compatible with latest webui changes (since wrote), has more stats etc

murrant:
InsaneSplash, I just rembered the oid is defined in snmpd.conf.  Can you paste yours here?

InsaneSplash:
I dont have any OID definitions in my snmpd.conf

```
com2sec readonly  default         community
group MyROGroup v2c        readonly
view all    included  .1                               80
access MyROGroup ""      any       noauth    exact  all    none   none
syslocation Space
syscontact Richard Mayhew

#Distro Detection
extend .1.3.6.1.4.1.2021.7890.1 distro /usr/bin/distro
extend apache /etc/snmp/apache-stats.py
extend ntp-client /etc/snmp/ntp-client.sh
extend osupdate /etc/snmp/os-updates.sh
extend rrdcached /etc/snmp/rrdcached
extend mysql /etc/snmp/mysql-stats.py

```

InsaneSplash:
Based on the distro example, maybe this should be configured to set a standard?

laf:
I don't know what to suggest here other than we just merge, I'd hate to break existing installs though

crcro:
@InsaneSplash ... you should have a line like:

extend nfs-stats /etc/snmp/nfs-stats.sh

strange i can't see any info about it in the docs, i remember adding them at the time ...

murrant:
@InsaneSplash. Yes, you need to update your snmpd.conf.  

We probably need to make sure the docs have the correct info.

InsaneSplash:
I do have the exact config line but it's not prepended to config line like the distro example with an oid.

I was wondering if this should be set to something to keep the oid the same across installs?

murrant:
This is a step in the right direction but I'm not sure if it is far enough.
Should we rename nfs-v3-stats to nfs-stats?
Might as well get it all cleaned up if we are going to need to send a notification to users...

crcro:
i think that we can leave it as nfs-v3-stats as there may be a need for nfs-v4-stats (meaning the nfs version 3 or 4 not regarding naming of the app)

murrant:
@crcro True, but would the stats be very different?

laf:
@InsaneSplash what are your thoughts on what murrant proposed above.

crcro:
@murrant: nfs-v4 has 50 extra datas from v3

murrant:
@crco, so v4 is a superset of v3...  I think merging them would be doable.

svennd:
I believe v4 is not a "superset" (meaning its not including the same data), its a different protocol; clients can talk to NFS over v2/v3 or v4. A server can have all three. Although most current Linux distro's by default use v3. Due to the changes in protocol more stats are being generated for v4. The script currently in the repository is just checking for v3 stats and general stats. (actually for v2 stats see https://github.com/librenms/librenms/pull/6304) 

nfs-v3-stats already incorporate stuff like file handlers, reply cache ...  it would be more logically to drop nfs-stats and extend nfs-v3-stats to include all stats found in /proc/net/rpc/nfsd
the problem there is, that they differ from kernel to kernel. For Centos 6/7 proc4 (nfsv4) have 3 value's and nfsv4ops ~22.

I am currently writing a script to read out these value's... never looking here. nfsv4 however is badly documented and kernel module reading is tough.

murrant:
Leave the OID, remove the "old" nfs stats rename the v3, update database, update docs...

murrant:
I feel like this is invalid now and we should try to migrate everyone to the new nfs-stats script.

PR #5778, opened 2017-02-05
Disco / Poller tidy up + Added display of module mem usage

laf:
DO NOT DELETE THIS TEXT

#### Please note

This mostly just tidies up a lot of variable and array use (unsets to clear up mem).

Also added in the display of per module mem usage.

Changed sensors polling to not poll all sensors types and only what the device has defined.

Changed variable names in some cisco modules as they used `$module` which was globally used and was then unset. This caused these poller module graphs to not be recorded.

### DIFF
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -140,11 +140,13 @@ function discover_device($device, $options = null)
             ($os_module_status && !isset($attribs['discover_' . $module])) ||
             ($module_status && !isset($os_module_status) && !isset($attribs['discover_' . $module]))) {
             $module_start = microtime(true);
+            $start_memory = memory_get_usage();
             echo "\n#### Load disco module $module ####\n";
             include "includes/discovery/$module.inc.php";
             $module_time = microtime(true) - $module_start;
             $module_time = substr($module_time, 0, 5);
-            printf("\n>> Runtime for discovery module '%s': %.4f seconds\n", $module, $module_time);
+            $module_mem  = (memory_get_usage() - $start_memory);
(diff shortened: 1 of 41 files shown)


includes/discovery/functions.inc.php
```diff
@@ -140,11 +140,13 @@ function discover_device($device, $options = null)
             ($os_module_status && !isset($attribs['discover_' . $module])) ||
             ($module_status && !isset($os_module_status) && !isset($attribs['discover_' . $module]))) {
             $module_start = microtime(true);
+            $start_memory = memory_get_usage();
             echo "\n#### Load disco module $module ####\n";
             include "includes/discovery/$module.inc.php";
             $module_time = microtime(true) - $module_start;
             $module_time = substr($module_time, 0, 5);
-            printf("\n>> Runtime for discovery module '%s': %.4f seconds\n", $module, $module_time);
+            $module_mem  = (memory_get_usage() - $start_memory);
```
murrant: Add memory_get_peak_usage() too?

murrant:
Looks good, I'd be interested to know the amount of improvement the changes make.

laf:
It varies based on the amount of data stored in the arrays, so ports for instance was one of the biggest savings (especially in discovery). Don't get me wrong none of these are huge savings in the scheme of things but it all helps slightly especially when running multiple concurrent poller.php processes.

Rosiak:
I'll see if I can test this.

exarv:
There are a few small bugs in the commit, caused by unix-agent / hddtemp disks to report suddenly 0 degrees.

I've created a pull request #5930 with fixes

PR #393, opened 2015-01-05
Added Hikvision support

laf:
This adds support for Hikvision CCTV camers. The snmp implementation is poor to say the least but this will at least now discover them so you can do up / down monitoring and have syslog / snmptraps sent.

PR #5781, opened 2017-02-06
refactor: Convert Hikvision discovery to yaml

murrant:
DO NOT DELETE THIS TEXT

#### Please note


laf:
I'm pretty certain (at least with some older firmware for these) that sysObjectId isn't available.

I have the cameras at home (but now use an NVR so can't snmp directly to them) but in the original PR: https://github.com/librenms/librenms/pull/393/files you can see that I did a specific check that no sysObjectId existed and then did an additional snmp call.

murrant:
Ok, please update or close this if you can check them, thanks :)

laf:
I can't :( It means removing them from my NVR and I'm that lazy right now :D

murrant:
haha, fair enough.  I could possibly change it to support both, not sure how though.

laf:
Sod it, the units are awful enough on snmp so let's just hope it's fixed in newer firmware.

PR #5792, opened 2017-02-06
Add (very) simple Toshiba TEC (printer) support

rockyluke:
DO NOT DELETE THIS TEXT

#### Please note


includes/definitions/toshiba.yaml
```diff
@@ -0,0 +1,14 @@
+os: toshiba
```
laf: I think we should change this to toshiba-printer, toshiba is too generic for a printer.

rockyluke: Good idea. It's fixed.

rockyluke:
I've switched from toshiba to toshiba-printer.

laf:
I've put the logo back as toshiba can remain generic for future devices.

laf:
Actually toshiba-tec seemed more appropriate. I've updated and add unit tests.

Issue #5634, opened 2017-01-27
Issue: Generex CS141 discovery


sirHulda:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.

Component | Version
--------- | -------
LibreNMS  | 3fb65da3396b9f94f36c7c109f69a6467131162e
DB Schema | 157
PHP       | 5.6.29-0+deb8u1
MySQL     | 5.5.54-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1

- [ ] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).

- [x] Please provide as much detail as possible.

Discovery doesn't work on my CS141 ups card. It works on my cs121 cards though. I checked sysobjectid and the output is "The Cs141 SNMP Agent". After changing "CS141" to "Cs141" in generex-ups.yaml the discovery worked fine.

- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

10.2.255.10 389 generic SNMP[/usr/bin/snmpget -v2c -c COMMUNITY -Ovq -M /opt/librenms/mibs udp:HOSTNAME:161 SNMPv2-MIB::sysDescr.0]
The Cs141 SNMP Agent

SNMP[/usr/bin/snmpget -v2c -c COMMUNITY -Ovqn -M /opt/librenms/mibs udp:HOSTNAME:161 SNMPv2-MIB::sysObjectID.0]
.*.2.1.33

| The Cs141 SNMP Agent | .1.3.6.1.2.1.33 |

sirHulda:
Hey,
Thanks for  the fix. Now it does discover properly but polling is no go.

Example:
"Checking (snmp) current Output Phase 1..."
[...]

PR #5794, opened 2017-02-06
update tripplite divisor for certain versions #5364

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5364

### DIFF
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -894,7 +894,15 @@ function get_device_divisor($device, $serial, $sensor)
         if ($sensor == 'current' || $sensor == 'frequencies') {
             if (version_compare($serial, '12.06.0068', '>=')) {
                 $divisor = 10;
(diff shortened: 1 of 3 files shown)


murrant:
BTW, I think the UPS attached to the SNMPcard determines the value.

UPS-MIB:upsIdentModel.0 (.1.3.6.1.2.1.33.1.1.2.0) seems show the UPS model maybe we can correlate that with the divisor?

laf:
Yeah it's a bit silly this way. This is standard rfc mib so these guys have just chosen to ignore the mib specs :(

FTBZ:
As mentioned in the IRC channel, a ticket was opened with Generex. Perhaps it's better to wait for the final response before trying to create a workaround.

laf:
I think it is.

FTBZ:
Issue #5634 was updated and this PR is no longer needed.

laf:
That's the wrong issue number, this fixes #5364 and the device is powerlite.

PR #5796, opened 2017-02-06
Add APC InfraStruXure Manager

rockyluke:
DO NOT DELETE THIS TEXT

#### Please note


rockyluke:
I don't understand where's the issue I only add one line.

laf:
It's just auto deploy that's failed so can be ignored.

This seems to be a separate device type and should be added as such (maybe as an appliance). Does it re-expose all of the information each APC unit would do or is it purely for detection?

rockyluke:
I close theus PR, I'll work more on that case.

Issue #5793, opened 2017-02-06
F5 Platform/Model not being discovered.


thejimmahknows:
Curious if anyone else is having this issue? I've added serveral F5 LTMs to test the Pool status feature, which works great! The only thing that isn't reporting back F5 related is the Platform/Model...I have direct access to these LTMs running codebase 11.5.1.

![screenshot from 2017-02-06 11-20-16](https://cloud.githubusercontent.com/assets/15802946/22657130/59d170c2-ec63-11e6-863e-a09361fdf1be.png)

![screenshot from 2017-02-06 11-51-29](https://cloud.githubusercontent.com/assets/15802946/22657136/60bdea8c-ec63-11e6-8d0e-7072f12a857c.png)

DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
./daily.sh 
Re-running /opt/librenms/daily.sh as librenms user
Updating to latest codebase                        OK
Updating SQL-Schema                                OK
Updating submodules                                OK
Cleaning up DB                                     OK
Fetching notifications                             OK

- [x ] Please include all of the information between the `====================================` 
section of `./validate.php` which you can run from the cli.
==========================================================
Component | Version
--------- | -------
LibreNMS  | aab1cdac49c97a9c5b28f21af572648861596f09
DB Schema | 159
PHP       | 5.6.29-0+deb8u1
MySQL     | 5.5.54-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1
==========================================================
- [x ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
[discovery-dcltm01.txt](https://github.com/librenms/librenms/files/755344/discovery-dcltm01.txt)
[poller-dcltm01.txt](https://github.com/librenms/librenms/files/755348/poller-dcltm01.txt)
[...]

PR #5797, opened 2017-02-06
Add F5 Hardware and S/N detection

geordish:
Fix for #5793 5793

DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/includes/polling/os/f5.inc.php
+++ b/includes/polling/os/f5.inc.php
@@ -1,4 +1,4 @@
 <?php
-$version = trim(snmp_get($device, '.1.3.6.1.4.1.3375.2.1.4.2.0', '-OQv', '', ''), '"');
-$hardware = snmp_get($device, '.1.3.6.1.4.1.3375.2.1.3.5.2.0', '-OQv', '', '');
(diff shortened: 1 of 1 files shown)


includes/polling/os/f5.inc.php
```diff
@@ -1,2 +1,4 @@
 <?php
-$version = trim(snmp_get($device, '.1.3.6.1.4.1.3375.2.1.4.2.0', '-OQv', '', ''), '"');
+$version = trim(snmp_get($device, '.1.3.6.1.4.1.3375.2.1.4.2.0', '-OQv'), '"');
+$hardware = snmp_get($device, '.1.3.6.1.4.1.3375.2.1.3.5.2.0', '-OQv');
```
murrant: You will need to trim quotes from these as well.

The other way around that is to use the MIB aka `F5-BIGIP-SYSTEM-MIB::sysGeneralChassisSerialNum.0`

geordish: Which would you prefer?

murrant:
No worries man, we all make mistakes, especially when you are trying to quick fix something ;)

That is what review is for.  Thanks for the PR!

Issue #5229, opened 2016-12-23
Add support for Turbo NAS TS-431


NK74:
- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php`.

| Component | Version|
| --- | --- |
| LibreNMS   | 5aef36abd653061db7fe547997bac8b87a9f10c6|
| DB Schema | 159 |
| PHP       | 7.0.8-3ubuntu3
| MySQL     | 10.0.28-MariaDB-0ubuntu0.16.10.1
| RRDTool   | 1.6.0
| SNMP      | NET-SNMP 5.7.3

- [x] If you would like us to add a new device or your device is listed as Generic then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)
- [x] Please provide as much detail as possible.

![image](https://cloud.githubusercontent.com/assets/17825769/21432838/9cedecce-c86d-11e6-9006-b8045caaabb1.png)

Turbo NAS TS-431 MIB
Discovery : http://pastebin.com/QKF6aT5C
Poller: http://pastebin.com/4P72gP6i
snmpbulkwalk: http://pastebin.com/fBknPQx4

http://pastebin.com/zSfqZJkX
http://www115.zippyshare.com/v/3lzdM7Vo/file.html

laf:
What are you actually asking for here? If this is to add detection for a device then you need to provide the information we ask for at the top.

NK74:
Newly added files 👍

NK74:
Up :)

NK74:
Hi,

Do you have domr news about my device ?

laf:
The snmpwalk doesn't seem to contain the info we need. Did you specify the `.` at the end of the command?

NK74:
Hi @laf ,

snmpbulkwalk : http://pastebin.com/wiK4v6rY

laf:
That's marked as private

NK74:
My apologies...

Try now it's "Unlisted"

laf:
It still doesn't contain things like sysDescr and sysObjectId - what command have you run?

NK74:
snmpbulkwalk -OUneb -v2c -c public 192.168.1.144 . > smpbulwalk.log

NK74:
[...]

PR #5804, opened 2017-02-06
newdevice: Added further QNAP Turbo NAS detection #5229

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5229

Adds a new option to os discover, it's a generic check (manufacturer name).

Chose to always have sysMfgName as regex compatible as I doubt people will search for the full thing here.


includes/functions.php
```diff
@@ -101,8 +101,9 @@ function getHostOS($device)
 
     $sysDescr    = snmp_get($device, "SNMPv2-MIB::sysDescr.0", "-Ovq");
     $sysObjectId = snmp_get($device, "SNMPv2-MIB::sysObjectID.0", "-Ovqn");
+    $sysMfgName  = snmp_get($device, "ENTITY-MIB::entPhysicalMfgName.1", "-Ovqn");
```
murrant: Will this always be .1, can it be .0?

murrant:
I'm unsure how I feel about this PR.  Something like this seems to be required in this case, but often using manufacturer name is obvious but often the wrong thing.  So I don't think we need to encourage it's use by making it easier.

I created a ticket with Qnap for the missing sysObjectID.

laf:
I'm not happy either but it's the only way I could think of getting detection done without going back to doing a php file.

includes/functions.php
laf: Here yes, we also use it in siklu discovery but in other vendors no it can be quit varied, Cisco seems to be [0-9]!

murrant:
I feel like I would prefer it as a php file for now.  You don't have to remove the sysObjectId detection from the yaml file.

laf:
Updated

filltr:
Plase add HDD temperatures.
HDD1 - 1.3.6.1.4.1.24681.1.2.11.1.3.1
HDD2 - 1.3.6.1.4.1.24681.1.2.11.1.3.2
HDD3 - 1.3.6.1.4.1.24681.1.2.11.1.3.3
HDD4 - 1.3.6.1.4.1.24681.1.2.11.1.3.4

NK74:
Hi @filltr , it's already mentioned here https://github.com/librenms/librenms/issues/5229#issuecomment-278596540

PR #5805, opened 2017-02-07
Improve ports polling

laf:
DO NOT DELETE THIS TEXT

#### Please note

Background: https://community.librenms.org/t/looking-for-testers-for-port-polling-update/524/1

Looking for code review.


geordish:
I like what you have done, and it looks fairly sane (although I've not ran it!)

Would checking for ifOper in your initial walk, and only collecting stats on those that are up be beneficial also?

laf:
Someone else mentioned that and I considered it before creating the PR last night. The issue I see is that the port could have just gone down 1 second before and that means we lose the last 5 minutes of data. It could flap or be mis-reported.

It's easy enough to add in but I wasn't sure it was a safe thing to do.

geordish:
That is a fair point.

What about storing the previous ifOper status of the port (We might already?) and not polling if it was down, and is currently down.

If it is up, or was up, then collect the stats and update to the new ifOper status?

laf:
We do indeed, both ifOper and ifAdmin so yes, an excellent idea :)

rucarrol:
ok, I enabled this on some QFabric and EX3300 stacks. While it's a bit early, I can see a VERY huge improvement (About 50%). 

`Someone else mentioned that and I considered it before creating the PR last night. The issue I see is that the port could have just gone down 1 second before and that means we lose the last 5 minutes of data. It could flap or be mis-reported.`


Yeah that was me. I understand the concern, however to count the point - you're assuming that ifoper and ifadmin are going to be constant across the poll. 

If you wanted to tighten the timing around then, then each port should have ifOper and ifAdmin polled _just before_ fetching port stats. 

Anyway, I'm very happy with this patch so far. I'll leave it run 1-2 days before enabling globally.

laf:
@geordish Actually just went do this update and then realised, we'd have to add ifOperStatus and/or ifAdminStatus to the list of full walks we do which is where things slow down slightly. What do you reckon, still worth it?

geordish:
Would those oids not need to be walked regardless? At least for alerting
etc?

laf:
Not walked but multi get. If the all ports were up then in theory this would be slower, if some were down then it would be quicker depending on how many.

I'll ask one of the testers to benchmark a walk so we can see.

laf:
This is possibly ok for a merge as is at present, the community post shows this working well and it's not enabled until someone chooses to.

laf:
Ok from one users test output and a back of a fag packet approach:

We take the total time to poll ports, and add 50% on (we do 4 walks and would be adding 2 more). 

Test 1. This ran the ports module in 19 seconds without the patch and 9 seconds with. We could expect that then to change to 13.5 seconds so still under the original but not as effective overall. This doesn't however take into account ports that are down and have been down for the last poll.

Test 2. This ran the ports module in 217 seconds without the patch and 40 seconds with. We could expect that then to change to 60 seconds so way under the original still.

Seems like it's worth doing that and tracking the port status, here I've assumed we should track ifOperStatus and ifAdminStatus as both would result in no data. It would mean we wouldn't pick up things like ifType changes but I think that's an acceptable compromise.

laf:
Ok updated.

geordish:
is ifAdmin strictly necessary? If a port is admin up, the Oper state may be up or down. If the port is admin down, the port will be down.

ifAdmin | ifOper | Port State
----------|--------|-----------
Up | Up | Up
Up | Down | Down
Down | Down | Down

The missing truth (right word here?) from that table is impossible.

Seems to be that the Port State will always be reliant on the ifOper state.

laf:
You're correct that's it not necessary as if the port is admin down then ifOper would be down (or should be), however having seen many vendors IF-MIB implementation I don't think we should rely on it + it looks cleaner so people know what it's doing rather than having to know what's intended.

philmd:
poller_perf

tested with Mikrotik RouterOS 6.37.4, working well.

murrant:
Any thoughts on using this automatically when we think it will be faster instead of having a config option?

Or do you want wait till it is better tested?

laf:
How do you propose we auto-enable it?

query ports table for count of total, deleted, ifOperStatus == ifOperStatus_prev and ifAdminStatus == ifAdminStatus_prev and if we have 50% of deleted or down ports turn it on?

murrant:
That is basically what I had in my mind.  Perhaps we add this as is, then we can collect some performance data and see how that would work later.  We can figure out where the performance tipping point roughly is.

laf:
Yeah the issue at the moment is we don't have the stats to show what's good/bad thresholds. 

I'm happy with the code, running it at work and 3/4 users have confirmed it  works for them.

PR #5806, opened 2017-02-07
Improve OS discovery tests data 

murrant:
DO NOT DELETE THIS TEXT

#### Please note

Add data collected from http://stats.librenms.org.  Try to avoid modifying discovery.  Dragonfly had no detection, so add that...


laf:
How come you've got rid of multimatic.yaml?

Aside from that and a blank line I've pushed an update for all good from me.

murrant:
multimatic was renamed to generex-ups.  The data was exactly the same for the devices, so we can't have both.  I'm not sure why we even had a multimatic.yaml file...

https://github.com/librenms/librenms/commit/4898617ef52bed41c36ad33afc3a0af5570484c7

Issue #5542, opened 2017-01-21
Request for adding new OS/MIB Cyberoam CRiV-1C virtual appliance


ursus69:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Component | Version
--------- | -------
LibreNMS  | eca168ea68446936c606d0ed6e29a209e63248eb
DB Schema | 156
PHP       | 5.5.9-1ubuntu4.19
MySQL     | 5.5.50-0ubuntu0.14.04.1
RRDTool   | 1.4.7
SNMP      | NET-SNMP 5.7.2

Hello I would like to have the Cyberoam CRi-1V Virtual appliance adde to LibreNMS so I can monitor the IPSec tunnels.
At this momment is recognized as a generic snmp device and only one IPSec tunnel shows when it has at least 5.
Maybe call the OS Cyberoam

I have attached the never expiring public links  from pastebin for:
[Discovery](http://pastebin.com/zsemajqf)
[Poller](http://pastebin.com/Fnc8TkY2)
[smnpbulkwalk](http://pastebin.com/R3j83rvj)
[Cyberoam MIB from vendor site](https://kb.cyberoam.com/redirfile.asp?id=5425&fstore=&SID=)

laf:
The walk doesn't contain any data from  .1.3.6.1.4.1.21067.2 which is the OID all of the MIB data is in, can you please provide the output of:

`snmpbulkwalk -OUneb -v2c -c COMMUNITY HOSTNAME .1.3.6.1.4.1.21067.2`

ursus69:
[...]

PR #5819, opened 2017-02-07
newdevice: Added support for Cyberoam UTM devices #5542

laf:
DO NOT DELETE THIS TEXT

#### Please note

#5542


murrant:
Nice one.

Btw, don't embed pngs in svg files ;)

laf:
I used inkscape to do the svg, had png in the first place :/

murrant:
@laf, you have to trace the png and more to convert it to actual SVG.  When you just import the PNG in Inkscape, it is just a PNG embedded in an SVG.

If you don't have an SVG readily available, it is probably not worth your time to convert it, just add the SVG.  Others may find or create SVG resources for it.  It took me about an hour to convert a logo that was much simpler than this one.

Issue #5723, opened 2017-02-02
IP address not displayed on Overview page


towster:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x ] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x ] Please provide as much detail as possible.
- [x ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

==========================================================
Component | Version
--------- | -------
LibreNMS  | 6bc7b1991e3602675364186aac35962ec038d114
DB Schema | 157
PHP       | 7.0.14
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
==========================================================

Recently added this device. LNMS does show the IP if you click on the icon to telnet to the device but it is not displayed on the overview page.

![capture](https://cloud.githubusercontent.com/assets/1965956/22533130/9f0d82c6-e8b1-11e6-9dba-301458a5dff2.PNG)

laf:
Is this using a proper dns name or /etc/hosts?

towster:
The device was added by IP.
There is no DNS or hosts entry for the device.
The SNMP data was not even configured properly.

laf:
Then we don't show an 'ip' field as that's when we do a lookup of the hostname to the IP.

towster:
Could we have it show the IP when DNS lookups fail?

laf:
If you add the device by IP, we don't do a dns lookup to get the IP so the 'IP' field is never populated. We could change that so we just copy it across I guess.

PR #5822, opened 2017-02-07
webui: Show device hostname as IP if force_ip_to_sysname enabled and no resolved IP #5723

laf:
DO NOT DELETE THIS TEXT

#### Please note

Tested with force_ip_to_sysname enabled and disabled, devices with hostnames and IPs and this works as expected.

Fixes: #5723

### DIFF
--- a/html/includes/dev-overview-data.inc.php
+++ b/html/includes/dev-overview-data.inc.php
@@ -30,16 +30,10 @@ echo '<tr>
         <td>'.$device['sysName'].' </td>
       </tr>';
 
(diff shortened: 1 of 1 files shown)


murrant:
@laf something like this would be a bit easier to follow.

```php
if (!empty($device['ip']) {
     echo "<tr><td>Resolved IP</td><td>{$device['ip']}</td></tr>"; 
} elseif ($config['force_ip_to_sysname'] === true)) {
     echo "<tr><td>Resolved IP</td><td>{$device['hostname']}</td></tr>"; 
}
```

laf:
Thanks murrant :)

Issue #5766, opened 2017-02-04
GUI Display Issue Unifi Wifi 


TyShawn76:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
==========================================================
Component | Version
--------- | -------
LibreNMS  | 424ac80bff355ebcb503d1730928e06cdb70f7af
DB Schema | 159
PHP       | 5.6.29-0+deb8u1
MySQL     | 5.5.54-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1
==========================================================
- [x] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.
Multiple sections are off in the GUI most are over Chrome and Safari and one is only over Safari.

The following section is in the Unifi wireless area, and the text is cut off over both browsers.
![screen shot 2017-02-04 at 6 50 58 am](https://cloud.githubusercontent.com/assets/21324034/22618504/038c367e-eaac-11e6-8e87-ecccecbc257a.png)
![screen shot 2017-02-04 at 6 51 14 am](https://cloud.githubusercontent.com/assets/21324034/22618505/03920e0a-eaac-11e6-8fe4-11df0fcc2e8f.png)

laf:
Can you try this patch please:

```
diff --git a/html/includes/graphs/device/ubnt_unifi_RadioCu_0.inc.php b/html/includes/graphs/device/ubnt_unifi_RadioCu_0.inc.php
index 3d0e246..3968db0 100644
--- a/html/includes/graphs/device/ubnt_unifi_RadioCu_0.inc.php
+++ b/html/includes/graphs/device/ubnt_unifi_RadioCu_0.inc.php
[...]

PR #5826, opened 2017-02-08
fix: Use generic_multi_line for unifi wifi graphs #5766

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5766 confirmed in the issue.


jquagga:
This doesn't work though.  The capacity used graphs are area graphs.  Moving them to line graphs means they don't stack on top of one another.  Basically instead of being a "how much is used gauge" they become a bunch of lines.

laf:
@jquagga Can you fix the initial issue then please?

jquagga:
Maybe, let me see what I can do in a couple of hours.  I'm not sure why generic multi pushes the TOC that far over.  That's why it "falls off the edge".

PR #5830, opened 2017-02-08
webui: New eventlog severity classification

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

The idea behind this change is : 

- Provide the ability to tag an event with a severity at the time it is written to the eventlog table
- Be able to modify the colours of the events displayed in one place.
- Not have to try run a regex against the text to determine the icon type.
- Provide 5 levels of severity (based on defcon) :)


InsaneSplash:
I have updated some of the log_event statements with a severity to kick things off in this PR.

sql-schema/161.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `librenms`.`eventlog` ADD COLUMN `severity` INT(1) NULL DEFAULT 0 AFTER `reference`;
```
laf: You can't specify the database here, not everyone uses librenms.

I also still think default should be 1:

<pre>`ALTER TABLE `eventlog` ADD COLUMN `severity` INT(1) NULL DEFAULT 1 AFTER `reference`;</pre>

includes/functions.php
```diff
@@ -796,7 +796,7 @@ function get_astext($asn)
 }
 
 # Use this function to write to the eventlog table
-function log_event($text, $device = null, $type = null, $reference = null)
+function log_event($text, $device = null, $type = null, $reference = null, $severity = 0)
```
laf: I think default should be 1.

alerts.php
```diff
@@ -349,13 +349,13 @@ function ExtTransports($obj)
             $prefix[4] = &$prefix[0];
             if ($tmp === true) {
                 echo 'OK';
-                log_event('Issued '.$prefix[$obj['state']]." for rule '".$obj['name']."' to transport '".$transport."'", $obj['device_id']);
+                log_event('Issued '.$prefix[$obj['state']]." for rule '".$obj['name']."' to transport '".$transport."'", $obj['device_id'],'','',1);
```
laf: Don't specify `''`, instead use `null`

laf:
Quite a few coding style fixes needed: https://travis-ci.org/librenms/librenms/jobs/199679004

InsaneSplash:
Ok. Let me run through these changes. 

If you don't think the default should be grey (0) I think it should rather be information (2) as we should keep ok (1) for OK events rather.

laf:
2 Is fine with me :)

html/includes/print-event-short.inc.php
```diff
@@ -7,15 +7,9 @@
...
-$icon_type = $icon_returned['icon'];
-$icon_colour = $icon_returned['colour'];
+$severity_colour = eventlog_severity($entry['severity']);
 
-if ($icon_type) {
-    $icon = "<i class='fa $icon_type fa-lg' style='color:$icon_colour' aria-hidden='true'></i>";
-} else {
-    $icon = "<i class='fa fa-bookmark-o fa-lg' style='color:black' aria-hidden='true'></i>";
-}
+$icon = "<i class='fa fa-bookmark fa-lg' style='color:$severity_colour' aria-hidden='true'></i>";
```
laf: actually sorry to be a pain, one last change. Rather than doing style='color' can you not just create styles in styles.css called log-$colourname and then tack log-darkgrey on the end of fa-lg?

InsaneSplash: Hmm worth a shot! Let me give it a bash, i don't mind.

InsaneSplash: OK, giving it a little thought, I dont think that would work as it would not allow for any future customization of the colours (mine might not be so great!). I think from what I can see we have 2 options,....

1. Leave it as is
2. Move the "style=color:xxx" attribute in to the function that determines the colour and just have the $severity_colour variable included?

laf:
Works for me :)

@librenms/reviewers

murrant:
Honestly, it looks like severity should be before reference in the function call.

If I get a chance later I can do that it is about a 30s refactor with phpstorm.

InsaneSplash:
Great! We will just need to revise the severity of each log_event depending on what the type is. There about about 90+ events that need classifying from what I initially looked at.

laf:
trigger ci

PR #5831, opened 2017-02-08
device: Improved hardware detection for Xerox (case issue)

rockyluke:
DO NOT DELETE THIS TEXT

#### Please note

Simply DES vs. Des.


includes/polling/os/xerox.inc.php
```diff
@@ -2,14 +2,21 @@
 
 // ...253.8.51.1.2.1.20.1 = STRING: "MFG:Xerox;CMD:Adobe PostScript 3,PCL;MDL:Phaser 4510N;CLS:Printer;DES:Xerox Phaser 4510 Laser Printer, PostScript 3, Letter/A4 Size"
 // ...253.8.51.1.2.1.20.1 = STRING: "MFG:Xerox;CMD:Adobe PostScript 3,PCL;MDL:ColorQube 8880DN;CLS:Printer;DES:Xerox ColorQube 8580 & 8880, Color Printer,PostScript 3,Letter/A4 Size "
+// ...253.8.51.1.2.1.20.1 = STRING: "MFG:Xerox; CMD:PCL, PJL, POSTSCRIPT; MDL:WorkCentre 7535; CLS:Printer; Des:Xerox WorkCentre 7535"
 $xinfo = explode(';', trim(snmp_get($device, '1.3.6.1.4.1.253.8.51.1.2.1.20.1', '-OQv', '', ''), '" '));
 
 foreach ($xinfo as $xi) {
     list($key,$value) = explode(':', trim($xi));
```
laf: Would it not be easier to just run http://uk1.php.net/manual/en/function.strtoupper.php on `$key` before it's used in `$xerox` then you don't need to worry about case?

rockyluke: Yes, easier and cleaner !

Issue #5811, opened 2017-02-07
SNMP contact is not updated in webgui when removed from the device.


eikik87:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [ x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x ] Please provide as much detail as possible.
- [x ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

==========================================================
Component | Version
--------- | -------
LibreNMS  | 5aef36abd653061db7fe547997bac8b87a9f10c6
DB Schema | 159
PHP       | 5.5.9-1ubuntu4.20
MySQL     | 5.5.54-0ubuntu0.14.04.1
RRDTool   | 1.4.7
SNMP      | NET-SNMP 5.7.2
==========================================================

http://pastebin.com/UG8DkNHj

I have a Cisco router.

I put in this command - "snmp-server contact none@none.com"

After running debug command "./discovery.php  -m sensors  -h HOSTNAME -d" I get the contact instantly into the webgui.

But when I remove the contact, it does not get updated and the old 'none@none.com' contact is stuck. And only way I can "remove" it, is by putting in some random words.

This might be issue as I'm planning on having email of our customer and if they would not want to have email anymore, I would remove this contact command from the device. Instead of having librenms trying to send email to random words I put into the contacts. It would be better to have it empty.

PR #5836, opened 2017-02-08
Various variables will all be updated if they are blank #5811

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5811


murrant:
My only concern is if one of these variables is null or something and the database returns ''.  Causing it to b e updated every poll.

I'll try to test this, but my testing bandwidth has been pretty low of late :(

laf:
```
-bash-4.2$ cat /tmp/n.php
<?php

$t = null;
$s = '';

if ($t != $s) {
    echo 'notequals';
} else {
    echo 'equals';
}
-bash-4.2$ php /tmp/n.php
equals
```

Should be ok

murrant:
Ah yes, you aren't using strict comparison.

laf:
Yeah I only use strict when I know or trust the data

PR #5837, opened 2017-02-08
feature: adds snmptable command and 'ports_snmptable' config option

VimCommando:
DO NOT DELETE THIS TEXT

#### Please note

- [✔️] Have you signed the [Contributors agreement](http://docs.librenms.org/General/Contributing/) - please do NOT submit a pull request unless you have (signing the agreement in the same pull request is fine). Your commit message for signing the agreement must appear as per the docs.
- [✔️] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)


laf:
Interesting concept. out of interest why use tables? A quick benchmark on a device with circa 48 ports shows no noticeable performance improvements.

murrant:
Isn't snmptable just a different output format than snmpget/walk?

murrant:
No response, closing.

VimCommando:
So, I found two major differences with snmptable vs. snmpwalk/bulkwalk.

1.  If you are using snmp v1, you can forget about bulkwalk, it is not supported. However *snmptable* will work through snmp v1, with much of the speed benefit.
2. If you are dealing with devices that have less than stellar support for snmpbulkwalk *cough*mikrotik*cough* or broken implementations for specific OIDs, it gives you an alternate method of pulling the data without resorting to a full walk.

The device in this attached screenshot is a great case study.

screen shot 2017-02-07 at 09 41 39

This is a 24-port Mikrotik hotspot controller that manages and apartment building with a *lot* of access points. We average 150+ wireless clients connected at any given time. Each apartment is setup on a separate VLAN that is captured as a port entry--we want all of the data collected from the *ports* module, but it was causing serious performance issues (check the 404.20 second max time!) The big change on Feburary 2nd, where the average polling time dropped from ~120sec to ~80 sec, was when I added an option to the *ports* module to use this *snmptable* command. (not included in this commit)

On a device that implements both *snmptable* and *snmpbulkwalk* correctly, the performance difference will be negligible. I found times when table was faster, and others when bulkwalk was faster; it was dependent mainly on the *max_repeaters* setting. Whatever the default is gives the edge to table, and once you've increased it bulkwalk can take the lead.

The first time I used the *snmptable* command was to build this graph:
hotspotusers

This is this week's data from that same device. I needed to graph how many uses are connected to our hotspot controllers at any given time for historical reference. This is not a data point available anywhere in the Mikrotik MIB. However, data on each connected user is stored in an ActiveHotspotUsers table. So I could just pull the table and count the number of rows. It only added 3-5 seconds of polling time, which was acceptable on this device. On smaller devices it was <2 seconds, and gave us the ability to graph what would've otherwise required the experimental MIB polling modules (which we tried with... disastrous results.)

laf:
It's certainly an extremely interesting concept. I'll do some testing on devices we've got with large number of ports to see.

This will need rebasing though but I'd wait. In the time since this was done we've added the ability to only poll ports which are not down and disabled/deleted to help with performance in the ports polling.

VimCommando:
@laf In this case the hardware only has 24 physical ports. However, every apartment has a "VLAN Subif" listed as separately. Since each port (including VLAN) reports performance data, the ports tab ends up with 159 entries and 154 *with active traffic*. In the last 24 hours it handled had over 2 TB, which is pretty typical:
24hourtraffic

The ability to improve polling performance when using SNMP v1 or on devices with poor/broken bulkwalk support was why I felt it was worth a PR. In most cases, bulkwalk does the job. Unfortunately for us we rely heavily on devices from this manufacturer (Mikrotik), so our server was hitting CPU usage limits from all the long-running LibreNMS' pollers. So this helped reduce our average CPU usage as well.

f0o:
@VimCommando Any chance you could resolve these conflicts?

VimCommando:
@f0o I'll try to get them resolved.

f0o:
@VimCommando Any updates? :)

laf:
Feel free to re-submit this when you've resolved the conflicts.

PR #5838, opened 2017-02-08
feature: Added ability to output more info for snmp-scan.php

laf:
DO NOT DELETE THIS TEXT

#### Please note

https://community.librenms.org/t/snmp-scan-php-more-info/546


murrant:
I usually use -v -v for something like this (then with getopt() you set the verbosity by count($options['v]).

No problem with the way it is done here, just wondering if you might like using multiple -v options to increase verbosity.

laf:
-v adds all the output from db queries and snampwalk. The requisite just wants to see what hosts were tried.

PR #5847, opened 2017-02-09
allow interface names > 64 chars

sorano:
As discussed on IRC, this will allow interface names up to 255 chars to prevent eventlog updates from rinse and repeat ad infinitum

DO NOT DELETE THIS TEXT

#### Please note


InsaneSplash:
Do you think we should rather limit it to 128 characters so it doesn't wrap all over the page. It is strange though to have such a long name for an interface. ☺

laf:
The mib definition says it can be 255 so we need to match that. Thanks @sorano

PR #5848, opened 2017-02-09
Initial Mikrotik/RouterOS WiFi Stuff 

geordish:
Continuation of #5509 

I've added a commit experimenting with extending discovery for WiFi stuff.

Should give you an idea about what I'm thinking... there is a lot of junk there mind.

Thoughts?

DO NOT DELETE THIS TEXT

#### Please note


barryodonovan:
re the original post about number of radios in the original issue thread,

I had the same issue with ceragon (ceraos) radios, various models could have between 1 and 4 (but actually no limit as there is blade style rack available) radios.

I used a hack to put the count of radios into the $device table and then used that at various places to determine how many radios to poll / graph.

I'd like to make it cleaner if we can create a common architecture for wireless devices since number of radios does vary but for now something similar might help with mikrotiks.

murrant:
trigger ci

geordish:
This isn't ready to merge yet. waiting feedback on my approach to collecting WiFi stats.

Also @barryodonovan's comments make me doubt how I am displaying the stats...

geordish:
@librenms/reviewers any thoughts?

laf:
Not on the wifi side of things at least but few things code wise.

Don't use `file_exists` use `rrdtool_check_rrd_exists` instead.

The sensors discovery code shouldn't be used like it is here. We're fudging something in which either belongs in it's own module entirely or needs to be made to work with the way sensors do now.

murrant:
@geordish how about a radios table?  you can relate it by device_id and port_id if available.

geordish:
Will think about this some more, and reopen in the future.

Issue #5494, opened 2017-01-17
Remove Unwanted Modules Automatically Upon Detection


Lonecrowe:
Many OS'es will automatically enable all the services and modules per device.  It then creates dummy files that are never used or graphed in each device. 

BGP, OSPF etc
Cisco based modules
Juniper etc
Printer toner 

It should know that if it is a Cisco it does not need to add printer toner or Juniper based modules! 

I'll extrapolate more specifically what we need later.  

For now I'd like to start with the cambium / canopy OS and make sure everything is disabled except the interface/ports and the actual wireless modules.

laf:
imho we should tackle this purely by enabling only the modules needed for the device type. We've started on this already.

One thing we could do would be to disable polling modules when discovery finds no data. The benefits won't be that great though.

laf:
moved to https://community.librenms.org/t/remove-unwanted-modules-automatically-upon-detection/666

PR #5851, opened 2017-02-09
feature: Discovery modules will now auto-disable poller modules when no data found #5494

laf:
DO NOT DELETE THIS TEXT

#### Please note

Throwing this out there for feedback rather than merging.

One issue I can think of is if someone turns off the sensors polling module but leaves disco on, if discovery finds data we will turn the module back on.

murrant talked about linking both disco / poller modules together so you toggle them both on / off at the same time (from a users perspective) which would help with the above.

Thoughts @librenms/reviewers

### DIFF
--- a/includes/discovery/bgp-peers.inc.php
+++ b/includes/discovery/bgp-peers.inc.php
@@ -10,6 +10,7 @@ if ($config['enable_bgp']) {
 
     $bgpLocalAs = trim(snmp_walk($device, '.1.3.6.1.2.1.15.2', '-Oqvn', 'BGP4-MIB'));
 
+    $found_data = 0;
     foreach ($vrfs_lite_cisco as $vrf) {
             $device['context_name'] = $vrf['context_name'];
         if (is_numeric($bgpLocalAs)) {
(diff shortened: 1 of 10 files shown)


wriedel:
Hi Folks, I think it would be beneficial to have some tweaking especially if we're heading for 1 minute fast polling. I guess it would make sense to have something like enable modules per device group or bake more intelligence into the polling. Why even try polling for OTV on an ASA or a Linux server or IPMI on a virtual machine?
Also I just wonder if it's feasible to group devices or modules by OID response time. Like while polling IPSLA every minute don't do that for my full BGP table.

murrant:
Looks good to me, but I worry about modules that might not have data at discovery, but could change later, like disks or maybe bgppeers.

Stuff like memory pools and entity-physical should be fine no matter what.

laf:
I don't think this is in anyway ready for merging btw it was merely to invoke more discussion as it will have issues where people disable polling modules specifically and they get re-enabled.

With regards to data not available at disco, not sure this is an issue as most poller modules don't discover what they should be doing - the exception is things like ports which can be discovered through polling as well).

PR #5858, opened 2017-02-10
Canopy frame utilization and updated MIBs

barryodonovan:
DO NOT DELETE THIS TEXT

#### Please note

I've added frame utilization graphing and the updated MIBs to poll for it.
I think that's all for (me for) now on canopy_generic 
There are a lot more useful metrics but if there is going to be a big refactoring of wireless devices then these could be part of the mix.


murrant:
@barryodonovan Actually, if you know what kind of stats you are looking for on Canopy devices, can you list them on the community post to help out with the refactor, because we are a lacking a lot of info in that area.

https://community.librenms.org/t/improve-wifi-collection-and-graphing-code/453

barryodonovan:
Yup, will do. Just trying to catch up with getting LibreNMS into production.
I lost January to everything that went wrong over christmas when I took time off! ;-)

includes/polling/wireless/canopy-generic.inc.php
```diff
@@ -290,4 +290,21 @@
         $graphs['canopy_generic_freq'] = true;
         unset($rrd_filename, $freq);
     }
+
+    $downlinkutilization = str_replace('"', "", snmp_get($device, "frUtlLowTotalDownlinkUtilization.0", "-Ovqn", "WHISP-APS-MIB"));
```
murrant: Do you need the str_replace here?  You should get nicely formatted data here and not need that. (and the line below)

murrant:
Also, can you `chmod 644` the mib files you added.  Thanks.

includes/polling/wireless/canopy-generic.inc.php
barryodonovan: yeah, just kept things the same as the existing code so as not to confuse someone who's looking at it later, The entire thing need to be rewritten to use a multiget, I was just trying to shoehorn my requirements in within the time window I had.

I'll see if I can find time to rewrite today.

PR #5860, opened 2017-02-10
Tidy up tripplite os polling.

murrant:
DO NOT DELETE THIS TEXT

#### Please note


laf:
How come the MIB names have been moved to he oid? I'm not a fan of that way personally

murrant:
Oh, I am, allows for easy mixing of MIBs, but can change it back.  I was looking into my poweralert devices to see why they didn't have platform info.  Turns out they just suck and don't answer on any of these oids...

laf:
You can just pass multiple mibs with IF-MIB:SNMPv2-MIB

murrant:
True, but I like the full OID name, it probably saves NetSNMP from searching for the oid.

PR #5865, opened 2017-02-11
refactor: small function cleanups

murrant:
found while working on my snmp refactor
mac_clean_to_readable is about 33% improvement
hex2str is a moderate improvement

DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -65,14 +65,7 @@ function array_sort($array, $on, $order = SORT_ASC)
 
 function mac_clean_to_readable($mac)
 {
-    $r = substr($mac, 0, 2);
-    $r .= ":".substr($mac, 2, 2);
-    $r .= ":".substr($mac, 4, 2);
-    $r .= ":".substr($mac, 6, 2);
-    $r .= ":".substr($mac, 8, 2);
-    $r .= ":".substr($mac, 10, 2);
-
-    return($r);
+    return rtrim(chunk_split($mac, 2, ':'), ':');
 }
 
 function only_alphanumeric($string)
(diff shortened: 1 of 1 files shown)


laf:
I'm never sure if we should be test all functions but this one seems like we should as we rely on the output. If you agree then:

```
diff --git a/tests/CommonFunctionsTest.php b/tests/CommonFunctionsTest.php
index 9041d43..70da2e9 100644
--- a/tests/CommonFunctionsTest.php
+++ b/tests/CommonFunctionsTest.php
@@ -83,4 +83,10 @@ class CommonFunctionsTest extends \PHPUnit_Framework_TestCase
         $data = 'Toner, S/N:CR_UM-16021314488.';
         $this->assertEquals('Toner, S/N CR_UM-16021314488.', safedescr($data));
     }
+
+    public function testCleanMac()
+    {
+        $this->assertEquals('00:00:0c:9f:f1:f8', mac_clean_to_readable('00000c9ff1f8'));
+    }
+
 }

```

If not just merge it as I've tested the mac_clean_to_readable() function which works as expected.

hex2str() isn't used anywhere so we could just bin it off......

murrant:
I actually used these in my snmp rewrite, Instead of committing there, I pushed it upstream.  I can add tests.

laf:
It was more a question of should we have tests for things like this? I've included a working test above unless you want to extend it further.

Either way feel free to merge this :)

murrant:
The amount of refactoring for functions like this is likely small, but also the amount of time added by the test is extremely minimal.

murrant:
Huh, I thought I pushed tests to this...

PR #5866, opened 2017-02-11
poller: Linux Server hardware detection

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

```
/etc/snmpd.conf
extend .1.3.6.1.4.1.2021.7890.2 hardware '/usr/sbin/dmidecode -s system-product-name'
extend .1.3.6.1.4.1.2021.7890.3 manufacturer '/usr/sbin/dmidecode -s system-manufacturer'
extend .1.3.6.1.4.1.2021.7890.4 serial '/usr/sbin/dmidecode -s system-serial-number'
```


includes/polling/os/unix.inc.php
```diff
@@ -43,9 +43,30 @@
         $hardware = "Dell " . $hw;
     }
 
+    # Try detect using the extended option (dmidecode)
+    if (!$hw) {
```
laf: May as well just do `} else {` with the above if statement here.

includes/polling/os/unix.inc.php
```diff
@@ -43,9 +43,30 @@
...
+        $version_dmide = str_replace("\"", "", snmp_get($device, ".1.3.6.1.4.1.2021.7890.2.4.1.2.8.104.97.114.100.119.97.114.101.1", "-Oqv", "NET-SNMP-EXTEND-MIB"));
```
laf: You've specified the MIB to use but have a numeric OID, use the named OID instead.

includes/polling/os/unix.inc.php
```diff
@@ -43,9 +43,30 @@
...
+        $version_dmide = trim(str_replace("\"", "", $version_dmide));
+
+        $hardware_dmide = str_replace("\"", "", snmp_get($device, ".1.3.6.1.4.1.2021.7890.3.4.1.2.12.109.97.110.117.102.97.99.116.117.114.101.114.1", "-Oqv", "NET-SNMP-EXTEND-MIB"));
```
laf: You've specified the MIB to use but have a numeric OID, use the named OID instead.

includes/polling/os/unix.inc.php
```diff
@@ -43,9 +43,30 @@
...
+        $hardware_dmide = trim(str_replace("\"", "", $hardware_dmide));
+        if ($hardware_dmide) {
+            $hardware = $hardware_dmide;
```
laf: Why not just assign `$hardware` to the str_replace call above?

includes/polling/os/unix.inc.php
```diff
@@ -43,9 +43,30 @@
...
+            }
+        }
+    }
+
     $serial = snmp_get($device, ".1.3.6.1.4.1.674.10892.1.300.10.1.11.1", "-Oqv", "MIB-Dell-10892");
     $serial = trim(str_replace("\"", "", $serial));
 
+    # Try detect using the SNMP Extend option (dmidecode)
+    if (!$serial) {
+        $serial = str_replace("\"", "", snmp_get($device, ".1.3.6.1.4.1.2021.7890.4.4.1.2.6.115.101.114.105.97.108.1", "-Oqv", "NET-SNMP-EXTEND-MIB"));
```
laf: You've specified the MIB to use but have a numeric OID, use the named OID instead.

includes/polling/os/unix.inc.php
InsaneSplash: I was trying to not replace the pre-built $hardware variable above with a blank one if the dmidecode smpget failed to retrieve anything.

includes/polling/os/unix.inc.php
InsaneSplash: bit confused here as I used the same format as all the other distro extended snmp calls? should I remove the "NET-SNMP-EXTEND-MIB" reference?

includes/polling/os/unix.inc.php
laf: Fair enough

includes/polling/os/unix.inc.php
laf: If you use numerical oids then yes as they don't need a mib to do lookups

Issue #5862, opened 2017-02-10
OS reverts to generic when the device is unreachable


murrant:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

I'm not 100% sure when this happens, but I am putting this here as a reminder.  Or if anyone else wants to investigate.

PR #5870, opened 2017-02-12
refactor: Centralise device up/down check and use in disco #5862

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5862 

The reason the icon/os changed before was that the response from getHostOS() returned generic if no specific match was found. As a device not responding to snmp didn't return a match we then changed os/icon to the newly found generic.

I've centralised the code into function from poller which decides if the device is up/down before polling carried on, we now do the same in disco. No point in running through a full discovery for a device that's down.


includes/discovery/functions.inc.php
```diff
@@ -102,78 +102,83 @@ function discover_device($device, $options = null)
...
 
-    if ($device['os'] == 'generic') {
-        // verify if OS has changed from generic
-        $device['os'] = getHostOS($device);
+    $response = device_is_up($device, true);
 
-        if ($device['os'] != 'generic') {
-            echo "\nDevice os was updated to " . $device['os'] . '!';
-            dbUpdate(array('os' => $device['os']), 'devices', '`device_id` = ?', array($device['device_id']));
+    if ($response['status'] == '1') {
```
murrant: Can you do:
```php
if ($response['status'] !== '1') {
    return;
}
```
I like to avoid long if blocks, they make things harder to read.

laf:
I need to redo anyway and get rid of the updates to /vendor

murrant:
We should probably update those anyway (but in their own commit)

includes/functions.php
```diff
@@ -1993,3 +1993,50 @@ function recordSnmpStatistic($stat, $start_time)
...
 }
+
+/**
+ * @param $device
+ * @param bool $record_perf
+ * @return array
+ */
+function device_is_up($device, $record_perf = false)
+{
+    global $config;
```
murrant: This seems to be unused.

includes/functions.php
```diff
@@ -1993,3 +1993,50 @@ function recordSnmpStatistic($stat, $start_time)
...
+
+    $address_family = snmpTransportToAddressFamily($device['transport']);
+
+    $ping_response = isPingable($device['hostname'], $address_family, $device['attribs']);
+
+    $device_perf              = $ping_response['db'];
+    $device_perf['device_id'] = $device['device_id'];
+    $device_perf['timestamp'] = array('NOW()');
+
+    if ($record_perf === true && can_ping_device($device['attribs']) === true && is_array($device_perf)) {
```
murrant: `is_array($device_perf)` is unescesary

includes/functions.php
```diff
@@ -1993,3 +1993,50 @@ function recordSnmpStatistic($stat, $start_time)
...
+        dbInsert($device_perf, 'device_perf');
+    }
+
+    $response             = array();
+    $device['pingable']   = $ping_response['result'];
```
murrant: This will not update the global $device array(), just replace `$device['pingable']` below with `$ping_response['result']` and drop this line.

includes/functions.php
```diff
@@ -1993,3 +1993,50 @@ function recordSnmpStatistic($stat, $start_time)
...
+    $response['ping_time']= $ping_response['last_ping_timetaken'];
+    if ($device['pingable']) {
+        $device['snmpable'] = isSNMPable($device);
```
murrant: same as pingable

murrant:
I know you just copy and pasted the code, but it is a good time to give it a once over ;)

laf:
Closing in favour of #5897 

Should take into account your comments @murrant

PR #5879, opened 2017-02-13
webui: Add additional eventlog severity levels

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note


InsaneSplash:
This is a start to provide a severity for each log_event. From this point we may need to review the severity level attached to each event and modify if required. I think I have set up a good base. but fear that most events are seen as notices......

thoughts?

Issue #5809, opened 2017-02-07
Blank Interfaces in NMS


mr-ravirajput:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ x] Please provide as much detail as possible.
- [x ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Hello @laf 

I am re-creating the issue which was closed here

https://github.com/librenms/librenms/issues/5539#issuecomment-274295238

All the required details are now given

![blank interfaces](https://cloud.githubusercontent.com/assets/24430380/22686238/51517208-ed49-11e6-92dc-3baf695c3f41.png)
![validate php](https://cloud.githubusercontent.com/assets/24430380/22686237/514e99b6-ed49-11e6-8362-4a5d7372c639.png)

[discovery logs.docx](https://github.com/librenms/librenms/files/757173/discovery.logs.docx)
[poller logs.docx](https://github.com/librenms/librenms/files/757174/poller.logs.docx)
[snmp logs.docx](https://github.com/librenms/librenms/files/757175/snmp.logs.docx)

discovery

https://gist.github.com/mr-ravirajput/184b6e751447ad2af1b5e11f97c461a7

Poller log
https://gist.githubusercontent.com/mr-ravirajput/1221f0dbb60db0780a600bbc51f56e79/raw/e828909c5c0f2e3ee7fd29b17265acac002ff0b0/Blank%2520Interfaces%2520poller%2520log.rtf

snmp logs

[...]

PR #5882, opened 2017-02-13
tidy up some ports polling/disco #5809

laf:
DO NOT DELETE THIS TEXT

#### Please note

I made these changes when debugging #5809 and whilst I don't think was the full cause of the issue they certainly shouldn't be an issue in changing them.

FIxes: #5809

### DIFF
--- a/includes/discovery/ports.inc.php
+++ b/includes/discovery/ports.inc.php
@@ -42,6 +42,8 @@ foreach ($port_stats as $ifIndex => $port) {
     // Store ifIndex in port entry and prefetch ifName as we'll need it multiple times
     $port['ifIndex'] = $ifIndex;
     $ifName = $port['ifName'];
+    $ifAlias = $port['ifAlias'];
+    $ifDescr = $port['ifDescr'];
 
     // Get port_id according to port_association_mode used for this device
     $port_id = get_port_id($ports_mapped, $port, $port_association_mode);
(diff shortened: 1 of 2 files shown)


includes/discovery/ports.inc.php
```diff
@@ -42,19 +42,21 @@
...
         // Port newly discovered?
         if (! is_array($ports_db[$port_id])) {
-            $port_id         = dbInsert(array('device_id' => $device['device_id'], 'ifIndex' => $ifIndex, 'ifName' => $ifName), 'ports');
+            $port_id         = dbInsert(array('device_id' => $device['device_id'], 'ifIndex' => $ifIndex, 'ifName' => $ifName, 'ifDescr' => $ifDescr, 'ifAlias' => $ifAlias), 'ports');
             $ports[$port_id] = dbFetchRow('SELECT * FROM `ports` WHERE `device_id` = ? AND `port_id` = ?', array($device['device_id'], $port_id));
             echo 'Adding: '.$ifName.'('.$ifIndex.')('.$port_id.')';
         } // Port re-discovered after previous deletion?
         elseif ($ports_db[$port_id]['deleted'] == '1') {
-            dbUpdate(array('deleted' => '0'), 'ports', '`port_id` = ?', array($port_id));
+            dbUpdate(array('deleted' => '0'), 'ports', "`port_id` = ?, `ifName` => '?', `ifDescr` => '?', `ifAlias` => '?'", array($port_id, $ifName, $ifDescr, $ifAlias));
```
ekoyle: Does mysqli not handle quoting?

ekoyle: Also, it looks like you are using `=>` in an SQL statement; should those be moved to the first array instead? (ie. set the values instead of adding to the condition)

laf: Thanks for spotting that.

As for the quotes, we've had it recently where dbUpdates with strings weren't being quoted so people were getting an error in mysql and had to be updated: https://github.com/librenms/librenms/pull/5842/files

I don't know why so I've took the safe option here for these strings.

laf: Actually I've removed it completely as we do some checks in polling for those variables so we may just end up changing them constantly.

PR #5885, opened 2017-02-14
tidy up some more eventlog things

laf:
DO NOT DELETE THIS TEXT

#### Please note

Few log_event calls that needed fixing + also the schema hadn't applied for me due to running sql_mode with certain options. I've fixed that with a new schema file and also added the same fix to the previous schema for when we allow people to run mysql strict mode.


sql-schema/163.sql
```diff
@@ -1 +1,2 @@
+ALTER TABLE `eventlog` CHANGE `datetime` `datetime` DATETIME NOT NULL DEFAULT '1970-01-01 00:00:01';
```
murrant: Is this query supposed to be in both 163.sql and 164.sql?

laf: Yes, when we tell people they don't have to turn off sql_mode anymore then this schema without the change would break as you can't modify the table until the datetime is fixed.

murrant: Ok

PR #5886, opened 2017-02-14
added local radios Rx level and graph to ubnt airfibre

barryodonovan:
DO NOT DELETE THIS TEXT

#### Please note

Ubiquiti Airfibre radios:
Added local radio Rxlevels to mib oid array.
Added graph for local Rxlevels.

In order to get it to graph it is necessary to add new DS to RRD, either by deleting existing RRD or by using rrdtune (if rrd >1.5.0)

rrdtool tune PATH-TO-RRD/RRD-FILENAME.RRD DS:rxPower0:GAUGE:600:U:100000000000 DS:rxPower1:GAUGE:600:U:100000000000 RRA:AVERAGE:0.5:1:2016 RRA:AVERAGE:0.5:6:1440 RRA:AVERAGE:0.5:24:1440 RRA:AVERAGE:0.5:288:1440 RRA:MIN:0.5:1:720 RRA:MIN:0.5:6:1440 RRA:MIN:0.5:24:775 RRA:MIN:0.5:288:797 RRA:MAX:0.5:1:720 RRA:MAX:0.5:6:1440 RRA:MAX:0.5:24:775 RRA:MAX:0.5:288:797 RRA:LAST:0.5:1:1440

ref: https://community.librenms.org/t/airfiber-radio-rx-level/529/8


barryodonovan:
@laf @murrant 
I really think it would be worthwhile to put out a notification to see if this change will actually affect anyone.
The workaround (a seperate poll and RRD for Rx only) is hacky and will be impossible to clean up going forward. Once we push it, we are stuck with it.

includes/definitions/airos-af.yaml
```diff
@@ -1,6 +1,6 @@
 os: airos-af
 text: 'Ubiquiti AirFiber'
-type: network
+type: wireless
```
laf: This will require us to send a notification out before it's merged.

includes/polling/wifi.inc.php
```diff
@@ -7,6 +7,8 @@
     } elseif ($device['os'] == 'airos-af') {
         echo 'It Is AirFIBER' . PHP_EOL;
         include 'includes/polling/mib/ubnt-airfiber-mib.inc.php';
+        echo 'It Is AirFIBER - Rx' . PHP_EOL;
+        include 'includes/polling/wireless/ubnt-airfibre-rx.inc.php';
```
laf: It could be better to just modify poll_mib_def() to accept an rrd_name to be passed into $tags so you can run it more than once potentially.

If that's not possible then I still think the rx code should just go in the includes/polling/mib/ubnt-airfiber-mib.inc.php file.

laf:
@barryodonovan I've just put a comment in about updating poll_mib_def() to support being run more than once.

includes/polling/wifi.inc.php
barryodonovan: @laf I agree the rx code should go in the airfibre-mib file.

I think a notification would be the best approach and see if there are objections. I think there are few, if any, installations graphing airfibre and we are just making extra work trying to keep things simple (for an non-existent user)

barryodonovan: Of course

includes/definitions/airos-af.yaml
barryodonovan: Of course

blue-yu:
Hey guys, I don't know the procedure with development, but when can I expect this option to be available?

BR...

barryodonovan:
@murrant @laf 

Could you guys decide how to proceed with this? As I've said, I think a notification about the initial implementation extending the MIB polling is the right approach but both are on offer.

I only have one link using this equipment so I'm moving on from it until there is clarity.

laf:
@barryodonovan did you see my comment above re: poll_mib_def()? Looks trivial to make it support multiple rrd files for this PR. I can do the code if it helps?

barryodonovan:
@laf yes, it looks fairly trivial but it still leaves the situation of having two polling requests and two RRDs.
What would you think of the idea of scripting the rrdtune DS insertion?
At least a framework like that would allow similar issues to be addressed going forwards...

laf:
Tune doesn't work over rrdcached :(

I don't see an issue with two rrds, we do that for sensors, poller performance, mib polling.

barryodonovan:
@laf I may get a chance to look at the mib_poll_def tomorrow.

Are you unwilling to just use the current Rx poll?

laf:
As I would say I am, it's in a totally separate file but I honestly think we should adapt mib_poll_def()

barryodonovan:
Once #5993 has been merged I'll refactor this to use poll_mib_def()

blue-yu:
Any news on this issue? When do You expect it to be implemented?

BR...

murrant:
@barryodonovan You've got a bit of an issue with your git here.  Something got merged in a bad way.

I'm not 100% sure of the correction.  But assuming you have an upstream branch:
```
git rebase upstream/master
```

barryodonovan:
Closing this and creating new PR

Issue #5895, opened 2017-02-14
Custom thresholds for progress bars


rpenziol:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

New feature:

As described here: https://community.librenms.org/t/custom-thresholds-for-progress-bars/572
Using storage warning percentage from a devices' edit page to change the progress bars red on the storage health page and the device's overview page.

Additionally, adding the ability to set warning percentages to CPUs and memory pools from a devices 'edit' pages, then using these values to set the red progress bar threshold for their health pages and the device's overview page.

rpenziol:
I have a working solution for this features. Made the issue for tracking purposes.

laf:
Merged in

PR #5901, opened 2017-02-15
webgui: Added ability to set warning percentage for CPU, mempools fro…

rpenziol:
…m device edit page (#5895)

	*Set health progress bars for store, CPU, mempools to reach red threshold based on warning percentage

DO NOT DELETE THIS TEXT

#### Please note


build.sql
```diff
@@ -43,7 +43,7 @@ CREATE TABLE IF NOT EXISTS `processors` ( `processor_id` int(11) NOT NULL AUTO_I
...
-CREATE TABLE IF NOT EXISTS `storage` ( `storage_id` int(11) NOT NULL AUTO_INCREMENT, `device_id` int(11) NOT NULL, `storage_mib` varchar(16) NOT NULL, `storage_index` int(11) NOT NULL, `storage_type` varchar(32) DEFAULT NULL, `storage_descr` text NOT NULL, `storage_size` bigint(20) NOT NULL, `storage_units` int(11) NOT NULL, `storage_used` bigint(20) NOT NULL, `storage_free` bigint(20) NOT NULL, `storage_perc` text NOT NULL, `storage_perc_warn` int(11) DEFAULT '60', PRIMARY KEY (`storage_id`), KEY `device_id` (`device_id`) ) ENGINE=InnoDB  DEFAULT CHARSET=utf8;
+CREATE TABLE IF NOT EXISTS `storage` ( `storage_id` int(11) NOT NULL AUTO_INCREMENT, `device_id` int(11) NOT NULL, `storage_mib` varchar(16) NOT NULL, `storage_index` int(11) NOT NULL, `storage_type` varchar(32) DEFAULT NULL, `storage_descr` text NOT NULL, `storage_size` bigint(20) NOT NULL, `storage_units` int(11) NOT NULL, `storage_used` bigint(20) NOT NULL, `storage_free` bigint(20) NOT NULL, `storage_perc` text NOT NULL, `storage_perc_warn` int(11) DEFAULT '90', PRIMARY KEY (`storage_id`), KEY `device_id` (`device_id`) ) ENGINE=InnoDB  DEFAULT CHARSET=utf8;
```
laf: I'm not personally a fan of 90 as the default, think it should be more 75 or so.

laf:
Absolutely brilliant, thanks for contributing this :)

I'm not sure I'd agree the defaults for all of this should be 90%, it's too close to the max to be potentially useful. 60% or upto 75% seems more appropriate imho ( @librenms/reviewers ?)

The other alternative is we have a $config array that we use to set this and people can override:

```
$config['health']['limits']['storage'] = array(
    'high' => 90,
); 
$config['health']['limits']['memory'] = array(
    'high' => 90,
);
```

rpenziol:
Thanks @laf ! The CPU warning level defaulted to 60% before. I changed it to 90% to match the normal 'red' colored progress bar threshold. In hindsight I probably shouldn't have touched that value. Keeping it at 60% to not mess up anyone's existing settings sounds like the best course of action.

build.sql
murrant: 80 sounds good to me.
I'm ok with making this change only affect new installs.

murrant:
trigger ci

laf:
@rpenziol are you updating this to revert the default back to 60%?

rpenziol:
@laf Yes, I will revert it and push my changes. I'll try to push my changes either tomorrow or Monday

laf:
Ace, thanks :)

laf:
You've still got some defaults of 90 in there and imho a default of 90 is too high.

crcro:
it think a value around 70-75% should be more eloquent then 60 or 90 ...

rpenziol:
@laf @crcro My motivation to setting it to 90% is that this value will determine the threshold on progress bars for CPU/mempool usage to turn bright red. Currently that threshold is 90% for all progress bars. The thought being that if this is lowered it may alarm users - as many more progress bars will be bright red than before. I also realize people will be able to set alerts based off of these new values, and 90% could be considered too high for many. I have no problem changing it per your recommendations.

laf:
It's my opinion (and crcro as well) so I'd be interested in other opinions ( @librenms/reviewers ) but yes, 75% would be an acceptable value.

I realise this may show more red visually but the alerts are what are more important as those are automated, the former relies on people actually browsing to those specific pages.

rpenziol:
@laf I don't have an objection with 75%. I'll wait for more opinions or until you give the go-ahead to update my pull request with an agreed-upon percentage.

geordish:
75% sounds good to me.

html/includes/functions.inc.php
```diff
@@ -190,10 +190,16 @@ function generate_overlib_content($graph_array, $text)
 }//end generate_overlib_content()
 
 
-function get_percentage_colours($percentage)
+function get_percentage_colours($percentage, $component_perc_warn = null)
 {
+    $perc_warn = '90';
```
laf: I think just this one to adjust

PR #5907, opened 2017-02-15
newdevice: Basic support for HPE MSA storage

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

screen shot 2017-02-15 at 16 59 13
screen shot 2017-02-15 at 16 59 04


includes/polling/os/hpe-msa.inc.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+$oids = 'cpqSiProductName.0 cpqSiSysSerialNum.0';
+
+$data = snmp_get_multi($device, $oids, '-OQUs', 'CPQSINFO-MIB');
+
+$hardware = preg_replace('/"/', '', $data[0]['cpqSiProductName']);
```
murrant: What is the output for these?  I don't like the preg_replace here when `str_replace()` or `trim($sr, '"')` would likely work.

Adding the output of these to the snmprec wouldn't be a bad idea either. like this:
```
1.3.6.1.4.1.232.2.2.4.2.0|4|Product name string here
1.3.6.1.4.1.232.2.2.2.1.0|4|Serial number here (Change actual number to a similar format/length)
```

includes/definitions/hpe-msa.yaml
```diff
@@ -0,0 +1,11 @@
+os: hpe-msa
```
laf: Is storageworks not a better os name?

tests/snmpsim/hpe-msa.snmprec
```diff
@@ -0,0 +1,2 @@
+1.3.6.1.2.1.1.1.0|4|"HP MSA 2040 SAN"
+1.3.6.1.2.1.1.2.0|6|"DRSAKZR-123456789"
```
laf: This should be the sysObjectId

includes/definitions/hpe-msa.yaml
FTBZ: StorageWorks isn't used anymore by HPE, this is the MSA family now.

tests/snmpsim/hpe-msa.snmprec
FTBZ: Not sure to understand, I pushed a PR tell me if it's wrong again. The OIDs should be standardised of the real OID of the device?

PR #5919, opened 2017-02-16
Update exim-stats.inc.php

alchemyx:
Output from debug of poller.php:

```
SNMP[/usr/bin/snmpget -v2c -c COMMUNITY -Oqv -M /opt/librenms/mibs:/opt/librenms/mibs/supermicro udp:HOSTNAME:161 .1.3.6.1.4.1.8072.1.3.2.3.1.2.10.101.120.105.109.45.115.116.97.116.115]
"0
0"

 exim-statsRRD[update /opt/librenms/rrd/monitoring2/app-exim-stats-5.rrd N:0:0]

```
So we have two resulting variables from that outpu (note additional quotes there)
```
$frozen = '"0';
$queue = '0"';
```

And it fails. My graphs contain nan instead of real values. I think same issue is with apache application.

DO NOT DELETE THIS TEXT

#### Please note

### DIFF
--- a/includes/polling/applications/exim-stats.inc.php
+++ b/includes/polling/applications/exim-stats.inc.php
@@ -36,8 +36,8 @@ $rrd_def = array(
 );
 
 $fields = array(
-    'frozen' => intval($frozen),
-    'queue' => intval($queue)
+    'frozen' => intval(trim($frozen, '"')),
+    'queue' => intval(trim($queue, '"'))
 );
 
 $tags = compact('name', 'app_id', 'rrd_name', 'rrd_def');
--- a/includes/polling/applications/exim-stats.inc.php
+++ b/includes/polling/applications/exim-stats.inc.php
@@ -36,8 +36,8 @@ $rrd_def = array(
 );
 
 $fields = array(
-    'frozen' => $frozen,
-    'queue' => $queue
+    'frozen' => intval($frozen),
+    'queue' => intval($queue)
 );
 
 $tags = compact('name', 'app_id', 'rrd_name', 'rrd_def');


laf:
Feels like we really should fix this in the agent file itself.

alchemyx:
Hi Neil,

I don't think it is the agent fault. If you run it from the commandline it returns:

# /etc/snmp/exim-stats.sh
0
1

Quotes are added by snmpd or snmpget. 

-- 
Michał Margula, michal@margula.pl
"W życiu piękne są tylko chwile" [Ryszard Riedel]

Dnia 16.02.2017 o godz. 19:39 Neil Lathwood <notifications@github.com> napisał(a):

laf:
Fair enough :)

laf:
Actually doesn't this always then just output 0?

```
-bash-4.2$ cat /tmp/l.php
<?php

$init_modules = [];
include '/opt/librenms/includes/init.php';


$data = '"0';
print_r(intval($data));
$data = '"8';
print_r(intval($data));
-bash-4.2$ php /tmp/l.php
00-bash-4.2$
```

alchemyx:
Yes. It seems that you are right. I didn't check it thoroughly because I always had 0 as $frozen and I assumed it was working fine.

Lets change it to this:

```
 $fields = array(
     'frozen' => intval(trim($frozen, '"')),
    'queue' => intval(trim($queue, '"'))
  );		  

```

Test:

```
alchemyx@phoenix ~> php
<?php
$a = '"3';
$b = '4"';
print_r(trim($a, '"'));
print_r(trim($a, '"'));
?>
33
```

Would it be acceptable solution?

laf:
I think so yes :)

alchemyx:
I updated it with that version. Hope it will be accepted now. Thanks!

Issue #5904, opened 2017-02-15
hddtemp sensors not updating correctly


tsumaru720:
Hi Fellas, 

I noticed that last Thursday, graphs for my hddtemps are not displaying correctly in LibreNMS

https://s3-eu-west-1.amazonaws.com/uploads-eu.hipchat.com/23738/290850/cNy6rMcoRcNJD1h/libre.png
https://s3-eu-west-1.amazonaws.com/uploads-eu.hipchat.com/23738/290850/uS0my0ZLfJXWpFH/upload.png

My daily.sh cron runs at 00:15 every day which is about the time this seems to have occured (as per the graph)

As per request from @laf on IRC - Output of (snipped ofc):

`./poller.php -r -f -m applications,unix-agent  -h HOSTNAME -d`

http://pastebin.com/y7sCfhK7

Permissions appear to be correct (and the files are being updated, by rrdcache) however it seems zero value is being inserted into the RRDs themselves - no other issues with RRD files as far as I can tell

```
# ls -alctr | grep hddtemp-4
-rw-rw-r--  1 librenms librenms  125000 Feb 15 11:05 sensor-temperature-hddtemp-4.rrd

# rrdtool fetch sensor-temperature-hddtemp-4.rrd AVERAGE -s 1486598400 -e 1486602000
                         sensor

1486598700: 3.0000000000e+01
1486599000: 3.0000000000e+01
1486599300: 1.2000000000e+00
1486599600: 0.0000000000e+00
1486599900: 0.0000000000e+00
1486600200: 0.0000000000e+00
1486600500: 0.0000000000e+00
1486600800: 0.0000000000e+00
1486601100: 0.0000000000e+00
1486601400: 0.0000000000e+00
1486601700: 0.0000000000e+00
1486602000: 0.0000000000e+00
1486602300: 0.0000000000e+00
```

As evident by the paste, the temps seem to be exposed correctly to LibreNMS

- [x] Is your install up to date? **within last 24 hours**

```
# date
Wed Feb 15 11:01:49 GMT 2017

[WARN]  Your install is out of date, last update: Wed, 15 Feb 2017 00:06:00 +0000
```

- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.

Component | Version
--------- | -------
LibreNMS  | 49b5844346bdfe54616a0e838a7ef634b0e7bdd1
DB Schema | 164
PHP       | 5.5.9-1ubuntu4.21
[...]

PR #5923, opened 2017-02-16
refactor: Updated storing of sensors data to be used in unix-agent #5904

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5904 

I broke this with the sensors refactor. Centralised the sensor data update into a function so it's re-usable elsewhere. A bit of a hack for unix-agent but that really needs some love at some stage.


laf:
Yeah sorry I hadn't add the db update call I meant to. Just pushed that now.

The issue with the old way was the total reliance on running unix-agent and sensors within the same poller run so you have to have both for it to work. We shouldn't have any reliance on modules with one another so this now splits that out.

It's still hacky but the unix-agent needs a lot of work anyway so it will do for now.

laf:
I've pulled in your changes from your PR into this as well.

Issue #5739, opened 2017-02-02
OS Ticket Transport API Broken


jmoore2016:
In includes/alerts/transport.osticket.php the line 'email' => $_SERVER['SERVER_NAME'], does not return a valid email to OS Ticket and breaks the OS Ticket API Transport. I had to set 'email' => 'myemail@mydowmain.com', for this to work.

DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ ] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ ] Please provide as much detail as possible.
- [ ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

laf:
We can provide a text box to enter an email or we could use sysContact from the device. What would be preferred.

setiseta:
It seems that its not only the OS Ticket.
On my instance E-Mail Alert is Broken.
I'm using SMTP but can't even see a connect from the librenms server on my smtp server.

laf:
@jmoore2016 can you please test: https://github.com/librenms/librenms/pull/5927

It now default to using the from email in the email transport.

PR #5927, opened 2017-02-16
fix: Update OSTicket transport to use the from email address #5739

laf:
DO NOT DELETE THIS TEXT

#### Please note

Fixes: #5739 
If this works and is accepted then we will need to send a notification out. Let me know when it's approved and I'll do that.

I've added a process_config include file so clean up the config we expect to see (I'm sure it's useable for other things). I did this rather than having a function that needed to be called as it's variables that should just be available.

### DIFF
--- a/includes/alerts/transport.osticket.php
+++ b/includes/alerts/transport.osticket.php
@@ -11,9 +11,15 @@
 
 $url = $opts['url'];
 $token = $opts['token'];
+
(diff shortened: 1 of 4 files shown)


laf:
@librenms/reviewers ?

laf:
bumpy bump bump bump @librenms/reviewers

PR #5930, opened 2017-02-16
Fixed a few small bugs, introduced by the Poller tidy up of laf

exarv:
DO NOT DELETE THIS TEXT

#### Please note


includes/polling/functions.inc.php
```diff
@@ -56,7 +56,7 @@ function poll_sensor($device, $class)
     $misc_sensors = array();
     $all_sensors = array();
 
-    foreach (dbFetchRows("SELECT * FROM `sensors` WHERE `sensor_class` = '?' AND `device_id` = ?", array($class, $device['device_id'])) as $sensor) {
+    foreach (dbFetchRows("SELECT * FROM `sensors` WHERE `sensor_class` = ? AND `device_id` = ?", array($class, $device['device_id'])) as $sensor) {
```
laf: This was actually changed due to some users complaining that queries weren't being quoted so it was doing WHERE `sensor_class` = temperature and failing.

laf:
You also mention in the other PR that this broke unix-agent hddtemps, does this actually fix it as from testing today and even just now it doesn't seem to. I've put another PR in place that does fix hddtemp graphs.

includes/polling/functions.inc.php
exarv: Hmm. strange. I always learned that, when using ? for a string field, you never quote it...

laf: Ditto and we don't all over the show but for some reason people have been having issues recently and no one's been able to track it down. So unless this is causing you problems then we should leave it for now.

exarv: I also see now a dbMakeQuery in includes/dbFacile.php, which combines the query and the data. And as far as I can see, it's not using the PDO prepare / execute

exarv:
My patch fixed the graphs and the sensor (health) alert checking for the unix-agent hddtemp.
At least on my installation (last update from about 24 hours ago)
CentOS 7.3.1611, PHP 5.4.16, MariaDB 5.5.52

includes/polling/functions.inc.php
exarv: Well, the code with the quoted '?' does give me problems. Then the hddtemp drop to 0 degrees in my graphs, and I have alerts for alle servers with unix-agent / hddtemp enabled. (temp = 0 C)

laf: Is that not just because it's broke anyway?

laf:
The issue overall with how this is now is the reliance on running sensors module as well, without this doesn't work so it makes more sense to separate them out as per my pr.

PR #5932, opened 2017-02-17
Db collation standardisation

laf:
DO NOT DELETE THIS TEXT

#### Please note

Ok this should be complete now unless we have any suggestions from people.

utf8_unicode_ci selected as the default collation. utf8mb4 should be used but only introduced in mysql 5.5 which isn't the default in CentOS 5 or 6 so decided not to use it.

It is slow to run on heavy DBs, my 3k+ device install at a best guess took a total of 10-12 minutes to run. This is on varying tables so isn't a total impact for that time. I don't really see a way round it :(

Ideally we should tidy up the build base and schema files as we have a lot of duplication + collations being set there but overall it's not necessary here as for existing installs it's too late and this is need to clean it up, for new installs no data exists so this change will be quick and run for all new users once merged.

I've tested running build-base.php manually, web installer and they still all work fine.

Updated validate to check for this.

Added test units to make sure no issues are re-introduced.


laf:
tagging @librenms/reviewers

murrant:
Looking pretty good. After  running the schema update, I still had these:
```
 mysqldump -d -p librenms|egrep -i "TABLE|ENGINE"|grep -i latin1 -B 1
Enter password:
CREATE TABLE `_config` (
) ENGINE=InnoDB AUTO_INCREMENT=930 DEFAULT CHARSET=latin1;
--
CREATE TABLE `cmpMemPool` (
) ENGINE=MyISAM DEFAULT CHARSET=latin1;
--
CREATE TABLE `current` (
) ENGINE=MyISAM AUTO_INCREMENT=189 DEFAULT CHARSET=latin1;
--
CREATE TABLE `fanspeed` (
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
--
CREATE TABLE `frequency` (
) ENGINE=MyISAM AUTO_INCREMENT=189 DEFAULT CHARSET=latin1;
--
CREATE TABLE `voltage` (
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```

laf:
_config never exist (assuming you backed it up maybe).

Others are legacy tables now cleaned up so you can safely delete them.

murrant:
_config was from a failed schema update ;)

murrant:
Tested and looks good.  Tested after upgrading an existing db and buildbase.

laf:
Releasing on Monday the 27th Feb. Will rebase just before.

guipoletto:
Is there a housekeeping-guide (cleanup) that i can do to the DB before this conversion?

murrant:
@guipoletto You might be looking for this: http://docs.librenms.org/Support/Configuration/#cleanup-options

But really nothing should need to be done in preparation.  If you want to do it ahead of time or if the update fails, you can simply run the sql file from this PR.

wriedel:
Hi Folks,
just for my own education what's the reason going for utf8_unicode_ci / utf8 instead of utf8mb4_unicode_ci / utf8mb4 ?

Thanks,
Wolfgang

laf:
The reason is at the start of this pull request @wriedel :)

wriedel:
@laf OK, thanks ;-)
So like me, if you're already on 10.2.4-MariaDB with utf8mb4_unicode_ci / utf8mb4 is there anything which needs to be taken care of?
Thanks
Wolfgang

laf:
if you look at a lot of the tables and columns you'll find they aren't utf8mb4_unicode_ci. whilst we'd love to have a config option for people to choose the encoding type it's not possible right now.

Basically this pr will force utf8_unicode_ci

PR #5933, opened 2017-02-17
fix: Remove the default date from the maintenance form when editing

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

- Removing the default value of 1970 for this field to prevent to lapse from maintenance when editing.


FTBZ:
Credit to @schubling

laf:
I'll have to test but I doubt this will work for MySQL with sql_mode on

laf:
Unfortunately this doesn't work:

`Incorrect date value: '' for column 'end_recurring_dt' at row 1`

`1970-01-01 00:00:01` is the minimum default for MySQL with `NO_ZERO_IN_DATE,NO_ZERO_DATE` set for sql_mode which we are trying to strive to achieve support for.

FTBZ:
Thanks for the tests. And can we use a date in the future? If no, this PR can be closed but it's not very friendly when editing.

laf:
Explain the issue you are having.

FTBZ:
We're using this maintenance window during the backup session to be not annoyed with false-positive alerts. So we need to edit regularly the maintenance rule to add new service and remove old.

But when editing, the field "End maintenance date" will always be filed with the 1970-01-01 00:00:01. So if you don't take care and validate. The maintenance will be lapsed and you need to create a new one will all your service.

When writing this, I think that the best method isn't to leaving the field blank when creating the maintenance rule. I can change to a date far far far away...

laf:
So we just need to blank out the fields with the default value in then?

FTBZ:
Yes, mostly in the edit form.

Issue #5903, opened 2017-02-15
Service module has conflicted configuration files


FTBZ:
- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)

Component | Version
--------- | -------
LibreNMS  | 3fd56d729729ad036f6ab948d8f22c611ce58c9b
DB Schema | 164
PHP       | 7.0.11
MySQL     | 5.5.50-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2

### Service module has conflicted configuration files

I noticed that the service module has some conflict in the files that make some variables to be ignored. To explain my problem, we're using this example only the Nagios check_http module.

When configuring LibreNMS to check a service that's NOT linked to a device, I noticed that the field "IP Address" is ignored with the check_http module and not with the check_ldap module.

Browsing the source code to understand my problem I noticed this problem:

- librenms/includes/services.inc.php has a way to handle services that have not a specific file present in librenms/includes/services/ (line 128)
- The code presents at this line include the ability to use the IP address field
- In the other hand, the file check_http.inc.php ignores the IP variable field

**services.inc.php code:**
`$config['nagios_plugins'] . "/check_" . $service['service_type'] . " -H " . ($service['service_ip'] ? $service['service_ip'] : $service['hostname']);`

**check_http.inc.php code:**
`$config['nagios_plugins'] . "/check_http -H ".$service['hostname']." ".$service['service_param'];`

So the question is why using custom files for services that's compatible with the "standard" method present on line 128 using the -H argument? This is redundant?

To correct this problem we need to implement the $ip variable to all the files in librenms/includes/services/ or to remove the files that use exactly the same format at the line 128 I think. If technically there is no reason to use a separated file, I prefer the second method.

Perhaps @adaniels21487 can help for this?

FTBZ:
So, without more advice I will PR the second solution. @InsaneSplash (Latest committer)  can give its opinion before?

laf:
I gave you my advice last night on irc :/

FTBZ:
Sure, this is why I say that I will PR the second solution. But I just want the option of @InsaneSplash because we have talked together about her last commits to these files.

adaniels21487:
Sorry, I'm a little late to the party...
The code that's there came from the original, forked codebase. All I added was the statistics collection.
If the computed $check_cmd doesn't differ from the default one, I see no need to keep the separate files.

Thanks,
Aaron

[...]

PR #5934, opened 2017-02-17
fix: Service module has conflicted configuration files #5903

FTBZ:

--------------

Fix for #5903 

- This PR tries to solve a problem where the $ip variable is ignored with many Nagios services. The files using exactly the same command as the default in services.inc.php have been removed.
- Tested with some services, but not all a double check from other members will be appreciated.


laf:
Looks good, the additional files do just what poll_service() does so makes sense to remove.

I don't think we need to notify users with this even though in theory it could change the services but only to how they thought it should have worked.

PR #5936, opened 2017-02-17
fix: Limit get_graph_by_port_hostname() to one port and exclude deleted

arjitc:
DO NOT DELETE THIS TEXT

#### Please note


arjitc:
In my case, there were 3 ports being returned when polling 1 port, out of these 3 ports, 2 were deleted (same name but deleted) and 1 was the actual/current port graph. This patch pulls in only the non-deleted port.

Thanks to @laf for the help over IRC.

PR #5912, opened 2017-02-15
add PHP-FPM

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [X] Have you signed the [Contributors agreement](http://docs.librenms.org/General/Contributing/) - please do NOT submit a pull request unless you have (signing the agreement in the same pull request is fine). Your commit message for signing the agreement must appear as per the docs.
- [X] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

At current time, this is limited to a single pool.

PR #5937, opened 2017-02-17
Add php-fpm graphs to application overview (#5912)

florianbeer:
DO NOT DELETE THIS TEXT

#### Please note


florianbeer:
Just found out that the mouseover graphs don't show up correctly. No clue why.
@laf help, please?

VVelox:
The mouse over does not work as from what I can tell that page does not like the way it is named.

Was just mucking around with that here with fail2ban and it looks like the graph would best be renamed to php-fpm_stats then just include 'stats' in the array instead of ''.

florianbeer:
Great! Makes sense.
How/where do I rename the graph?

laf:
That does indeed seem to the best approach.

VVelox:
mv html/includes/graphs/application/php-fpm.inc.php html/includes/graphs/application/php-fpm_stats.inc.php

Then update html/pages/device/apps/php-fpm.inc.php and html/pages/apps.inc.php to reflect that the graph has been renamed from php-fpm to php-fpm_stats.

Just did that here and it fixes the issue.

florianbeer:
Done. Thanks for the assistance!

VVelox:
Null sheen!

florianbeer:
Thanks for creating this whole php-fpm thing, chummer!

Issue #5935, opened 2017-02-17
Exim-stats app queue graph showing the amount of frozen messages


valikvicious:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.

==========================================================
Component | Version
--------- | -------
LibreNMS  | 1140be1a01d3974c1c15b6661af42ee0437e30b2
DB Schema | 164
PHP       | 7.0.14
MySQL     | 5.5.50-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
==========================================================

- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

The file librenms/html/includes/graphs/application/exim-stats_queue.inc.php refers to the 'frozen' array element instead of 'queue' in line 38.

Due to this, the exim-stats app shows frozen messages in the queue graph, instead of the full queue amount.

Correcting line 38 as in the attached file leads to the correct graph being displayed:

librenms/html/includes/graphs/application/exim-stats_queue.inc.php
37 $array = array(
38     'queue' => array('descr' => 'Queue emails','colour' => 'c13a38',),
39 );

Hope this helps.

[exim-stats_queue.inc.php.tar.gz](https://github.com/librenms/librenms/files/784050/exim-stats_queue.inc.php.tar.gz)

laf:
Can you submit this as a pull request?

PR #5945, opened 2017-02-18
Fix exim queue graph

valikvicious:
DO NOT DELETE THIS TEXT

The exim-stats app shows frozen messages in the queue graph, instead of the full queue amount. Refer to issue #5935 for more details.

This is corrected here.


laf:
Thanks for the fix. You need to submit a commit inline with the contributors agreement.

Issue #5946, opened 2017-02-19
QNAP snmp readings broken since #5925


tuxi56:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Component | Version
--------- | -------
LibreNMS  | bf7656923e951d8b596e6be8cc232d9fb5a43845
DB Schema | 165
PHP       | 7.0.16-1~dotdeb+8.1
MySQL     | 5.5.54-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1

Hi

I have no readings from my two QNAP NAS (TS-112P and TS-259 Pro+) in LibreNMS (no bandwith, no cpu usage, no memory, no storage, nothing) since 2017-02-18 shortly after the daily.sh, all I could find by this period of time was the #5925 pr. 

Event log for Turbo NAS TS-259 Pro+ : 
![image](https://cloud.githubusercontent.com/assets/22138991/23103278/26762ca8-f6b9-11e6-9c84-7b76fdb3b080.png)

Event log for Turbo NAS TS-112P : 
![image](https://cloud.githubusercontent.com/assets/22138991/23103510/60609518-f6bc-11e6-93be-2b8bc7150edf.png)

I'm unable to provide you discovery.php or poller.php logs for those devices, because of `Timeout: No Response from udp:<IP_ADDRESS>:161.` It seems that the snmp process on the NAS crashes when the poller runs (when manually restarting, all I can do is snmpbulkwalk ; if I run poller.php just after it hangs with timeout message). 

[...]

PR #5948, opened 2017-02-19
feature: Added snmp_check() to decide if a device is up/down for snmp #5946

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Pretty sure this fixes: #5946 

It falls back to the old way if this doesn't work (or should). Removed the hack for hikvision as it shouldn't have been done like that.

The idea behind this is that we only care if we have a response code of 0 from the call, which any lookup that gets a form of snmp response has. I checked with snmpsimd and added a file with just sysDescr response, this is detected as being ok and returns true (we don't care we have a sysObjectId we just want to know snmp is up).

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -557,15 +557,16 @@ function isSNMPable($device)
 {
     global $config;
 
-    $pos = snmp_get($device, "sysObjectID.0", "-Oqv", "SNMPv2-MIB");
-    if (empty($pos)) {
-        // Support for Hikvision
-        $pos = snmp_get($device, "SNMPv2-SMI::enterprises.39165.1.1.0", "-Oqv", "SNMPv2-MIB");
-    }
-    if ($pos === '' || $pos === false) {
-        return false;
-    } else {
+    $pos = snmp_check($device);
+    if ($pos === true) {
         return true;
+    } else {
+        $pos = snmp_get($device, "sysObjectID.0", "-Oqv", "SNMPv2-MIB");
+        if ($pos === '' || $pos === false) {
+            return false;
+        } else {
+            return true;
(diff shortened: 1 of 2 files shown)


includes/snmp.inc.php
```diff
@@ -255,6 +255,25 @@ function snmp_get($device, $oid, $options = null, $mib = null, $mibdir = null)
     }
 }//end snmp_get()
 
+/**
+ * @param $device
+ * @return bool
+ */
+function snmp_check($device)
+{
+    $time_start = microtime(true);
```
murrant: $time_start is unused ;)

includes/functions.php
```diff
@@ -557,15 +557,16 @@ function isSNMPable($device)
...
-        // Support for Hikvision
-        $pos = snmp_get($device, "SNMPv2-SMI::enterprises.39165.1.1.0", "-Oqv", "SNMPv2-MIB");
-    }
-    if ($pos === '' || $pos === false) {
-        return false;
-    } else {
+    $pos = snmp_check($device);
+    if ($pos === true) {
         return true;
+    } else {
```
murrant: Why try two times?

laf: For now it's catch all. Whilst I'm sure this change will work I can't be 100% sure, all I've been able to test against is net-snmpd and snmpsimd. Up / down is the bit I know works but actually checking we have sysObjectId is where I've had limited testing.

We can remove it but I just didn't want to.

includes/snmp.inc.php
laf: Didn't mean to remove the line that used it, added that back in.

PR #5949, opened 2017-02-20
Adding hosts through addhost.php on SNMP v3 - Documentation

ahmadkhudeish:
This change is just to display the list of parameters required for SNMPv3 configuration for addhost.php. I think, it is easier to use this method when the parameters are available. I also include an example of how this method can be used :)

Kind Regards
Ahmad

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
The 14.04 are old docs now and aren't possibly 100% accurate anymore. I don't see the benefit of adding noise to the install docs like this, it should be in it's own docs with examples of adding devices but overall I'm not sure even that's worth it above what we already have.

PR #5947, opened 2017-02-19
feature: Allow customisation of rrd step/heartbeat when creating new rrd files

laf:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you signed the [Contributors agreement](http://docs.librenms.org/General/Contributing/) - please do NOT submit a pull request unless you have (signing the agreement in the same pull request is fine). Your commit message for signing the agreement must appear as per the docs.
- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

I give you 1 minute polling:

image

Done mostly with sed so hopefully it's not broken anything. Needs script to convert new files + docs.

PR #5950, opened 2017-02-20
RFC: RRD Definition object

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Move RrdDefinition to an Object to make it easier to create and remve the possibility of typos.
@laf: Combine this with #5947 ?

This code is not mergable as-is, it requires updating all definitions.

### DIFF
--- a/LibreNMS/Exceptions/InvalidRrdTypeException.php
+++ b/LibreNMS/Exceptions/InvalidRrdTypeException.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ * InvalidRrdTypeException.php
+ *
+ * -Description-
+ *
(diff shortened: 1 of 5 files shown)


laf:
@murrant are you looking to merge those two prs and update all the files to make use of this new class at the same time?

murrant:
@laf yes, if it seems like this is something we want.

laf:
Yes indeed, just don't know if I've got the time to go and update all the files to use this straight away :/

murrant:
I can take a shot at it.  A few regex replacements should do it.

laf:
If you do just add it to this PR.

I'll do the webui config and docs if you like.

PR #5952, opened 2017-02-20
Update LLDP discovery to use IP address if discovery_by_ip is enabled

geordish:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fix for #5872

### DIFF
--- a/includes/discovery/discovery-protocols.inc.php
+++ b/includes/discovery/discovery-protocols.inc.php
@@ -191,8 +191,8 @@ if ($device['os'] == 'pbn' && $config['autodiscovery']['xdp'] === true) {
                                 $ptopo_array = snmpwalk_cache_oid($device, 'ptopoConnEntry', array(), 'PTOPO-MIB');
                                 d_echo($ptopo_array);
                                 foreach (array_keys($ptopo_array) as $ptopo_key) {
-                                    d_echo("Testing '" . $ptopo_array[$ptopo_key]['ptopoConnRemoteChassis']. "' == '${lldp['lldpRemChassisId']}'\n");
-                                    if (strcmp($ptopo_array[$ptopo_key]['ptopoConnRemoteChassis'], $lldp['lldpRemChassisId']) == 0) {
+                                    d_echo("Testing '" . trim($ptopo_array[$ptopo_key]['ptopoConnRemoteChassis']). "' == '" . trim($lldp['lldpRemChassisId']) . "'\n");
+                                    if (strcmp(trim($ptopo_array[$ptopo_key]['ptopoConnRemoteChassis']), trim($lldp['lldpRemChassisId'])) == 0) {
                                         $ip_arr = explode(" ", $ptopo_array[$ptopo_key]['ptopoConnAgentNetAddr']);
 
                                         $a = hexdec($ip_arr[0]);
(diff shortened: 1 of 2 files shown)


includes/discovery/discovery-protocols.inc.php
```diff
@@ -157,6 +157,10 @@
     d_echo($lldp_array);
     $dot1d_array = snmpwalk_cache_oid($device, 'dot1dBasePortIfIndex', array(), 'BRIDGE-MIB');
     d_echo($dot1d_array);
+    if ($config['discovery_by_ip'] == true) {
+        $ptopo_array = snmpwalk_cache_oid($device, 'ptopoConnEntry', array(), 'PTOPO-MIB');
```
laf: May as well move these two lines into the next `if ($config['discovery_by_ip'] == true) {`

includes/discovery/discovery-protocols.inc.php
```diff
@@ -186,7 +190,24 @@
...
+                                        $a = hexdec($ip_arr[0]);
+                                        $b = hexdec($ip_arr[1]);
+                                        $c = hexdec($ip_arr[2]);
+                                        $d = hexdec($ip_arr[3]);
+                                        
+                                        $discover_hostname = "$a.$b.$c.$d";
+                                    }
+                                 }
+                            } else {
+                                $discover_hostname = $lldp['lldpRemSysName'];
```
laf: Is lldpRemSysName always a hostname? If it's not then we should try that along with the new code.

geordish: You mean if discover by IP is enabled, do discover on lldpRemSysName, then discover on IP, if its not, just do discover on lldpRemSysName?

laf: Well we possibly should actually just run them all regardless, discover_new_device() does checks on discovery_by_ip.

towster:
tested it - detects those devices properly now

laf:
I actually more meant like this. Patch against your branch:

https://gist.githubusercontent.com/laf/3fe8ab5084d896d4cb260124b234bede/raw/00419fbf674a330c4af30a555a7c02c790db5a21/gistfile1.txt

geordish:
Bizarly your patch wouldn't apply!

This should be roughly what you are suggesting. @towster Any chance of testing this updated patch?

towster:
[discovery-10.110.200.60.txt](https://github.com/librenms/librenms/files/791096/discovery-10.110.200.60.txt)

Here is rediscover with patch applied.
It did not add the switch properly..

laf:
I can't see why initially but the logs indicate a bad hostname/ip is being passed. It definitely gets the right part of the code though. @geordish feel free to revert to your known working one.

geordish:
I've added more debugging. @towster - mind providing the output now?

towster:
I re-ran it. Think key is there appears to be a trailing space..
`Testing 'A8 97 DC 98 D2 00' == 'A8 97 DC 98 D2 00 '
`

[discovery-10.110.200.60.txt](https://github.com/librenms/librenms/files/794149/discovery-10.110.200.60.txt)

geordish:
Good spot @towster. That should be fixed now!

towster:
tested it - switch was added properly
Thanks @geordish

towster:
capture

geordish:
was that working before the patch?

towster:
dont believe so - either way the discovery is adding the device properly - to see about resolving the fact that one port links but the other doesnt I would think another issue might be in order - lets see what @laf says

laf:
A debug at the time of it adding will be needed to see what's next but some port info is built up from mac addresses then a discover_link is run.

@geordish If you can remove the d_echo("Testing line then I'm ok to merge this unless you want to look at the link issue as well.

geordish:
Since the issue isn't related to this patch, I will leave it. Don't fancy heading down that rabbit hole right now!

Removed excess debug statements.

PR #5957, opened 2017-02-20
fix: RF1628 UPS have wrong state information

FTBZ:
screen shot 2017-02-20 at 20 30 20

- Removing the none working state elements
- Adding two standard state information from UPS-MIB

The code used before for "Seconds on Battery" and "Battery: Estimated minutes remaining" will be moved to a new section using an other PR in the a few hours. (more easy if I broke something :p)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/sensors/states/rfc1628.inc.php
+++ b/includes/discovery/sensors/states/rfc1628.inc.php
@@ -1,43 +1,78 @@
 <?php
 
-echo 'RFC1628 ';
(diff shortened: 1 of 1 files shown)


laf:
Is there only ever 1 battery as the codes changed from walking a tree to just a single get.

FTBZ:
Oh, didn't think about that. Our UPS have many batteries but the status is always showing one. So I don't know how to handle this. Will close the PR an open an issue.

PR #5971, opened 2017-02-21
api: Add ability to return all device info for get_devices_by_group()

gilrim:
Adds the option for GetDevicesFromGroup to output full device details for members of a group. also fixes utf8-output from json_encode

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
@gilrim I'm thinking of dropping the:

`
    array_walk_recursive($data, function (&$val) {
        $val = utf8_encode($val);
    });
`

Code now we have utf8 for the schema, can you jump on irc at some stage to discuss pls.

PR #5985, opened 2017-02-23
Fix hardware detection for 2.1.7 hardware

thomseddon:
Looks like they changed the output in the 2.1.7 release

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/sub10.inc.php
```diff
@@ -2,6 +2,8 @@
 
 if (preg_match("/Sub10 Systems - ([\s\d\w]+)/", $poll_device['sysDescr'], $hardware)) {
     $hardware = $hardware[1];
+} else {
+    $hardware $poll_device['sysDescr'];
```
geordish: Should this line not have an = int it?

thomseddon:
@geordish thanks for spotting that, bad copy and past on my part

PR #5991, opened 2017-02-23
fix: Adding state sensor support for Synology DSM

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

-----------

- Tested on a RS815rp+ and on a RS2212+
- Removed the support for UPS Status (probably working only when a UPS is plugged in through USB, can't test it without feedback)

screen shot 2017-02-23 at 21 13 16
screen shot 2017-02-23 at 21 13 01


FTBZ:
I kept the multi_oid for disks and for raid because it's OID table and I think it possible to have more than one raid on big systems.

laf:
image

PR #5996, opened 2017-02-24
webui: fix device edit health update icons

murrant:
Allow 0 and "" for valid health update values.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
How come allow ""?

murrant:
Many of my fields are empty, so I assume "" is perfectly fine.  I can definitely change the code to not allow it.  My main goal was to allow 0.

laf:
We should probably fix why they are empty as well.

PR #5999, opened 2017-02-24
webui: added opensuse svg logo

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
This looks like it should be in /logos and it needs to be minimized.
http://docs.librenms.org/Developing/Support-New-OS/#icon-and-logo

crcro:
correct, missed that docs part, my bad ...
added logo and icon, minified them also, thanks!

laf:
You need to undelete the os png :)

crcro:
why is that needed if we have the svg?

laf:
We have to areas for logos as per the link murrant added.

html/images/os/ is for the logo we use in the device list, html/images/logos/ are the ones we use for the device page itself.

crcro:
i've added both of them as svg in the second push

PR #6002, opened 2017-02-25
Postfix

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

Adds detailed polling for Postfix.

This requires https://github.com/librenms/librenms-agent/pull/92 .
#### Testers


html/includes/graphs/application/postfix_deferral.inc.php
```diff
@@ -0,0 +1,37 @@
+<?php
+require 'includes/graphs/common.inc.php';
```
laf: Could you remove `require 'includes/graphs/common.inc.php';` from each of the graph files and see if it's ok still. This is already included so doesn't need to be here.

VVelox: Removed and it works well with out it.

PR #6005, opened 2017-02-25
Add support for dynamic interfaces in ifAlias script

ollb:
Sometimes there are dynamic interfaces and retrieving aliases from config files is not possible. The idea is to use alias from netlink if there is no alias information in config file. It is useful for example when you have ppp links or dynamic tunnel links. In such cases, you can use the `ip link set IFACE alias ALIAS` command when initializing your interface.

* netlink and ip support alias (`ip link set IFACE alias ALIAS`). In an
  environment with dynamic interfaces, it is useful to be able retrieve
  aliases from ip.

* we cannot assume next iface oid is n + 1 since they do not follow
  incrementally when some interfaces has been destroyed and recreated.

* use straight POSIX shell

Tested on Debian GNU/Linux Jessie with snmpwalk & snmpbulkwalk.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -183,6 +183,7 @@ LibreNMS contributors:
 - Stefan Behte <craig at haquarter.de> (craig)
 - Zane C. Bowers-Hadley <vvelox@vvelox.net> (vvelox)
 - Valentin Polonuyer <valik.vicious@gmail.com> (ValikVicious)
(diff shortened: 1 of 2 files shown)


patrick7:
Works for me with both ip link alias and /etc/network/interfaces, the same goes for snmpwalk / snmpbulkwalk.

PR #6026, opened 2017-02-27
newdevice: Added Coriant support

xbeaudouin:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This PR is the first part or Coriant devices support using TNMS.

It will, using MEF-EVC MIB get a status of the links.

Since MEF support can be used on some other devices it can be added to some other devices.

Coriant have accepted that TNMS-NBI-MIB can be added into Librenms (I can give the official mail for that). Other MIBs are opensource or equivalent.


html/images/os/coriant.svg
```diff
@@ -0,0 +1,63 @@
...
+     id="circle97"
+     r="13.167"
+     cy="101.834"
+     cx="491.83301" /><path
+     style="fill:#008c99"
+     id="path99"
+     d="m 487.334,64.95 -8.667,-18.283 c 0,0 59.999,-14.333 70.333,55.167 0,59.5 -80,79.833 -109.666,21.166 14,3.667 25.333,1.667 25.333,1.667 0,0 8.334,12.999 27.167,12.333 18.833,-0.666 35.833,-11.999 35.833,-33.666 0,-21.667 -15.999,-37.768 -40.333,-38.384 z" /><path
+     style="fill:#00467f"
+     id="path101"
+     d="M 335.667,58.334 C 334.334,45.667 324,35.001 304,35.001 c -20,0 -31.333,13 -31.333,13 l 10.667,10.333 c 0,0 7.666,-7.333 21.333,-7.333 13.667,0 14,13.333 14,13.333 0,0 -1.667,0 -16.204,0 -14.537,0 -33.796,8 -33.796,23.667 0,15.667 12.333,24 26,25 13.667,1 23.333,-9.667 23.333,-9.667 v 7.334 h 17.667 c 0,0 1.333,-39.667 0,-52.334 z m -19,28.333 c 0,0 -2.667,10.667 -14.204,10.667 C 294,97.334 289,94 289,88.667 c 0,-5.333 9.334,-10.666 17.667,-10.333 8.333,0.333 10,0 10,0 z" /></svg>
```
geordish: Can you run this through something like svgo to compress it down a bit (https://github.com/svg/svgo)

xbeaudouin: Done :)

tests/snmpsim/coriant.snmprec
```diff
@@ -0,0 +1 @@
+1.3.6.1.2.1.1.2.0|6|1.3.6.1.4.1.42229.6.22
```
geordish: Can you also add the sysDescr here too please.

xbeaudouin: sysDescr is not relevant. Since Coriant TNMS is a proxy.
On my TNMS we have : 
 snmpwalk -v2c -c OsB24 par-eqx-tnms-01:50161 1.3.6.1.2.1.1.1.0
iso.3.6.1.2.1.1.1.0 = STRING: "Linux version 3.10.0-327.el7.x86_64"

But you can have also a Windows. So sysDescr is not usuable.

laf: Whilst it can vary this is purely for testing so you should add what you have available, the idea is that we can then ensure other changes don't break detection

includes/definitions/coriant.yaml
```diff
@@ -0,0 +1,47 @@
...
+    ntp: 0
+    ospf: 0
+    stp: 0
+    ucd-diskio: 0
+    ucd-mib: 0
+    ipmi: 0
+    bgp-peers: 0
+    toner: 0
+    services: 0
+    loadbalancer: 0
```
laf: Can be removed, it's already 0 by default

includes/definitions/coriant.yaml
```diff
@@ -0,0 +1,47 @@
...
+icon: coriant
+discovery_modules:
+    ntp: 0
+    ospf: 0
+    stp: 0
+    ucd-diskio: 0
+    ucd-mib: 0
+    ipmi: 0
+    bgp-peers: 0
+    toner: 0
```
laf: Can be removed, it's already 0 by default

includes/definitions/coriant.yaml
```diff
@@ -0,0 +1,47 @@
...
+    services: 0
+    loadbalancer: 0
+    libvirt-vminfo: 0
```
laf: Can be removed, it's already 0 by default

includes/definitions/coriant.yaml
```diff
@@ -0,0 +1,47 @@
+os: coriant
+text: 'Coriant TNMS'
+type: network
+icon: coriant
+discovery_modules:
```
laf: From discovery_module you can remove the following as they are all 0 by default:

loadbalancers
toner
libvirt-vminfo
vmware-vminfo
junose-atm-vp

includes/definitions/coriant.yaml
```diff
@@ -0,0 +1,47 @@
...
+    loadbalancer: 0
+    libvirt-vminfo: 0
+    vmware-vminfo: 0
+    vlans: 0
+    junose-atm-vp: 0
+    arp-table: 0
+    mef-evc: 1
+    cisco-vrf-lite: 0
+    storage: 0
+poller_modules:
```
laf: From poller_modules you can remove the following as they are all 0 by default:

junose-atm-vp
toner
loadbalancers

includes/discovery/mef-evc.inc.php
```diff
@@ -0,0 +1,72 @@
...
+        if ($mefMtu == 0) {
+            $mefMtu = 1600;
+        }
+
+        /*
+         * Check if the MEF EVC is already known for this host
+         */
+        if (dbFetchCell("SELECT COUNT(id) FROM `mefinfo` WHERE `device_id` = ? AND `mefID` = ?", array($device['device_id'], $index)) == 0) {
+            $mefid = dbInsert(array('device_id' => $device['device_id'], 'mefID' => $index, 'mefType' => mres($mefType), 'mefIdent' => mres($mefIdent), 'mefMTU' => mres($mefMtu), 'mefAdmState' => mres($mefAdmState), 'mefRowState' => mres($mefRowState)), 'mefinfo');
+            log_event("MEF link: ". mres($mefIdent) . " (" . $index . ") Discovered", $device, 'system', $mefid);
```
laf: `$mefid` is actually what the severity should be:

```
        case 1:
            return "severity-ok"; //OK
            break;
        case 2:
            return "severity-info"; //Informational
            break;
        case 3:
            return "severity-notice"; //Notice
            break;
        case 4:
            return "severity-warning"; //Warning
            break;
        case 5:
            return "severity-critical"; //Critical
            break;
```

includes/discovery/mef-evc.inc.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+
+    $sql = "SELECT id, mefID, mefIdent FROM mefinfo WHERE device_id = '".$device['device_id']."'";
+    foreach (dbFetchRows($sql) as $db_mef) {
+        /*
+         * Delete the MEF Link that are removed from the host.
+	 */
+        if (!in_array($db_mef['mefID'], $mefevc_list)) {
+            dbDelete('mefinfo', '`id` = ?', array($db_mef['id']));
+            log_event("MEF link: ".mres($db_mef['mefIdent']).' Removed', $device, 'system', $db_mef['mefID']);
```
laf: Same here for severity

includes/defaults.inc.php
```diff
@@ -712,6 +712,7 @@
 $config['poller_modules']['ntp']                         = 1;
 $config['poller_modules']['services']                    = 1;
 $config['poller_modules']['loadbalancers']               = 0;
+$config['poller_modules']['mef-evc']                     = 0;
```
laf: mef-evc seems very generic, should we do this as just mef or is that to generic?

xbeaudouin: Well, I dunno. Seems the mef-evc is kinda related to the mib itself. It should be generic and can be used on some other network devices (like adva for example).
You choose what you will like for that... :)

tests/snmpsim/coriant.snmprec
xbeaudouin: Ok I have added it.

includes/discovery/mef-evc.inc.php
xbeaudouin: I don't understand what you mean there?

xbeaudouin: Fixed.

includes/discovery/mef-evc.inc.php
xbeaudouin: Fixed.

includes/discovery/mef-evc.inc.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+/*
+ * Try to discover any MEF Links
+ */
+
+if ($device['os'] == 'coriant') {
```
geordish: Should this not be checking for the mef-evc discovery item being set?

includes/polling/mef-evc.inc.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+if ($device['os'] == 'coriant') {
```
geordish: Should this not be checking for the mef-evc poller item being set?

includes/discovery/mef-evc.inc.php
xbeaudouin: Well, as you want. Tell me if you would like I change that.

laf: If it's a standard that devices should support then the if check can be completely removed as the module will only be on for relevant devices anyway.

includes/defaults.inc.php
laf: It's a question rather than an ask. I just don't want to see us end up with 10 modules all called mef-this, mef-that, etc.

murrant: I'd say start with just met we can rename it later if needed.

includes/discovery/mef-evc.inc.php
murrant: I believe the poller code takes care of that.

xbeaudouin: Ok so let's do it :) Commit is comming :)

murrant:
I think the current svg should be moved to the logos dir and the os image should be the semi-circle with the dot.  (Which matches the favicon on their website)

murrant:
Thanks @xbeaudouin this is pretty close.

It seems like the discovery and poller modules do much the same thing.  What is the difference?  If they do the same thing why do we need both?

xbeaudouin:
@murrant mostly because the links can change in between too polls. I took what was done on VM stuff on vmware esx. Maybe this was not a good example?

laf:
Can you switch the modules to just mef @xbeaudouin ?

PR #6029, opened 2017-02-27
feature: HPE Comware Processor Discovery

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/processors/comware.inc.php
+++ b/includes/discovery/processors/comware.inc.php
@@ -3,6 +3,7 @@
  * LibreNMS
  *
  * Copyright (c) 2017 Søren Friis Rosiak <sorenrosiak@gmail.com>
+ * Copyright (c) 2017 Tony Murray <murraytony@gmail.com>
  * This program is free software: you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by the
  * Free Software Foundation, either version 3 of the License, or (at your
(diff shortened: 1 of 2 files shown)


murrant:
@Rosiak I was wondering if we could use the entPhysicalData in the database, soooo, I re-wrote this.  Do you like this better?
```

if ($device['os'] == 'comware') {
    echo 'HPE Comware ';

    if (dbFetchCell('SELECT COUNT(`entPhysical_id`) FROM `entPhysical` WHERE `device_id` = ?', array($device['device_id']))) {
        $indexes = dbFetchColumn('SELECT `entPhysicalIndex` FROM `entPhysical` WHERE device_id = ? AND entPhysicalName = ?', array($device['device_id'], 'Board'));
    } else {
        $entity_data = snmpwalk_cache_oid($device, 'entPhysicalName', array(), 'ENTITY-MIB');
        $indexes = array_keys(array_filter($entity_data, function ($entry) {
            return $entry['entPhysicalName'] == 'Board';
        }));
    }

    if (!empty($indexes)) {
        $procdata = snmpwalk_cache_oid($device, 'hh3cEntityExtCpuUsage', array(), 'HH3C-ENTITY-EXT-MIB');

        foreach ($indexes as $i => $entIndex) {
            if (isset($procdata[$entIndex])) {
                $cur_oid = ".1.3.6.1.4.1.25506.2.6.1.1.1.1.6.$entIndex";
                $cur_value = $procdata[$entIndex]['hh3cEntityExtCpuUsage'];
                $descr = 'Slot ' . ++$i;
                discover_processor($valid['processor'], $device, $cur_oid, $entIndex, 'comware', $descr, '1', $cur_value);
            }
        }
    }
}
```

includes/discovery/processors/comware.inc.php
```diff
@@ -0,0 +1,34 @@
...
+
+    if (!empty($entphydata)) {
+        $procdata = snmpwalk_cache_multi_oid($device, 'hh3cEntityExtCpuUsage', array(), 'HH3C-ENTITY-EXT-MIB');
+        foreach ($entphydata as $index => $phyvalue) {
+            if (stristr($phyvalue['entPhysicalName'], 'Board')) {
+                foreach ($procdata as $procindex => $value) {
+                    if ($index == $procindex) {
+                        $cur_oid = '.1.3.6.1.4.1.25506.2.6.1.1.1.1.6.';
+
+                        discover_processor($valid['processor'], $device, $cur_oid . $procindex, $procindex, 'comware', 'Slot ' . $x, '1', $value['h3cEntityExtCpuUsage']);
```
murrant: Should be hh3cEntityExtCpuUsage

Rosiak:
Perfectly fine with that @murrant !

laf:
lol, I'm against that personally as per the other PR which did the same.

Doing it this way requires entity-physical discovery module to be enabled and have run before this (admittedly after the second run it won't matter if entity-physical runs first or not). However the dependancy on another module imho is a no no as you're restricting users who have sensors enabled because they want to see them but not entity-physical.

Other discovery modules which require entityphysical info do another walk afaik - which if we can get the updated snmp Class finished and merged won't be an issue as the results will be cached.

murrant:
@laf I specifically made a fallback so this doesn't rely on the entity-physical module, only takes advantage of it if it exists.

laf:
Yeah I didn't realise that until you pointed it out as I'd only read the reply on email where it said you rewritten it to pull from DB.

As you pointed out in irc, this still does mean that someone could have stale data if entity-physical was disabled 6 months ago as an example. It's not a huge concern but does still swing me more to why bother checking the DB if we can't be sure when that data was from.

Rosiak:
As it's only disco job, I'm not concerned regarding the "extra" snmp req as this not that "expensive" at this point.

PR #6032, opened 2017-02-27
feature: Store the username in eventlog for any entries created through the Webui

towster:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Great idea - I'd personally say we should add another column and record this properly. @librenms/reviewers thoughts?

geordish:
Great idea, but I agree with @laf. A column on the event to relate it back to a person would be much better.

laf:
I've updated this now to store username. Chose not to do user_id due to varying auth methods and how inconsistence this could be. @librenms/reviewers can someone review pls.

towster:
Thanks for doing this more properly laf

laf:
No worries, thanks for prompting us into it :)

laf:
Updated schema file.

html/includes/table/eventlog.inc.php
```diff
@@ -60,12 +60,16 @@
         $type = $eventlog['type'];
     }
     $severity_colour = $eventlog['severity'];
+    if ($eventlog['username'] === 0) {
```
murrant: Are you sure the identical comparison (===) is the correct one here?
Aren't all db results strings?

includes/functions.php
```diff
@@ -830,13 +830,19 @@ function log_event($text, $device = null, $type = null, $severity = 2, $referenc
         $device = device_by_id_cache($device);
     }
 
+    $username = '';
+    if (function_exists(get_userid) && !empty($_SESSION['username'])) {
```
murrant: function_exists() a left over?

laf: Yeah I had userid as the one being stored originally.

Updated

PR #6037, opened 2017-02-28
refactor: Further speed improvements to port poller

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Improved the selected ports polling section, we don't need to do the full walk as that still causes issues with devices with a lot of ports. It may add a small increase for low port number devices with all enabled and up ports but this feature didn't help with that before.

For me on a device with high latency, ports went from 170~ to 120~.

### DIFF
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -181,17 +181,12 @@ if ($device['os'] === 'f5' && (version_compare($device['version'], '11.2.0', '>=
 } else {
     if ($config['polling']['selected_ports'] === true || $device['attribs']['selected_ports'] == 'true') {
         echo('Select ports polling');
-        foreach ($table_base_oids as $oid) {
-            $data = snmpwalk_cache_oid($device, $oid, $data, 'IF-MIB');
-        }
-        unset(
-            $oid,
-            $table_base_oids
-        );
         $lports = dbFetchRows("SELECT * FROM `ports` where `device_id` = ? AND `deleted` = 0 AND `disabled` = 0", array($device['device_id']));
         foreach ($lports as $lport) {
             if (is_port_valid($lport, $device)) {
                 $i = $lport['ifIndex'];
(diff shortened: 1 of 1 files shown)


wriedel:
Hi Laf, very cool. For a Cat4503/SUP8e with 96ports where I already had disabled all non needed modules I am now down to 25sec from 50sec!  
Question: For Pollers, Poller history, there is a line-item listing for every device. Is there a way to have a graph visualizing all this and also give me an aggregated number for a polling cycle? 
Would make the saving more visual ;-)

wriedel:
Hi Laf, is this by design that ports being identified by that feature gets deleted?

laf:
No

wriedel:
you need more debugs or anything else?

laf:
Full port debug for poller and discovery

laf:
I still think this is ok to merge.

https://community.librenms.org/t/do-not-pool-virtual-interfaces/619/14

LoZio:
Running LibreNMS  | 1.37-40-g6b68d5f
I checked one of the files in your commit and see different code in my setup as of today (Mar 13). There is some special procedure to do to have your patch without switching to a specific PR?
I'm desperately trying to lower my poll times and this coud be very useful.
Do I need to configure something to enable this patch?
Sorry for my dumbness

murrant:
You already have this code.

Please read this https://docs.librenms.org/Support/Performance/

LoZio:
I read that doc and I'm experimenting with all the parameters to find the better setup for my environment.
I supposed to have different code since if I look at the commit in this thread and files on my disk they are different.
Daily.sh says I have the latest code though.

murrant:
@LoZio code changes over time.  There has been improvements/bugfixes since then.
If `./validate.php` says you are all good, you are up to date.

Make sure to pay attention to the per-port polling section of the performance docs that may help you.

PR #6049, opened 2017-03-01
fix: Updated set_null() function to be more useful #6030

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6030 (hopefully).

@murrant ?


laf:
Yes and no.

It was supposed to drop `!isset($value) || ` so 0 would be valid but I mustn't have pushed my second commit. I'll do it later tonight.

laf:
@murrant Pushed. I also added a unit test

tests/CommonFunctionsTest.php
```diff
@@ -83,4 +83,12 @@ public function testRrdDescriptions()
         $data = 'Toner, S/N:CR_UM-16021314488.';
         $this->assertEquals('Toner, S/N CR_UM-16021314488.', safedescr($data));
     }
+
+    public function testSetNull()
+    {
```
murrant: Needs some more tests for default and min:
```
 $this->assertEquals(99, set_null(' ', 99));
$this->assertNull(set_null(-25, null, 0));
$this->assertEquals(2, set_null(2, 0, 2));
```

laf:
Updated

PR #6058, opened 2017-03-02
add get_disks function

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Wrote this as I want to go write some smart stuff. As we don't want to glob the RRDs, this will allow for a lists of disks to manually be fetched.

Also this way if the SQL ever changes or the like, it is easier to update as it can be changed here instead of having to look ever where that may use it.


VVelox:
Wrote this as a run up to writing some SMART stuff.

PR #6059, opened 2017-03-02
fix: Correct the use of GetContacts() #5012

### LINKED ISSUE #5012: Email alerts not working with AD authentication
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php`.
==========================================================
Component | Version
--------- | -------
LibreNMS  | c46de37197b7b7db70f50c5ef7a3f00c423789f4
DB Schema | 150
PHP       | 5.5.9-1ubuntu4.20
MySQL     | 5.5.49-0ubuntu0.14.04.1
RRDTool   | 1.4.7
SNMP      | NET-SNMP 5.7.2
==========================================================

- [ ] If you would like us to add a new device then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)

Email alerts stops working when AD authentication is enabled. Eventlog gives the following error: Could not issue critical alert for rule 'Service up/down' to transport 'mail' Error: You must provide at least one recipient email address. This error is encountered on both my Librenms test boxes running Ubuntu 14.04 and 16.04.

Doing a test mail transport, it is able to send the test mail to all users in the AD group. Even with only one user being a member of the AD group, no alerts are sent.

The setup is the following: A "Networking" AD group contains three user accounts each with assigned email addresses.
In config.php, the following is added: $config['auth_ad_groups']['Networking']['level'] = 10;

Alerting settings: https://drive.google.com/open?id=0B8WCN9DjlJUca1pXajQxYXNGMlk
Alert capture: https://drive.google.com/open?id=0B8WCN9DjlJUcYkw4X3JMaUZFek0
Alert rules: https://drive.google.com/open?id=0B8WCN9DjlJUcNENsSGd6VWdDYTg

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Possibility that this fixes #5012

### DIFF
--- a/html/includes/output/query.inc.php
+++ b/html/includes/output/query.inc.php
@@ -38,14 +38,15 @@ switch ($type) {
         $device = device_by_id_cache($device_id);
         $rules = GetRules($device_id);
         $output = '';
-        $qry = array();
+        $results = array();
         foreach ($rules as $rule) {
             if (empty($rule['query'])) {
                 $rule['query'] = GenSQL($rule['rule']);
             }
             $sql = $rule['query'];
-            $qry[] = dbFetchRow($sql, array($device_id));
-            if (count($qry) > 0) {
+            $qry = dbFetchRow($sql, array($device_id));
+            if (is_array($qry)) {
+                $results[] = $qry;
                 $response = 'matches';
             } else {
                 $response = 'no match';
@@ -56,7 +57,7 @@ switch ($type) {
             $output .= 'Rule match: ' . $response . PHP_EOL . PHP_EOL;
         }
         if ($config['alert']['transports']['mail'] === true) {
-            $contacts = GetContacts($qry);
+            $contacts = GetContacts($results);
             if (count($contacts) > 0) {
                 $output .= 'Found ' . count($contacts) . ' contacts to send alerts to.' . PHP_EOL;
             }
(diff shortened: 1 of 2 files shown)


html/includes/output/query.inc.php
```diff
@@ -43,7 +43,7 @@
                 $rule['query'] = GenSQL($rule['rule']);
             }
             $sql = $rule['query'];
-            $qry = dbFetchRow($sql, array($device_id));
+            $qry[] = dbFetchRow($sql, array($device_id));
```
murrant: Won't this work better `$qry = dbFetchRows($sql, array($device_id));`?
`Row`->`Rows`

html/includes/output/query.inc.php
```diff
@@ -43,7 +43,7 @@
...
             if (is_array($qry)) {
```
murrant: This is now always true.

html/includes/output/query.inc.php
laf: No, $qry is passed later on and needs to be an array, the $sql is a single row lookup but we're in a loop so need to build that $qry array up.

html/includes/output/query.inc.php
laf: Updated

Issue #5068, opened 2016-11-27
Alerting Based on Application Status / Variables


arrmo:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

- [ X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ X] Please include all of the information between the `====================================` section of `./validate.php`.
- [ X] If you would like us to add a new device then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)
- [X ] Please provide as much detail as possible.

One item that may be nice (and may exist, but I have been trying to use it with no luck unfortunately) - alerting based on values / variables inside applications. For example, create an alert when os_updates is greater than some threshold.

Thanks!

laf:
Take a look at the applications mysql table, see if it's got what you want in there and then base your alert on that data.

arrmo:
OK, will try to fiddle with it. I was looking the right place it seems - that's good .. :). Trying this for os_updates, I do see 4 rows in the table, with device_id matching the right machines. But still trying to find the actual data (all I see is app_state = UNKNOWN). Will keep digging, thanks!

laf:
If that's all that's in there then you might be out of look at the moment.

Applications support isn't 100% so it's possible the info isn't logged to the DB.

arrmo:
NP, thanks for the pointers. Do we close this, or leave it for future enhancement? Whatever you prefer is fine with me.

laf:
I'd say leave it, it should be something that works.

arrmo:
Sorry, missed this going by somehow ... :-(. I see app_state, but that's not showing the number of OS Updates that are pending - is it?

Thanks!

PR #6061, opened 2017-03-02
feature: Added app_state support for applications #5068

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #5068

Application support needs some major loving and a rewrite, for now this will provide a status + prev_status to be useful for alerting.


laf:
Rebased

includes/polling/functions.inc.php
```diff
@@ -564,3 +564,37 @@ function get_device_oid_limit($device)
         return 10;
     }
 }
+
+/**
+ * Update the application status and output in the database.
+ *
+ * @param array $device
```
murrant: Why do we need $device?

laf: Good spot, I've made more use of it by adding it to the dbUpdate call where it should have been included anyway.

includes/polling/functions.inc.php
```diff
@@ -596,5 +596,5 @@ function update_applications($device, $app, $response)
     if ($data['app_state'] != $app['app_state']) {
         $data['app_state_prev'] = $app['app_state'];
     }
-    dbUpdate($data, 'applications', '`app_id` = ?', array($app['app_id']));
+    dbUpdate($data, 'applications', '`app_id` = ? AND `device_id` = ?', array($app['app_id'], $device['device_id']));
```
murrant: app_id is unique, why do we need to check device_id too?

laf: Being cautious?

Issue #4930, opened 2016-11-03
Geolocation on pollers causes duplicates because of caching and race conditions


drakej:
==========================================================
LibreNMS Version: 6bed6cbea187b1a59b3cd049a3aa29ca6081723d
DB Schema: 148
PHP: 5.6.26
MySQL: 5.1.73
RRDTool: 1.4.8
SNMP: NET-SNMP 5.7.2
==========================================================

After turning on $config['geoloc']['latlng'] = true; in config.php, it creates a ton of duplicates inside of the locations DB table.

I found two problems with the setup. 

First, the query to check for an existing location in the function location_to_latlng (includes/polling/functions.inc.php) runs a dbFetchRow query against the cache (why would you ever want to do that if it's a check for existing data in the DB?!).

The second problem occurs even if you do turn off the memcached usage for the call via the 3rd argument being set to true. There's a race condition because devices are polled and updated asynchronously and if two devices in the same location both check the locations table for a preexisting row before either of them write the result of the google lookup, it creates duplicates.

You need to first off, turn off the memcached usage, which is really an amazing mistake, might want to blacklist that developer from contributing because their skill is lacking.

Second you need to make sure that the locations table provides a locking mechanism to prevent multiple reads at the same time from this particular check. This is probably going to be something that isn't easy given I would expect the need to make the actual dbFetchRow call to support a full table lock on reads since the code treats this as a critical section without actually locking it!

You might want to consider reeducation when it comes to locking mechanics. These things are vital when working with databases and parallel programming. Nobody has any business writing code using parallel processing and async methodology if they don't understand how locks work.

laf:
I'm actually not sure how to respond to this.

You may or may not be aware of our history but in general we inherited a pretty horrible code base and people have worked to improve it, is it perfect still - no, will it ever be - probably not.

'That' developer who added memcached did so quickly to fix an utterly broken memcache code base so it's a hell of a lot better than it was. And no, we won't blacklist them, they've contributed a hell of a lot to this project.

Duplicate information in the locations table is hardly a major issue, are you seeing 1,00's of duplicates?

We appreciate feedback but please be constructive, telling us to blacklist contributors from the project when you in turn haven't contributed anything isn't going to do anything but annoy people.

We look forward to seeing your pull requests to fix these issues.

drakej:
[...]

PR #6062, opened 2017-03-02
fix: Stop duplication locations within the DB #4930

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Removes duplicates and then creates a unique index on lat,lng.

Moved location lookup to discovery to save lookups on busy networks, I don't think it's a concern that we don't pick the change up within 5 minutes.


towster:
I have locations that are physically same lat/lng but with different names. For example a building with different floors. I think from reading this that it would not allow me to have two locations with same lat,lng.

laf:
We only do the following lookup: SELECT `lat`,`lng` FROM `locations` WHERE `location`=? LIMIT 1 which means that if lat/lng are seen already then we don't do another lookup which means this change doesn't affect how things work already

towster:
So the schema change will delete 144 of my locations. 

```
MariaDB [librenms]> select count(distinct(l1.id)) FROM `locations` `l1`, `locations` `l2` WHERE `l1`.`lat` = `l2`.`lat` AND `l1`.`lng` = `l2`.`lng` AND `l1`.`id` > `l2`.`id`;
+------------------------+
| count(distinct(l1.id)) |
+------------------------+
|                    144 |
+------------------------+

```

sql-schema/174.sql
```diff
@@ -0,0 +1,2 @@
+DELETE `l1` FROM `locations` `l1`, `locations` `l2` WHERE `l1`.`lat` = `l2`.`lat` AND `l1`.`lng` = `l2`.`lng` AND `l1`.`id` > `l2`.`id`;
+ALTER TABLE `locations` ADD UNIQUE( `lat`, `lng`);
```
geordish: Based on @towster's comments, should this be perhaps:

```SQL
DELETE `l1` FROM `locations` `l1`, `locations` `l2` WHERE `l1`.`lat` = `l2`.`lat` AND `l1`.`lng` = `l2`.`lng` AND `l1`.`location` = `l2`.`location` AND `l1`.`id` > `l2`.`id`;
ALTER TABLE `locations` ADD UNIQUE( `lat`, `lng`, `location`);
```

towster: Much better - I can have a building with multiple floors again..

```
MariaDB [librenms]> select count(distinct(l1.id)) FROM `locations` `l1`, `locations` `l2` WHERE `l1`.`lat` = `l2`.`lat` AND `l1`.`lng` = `l2`.`lng` AND `l1`.`location` = `l2`.`location` AND `l1`.`id` > `l2`.`id`;
+------------------------+
| count(distinct(l1.id)) |
+------------------------+
|                      0 |
+------------------------+
```

PR #6066, opened 2017-03-02
Adds column sort persistence

rmedlyn:
Adds session storage variables to persist the bootgrid column sorting for the alerts after both page refresh and automated ajax refresh.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/common/alerts.inc.php
+++ b/html/includes/common/alerts.inc.php
@@ -264,6 +264,17 @@ var alerts_grid = $("#alerts_'.$unique_id.'").bootgrid({
         }
       });
     });
+    if(sessionStorage.getItem("refresh")) {
+        sessionStorage.removeItem("refresh");
+        var $sortDictionary = JSON.parse(sessionStorage.getItem("sortDictionary"));
+        alerts_grid.bootgrid("sort", $sortDictionary);
+    }
+    alerts_grid.find("[data-column-id=rule], [data-column-id=hostname], [data-column-id=timestamp], [data-column-id=severity]").on("click", function(e) {
+        setTimeout(function() {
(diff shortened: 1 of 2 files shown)


laf:
Thank you very much for contributing this.

It's obviously very specific to alerts at present, seems it would make more sense to put this in a function and make it available to other bootgrid tables that we have in use today.

laf:
@rmedlyn Any feedback?

rmedlyn:
That does sound like a good idea. I'll look into it and get back to you.

rmedlyn:
I would think that column-sort persistence is a feature that should be implemented within bootgrid, but seeing as it has not been updated in a while (last update [Sept. 2015][1]), I don't see that happening anytime in the near future. My recommendation would be looking into [DataTables][2] ([GitHub repo][3]). It is well maintained and has the current desired functionality built-in. My code could be merged as a temporary fix until a switch is made. 

I did look into generalizing the function, but it seems like it will get really complicated really quickly for what should be a simple UI fix.

[1]: https://github.com/rstaib/jquery-bootgrid/commits/master
[2]: https://datatables.net/
[3]: https://github.com/DataTables/DataTables

laf:
We use datatables in v2 but no plans to port this to v1.

Imho a part fix isn't worth merging in.

I'll leave @librenms/reviewers to comment further if they want.

murrant:
Perhaps the fix is to disable the automated page refresh here?

laf:
lol, that's the simple fix :)

The small pain with that is the pages where we show graphs as they will stop auto refreshing but I see that as a small price to pay.

laf:
Closing as no further feedback provided.

PR #6081, opened 2017-03-03
feature: Added rrd step conversion script

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Works for me :)

### DIFF
--- a/scripts/rrdstep.php
+++ b/scripts/rrdstep.php
@@ -35,12 +35,14 @@ if (empty($hostname)) {
     echo "*********************************************************************\n";
     echo "We highly suggest that you back up your rrd files before running this\n";
     echo "*********************************************************************\n";
-    echo "-h <hostname or device id>       Device to process the rrd file for\n";
+    echo "-h <hostname or device id or all>  Device to process the rrd file for\n";
     echo "\n";
     exit;
 }
 
-$hostname = !ctype_digit($hostname) ? $hostname : gethostbyid($hostname);
+if ($hostname !== 'all') {
+    $hostname = !ctype_digit($hostname) ? $hostname : gethostbyid($hostname);
+}
 
 if (empty($hostname)) {
     echo "Invalid hostname or device id specified\n";
@@ -53,7 +55,10 @@ $rrd_path  = $config['rrd_dir'];
 $rrdtool   = $config['rrdtool'];
 $tmp_path  = $config['temp_dir'];
 
-$files = glob($rrd_path.'/'.$hostname.'/*.rrd');
+if ($hostname === 'all') {
+    $hostname = '*';
+}
+$files = glob($rrd_path . '/' . $hostname . '/*.rrd');
 
 $run = readline("Are you sure you want to run this command [N/y]: ");
 if ($run != 'y' || $run == 'Y') {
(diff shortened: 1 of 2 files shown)


wriedel:
Hi Laf, it's a little bit cumbersome to convert 90 devices that way but it's being worth every minute to get this 1 minute high resolution graphs ;-) Script worked like a charm and I hope we can get a wildcard option in the near future.

laf:
Added -h all.

Don't see the point in adding a wildcard.

scripts/rrdstep.php
```diff
@@ -0,0 +1,85 @@
...
+$rrdtool   = $config['rrdtool'];
+$tmp_path  = $config['temp_dir'];
+
+if ($hostname === 'all') {
+    $hostname = '*';
+}
+$files = glob($rrd_path . '/' . $hostname . '/*.rrd');
+
+$run = readline("Are you sure you want to run this command [N/y]: ");
+if ($run != 'y' || $run == 'Y') {
```
murrant: Looks like this should be: 
`+if ($run != 'y' || $run != 'Y')`

scripts/rrdstep.php
```diff
@@ -0,0 +1,85 @@
...
+    exit;
+}
+
+foreach ($files as $file) {
+    $random = $tmp_path.'/'.mt_rand() . '.xml';
+    $tmp = explode('/', $file);
+    $rrd_file = array_pop($tmp);
+    echo "Converting $file: ";
+    $command = "$rrdtool dump $file > $random;
+        sed -i 's/<step>300/<step>$step/' $random;
```
murrant: Perhaps use && in between these commands.  That way if one fails the rest are not executed.

murrant:
@laf I know you mentioned this elsewhere, but why can't you use rrdtune?

laf:
Because it doesn't work, try it. If you tune a file and then dump it you will see the step will remain the same.

I'm unsure if this is all versions but (iirc) I tested on 1.4.7 and 1.6.x

nrm21:
Hi,

This script works great!  However my graphs only update every 5 mins and they update with 5 new values.  On further inspection it appears the code is doing some sort of:
what is the current unix timestamp? ->
ok now lets round to the nearest 5 min step of that timestamp ->
we'll use that in the URL request for 'from=' and 'to=' ->
now we generate the graph using those values

...or it appears something like that bawsd on studying web behavior.

Maybe the graph portion of code could be changed to round to the nearest minute of the current timestamp instead of 5 min, once the script is applied?

nrm21:
Found it... in ./includes/definitions.inc.php

Change:     $config['time']['now']     -= ($config['time']['now'] % 300);
from mod 300 to mod 60

and voila!  Graphs now show by the min.  :)

laf:
Thanks but this has nothing to do with this script.

nrm21:
Ok maybe it doesn't but.  I assume that others like me might want to not just poll every min but also see an up to the min graph of their bandwidth data.  In that case, it might be necessary to do that to the default code to achieve it.

laf:
Feature requests need to be posted on https://community.librenms.org/c/feature-requests

murrant:

https://community.librenms.org/t/1-minute-polling-and-graphs/787/9

Issue #6063, opened 2017-03-02
Unable to create API token while using active-directory authentication


towster:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x ] Please provide as much detail as possible.
- [x ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

When clicking on create API token button the dropdown for the user list is empty.

laf:
Could you test #6082

lunacite:
This worked for me when I manually hacked it into my docker container, however I was unable to continue using old api tokens and had to generate new ones

PR #6082, opened 2017-03-03
fix: Update api-access page to use get_userlist() #6063

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6063

### DIFF
--- a/html/pages/api-access.inc.php
+++ b/html/pages/api-access.inc.php
@@ -54,7 +54,7 @@ if (is_admin() === true) {
               <div class="col-sm-4">
                 <select class="form-control" id="user_id" name="user_id">
 <?php
-foreach (get_userlist() as $users) {
+foreach ($userlist = get_userlist() as $users) {
     echo '<option value="'.$users['user_id'].'">'.$users['username'].'</option>';
 }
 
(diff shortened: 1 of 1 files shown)


laf:
Confirmed working in the issue.

html/pages/api-access.inc.php
```diff
@@ -136,15 +136,21 @@
         </tr>
 ';
 
-foreach (dbFetchRows('SELECT `AT`.*,`U`.`username` FROM `api_tokens` AS AT JOIN users AS U ON AT.user_id=U.user_id ORDER BY AT.user_id') as $api) {
+$tmp_users = get_userlist();
```
murrant: Why call get_userlist() again when we already got that data a little bit ago?

html/pages/api-access.inc.php
```diff
@@ -136,15 +136,21 @@
...
+foreach (dbFetchRows('SELECT * FROM `api_tokens` ORDER BY user_id') as $api) {
     if ($api['disabled'] == '1') {
         $api_disabled = 'checked';
     } else {
         $api_disabled = '';
     }
+    foreach ($tmp_users as $tmp_user) {
+        if ($tmp_user['user_id'] === $api['user_id']) {
+            $user_details = $tmp_user;
```
murrant: A `break;` here would probably be a good idea.

murrant:
Any thoughts on how to fix the incorrect user ids for ldap auth?  changes in the auth?

laf:
ldap auth get_userlist() and get_userid() looks like it gets those values from ldap itself so we would then insert that into the db. I'm not sure why you'd get different IDs unless LDAP itself does that.

murrant:
Hmm, likely because my ldap and sql username matches...

murrant:
OK, well this bug is in the ldap auth, so I think this is ok.

PR #6093, opened 2017-03-05
Add error checking for DS name length

geordish:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/RRD/RrdDefinition.php
```diff
@@ -50,13 +51,20 @@ public static function make()
...
      * @param int $heartbeat Heartbeat for this dataset. Uses the global setting if null.
      * @return $this
+     * @throws InvalidRrdNameException
      */
     public function addDataset($name, $type, $min = null, $max = null, $heartbeat = null)
     {
         global $config;
+        $name = $this->escapeName($name);
+
+        if (strlen($name) == 0 || strlen($name) > 19) {
```
laf: The 0 length checks is fine but > 19 isn't as escapeName returns `substr($name, 0, 19);` so it will never match.

geordish: I agree that > 19 should probably never happen, but I put the check in, in the case that code below changes at some point. I can remove if you want.

murrant: Add a phpunit test instead of a runtime check.

geordish:
@murrant good point. Makes sense.

My first attempt at unit tests. Be nice!

murrant:
FYI, I looked and the ds name being too long is already covered by the unit tests.

https://github.com/librenms/librenms/blob/master/tests/RrdDefinitionTest.php#L50

geordish:
Doh, so it is. Would the test to make sure the DS name is > 0 characters not be useful stop though?

murrant:
@geordish I think so, the rrdcreate manual says the DS name size must be 1-19.

LibreNMS/RRD/RrdDefinition.php
```diff
@@ -50,13 +51,20 @@ public static function make()
...
      * @param int $heartbeat Heartbeat for this dataset. Uses the global setting if null.
      * @return $this
+     * @throws InvalidRrdNameException
      */
     public function addDataset($name, $type, $min = null, $max = null, $heartbeat = null)
     {
         global $config;
+        $name = $this->escapeName($name);
+
+        if (strlen($name) == 0) {
```
murrant: `if (empty($name)) {`  will work fine here.  empty() is true for NULL or ''

LibreNMS/RRD/RrdDefinition.php
```diff
@@ -50,13 +51,20 @@ public static function make()
...
+            $msg = "$name is an Invalid name for a DS";
```
murrant: Update the message to indicate that DS name must be set to a non-empty string.

laf:
Looks good to me.

Worth asking though and this is the same for the other code in this file that throws exceptions - should we really stop all of poller.php just because one rrd file couldn't be created?

murrant:
@laf I hope the exception never gets thrown during production.  This is meant to catch errors during development.

laf:
@murrant How's this going to get caught in dev? The test is for the function to show it works as expected but if someone adds code where the DS ends up being null then poller stops. Now you might say someone would notice if they did this before submitting but then that's leaving this down to trust.

murrant:
I agree, I don't think we can test for this right now.  Perhaps if we can go through all definitions in tests, but that isn't an option right now with our code structure.

laf:
Indeed but the issue for me still remains that we stop polling for one issue with a single rrd. Imho we shouldn't do that.

murrant:
So, should we just close this?

laf:
Personally it just needs to be updated to not throw the error and stop polling, we should just add it to the debug output or log to logs/librenms.log like we do with the sql queries.

Issue #6080, opened 2017-03-03
Add Extrahop Appliance as a Supported Device.


thejimmahknows:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
:/opt/librenms# ./validate.php 
==========================================================
Component | Version
--------- | -------
LibreNMS  | 41824dda15641eeba3f5af00b9d537642fe13e55
DB Schema | 173
PHP       | 5.6.30-0+deb8u1
MySQL     | 5.5.54-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1
==========================================================

- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
[extrahop_discovery.txt](https://github.com/librenms/librenms/files/818019/extrahop_discovery.txt)
[extrahop_poller.txt](https://github.com/librenms/librenms/files/818020/extrahop_poller.txt)
[extrahop_snmpwalk.txt](https://github.com/librenms/librenms/files/818021/extrahop_snmpwalk.txt)

##heres the  MIB
[EXTRAHOP-MIB.txt](https://github.com/librenms/librenms/files/818032/EXTRAHOP-MIB.txt)

- [x] Please provide as much detail as possible.
Request is to add  LibreNMS support for Extrahop Appliances ( https://www.extrahop.com/ )
- [x ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

geordish:
I just loaded your snmpwalk into my install, and it looks fairly well supported. What are you missing?

thejimmahknows:
The 10Gig NIC (eth4+5) are showing connected(bold) but not showing stats. I'm not sure if the SNMP walk contains the serial number, Extrahop Discover Appliance version running, or any other sensors.

[...]

Issue #4508, opened 2016-09-20
CiscoSB PPS Multicast/Broadcast Issue


Rosiak:
Just registering an issue on this, don't have the time for a fix right now.
Multicast/Broadcast only works using ifXEntry however Unicast only works with ifEntry..

ifXEntry

```
ifName.49 = gi1
ifName.50 = gi2
ifName.51 = gi3
ifName.52 = gi4
ifName.53 = gi5
ifName.54 = gi6
ifName.55 = gi7
ifName.56 = gi8
ifName.57 = gi9
ifName.58 = gi10
ifName.1000 = Po1
ifName.1001 = Po2
ifName.1002 = Po3
ifName.1003 = Po4
ifName.1004 = Po5
ifName.1005 = Po6
ifName.1006 = Po7
ifName.1007 = Po8
ifName.3000 = tunnel1
ifName.7000 = loopback1
ifName.20000 = Logical-int 1
ifName.100000 = 1
ifName.100009 = 10
ifName.100019 = 20
ifName.100029 = 30
ifName.100039 = 40
ifName.100041 = 42
ifName.100049 = 50
ifName.300000 = 42
ifInMulticastPkts.49 = 0
ifInMulticastPkts.50 = 0
ifInMulticastPkts.51 = 0
ifInMulticastPkts.52 = 0
ifInMulticastPkts.53 = 0
ifInMulticastPkts.54 = 0
ifInMulticastPkts.55 = 131376
ifInMulticastPkts.56 = 0
ifInMulticastPkts.57 = 0
ifInMulticastPkts.58 = 579334
ifInMulticastPkts.1000 = 0
ifInMulticastPkts.1001 = 0
ifInMulticastPkts.1002 = 0
ifInMulticastPkts.1003 = 0
ifInMulticastPkts.1004 = 0
ifInMulticastPkts.1005 = 0
ifInMulticastPkts.1006 = 0
ifInMulticastPkts.1007 = 0
ifInBroadcastPkts.49 = 0
ifInBroadcastPkts.50 = 0
ifInBroadcastPkts.51 = 0
ifInBroadcastPkts.52 = 0
ifInBroadcastPkts.53 = 0
ifInBroadcastPkts.54 = 0
ifInBroadcastPkts.55 = 209210
ifInBroadcastPkts.56 = 0
ifInBroadcastPkts.57 = 0
ifInBroadcastPkts.58 = 86490
ifInBroadcastPkts.1000 = 0
[...]

PR #6106, opened 2017-03-06
Use HC counters even if one direction is reporting as 0

geordish:
Second fix for #6080 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -446,7 +446,8 @@ foreach ($ports as $port) {
         }
 
         // use HC values if they are available
-        if (isset($this_port['ifHCInOctets'])) {
+        if (!isset($this_port['ifInOctets')) {
+            echo "HC ";
             $this_port['ifInOctets'] = $this_port['ifHCInOctets'];
             $this_port['ifOutOctets'] = $this_port['ifHCOutOctets'];
             $this_port['ififInUcastPkts'] = $this_port['ifHCInUcastPkts'];
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -446,13 +446,18 @@ foreach ($ports as $port) {
         }
 
         // Copy ifHC[In|Out]Octets values to non-HC if they exist
-        if ($this_port['ifHCInOctets'] > 0 && is_numeric($this_port['ifHCInOctets']) && $this_port['ifHCOutOctets'] > 0 && is_numeric($this_port['ifHCOutOctets'])) {
+        if ($this_port['ifHCInOctets'] > 0 && is_numeric($this_port['ifHCInOctets'])) {
             echo 'HC ';
             $this_port['ifInOctets']  = $this_port['ifHCInOctets'];
+        }
+        if ($this_port['ifHCOutOctets'] > 0 && is_numeric($this_port['ifHCOutOctets'])) {
+            echo 'HC ';
             $this_port['ifOutOctets'] = $this_port['ifHCOutOctets'];
         }
-        if (is_numeric($this_port['ifHCInUcastPkts']) && $this_port['ifHCInUcastPkts'] > 0 && is_numeric($this_port['ifHCOutUcastPkts']) && $this_port['ifHCOutUcastPkts'] > 0) {
+        if (is_numeric($this_port['ifHCInUcastPkts']) && $this_port['ifHCInUcastPkts'] > 0) {
             $this_port['ifInUcastPkts']  = $this_port['ifHCInUcastPkts'];
+        }
+        if (is_numeric($this_port['ifHCOutUcastPkts']) && $this_port['ifHCOutUcastPkts'] > 0) {
             $this_port['ifOutUcastPkts'] = $this_port['ifHCOutUcastPkts'];
         }
 
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -445,17 +445,12 @@ foreach ($ports as $port) {
             $port['update']['poll_period'] = $polled_period;
         }
 
-        // usee HC values if they are available
-        $hc_names = array(
-            'ifHCInOctets' => 'ifInOctets',
-            'ifHCOutOctets' => 'ifOutOctets',
-            'ifHCInUcastPkts' => 'ifInUcastPkts',
-            'ifHCOutUcastPkts' => 'ifOutUcastPkts',
-        );
-        foreach ($hc_names as $hc => $non_hc) {
-            if (is_numeric($this_port[$hc]) && $this_port[$hc] > 0) {
-                $this_port[$non_hc] = $this_port[$hc];
-            }
+        // use HC values if they are available
+        if (is_numeric($this_port['ifHCInOctets']) && $this_port['ifHCInOctets'] >= 0) {
+            $this_port['ifInOctets'] = $this_port['ifHCInOctets'];
+            $this_port['ifOutOctets'] = $this_port['ifHCOutOctets'];
+            $this_port['ififInUcastPkts'] = $this_port['ifHCInUcastPkts'];
+            $this_port['ififOutUcastPkts'] = $this_port['ifHCOutUcastPkts'];
         }
 
         if ($device['os'] === 'airos-af' && $port['ifAlias'] === 'eth0') {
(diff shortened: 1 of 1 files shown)


laf:
Is this the correct way to do this? Seems to me mixing HC / NonHC together to fix bad snmp from one vendor might not be the best thing to do.

geordish:
Its more the > 0 part that is bad. IE the counters ingress don't update, so
the egress counters don't get looked at. There is an example on the linked
to issue.

Perhaps just remove the test for > 0?

laf:
I'm guessing the check is originally there so that if we don't have HC counters it switches back.

I've not looked at the output in the issue much but is it not a cause of (might have got this the wrong way round) ifHCInOctets is > 0 but Out isn't so we use HC for in but non-HC for out?

geordish:
No, its a case of ifHCInOctets  = 0 (ie never increased. This looks to be
right) so both are ignored.

murrant:
@geordish counters should almost never be 0.  If ifHCInOctets was 1255 and receives no packets, next poll it will be 1255.

geordish:
@murrant  But if its 0, and it receives no packets, it will be 0. Never receiving packets on a port doesn't mean you will never transmit (or vice-versa)

murrant:
That is true.

laf:
The loop means that you can have IN pulling it's stats from HC and OUT pulling from non-HC. I don't see why we would do this.

I much prefer the original code to just have the addition of >= 0 so we either use HC or non-HC for both In and Out.

murrant:
That's true @laf, but that is the point of this PR.  What if you have an interface that only transmits or only receives?

The HC counter and normal counter should match as long as it is less than uint32...

laf:
What about only having an interface that only transmits or receives? Do you expect to see something other than 0 or greater in that situation?

I don't see why we would choose to chop and change with HC and non-HC counters for a single interface. Pick one (which we already do) and stick to it. The issue I think we're trying to solve is that a counter of 0 would force us to non-HC, >= fixes that.

geordish:
In fairness, only ifOctets or ifHCOctets are polled now, right?

laf:
Not sure what you mean by that?

geordish:
I don't really know what I meant...

If ifHCInOctets is present, ifInOctets and ifOutOctets won't be collected. Just checking ifHCInOctets is present and correct should be enough to copy both across.

laf:
That's what I think we should do and is what the current code does.

geordish:
That is not quite what the current code does. It checks for both ifHCInOctets and ifHCOutOctets, as well as the unicast packets being set.

Pushing another go at this.

laf:
I didn't mean strictly octets, I know it's also doing unicast HC as well

murrant:
FYI, I do seem to remember some device always returning 0 for HC counters.

laf:
Then this PR will break those devices. I guess the question then is, is returning 0 for ifHCInOctet valid when that device doesn't support / use them (i.e they are always 0). If it is then we need to look at another way to achieve this, if it's not then we should merge but have a device config flag to skip hc.

geordish:
Those devices will already be broken then for a long time. As it stands, it
fetches the HC counters, and if they exist it doesn't fetch the standard
counters.

laf:
But only if HC is > 0 so broken devices that return 0 in HC will skip to non-HC in the current code.

geordish:
No, it just tests ifHCInOctets exists, and its numeric:
```php
        if (!isset($hc_test[0]['ifHCInOctets']) && !is_numeric($hc_test[0]['ifHCInOctets'])) {
            $port_stats = snmpwalk_cache_oid($device, 'ifEntry', $port_stats, 'IF-MIB', null, '-OQUst');
        }
```

laf:
That's a different part of the code than what you've changed here.

geordish:
Yes, but that is the bit of code that decides which OIDs are queried. If that test fails, only HC counters are polled, so it doesn't matter if they are 0 or not, as the standard ones don't exist anyway.

includes/polling/ports.inc.php
```diff
@@ -445,15 +445,12 @@
             $port['update']['poll_period'] = $polled_period;
         }
 
-        // Copy ifHC[In|Out]Octets values to non-HC if they exist
-        if ($this_port['ifHCInOctets'] > 0 && is_numeric($this_port['ifHCInOctets']) && $this_port['ifHCOutOctets'] > 0 && is_numeric($this_port['ifHCOutOctets'])) {
```
murrant: In that case, how about just:
```
if (!isset($this_port['ifInOctects']) {
```

includes/polling/ports.inc.php
```diff
@@ -445,15 +445,12 @@
...
-            echo 'HC ';
-            $this_port['ifInOctets']  = $this_port['ifHCInOctets'];
+        // use HC values if they are available
+        if (is_numeric($this_port['ifHCInOctets']) && $this_port['ifHCInOctets'] >= 0) {
```
murrant: In that case, how about just:
```
if (!isset($this_port['ifInOctects']) {
```

laf:
Then do we not just need to change to:

`if (!is_numeric($hc_test[0]['ifHCInOctets']) || $hc_test[0]['ifHCInOctets'] < 1) {` or something similar?

This in place of: `if (!isset($hc_test[0]['ifHCInOctets']) && !is_numeric($hc_test[0]['ifHCInOctets'])) {`

includes/polling/ports.inc.php
laf: Isn't that still the same issue as @geordish just pointed out that we actually haven't polled for ifInOctets at that point in some situations because we've already decided that HC counter being 0 is valid (which apparently isn't always the case).

geordish:
That is probably quite bad. If the first port has no traffic on it, then only standard counters will be collected regardless.

@murrant  Do you know of an example device where HC counters are reported, but blank?

Considering this code which has been in place now for probably around 6 months without @murrant's bug being reported, is it an issue?

Do we just use @murrant's shorter check in the comment above in that case?

murrant:
@laf my point is we checked `if (!isset($hc_test[0]['ifHCInOctets']) && !is_numeric($hc_test[0]['ifHCInOctets']))`  on line 221. shouldn't we trust that?  Since we don't have the data anyway if that check is wrong.

geordish:
@Rosiak mentioned #4508

laf:
@murrant if that line passes then we actually have both HC and non-HC counters all because we say that ifHCInOctets is not numeric, so we could have all the other HC counters but miss In and then poll for non-hc.

murrant:
@laf That seems hypothetical.  But if that is the case then the code currently has that bug.

All I'm saying is check once.
IF we have valid HC data, copy it to the normal oids.
ELSE fetch the normal oids.

laf:
That's what we're trying to get to (although I'm getting lost with this now as we're all chopping and changing with the sections of code we're talking about).

Are we now just talking about changing: `if ($this_port['ifHCInOctets'] > 0 && is_numeric($this_port['ifHCInOctets']) && $this_port['ifHCOutOctets'] > 0 && is_numeric($this_port['ifHCOutOctets'])) {` to be >=?

geordish:
@laf - I'm not sure that is necessary. As the code which collects the HC data only checks ```if (!isset($hc_test[0]['ifHCInOctets']) && !is_numeric($hc_test[0]['ifHCInOctets']))``` to decide weather to poll the non HC data, the extra checks later on are redundant.

The bit where the HC data is copied across just needs to check ```isset($hc_test[0]['ifHCInOctets'])```.

I will push what I mean.

includes/polling/ports.inc.php
```diff
@@ -445,15 +445,12 @@
             $port['update']['poll_period'] = $polled_period;
         }
 
-        // Copy ifHC[In|Out]Octets values to non-HC if they exist
-        if ($this_port['ifHCInOctets'] > 0 && is_numeric($this_port['ifHCInOctets']) && $this_port['ifHCOutOctets'] > 0 && is_numeric($this_port['ifHCOutOctets'])) {
-            echo 'HC ';
-            $this_port['ifInOctets']  = $this_port['ifHCInOctets'];
+        // use HC values if they are available
+        if (isset($this_port['ifHCInOctets'])) {
```
murrant: Won't this always be set?  Shouldn't it be `if (!isset($this_port['ifInOctets')) {`

laf:
That's fine by me (although it's worth adding `echo 'HC ';` back in to help with debugging.

murrant:
Looks good to me.  I will try test it a bit.

Issue #6092, opened 2017-03-05
Alert History not displaying correctly for access_point entity


djamp42:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.
==========================================================
Component | Version
--------- | -------
LibreNMS  | a9ad228be444e1b5d1b8040e431f9d50fc8e2dcb
DB Schema | 173
PHP       | 7.0.15-0ubuntu0.16.04.4
MySQL     | 10.0.29-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3
==========================================================
I have a Cisco WLC that is working perfectly with LibreNMS I get a Alerts with the following rule. "%access_points.nummonclients >= "10"" 

This Alert displays correctly in the Notification (While the alert is happening). There is a **+** box next to the IP of the WLC to show the access points affected.

It does NOT display correctly in the Alert History. The same **+** box is there, but does not do anything, it does not display the list of access point affected by the alert that cleared or acknowledged. It just changes to **-**

![alert log access point expansion broken final](https://cloud.githubusercontent.com/assets/25335691/23584370/893c8e1c-012d-11e7-861f-1b40757bf397.jpg)

djamp42:
Actually this does it for any alert, even the default port up/down doesn't display the ports.

PR #6107, opened 2017-03-06
fix: Fix the displaying of alert info for historical alerts #6092

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

The way alerts work is they register an alert_log entry for every device and every rule (not sure why). Because of this rules / devices that have never triggered will have blank details (now shows 'Rule created, no faults found') and rules which have previously triggered we look for the last alert where the state = 1 which means it was triggered and grabs the data we need from that.

Fixes: #6092


djamp42:
Just applied this, working perfectly fine for me! Thanks!

PR #6109, opened 2017-03-06
security: Stop multiport_bits_separate graphs for showing regardless of auth #6101

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6101

### DIFF
--- a/html/includes/graphs/multiport/auth.inc.php
+++ b/html/includes/graphs/multiport/auth.inc.php
@@ -1,10 +1,8 @@
 <?php
 
-$auth = true;
-
 foreach (explode(',', $vars['id']) as $ifid) {
-    if (!$auth && !port_permitted($ifid)) {
-        $auth = false;
+    if ($auth || port_permitted($ifid)) {
+        $auth = true;
     }
 }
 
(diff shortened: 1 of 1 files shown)


html/includes/graphs/multiport/auth.inc.php
```diff
@@ -1,10 +1,8 @@
 <?php
 
-$auth = true;
-
 foreach (explode(',', $vars['id']) as $ifid) {
-    if (!$auth && !port_permitted($ifid)) {
-        $auth = false;
+    if ($auth || port_permitted($ifid)) {
+        $auth = true;
```
geordish: Would this not be worth writing as
if (port_permitted($ifid)) {
  $auth = true;
  break;
}

murrant: These both seem wrong.  Only should be allowed if $auth or all ports are permitted.
```
if (!$auth) {
    foreach ($ids as $ifid) {
        if ($ifid) {
            $auth = true;
        } else {
            $auth = false;
            break;
        }
    }
}
```

laf:
Updated

PR #6114, opened 2017-03-07
newdevice: Improve 3com switch detection

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixes: #6104


murrant:
`.1.3.6.1.4.1.43.` showed no devices we shouldn't have as this OS on stats, but only  `.1.3.6.1.4.1.43.1.` and `.1.3.6.1.4.1.43.10.` were present as well.  Preference?

laf:
Why replace the test, seems valid data:

`|    2|3com   |.1.3.6.1.4.1.43.1.16.4.3.45   | 3Com Switch 4210 52-Port Software Version 3.10 Release 2215P12 |`

murrant:
I thought we had enough data, no need for an extra test.  I can add an extra test instead.

PR #6127, opened 2017-03-08
alphabatize doc/Extensions/Applications.md 

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
This doesn't really make sense to me, it's not following the order of the list. I think Agent should be always at the top anyway. Actually this whole section could do with a revamp, maybe one application per file would be best seeing we have quite a few.

murrant:
I don't like the Agent being at the top because then you read from top down and conclude you need to install the agent.  What do you think about moving the agent install to a separate page?

laf:
No ones going to read that entire page and realise they need the agent at the end. Logically you'd assume the first step was to install the agent but yes the doc needs a revamp and the agent install can be one of the first on it's own.

murrant:
Agent Setup broken out to it's own page.

PR #6128, opened 2017-03-08
add SMART support

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Depends on https://github.com/librenms/librenms-agent/pull/101 .


laf:
imho the script either needs to auto detect disks (df?) or the disks to poll data for needs to be in a separate config file.

VVelox:
There is sadly no good way to choose what devices too poll automatically. Especially once you begin factoring things in like how Linux and FreeBSD display USB mass storage devices, which don't no SMART.

If we are going to extract the disk config, from script, we should should really just do it with the entire config.

laf:
Sorry I should have been clear, I was talking about the perl script (conversations are just easier here).

murrant:
The poller module could be shortened quite a bit: 
```php
<?php

use LibreNMS\RRD\RrdDefinition;

echo('S.M.A.R.T.');

$name = 'smart';
$app_id = $app['app_id'];

$oid = 'nsExtendOutputFull.5.115.109.97.114.116';
$output = snmp_walk($device, $oid, '-Oqv', 'NET-SNMP-EXTEND-MIB');

$data_points = array(
    'id5',
    'id10',
    'id173',
    'id177',
    'id183',
    'id184',
    'id187',
    'id188',
    'id190',
    'id194',
    'id196',
    'id197',
    'id198',
    'id199',
    'id231',
    'id233',
    'completed',
    'interrupted',
    'readfailure',
    'unknownfail',
    'extended',
    'short',
    'conveyance',
    'selective',
);

foreach (explode("\n", $output) as $line) {
    $full_data = explode(',', $line);
    $disk = array_shift($full_data);  // remove the first item from the array
    $rrd_name = array('app', $name, $app_id, $disk);
    
    $data = array_combine($data_points, $full_data);
    $fields = array();
    $rrd_def = new RrdDefinition();
    foreach ($data as $ds => $value) {
        $rrd_def->addDataset($ds, 'GAUGE', 0);
        $fields[$ds] = is_int($value) ? $value : null;
    }

    $tags = compact('name', 'app_id', 'rrd_def, 'rrd_name');
    data_update($device, 'app', $tags, $fields);
}
```

VVelox:
Updated it to use a separate con fig file now and documented it in the script. Still need to update the docs.

VVelox:
@murrant Tested out your suggestion there and that is breaking polling for me, even after replacing the $tags line with the normal one. Replacing that causes the polling to work, but it stops updating the RRDs.

VVelox:
Updated the extend again it now has a -g option that will guess and test and print out what is most likely a good config for most common disk devices.

VVelox:
Wooho! This is now done.

laf:
Need to use update_application(); now in the poller file + needs a rebase.

VVelox:
rebased and seems to of not gone well

VVelox:
Replaced by https://github.com/librenms/librenms/pull/6181 .

PR #6135, opened 2017-03-08
ensure mib-name passed as parameter

barryodonovan:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/functions.inc.php
+++ b/includes/polling/functions.inc.php
@@ -386,15 +386,14 @@ function poll_mib_def($device, $mib_name_table, $mib_subdir, $mib_oids, $mib_gra
     echo "This is poll_mib_def Processing\n";
     $mib = null;
 
-    if (stristr($mib_name_table, 'UBNT')) {
-        list($mib,) = explode(':', $mib_name_table, 2);
-        $measurement_name = strtolower($mib);
-    } else {
-        list($mib,$file) = explode(':', $mib_name_table, 2);
-        $measurement_name = strtolower($file);
-    }
+    list($mib, $file) = explode(':', $mib_name_table, 2);
+
     if (is_null($rrd_name)) {
-        $rrd_name = $measurement_name;
+        if (str_contains($mib_name_table, 'UBNT', true)) {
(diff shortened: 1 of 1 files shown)


includes/polling/functions.inc.php
```diff
@@ -386,16 +386,12 @@ function poll_mib_def($device, $mib_name_table, $mib_subdir, $mib_oids, $mib_gra
...
 
-    if (is_null($rrd_name)) {
-        if (stristr($mib_name_table, 'UBNT')) {
-            list($mib,) = explode(':', $mib_name_table, 2);
-            $measurement_name = strtolower($mib);
-        } else {
-            list($mib,$file) = explode(':', $mib_name_table, 2);
-            $measurement_name = strtolower($file);
-        }
+    if (stristr($mib_name_table, 'UBNT')) {
```
murrant: we have a function called str_contains(), please use that.

murrant:
This code seems confusing to me, can it just be simplified to (including the section below):
```
    if (is_null($rrd_name)) {
        list($mib,$file) = explode(':', $mib_name_table, 2);
        if (str_contains($mib_name_table, 'UBNT', true)) {
            $rrd_name = strtolower($mib);
        } else {
             $rrd_name = strtolower($file);
        }
    }
```

barryodonovan:
Aaarrrgghhh!

I can't understand what is happening!

The only change need to be a parameter, rrd_name =null and then an if rrd_name == null, rrd_name = measurement_name

For some reason the code is correct on my dev box but not on github...

Regarding the str_contains, that's the original code, I'm all in favor of cleaning it up, but that's how it was merged last year... Then again, I'm also in favor of native functions where available...

Barry

murrant:
strstr() does not always behave as expected.  The function simply extracts a string from another string.   

The reason we added that function is because we have strops() strstr() and their case insensitive variants all over the place inconsistently.

As for the code it is a very complicated piece of code with too many temp variables just to set $rrd_name to one of three values.

barryodonovan:
Fair enough, good points.

I'll give it all another shot today.

PR #6139, opened 2017-03-08
refactor: Cisco ASA Polling Performance Improvements

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/os/asa.inc.php
+++ b/includes/polling/os/asa.inc.php
@@ -10,25 +10,26 @@
  * option) any later version.  Please see LICENSE.txt at the top level of
  * the source code distribution for details.
  */
+if (!str_contains($device['hardware'], 'sc')) {
+    $oids = 'entPhysicalModelName.1 entPhysicalSoftwareRev.1 entPhysicalSerialNum.1 entPhysicalModelName.4 entPhysicalSoftwareRev.4';
 
-$oids = 'entPhysicalModelName.1 entPhysicalSoftwareRev.1 entPhysicalSerialNum.1 entPhysicalModelName.4 entPhysicalSoftwareRev.4';
+    $data = snmp_get_multi($device, $oids, '-OQUs', 'ENTITY-MIB');
 
-$data = snmp_get_multi($device, $oids, '-OQUs', 'ENTITY-MIB');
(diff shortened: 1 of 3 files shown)


includes/polling/os/asa.inc.php
```diff
@@ -10,25 +10,26 @@
...
-
-if (isset($data[1]['entPhysicalModelName']) && $data[1]['entPhysicalModelName'] != '') {
-    $hardware = $data[1]['entPhysicalModelName'];
-} elseif (isset($data[4]['entPhysicalSoftwareRev']) && $data[4]['entPhysicalSoftwareRev'] != '') {
-    $hardware = $data[4]['entPhysicalModelName'];
-}
-
-if (isset($data[1]['entPhysicalSerialNum']) && $data[1]['entPhysicalSerialNum'] != '') {
-    $serial = $data[1]['entPhysicalSerialNum'];
+if (!str_contains($device['hardware'], 'sc')) {
```
laf: What's the sc here?

includes/polling/ports.inc.php
```diff
@@ -227,7 +227,7 @@
 if ($config['enable_ports_etherlike']) {
     echo 'dot3Stats ';
     $port_stats = snmpwalk_cache_oid($device, 'dot3StatsEntry', $port_stats, 'EtherLike-MIB');
-} else {
+} elseif ($device['os'] != 'asa') {
```
laf: Seems like we should really add something like:

poller_sub_modules:
    ports_dot3: 0
    ports_dot1q: 0

Into the yaml and check that as we'll have more devices to add to that if() list.

I've made those array names up btw, we need to pick something decent.

includes/polling/os/asa.inc.php
Rosiak: System Context / Multi Context device.

includes/polling/ports.inc.php
Rosiak: Fully agree here, however personally I don't have much time at hand right now :(

laf: @murrant what's your thoughts on this. It can be extended out to other poller modules where we do checks for varying OS' to include / exclude stuff to poll.

murrant: Maybe, don't want to end up with tons of modules/submodules.
It does seem to make sense in this circumstance.

PR #6143, opened 2017-03-09
feature: Application discovery

murrant:
Automatically detects and enables/disables snmp extend applications, will not disable user-enabled applications
Application page revamp
Only enabled by default for Linux
Work around some issues with bootstrap-switch

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Haha, Only just seen this after I updated FTBZ's PR:

https://github.com/librenms/librenms/pull/6153/files

It's only a single OID so a walk isn't needed. I wasn't sure if seconds on batter was right for runtime, the values for this are in minutes so if we do use it then it should have a divisor of 60

laf:
I've updated the ajax stuff to use json + toastr for notifications.

Also changed the toggle switch to a smaller size.

All tested so feel free to merge if you don't think I've made it worse :)

laf:
It would potentially only revert if the http call completely failed, if it just hit the `if (result.status == 0) {` block and below it then I don't think it would.

Issue #5977, opened 2017-02-22
Issues with the Generex UPS card and RFC1628 support


FTBZ:
- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.

------------------

Component | Version
--------- | -------
LibreNMS  | a8691680abb479ba7398d520fff01c1cffe228a0
DB Schema | 166
PHP       | 7.0.11
MySQL     | 5.5.50-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2

------------------

Discovery: http://pastebin.com/t0FNfSDS
Poller: http://pastebin.com/KQHUJHVi
Bulk: http://pastebin.com/sWPM4yBC

------------------

The Generex SNMP Card is a generic UPS card using the RFC1628 implementation in LibreNMS. Globally, it's working, but some glitches make LibreNMS not the best choice to monitor this kind of equipment.

1 - The state sensors are not working and need to be moved to the new format with more relevant sensors (like upsBatteryStatus and upsOutputSource):
<img width="737" alt="screen shot 2017-02-22 at 15 59 49" src="https://cloud.githubusercontent.com/assets/1927109/23216994/1f9c2f8e-f918-11e6-88ec-28534e59835f.png">
The graphs "Seconds on Battery" and "Battery: Estimated minutes remaining" should be moved to the charge section to not lost the data.

2 - At the top right, the "Battery Voltage" isn't working. When clicking on, a permission error is returned:

<img width="1111" alt="screen shot 2017-02-22 at 16 04 37" src="https://cloud.githubusercontent.com/assets/1927109/23217208/c478bdba-f918-11e6-8536-f40cf83c3f49.png">

<img width="638" alt="screen shot 2017-02-22 at 16 08 10" src="https://cloud.githubusercontent.com/assets/1927109/23217321/2015654c-f919-11e6-8a83-d2bd23f00aa4.png">

3 - The load values are not using a correct divisor:
<img width="528" alt="screen shot 2017-02-22 at 16 10 41" src="https://cloud.githubusercontent.com/assets/1927109/23217428/7884478e-f919-11e6-9aee-b51017057315.png">

[...]

PR #6153, opened 2017-03-09
fix: Adding state sensor support for RFC1628 UPS

FTBZ:

#### Testers

------

screen shot 2017-03-09 at 13 23 54
screen shot 2017-03-09 at 13 22 36

- Adding two state sensors for RFC1628 using the new format (based on the RFC it's a single OID, don't need to get multiple).
- Removed the two precedent informations that were not a sensor (Estimated minutes remaining & Minutes on Battery) because I don't know where to put this information, no category for these kind of graphs.

Details: #5977


laf:
Thanks for this @FTBZ .

One of the removed items is runtime so I've recreated that.

If you don't mind pulling that change in and testing it that would be ace.

FTBZ:
@laf, @murrant has create the runtime support in the PR #6158.

PR #6156, opened 2017-03-09
fix: Fix Generex voltage sensor divisor

FTBZ:

#### Testers

screen shot 2017-03-09 at 15 03 30

- Fixed the sensor divisors for voltages (issue confirmed by many members)
- Added an option to exclude a specific sensor because Generex card use different divisors for the same sensor type

Details: #5977

### DIFF
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -897,7 +897,7 @@ function avtech_add_sensor($device, $sensor)
  * @param $sensor
  * @return int
  */
-function get_device_divisor($device, $serial, $sensor)
+function get_device_divisor($device, $serial, $sensor, $exclude)
 {
     if ($device['os'] == 'poweralert') {
         if ($sensor == 'current' || $sensor == 'frequencies') {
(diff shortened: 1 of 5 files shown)


murrant:
FTBZ: this won't work this way.

get_device_divisor() gets called by the poller which doesn't have the "exclude" variable set.

Also, if you modify the function definition for get_device_divisor(), you need to update all calls to it.

Did you try using OID as I suggested  on IRC?

FTBZ:
Not sure to understand why you say that won't work, because I tested et re-rested and its working fine?

For the calls to get_device_divisor(), I noticed a lot of other functions that wasn't complete, this is why I dared do that.

I didn't find an easy way with the OID, if you've an idea?

Thanks for the feedback.

murrant:
You can only leave off arguments if there is  a default value for that argument :)

Did you test `./poller.php -h HOSTNAME -d -r -f -m voltage`?

I assume it works fine in the discovery, but the poller will still have the wrong divisor.

murrant:
@FTBZ I got the logic switched around.  That should do it.

FTBZ:
unknown

Don't understand how, but it's working fine!

PR #6158, opened 2017-03-10
fix: Change rfc1628 'state' (est. runtime and on battery) to runtime

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/graphs/sensor/runtime.inc.php
+++ b/html/includes/graphs/sensor/runtime.inc.php
@@ -1,8 +1,8 @@
 <?php
 $scale_min = '0';
 require 'includes/graphs/common.inc.php';
-$rrd_options .= " COMMENT:'                                 Last   Max\\n'";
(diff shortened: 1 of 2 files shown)


FTBZ:
Thanks for all your job @murrant. Not sure that this graph is working enough in "runtime" type. Because when you're not on battery it returning nothing.

screen shot 2017-03-10 at 07 27 06

The "seconds on battery" graph a value of 0, but the "minutes remaining" graph isn't working (Error Drawing Graph).

murrant:
Yeah, not sure what the issue is with the graphs.  I'l check it out when I get a chance.

laf:
I thought I'd commented on this last night, I updated #6153 before seeing this.

I can remove that change no problem but the runtime code should just be an snmp_get shouldn't it as it's only one OID.

murrant:
You commented on the application discovery PR, I was confused at first  :-). Let me check the MIB, unless you already did...

laf:
Was pushing my lack of sleep last night :/

image

PR #6165, opened 2017-03-10
Send default mail when no email destinations found

towster:
When no email destinations are found it simply would log that there were no email destinations.
This change will cause it to send an email to the default mail address if there are no email destinations set.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


towster:
It does change the "To:" to use the pretty name of "Default". Like this:
`To: Default <librenms@work.com>`
Is that sufficient?

laf:
This is a major change to the way alerts work - this should be it's own toggle switch "Fall back to default contact if no other contacts can be found"

laf:
Thanks @towster - It needs also to be in the webui:

`html/pages/settings/alerting.inc.php`

+ a config entry adding to db, see:

`sql-schema/166.sql` as an example

murrant:
trigger-ci

murrant:
trigger ci

laf:
I've just updated Default to NOC to match other areas - tested though and works fine.

PR #6166, opened 2017-03-10
Added new alert rules to collection

towster:
Added two new alert rules to the collection

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/misc/alert_rules.json
+++ b/misc/alert_rules.json
@@ -33,7 +33,7 @@
   },
   {
     "rule": "%ports.ifOutErrors_rate >= \"100\" || %ports.ifInErrors_rate >= \"100\"",
-    "name": "Interface Errors Rate"
+    "name": "Interface Errors Rate greater than 100"
   },
   {
     "rule": "%eventlog.type = \"discovery\" && %eventlog.message ~ \"@autodiscovered@\" && %eventlog.datetime >= %macros.past_60m",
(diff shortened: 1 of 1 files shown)


misc/alert_rules.json
```diff
@@ -30,5 +30,13 @@
   {
     "rule": "%ports.ifOperStatus = \"down\" && %ports.ifOperStatus_prev = \"up\" && %macros.device_up = \"1\"",
     "name": "Port status change from up to down"
+  },
+  {
+    "rule": "%ports.ifOutErrors_rate >= \"100\" || %ports.ifInErrors_rate >= \"100\"",
```
laf: Isn't this too specific, 100 error rate may not suite everyone.

misc/alert_rules.json
```diff
@@ -30,5 +30,13 @@
...
+    "name": "Interface Errors Rate"
+  },
+  {
+    "rule": "%eventlog.type = \"discovery\" && %eventlog.message ~ \"autodiscovered\" && %eventlog.datetime >= %macros.past_60m",
```
laf: Should be: `\"@autodiscovered@\"`

The name should also be Device discovered within the last 60 minutes.

misc/alert_rules.json
towster: It is true it may not but then I see these as more of examples of rules rather than absolutes. Before I had that rule I had no idea how to determine which ports were getting excessive errors.

laf: Ok no worries. The only other issue with this is that || don't work well right now so it needs to be two rules or done as a macro

PR #6167, opened 2017-03-10
feature: Add APC NetBotz State Sensor Support

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/sensors/states/netbotz.inc.php
+++ b/includes/discovery/sensors/states/netbotz.inc.php
@@ -1,0 +1,51 @@
+<?php
+/*
+ * LibreNMS
+ *
+ * Copyright (c) 2017 Søren Friis Rosiak <sorenrosiak@gmail.com> 
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+$temp = snmpwalk_cache_multi_oid($device, 'otherStateSensorTable', array(), 'NETBOTZ410-MIB');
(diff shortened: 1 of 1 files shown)


includes/discovery/sensors/states/netbotz.inc.php
```diff
@@ -0,0 +1,51 @@
...
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+$temp = snmpwalk_cache_multi_oid($device, 'otherStateSensorTable', array(), 'NETBOTZ410-MIB');
+$cur_oid = '.1.3.6.1.4.1.5528.100.4.2.10.1.3.';
+
+if (is_array($temp)) {
+    //Create State Index
+    $state_name = 'otherStateSensorErrorStatus';
```
murrant: This state name seems odd, is it meaningful?

laf:
I've been thinking about state support a little, the state translations are basically static but if we do need to update them then it means schema files to do so.

Would it make more sense to add them into the os yaml file?

murrant:
@laf I was thinking of converting it to something similar to the RrdDefinition class.

laf:
I'm fine with either way, was just hoping for something a bit tidier in the sensors code. The state translations are static at the end of the day so we probably don't need them in the DB.

murrant:
laf ah yes, I see what you mean.

Rosiak:
Interesting thought laf, I've been wanting to refactor the code for a while makes perfectly sense to me.

includes/discovery/sensors/states/netbotz.inc.php
laf: It's not used anywhere and is referenced to the mib so seems appropriate if a little obscure.

laf:
Actually, sod moving to yaml, definitely would be worse doing that.

PR #6173, opened 2017-03-11
Add missing datasources for postfix app

florianbeer:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Add missing DS for postfix app. Make sure to also update agent script and possibly (?) re-create rrd file.

### DIFF
--- a/html/includes/graphs/application/postfix_rejects.inc.php
+++ b/html/includes/graphs/application/postfix_rejects.inc.php
@@ -84,6 +84,12 @@ if (is_file($rrd_filename)) {
         ),
         array(
             'filename' => $rrd_filename,
+            'descr'    => 'Msg ex limit',
+            'ds'       => 'msefl',
+            'colour'   => '0053EE'
+        ),
+        array(
+            'filename' => $rrd_filename,
             'descr'    => 'Config Error',
             'ds'       => 'sce',
             'colour'   => '005300'
--- a/includes/polling/applications/postfix.inc.php
+++ b/includes/polling/applications/postfix.inc.php
@@ -16,7 +16,7 @@ list($incomingq, $activeq, $deferredq, $holdq) = explode("\n", $mailq);
 
 list($received, $delivered, $forwarded, $deferred, $bounced, $rejected, $rejectw, $held, $discarded, $bytesr,
      $bytesd, $senders, $sendinghd, $recipients, $recipienthd, $deferralcr, $deferralhid, $chr, $hcrnfqh, $sardnf,
-     $sarnobu, $bu, $raruu, $hcrin, $sarnfqa, $rardnf, $rarnfqa, $iuscp, $sce, $scp, $urr) = explode("\n", $detail);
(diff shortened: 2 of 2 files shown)


laf:
You can't just add an extra DS, it will break existing installs and only work for newly created files.

florianbeer:
I am aware, but since the app is broken at the moment anyway I just went for it on my system. The other possibility would be to exclude this statistic from the whole app.

VVelox:
This ain't as bad as I thought it was a first. Apparently I missed this in the SNMP extend script as well. This means there is no off by one error as shown in your poller patch. The new variable should be moved to the end of the line. Easier to just append it than move multiple variables.

florianbeer:
@VVelox: How would you like propose to proceed? Should we get rid of the DS entirely (also in the snmp script)? I don' think appending it to the end will do us anything good, the old RRDs will still not be usable anymore.

Also: did you read about my proposal to shift from GAUGE to COUNTER RRD's and dramatically simplify the snmp script? In the current form, the script does a lot of the work that RRD could do much better by itself. RRD was built to deal with increasing counters after all.

VVelox:
Below will fix it. The agent fix will also fix $iuscp not being tracked as well.

https://github.com/librenms/librenms/pull/6180
https://github.com/librenms/librenms-agent/pull/106

As to moving from gauge to counter or derive for mail, the problem is it results in the measurement being a per second value in the end.

florianbeer:
Erm, I already submitted a PR for the agent that fixes a few other issues as well: librenms/librenms-agent/pull/105. You missed a few instances where there can be more than one output line for a metric. Without accommodating for those, you won't get correct calculations, resulting in the values staying the same on every poll.

So if I read #6180 correctly it just adds an additional rrd file for the missing metric? That should certainly be a good way to move forward without breaking existing installs.

Are you sure COUNTER wouldn't work? If you just parse pflogsumm (or probably better the logfile itself like [mailgraph](http://mailgraph.schweikert.ch) does) and read it on every poll, you'd just get the increased counts. That's exactly how RRD works. I don't understand how this would result in "per second values", could you explain it to me?

VVelox:
Hmm.... just noticed a few other differences in yours. Will have to look those over. It should be added to the end. That way it is safer. If some one is late updating their extend, it means there won't be an off by one issue.

I know DERIVE will work, it just will result in the graph being rendered in per second though, which for mail is a bit annoying.

COUNTER will be buggy though. It tends to misbehave a lot when it comes to using it for stuff like this when log file rotation happens, resulting in massive false spikes some times when it happens.

VVelox:
Looked over the rest of your other PR and merged those into mine in https://github.com/librenms/librenms-agent/pull/106 .

laf:
@florianbeer If you update this so the stat is added to the end then we can get this merged.

florianbeer:
Ok, I can do that tomorrow. Today I'm in weekend mode ;)

laf:
This also needs the new variable to be moved in the agent: https://github.com/librenms/librenms-agent/pull/105/files

Otherwise it will break existing installs.

florianbeer:
I think this should do it: librenms/librenms-agent@25d73ba

Only the data output for snmp is altered form my original commit, the rest of the order can stay the same I think. Please glance over and tell me if that is correct.

PR #6180, opened 2017-03-12
properly track size based rejections($msefl) for Postfix now

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This depends on https://github.com/librenms/librenms-agent/pull/106 .


murrant:
I don't like this workaround.    Try adding the stat at the end, it will complain about too many values, but should update the first values.   Perhaps I should try again on my "rrd schema" patch...

VVelox:
@murrant where are you seeing that? The code works fine here.

The only option outside of doing this would be to rewrite it so that the old RRD needs removed.

murrant:
No, I said I don't like adding an extra rrd for one stat.  Then, I suggested an alternative.

VVelox:
close this as it is a bit of a dupe

PR #6187, opened 2017-03-13
Coriant Network Hardware Page.

xbeaudouin:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

#### Coriant TNMS support

This will add a new page that allow Coriant Hardware to be seen and get the right status.


laf:
I did mention in the other PR that it wouldn't be good to see mef-this, mef-that poller/discovery modules - this is what I mean. How comes this is separate, what is it actually giving the user?

laf:
Also the sql schema file should be all on one line.

xbeaudouin:
@laf: the SQL is fixed. About MEF, this is constructor idenpendant stuff, totaly outside Coriant. So that's why this is yet another poller/discovery stuff.

murrant:
What exactly does this bring?  Could components be used?

laf:
That doesn't explain what this gives though?

xbeaudouin:
Here some screen shot, that explain better the output.
capture d ecran 2017-03-15 a 16 01 25

Coriant TNMS is somewhat an proxy to several devices, this give us status to each devices and allow us to raize alerts :)

murrant:
Do you think any other device os will benefit from this module?

If not, it might just make sense to put this in the os polling file `includes/polling/os/coriant.inc.php`

Rosiak:
@murrant 
Wireless controllers? Not saying it directly could benefit from this MEC stuff, but the general approach could perhaps be used? List connected AP's, their state etc.

html/pages/device.inc.php
```diff
@@ -173,6 +173,14 @@
                 </li>';
         }
 
+        if (@dbFetchCell("SELECT COUNT(id) FROM tnmsneinfo WHERE device_id = '".$device['device_id']."'") > '0') {
```
murrant: Check the os before running a query.  Like this: `if ($device['os'] == 'coriant' &&` here

murrant:
Perhaps, this should either be made a little more generic or not be a module.

xbeaudouin:
I am opened to suggestions. Coriant snmp stuff is really a mess IMHO

laf:
Maybe a module called hardware?

xbeaudouin:
@laf: as you wish, I am opened to all suggestions on your side.

laf:
Let's get a consensus on it first.

murrant:
I feel like just putting this in the os polling module is correct it doesn't seem like it can be shared with other os.

xbeaudouin:
Since Coriant SNMP implementation is mostly working only on Coriant (the snmp this stuff is kinda strange BTW), I don't think we can share anything on other OS / hardware...

murrant:
So you agree, move the module polling code to `includes/polling/os/coriant.inc.php`

laf:
I personally don't as os module should be simply for that purpose - yes I know we have other cruft in now but that will need tidying up as well.

murrant:
It seems like a reasonable place for something like this.  I don't want a poller module for everything.

laf:
Neither do I but this seems like an excessive use of the os poller module.

xbeaudouin:
@laf: the Coriant TNMS SNMP is somewhat brain damage, I can provide in private the documentation of this. Since this PR is mostly the begining of the support.

guilhermefigueiredo:
Hello!

 How to do to test and add a coriant host in my librenms? I already did github-apply 6187 but apparently do not have the option yet .. remembering that we need the context in snmpv3, could they help me?

laf:
I've pushed an update to this to move the module into the os polling instead.

For me to approve this needs the devices table page moving to http://www.jquery-bootgrid.com/. We shouldn't have unpaginaged tables.

I also don't see why we'd have SqlSNMP(), looks like the function wrapper can be removed and we can just call the code.

murrant:
I also, think this should be converted to use "Components":
http://docs.librenms.org/Extensions/Component/

xbeaudouin:
@guilhermefigueiredo: you need to have access to TNMS system and enable SNMP
@laf: the "SqlNMP" function is a basis of the way Coriant store the data... eg : 
capture d ecran 2017-03-14 a 15 00 54
(Yes this is like a database)
@murrant: Will have a look about Components, thanks for the hint.

laf:
@murrant I'm not a huge fan of components, it makes alerting that much more difficult for people to understand and use.

@xbeaudouin I don't see the relevance still. All you are doing is running an snmpwalk and looping through the data to insert/delete/update.

xbeaudouin:
@laf: Started to read the Components, I am not sure this can be used with Coriant SNMP. I have several snmp tables with relational data between them. The fact is I have to add all data into a DB to make join between them to get some more data. I have some documentation from Coriant that really give me an headache. I can show you in private

guilhermefigueiredo:
@xbeaudouin ,

I have access to TNMS and snmp is already configured (version 3), but I need to add the "context" (-n option in snmpwalk) or else I can not do the collection.

root@nms:/opt/librenms# snmpwalk -v3 -a MD5 -l authPriv -u Administrator -A passWorD -x AES -X passWorD X.X.Y.Y
Error in packet.
Reason: noAccess
root@nms:/opt/librenms# 

root@nms:/opt/librenms# snmpwalk -v3 -a MD5 -n tnms -l authPriv -u Administrator -A passWorD -x AES -X passWorD X.X.Y.Y
SNMPv2-MIB::sysDescr.0 = STRING: P42022-P5173-C547-01-50VB hiT 7300 5.40 70.02.00 IC0220101160620 Copyright 2016 Coriant. All rights reserved.                   
SNMPv2-MIB::sysObjectID.0 = OID: SNMPv2-SMI::enterprises.42229.1.1.2863311536
DISMAN-EVENT-MIB::sysUpTimeInstance = Timeticks: (846062685) 97 days, 22:10:26.85
SNMPv2-MIB::sysContact.0 = STRING: eng
SNMPv2-MIB::sysName.0 = STRING: CORIANT-F
SNMPv2-MIB::sysLocation.0 = STRING: F-SITE
SNMPv2-MIB::sysServices.0 = INTEGER: 1
SNMPv2-MIB::sysORLastChange.0 = Timeticks: (0) 0:00:00.00
Error in packet.
Reason: noAccess
Failed object: SNMPv2-MIB::sysORLastChange.0

root@nms:/opt/librenms#

xbeaudouin:
@guilhermefigueiredo: hum... I don't use snmp v3 with them (mostly because our TNMS is behind a private network), did you try with v2c ? 
If it work you can ask them to fix their software ....

guilhermefigueiredo:
@xbeaudouin So I understand hiT 7300 does not work with snmp v2c, only with v3

murrant:
I think the core design for components is good, but it could use some Polish.

xbeaudouin:
@guilhermefigueiredo:  hum... Seems we don't same version of TNMS server. I have TNMS version 16.10.4.24.981 on a Linux server. What is yours ?

guilhermefigueiredo:
@xbeaudouin my version is 16 on a windows 7, but is the TNMS-NCT I think this version is some limited. We are trying to see with the coriant how to release access using snmp v2c.

laf:
Any update on this?

Imho, this needs the tables moving to bootgrid and the code moving out of the function. I know you say it will change later but from everything you said I'm just not sure how practical some of the additional code will be.

xbeaudouin:
@laf: what do you mean by bootgrid ? (I wasn't sure this is for me).

laf:
Basically we want to have paginated tables so that we don't end up with 100's of items on one page. It's not that difficult to do.

See:

`html/pages/peering/ix-list.inc.php`
`html/includes/table/ix-list.inc.php`

If you get stuck I can probably do it for you.

xbeaudouin:
@laf: I started to convert it. Mostly working, but I have issue with formating with nice colors the labels like in the screen shot. Since I am not a guru of javascript... :p

xbeaudouin:
@laf: just finished the conversion into bootgrid :p

laf:
image

PR #6189, opened 2017-03-13
fix: Alter the database to set the proper character set and collation

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/sql-schema/update.php
+++ b/includes/sql-schema/update.php
@@ -59,6 +59,9 @@ if ($tmp[0] <= $db_rev) {
     return;
 }
 
+// Set Database Character set and Collation
+dbQuery('ALTER DATABASE ? CHARACTER SET utf8 COLLATE utf8_unicode_ci;', array(array($config['db_name'])));
+
 $limit = 150; //magic marker far enough in the future
 foreach ($filelist as $file) {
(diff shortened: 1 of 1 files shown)


laf:
This will run every 5 minutes as it's called from discovery.php.

Is it worth moving to validate.php and not throwing a fail just fixing it?

murrant:
@laf If you look closely, it will only run if the schema is out of date.

Also, the ALTER query (assuming no change is needed) takes about the same time as the query to check what character set and collation are set to.

laf:
That's the issue when diffs only show the changes and no info is provided :)

includes/sql-schema/update.php
```diff
@@ -59,6 +59,9 @@
     return;
 }
 
+// Set Database Character set and Collation
+dbQuery('ALTER DATABASE ? CHARACTER SET utf8 COLLATE utf8_unicode_ci;', array(array($config['db_name'])));
```
laf: I'm assuming the double array is needed?

murrant: Yes, the second array removes the surrounding quotes.

Otherwise, the query is `ALTER DATABASE 'librenms' CHARACTER SET utf8 COLLATE utf8_unicode_ci;`

PR #6190, opened 2017-03-13
update pbin defaults (secure always on, no-expire and dropped hostname from name)

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Pastes are now secure by default with non-expiring links.

Also removed hostname in the from field so we don't leak info.

### DIFF
--- a/pbin.sh
+++ b/pbin.sh
@@ -29,6 +29,15 @@ PROGRAM=${0##*/}
 # paste. take input from stdin
 pastebin() {
 	# show params
+
+        if [[ "${expire}" == "" ]]; then
+            expire=0;
+        fi
+
+        if [[ "${private}" == "" ]]; then
+            private=1;
+        fi
+
 	sed -e '/^$/d' >&2 <<-EOF
 		${title+title: "$title"}
 		${name+name: "$name"}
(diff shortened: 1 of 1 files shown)


pbin.sh
```diff
@@ -100,13 +106,13 @@ set_defaults() {
 	unset reply
 
 	# default to user@hostname
-	name=${SUDO_USER:-$USER}@${HOSTNAME:-$(hostname)}
+	name=${SUDO_USER:-$USER}
```
murrant: I would say this is fine as is.  If the user doesn't want username@hostname, they can just specify name.

laf: But people won't know to do that and what does it matter if we have the full thing?

murrant: Ok, I thought it would just be more informative that root and librenms for everything :)  Then you could tell if multiple pastes were made by he same user.

laf: You won't see them in the list of recent or trending though as secure pastes don't show there.

So you'd need the link in the first (unless you fish from the DB) so at that point you'd know who they were from hopefully. If it's a public paste then I think we shouldn't put this info (security concern?) up but that's just imho and I'm not overly precious about it.

Issue #6150, opened 2017-03-09
Oxidized model/os for DNOS v6 should be powerconnect. v8/9 is DNOS.


BlackDex:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Component | Version
--------- | -------
LibreNMS  | b4c93cf9e699511196835f739d46f342e76f8d15
DB Schema | 173
PHP       | 7.0.16-3+deb.sury.org~xenial+1
MySQL     | 10.0.29-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3

As discussed on IRC the Dell DNOS v6 models do not work with the DNOS model, they work with the powerconnect model.

Here the IRC conversation.
https://gist.github.com/BlackDex/49bdc45eb3a57c6dbb72f8193193af0d

PR #6194, opened 2017-03-14
refactor: Split out dnos to dnos-6 and dnos-v9 OS #6150

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6150 

Reason for this change is that dnos has different versions in this range. For us it doesn't matter too much but oxidized needs to handle them differently so you have to map the os name dnosv6 -> powerconnect and dnosv9 -> dnos.

Will need to send out a notification of change so don't merge, just need confirmation this is ok to put a date in for merging.


murrant:
Why not just detect dnos6 as power connector?

laf:
Because it's not powerconnect, that just so happens to be what you can use to backup using Oxidized.

laf:
Ac

PR #6206, opened 2017-03-15
refactor: Move some DNOS detection to PowerConnect #6150

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Will need to send out a notification of change so don't merge, just need confirmation this is ok to put a date in for merging.

Moves dnos v6 (powerconnect based) OS to powerconnect as they should be.

Fixes: #6150 

Will be merged on the 24th.

This will change dnos v6 to powerconnect which is in line with the actual OS running.


murrant:
Looks good any consideration for making the display name something like `Powerconnect / DNOS6`?

Issue #6161, opened 2017-03-10
Added detection for Hirschmann Railswitch


mail4git38:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.

Component | Version
--------- | -------
LibreNMS  | 1d80a0d2a689e2f91ed72d614313aabc992529f2
DB Schema | 173
PHP       | 7.0.15
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2

- [X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).

discovery : http://pastebin.com/raw/EE9x6Mhj
poller : http://pastebin.com/raw/KKPyCLWN
snmpbulkwalk : https://gist.githubusercontent.com/mail4git38/07dee7f178740860d757a8a99b40e9e1/raw/84bad476937390d2de9315fd20eb8594809ec321/gistfile1.txt

- [X] Please provide as much detail as possible.
Tests are done with Hirschmann Railswitch RS30

- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

laf:
Are these the same devices

What OS do they run?

mail4git38:
RS30 is a switch and EAGLE is a firewall. There are both products from Hirschmann (Industrial network devices manufacturer).

They are also both running a dedicated OS, Hirschmann Railswitch OS and Hirschmann EAGLE OS.

laf:
Have you included the debug info from both devices, I only see one?

mail4git38:
You're right. At this time, I've just published the RS30's discovery. 

I have to connect an EAGLE on my network to be able to get the discovery, poller and snmpwalk files.

laf:
[...]

PR #6207, opened 2017-03-16
newdevice: Added detection for Hirschmann Railswitch #6161

mail4git38:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/discovery/os/hirschmann.inc.php
```diff
@@ -0,0 +1,7 @@
+<?php
```
murrant: Remove this file, it is handled by the yaml.

includes/definitions/hirschmann.yaml
```diff
@@ -0,0 +1,12 @@
+os: hirschmann
+text: 'Hirschmann Railswitch'
+type: network
+icon: hirschmann
+over:
+    - { graph: device_bits, text: 'Device Traffic' }
+    - { graph: device_processor, text: 'CPU Usage' }
+    - { graph: device_mempool, text: 'Memory Usage' }
+discovery:
+    - sysDescr:
```
murrant: Please use sysObjectId instead of sysDescr.

mail4git38:
Sorry, I'm a new user with git :-(

The requested changes are done on my dev server but I can't push my commits to issue-6161.

Here is the message I have : 'updates were rejected because the tip of your current branch is behind its remote counterpart'

How can I push my changes ?

includes/polling/os/hirschmann.inc.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+$hardware = trim(snmp_get($device, "hmPNIOOrderID.0", "-OQv", "HMPRIV-MGMT-SNMP-MIB"), '"');
```
mail4git38: trim() is indeed not needed here and in the lines below. I remove it.

includes/discovery/os/hirschmann.inc.php
mail4git38: This file is an old file from a previous LibreNMS installation. I've removed it.

includes/definitions/hirschmann.yaml
mail4git38: sysDescr is now replaced by sysObjectId.

murrant:
@mail4git38 sorry for the late response.  Run `git pull` first to pull down remote changes.

Sorry about that, I was fixing merge conflict.

mail4git38:
Hello @murrant. Thanks for your response and no problem with the delay.

I've tried to run `git pull upstream master` command before `git push origin issue-6161` but I still have the same issue (rejected updates).

laf:
You need to run `git pull` or if that doesn't work `git pull origin issue-6161` (i've assumed your remote is called origin - replace that name if it's not)

mail4git38:
Thanks a lot for your help on git for my first contribution. 

I successfully pushed my new commits to the issue 6161.

laf:
Only two things from me (aside from the merge conflict).

You have two files for state, one in `includes/discovery/sensors/states/` which is the old location and needs to be deleted.

The other is you've got the same `$index` value in all of the state support so this shouldn't be working. You would be better off updating these to use `name.$index` ie.: `hmPowerSupplyStatus.0`, `hmPowerSupplyStatus.1`, etc.

laf:
Just the issue with state indexes now and then good for a merge.

laf:
@mail4git38 Thanks for submitting this. Unfortunately you've removed yourself from the AUTHORS.md at the end of this PR. Please submit a new pull request adding yourself back in and agreeing to the contributors agreement otherwise we will have to revert this pull request.

PR #6208, opened 2017-03-16
fix: yaml files for edgeos and edgeswitch

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

1) discovery based only on sysdescr is not relevant in case of edgeos because that value can be changed via cli: set service snmp description command so i've added the sysObjectID match (which is already added in the tests files).

2) edgeswitch yaml file was malformed

### DIFF
--- a/includes/definitions/edgeos.yaml
+++ b/includes/definitions/edgeos.yaml
@@ -7,6 +7,7 @@ over:
     - { graph: device_processor, text: 'Processor Usage' }
     - { graph: device_ucd_memory, text: 'Memory Usage' }
 discovery:
+    - sysObjectId: .1.3.6.1.4.1.41112.1.5
     - sysDescr_regex:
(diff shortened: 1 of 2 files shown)


murrant:
What are you trying to fix here?
Both these devices share the same sysObjectID which is why they are problematic.

murrant:
What are you trying to fix here?

Both of these have generic sysObjectIDs so using sysObjectID alone is problematic.

crcro:
on edgeswitch is just alignment issue.
on edgeos i've added sysObjectID.

murrant:
Do we need a new test to validate these changes?

crcro:
no, the sysobjectid is already added in the edgeos.snmprec

murrant:
Looks good.  Thanks!

PR #6221, opened 2017-03-19
use components for fail2ban jails list

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

No longer using glob on the RRD directory to get a list of jails for fail2ban.


includes/polling/applications/fail2ban.inc.php
```diff
@@ -44,3 +49,43 @@
 
     $int++;
 }
+
+//
+// component processing for fail2ban
+//
+$device_id=$device['device_id'];
```
VVelox: I am a bit lost as to what that will accomplish. That just returns an array for jail names.

Or do you mean modifying it and adding a second function that returns the fail2ban component? That could also be use used there as well.

On another note, what is the proper way to bail out of a poller? The problem I found with this I found that if a host goes unresponsive in the middle of polling, it will update the component to say it has no jails.

murrant: It looks like this code is almost the same as get_fail2ban_jails(), I assumed it could reduce code duplication.

You have to add a check.  Fetch all the snmp data, validate it, then update the components.

Issue #6134, opened 2017-03-08
Notifications by email to admins isn't working when using Active Directory authentication


FTBZ:
Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.

Component | Version
--------- | -------
LibreNMS  | 1c48fe399f78b7acd93384646ef159c0d00d0ac1
DB Schema | 173
PHP       | 7.0.11
MySQL     | 5.5.50-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2

When trying to send mails notifications to admins users that are using Active Directory authentication, I noticed that was not working. Based on the @laf debugging by IRC it's seems to be not possible.

<img width="1018" alt="unknown" src="https://cloud.githubusercontent.com/assets/1927109/23708175/568d6276-0415-11e7-97e7-20609c822f24.png">

Debug code:
```
$init_modules = ['alerts'];
include '/opt/librenms/includes/init.php';

print_r(get_userlist());
```

Using this code returns an empty array even if the alerts capture debug is returning the correct informations.

FTBZ:
Not sure that's related to AD, because sysContact isn't working too...

laf:
It is related to AD, we can't auth to get user emails for AD so that will absolutely not work right now.

PR #6255, opened 2017-03-23
feature: bind user for active_directory auth

murrant:
Optional, allows the use of "remember me", API, and user alerting.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixes: #6134

### DIFF
--- a/doc/Extensions/Authentication.md
+++ b/doc/Extensions/Authentication.md
@@ -41,7 +41,14 @@ When using SELinux on the LibreNMS server, you need to allow Apache (httpd) to c
 setsebool -P httpd_can_connect_ldap=1
 ```
 
-#### MySQL Authentication
+#### Testing authentication
+You can test authentication with this script:
+```shell
+./scripts/auth_test
+```
+Enable debug output to troubleshoot issues
+
+### MySQL Authentication
 
 Config option: `mysql`
 
@@ -54,7 +61,7 @@ $config['db_pass'] = "DBPASS";
 $config['db_name'] = "DBNAME";
 ```
 
-#### HTTP Authentication
+### HTTP Authentication
 
 Config option: `http-auth`
 
@@ -69,7 +76,7 @@ $config['http_auth_guest'] = "guest";
 ```
 This will then assign the userlevel for guest to all authenticated users.
 
-#### LDAP Authentication
+### LDAP Authentication
 
 Config option: `ldap`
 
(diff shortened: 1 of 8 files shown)


FTBZ:
I don't know if it's possible, but when the bind user isn't working. It's showing a floating "Invalid credentials" on all pages and allow login of standard AD account (but without access to anything).

A better message should be useful I think.

screen shot 2017-03-24 at 07 38 14

FTBZ:
Noticed just a second annoying thing, the base DN cannot be specified for the bind user? Because our standard users and our technical users like a "bind user" isn't on the same Active Directory OU.

murrant:
Thanks for testing FTBZ, I'll see what I can do.  I think I need an additional check in reauthenticate to make sure the bind user bind was successful.

I'm not sure about the bind user DN,  one way to fix that is to set your base DN higher up the tree.

Maybe, it would make sense to support full DN for the bind user.

murrant:
Actually the bind user doen't use base dn, it uses `${config['auth_ad_binduser']}@${config['auth_ad_domain']}`

FTBZ:
Ref. to #6134 for tracking.

doc/Extensions/Authentication.md
```diff
@@ -41,7 +41,14 @@ When using SELinux on the LibreNMS server, you need to allow Apache (httpd) to c
 setsebool -P httpd_can_connect_ldap=1
 ```
 
-#### MySQL Authentication
+#### Testing authentication
+You can test authentication with this script:
+```shell
+./scripts/auth_test
```
laf: scripts/auth_test.php?

laf:
@FTBZ Would you  mind retesting this PR once @murrant is done?

murrant:
All,  done I think I gave it a pretty good round of testing myself.  I would like an additional confirmation from @FTBZ 

I also learned that apparently Firefox does some sort of session caching across browser closes...

FTBZ:
- The floating text message has disappeared and was replaced by the red box on the login screen. It's better, but it's strange to get this kind of message before trying to authenticate. The documentation needs to notified about this if it's not possible to replace the message with the exact error.

screen shot 2017-03-28 at 07 55 15

- I confirm that having two different OU for users and bind user has no impact
- Notification email is working fine, now you can use "Issue alerts to admins" with AD account
- API Access token can be created for AD user now, wasn't possible before without editing the DB

I've a Dashboard problem linked to AD account probably, but the patch returns me a lot of errors when trying to be applied to my production system.

murrant:
Thanks for the update these are all existing issues.

I just made a PR for the dashboard and twofactor bugs. #6286.

The error message can possibly be avoided... might look in to it sometime in a separate PR. It will give verbose output if it is enabled.

Issue #6147, opened 2017-03-09
Globe location issue


thebigsmileXD:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/) `No, see #6146`
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
```
Component | Version
--------- | -------
LibreNMS  | 9c780a46cb95db123991cb9e15c049531289156f
DB Schema | 173
PHP       | 7.0.16-1~dotdeb+8.1
MySQL     | 10.0.29-MariaDB-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1
```

- [ ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Hi. We experience a slight issue with the globe map in the dashboard.
The location is at a wrong position, i hope its not just that our location string is set wrong.
The world map shows it at the correct position though.
![image](https://cloud.githubusercontent.com/assets/8733998/23742626/e36489f2-04ad-11e7-9d9c-ed2a390c7dc3.png)
The location string i've entered: `Serverroom, [53.214628, 12.216470]`

laf:
Fix submitted, please test.

laf:
https://developers.google.com/chart/interactive/docs/gallery/geochart

The geo charts don't support locations and only areas, please use the worldmap.

PR #6258, opened 2017-03-23
webui: Added support for lat/lng to Globe widget #6147

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6147

### DIFF
--- a/html/includes/common/globe.inc.php
+++ b/html/includes/common/globe.inc.php
@@ -37,9 +37,31 @@ $temp_output .= "
         data.addRows([
 ";
 
-$locations = array();
-foreach (getlocations() as $location) {
-    $location = mres($location);
+if ($_SESSION['userlevel'] >= '5') {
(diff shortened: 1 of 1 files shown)


murrant:
Seems to be all nonsense for me :(   Also, really slow to populate.  Changes the names of things that are just towns to GPS.  Maybe I can take a closer look sometime but I can't really evaluate what is going on here.

Before:
before

After:
screenshot_20170325_143155

laf:
Thanks for checking, I'm not quite sure if you mean that the change is showing issues now of if that's how it was before and now.

This fix was purely to place the markers in the correct place, I'm guessing from your comment about Chicago that's not the case :)

murrant:
I think before was more correct...  They are both quite incorrect.   Some locations are lat,long, others are just town, state.

laf:
Unless I'm reading this wrong, it doesn't seem that geochart doesn't support actual locations and just areas: https://developers.google.com/chart/interactive/docs/gallery/geochart

murrant:
Makes sense, seems to explain the behavior.

Issue #6213, opened 2017-03-17
Disabled Devices | Still show in Alerts Widget


dneto82:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

==========================================================
Component | Version
--------- | -------
LibreNMS  | 559a22157f7197576c52542665486585344d78b5
DB Schema | 177
PHP       | 7.0.16
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.6.0
SNMP      | NET-SNMP 5.7.2
==========================================================
Disabled device correctly isn't showed in Summary Widget, WorldMap but still present in Alerts Widget:

![image](https://cloud.githubusercontent.com/assets/1860114/24024163/eae9193a-0a8f-11e7-9d7d-af82ed3e8983.png)

![image](https://cloud.githubusercontent.com/assets/1860114/24024176/0bf177d0-0a90-11e7-8cf0-4b60ff2c2676.png)

Maybe to just make sure to show only enabled devices in Alert widget.

laf:
You need to post your rule as the showing of disabled devices in alerts is down to that and nothing else

dneto82:
Hi @laf . Thanks.

This rule is a default / predefined rule from install.

![image](https://cloud.githubusercontent.com/assets/1860114/24055378/864ecdbe-0b1e-11e7-8f8c-0aaa330846d5.png)

On widget, I'm using these filters:

![image](https://cloud.githubusercontent.com/assets/1860114/24055401/9e5dc2f2-0b1e-11e7-84b5-d491a2fb55d9.png)

[...]

PR #6263, opened 2017-03-23
webui: Do not show disabled devices in alerts list as they stale #6213

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6213 

I've not provided an option to show hidden alerts, we might get asked for that but it adds unnecessary complicates.

### DIFF
--- a/html/includes/table/alerts.inc.php
+++ b/html/includes/table/alerts.inc.php
@@ -2,7 +2,7 @@
 
 require_once $config['install_dir'].'/includes/device-groups.inc.php';
 
-$where = 1;
+$where = ' `devices`.`disabled` = 0';
 
 $alert_states = array(
     // divined from librenms/alerts.php


murrant:
Alerts are still visible at the device level, or this affects both?

laf:
Both, I can add it back to the device page itself but the initial issue would still be valid for that page.

murrant:
I would think it would be expected to see all device alerts on the device page even when it is disabled.

ekoyle:
The problem is that the device is not polled and the alert isn't updated
when disabled, so the condition may have cleared in the meantime.  Maybe it
would be better to clear all alerts when a device gets disabled?

laf:
I'm actually of the opinion we shouldn't show these on the device page as they aren't alerts. The alternative is to change the status box for the alert to 'disabled'

murrant:
Sounds good to me.  Just wanted to discus it a bit.

PR #6271, opened 2017-03-26
Update Installation-CentOS-7-Apache.md

erocsid:
The -p option is not needed by the default installation of mariadb-server on CentOS Linux release 7.3.1611 (Core)

Including it without knowing to press enter for no password results in an authentication error:

root@netmon ~# mysql -u root -p
Enter password:
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)

Alternatively, it could be noted to leave the password field blank when asked.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
This also needs updating in the nginx doc as well if you can update that please.

Issue #5197, opened 2016-12-19
SNMP applications not working


Compizfox:
After not getting any response in my [forum thread](https://community.librenms.org/t/trying-to-get-snmp-applications-working/122/1) I guess I'll open a bug report.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.

Yes

- [x] Please include all of the information between the `====================================` section of `./validate.php`.

Component | Version
--------- | -------
LibreNMS  | ab29261f0c7c0578fd50ab491056b7f7db025671
DB Schema | 152
PHP       | 7.0.13-2
MySQL     | 10.0.28-MariaDB-2
RRDTool   | 1.6.0
SNMP      | NET-SNMP 5.7.3

- [x] Please provide as much detail as possible.

I'm trying to get applications through SNMP extend working. I've added the extend lines in my snmpd.conf, and snmpwalk shows it's working, I think:

	# snmpwalk -v2c -c public localhost NET-SNMP-EXTEND-MIB::nsExtendOutput2Table
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".1 = STRING: 8388608
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".2 = STRING: 8616368245
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".3 = STRING: 8616368245
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".4 = STRING: 202271
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".5 = STRING: 1037
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".6 = STRING: 0
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".7 = STRING: 332
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".8 = STRING: 3459
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".9 = STRING: 35699
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".10 = STRING: 10707
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".11 = STRING: 14649
	NET-SNMP-EXTEND-MIB::nsExtendOutLine."mysql".12 = STRING: 15081
[...]

PR #6277, opened 2017-03-26
fix: Update applications poller to use numeric oid instead of nsExtendOutputFull

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #5197 

Seen a few issues now where nsExtendOutputFull doesn't work but the OID does. I don't know why so switched the applications around to use numerical oid.


murrant:
@laf Not sure about this.
nsExtendOutLine.* = each line at a different oid
nsExtendOutputFull = all lines at one OID

I think you glazed over that a bit in these changes.

If you change all of them to nsExtendOutputFull  ".1.3.6.1.4.1.8072.1.3.2.3.1.2", then it should just be snmp_get(), not snmp_walk();

laf:
Sorry I'm not sure what you mean by the comment in the code block.

As for _get instead of _walk, this change shouldn't alter us looking up a table, the lookup is the same as before but we aren't using the named oid.

murrant:
I'm saying there is confusion of nsExtendOutputFull and nsExtendOutLine all over the applications.

You've replaced them all with nsExtendOutputFull (numeric).   Which is a single OID so all of the SNMP calls should just be snmp_get.

laf:
Rebased

murrant:
Odd that this is needed.  But no issues with it.

PR #6279, opened 2017-03-27
fix: APC/MGE Galaxy series UPS graphs

FTBZ:

#### Testers

--------------------

This device uses standard RFC1628 and its own OIDs. I added the new graphs using the RFC1628 and kept the old graphs with the standard OIDs to broke nothing. 

⚠️ FYI, I tested using the RFC1628 default and it's working fine, but it will lose the archived graphs.

- Add more overview graphs
- Add charge sensor (based on PowerWalk code)
- Add load sensor (based on RFC1628 code)
- Add power sensor (based on RFC1628 code, only input because my UPS returns wrong information for other values like output and bypass, so don't want to add bad code)
- Add runtime sensor (based on RFC1628 code)
- Add state sensor (based on RFC1628 code)
- Remove non working temperature sensor
- Add temperature sensor (based on RFC1628 code)

screen shot 2017-03-27 at 09 36 32
screen shot 2017-03-27 at 09 36 25

screen shot 2017-03-27 at 09 36 18
screen shot 2017-03-27 at 09 36 14


murrant:
Is most of this code copy/paste?  Can you set it as RFC1628 and override just the ones needed?  Or maybe just include the RFC1628 files?

FTBZ:
Yes a lot of copy/past with adjustments, but we can set it to use RFC but it will change everything and don't kept old graphs. So I don't think that we can set the RFC only for some graphs, but in my opinion it will be better to use the RFC for all 😜

murrant:
I really need to get my rrd renaming patch up to snuff...

FTBZ:
It's not possible to see in statistics DB if it's used or not? Galaxy UPS are big beast and probably used only by a few users. Perhaps we can send a notification and changed the device to RFC1628 now that many things were fixed in?

murrant:
Perhaps a simple notification would do the trick.

  Items missing from stats does not mean they are not used ;-)

FTBZ:
Yes, but it gives an idea as to whether it's used a lot or not.

If the notification is validated, I can work on the RFC1628 support instead of the current changes. I noticed only minor adjustment with the divisors.

laf:
So the only thing new here is the temp sensors, all others are new sensors so we can just switch everything to rfc code and leave the temp sensors - it will use both specific and rfc at that point.

I'll update now.

laf:
Updated

murrant:
@FTBZ can you give this test or provide an snmpwalk as here:
http://docs.librenms.org/Support/FAQ/#what-information-do-you-need-to-add-a-new-os

FTBZ:
Here only what I see that's not working as excepted:

- **Temperature Sensor** - The Environment sensor returns nothing, it was a problem before migrating to the RFC1628. This sensor can probably be removed.
screen shot 2017-04-05 at 13 33 12

- **Voltage Sensor** - Some divisor fine tuning here and some appears twice.
screen shot 2017-04-05 at 13 34 52

- **Current Sensor** - Some appears twice but values are correct.
screen shot 2017-04-05 at 13 38 16

- **Load Sensor** - Some divisor fine tuning here only.
screen shot 2017-04-05 at 13 39 02

laf:
How sure are you that this sensor doesn't work for anyone else?

FTBZ:
This is a good question, I can't. I was sure that the OID isn't existing, but I checked and it's returning 0. So it's not working on my two Galaxy 5000 but perhaps working with other models. Note that the battery temp information is corresponding to the value of the environmental temperature sensor for me.

laf:
So ok to merge then @FTBZ ?

FTBZ:
Divisors are still wrong and in my opinion, it's better to create a notification and remove the sensors that appear twice?

laf:
I didn't see any difference to the divisors in the code you copied across though, it looked like everything was exactly the same apart from some names.

You could remove the mgeup.inc.php files from discovery sensors where we have rfc ones already and see. We then also have the ability to get custom divisors within the rfc code anyway.

We'd need a device walk though to help with the code more as we don't know what values get returned.

FTBZ:
This is because the RFC code use not the same divisor as the default code used in my files.

This is a v1 only device, what I can do to get something like the snmpbulk?

Discovery: https://pastebin.com/80FVxwtn
Poller: https://pastebin.com/Us1ALEZc

murrant:
@FTBZ http://docs.librenms.org/Support/FAQ/#faq20

murrant:
@FTBZ Simply replace snmpbulkwalk with snmpwalk

PR #6281, opened 2017-03-27
Rocketchat no message

svennd:
If some people also face this issue : the transport works in rocket.chat, but contains no message. This is the script "required" for the webhook integration : 

```
class Script {
  process_incoming_request({ request }) {
    console.log(request.content);

    return {
      content:{
        text: request.content.text
       }
    };

    return {
       error: {
         success: false,
         message: 'Error example'
       }
     };
  }
}
```
Once this change has been done, it works for me. (nothing fancy as coloring) 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -195,6 +195,7 @@ LibreNMS contributors:
 - Hubert Ladet <mail4git38@gmail.com> (mail4git38)
 - Deeps (deepseth)
 - Jari Schäfer <jari.schaefer@gmail.com> (jarischaefer)
+- Svennd <admin@svennd.be> (svennd)
 
 Observium was written by:
 - Adam Armstrong
--- a/includes/alerts/transport.rocket.php
+++ b/includes/alerts/transport.rocket.php
@@ -28,7 +28,7 @@ foreach( $opts as $tmp_api ) {
     $color = ($obj['state'] == 0 ? '#00FF00' : '#FF0000');
     /*
     $data = array(
-        "attachments' => array(
+        'attachments' => array(
             0 => array(
                 'fallback' => $rocket_msg,
                 'color' => $color,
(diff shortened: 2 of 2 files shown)


laf:
I'm not a rocket.chat user but the original looks like it should work ok, I'd prefer to fix it if needed rather than change how it works.

svennd:
@laf I searched the rocket.chat for a solution why it did not work (as described in the docs), but they all pointed towards this solution. So I kinda wanted to share it. However the current code is just a copy of slack, which might have worked before (?) but for the latest version for sure does not work.

For sure, close it, if someone ever has a similar problem you can pick it up later. :) no bad feelings.

laf:
Do you have any links to show that this is the way it should be done? Sending text= is also fine but so is sending as attachments.

svennd:
https://rocket.chat/docs/administrator-guides/integrations/

Both methods seem fine, but I did not get the attachment method working. Perhaps @ToeiRei can pitch in ? Perhaps its just documenting it better on how to use it ?

ToeiRei:
In my case, notifications look like that:

> 
> LibreNMS @rocket.cat BOT 23:56:02 
> Device 10.0.0.8 recovered from Device rebooted 
> Device 10.0.0.8 recovered from Device rebooted 
> Severity: critical 
> Time elapsed: 5m 
> Timestamp: 2017-03-28 21:55:05 
> Unique-ID: 1885 
> Rule: Device rebooted 
> Faults: 
> #1: sysObjectID => enterprises.8072.3.2.8; sysDescr => Hardware: amd64; 
> Alert sent to: NOC

That's all I need. Not sure what the problem really is.

svennd:
Ah well, then I guess its a problem on my end. Closing this :) thx all !

laf:
@svennd you can run alerts.php with -d to get more debug output, you will only get it if an alert is due to be sent which means if the cron job gets to it before you do then you won't see anything. Could be worth just disabling the alerts.php in cron before testing.

ToeiRei:
@svennd what did you do at the rocket.chat config side?

PR #6284, opened 2017-03-27
fix: add semicolons in sql scripts

jarischaefer:
Adds missing semicolons in SQL scripts.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Thanks, good to tidy these up but they aren't strictly needed unless you directly import them into mysql.

jarischaefer:
After I had encountered errors running the build-base script, I tried importing the SQLs manually via MySQL's CLI, which resulted in the errors fixed by this PR.

It seems the build-base script ignores all SQL errors and just keeps going with no visible error messages.

I will probably submit another PR shortly, which fixes some of the "real" errors that forced me to use the CLI directly in the first place.

laf:
@jarischaefer Running build base gives you an error count (we have unit tests against these - slightly broken right now). I've recently been through and fixed all the errors, please ensure you don't have mysql strict mode enabled.

jarischaefer:
I just tested everything with the official MySQL docker image. MySQL 5.7 does not work at all (they removed ALTER IGNORE TABLE in 5.7.4, which breaks 109.sql) and 5.6 still fails at https://github.com/librenms/librenms/pull/6289.

Issue #5094, opened 2016-11-30
Two Factor not working when using LDAP


jallakim:
Hi,

Two Factor authentication does not work (i.e. it's not initialized) when using LDAP. Haven't tested with other authentication sources, but as long as they don't create a database entry in the `users` table, it won't work (which is suspect is common for most of them?).

It's the [`dbUpdate` that fails](https://github.com/librenms/librenms/blob/043de43365e2feda3cc883d18f537dffcb7e1a2d/html/pages/preferences.inc.php#L150), since there are no row having a `username` value that matches `$_SESSION['username']`.

jallakim:
```
Component | Version
--------- | -------
LibreNMS  | 3390be5107955e61d74f93a94eca8787911294ac
DB Schema | 151
PHP       | 7.0.13-1~dotdeb+8.1
MySQL     | 5.7.13-6
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1
```

jallakim:
Not sure where/how this should be handled, but new users signing in via LDAP should be added as a user in the database (w/o password being stored). That goes for the rest of the authentication methods as well (AD/RADIUS/etc).

laf:
We add users to the DB for quite a few auth methods. I'd say it's more of a work around than a fix but you're free to do the same for this auth module.

laf:
No further info provided

jallakim:
@laf, what more info should be provided? There is no need for more information -- just a PR from Someone™ that fixes it (and yes, I have it on my TODO-list). Issue should be kept open until it's resolved?

Issue #3376, opened 2016-04-15
Can't edit default dashboard with active directory auth provider


wiad:
Trying to do "Edit user" while using the AD auth provider yields an error message:

`Authentication method doesn't support updating users`

This makes it impossible to edit a users default dashboard. Since this setting is stored in the db it should be editable regardless of auth driver.

wiad:
A possible solution to this could be to disable the form elements which are not permitted to update if can_update_users() does not return 1.

PR #6286, opened 2017-03-28
fix: move user preferences dashboard and twofactor out of users table

murrant:
This allows them to work with any authentication method
Add set_user_pref() and get_user_pref() helper functions

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixes: #5094, #3376


sql-schema/183.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE users_prefs DROP PRIMARY KEY;
+ALTER TABLE users_prefs DROP INDEX pref;
+INSERT INTO `users_prefs` (`user_id`, `pref`, `value`) SELECT `user_id`, 'dashboard', `dashboard` FROM `users` WHERE `dashboard` != '0';
+INSERT INTO `users_prefs` (`user_id`, `pref`, `value`) SELECT `user_id`, 'twofactor', `twofactor` FROM `users` WHERE `twofactor` != '0';
+ALTER TABLE `users` DROP `twofactor`, DROP `dashboard`;
```
laf: This doesn't run:

`ERROR 1067 (42000): Invalid default value for 'updated_at'`

That is with `ONLY_FULL_GROUP_BY,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION` set in mysql but I've been through before and resolved all of those.

`1970-01-01 00:00:01` is what the default should be so will need that updating in the schema before the final ALTER TABLE line.

This has shown me that the db unit tests are broke. When you run `DBTEST=true phpunit tests/DBSetupTest.php` it's fine but if you just run `DBTEST=true phpunit` then it fails as it can't select the database on init as it's not been created yet. I'm not sure how to fix that other than try and create the db on initi if we can't switch.

laf: Actually this patch fixes it:

```
diff --git a/includes/init.php b/includes/init.php
index 8ccad05..3a5a6d4 100644
--- a/includes/init.php
+++ b/includes/init.php
@@ -110,6 +110,10 @@ if (!module_selected('nodb', $init_modules)) {
         die;
     }
     $database_db = mysqli_select_db($database_link, $config['db_name']);
+    if (!$database_db) {
+        mysqli_query($database_link, "CREATE DATABASE " . $config['db_name'] . " CHARACTER SET utf8 COLLATE utf8_unicode_ci");
+        $database_db = mysqli_select_db($database_link, $config['db_name']);
+    }
     dbQuery("SET NAMES 'utf8'");
     dbQuery("SET CHARACTER SET 'utf8'");
     dbQuery("SET COLLATION_CONNECTION = 'utf8_unicode_ci'");

```

sql-schema/183.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE users_prefs DROP PRIMARY KEY;
+ALTER TABLE users_prefs DROP INDEX pref;
+INSERT INTO `users_prefs` (`user_id`, `pref`, `value`) SELECT `user_id`, 'dashboard', `dashboard` FROM `users` WHERE `dashboard` != '0';
+INSERT INTO `users_prefs` (`user_id`, `pref`, `value`) SELECT `user_id`, 'twofactor', `twofactor` FROM `users` WHERE `twofactor` != '0';
+ALTER TABLE `users` CHANGE `updated_at` `updated_at` TIMESTAMP NOT NULL DEFAULT '1970-01-01 00:00:01';
```
laf: Nothing's ever simple. So this still doesn't work when your TZ is not UTC :(

Seems `DEFAULT CURRENT_TIMESTAMP` works.

laf: But you can't have two of those, ffs

laf: Not really sure how best to handle, looks like I've actually done 1970-01-02 00:00:01 in sql-schema/036.sql before

murrant: Yeah two as current timestamp is how we got here in the first place.

laf: So switch to 1970-01-02? That way TZ's don't come into play.

Issue #2703, opened 2016-01-06
Port name containing HTML Special Chars like < or >


Kliwer:
Hello. If port name contains html special char like < or > it is not shown correctly - it is used as HTML tag in the code. Using htmlspecialchars() fixes this. Please correct this.

f0o:
Haha priceless.
Will take care of it asap

f0o:
turns out it's a bit more complicated due to the bad codebase we inherited ...
Got it almost settled, just one thing keeps altering the ifLabel back for some reason

Rosiak:
@f0o Able to finish the mentioned PR or?

laf:
I've added two functions clean() and display() to clean input and display output - we can use display() here to deal with this now.

Kliwer:
On version: 969b963 - Thu Dec 15 2016 09:23:25 GMT+0100 this problem still exists.

laf:
Can you actually provide some more info on this, poller debugs, output from the device?

Kliwer:
![libre](https://cloud.githubusercontent.com/assets/11368153/21266866/3088571e-c3a8-11e6-85d3-9a651ee534b8.png)
This is how it looks. "<pppoe-mlody>" is the interface name. 
I'll add debug and output later.

genabi:
This is a potential XSS problem - SNMP response can be crafted to load an unwanted JS into browser

PR #6288, opened 2017-03-28
refactor: Moved ifLabel -> cleanPort and updated the usage

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #2703 

This should resolve issues with ports that need to have information escaped / encoded before displaying.

Also fixed the notes widget to allow people to put certain html tags in.

### DIFF
--- a/includes/common.php
+++ b/includes/common.php
@@ -1508,6 +1508,7 @@ function clean($value)
 
 /**
  * @param $value
+ * @param array $purifier_config (key, value pair)
  * @return string
  */
 function display($value, $purifier_config = array())
@@ -1515,6 +1516,9 @@ function display($value, $purifier_config = array())
     /** @var HTMLPurifier $purifier */
     global $config, $purifier;
 
+    // If $purifier_config is non-empty then we don't want
+    // to convert html tags and allow these to be controlled
+    // by purifier instead.
     if (empty($purifier_config)) {
         $value = htmlentities($value);
     }
@@ -1522,7 +1526,7 @@ function display($value, $purifier_config = array())
         // initialize HTML Purifier here since this is the only user
         $p_config = HTMLPurifier_Config::createDefault();
         $p_config->set('Cache.SerializerPath', $config['temp_dir']);
-        foreach ((array)$purifier_config as $k => $v) {
+        foreach ($purifier_config as $k => $v) {
             $p_config->set($k, $v);
         }
         $purifier = new HTMLPurifier($p_config);
--- a/includes/common.php
+++ b/includes/common.php
@@ -1510,14 +1510,21 @@ function clean($value)
  * @param $value
  * @return string
  */
-function display($value)
+function display($value, $purifier_config = array())
 {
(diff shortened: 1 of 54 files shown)


includes/common.php
```diff
@@ -1510,14 +1510,21 @@ function clean($value)
...
     global $config, $purifier;
+
+    if (empty($purifier_config)) {
+        $value = htmlentities($value);
+    }
     if (!isset($purifier)) {
         // initialize HTML Purifier here since this is the only user
         $p_config = HTMLPurifier_Config::createDefault();
         $p_config->set('Cache.SerializerPath', $config['temp_dir']);
+        foreach ((array)$purifier_config as $k => $v) {
```
murrant: no need for a cast here, because you are using keys it won't work (keys will be numeric on something casted to an array.

(Update the phpdoc to indicate $purifier_config should only be an array, maybe a note that it is key->value too

includes/common.php
```diff
@@ -1510,14 +1510,21 @@ function clean($value)
  * @param $value
  * @return string
  */
-function display($value)
+function display($value, $purifier_config = array())
 {
     /** @var HTMLPurifier $purifier */
     global $config, $purifier;
+
+    if (empty($purifier_config)) {
```
murrant: Can you add a comment here, I can't figure out why this is here by looking at the code alone.

includes/rewrites.php
```diff
@@ -58,15 +58,17 @@ function rewrite_entity_descr($descr)
 
 function ifNameDescr($interface, $device = null)
 {
-    return ifLabel($interface, $device);
+    return cleanPort($interface, $device);
 }
 
 
-function ifLabel($interface, $device = null)
+function cleanPort($interface, $device = null)
```
murrant: can you add phpdoc here, mostly to signify that both variables should be arrays.

includes/common.php
laf: Done, it's so that we can pass on purifier config options at which point we don't want to have munged the html - see html/includes/common/notes.inc.php as an example.

murrant:
I started to test this and it was inconsitent.
For example, look at the ifAlias line on any port overlib link.

I'll have to think on it more.

laf:
I have looked at the ifAlias line on port overlib - what are you seeing and what's your sample ifAlias data?

murrant:
I set my sample data to `eth0&amp;<Descr>` for ifName, ifDescr, and ifAlias

laf:
Pushed an update to fix this. I'd already done it on the line above. Basically display() is already run so gives you the clean data but we then add that into the html - which at that stage looks fine, but as this is an overlib popup the html data is reparsed by the browser and &amp; becomes &. Double display() basically fixes this.

murrant:
@laf There is an issue with ajax tables getting double encoded.
`eth0&amp;amp;amp;&amp;lt;Alias&amp;gt;`

We could remove the cleanPort(), but I don't think that is correct.
Perhaps a way to skip htmlentities when calling cleanPort()?

PR #6297, opened 2017-03-29
Change Cisco UCM category from tele to telephony

aldemira:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I think tele doesn't mean much.


laf:
Thanks for this, to change a type you will need to also submit a notification that this is changing: https://github.com/librenms/librenms.github.io/pull/83

However it will also need more work doing, tele at the moment must be broke but it also needs at least something in includes/definitions.inc.php:

```
$config['device_types'][$i]['text'] = 'Printers';
$config['device_types'][$i]['type'] = 'printer';
$config['device_types'][$i]['icon'] = 'printer.png';
```

As an example.

laf:
Thanks @aldemira. This just needs the notification now adding, let us know when that's been done.

aldemira:
I'm sorry I couldn't get what you mean by notifications?

adaniels21487:
I would prefer this changed to 'collaboration' to be consistent with: Acano, Expressway, VC Endpoints and TPS.

laf:
@aldemira What do you think?

@adaniels21487 If we do stick with that then it needs a group definition like this PR has, that looks to be missed when collaboration was added.

aldemira:
I'm fine with that as long as it's not "Tele". So shall I update the code then?

laf:
@aldemira Yes please, I'll do the notification once you have.

laf:
Ok, the notification has gone out. Will merge Monday 10th.

PR #6303, opened 2017-03-30
feature: Added ability to validate database schema

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This is not ready for merging yet as it needs the deploy script finishing and adding to travis.

For users, validate will compare the database schema against a known working one and then provide queries to bring the schema into line. Some things...

  - I've not yet added support for creating missing tables.
  - I've not yet added support for creating missing columns.
  - I have seen an issue with users table where an update to created_at and updated_at still doesn't work as the order needs to be particular which this doesn't cover.
  - I've not done auto fixing but this is possible, for now I've asked users to report the recommended schema changes so we can validate them until we are happy.

When a PR is merged in, travis will rebuild the schema file and if it's changed then it will commit that to the master branch automatically. The only issue with this is if someone pulls in an update in the time it takes for a merge to happen and travis to rebuild the file - that's acceptable to me.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
   - php: 5.5
     env: SKIP_STYLE_CHECK=1
   - php: 5.6
-    env: SKIP_STYLE_CHECK=1 EXECUTE_BUILD_DOCS=true
+    env: SKIP_STYLE_CHECK=1 EXECUTE_BUILD_DOCS=true EXECUTE_BUILD_SCHEMA=true
 #  - php: hhvm
 #    env: SKIP_STYLE_CHECK=1
 
@@ -39,6 +39,7 @@ before_script:
 
 after_success:
   - test $TRAVIS_PULL_REQUEST == "false" && test $TRAVIS_BRANCH == "master" && test $EXECUTE_BUILD_DOCS == "true" && bash scripts/deploy-docs.sh
+  - test $TRAVIS_PULL_REQUEST == "false" && test $TRAVIS_BRANCH == "master" && test $EXECUTE_BUILD_SCHEMA == "true" && bash scripts/deploy-schema.sh
 
 after_failure:
   - cat /tmp/snmpsimd.log
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2186,6 +2186,9 @@ function cache_peeringdb()
 function dump_db_schema()
 {
     global $config;
+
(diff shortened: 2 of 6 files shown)


murrant:
I like where this headed.

I don't mean to deflate your steam, but this would be a lot better with a DBAL/schema manager like Doctrine.

laf:
I'm aware the more permanent fix is using a dbal for this but we don't have that currently.

We do have inconsistencies within users db schemas right now and this is done to purely fix this and bridge the gap until we get something better in place. At present we assume users db's are correct so when helping to debug things we can end up on a wild goose chase until we realise it's because a schema update has been missed. This should stop that from happening for now.

It's automatic in it's approach so we don't need to do anything to maintain this at present, I expect it will generate some noise whilst we validate the output it produces but once we're happy we can just enable auto fixing.

It was hardly anytime to put it together so I'm not particular about it being merged in but I do prefer to spend my time enhancing the software rather than bug hunting :)

murrant:
I think json would be fine for this since it is system generated and consumed (no humans in the loop generally)

laf:
Why would you need the schema number? The version of the yaml/json file should match the the schema in that current commit.

laf:
With regards to json over yaml. If we use formatted json it's bulky, if we don't it looks a mess and if humans do want to read it it's difficult.

So yaml or formatted json?

murrant:
I compared the outputs:
json_encode was smaller but all on one line
_json_encode was twice the size and a little harder to read than yaml also

So yaml looks the best here. :)

I've added checks for extra tables and columns, is that ok?

Also, I noticed there is no check for the indexes.  I think we should check that too if we can.

laf:
Yeah I didn't do index checks yet as that needs more work as they need to be parsed.

murrant:
I'll go through this tonight a little more to see if I can complete the table creation bit.  If we don't get indexes tackled, I think we can merge this PR and handle that later.

murrant:
@laf I think this is in decent shape.  All that is left is hooking up the travis-ci part.

Also, why the 60 second sleep in build-schema.php?  exec calls wait until the command returns right?

laf:
@murrant I just put it in as a safe guard as locally when testing I had a couple of times where the schema change hadn't applied fully.

laf:
This is either going to go well or not - merging :)

Issue #6298, opened 2017-03-29
NFS application extension : confusing/bug


svennd:
https://github.com/librenms/librenms/blob/master/scripts/agent-local/nfs-stats.sh
states that : 
```
# edit your snmpd.conf and add the below line and restart: #
# extend nfs-stats /opt/nfs-stats.sh 

```
this is incorrect as https://github.com/librenms/librenms/blob/master/includes/polling/applications/nfs-stats.inc.php which is suppose to "accept" this data, is expecting way lower amount of data.  The correct version is : https://github.com/librenms/librenms/blob/master/includes/polling/applications/nfs-v3-stats.inc.php  However the oid given in that files does not match up with "nfs-v3-stats". It looks like there is an overlap in functionality to me. 

I will look for a solution and try to document it tomorrow. Thx for all your help (@irc)

DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

laf:
Can this be closed?

svennd:
Yes; there is currently a PR trying to fix it  : https://github.com/librenms/librenms/pull/5767

PR #6304, opened 2017-03-30
bugfix for nfs-v3 application

svennd:
bugfix for nfs-v3 application;
- in nfs-stats.sh, line 8 is refering to proc2 which are stats for
nfsv2.
- nfs-stats.inc.php is using an oid that is using the results of this nfs-stats.sh script. This is incorrect but I don't want to break stuff.
- nfs-v3-stats.inc.php could not find data using the oid provided, so I changed it.

These changes made it work for me, but I think it would be good if
@crcro could have a look.

bug : https://github.com/librenms/librenms/issues/6298

env: centos 6/7 tested

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


svennd:
most likely should have been posted here : https://github.com/librenms/librenms/pull/5767

sorry for duplicate.

svennd:
wrong oid.

PR #6306, opened 2017-03-30
newdevice: Add support for AvediaStream Encoder

FTBZ:

#### Testers

--------------

- Add support for AvediaStream Encoder
- Detected as generic Linux, can't use the sysObjectId only, because is a Linux standard
- OS name returned by SNMP is Ardmore Encoder VX


laf:
This is missing unit tests: http://docs.librenms.org/Developing/Support-New-OS/#os-test-units

+ 1 Commet I left about sysDescr.

tests/OSDiscoveryTest.php
```diff
@@ -206,6 +206,11 @@ public function testApic()
     {
         $this->checkOS('apic');
     }
+    
+    public function testArdmore-encoder()
```
laf: Change to: `public function testArdmoreEncoder()`

PR #6320, opened 2017-04-02
NFS-server

svennd:
Hey,

I rewrote the pulling of NFS stat, at current there are 3 implementations : 
1) freeBSD server polling
2) [nfs-stats](https://github.com/librenms/librenms/blob/master/includes/polling/applications/nfs-stats.inc.php) (which is very limited, and I don't know where the script is)
3) [nfs-v3-stats](https://github.com/librenms/librenms/blob/master/includes/polling/applications/nfs-v3-stats.inc.php) : which uses this [script](https://github.com/librenms/librenms/blob/master/scripts/agent-local/nfs-stats.sh) (although the documentation is wrong, the oid is pointing to nfs-v3-stats)

The problem with the last option is, it depends on a script that does not always report the correct value's.  Line 8 of /proc/net/rpc/nfsd on older systems (centos 6) reports nfsv2 stats, and on newer systems (centos 7, debian) nfsv3. These are similar but not exactly the same. Therefore I believe the collected data is faulty in allot of cases. 

I used nfs-v3-stats as template to write nfs-server, this also makes it possible to later add nfs-client. (which is also relevant data)

So my personal opinion would be to official drop the second option and leave in nfs-v3-stats for users that trust that the data they collected are correct. 

I rewrote it, but took another approach on storing the data, I let rrd do the calculation instead of doing it with a script. This means : 
`extend nfs-server /bin/cat /proc/net/rpc/nfsd`
is enough to get it working.

I do have a few remarks tho, where your advice would be helpful
- nfs reports tons of stats, many of those however are not used.  (fh, th, ... all 0) should we store those ?
- nfs reports allot of stats, would it be useful to split the rrd's ? This idea I like allot since v2 is being dropped and v4 is changing rapidly (see further) we could keep data even is there is an update in the data-structure later. It would also be possible to disable part of the application. 
- nfsv4ops has changed at least two times, I have seen systems of 40 value's, 59 value's and 71 value's I currently only found what the 59 where... Any help would be great. ( I assume the 59 are inc. in the 71 but I'm not sure on the order)
- overall advice is also welcome ( be gentle ;) )

Considering I used the code of @crcro and @InsaneSplash , I would be very happy if you could help with advice of your opinions ! 

// addition
note : after talking to laf, I decided that splitting up would be best as currently some value's can be empty. (v2 in new systems, v4ops in old/new systems) which rrd can't handle.

#### Please note

#### Testers


murrant:
@svennd You have a decent amount of style fixes needed here.
You can check with `./scripts/pre-commit.php`

laf:
So I guess to answer your questions:

- ***nfs reports tons of stats, many of those however are not used.  (fh, th, ... all 0) should we store those ?***

If they are truly never used then drop them.

- ***nfs reports allot of stats, would it be useful to split the rrd's ? This idea I like allot since v2 is being dropped and v4 is changing rapidly (see further) we could keep data even is there is an update in the data-structure later. It would also be possible to disable part of the application. ***

I'd be happy for the split.

- ***nfsv4ops has changed at least two times, I have seen systems of 40 value's, 59 value's and 71 value's I currently only found what the 59 where... Any help would be great. ( I assume the 59 are inc. in the 71 but I'm not sure on the order)***

This seems if we don't know how much data we are getting back. A lot of the time this is why agent/snmp scripts are written.

svennd:
@laf thanks for your answers! Well the file tells you how much value's you are to receive : its the first value of the line. The issue is that I can't find what the value's mean. Its a bit more work to find out... if someone knew it would be useful

laf:
rrd is a fixed format though so you need to know how many entries you are going to insert - it has to be predictable really.

svennd:
please see https://community.librenms.org/t/nfs-server/975

Code is not working.

includes/polling/applications/nfs-server.inc.php
```diff
@@ -0,0 +1,278 @@
...
+                        
+// parse each output line, by the id
+// then 'map' the values to the arrays from $keys_nfs_server
+$lines     = explode("\n", $nfsstats);
+$default_fields = array();
+$fields = array();
+foreach ($lines as $line)
+{
+    $line_values     = explode(" ", $line);
+    $line_id         = $line_values[0];
```
murrant: array_shift() returns the value, you can merge these two lines like this:
```
$line_id = array_shift($line_values);
```

laf:
Fair few travis issues that need fixing :)

https://travis-ci.org/librenms/librenms/builds/219620508

svennd:
Yeah, but white-spaces/tabs/enters won't fix the code.

laf:
What do you mean won't fix the code? They will fix travis, are you having other issues?

svennd:
https://community.librenms.org/t/nfs-server/975

When the "normal" poller runs it generates device/A.rrd instead of the device/app-nfs-server-default.rrd. the app-nfs-server-proc[2,3,4,4ops].rrd aren't generated either. While manually running 
`./poller.php -r -f -m applications -h device -d `

Shows the correct rrd creation/update.

svennd:
Ok, the code works and I made it PSR2 valid (?).  It's running here now for a few hours and pulling data in correctly. Perhaps It needs some more documentation.

includes/polling/applications/nfs-server.inc.php
```diff
@@ -0,0 +1,269 @@
...
+// nfs server stats processor 
+
+use LibreNMS\RRD\RrdDefinition;
+
+$name = 'nfs-server';
+$oid = '.1.3.6.1.4.1.8072.1.3.2.3.1.2.10.110.102.115.45.115.101.114.118.101.114';
+
+echo ' ' . $name;
+
+$nfsstats = snmp_walk($device, $oid, '-Oqv', 'NET-SNMP-EXTEND-MIB');
```
murrant: You should use snmp_get() as this is only a single oid.

murrant:
yeah, updating the docs would be good :-)

html/pages/device/apps/nfs-server.inc.php
```diff
@@ -0,0 +1,37 @@
...
+    'nfs-server_io' => 'IO',
+    'nfs-server_fh' => 'File handler',
+    'nfs-server_rc' => 'Reply cache',
+    'nfs-server_ra' => 'Read ahead cache',
+    'nfs-server_net' => 'Network stats',
+    'nfs-server_rpc' => 'RPC Stats',
+
+);
+
+foreach ($graphs as $key => $text) {
```
murrant: You should not display headers for sections that do not have an rrd.

laf:
Just needs the contributor PR fixing and the order for the Applications.md (it should be alphabetical) and then looks good.

html/includes/graphs/application/nfs-server_fh.inc.php
```diff
@@ -0,0 +1,36 @@
...
+$colours       = 'mixed';
+$unit_text     = 'File Handle Stats';
+$unitlen       = 15;
+$bigdescrlen   = 15;
+$smalldescrlen = 15;
+$dostack       = 0;
+$printtotal    = 0;
+$addarea       = 1;
+$transparency  = 33;
+$rrd_filename  = $config['rrd_dir'].'/'.$device['hostname'].'/app-nfs-server-default-'.$app['app_id'].'.rrd';
```
laf: `$rrd_filename` needs building with `rrd_name()`

html/includes/graphs/application/nfs-server_fh.inc.php
```diff
@@ -0,0 +1,36 @@
...
+    'fh_lookup' => array('descr' => 'fh_lookup'),
+    'fh_anon' => array('descr' => 'fh_anon'),
+    'fh_ncachedir' => array('descr' => 'fh_ncachedir'),
+    'fh_ncachenondir' => array('descr' => 'fh_ncachenondir'),
+    'fh_stale' => array('descr' => 'fh_stale'),
+);
+
+$i = 0;
+
+if (is_file($rrd_filename)) {
```
laf: Please use `rrdtool_check_rrd_exists()` instead of `is_file()`

laf:
lgtm @librenms/reviewers

laf:
@librenms/reviewers I'm merging this in tomorrow unless anyone has any objections.

PR #6341, opened 2017-04-04
refactor: Update some snmpwalks for ports polling to improve speed

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I'll do some inline comments on this one but for me it drops port polling on Arista down from 160 -> 80 seconds. It also helps our Cisco switches, with about 800 devices I've dropped 20 seconds on each distributed poller.

### DIFF
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -66,23 +66,6 @@ $stat_oids_db_extended = array(
     'ifOutMulticastPkts',
 );
 
-// From above for DB
-$etherlike_oids = array(
-    'dot3StatsAlignmentErrors',
-    'dot3StatsFCSErrors',
-    'dot3StatsSingleCollisionFrames',
-    'dot3StatsMultipleCollisionFrames',
-    'dot3StatsSQETestErrors',
-    'dot3StatsDeferredTransmissions',
-    'dot3StatsLateCollisions',
-    'dot3StatsExcessiveCollisions',
-    'dot3StatsInternalMacTransmitErrors',
-    'dot3StatsCarrierSenseErrors',
-    'dot3StatsFrameTooLongs',
-    'dot3StatsInternalMacReceiveErrors',
-    'dot3StatsSymbolErrors',
-);
-
 $cisco_oids = array(
     'locIfHardType',
     'locIfInRunts',
@@ -99,6 +82,9 @@ $cisco_oids = array(
 
 $pagp_oids = array(
     'pagpOperationMode',
+);
+
+$pagp_extended_oids = array(
     'pagpPortState',
     'pagpPartnerDeviceId',
     'pagpPartnerLearnMethod',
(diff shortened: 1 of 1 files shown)


includes/polling/ports.inc.php
```diff
@@ -66,23 +66,6 @@
     'ifOutMulticastPkts',
 );
 
-// From above for DB
-$etherlike_oids = array(
```
laf: We appear to not use these at all.

includes/polling/ports.inc.php
```diff
@@ -224,11 +210,11 @@
     }
 }
 
-if ($config['enable_ports_etherlike']) {
-    echo 'dot3Stats ';
-    $port_stats = snmpwalk_cache_oid($device, 'dot3StatsEntry', $port_stats, 'EtherLike-MIB');
```
laf: We don't use any of this table.

includes/polling/ports.inc.php
```diff
@@ -224,11 +210,11 @@
...
-} elseif ($device['os'] != 'asa') {
+if ($device['os'] != 'asa') {
     echo 'dot3StatsDuplexStatus';
+    if ($config['enable_ports_poe'] || $config['enable_ports_etherlike']) {
+        $port_stats = snmpwalk_cache_oid($device, 'dot3StatsIndex', $port_stats, 'EtherLike-MIB');
```
laf: We use only this + dot3StatsDuplexStatus within poe and etherlike port polling files so only look those up.

includes/polling/ports.inc.php
```diff
@@ -265,13 +251,17 @@
...
 }
 
-// FIXME This probably needs re-enabled. We need to clear these things when they get unset, too.
-// foreach ($etherlike_oids as $oid) { $port_stats = snmpwalk_cache_oid($device, $oid, $port_stats, "EtherLike-MIB"); }
-// foreach ($cisco_oids as $oid)     { $port_stats = snmpwalk_cache_oid($device, $oid, $port_stats, "OLD-CISCO-INTERFACES-MIB"); }
-// foreach ($pagp_oids as $oid)      { $port_stats = snmpwalk_cache_oid($device, $oid, $port_stats, "CISCO-PAGP-MIB"); }
 if ($device['os_group'] == 'cisco' && $device['os'] != 'asa') {
     foreach ($pagp_oids as $oid) {
-        $port_stats = snmpwalk_cache_oid($device, $oid, $port_stats, 'CISCO-PAGP-MIB');
+        $pagp_port_stats = snmpwalk_cache_oid($device, $oid, array(), 'CISCO-PAGP-MIB');
```
laf: Later on we only use the check for pagpOperationMode so we poll that first, check it has data and if so apply it to $port_stats and then poll the rest of pagp data.

includes/polling/ports.inc.php
```diff
@@ -280,7 +270,7 @@
     $port_stats = snmpwalk_cache_oid($device, 'vlanTrunkPortEncapsulationOperType', $port_stats, 'CISCO-VTP-MIB');
     $port_stats = snmpwalk_cache_oid($device, 'vlanTrunkPortNativeVlan', $port_stats, 'CISCO-VTP-MIB');
 } elseif ($device['os'] != 'asa') {
-    $port_stats = snmpwalk_cache_oid($device, 'dot1qPortVlanTable', $port_stats, 'Q-BRIDGE-MIB');
+    $port_stats = snmpwalk_cache_oid($device, 'dot1qPvid', $port_stats, 'Q-BRIDGE-MIB');
```
laf: We only used dot1qPvid out of the entire table!

laf:
Comments on both old and new changes in: https://github.com/librenms/librenms/pull/6341/files

murrant:
@laf Looks nice!  Not sure when I will have time to test...

laf:
You can have faith :)

murrant:
I have faith, but for some portions of the code I require myself to test before approving ;)

laf:
Wimp :D

PR #6345, opened 2017-04-04
fix whitespace display on RRDTool Command

tsumaru720:
*edit* update images*

The RRDTool command section removes extra white-space which is used in formatting

See example images

rrd-old
rrd-new

#### Please note

#### Testers

### DIFF
--- a/html/css/styles.css
+++ b/html/css/styles.css
@@ -838,7 +838,7 @@ a.interface-upup:hover, a.interface-updown:hover, a.interface-admindown:hover {
   margin: 8px;
 }
 
-.infobox pre {
+.infobox .rrd-pre {
   background-color: #CCCCFF;
   font-size: 12px;
   white-space: pre-wrap;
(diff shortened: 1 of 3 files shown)


html/css/styles.css
```diff
@@ -838,6 +838,13 @@ a.interface-upup:hover, a.interface-updown:hover, a.interface-admindown:hover {
   margin: 8px;
 }
 
+.infobox pre {
```
murrant: This css selector is a little generic.  I'd prefer you make it more specific (with a class for example) or just put the style in the html code.

tsumaru720: will make it a new class - i did it this way so the bg colour is the same as infobox so if im specifying the colour, i feel it should be with the rest of the CSS instead of hidden away in code somewhere

new change coming shortly

tsumaru720:
hows that @murrant ?

PR #6349, opened 2017-04-04
Add OS Polling for Opengear ACM7008

EnzoZafra:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/opengear.inc.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+// SNMPv2-SMI::enterprises.25049.17.1.1 = STRING: "3.16.6 058e5600 ()" - ogFirmwareVersion
+// SNMPv2-SMI::enterprises.25049.17.1.2 = STRING: "7006851296650" - ogSerialNumber
+//
+$ogFirmwareVersion = snmp_get($device, '1.3.6.1.4.1.25049.17.1.1', '-OQv', '', '');
+$ogFirmwareVersion = explode(' ', $ogFirmwareVersion);
+$version = 'v' . trim($ogFirmwareVersion[0], '" ');
```
laf: Do we need to prepend the v here? We don't do that elsewhere and I think it's better to leave it as shown from the device.

EnzoZafra: Done, also removed (... ,'', '') in the snmp_get calls

PR #6367, opened 2017-04-05
Unit test to validate the db schema

murrant:
Changes build-schema.php to just write misc/db_schema.yaml

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
One issue with this is that we are trusting the users existing db is clean, if it's not then they will submit a bad yaml file which could lead to issues - it would get picked up by testing (if we get this bit working :)) so at least we have that safety net.

murrant:
This test would have benefits in both directions.

I'm not sure why it doesn't work on Travis.

laf:
Some debugging is showing that `dbFetchRows("SELECT TABLE_NAME FROM information_schema.TABLES WHERE TABLE_SCHEMA = '{$config['db_name']}' ORDER BY TABLE_NAME")` never returns any data and I've not been able to figure out why.

murrant:
@laf the database was empty.

laf:
@murrant not quite. The db is always built (doing a drop table .... shows this). However the connection (or lack of) to the db within phpunit seems to be the issue, the tables are created fine but the test unit isn't connected so is not returning anything.

```
diff --git a/tests/DBSetupTest.php b/tests/DBSetupTest.php
index 69fbcef..2bd8243 100644
--- a/tests/DBSetupTest.php
+++ b/tests/DBSetupTest.php
@@ -39,6 +39,7 @@ class DBSetupTest extends \PHPUnit_Framework_TestCase
     {
         if (getenv('DBTEST')) {
             global $config, $database_link;
+            $database_link = mysqli_connect('p:' . $config['db_host'], $config['db_user'], $config['db_pass'], null
             self::$sql_mode = dbFetchCell("SELECT @@global.sql_mode as sql_mode");
             dbQuery("SET NAMES 'utf8'");
             dbQuery("SET CHARACTER SET 'utf8'");

```

Is a fix, doesn't seem to be a bodge from what I've read and is the recommended way (well, the real recommended way is to do the connection per test unit. It will need updating to self::$database_link to be passed to each test unit or we can just do the full connection each time.

murrant:
@laf I see this  PR #5724 should fix that

PR #6370, opened 2017-04-06
refactor: schema updates

murrant:
Use transactions for schema updates
Test the schema files against 'NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION,STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO' and fix any issues

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/functions.php
```diff
@@ -2267,3 +2267,49 @@ function index_schema_to_sql($index_data)
...
+ * Check if the database schema is up to date.
+ *
+ * @return bool
+ */
+function db_schema_is_current()
+{
+    $current = get_db_schema();
+
+    $schemas = get_schema_list();
+    end($schemas);
```
f0o: are you sure they're sorted correctly?
Not that `1.sql 10.sql 100.sql` come before `2.sql`

murrant: Yes, glob() returns files alphabetically.  All existing sql files are prefixed with 0s.

laf:
Won't this introduce a higher possibility of breaking polling in the scenario where let's say we make 4 changes in one schema file and only 1 fails but the first 3 made changes to columns that the new code expects to use (say column name change in ports) - at this stage all 4 changes will be rolled back and polling can break from a single module all the way through to the entire thing.

Yes if the same thing happened now we would also experience issues but only if the 1 failed schema change was something the code was reliant on, the first 3 would have worked and not rolled back.

The above scenario gets worse the longer someone is from the last update, monthly releases are in that mix.

murrant:
Yes, but it would give us a chance to fix the schema file.

murrant:
I'm not strictly convinced using transactions is a good idea.

laf:
Hmm, surely the unit tests should be detecting the bad schemas (which I thought it already did)

murrant:
The other PR fixes that, and after that is merged, we can add STRICT_TRANS_TABLES and apply the fixes in the PR.

murrant:
Rebased and STRICT_TRANS_TABLES added.

laf:
I'm still not sure the transactions is a good thing for us, just not sure how rolling back a db schema change but not the code could be of any use.

murrant:
That is a very good point laf.   I'll pull the transaction so we can merge the rest.  Perhaps something more complex would be needed.

Pull schema update, apply update, if the SQL applies, update code, etc.  More things that could go wrong.

laf:
Don't forget we do schema updates in discovery as well. not just daily.sh

My take on it now (the schema side) is that we are fairly well covered now with the recent schema  + collation.

murrant:
I've removed the transactions.  Sorry I did a rebase, should have left the merge separate so it was easier to look at.

PR #6371, opened 2017-04-06
Fix redirect on login for instances behind reverse proxies

putnam:
On instances where base_url has been set for use behind a reverse proxy, logins are incorrectly redirected.

This happens because REQUEST_URI is set by the proxy:
  1. librenms has base_url set to http://site.com/nms/
  2. Browser requests http://site.com/nms/
  3. nginx reverse proxies /nms/ to librenms at http://somehost:1234/
  4. librenms sees REQUEST_URI as "/"
  5. librenms logs the user in, but sends "Location: /" to the browser. This redirects to the wrong location. Similar behavior occurs if REQUEST_URI points to some resource within librenms.

To resolve, concatenate REQUEST_URI (which is relative) to base_url. As base_url is slash-terminated, crop the trailing slash. This should have no effect on users with default settings and will correctly redirect instances behind reverse proxies.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/authenticate.inc.php
+++ b/html/includes/authenticate.inc.php
@@ -97,7 +97,7 @@ if ((isset($_SESSION['username'])) || (isset($_COOKIE['sess_id'],$_COOKIE['token
         $permissions = permissions_cache($_SESSION['user_id']);
         if (isset($_POST['username'])) {
             // Trim the trailing slash off of base_url and concatenate the (relative) REQUEST_URI
-            header('Location: '.substr($config['base_url'], 0, -1).$_SERVER['REQUEST_URI'], true, 303);
+            header('Location: '.rtrim($config['base_url'], '/').$_SERVER['REQUEST_URI'], true, 303);
             exit;
         }
     } elseif (isset($_SESSION['username'])) {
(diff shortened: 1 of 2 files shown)


html/includes/authenticate.inc.php
```diff
@@ -96,7 +96,8 @@
 
         $permissions = permissions_cache($_SESSION['user_id']);
         if (isset($_POST['username'])) {
-            header('Location: '.$_SERVER['REQUEST_URI'] ?: $config['base_url'], true, 303);
+            // Trim the trailing slash off of base_url and concatenate the (relative) REQUEST_URI
+            header('Location: '.substr($config['base_url'], 0, -1).$_SERVER['REQUEST_URI'], true, 303);
```
f0o: This variable does not always have a slash at the end.
Please add a sanity check before cutting off a character from the base_url; worst case you kill the tld.

laf:
Probably easier to just do something like rtrim($val, '/');

putnam:
Good catch. I didn't realize rtrim had that character mask param. Looks like that is both cleaner and solves the use case where there's no trailing slash.

PR #6381, opened 2017-04-07
feature: Added rrdtool version check to compare installed version with defined version

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Came across a user this week who had blindly set `$config['rrdtool_version'] = '1.6.0';` when running 1.4.3 resulting in poor performance apparently. Easy enough to test for.

### DIFF
--- a/validate.php
+++ b/validate.php
@@ -334,8 +334,8 @@ foreach ($bins as $bin) {
 }
 
 // Check that rrdtool config version is what we see
-if (isset($config['rrdtool_version']) && ($versions['rrdtool_ver'] != $config['rrdtool_version'])) {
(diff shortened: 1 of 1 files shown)


murrant:
The inverse is fine though, having a lower version defined than installed.  This effectively limits the features rrdtool can use (falls back to is_file() checks)

laf:
What benefit does that give the user? Imho users should set the version they use, it took ages to work out last time the version mismatch was the cause. 

The better solution is to remove that variable and auto detect the version.

murrant:
Perhaps, what is your rrdtool version and the variable on your main setup?

laf:
I don't specify it, on my dev, home and work installs it's left to just use the binary.

murrant:
I think for validate we should only raise the alarm if the version is newer than the binary

PR #6397, opened 2017-04-11
Added 'Video' device group and moved Axis cameras to this group'

EnzoZafra:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
I'm not a fan of video, from the sysDescr this could be a variety of things so appliance may be a better fit but overall I'm not sure we should change it.

EnzoZafra:
@laf Is it possible to at least move the network cameras to a separate group? Something like 'Surveillance' or 'Cameras'? The video encoders and video servers could stay in the 'Network' group.

laf:
That would mean splitting out this OS discovery into 3 separate OS'.

You are more than welcome to submit a pull request to do this but my opinion (not the projects) is that this should be kept simple - basically as it is with the appropriate type, appliance potentially fits better here.

EnzoZafra:
Okay, that makes sense. The important part for us is to separate the Axis cameras from the network group as we have hundreds of cameras and the switches are getting pushed back to the last few pages - so this will solve our problem

laf:
Needs a rebase and also a notification adding:

https://github.com/librenms/librenms.github.io/blob/master/notifications.rss

Use 'Change to Cisco UCM type from tele to collaboration' as an example, needs to give around 4 days notice.

laf:
I've merged the notification.

This should be merged on the 17th.

PR #6410, opened 2017-04-13
feature: use RrdDefinition to format and validate data

murrant:
Converted ucd-mib
Added simple array_column() function
Tests

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Out of interest what issue is this fixing?

murrant:
Trying to reduce code duplication and make it easier to get things correct.  This makes sure every DS is set and in the correct order.  Possibly could remove a variable from the data_update function and expand on this idea.

I don't feel that strongly about this approach.  This PR is guaging interest, that is why I didn't convert every data_update call.

murrant:
I tried to take this all the way to encapsulate full data definition for all data storages and things got a little weird (with extra data needed for rrd and name length restrictions)

I'm going to toss this on the back-burner for now and let it stew a bit.

PR #6413, opened 2017-04-13
Update Shebangs and daily.sh for FreeBSD compatibility

Pricetx:
Updated the shebangs on any files that did not have flags set to use /usr/bin/env, to provide support for FreeBSD in addition to Linux.
Updated daily.sh to support FreeBSD (su has different args, replaced with sudo when not on linux)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


daily.sh
```diff
@@ -113,7 +113,11 @@ main () {
         # only try to su if we are root (or sudo)
         if [[ "$EUID" -eq 0 ]]; then
             echo "Re-running ${DAILY_SCRIPT} as ${LIBRENMS_USER} user"
-            su -l "$LIBRENMS_USER" -c "$DAILY_SCRIPT $@"
+            if [[ "$(uname)" == "Linux" ]]; then
+                su -l "$LIBRENMS_USER" -c "$DAILY_SCRIPT $@"
+            else
+                sudo -u "$LIBRENMS_USER" "$DAILY_SCRIPT"
```
murrant: You are missing the arguments here `$@`

Also, we could probably use sudo for both Linux and FreeBSD...

Pricetx: Thanks, I had initially left it off due to it causing the script to break, but later realised it was just a quoting issue. I had intended to add it back but apparently forgot to. I will fix that now.

Regarding sudo on Linux, I'm perfectly fine to do so, but I didn't want to unnecessarily impose the additional dependency on Linux users if I didn't have to. If you're happy for everything to use sudo then i'll go ahead and remove the if statement altogether.

murrant: I think sudo is very standard.

Pricetx:
I have updated the pull request with the following changes:
- Added missing arguments to sudo command
- Removed su command so all platforms use sudo

laf:
lgtm

PR #6415, opened 2017-04-13
Create check_mssql_health.inc.php

premtimibrahimi:
This could be very usefull for the people who want to monitor MS SQL Servers. It is a very cool service check.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/services/check_mssql_health.inc.php
+++ b/includes/services/check_mssql_health.inc.php
@@ -1,0 +1,10 @@
+<?php
+
+$check_cmd = $config['nagios_plugins'] . "/check_mssql_health --server ";
+
+if ($service['service_ip']) {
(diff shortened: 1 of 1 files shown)


murrant:
check_mssql_health does not seem to be in the standard nagios/monitoring plugins.  @librenms/reviewers How should we deal with that?  Check for it's existence before listing? Add an note?

premtimibrahimi:
Honestly not sure, but integrating it within librenms would be a good idea. To make this service check work its simple.

laf:
@murrant The list of nagios checks is pulled from the plugin dir for the user so if you add any check_* file it will be listed so this is fine as it's only used when someone selects it.

murrant:
Oh great! Didn't even look at the code 😊

Issue #6412, opened 2017-04-13
EMC Isilon OneFS v8 discovery and polling broken


dragans:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Component | Version
--------- | -------
LibreNMS  | 737865c94f9ac5f5874c71db7cbe91823601570d
DB Schema | 182
PHP       | 7.0.17
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2

[Discovery](https://gist.github.com/dragans/179b707ab2b8a4405206ca62dd83d918)
[Poller](https://gist.github.com/dragans/a8ef71625d62de27cdec29c18da0667f)
[SNMP](https://gist.github.com/dragans/1109e0c9d97044f7686947cf3ba9357c)
[MIB for OneFS v8](https://github.com/librenms/librenms/files/919472/ISILON-MIB.txt)

Device is currently recognized as Generic Device and only device ports are discovered.

Seems like they changed SystemObjectID value to ".1.3.6.1.4.1.12325.1.1.2.1.1".

PR #6416, opened 2017-04-13
newdevice: Added detection and sensor support for EMC OneFS v8

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6412 

Also adds various sensors and storage support.


includes/definitions/onefs.yaml
```diff
@@ -10,3 +12,5 @@ over:
 discovery:
     - sysObjectId:
         - .1.3.6.1.4.1.12124.1
+    - sysDescr_regex:
```
murrant: Two questions: why include the ' v8'? and why _regex what you have here is the same as a normal sysDescr

PR #6433, opened 2017-04-15
Increased speed of loading syslog pages.

TimStallard:
When using the syslog functionality with many records (I tested with ~1.5 million), the UI becomes slow in several places:
- Loading the main syslog page or the page for each device is very slow. This is due to the priorities dropdown, which is ordered by level. Adding an index on the levels and priorities combined resolves this.
- Moving around the table for a single device is fairly slow, due to the count at the bottom being updated. A COUNT(timestamp) is currently performed, but there should not be any items with a null timestamp, so changing this to simply count rows using COUNT(*) significantly increases performance.
These issues are still present when using additional filters, but I can't think of a nice way in which this would be resolved without adding many more combined indexes (or maybe one large index for device, program and priority?).

I agree to the conditions of the Contributor Agreement contained in doc/General/Contributing.md.

#### Please note

#### Testers


laf:
@TimStallard How long did this index take to create for you?

TimStallard:
@laf I just ran a test on an instance with 1.76 million records, it took ~22.3 seconds to create. That's running inside an LXC container with two allocated cores and 2GB RAM.

I also ran the same test on a $40/mo digitalocean droplet (just running mariaDB with the database imported) and saw a similar time, ~23.8s.

laf:
Waiting for #6469 then we can rebase and fix this.

@TimStallard We've gone live with a new CLA, can you please sign that one so we are able to merge.

laf:
@TimStallard We're ready to merge this, can you please sign the new CLA please?

TimStallard:
Sorry, I think something weird must have happened last time I tried. Should be signed now.

laf:
Thanks very much for contributing :)

Issue #6434, opened 2017-04-16
Cisco devices changed to Generic Device


claushc:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)

Yes, running the daily.sh script daily, but tonight something fishy started to appear.

- [X] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.

Component | Version
--------- | -------
LibreNMS  | 924b751be46229bec2ff0a33330ef48553fcb80f
DB Schema | 185
PHP       | 5.5.9-1ubuntu4.21
MySQL     | 5.5.54-0ubuntu0.14.04.1
RRDTool   | 1.4.7
SNMP      | NET-SNMP 5.7.2

- [X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).

[discovery-as1.ker1.nefnet.dk.txt](https://github.com/librenms/librenms/files/924052/discovery-as1.ker1.nefnet.dk.txt)

[poller-as1.ker1.nefnet.dk.txt](https://github.com/librenms/librenms/files/924053/poller-as1.ker1.nefnet.dk.txt)

[generic-as1.ker1.nefnet.dk.snmpwalk.txt](https://github.com/librenms/librenms/files/924054/generic-as1.ker1.nefnet.dk.snmpwalk.txt)

- [X] Please provide as much detail as possible.

After tonights daily.sh run, LibreNMS decided to report all my Cisco devices as Generic Device, changing device icons and stuff.  This happened to three different types of devices, but I have only included one, since it's probably a very simple problem.

- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

laf:
Please update and re-test

claushc:
Problem fixed.  Thank you very much for your help :-)

murrant:
Thanks for the revert @laf any idea what went wrong?

laf:
[...]

PR #6443, opened 2017-04-18
fix: Allow line returns in snmprec files with the 4x data type

murrant:
Fill in IOS data files
IOS-XE discovery fails, fix that

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

issue: #6434


murrant:
trigger ci

laf:
No better way to do this to leave the test units human readable? Is it the detection that breaks or the test units that break without this?

murrant:
Tests are incorrect without this, which allowed the incorrect detection to slip through this weekend.  Unfortunately, no way to keep it human readable and stay compatible with snmpsim.

I did change it so it will print out the decoded string if tests fail.

laf:
:(

PR #6456, opened 2017-04-18
refactor: move ios detection to yaml using new sysDescr_except

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/ios.yaml
```diff
@@ -8,6 +8,15 @@ over:
...
 icon: cisco
+discovery:
+    -
+      sysDescr:
+          - 'Cisco Internetwork Operating System Software'
+          - 'IOS (tm)'
+          - 'Cisco IOS Software'
+          - 'Global Site Selector'
+      sysDescr_except:
+          - 'LINUX_IOSD'
```
laf: Do we not need to include (exclude?) IOS-XE here?

murrant: Sure, I had removed it earlier as I couldn't find any devices containing IOS-XE.  As I said on the other PR, no harm in adding it.

murrant: Ok, went back to look through the stats data.  Not sure how I missed that, but a **few** people have C4500 and C3850 switches, which needs the IOS-XE.  Adding test data.

Issue #6225, opened 2017-03-20
Incomplete CPU and RAM graphing on Enterasys N-series switches


breeman1:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the `Component | Version
--------- | -------
LibreNMS  | cd16a071ef6ad59752312b5e0f924b3a62048e47
DB Schema | 177
PHP       | 7.0.8-0ubuntu0.16.04.1
MySQL     | 10.0.25-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3` section of `./validate.php` which you can run from the cli.
- [X ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

[Entera_N3_snmpbulkwalk.txt](https://github.com/librenms/librenms/files/855376/Entera_N3_snmpbulkwalk.txt)
The snmpbulkwalk command does not finish as these switches apparently have issues with VLAN information polling, hence a huge file which I had to kill while running. Probably related to #4988.
[Entera_N3_discovery.txt](https://github.com/librenms/librenms/files/855377/Entera_N3_discovery.txt)
[Entera_N3_poller.txt](https://github.com/librenms/librenms/files/855378/Entera_N3_poller.txt)

This issues is related to the #5974, however on the N-series switches whether be they N3 or N7 modular switches we have a few slots per switches which can each its own CPU and RAM hardware, See below;
![image](https://cloud.githubusercontent.com/assets/16717388/24106041/737f0564-0d8f-11e7-9b9b-a8fd9b34180a.png)
[...]

PR #6458, opened 2017-04-18
fix: Updated enterasys mempools disco/polling to support multiple ram…

laf:
… devices

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This is now changed to dynamically discover the memory for enterasys. Existing mempools will break as it uses an index value of 0 which seems to not be valid in the snmp output. I'll happily do a notification for this.

Fixes: #6225


murrant:
I think users that are working right now will continue to work right?

laf:
I'm not 100% sure it will but have no devices to test :(

PR #6466, opened 2017-04-19
docs: Updated install docs + perf to support compressing file types and using http/2

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I'm not sure all these file types are needed but I've added more JS to be compressed which helps with page load times. Standardised the file types across apache + nginx.

Added an entry to Performance docs to say to enable http/2 where possible.


murrant:
Nice :)

PR #6469, opened 2017-04-19
refactor: DB Updates will now file level lock to stop duplicate updates

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I've tested `php build-base.php` and web installer both whilst running cron, validate and discovery to see if I could a) break the lock feature and/or b) have validate turn up bad schema which I couldn't on both accounts.

This is to stop the race condition we have when discovery.php could invoke a schema update when daily.sh is already doing the same. Works for me.....

Should now allow #6433 to run now without worrying about it over-running.


murrant:
@laf Do you think we should include lock expiration?   That would allow us to break stuck locks eventually.

laf:
@murrant shouldn't be difficult, get_lock() can probably do that check

laf:
@murrant I actually think it does what you want already. So the lock isn't only checked if the file exists, it looks to see via `ps` if the pid previously recorded is still running, only if it is does the lock stay.

includes/functions.php
```diff
@@ -2317,3 +2317,50 @@ function db_schema_is_current()
...
+    } else {
+        file_put_contents($lock_file, getmypid());
+    }
+}
+
+/**
+ * Clean up lock file for a given name
+ * @param $name
+ */
+function clean_lock($name)
```
murrant: I feel like this should be called release_lock()

laf: Agreed and done.

murrant:
Sorry for not getting this tested sooner, but I think I'd like to push it to the next release so we can have a good lead time of testing.

laf:
No worries on my part.

laf:
@murrant you ok with this now?

murrant:
Yeah, I was going to apply it and try an update then merge it, but I got busy.  If you are happy with it, you can merge.

Issue #6318, opened 2017-04-02
Version in /about is dated on 0sec UNIX time (1970)


genabi:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

validate.php:
Component | Version
--------- | -------
LibreNMS  | a0abcbb6496def0bacc77fb811064c314c6d087e
DB Schema | 184
PHP       | 5.6.30
MySQL     | 5.6.35-log
RRDTool   | 1.6.0
SNMP      | NET-SNMP 5.7.3

Hi,

I've tried to find a solution to this small problem, however my PHP knowledge is too poor: 

There's following code in html/pages/about.inc.php:
`$version_date    = $versions['local_date'];`
I can see that local_date is defined in includes/common.php:
`list($local_sha, $local_date) = explode('|', rtrim(`git show --pretty='%H|%ct' -s HEAD`));`
This command ran on my systems shows proper epoch date:
`$ git show --pretty='%H|%ct' -s HEAD`
`a0abcbb6496def0bacc77fb811064c314c6d087e|1491081724`
However on /about page i can see:
`Version	a0abcbb64 - Thu Jan 01 1970 01:00:00 GMT+0100`

laf:
Please post the full output from validate

genabi:
2 systems with similar problem:
FreeBSD 10.3-RELEASE-p17
FreeBSD 11.0-STABLE r314290M
Completely different setups, different locations, both with nginx, php-fpm, oracle-mysql


`==========================================================
Component | Version
--------- | -------
[...]

PR #6474, opened 2017-04-20
webui: Fixed the display date for the current version

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6318 

which doesn't work via the webinterface to detect git, whereis does but I'm not sure how universal that is. We now check the git response code to see if it exists. 1 is expected as we don't actually run a valid git command.

### DIFF
--- a/includes/common.php
+++ b/includes/common.php
@@ -1643,7 +1643,7 @@ function set_numeric($value, $default = 0)
 
 function check_git_exists()
 {
-    exec('git >> /dev/null 2>&1', $response, $exit_code);
+    exec('git > /dev/null 2>&1', $response, $exit_code);
     if ($exit_code === 1) {
         return true;
     } else {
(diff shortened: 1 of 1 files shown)


includes/common.php
```diff
@@ -1643,7 +1643,8 @@ function set_numeric($value, $default = 0)
 
 function check_git_exists()
 {
-    if (`which git`) {
+    exec('git >> /dev/null 2>&1', $response, $exit_code);
```
murrant: No need for double chevron `>>` here right?

laf: Technically no but we don't care for the output at all so I though suppressing it would be better. The exit code is all we care about here.

murrant: I mean isn't normally this?  `git > /dev/null 2>&1`

laf: I actually just took that from our cron file, I can update to > though

laf: Updated

PR #6485, opened 2017-04-20
location map regex replace pattern only 

MrDragon:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Support/Configuration.md
+++ b/doc/Support/Configuration.md
@@ -394,6 +394,10 @@ Regex Matching:
 ```php
 $config['location_map_regex']['/Sink/'] = "Under The Sink, The Office, London, UK";
 ```
+Regex Match Substitution:
+```php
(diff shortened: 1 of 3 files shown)


MrDragon:
I don't know if I did anything right... See https://community.librenms.org/t/location-mapping-regex-substitute/1054/3 - that's my submission, use it as you see fit :)

murrant:
I think this needs some documentation updates and you need to fix the one style issue.

laf:
Docs are definitely needed :)

MrDragon:
Thank you for pointing me at the docs. As to the style issue, I did find two lines with trailing spaces (fixed). If there's another style issue, I may need a bigger hint :(
Thank you for your patience and guidance!

laf:
imho it needs a bit more clarity in the docs, I can't tell from the example what it's doing.

PR #6488, opened 2017-04-20
newdevice: Etherwan managed switches

EnzoZafra:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Do all of the switches Etherwan provide run the same OS/Firmware? If not then I'd say the OS name should be updated to be less generic here.

EnzoZafra:
Didn't find much online however I know they supply one MIB for all their switches. The MIB used to be called Etherwan SW72000 which is not a name for any of their hardware so maybe its what they call their OS? Their MIBs are now called 'ewnSwitchM' (Etherwan Switch Management ?) though

includes/definitions/etherwan.yaml
```diff
@@ -0,0 +1,10 @@
+os: etherwan
+text: 'Etherwan Managed Switch'
+type: network
+icon: etherwan
+over:
+    - { graph: device_bits, text: 'Device Traffic' }
+    - { graph: device_processor, text: 'CPU Usage' }
+discovery:
+    - sysObjectId:
+        - .1.3.6.1.4.1.2736.1.1
```
laf: We have nothing in stats for this OID, I'd be inclined to change it to `.1.3.6.1.4.1.2736.`

includes/polling/os/etherwan.inc.php
```diff
@@ -0,0 +1,38 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2017 Lorenzo Zafra
+ * @author     Lorenzo Zafra<zafra@ualberta.ca>
+ */
+
+// SNMPv2-MIB::sysDescr.0 = STRING: Etherwan 6TX+2G Managed Switch, Firmware rev: 1.94.3 05/11/15 16:31:18
```
laf: Can remove that, we have the test data.

includes/polling/os/etherwan.inc.php
```diff
@@ -0,0 +1,38 @@
...
+
+$description = $poll_device['sysDescr'];
```
laf: No point in assigning this variable, just use `$poll_device['sysDescr']` directly below

PR #6490, opened 2017-04-21
Add Cisco FSM to ios.yaml

pblasquez:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/ios.yaml
+++ b/includes/definitions/ios.yaml
@@ -15,6 +15,7 @@ discovery:
           - 'IOS (tm)'
           - 'Cisco IOS Software'
           - 'Global Site Selector'
+          - 'Cisco Firewall Services Module'
       sysDescr_except:
           - 'LINUX_IOSD'
           - 'IOS-XE'
--- a/includes/definitions/ios.yaml
+++ b/includes/definitions/ios.yaml
@@ -15,7 +15,6 @@ discovery:
           - 'IOS (tm)'
           - 'Cisco IOS Software'
           - 'Global Site Selector'
-          - 'Cisco Firewall Services Module'
       sysDescr_except:
           - 'LINUX_IOSD'
           - 'IOS-XE'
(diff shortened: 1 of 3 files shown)


murrant:
You will need to add tests too.
Update OSDiscovery and add a new snmprec.

laf:
Does Cisco Firewall Services Module run IOS? For the 6500 it run fwsm: https://software.cisco.com/download/release.html?mdfid=277413409&softwareid=280775068&release=4.1(11)

laf:
@pblasquez Any feedback?

murrant:
Found this: http://www.cisco.com/c/en/us/products/interfaces-modules/catalyst-6500-series-firewall-services-module/index.html

`Based on Cisco PIX Firewall technology, the Cisco FWSM`

So, looks like it should be classified as PIX maybe?

pblasquez:
I looked for an fwsm OS but didn't find one, my short-term goal was to activate polling for CPU/RAM/Disk, which works fine using the IOS profile.

Can we add this PR to address this for the short term so I don't have to maintain my local patch, and open the new FWSM/PIX OS as a feature request discussion on the community site?

@murrant: For the test, do I just add the 'Cisco Firewall Services Module' entry to the correct OS function? Or something more involved than that?

laf:
We already have pix (pixos) as an OS this can be tagged onto. I'm not sure if it does really belong in it's own fwsm as that's what it's referred to as is the firmware you can download from it.

We shouldn't merge this into IOS as a work around, it needs to be done correctly first time.

pblasquez:
ok, didn't see pixos since I wasn't looking in the definitions directory yet.  I will update and add the test as well.

laf:
Does cpu/ram/disk work still now?

pblasquez:
I'm going to close this out and submit a new, rebased, clean patch.

PR #6492, opened 2017-04-22
docs: Updated Support-New-OS doc to provide clearer information

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I've tried to segment out the docs for adding new support so people can read / focus on what they want to do.


murrant:
Thanks @laf 
Probably could use some refinement like an easy way to move between the individual documents.  But merging this now.

PR #6495, opened 2017-04-23
fix: Updated ipv4/ipv6 discovery to exclude IPs with invalid port_ids

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

We just check we have a valid port_id before we carry on, this should clear out stale entries if the device still exists in librenms.

Fixes: #6302

### DIFF
--- a/includes/discovery/arp-table.inc.php
+++ b/includes/discovery/arp-table.inc.php
@@ -109,6 +109,11 @@ foreach ($vrfs_lite_cisco as $vrf) {
             d_echo(null, '-');
         }
     }
+
+    // remove entries that no longer have an owner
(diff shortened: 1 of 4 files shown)


murrant:
What IP address are you removing with this?  Other than that question the code change looks straight forward enough.

laf:
Ones that don't have a valid port id

murrant:
Sorry, I mean I don't understand what kind of ips would not have a valid port id and why I wouldn't want them.

Perhaps ports in badif?

laf:
Probably would have helped if I'd have linked to #6302

murrant:
Sure would have ;)

murrant:
Oh, it was an issue I created :blush: 

I don't think this fixes the problem.  I think the problem is in the $valid cleanup.

PR #6505, opened 2017-04-25
add to ipmi definitions

tsumaru720:
Add "Amps" to ipmi_unit definitions

```
# ipmitool sensor | grep Amps
Current 1        | 0.400      | Amps       | ok    | na        | na        | na        | na        | na        | na        
Current 2        | 0.200      | Amps       | ok    | na        | na        | na        | na        | na        | na
```

I've added this as the "current" metric as that seems to be whats used for my APC ups for tracking Amps on that.

---

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions.inc.php
+++ b/includes/definitions.inc.php
@@ -721,6 +721,7 @@ $config['ipmi_unit']['degrees C'] = 'temperature';
 $config['ipmi_unit']['RPM']       = 'fanspeed';
 $config['ipmi_unit']['Watts']     = 'power';
 $config['ipmi_unit']['Amps']      = 'current';
+$config['ipmi_unit']['percent']   = 'load';
 $config['ipmi_unit']['discrete']  = '';
 
 // Define some variables if they aren't set by user definition in config.php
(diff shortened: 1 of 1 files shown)


tsumaru720:
One of my boxes exposes fan speed as a percentage of its maximum via ipmi

```
Fan 1            | 10.976     | percent    | ok    | na        | na        | na        | na        | na        | na        
```

As we cant really tell what metric should be tracked here - For now I've opted for "Load". I think this should suffice with the description to explain what it is.

Screenshot (Its only polled once on my env so graphs no graphs yet drawn yet )
x

tsumaru720:
some more samples of "generic percentages"

```
CPU Usage        | 81.000     | percent    | ok    | na        | na        | na        | 101.000   | na        | na
IO Usage         | 0.000      | percent    | ok    | na        | na        | na        | 101.000   | na        | na
MEM Usage        | 3.000      | percent    | ok    | na        | na        | na        | 101.000   | na        | na
SYS Usage        | 78.000     | percent    | ok    | na        | na        | na        | 101.000   | na        | na
```

The description (first field) I think is freeform for vendors to put what they like so there's no reliable way of determining what "sensor type" it is so i feel best to lump all this under one metric (Load, because % based)

PR #6513, opened 2017-04-27
Use sysName for Application graph titles if not empty

aldemira:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Auto discovered devices with no reverse dns names are shown with IP addresses on graph titles. So I'm proposing to use sysName if not empty. That makes more sense IMHO.


laf:
I'm not sure on this one personally, seems to be specific. Why the change?

aldemira:
If a device with no reverse dns is auto discovered then the graph titles show IP addresses rather than a friendly name. I must admit I wasn't totally sure when sending this request :), but I am totally ok to delete this or to find another way.

florianbeer:
I would welcome this as well. I often set up large(-ish) networks with a lot of network components that only have IPs and sysNames, there is neither time nor need to set up DNS, rDNS for these.

laf:
Use this badly named function:

`ip_to_sysname($device, $hostname)`

Probably like this:

`shorthost(ip_to_sysname($device, $app_device['hostname']));`

You need to use the config option `$config['force_ip_to_sysname'] = true;` for this to work but basically anywhere an IP is shown for the hostname we switch out to sysName instead. It's not in place all over (just as you've seen here) but slowly it's being fixed.

Issue #3592, opened 2016-06-02
Look into Oxidized reload_nodes call during auto-addhost


Rosiak:
Seems like it's only doing it's thing during manual addhost?

laf:
Code indicates it runs anytime a device is added not based on the method.

murrant:
It doesn't actually work for some reason. I took a look at it at one point, but couldn't figure out why.

murrant:
Oddly, it does seem to work for the previously added device when adding the next device.  So it could be some sort of race condition.

laf:
Hi,

We are trying to tidy up old issues by working out which ones are still required or have not yet been resolved. If this issue should remain open - please update it with the latest debug information ensuring your install is up to date.

_TAG_TIDYUP

murrant:
This is still an issue.  I couldn't figure out how to fix it when I investigated.

laf:
Is this still an issue?

If it is, are we sure it's not an issue on the oxidized end? The code runs only when a device is added and we only serve the devices from the db for the api call so seems odd that this wouldn't work all the time.

PR #6515, opened 2017-04-27
fix: Actually reload oxidized when we should not when we think we should

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #3592 

In the end managed to replicate this so I'm confident this fix works + added support for reloading when deleting devices.

The short of it is, I don't think the OS had been discovered by the time the API call was made from Oxidized back to LibreNMS to tell it a new device was ready. This call is now done when new device discovery is called so every 5 mins which I think is acceptable. It's probably possible to move the call somewhere else but where it was before was pretty much at the end of the process of adding a new device anyway, the only other place would be in `addHost()` but I don't think it will make a difference.


murrant:
Awesome!

image

PR #6518, opened 2017-04-28
fix: inconsistent device discovery

murrant:
There were two or more separate $device arrays during discovery. When one was updated, others were not.
Combine those all.
Make sure the new os is loaded if the os changed.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
To reproduce this add a device and run discovery.  The most obvious thing is the OS gets set and the icon changes to generic.

I was just changing the community on my snmpsim for testing.

PR #6531, opened 2017-04-30
Added more sensors for Exalt ExtendAir devices

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/sensors/state/extendair.inc.php
+++ b/includes/discovery/sensors/state/extendair.inc.php
@@ -54,10 +54,83 @@ if (is_numeric($temp)) {
         }
     }
 
-    $descr = 'Link status for end radio';
+    $descr = 'Link status (far end radio)';
     discover_sensor($valid['sensor'], 'state', $device, $cur_oid, $index, $state_name, $descr, '1', '1', null, null, null, null, $temp);
     create_sensor_to_state_index($device, $state_name, $index);
 }
+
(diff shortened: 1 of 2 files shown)


includes/discovery/sensors/state/extendair.inc.php
```diff
@@ -131,6 +204,78 @@
...
+$cur_oid = '.1.3.6.1.4.1.25651.1.2.4.2.3.1.15.0';
+
+if (is_numeric($temp)) {
+    $state_name = 'locLinkStateV';
+    $index = $state_name;
+    $state_index_id = create_state_index($state_name);
+
+    if ($state_index_id !== null) {
+        $states = array(
+            array($state_index_id,'almNORMAL',1,0,0) ,
```
Rosiak: Wouldn't it be worth to consolidate the state translations, in this case, as they practically all are the same?

laf: You mean one state name?

esorin0:
https://pastebin.com/c71afa8x

laf:
According to that output, the two new sensors have been detected. What's actually wrong for you?

esorin0:
I have just posted the output, nothing is wrong.
You've added Frequency (MHz), which is something that is not really changing for a radio of this type and it's not really useful. Same with 'Link security status' state, which is a constant for a given link. Can we show TX/RX power on the device main page? Thanks!

Issue #6549, opened 2017-05-02
Add support for GE Digital Energy UPS


olivierbeytrison:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Component | Version
--------- | -------
LibreNMS  | eca2436d5436807da36367e87854dac7430a42ba
DB Schema | 188
PHP       | 7.0.15-0ubuntu0.16.04.4
MySQL     | 10.0.29-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3

discovery : https://p.libren.ms/view/8483b77d
poller : https://p.libren.ms/view/cc2b7c2c
snmpwalk : https://p.libren.ms/view/c988ea0e

If you need anything more just ask ;)
Thanks !

olivierbeytrison:
And this would be the MIB : https://p.libren.ms/view/6befc276

olivierbeytrison:
working on the code myself, should do a PR in an hour or so

PR #6553, opened 2017-05-02
Add support for GE UPS (#6549)

olivierbeytrison:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/ge-digitalenergy.yaml
```diff
@@ -0,0 +1,16 @@
...
+rfc1628_compat: 1
+over:
+    - { graph: device_current, text: Current }
+    - { graph: device_voltage, text: Voltage }
+    - { graph: device_load, text: Load }
+discovery:
+    -
+        sysObjectId:
+            - .1.3.6.1.4.1.818.1.100.1.1
+        sysDescr_regex:
```
laf: Do you need this as well? Using just sysObjectId on it's own is the preferred way unless this OID conflicts with another OS

includes/definitions/ge-digitalenergy.yaml
```diff
@@ -0,0 +1,16 @@
+os: ge-digitalenergy
```
laf: Is this the best name for the os? It's usually based on the firmware or actual OS the device runs like IOS, JunOS, etc. If it is then that's not an issue.

olivierbeytrison: tbh I don't even think there's an OS behind it. It's a simple extension card in the UPS to provide basic SNMP support. it *could* be changed to ge-ups

includes/definitions/ge-digitalenergy.yaml
olivierbeytrison: I don't think it would lead to conflicts to remove it.

Being new to librenms I'm slowly getting my hands on the code ;)

PR #6555, opened 2017-05-02
refactor: OS discovery tests are now dynamic

murrant:
All that needs to be done now is supply snmprec files with the proper name. os.snmprec or os_description.snmprec
I also tweaked the loading of yaml files to speed up the tests substantially.
Verifies all os have at least one test and all snmprec files are tested.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


doc/Developing/os/Test-Units.md
```diff
@@ -2,28 +2,15 @@ source: Developing/os/Test-Units.md
...
-        $this->checkOS('pulse', 'pulse-sa6500');
-        $this->checkOS('pulse', 'pulse-vaspe');
-        $this->checkOS('pulse', 'pulse-sa');
-    }
-```
-
-The above example has multiple test files, the first argument in the function call is the os name we expect the test 
-to pass as, the second argument is the filename to use for the test. If not filename is passed then the os name is used 
-for the file.
+All that you need to do is create an snmprec file in tests/snmpsim with the proper name. If the adding the first test for
```
Rosiak: Small typo:
"If the adding the first test for.."

murrant: Thanks, took me a sec :)

murrant:
Thanks for the rebase ;)

Issue #5264, opened 2016-12-29
AlertLog /alert-log page not loading in chrome


mmarchand:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php`.
- [x] If you would like us to add a new device or your device is listed as Generic then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)
- [x] Please provide as much detail as possible.

In Chrome 57 (and probably before since it has happened for quite a long time now)
Javascript console shows a :  
"Uncaught RangeError: Maximum call stack size exceeded"

laf:
Isn't Chrome 57 a dev release? I'm not sure we should be reacting to issues in a dev piece of software.

mmarchand:
same thing in 55 (just tested here ;)

laf:
I've tested 55 and don't see an issue. What commit # are you on for LibreNMS?

Does the same happen on https://demo.librenms.org/alert-log/ as that also works.

mmarchand:
I am using latest master
it works on the demo site,

I suspect it could be related to the number of devices (I have 4k+) 
I guess Chrome does not want to create a 4k list on the selection list (it does not show here, see screenshot)

https://gyazo.com/7ff062e5647bd6cb5147eb672e0074f1

mmarchand:
Just tested to comment out the 
```
foreach (get_all_devices() as $hostname) {
```
in pages/alert-log.inc.php

and the page loads fine (without the list box of course)

mmarchand:
strange thing, the listbox works in the eventlog page ... so I don't know what's wrong here

laf:
I guess we could change this to be ajax powered input field instead.

f0o:
@mmarchand Do you see anything in chrome's error-console?

mmarchand:
Yes : 
Uncaught RangeError: Maximum call stack size exceeded  (index:1)

laf:
Needs converting to an ajax fed field.

PR #6557, opened 2017-05-03
fix: alart-log many devices, load selects from ajax

murrant:
fixes: #5264

The issue was all the string concatenation...

But I decided to try loading the select via ajax. Seems ok.  We could just use something like select2.  Opinions? Use this approach or load at page load.
I was able to reduce the loading of the 4000 select items from 1.6s to 0.6s on my laptop by optimizing the sql.

I added 4000 devices to my database to check this :)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/ajax_list.php
+++ b/html/ajax_list.php
@@ -4,6 +4,7 @@
  * LibreNMS
  *
  * Copyright (c) 2014 Neil Lathwood <https://github.com/laf/ http://www.lathwood.co.uk/fa>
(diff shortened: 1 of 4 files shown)


murrant:
perhaps I should separate these out.  Fix the bug in one PR, submit the ajax select loading in another.

laf:
Don't think you need to separate it out.

Issue #6581, opened 2017-05-04
Support for interfaces with "virtual" in the description


swolski:
The issue was identified when noticing that eth0/1 and eth0/2 on an Adtran 908e were not populating into the ports table due to their description containing the word "Virtual".

functions.php (line 1012-1013) appears to be the culprit since it marks all ports with "irtual" as invalid (with the exception provided for Avaya VSP ports).
```
   } elseif (strstr($port['ifDescr'], "irtual") && strpos($port['ifDescr'], "Virtual Services Platform") === false) {
        $valid = 0;
```
Additional details attached and in the comments.

DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
==========================================================
Component | Version
--------- | -------
LibreNMS  | bdf1616c76797e9accc0e96d4c28e971d71cfefe
DB Schema | 189
PHP       | 5.5.9-1ubuntu4.21
MySQL     | 5.5.55-0ubuntu0.14.04.1
RRDTool   | 1.4.7
SNMP      | NET-SNMP 5.7.2
==========================================================
- [ ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).

[librenms-issue-20170504-mysql-output.txt](https://github.com/librenms/librenms/files/976953/librenms-issue-20170504-mysql-output.txt)
[librenms-issue-20170504-discovery-ports-output.txt](https://github.com/librenms/librenms/files/976946/librenms-issue-20170504-discovery-ports-output.txt)

- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

swolski:
From the discovery:
```
[...]

Issue #3263, opened 2016-03-17
Ports types not shown on Enterasys switches


breeman1:
Hi,

I have an issue where the port types and numbers (ifNames) are shown on the Enterasys switches in the GUI; only ifDescr is shown although looking at the discovery and polling command debugs; ifNames are being picked up. Please help, the discovery dump is here, https://dpaste.de/giys and teh poller dump here, https://dpaste.de/moOD. The screenshot is below;

![image](https://cloud.githubusercontent.com/assets/16717388/13854091/8338c704-ec72-11e5-8f57-9adfabf19e24.png)

murrant:
Add this code to your config.php to fix.
```
$config['bad_if'] = array_diff($config['bad_if'], array('ras'));
```

breeman1:
HI @murrant 

The code you provided does not seem to fix the issue. I re-read my initial post and I can see what caused the confusion, sorry, my bad. I meant to say "I have an issue where the port types and numbers (ifNames) are **NOT** shown on the Enterasys switches in the GUI..." instead of "I have an issue where the port types and numbers (ifNames) are shown on the Enterasys switches in the GUI..."  
Below is the screenshot of a portion of the ./discovery.php -h x.x.x.x -d -m ports command and I have highlighted in arrows the information I need to be displayed in the GUI over and above the ifDescr information which is already being displayed.

![snap2](https://cloud.githubusercontent.com/assets/16717388/18224854/0958fdd0-71e4-11e6-88c1-ca71720fb717.png)

breeman1:
Wait, I can actually see a lot more ports now being displayed on the Enterasys switches after this change, however I am still not getting the ifNames information being displayed. This change also seems to have resolved the problems I raised in issue [https://github.com/librenms/librenms/issues/2838](url), see below;

![snap2](https://cloud.githubusercontent.com/assets/16717388/18226673/6361433e-720f-11e6-8a72-35a2cf6cacdf.png)

PR #6584, opened 2017-05-04
feature: Added ability to whitelist ifDescr values from being ignored with good_if

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6581 

I'm not 100% sure this is the best approach but it will also solve the issue we've had before where users want to remove an item from bad_if so have to unset the whole array and add the others back in.


murrant:
@laf what do you think about removing the global bad_if and moving it into yaml/per is?

laf:
I don't think we will know which devices do / don't have interfaces we will ignore and it's going to duplicate a lot of config in yaml (imho)

murrant:
Probably, but I feel like we could remove some from the global that only make sense on one OS.  However, tracking those down may not be feasible.

... If you can't tell, I'm not a fan of bad_if :D

laf:
It's the tracking down that would be hard, we could obviously just remove them, add the ones we have in our installs to the correct OS and then see what is reported to us, however.....

  - it's a lot of duplication imho hence why bad_if isn't actually that bad
  - it will generate more 'support' noise which I don't particularly want :)

This approach is blacklist bulk, whitelist specific which seems to be a more logical approach no?

murrant:
@laf  Ok, sounds good, but can you please add a good_if for this:  https://github.com/librenms/librenms/issues/3263

laf:
Sure thing. Will do it later on.

laf:
Done

Issue #4715, opened 2016-10-04
Feature Request: Display hostname instead of ip addresses when force_ip_to_sysname is enabled


breeman1:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.
- [ ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
    Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ ] Please include all of the information between the `====================================` section of `./validate.php`.
- [ ] If you would like us to add a new device then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)
- [ ] Please provide as much detail as possible.

There are various places in the GUI where I think displaying hostnames or sysnames instead of ip addresses would come in handy. Like when searching for an ip address as shown below, one might want to zoom in on a particular device whilst doing the search, but it not that easy remembering the ip addresses of all the devices particularly when you many on your install. The second scenario was when trying to edit User Permissions, being able to assign by sysname would be great as I have `$config['force_ip_to_sysname']    = true;` in my config file.

![snap6](https://cloud.githubusercontent.com/assets/16717388/19071978/10f5c61a-8a33-11e6-8f4c-adedd610a985.png)
![snap3](https://cloud.githubusercontent.com/assets/16717388/19071983/1b4b57e2-8a33-11e6-85f1-39f55a21af20.png)
![snap4](https://cloud.githubusercontent.com/assets/16717388/19071989/20144e0a-8a33-11e6-84db-3ca17f1220c0.png)

laf:
I'm confused by this, force_ip_to_sysname (whilst a confusing name) basically converts something to sysName, we only use that for hostname. So any device you add with a fqdn will just display that, if it's added by IP then we switch to use sysName instead of this flag is enabled.

So the title for this is confusing as we will never use hostname instead of IP as they are both the same and depend on how you add the device.

Do you mean that you have some areas of the web interface not honouring that feature?

breeman1:
[...]

PR #6585, opened 2017-05-04
webui: Updated remaining display options where we do not show sysName if hostname is IP

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #4715 

Think this catches the last of the hostnames that are IPs but ip_to_sysname option is on. The name of the function may not be the best choice - any suggestions welcome.

### DIFF
--- a/includes/common.php
+++ b/includes/common.php
@@ -206,11 +206,9 @@ function get_all_devices()
     // FIXME respect $type (server, network, etc) -- needs an array fill in topnav.
 
     if (isset($cache['devices']['hostname'])) {
-        $devices = array_keys($cache['devices']['hostname']);
+        $devices = array_keys($cache['devices']);
     } else {
-        foreach (dbFetchRows("SELECT `hostname` FROM `devices`") as $data) {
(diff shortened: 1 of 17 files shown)


murrant:
How about display_device_name() or format_hostname()

includes/common.php
```diff
@@ -1128,21 +1126,21 @@ function inet6_ntop($ip)
...
 /**
- * Convert IP to use sysName
+ * If hostname is an ip, use return sysName
  * @param array device
- * @param string ip address
+ * @param string hostname
  * @return string
 **/
-function ip_to_sysname($device, $ip)
+function process_hostname($device, $hostname)
```
murrant: You could make the second argument optional, that would make most of the calls cleaner.

aldemira:
Last time I've forced ip_to_sysname, the devices which didn't have any sysname became unclickable in the all devices page. Maybe I should put this as an issue.

laf:
@murrant all updated

@aldemira also taken care of in this, we only use sysName if it's available.

PR #6586, opened 2017-05-05
Add Gaia definition

pblasquez:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
`.1.3.6.1.4.1.8072.3.2.10` is the standard linux sysObjectID.  You'll have to use something else.

pblasquez:
Sorry about that! I'm trying to find out why a checkpoint is reporting a linux OID now...the other gaia tests also shows a sysdecr with "Linux.*" but they have unique OIDs... will report back.

pblasquez:
So looks like gaia gets picked up even without the sysObjectID definition...closing.

Issue #6499, opened 2017-04-25
Enhancement: Add option to disable alerting for normal users with device permissions


tobiasa1:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Component | Version
--------- | -------
LibreNMS  | dcc6c96ec4c646feade95eac7f91d402c4dd312b
DB Schema | 186
PHP       | 5.5.9-1ubuntu4.21
MySQL     | 5.5.54-0ubuntu0.14.04.1
RRDTool   | 1.4.7
SNMP      | NET-SNMP 5.7.2

When assigning device permissions to a normal user without admin privileges this user are receiving alert e-mails for that specific device and there seems to be no option to disable that. I suggest we add an option in the General Alert Settings tab to Issue alerts to normal users Yes/No

f0o:
I wouldn't call it a bug, it's by design.

if there's a general demand for this feature, I can easily make a PR for this

laf:
I'd definitely say it's a bug. When you assign device permissions to a user they start getting alerts with no way to turn it off other than disabling the rules.

PR #6590, opened 2017-05-05
alerting: Added option to enable/disable option for sending alerts to normal users

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6499 

By default, adding a normal user and assigning them a device means they will get alerts for this and can't disable it. With this you can now enable / disable that ability.

I've chosen to default this off as I think that should be the fresh install option, this will obviously stop alerts for users who now expect the old way is how it should be for which I've put a notification in. Overall I'm not fussed if we default to on or off but just felt that we shouldn't default to what was the old (broken imho) way and we should do it properly.


laf:
Actually added schema file now!

laf:
Feel free to push the change to my branch

laf:
If you do change it to default on, please remove the update from notifications as that's no longer relevant.

murrant:
How about different defaults for new and existing installs.

laf:
That would require adding it as false to config.php.default and means users can't just use the webui to make changes to the option. 

We should just pick one way or the other.

laf:
bump

PR #6593, opened 2017-05-05
fix: Restore access to existing wireless graphs

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Is their a point in adding this back if we no longer poll for data?

murrant:
So users can see their graphs still.   We didn't remove the graphs and I didn't intend too.

At a later time we would remove both the graphs and this.

PR #6598, opened 2017-05-06
newdevice: Added support for Terra sti410C

maciejkaczkowski:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


html/images/os/terra.svg
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
```
murrant: SVG size can be cut down by 1/3, you can trim it down with: https://jakearchibald.github.io/svgomg/

includes/definitions/sti410c.yaml
```diff
@@ -0,0 +1,10 @@
+os: sti410c
```
laf: Is this OS name too specific to one model? Ideally an OS covers a range of models in a product line.

includes/polling/os/sti410c.inc.php
```diff
@@ -0,0 +1,4 @@
+<?php
+
+$version = trim(snmp_get($device, ".1.3.6.1.4.1.30631.1.9.4.1.0", "-Oqv"), '"');
```
laf: Please convert these two snmp_get calls to snmp_multi_get_oid()

includes/definitions/terra.yaml
```diff
@@ -0,0 +1,10 @@
+os: terra
```
laf: That's one extreme to the other :)

Does this cover all terra devices?

maciejkaczkowski: It covers channel processor devices (DVB-T to IP, DVB-S to DVB-C etc, index 5-10 and 14-18).
All have custom embbeded OS (different for each model), I made tests only for sti410 (9) and sti440 (18). Each device has its own MIBs available after purchase. Only version and serial number are under the same OID.

It doesn't cover products with index:
1 - OD120 - optical receiver - discontinued
2 - MO411 - optical transmitter (other purpose)
3 - MO4x8 - optical transmitter (other purpose - not DVB processor)
4 - UC412 - can't find any info
11 - S2C16 - Similar GUI to sti410, should also work
12 - OT001M - can't find any info
13 - S2C16P - Similar GUI to sti410, should also work

includes/discovery/processors/terra.inc.php
```diff
@@ -0,0 +1,23 @@
...
+
+if ($device["os"] === "terra") {
+    $descr = "Processor";
+
+    $query = array(
+       array("sti410C", ".1.3.6.1.4.1.30631.1.9.1.1.3.0"),
+       array("sti440",  ".1.3.6.1.4.1.30631.1.18.1.326.3.0")
+    );
+
+    $sysDescr = strlen($device["sysDescr"]) == 0 ? snmp_get($device, ".1.3.6.1.2.1.1.1.0", "-Ovq") : $device["sysDescr"];
```
laf: What's the reason for doing this? $device['sysDescr'] should already have what you need.

includes/discovery/sensors/temperature/terra.inc.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+if ($device["os"] === "terra") {
+    $query = array(
+        array("sti410C", ".1.3.6.1.4.1.30631.1.9.1.1.4.0"),
+        array("sti440",  ".1.3.6.1.4.1.30631.1.18.1.326.4.0")
+    );
+
+    $sysDescr = strlen($device["sysDescr"]) == 0 ? snmp_get($device, ".1.3.6.1.2.1.1.1.0", "-Ovq") : $device["sysDescr"];
```
laf: Same here.

includes/discovery/sensors/voltage/terra.inc.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+if ($device["os"] === "terra") {
+    $query = array(
+        array("sti410C", ".1.3.6.1.4.1.30631.1.9.1.1.5.0"),
+        array("sti440",  ".1.3.6.1.4.1.30631.1.18.1.326.5.0")
+    );
+
+    $sysDescr = strlen($device["sysDescr"]) == 0 ? snmp_get($device, ".1.3.6.1.2.1.1.1.0", "-Ovq") : $device["sysDescr"];
```
laf: Same here.

includes/discovery/processors/terra.inc.php
maciejkaczkowski: $device["sysDescr"] is empty after adding and first discovery (at cron: /opt/librenms/discovery.php -h new). sysDescr is available only after manual discovery or after /opt/librenms/discovery.php -h all. Without that line first graphs are shown only after the discovery.php -h all (after a few hours in the worst case).

murrant: Is your install up to date?  This should have been fixed recently.

maciejkaczkowski: Created and tested on acaa29c583049ba31937e40e2b339bbc6851dab7 (5 may 2017).

laf: I've just tested this and adding a device does populate the sysDescr value straight away.

maciejkaczkowski:
Sorry for the last commit, it should be separate pull request - I don't know how to remove auto commit to current thread.

laf:
Thanks for the pull request.

I reverted your last commit by the way as it was for another device, can you submit that as a different pull request please?

maciejkaczkowski:
Ok, Thank you. I will make few updates and create another request.

PR #6606, opened 2017-05-07
Feature/add t echo

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

DO NOT MERGE

Thought this might be useful to try and tidy up poller/debug output as it's a pain to read through some times. php-cli-tools is a modified version which is within the librenms org, I've added support for outputting the index used for the arrays as we have a lot of data which don't just have 0,1,2,3,etc as indexes but 100, 501, 1501, etc (ifIndex).

We could probably roll our own but this has support for other cli processing + still works when you pipe the data to other programs.

Creating PR for feedback, if it's good I'll try and covert a lot of poller/disco to use it.


laf:
Example output https://www.irccloud.com/pastebin/MtpEzw0S/

murrant:
I made a change to make headers optional.
I also fixed the autoloader (`composer dump-autoload -o`)

laf:
You happy with it then if I convert some poller files over?

murrant:
If it makes things easier to read, go ahead.  We could also drop our console_color2 dependency, but that doesn't have to be done in this PR.

laf:
Hmm, started on ports polling and because of the size of the columns you get:

https://p.libren.ms/view/3c8c890a

I'm not sure that's actually anymore readable than:

https://p.libren.ms/view/6e750e02

So closing this down :(

murrant:
Hmm, yeah, it is just a lot of data...

Issue #6223, opened 2017-03-20
rrdcached - MySQL App -  Error Drawing Graph


wriedel:
Hi Folks,

if I enable
```
$config['rrdcached']    = "unix:/var/run/rrdcached.sock";
$config['rrdtool_version'] = '1.6.0';
```
everything works smoothly beside the graphs for the MySQL Application.

```
RRDTool Command
[...]

PR #6608, opened 2017-05-08
fix: MySQL app graphs fail with rrdcached

murrant:
Do not send -1 values when we set the minimum as 0.  Send the proper U for non-values.
fixes: #6223

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Looks good so far, you'll notice the graphs containing bad values fail stil:
image

PR #6611, opened 2017-05-09
fix: page/graph load speed: part 1

murrant:
Speed up load_all_os() by caching the full $config['os'] array
Function time goes from 1.13648s -> 0.00073s
The cache is updated during discovery, but only if it is >22hrs old.
Additionally, the cache is not used and a full yaml load is forced if the cache is >24hrs old.
The cache time is controlled by $config['os_def_cache_time'], I don't see any reason anyone would change that time.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,3 @@ rrd
 /vendor
 composer.phar
 librenms_theme
-/cache/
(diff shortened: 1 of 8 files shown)


laf:
lgtm, leaving to run overnight on my test install

PR #6612, opened 2017-05-09
Support for ADVA FSP150 EGX and GE11x devices - Temp Sensors

czilian:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
That's a lot of mibs!

I can already see duplicated mibs like LLDP-MIB.mib so you'll need to remove all of the ones we have as standard already.

czilian:
Got it. Will do. Thx,

includes/definitions/advafsp150eg-x.yaml
```diff
@@ -0,0 +1,13 @@
+os: advafsp150eg-x
```
laf: This os name is too specific, especially seeing as you say it covers another device.

The OS should be named after the operating system / firmware which is run on these devices. If they don't have one then it should be the group of products this covers.

includes/definitions/advafsp150eg-x.yaml
```diff
@@ -0,0 +1,13 @@
...
+text: ADVA FSP150 EGX
+type: network
+icon: adva
+group: MetroE
```
laf: What's this? The group would be to tie in multiple OS' to one vendor, like cisco with ios, ios-xe, etc.

includes/definitions/advafsp150eg-x.yaml
```diff
@@ -0,0 +1,13 @@
...
+over:
+    - { graph: device_bits, text: Traffic }
+    - { graph: device_temperature, text: Temperature }
+discovery:
+    - sysObjectId:
+      - .1.3.6.1.4.1.2544.1.12.1.1.7
+mib_dir:
+    - advafsp150eg-x
```
laf: The mib_dir should be the manufacturers name ideally.

includes/definitions/advafsp150eg-x.yaml
czilian: Ok. Making a product just work or having it structured vendor-centric to follow the overall concept its what missing on my side. I will try to learn from your approach to APC. There you cover also quite a wide range of products with different readings. For now just bin my pull request. Hope you can help here and there by looking at my fork upon request before I place a pull request again. Thx

Issue #6613, opened 2017-05-09
Pretty mysql for alerts breaks regex rules


XioNoX:
Running LibreNMS version 1.27

In https://github.com/librenms/librenms/blob/master/includes/alerts.inc.php#L39
The pretty print array converts  '('  to ' ( ' and  ')'  to ' ) ' (adds spaces on both sides).

If my alert term (or macro) looks like:
```
$config['alert']['macros']['rule']['is_main_intf'] = '%ports.ifAlias ~ "^(Cust|Transit|Peering|Core).*"';
```
The mysql query will be of such:

```
sudo php ./poller.php -h <hostname< -d -r -m system,os
[...]
SQL[SELECT * FROM ports WHERE (ports.device_id = '93') && ( ( ports.ifAlias REGEXP "^ ( Cust|Transit|Peering|Core ) .*" )  = "1"    &&   (  (  ( ports.ifInOctets_rate*8 )  / ports.ifSpeed ) *100 )  < "80" )] 
```
Because of those extra spaces, the part:
> ports.ifAlias REGEXP "^ ( Cust|Transit|Peering|Core ) .*"  

Does not match what its supposed to, as the regex processor is expecting extra spaces.

The workaround (and most likely fix), tested and validated on my side is to:
set
> '('  => '(', ')'  => ')' 

(Or remove the mapping for those parenthesis).
in https://github.com/librenms/librenms/blob/master/includes/alerts.inc.php#L39

Which works fine:
```
sudo php ./poller.php -h <hostname< -d -r -m system,os
[...]
SQL[SELECT * FROM ports WHERE (ports.device_id = '93') && ((ports.ifAlias REGEXP "^(Cust|Transit|Peering|Core).*") = "1"    &&  (((ports.ifInOctets_rate*8) / ports.ifSpeed)*100) < "80" )] 
[...]
Status: ALERT
```

I also confirmed it by running the mysql command manually.

Please let me know if you need more informations!

PR #6614, opened 2017-05-09
Pretty mysql for alerts breaks regex rules

XioNoX:
Fixes: #6613

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/alerts.inc.php
+++ b/includes/alerts.inc.php
@@ -36,7 +36,7 @@ function GenSQL($rule)
         return false;
     }
     //Pretty-print rule to dissect easier
-    $pretty = array('*'  => '*', '('  => '(', ')'  => ')', '/'  => '/', '&&' => ' && ', '||' => ' || ', 'DATE_SUB ( NOW (  )' => 'DATE_SUB( NOW()');
+    $pretty = array('&&' => ' && ', '||' => ' || ');
     $rule = str_replace(array_keys($pretty), $pretty, $rule);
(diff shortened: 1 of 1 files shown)


includes/alerts.inc.php
```diff
@@ -36,7 +36,7 @@ function GenSQL($rule)
         return false;
     }
     //Pretty-print rule to dissect easier
-    $pretty = array('*'  => '*', '('  => ' ( ', ')'  => ' ) ', '/'  => '/', '&&' => ' && ', '||' => ' || ', 'DATE_SUB ( NOW (  )' => 'DATE_SUB( NOW()');
+    $pretty = array('*'  => '*', '('  => '(', ')'  => ')', '/'  => '/', '&&' => ' && ', '||' => ' || ', 'DATE_SUB ( NOW (  )' => 'DATE_SUB( NOW()');
```
murrant: It looks like `'&&' => ' && '` and `'||' => ' || '` are the only two worthwhile mappings.  Probably just remove the rest.  It is interesting that this already contains two worthless replacements...

murrant: Or maybe the entire thing...

f0o:
Honestly i don't 100% recall why I escaped all of these with spaces. But i faintly recall that a mysql version disliked certain rules without spaces.
Not a blocker here, merge ahead if it works with upstream mysql and mariadb.

Issue #4127, opened 2016-08-19
PHP session contains user login credentials


ekoyle:
mneth in IRC reported that ibrenms is storing passwords in the PHP session file.  I think this is an unnecessary security risk.

It looks like every page request will call reauthenticate() with a chained call to authenticate() if there is no cookie.  It looks like this is the only place where the password is used.  Should this be using a local variable instead of $_SESSION? It appears to be set around [html/includes/authenticate.php line 39-47](https://github.com/librenms/librenms/blob/bf2c522c899a41c726830d5bc63f95dc8f6e7a72/html/includes/authenticate.inc.php#L39) and then only used on line 64.
#### Please note

> Please read this information carefully.
- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
- [ ] Please include all of the information between the `====================================` section of `./validate.php`.
- [ ] If you would like us to add a new device then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)
- [x] Please provide as much detail as possible.

f0o:
Correct, one of the many relics we inherit.

This will not be present in v2 :)

Luckily the attack vector for this implies a breach into the filesystem of the nms's server. At which point one could easily drain a database dump and break the terribly outdated pw-lib we use.

I guess I'm the pessimist again.

Not saying we shouldn't fix this! Likely just an 'if' and drop the pwd storage ;)

On 19 August 2016 19:20:47 CEST, Eldon Koyle <notifications@github.com> wrote:
>mneth in IRC reported that ibrenms is storing passwords in the PHP
>session file.  I think this is an unnecessary security risk.
>
>It looks like every page request will call reauthenticate() with a
>chained call to authenticate() if there is no cookie.  It looks like
>this is the only place where the password is used.  Should this be
>using a local variable instead of $_SESSION? It appears to be set
>around [html/includes/authenticate.php line
>39-47](https://github.com/librenms/librenms/blob/bf2c522c899a41c726830d5bc63f95dc8f6e7a72/html/includes/authenticate.inc.php#L39)
>and then only used on line 64.
>
>#### Please note
>
>> Please read this information carefully.
>
>- [x] Is your install up to date? [Updating your
>install](http://docs.librenms.org/General/Updating/)
>- [ ] Please include all of the information between the
>`====================================` section of `./validate.php`.
>- [ ] If you would like us to add a new device then please provide the
>information asked for
[...]

PR #4608, opened 2016-09-26
Trust PHP session and remove remember me cookie

ekoyle:
#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.
- [x] Have you signed the [Contributors agreement](http://docs.librenms.org/General/Contributing/)
- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)
- [ ] remove 'remember me' remnants from database schema

My rationale: By putting usernames/passwords in the session, we have already extended the ultimate level of trust there.  Reduce our exposure by removing username/password, and eliminate redundant handling caused by 'remember me' cookie.
- mitigate some session fixation concerns
- trust PHP sessions to handle 'remember me' functionality and login functionality
- add some session handling helper functions
- fix some active directory bugs (don't try to set user_id on insert to users table)

FIxes #4127

PR #6615, opened 2017-05-09
fix: Improve authentication load time and security

murrant:
page/graphs speedup part 2

**Write close the session as soon as we no longer need to write to it. Prevents the session from blocking other requests.**
Do not run through full authentication functions if the session is already authenticated.
Do not bind to AD every request when the session is already authenticated.
Extract login, logout, and remember_me functions.
Use Exceptions for authentication error handling.

Removes password from the session (See #4608)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixes: #4127

### DIFF
--- a/LibreNMS/Exceptions/AuthenticationException.php
+++ b/LibreNMS/Exceptions/AuthenticationException.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ * AuthenticationException.php
+ *
+ * Throw this when authentication fails.  Sets the default message to 'Invalid credentials' so we don't leak any info.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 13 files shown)


laf:
Still having graph issues with this and also getting kicked frequently.

One of the graph issues is that no width/height are set.

murrant:
@laf OK, I think this is ready for testing.  I didn't find anything obviously wrong.

I also found out calling ldap_bind() on include in ad auth was a huge waste of time if we aren't going to even use it most of the time.

rbax82:
Deployed on my setup. Wow what a difference in page load time.

olivierbeytrison:
Currently in test on our setup with radius auth. There's a clear speed improvement!

murrant:
Thanks for testing!

laf:
lgtm

laf:
@murrant Good for a merge from your perspective? If so, merge away

Issue #1196, opened 2015-06-04
Poll FDB tables and allow searching for MAC addresses


xenith:
Poll/store FDB/CAM tables from devices (mainly layer2 switching devices) and allow for searching of MAC addresses within the CAM tables.

Example: snmpbulkwalk -v2c -c 'COMMUNITY'@VLAN DEVICE -M mibs/ -m BRIDGE-MIB dot1dTpFdbTable -OqsX

dot1dTpFdbAddress[0:50:56:89:77:62] 0:50:56:89:77:62
dot1dTpFdbAddress[24:e9:b3:ae:e6:16] 24:e9:b3:ae:e6:16
dot1dTpFdbPort[0:50:56:89:77:62] 50
dot1dTpFdbPort[24:e9:b3:ae:e6:16] 50
dot1dTpFdbStatus[0:50:56:89:77:62] learned
dot1dTpFdbStatus[24:e9:b3:ae:e6:16] learned

markuspachali:
I also would like to have that feature. Would be great to have the option to search for the MAC of a client device and find the switchport to which is is connected.

aauvinet:
Hi, 
Me too.....
I have got  five handred devices, I need  to find the location's (switch/interface) Client, PC, printers...



Cdlt,

rlphnger:
With thousands of devices, this would be a killer feature. Now we have to use another tool for MAC-searching. Having it in LibreNMS would be awesome.

markuspachali:
If it helps to implement that feature:
NeDi (nedi.ch) has such a feature already implemented. The code is licensed under GPLv3, so it should maybe possible to adopt their implementation to have the feature also in LibreNMS.

aauvinet:
HI,
Where is the NeDi's code ?

markuspachali:
Hi,
you can download it from www.nedi.ch/download

awlx:
Just added some HP switches to my librenms install and encountered the same problem.

ekoyle:
I have actually worked on this before, we have some custom software that
polls fdb/arp tables and stores the data in a db.  We keep it for a month
or so and  It actually turns out to be a lot of data in the history, even
though we have configured it to only keep track of the fdb on edge ports.

I have some questions about implementation details.

How often would we want to poll these tables? IIRC, it seems like the
discovery already hits the fdb/arp tables, so storing the data there should
be relatively low impact.  Adding them to the normal polling could greatly
[...]

PR #6616, opened 2017-05-09
feature: fdb tables

EnzoZafra:
The code is heavily based on the implementation of the ARP tables and the following thread: #1196 

At the moment, this implementation of FDB tables is for IOS and Comware OS'.
I put this within the discovery module - either run all modules in ./discovery.php, or use '-m fdb-tables' 

Can we discuss whether we want it in the poller module instead? I was thinking it could be added into the poller module, however, manage it with cron so that it doesn't call it as often as the main poller.

Also suggestions/discussion about the code design or UI would be great.

Screenshot:
screenshot

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@EnzoZafra Looks cool.

What would you think about putting this data into the ipv4_mac table? It seems pretty similar.

ftaurino:
Can this branch cover also HP and Huawei switches? With Cisco, this will cover over 70% of ethernet switch market share.

EnzoZafra:
@murrant, I was looking at that, but it didn't seem like it would make sense. I believe that the ipv4_mac table maps mac addresses to the corresponding IP address. 
The FDB tables here store the mac address that has been learned and which port it was learned on. I tried comparing both data sets and I didn't find a good relation other than being able to map the ip address and mac address together.

@ftaurino, Unfortunately, we don't have any test devices for those switches. I would need to coordinate with somebody who has them available. 
Developing support for those should not be hard and I'm willing to help if somebody wants to develop it.

ftaurino:
hi enzo, I'd be happy to help and there's a network with over 30 hp switches (1910, 1920 and 3800) waiting for us!

ftaurino:
I can clone your branch and start looking in the includes/discovery/fdb-table.inc.php file

EnzoZafra:
@ftaurino The important part is finding the OIDs for the fdb table. If we can search the MIBs for it that would be good

murrant:
Having this information in LibreNMS is great, but I'm unsure about how it should be presented.
For example:
Do you show one entry for a mac address and list all the ports/devices it is shown on or list them all individually?
For a user that might not know what fdb is, what is the difference between fdb and mac addresses?

EnzoZafra:
@murrant 
I agree that the way it is presented right now could be changed since it may be confusing because of the similarities with the other tables.

At the moment it lists mac address entries individually. It is possible to have the same mac address for the same device and the same port, however they should each be in a different VLAN. This is what it looks like for that case:
image

I thought about grouping the entries by Vlans and each entries inside a vlan group would be unique, but the table would look different from the other search tables already in Librenms.

Maybe others have an input/suggestion?

murrant:
I think I would really like to see the mac data combined with this.  I'll try to give some more specific pointers when I have time.

alythemonk:
@murrant I would like to request that this be a separate function. We built this as a separate function into LibreNMS so that:

1. We can easily identify the device, mac, IP, port and vlan
2. There would be far too many columns in mac component, reducing its value and simplicity
3. Adding the FDB feature as a standalone component allows us to not impact the current mac functionality, allow for wider user feedback and device expansion for the component.

There is always the potential to integrate details from this component further down the line into other relevant areas.

alythemonk:
@murrant I think we may see benefits by keeping these components separate. Whilst they share some common values, the FDB has more targetted values for network admins. Also whilst this is a new component is maybe a good idea to keep them separate for now. Also it would allow us to further understand the wider user base's feelings on this.

Other contributors may expand past the Cisco and Nokia/ALU devices. The potential to integrate down the line into other parts of the LibreNMS system would exist. But day one this maybe a better way to go. Thoughts ? 

PS: I am waiting on the integration of this patch to mainline so that we can pick it up. Lorenzo has it in our dev instance here, but its not going to do us much good there :)

alythemonk:
@murrant I think we may see benefits by keeping these components separate. Whilst they share some common values, the FDB has more targetted values for network admins. Also whilst this is a new component is maybe a good idea to keep them separate for now. Also it would allow us to further understand the wider user base's feelings on this.

Other contributors may expand past the Cisco and Nokia/ALU devices. The potential to integrate down the line into other parts of the LibreNMS system would exist. But day one this maybe a better way to go. Thoughts ? 

PS: I am waiting on the integration of this patch to mainline so that we can pick it up. Lorenzo has it in our dev instance here, but its not going to do us much good there :)

murrant:
Looks like you have some sort of bug in the discovery:
```
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
SQL[UPDATE `ports_fdb` set `port_id` ='' WHERE `device_id` = '20' AND `vlan_id` = '1' AND `mac_address` = 'e8f724470bb7'] 
```

murrant:
I think this needs to be integrated a little better:
1. ipv4_mac and ports_fdb should be separate tables, otherwise, it will be hard to remove invalid entries.
2. ports_fdb should reference the vlan in the vlan table, not just be a raw vlan id.  May need to add entries if they are missing...
3. there should only be one mac address search which will search a union of both tables.  Obviously, entries from fdb will have more data.

includes/snmp.inc.php
```diff
@@ -651,6 +655,9 @@ function snmp_gen_auth(&$device)
     } elseif ($device['snmpver'] === 'v2c' or $device['snmpver'] === 'v1') {
         $cmd  = " -".$device['snmpver'];
         $cmd .= " -c '".$device['community']."'";
+        if ($vlan) {
+            $cmd .= '@' . $vlan;
```
EnzoZafra: I tried adding '-n vlan-[vlanid]' to the command but I'm not getting the same output as if I did 'communitystring@[vlan]'

https://pastebin.com/Q7uxfx6W

EnzoZafra:
1. I am not quite sure what you mean by this. Do you mean I should not be inserting into the table 'ipv4_mac'? I'm only using that table to map the mac_address to an IP address if its an entry.

2. Ah yes I should have been using the vlan table - I'm working on this.
I believe that the vlans are being inserted during the discovery module 'vlans'. Do you think its redundant to check and insert? Both the vlans and fdb module use the same OID for the vlans so it should be covered.

3. When you say search are you talking about the GUI? 

4. Are you doing a full discovery or are you just running this module?

includes/snmp.inc.php
murrant: snmp v3 has a separate switch for context
snmp v2 and v1?  use the `@context` format.

I can implement that for you if you like.

EnzoZafra: Ah I understand now. No worries, I'll do it!

murrant:
1. requires no action
2. I'm saying you can't guarantee the vlans discovery has been run before your module.
3. Yes, the webui
4.  I'll test again :D

ftaurino:
perhaps it could be useful to have two more columns in the db table, like "first_seen" and "last_seen" (timestamp), and a configurable parameter to show mac addresses in the last, say, 24 (or more) hours. and only remove entries oldest than 30/60/90 days.

alythemonk:
I agree @ftaurino with your request. Currently we have Lorenzo working on some other things internally here, waiting for this initial patch to be accepted. Once this is in main line, and in people's hands, we can actually spin up a second round patch that can be submitted to add your requested item.

EnzoZafra:
@ftaurino, can you please sign the Contributors License Agreement?

ftaurino:
sure! after the pull request there was no link or other steps if I remember correctly. perhaps in the next request? or in other ways?

html/pages/search/fdb.inc.php
```diff
@@ -0,0 +1,97 @@
...
+}
+
+$sql .= " WHERE F.port_id = P.port_id AND P.device_id = D.device_id $where GROUP BY `D`.`device_id`, `D`.`hostname` ORDER BY `hostname`";
+foreach (dbFetchRows($sql, $param) as $data) {
+    echo '"<option value=\"'.$data['device_id'].'\""+';
+    if ($data['device_id'] == $_POST['device_id']) {
+        echo '" selected "+';
+    }
+
+    echo '">'.$data['hostname'].'</option>"+';
```
laf: You need to use `format_hostname()` here. See html/pages/search/ipv4.inc.php for an example

sql-schema/193.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE IF NOT EXISTS `ports_fdb` ( `port_id` int(11) NOT NULL, `mac_address` varchar(32) NOT NULL, `vlan_id` int(11) NOT NULL, `device_id` int(11) NOT NULL ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
```
laf: The _id's in this can all be unsigned INTs can't they?

laf:
I've made a couple of inline comments +:

1. I think you should split out the os code into other includes like we do elsewhere, i.e: `include 'includes/discovery/fdb-table/ios.inc.php`.

2. The `includes/discovery/fdb-table.inc.php` contains db queries for vlan and port info which means we rely on those discovery modules to run. I don't think we should do this as we have no way to force these modules to run and they could be disabled. It would be much better to lookup the data in snmp.

trs80:
Just noting that Enterasys/Extreme switches support have a similar feature called [nodealias](http://www.extremenetworks.com/wp-content/uploads/resources/mibs/cabletron/ctron-alias-mib.txt) which includes various optimisations like being able to note the time you last queried the table, and then only query for new entries since then. So it'd be nice to at least think about that in the design.

EnzoZafra:
@laf 
1. Done
2. So instead of ids that refers to the tables that you mentioned would it be preferred to insert raw vlan and raw port data into the 'ports_fdb' table? 

Or: I could get the data through snmp, then check the database for a corresponding entry. If it doesn't exist then make an insert. However this process would be very similar to the vlans and ports module. How do you guys feel about calling the includes/discovery/vlans.inc.php and includes/discovery/ports.inc.php module if needed?

laf:
You should either store the raw index of the port so you can match with a query later on or do the same snmpwalks as the other modules but only use the data you need.

alythemonk:
@laf there seems to be a slight disjoint between your guidance and that of @murrant. Which way would you prefer it to proceed. We can do this either way but lets have consensus. I would like to polish this off our plate and re-task Lorenzo to other patches for submission. Thoughts ?

alythemonk:
Lorenzo and I discussed this a little more, if VLANs and Port discovery is turned off, there seems to be little to no value for FDB to run. We'd like to make it a pre-req that in order to use the FDB functionality that VLAN and Port discovery be running or enabled. Thanks.

laf:
You can't require users to run other modules, it's not part of the code base so unless you also add that support then you can't rely on those modules being run.

@murrant Let's clear up the issue on using other tables / module data.

EnzoZafra:
Looking at the implementation for arp-tables, it also uses an ifIndex => port_id dictionary by querying the DB, which is what I've done here for both the port_id and vlan_id. Does that mean that the implementation for arp-tables should be changed too?

murrant:
Perhaps we need dependencies for modules.

laf:
Existing modules aren't always correct in themselves but it's no reason to allow further code that would just need resolving later to be added at this level.

laf:
I'm going to leave this be, I'm fine with it being merged as is but would prefer the changes I said to be made if possible.

EnzoZafra:
The module now checks if any inserts to the DB are needed for vlans. The ports table on the other hand, I'm having a little trouble because that table has a lot more columns/data that I need to retrieve by snmp.

alythemonk:
@laf are you okay with us generating a database view and using that to feed info into the various tables ?

laf:
I guess show the view you'd create so we can see :) We don't use any at the moment but if it fixes an issue and doesn't cause problems then it's no problem.

murrant:
I'm not sure if I'm ok with a view.  They can be a pain when it comes to maintenance and all they are is basically a pre-defined query.

ftaurino:
Please, don't use view. A lot of LibreNMS users use Mysql as db, and with a lot of sql constructs (like aggregates, distinct, group by, having, limit...), this db server creates temporary tables on disk, negatively impacting performance. Look at https://dev.mysql.com/doc/refman/5.7/en/derived-table-optimization.html

murrant:
Ok, I went through this and ended up changing quite a bit so I created a PR against your branch.
https://github.com/EnzoZafra/librenms/pull/3

I'm ok with the separate search for now, but I would like to bring all mac related information into one ui.

EnzoZafra:
I merged the PR. I removed TiMOS support to make all the changes that this PR simpler. I can start developing it once this PR is merged

laf:
The two mibs you've added don't seem to be used so you can remove those. If they are needed, please put them in the relevant directory within /mibs/ related to the OS that would use them.

After that. Looks good to me. Thanks for contributing this :)

murrant:
I'm unsure about my change to add indexes to the table generally, it is a good idea, but could we perhaps overflow INT on the index and cause issues?

Options: Use BIGINT or remove the index again.

laf:
@murrant What's your specific change as I can't pick it out

misc/db_schema.yaml
```diff
@@ -1126,6 +1126,16 @@ ports_adsl:
...
     interface_id: { Name: interface_id, Columns: [port_id], Unique: true, Type: BTREE }
+ports_fdb:
+  Columns:
+    device_id: { Field: device_id, Type: 'int(11) unsigned', 'Null': false, Default: null, Extra: '' }
+    mac_address: { Field: mac_address, Type: varchar(32), 'Null': false, Default: null, Extra: '' }
+    ports_fdb_id: { Field: ports_fdb_id, Type: int(11), 'Null': false, Default: null, Extra: auto_increment }
+    port_id: { Field: port_id, Type: 'int(11) unsigned', 'Null': false, Default: null, Extra: '' }
+    vlan_id: { Field: vlan_id, Type: 'int(11) unsigned', 'Null': false, Default: null, Extra: '' }
+  Indexes:
+    PRIMARY: { Name: PRIMARY, Columns: [ports_fdb_id], Unique: true, Type: BTREE }
```
murrant: laf this is the key I added to the prots_fdb table

ckrez:
The addition of the following lines in this commit is breaking polling on some devices, specifically bgp polling on Junos. It's adding the '@' to the end of the community string being sent to the device:

if (isset($device['context_name'])) {
            $cmd .= '@' . $device['context_name'];
        }

SNMPD_AUTH_FAILURE: nsa_log_community: unauthorized SNMP community from <librenms> to <mgmt_ip> (<community>@)

Removing the '@' from the $cmd resolves the polling issue.

laf:
Sorry i dont have time to look the issue so I've had  to revert the pr

EnzoZafra:
@murrant, should we revert it back to appending the vlan instead of using contexts? Or should this be fixed in the bgp polling side for Junos?

laf:
It's been updated and merged in already so nothing more needs to be done :)

EnzoZafra:
Ah I missed the PR. Thank you!

laf:
No, thank you for the work you did :)

Issue #5385, opened 2017-01-10
Add support for APC netbotz exp modules.


sp1rr3:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ x] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ x] Please provide as much detail as possible.

Component | Version
--------- | -------
LibreNMS  | 322bebfd7ae8a872a5c59bd5708d018a2e175951
DB Schema | 153
PHP       | 7.0.8-0ubuntu0.16.04.3
MySQL     | 10.0.28-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3

Hi, 
When i add my netbotz 200 it only lists the sensors connected directly connected to it, not the sensors connected to the expansion modules. 

Please see attached pictures:

![libre-netbotz](https://cloud.githubusercontent.com/assets/25038742/21818961/0f555eea-d76a-11e6-991f-93556a5931b1.PNG)
![netbotz](https://cloud.githubusercontent.com/assets/25038742/21818962/0f56b5ba-d76a-11e6-94f0-5219e7fa8d82.PNG)
![all-netbotz](https://cloud.githubusercontent.com/assets/25038742/21819049/5f73bb38-d76a-11e6-804c-9fff7fc5ff59.PNG)

Last picture shows all sensors connected..

I can see them when i snmpwalk it.

For ex : 
1.3.6.1.4.1.318.1.1.10.4.2.3.1.5.0.5 shown in LibreNMS (GX1:5-C3)

Not shown: 1.3.6.1.4.1.318.1.1.10.4.2.3.1.5.1.3 (GT2:3-C5)

Rosiak:
Please provide the info asked:
./discovery.php -h HOSTNAME -d -m os
[...]

PR #6619, opened 2017-05-09
newdevice: Added more health information for APC units

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #5385 

I've fixed two current states that didn't show the translation. The change will actually break graphs but as these are only states then I've not put a notification in.


laf:
Confirmed working in the issue. @librenms/reviewers worth a notification for state graphs changing in this?

Rosiak:
Meh, personally I don't care about state graphs as they typically are pretty static. 
So IMO not worth it.

laf:
Rebased.

Yeah they aren't as useful day to day, it's mainly the alerting that is of importance.

PR #6621, opened 2017-05-10
refactor: move default alert rules into the collection

murrant:
Also adds the wireless rules to the default set.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/print-alert-rules.php
+++ b/html/includes/print-alert-rules.php
@@ -22,7 +22,7 @@ if (isset($_POST['create-default'])) {
             'device_id' => -1,
             'query'     => GenSQL($add_rule['rule']),
             'severity'  => 'critical',
-            'extra'     => '{"mute":false,"count":"1","delay":"300"}',
+            'extra'     => '{"mute":false,"count":"1","delay":"300","invert":false,"interval":300}',
             'disabled'  => 0,
             'name'      => $add_rule['name']
         );
--- a/html/includes/print-alert-rules.php
+++ b/html/includes/print-alert-rules.php
@@ -11,82 +11,23 @@ $no_refresh = true;
 </div>
 <?php
 if (isset($_POST['create-default'])) {
-    $default_rules[] = array(
-        'device_id' => '-1',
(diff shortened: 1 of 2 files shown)


html/includes/print-alert-rules.php
```diff
@@ -11,82 +11,23 @@
...
     require_once '../includes/alerts.inc.php';
+
     foreach ($default_rules as $add_rule) {
-        $add_rule['query'] = GenSQL($add_rule['rule']);
-        dbInsert($add_rule, 'alert_rules');
+        $insert = array(
+            'device_id' => -1,
+            'query'     => GenSQL($add_rule['rule']),
+            'severity'  => 'critical',
+            'extra'     => '{"mute":false,"count":"1","delay":"300"}',
```
laf: Only one alert to be sent? I know it's user editable but should we consider -1 still for unlimited?

html/includes/print-alert-rules.php
```diff
@@ -11,82 +11,23 @@
...
     require_once '../includes/alerts.inc.php';
+
     foreach ($default_rules as $add_rule) {
-        $add_rule['query'] = GenSQL($add_rule['rule']);
-        dbInsert($add_rule, 'alert_rules');
+        $insert = array(
+            'device_id' => -1,
+            'query'     => GenSQL($add_rule['rule']),
+            'severity'  => 'critical',
+            'extra'     => '{"mute":false,"count":"1","delay":"300","invert":false,"interval":300}',
```
laf: My old comment got hidden:

Only one alert to be sent? I know it's user editable but should we consider -1 still for unlimited?

murrant: This only affects when you click add default alerts.  If a user selects one from the collection, those default values are elsewhere. Defaults here we're copied exactly from the old code.

laf: The old code had a mix of 1 and -1 based on the type of rule :)

murrant: Oh, I see it now.  I'd say default to 1 and add extra to the rules that have -1...

PR #6626, opened 2017-05-10
newdevice: Rittal LCP Liquid Cooling Package

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

------------------

- ENTITY-SENSOR-MIB is used to returns data without specific support
- Humidity includes things like the SD Card usage or valve opening because it's values in percent

screen shot 2017-05-10 at 11 52 45

- Temperatures have wrong divisors, but I don't know how to correct this because this is a OID stating with the same values

temperature

- Voltage divisors are wrong but changing this in the functions.php file has no effect

screen shot 2017-05-10 at 11 52 49

- Cooling Capacity isn't a Power information, but it's in Power sensors because this is a value in kW

screen shot 2017-05-10 at 11 52 51

- Latest thing is that a new sensor type exists : l/min and it's not yet supported, it can be used to view the water flow in the cooling unit

For me, it's not a real problem. I prefer to have the information in bad categories than not to have it at all. Advices are welcome.


FTBZ:
Always waiting for an answer from Rittal about these problems to be sure to take the correct way.

FTBZ:
OK, finally got a response from Rittal about this. ENTITY-SENSOR-MIB/RFC 3433 isn't yet supported but OIDs are used... It's planned but I've no release date.

So what's the best way here? In my opinion problems are not really critical, I prefer to have these informations on the wrong sensor type rather than not having them.

PR #6628, opened 2017-05-10
feature: Added functionality to support recording status info

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

We don't currently store any data from the applications when polled. I've used the app_status column to store the response from the os-updates application poller. This could be done for other applications as well.

### DIFF
--- a/includes/polling/functions.inc.php
+++ b/includes/polling/functions.inc.php
@@ -549,7 +549,7 @@ function location_to_latlng($device)
  * @param array $app app from the db, including app_id
  * @param string $response This should be the full output
  */
-function update_application($app, $response)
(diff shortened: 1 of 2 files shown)


includes/polling/functions.inc.php
```diff
@@ -549,16 +549,17 @@ function location_to_latlng($device)
  * @param array $app app from the db, including app_id
  * @param string $response This should be the full output
  */
-function update_application($app, $response)
+function update_application($app, $response, $current = '')
```
murrant: Probably update the phpdoc here to indicate how $current should be used.

laf: Updated

PR #6636, opened 2017-05-12
refactor: modest speedup to database config population

murrant:
Page/Graph speedup part 3

use references to assign values into the config array more efficiently, prevents a lot of memory churn caused by the recursive approach

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
I dumped the config with both this and the previous code and the arrays were equal  (and array_diff was empty).  Some things were in different order, but that doesn't affect anything.

PR #6638, opened 2017-05-12
newdevice: HPE RT3000 UPS

marcuspink:
Initial support for HP(E) R/T3000 UPS which is basically a rebranded EATON 5P UPS.
That's why MG-SNMP-UPS-MIB is used instead of a specific HPE MIB.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/os/hpe-rtups.inc.php
+++ b/includes/polling/os/hpe-rtups.inc.php
@@ -1,8 +1,7 @@
 <?php
+$rtups_data = snmp_get_multi_oid($device, 'deviceSerialNumber.0 deviceFirmwareVersion.0', '-OQs', 'CPQPOWER-MIB');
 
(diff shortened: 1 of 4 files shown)


includes/polling/os/hpe-rtups.inc.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+$version = trim(snmp_get($device, 'upsmgIdentFirmwareVersion.0', '-OQv', 'MG-SNMP-UPS-MIB'), '" ');
```
laf: Thanks for the PR.

Can you change all of the snmp_get() calls here to one snmp_get_multi_oid(). See includes/polling/os/extendair.inc.php as an example

laf:
image

Issue #6622, opened 2017-05-10
Geist PDU


basn:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X ] Please provide as much detail as possible.
- [ X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Component | Version
--------- | -------
LibreNMS  | b0c1885d731edc7f129072fb19f3d48c88cc4d5c
DB Schema | 190
PHP       | 7.0.15-0ubuntu0.16.04.4
MySQL     | 10.0.29-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3

https://p.libren.ms/view/cc74adc7
https://p.libren.ms/view/041156cd
https://p.libren.ms/view/b640c04a

As far as i understand of the implementation of GEIST its only its watchdog appliance, but not the actual PDU.

I see that the voltage is being graphed but its at 0 all the time.

I can if someone on IRC wants access help out to check this issue if needed to be checked on the live system.

The PDU right now shows  Volts	230.10 Vrms etc.

PR #6646, opened 2017-05-15
newdevice: Added support for Geist PDU

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6622


includes/definitions/geist-watchdog.yaml
```diff
@@ -2,9 +2,13 @@ os: geist-watchdog
...
 type: environment
 icon: geist
+group: geist
 discovery:
-    - sysObjectId:
-        - .1.3.6.1.4.1.21239.
+    -
+        sysObjectId:
+            - .1.3.6.1.4.1.21239.
+        sysDescr_except:
```
murrant: Isn't this unneeded since geist-pdu is first?

laf: I like to be safe, removing anyway.

laf: Actually phpunit fails when I do. So leaving in place

murrant: That's strange, makes it seem like I'm missing something...

murrant: I removed it locally and tests ran fine.  I'm not bothered by it though.

laf: I'm also now getting errors when run pre-commit about other OS' so it could be my local install :/

PR #6648, opened 2017-05-16
refactor: Remove eval() from multi-tenancy code

murrant:
I couldn't figure out how much use this is since you can't have multiple databases.
If anyone wants to keep this, at least we should replace the eval() with array_replace_recursive()

Code origin: https://github.com/librenms/librenms/commit/371e394d8663991ad0f4997c7d83093f1e4af8ba

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/index.php
+++ b/html/index.php
@@ -84,14 +84,10 @@ if (strstr($_SERVER['REQUEST_URI'], 'widescreen=no')) {
 
 # Load the settings for Multi-Tenancy.
 if (isset($config['branding']) && is_array($config['branding'])) {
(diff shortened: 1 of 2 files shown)


laf:
What's multiple dbs got to do with it? It just uses an array with the server name as one of the keys to pull out custom logos/css etc.

I tried to remove this many moons ago but it got closed (back in 2013 / early 2014). I don't care for it but some people might - those who allow customers to log in and see stuff can have there own custom logo and things.

murrant:
Ok, sounds fine to me, I'll change it to remove the evals instead of dropping.

PR #6657, opened 2017-05-16
Add TiMOS temperature and power supply state sensors.

peelman:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/sensors/state/timos.inc.php
+++ b/includes/discovery/sensors/state/timos.inc.php
@@ -57,7 +57,7 @@ $power_supply_oids = array(
 );
 
 
-$pst = snmpwalk_cache_numerical_oid($device, 'tmnxChassisPowerSupplyTable', $power_supply_table = array(), 'TIMETRA-CHASSIS-MIB', 'aos', '-OQUsn')
+$pst = snmpwalk_cache_numerical_oid($device, 'tmnxChassisPowerSupplyTable', $power_supply_table = array(), 'TIMETRA-CHASSIS-MIB', 'aos', '-OQUsn');
 $power_supply_table = end($pst);
 
 foreach ($power_supply_oids as $data) {
--- a/includes/discovery/sensors/state/timos.inc.php
+++ b/includes/discovery/sensors/state/timos.inc.php
@@ -1,7 +1,5 @@
 <?php
 
-$power_supply_table = end(snmpwalk_cache_numerical_oid($device, 'tmnxChassisPowerSupplyTable', $power_supply_table = array(), 'TIMETRA-CHASSIS-MIB', 'aos', '-OQUsn'));
-
 $nokia_device_state_name  = 'tmnxDeviceState';
 $nokia_device_state_index = create_state_index($nokia_device_state_name);
 if ($nokia_device_state_index !== null) {
@@ -47,7 +45,7 @@ if ($nokia_device_ps_assigned_types_index !== null) {
     }
 }
 
-$base_oid          = '.1.3.6.1.4.1.6527.3.1.2.2.1.5.1';
+$base_oid          = '.1.3.6.1.4.1.6527.3.1.2.2.1.5.1.';
 $power_supply_oids = array(
     array('sub_oid' => '2.1.1', 'desc' => 'AC Status', 'type' => $nokia_device_state_name),
     array('sub_oid' => '3.1.1', 'desc' => 'DC Status', 'type' => $nokia_device_state_name),
@@ -58,8 +56,11 @@ $power_supply_oids = array(
     array('sub_oid' => '10.1.1', 'desc' => 'Power Out', 'type' => $nokia_device_state_name),
 );
 
+
+$power_supply_table = end(snmpwalk_cache_numerical_oid($device, 'tmnxChassisPowerSupplyTable', $power_supply_table = array(), 'TIMETRA-CHASSIS-MIB', 'aos', '-OQUsn'));
+
 foreach ($power_supply_oids as $data) {
-    $full_oid = $base_oid . "." . $data['sub_oid'];
+    $full_oid = $base_oid . $data['sub_oid'];
     $index = "tmnxChassisPowerSupplyTable." . $data['sub_oid'];
     discover_sensor($valid['sensor'], 'state', $device, $full_oid, $index, $data['type'], $data['desc'], 1, 1, null, null, null, null, $power_supply_table[$full_oid], 'snmp', $index);
     create_sensor_to_state_index($device, $data['type'], $index);
--- a/includes/discovery/sensors/state/timos.inc.php
+++ b/includes/discovery/sensors/state/timos.inc.php
@@ -57,7 +57,8 @@ $power_supply_oids = array(
 );
 
 
-$power_supply_table = end(snmpwalk_cache_numerical_oid($device, 'tmnxChassisPowerSupplyTable', $power_supply_table = array(), 'TIMETRA-CHASSIS-MIB', 'aos', '-OQUsn'));
+$pst = snmpwalk_cache_numerical_oid($device, 'tmnxChassisPowerSupplyTable', $power_supply_table = array(), 'TIMETRA-CHASSIS-MIB', 'aos', '-OQUsn')
+$power_supply_table = end($pst);
 
 foreach ($power_supply_oids as $data) {
     $full_oid = $base_oid . $data['sub_oid'];
--- a/includes/discovery/sensors/state/timos.inc.php
+++ b/includes/discovery/sensors/state/timos.inc.php
@@ -5,75 +5,75 @@ $power_supply_table = end(snmpwalk_cache_numerical_oid($device, 'tmnxChassisPowe
 $nokia_device_state_name  = 'tmnxDeviceState';
 $nokia_device_state_index = create_state_index($nokia_device_state_name);
 if ($nokia_device_state_index !== null) {
-  $nokia_device_states      = array(
-    array($nokia_device_state_index, 'Unknown', 1,1,3),
-    array($nokia_device_state_index, 'Not Equippped', 1,2,3),
-    array($nokia_device_state_index, 'Ok', 1,3,0),
-    array($nokia_device_state_index, 'Failed', 1,4,2),
-    array($nokia_device_state_index, 'Out Of Service', 1,5,1),
-    array($nokia_device_state_index, 'Not Provisioned', 1,6,3),
-  );
-  foreach ($nokia_device_states as $state) {
-    $insert = array(
-      'state_index_id'        => $state[0],
-      'state_descr'           => $state[1],
-      'state_draw_graph'      => $state[2],
-      'state_value'           => $state[3],
-      'state_generic_value'   => $state[4],
+    $nokia_device_states      = array(
+        array($nokia_device_state_index, 'Unknown', 1,1,3),
+        array($nokia_device_state_index, 'Not Equippped', 1,2,3),
+        array($nokia_device_state_index, 'Ok', 1,3,0),
+        array($nokia_device_state_index, 'Failed', 1,4,2),
+        array($nokia_device_state_index, 'Out Of Service', 1,5,1),
+        array($nokia_device_state_index, 'Not Provisioned', 1,6,3),
     );
-    dbInsert($insert, 'state_translations');
-  }
+    foreach ($nokia_device_states as $state) {
+        $insert = array(
+            'state_index_id'        => $state[0],
+            'state_descr'           => $state[1],
+            'state_draw_graph'      => $state[2],
+            'state_value'           => $state[3],
+            'state_generic_value'   => $state[4],
+        );
+        dbInsert($insert, 'state_translations');
+    }
 }
 
 $nokia_device_ps_assigned_types_name  = 'tmnxChassisPowerSupplyAssignedTypes';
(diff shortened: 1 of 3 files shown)


includes/discovery/sensors/state/timos.inc.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+$power_supply_table = end(snmpwalk_cache_numerical_oid($device, 'tmnxChassisPowerSupplyTable', $power_supply_table = array(), 'TIMETRA-CHASSIS-MIB', 'aos', '-OQUsn'));
```
laf: What's going on here?

includes/discovery/sensors/state/timos.inc.php
```diff
@@ -0,0 +1,79 @@
...
+    array('sub_oid' => '3.1.1', 'desc' => 'DC Status', 'type' => $nokia_device_state_name),
+    array('sub_oid' => '6.1.1', 'desc' => 'Power Supply 1', 'type' => $nokia_device_state_name),
+    array('sub_oid' => '7.1.1', 'desc' => 'Power Supply 2', 'type' => $nokia_device_state_name),
+    array('sub_oid' => '8.1.1', 'desc' => 'Power Supply Type', 'type' => $nokia_device_ps_assigned_types_name),
+    array('sub_oid' => '9.1.1', 'desc' => 'Power In', 'type' => $nokia_device_state_name),
+    array('sub_oid' => '10.1.1', 'desc' => 'Power Out', 'type' => $nokia_device_state_name),
+);
+
+foreach ($power_supply_oids as $data) {
+    $full_oid = $base_oid . "." . $data['sub_oid'];
```
laf: You should just add the trailing `.` to `$base_oid` to make this easier to read.

includes/discovery/sensors/temperature/timos.inc.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+if ($device['os'] === 'timos') {
```
laf: You don't need this check

includes/discovery/sensors/temperature/timos.inc.php
```diff
@@ -0,0 +1,30 @@
...
+    $oids = snmp_walk($device, 'tmnxHwID', '-Osqn', 'TIMETRA-CHASSIS-MIB', 'aos');
```
laf: Does one of the snmpwalk_* functions not do what you want here?

includes/polling/os/timos.inc.php
```diff
@@ -10,13 +10,16 @@
 
 $version = 'v' . $majorVersion . '.' . $minorVersion . '.' . $versionModifier;
 
-//SNMPv2-MIB::sysDescr.0 = STRING: TiMOS-B-9.0.R3 both/hops Nokia SAS-Sx 48Tp4SFP+ (PoE) 7210 Copyright (c) 2000-2017 Nokia.
-//All rights reserved. All use subject to applicable license agreements.
-//Built on Thu Jan 5 11:01:16 IST 2017 by builder in /home/builder/9.0B1/R3/panos/main
+$chassis_type_name_array = snmpwalk_cache_oid($device, 'tmnxChassisTypeName', $a = array(), 'TIMETRA-CHASSIS-MIB', 'aos', '-OQUs');
+$hardware = end(reset($chassis_type_name_array));
```
laf: Why is this being done, how big is the snmpwalk?

includes/polling/os/timos.inc.php
```diff
@@ -10,13 +10,16 @@
...
 
-$pattern = "~(cpm|both)\/(hops64|hops|x86_64) (?'hardware'.*)\sCopyright~";
-preg_match($pattern, $poll_device['sysDescr'], $matches);
-
-if ($matches['hardware']) {
-    $hardware = $matches['hardware'];
+$props = snmpwalk_cache_numerical_oid($device, 'tmnxHwEntry.7', $props = array(), 'TIMETRA-CHASSIS-MIB', 'aos', '-OQne');
```
laf: How big is this snmpwalk?

includes/discovery/sensors/state/timos.inc.php
peelman: This can probably be refactored; it was being used to verify what was being grabbed, but then I ended up reusing the table for the population.  We can probably replace it with a direct `snmp_get` in the loop below.

includes/discovery/sensors/state/timos.inc.php
peelman: good call.

includes/discovery/sensors/temperature/timos.inc.php
peelman: This is another one of those where its in so many things and in the examples, its hard to know what is necessary and what isn't.

includes/discovery/sensors/temperature/timos.inc.php
peelman: Quite possibly; but the context and use cases of the derivatives aren't made very clear.

includes/polling/os/timos.inc.php
peelman: To fetch the actual chassis model and type. without resorting to a regex of the sysDescr (which wasn't working on v9 of SROS, but was on v8)  

I almost used `tmnxChassisTypeDescription` to populate `$features` but decided against it.  

On a SAS-K the walk returns two nodes:

```
snmpwalk -M /usr/share/snmp/mibs/:Code/public_projects/librenms/mibs:Code/SNMP-mibs/nokia -m +TIMETRA-PORT-MIB -v2c -c public -OQun 10.1.86.31 tmnxChassisTypeName
.1.3.6.1.4.1.6527.3.1.2.2.1.6.1.2.20 = "7210 SAS-K 2F2T1C-1"
.1.3.6.1.4.1.6527.3.1.2.2.1.6.1.2.36 = "7210 VC"
```

If they ever get our SAS-T's powered, I'll test against those too; but I don't expect them to  return a lot of additional stuff, based on the documentation in the MIB.

The .20 and .36 endpoints are not readily identifiable through another direct `get` that I can find, so a narrow walk and reading the first object in the array seemed like the best past forward.

includes/polling/os/timos.inc.php
peelman: Relatively small, but will vary depending on the device.  On a SAS-K:

```
snmpwalk -M /usr/share/snmp/mibs/:Code/public_projects/librenms/mibs:Code/SNMP-mibs/nokia -m +TIMETRA-CHASSIS-MIB -v2c -c public -OQun 10.1.86.31 tmnxHwEntry.7
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.50331649 = physChassis
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.83886081 = powerSupply
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.134217729 = ioModule
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.150994977 = cpmModule
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.167772162 = fabricModule
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.184549633 = mdaModule
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.201327105 = flashDiskModule
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.201327106 = flashDiskModule
```
Again, I can find no way to otherwise identify those IDs and I don't yet have a broad enough pool of devices to test against to see if there is any significant consistency to them.  So walking them and looking for the one identified as "3" (PhysicalChassis/Chassis, depends on the MIB version which is returned) was the best solution I could concoct.

includes/discovery/sensors/state/timos.inc.php
peelman: @laf while looking at this this morning, wouldn't it really be more efficient to just grab the table rather than issuing X number of individual `snmp_get`'s on each device?

includes/discovery/sensors/temperature/timos.inc.php
peelman: I think I got it, after a little tinkering and poking around at some of the other includes.  It does result in a much cleaner, simpler bit of code.  👍

peelman:
What say you now @laf?

laf:
The two biggest things for me in this is us doing snmpwalks in polling for two simple items. What data do they provide back?

peelman:
#### `$chassis_type_name_array`

```
08:35:49 npeelman@npeelman-mini:[~] > snmpwalk -M /usr/share/snmp/mibs/:Code/public_projects/librenms/mibs:Code/SNMP-mibs/nokia -m +TIMETRA-PORT-MIB -v2c -c hobbit7 -OQun 10.1.86.31 tmnxChassisTypeName
.1.3.6.1.4.1.6527.3.1.2.2.1.6.1.2.20 = "7210 SAS-K 2F2T1C-1"
.1.3.6.1.4.1.6527.3.1.2.2.1.6.1.2.36 = "7210 VC"
```

From the MIB:

```        
tmnxChassisTypeName     OBJECT-TYPE
    SYNTAX      TNamedItemOrEmpty
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "The administrative name that identifies this type of Alcatel
         7x50 SR series chassis model.  This name string may be used in 
         CLI commands to specify a particular chassis model type."
    ::= { tmnxChassisTypeEntry 2 }``

They do not specify what comes after the .2; so there isn't a way I am comfortable with saying that it will always be `.20` as the endpoint containing the chassis type.

#### `$props`

```
08:37:44 npeelman@npeelman-mini:[~] > snmpwalk -M /usr/share/snmp/mibs/:Code/public_projects/librenms/mibs:Code/SNMP-mibs/nokia -m +TIMETRA-PORT-MIB -v2c -c hobbit7 -OQun 10.1.86.31 tmnxHwEntry.7
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.50331649 = physChassis
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.83886081 = powerSupply
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.134217729 = ioModule
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.150994977 = cpmModule
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.167772162 = fabricModule
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.184549633 = mdaModule
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.201327105 = flashDiskModule
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.7.1.201327106 = flashDiskModule
```

I have walked the entire tree offered by our devices (SAS-K and SAS-T.  There is no simpler, faster, or less intensive way to find `50331649`; if I could guarantee that `50331649` would always be the physChassis node, I would just skip this and call it directly.  And if that's the path you want to take, I will make that change and any bugs discovered down the line can be dealt with on a case-by-case basis.  But this seemed to be the best compromise between longevity, simplicity, and speed.

**Comment by laf** [comment]
What else is in `1.3.6.1.4.1.6527.3.1.2.2.1.8.1.5.1`?

**Comment by peelman** [comment]
It returns a table with serial numbers for all components (including empty strings for those without serial numbers).

```
09:27:27 npeelman@npeelman-mini:[~] > ] > snmpwalk -M /usr/share/snmp/mibs/:Code/public_prts/librenms/mibs:Code/SNMP-mibs/nokia -m +TIMETRA-PORT-MIB -v2c -c hobbit7 -OQun 10.1.86.31 .1.3.6.1.4.1.6527.3.1.2.2.1.8.1.5.1
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.5.1.50331649 = NS1530C0361
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.5.1.83886081 = 
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.5.1.134217729 = NS1530C0361
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.5.1.150994977 = NS1530C0361
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.5.1.167772162 = 
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.5.1.184549633 = NS1530C0361
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.5.1.201327105 = 
.1.3.6.1.4.1.6527.3.1.2.2.1.8.1.5.1.201327106 = 
```

laf:
I'm not 100% happy about merging this but I guess we'll just see if it has any impact for people.

Polling should be as lite as possible so walking unnecessary trees and even single OIDs should always be avoided.

laf:
Thanks for contributing the changes though :)

peelman:
I don't disagree. Were it up to me, setting the properties of the device like `$hardware`, `$features`, `$version`, etc. would be done in Discovery anyway, not at every polling interval.

laf:
@peelman That's certainly something that can be moved and I'd be happy about it. We would have to check that nothing else relied on it but it shouldn't.

Things like that change so infrequently anyway - @murrant?

peelman:
Would it be as simple as moving it from `polling/os` to `discovery/os`?  Or are there other moving pieces there?

laf:
It's not as simple as that no :(

Some OS polling does other things including basing the checks on group which won't have been set first time around.

peelman:
Well crap.

PR #6660, opened 2017-05-17
fix: Moved get_client_ip() to common

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes:

```
./scripts/auth_test.php -u username -d -v
SQL[SET NAMES 'utf8']
SQL[SET CHARACTER SET 'utf8']
SQL[SET COLLATION_CONNECTION = 'utf8_unicode_ci']
SQL[SELECT config_name,config_valueFROMconfig]
SQL[select * from graph_types]
SQL[DELETE FROM session WHERE session_expiry < '1494936175']
PHP Fatal error: Uncaught Error: Call to undefined function get_client_ip() in /opt/librenms/html/includes/authenticate.inc.php:83
Stack trace:
```

### DIFF
--- a/html/includes/functions.inc.php
+++ b/html/includes/functions.inc.php
@@ -1013,18 +1013,6 @@ function demo_account()
     print_error("You are logged in as a demo account, this page isn't accessible to you");
 }//end demo_account()
 
-
-function get_client_ip()
-{
-    if (isset($_SERVER['HTTP_X_FORWARDED_FOR'])) {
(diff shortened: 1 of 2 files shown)


laf:
Not sure what you mean by case but I moved this function to common.php so that we didn't have to add web to the init_modules array as that just includes unnecessary bloat.

laf:
Fixed travis.

murrant:
NACK on this, it is bloating the main application to save a bit of memory on a testing script.  I'll fix the script.

PR #6671, opened 2017-05-17
Quick fix for new Bind poller in agent mode

slashdoom:
$name and $app_id were not defined when using new bind poller in agent mode leading to NaN/U RRD data.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/applications/bind.inc.php
```diff
@@ -2,13 +2,11 @@
 
 use LibreNMS\RRD\RrdDefinition;
 
-echo ' bind ';
-
+$name = 'bind';
```
murrant: You have removed the echo.  I suggest this:
`echo " $name";`

slashdoom: Thanks, I didn't realize that was preferred as it seems kind of inconsistent in these inc files.  But I've added back in as suggested.

includes/polling/applications/bind.inc.php
```diff
@@ -2,20 +2,20 @@
...
-    $name = 'bind';
-    $app_id = $app['app_id'];
     $options = '-O qv';
     $mib          = 'NET-SNMP-EXTEND-MIB';
     $oid     = 'nsExtendOutputFull.4.98.105.110.100';
     $bind    = snmp_get($device, $oid, $options, $mib);
 }
 update_application($app, $bind);
 
+echo " $name";
```
laf: Pop this back up the top, it needs to be echoed before we do anything for debugging.

slashdoom: Ha, did I get it this time?  I used the apache, nginx and mysql files as examples so I'm a bit confused still.

PR #6672, opened 2017-05-17
fix: Two-Factor Authentication

murrant:
Moved library to a class to take advantage of namespacing and auto loading.
Update the two factor code to use the AuthenticationException for error messages.
Fix remember me to work with 2fa.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


novakin:
Unfortunately it does not work. 

Since I have applied the patch, the two factor auth is broken. After entering the login and password, it just redirects to a blank page. Can see error 500 in the dev console.

It never asks me for the 2FA token.

murrant:
@novakin I had a bad commit in there for a bit, you might have tested before and after I fixed it :D

novakin:
@murrant Indeed. Anyway huge thanks for the quick fix.

PR #6687, opened 2017-05-19
Feature: Add elasticsearch transport and docs

pblasquez:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@pblasquez the test is failing because the test is a little to simplistic.  You can either not use librenms or we can fix the test.  Sorry about that :/

laf:
@murrant 

```
diff --git a/tests/DBSetupTest.php b/tests/DBSetupTest.php
index fed6c2f..f5ebf05 100644
--- a/tests/DBSetupTest.php
+++ b/tests/DBSetupTest.php
@@ -60,7 +60,7 @@ class DBSetupTest extends DBTestCase
                 }

                 // cannot assume user use the librenms database name
-                if (str_contains($line, 'librenms')) {
+                if (str_contains($line, 'librenms.')) {
                     throw new PHPUnitException("Do not include the database name in schema files\n$file: $line");
                 }
             }

```
?

pblasquez:
@murrant np, the default value was only there for completeness.  Also the delay helped me catch that my state switch was incomplete. :)

murrant:
@laf yeah, that would be good, but possibly `librenms`.

pblasquez:
@murrant: sorry, just noticed I left an unused variable that isn't in the SQL config, might cause issues on the UI.

pblasquez:
@murrant @laf I've had this working stable in production for 6 days now, anything else needed for merge?

laf:
@pblasquez all good with me except you need to rename the schema file to 193.sql.

PR #6694, opened 2017-05-20
append Avaya VSP HW revision to serial number

ospfbgp:
The old way added the HW revision to the sysDescr.  When managing a bunch of switches with different versions this made it look like you had a lot of different kinds of switches.  These are all the switches.  Just add it to the serial number so you can have the hw version for reference if needed.  Just makes more sense not to appended to sysDescr.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
I don't think appending HW revision to serial makes sense. Seems like you should just get rid of HW revision full stop.

ospfbgp:
I agree I think the best option right now is to remove hardware revision.   Code has been fixed to remove the hardware revision.

laf:
Just the whitespace that needs fixing now:

https://travis-ci.org/librenms/librenms/jobs/234566100

ospfbgp:
this should be complete.

PR #6698, opened 2017-05-21
fix: rrdtool web settings

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

to fully implement web settings (or db config) the db configs should overwrite the global ones (from files). 

discovered this issue when tried to change the rrdtool bin from web but it was still using the defaults.

### DIFF
--- a/includes/mergecnf.inc.php
+++ b/includes/mergecnf.inc.php
@@ -37,7 +37,7 @@ function mergedb()
     foreach (dbFetchRows('SELECT `config_name`,`config_value` FROM `config`') as $obj) {
         assign_array_by_path($db_config, $obj['config_name'], $obj['config_value']);
     }
-    $config = array_replace_recursive($config, $db_config);
+    $config = array_replace_recursive($db_config, $config);
 }
 
 /**
--- a/includes/mergecnf.inc.php
+++ b/includes/mergecnf.inc.php
@@ -27,7 +27,7 @@
 
 /**
  * merge the database config with the global config
- * DB Config overwrites the files (to fully implement web settings)
+ * Global config overrides db
  */
 function mergedb()
 {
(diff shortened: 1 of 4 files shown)


laf:
I'm against this 100%. Breaking anything is best avoided but when this can break rrdtool then that's a pretty bad thing to do.

v2 is aiming to prioritise db config over config.php but I don't believe we should change v1, it's been that way since db config was added and I don't anyone else yet caught out by this.

The best option would be to just make an ajax call for the config options and test if they exist in config.php

crcro:
@laf ... i run this and doesn't break anything. without it settings the rrdtool path in web is useless, being overwriting always by default.inc.php or config.php (and then the docs should be changed as they say the preferred way to set rrdtool path is via web not via config.php).

laf:
That's different, the fix for rrdtool being in includes/defaults.inc.php is to remove it.

Changing the preference of db config over config.php is not something that should happen imho.

crcro:
@laf ... ok, reverted the config.php over db, and removed the rrdtool from defaults.inc.php

includes/mergecnf.inc.php
```diff
@@ -27,7 +27,7 @@
 
 /**
  * merge the database config with the global config
- * Global config overrides db
+ * DB Config overwrites the files (to fully implement web settings)
```
laf: This needs reverting as the original text still remains true.

murrant:
@crcro if you want to set a setting in the webui, either don't set it in config.php or use `unset()` to unset the default.

crcro:
@murrant, @laf how would be the best approach to make the webui rrrdtool path setting functional?

laf:
I've pushed an update to add a safeguard. @murrant I'm fine with this now, how about you?

Issue #6697, opened 2017-05-21
http_auth_guest broken?


KristijanL:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
==========================================================
Component | Version
--------- | -------
LibreNMS  | db3a0c26232494c7f3efc13ac71d801af7ba7898
DB Schema | 192
PHP       | 7.0.18
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
==========================================================
- [ ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ ] Please provide as much detail as possible.
- [ ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

after an updated in the last month i am no longer an admin when i login to librenms using this config:
$config['auth_mechanism'] = "http-auth";
$config['http_auth_guest'] = "admin";

select * from users;
| user_id | username  | password  | realname  | email | descr | level | can_modify_passwd | created_at          |
|  1 | admin | x | | x@gmail.com |  | 10 |  1 | 0000-00-00 00:00:00 |

Best Regards,
Kristijan

laf:
Please test https://github.com/librenms/librenms/pull/6699

PR #6699, opened 2017-05-21
fix: Fixed http-auth not honouring http_auth_guest

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6697 

Users don't have to exist in the db for http-auth, when they don't we fall back to checking the http_auth_guest variable as that's defined as a fall back.

### DIFF
--- a/html/includes/authenticate.inc.php
+++ b/html/includes/authenticate.inc.php
@@ -60,8 +60,6 @@ try {
             } elseif ($config['auth_mechanism'] === 'http-auth') {
                 if (isset($_SERVER['PHP_AUTH_USER']) && $_SERVER['PHP_AUTH_USER']) {
                     $username = clean($_SERVER['PHP_AUTH_USER']);
-                } elseif (isset($config['http_auth_guest'])) {
-                    $username = $config['http_auth_guest'];
                 }
             }
 
--- a/html/includes/authenticate.inc.php
+++ b/html/includes/authenticate.inc.php
@@ -56,9 +56,13 @@ try {
                 $username = clean($_REQUEST['username']);
                 $password = $_REQUEST['password'];
             } elseif (isset($_SERVER['REMOTE_USER'])) {
-                $username = $_SERVER['REMOTE_USER'];
-            } elseif (isset($_SERVER['PHP_AUTH_USER']) && $config['auth_mechanism'] === 'http-auth') {
-                $username = $_SERVER['PHP_AUTH_USER'];
+                $username = clean($_SERVER['REMOTE_USER']);
+            } elseif ($config['auth_mechanism'] === 'http-auth') {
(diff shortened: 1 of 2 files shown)


KristijanL:
after applying the patch i am logged in as admin, but i don't see the admin default Dashboard, if i switch back to mysql user management and log in as admin i see my default Dashboard.

html/includes/authentication/http-auth.inc.php
```diff
@@ -80,7 +80,12 @@ function user_exists($username)
 
 function get_userlevel($username)
 {
-    return dbFetchCell('SELECT `level` FROM `users` WHERE `username`= ?', array($username));
+    global $config;
+    $level = dbFetchCell('SELECT `level` FROM `users` WHERE `username`= ?', array($username));
+    if (empty($level)) {
+        $level = dbFetchCell('SELECT `level` FROM `users` WHERE `username`= ?', array($config['http_auth_guest']));
```
murrant: You could just use a single SQL query...

murrant:
@laf I applied your fix throught http-auth.  I also made sure `$username` is set properly in authenticate.inc.php.

Honestly, we could move setting `$username` into authenticate()

html/includes/authenticate.inc.php
```diff
@@ -56,9 +56,13 @@
...
                 $password = $_REQUEST['password'];
             } elseif (isset($_SERVER['REMOTE_USER'])) {
-                $username = $_SERVER['REMOTE_USER'];
-            } elseif (isset($_SERVER['PHP_AUTH_USER']) && $config['auth_mechanism'] === 'http-auth') {
-                $username = $_SERVER['PHP_AUTH_USER'];
+                $username = clean($_SERVER['REMOTE_USER']);
+            } elseif ($config['auth_mechanism'] === 'http-auth') {
+                if (isset($_SERVER['PHP_AUTH_USER']) && $_SERVER['PHP_AUTH_USER']) {
+                    $username = clean($_SERVER['PHP_AUTH_USER']);
+                } elseif (isset($config['http_auth_guest'])) {
```
laf: Would this elseif check even trigger. PHP_AUTH_USER should always contain someone who is authenticated.

murrant: Yeah, I wasn't sure and honestly it doesn't matter.  Inside the http-auth class we check things properly.

I just wanted to be sure that $username is set because of the `isset($username)` before authenticate().

If you are confident PHP_AUTH_USER will be set 100% we can drop the entire elseif.  Perhaps in fact if PHP_AUTH_USER is not set we shouldn't allow guest?

laf: Yeah if PHP_AUTH_USER isn't set then the user shouldn't have authed so we don't want to assign access/privs.

laf:
@murrant I've removed the elseif.

If you're happy then merge otherwise I'll do it in the morning.

PR #6701, opened 2017-05-22
newdevice: Added support for Ceragon FA2000 C/D

tamikkelsen:
Added support for Ceragon radios. FibeAir 2000C and FibeAir 2000D.

Currently supports TX/RX Rate, and Power/Signal, Frequency and Distance.
Also reports Signal on Overview page.

Not currently polling correct dBm for HSB for the new FA2000D radios. (Shows -120dBm, will fix that at a later time, no time to do it just yet.)

Tested and works on FA2000C Integrated, Connectorized and FA2000D Integrated, Connectorized radios.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
RX and TX power can be done in the LibreNMS/OS/Fibeair.php file (am I right @murrant ).

The other ones we should try and get those into wireless code rather than going down the custom MIB route.

tamikkelsen:
Having some GIT issues here. Will consider the proposed changes.

PR #6705, opened 2017-05-22
fix: take another crack at fixing uptime

murrant:
It looks like we weren't using sysUpTime before???
Simplify the code and just max all the time values.
Add bad_hrSystemUptime (I have another os that needs this)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
We could also detect and record counter rolls, but that could be a bit tricky

PR #6707, opened 2017-05-22
make login form more mobile-friendly

peelman:
Disable autocomplete, autocorrect, autocapitalize, and spellcheck on username and password input forms.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/logon.inc.php
+++ b/html/pages/logon.inc.php
@@ -16,12 +16,12 @@ if ($config['twofactor'] && isset($twofactorform)) {
                 <form class="form-horizontal" role="form" action="" method="post" name="logonform">
                     <div class="form-group">
                         <div class="col-md-12">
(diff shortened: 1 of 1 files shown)


laf:
What's wrong with autocomplete on the username field?

Also, does this stop browser saved logins working at all?

peelman:
In my case, on iOS, the input field defaults to capitalizing the first character.  So you end up with `Npeelman` instead of `npeelman`.  If you have a username that is spelled like a common word, it may autocorrect it every time (which mine will do occasionally on systems where my username is just 'peelman').

AFAIK, it doesn't affect saved browser logins.  It only affects how browsers treat the keyboard entry on those inputs.

laf:
I'm fine with all the others (like auto capital, etc), just wondering if the actual auto-complete should be left on for usernames?

murrant:
Also, I don't think those attributes are needed when `type="password"` so you can remove them all there.

laf:
murrant summed up what I was trying to say :)

murrant:
Thanks!

Issue #2124, opened 2015-10-13
updated to lastest commit, the dashboard can't work


alex1528:
see image:
![1](https://cloud.githubusercontent.com/assets/12593499/10443127/59194d4c-718c-11e5-9021-bbf6eb4396a1.png)
Page load for a long long time....

thanks

alex1528:
sorry! ignore this issue
because the google's jsapi can't request from China!

PR #6718, opened 2017-05-24
fix: dashboard slowness with offline browser

slashdoom:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


slashdoom:
Not sure if this a good idea or not...

In my environment, the browser I use to access LibreNMS doesn't have internet access. This makes the dashboard page load extremely slow. It seems like only the globe widget needs this Google JS file which I don't use so seems like a good speed fix.

Similar to issue #2124.

PR #6734, opened 2017-05-26
Add Current Connections Graph for Cisco WSA

network-guy:
Add current connections graph for AsyncOS, but only for the web filter appliances. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/asyncos.inc.php
```diff
@@ -20,3 +20,22 @@
 
 preg_match('/[[\w]+-[\w]+/', $serial, $regexp_results);
 $serial = $regexp_results[0];
+
+$sysobjectid = snmp_get($device, 'sysObjectID.0', '-OQnv', 'SNMPv2-MIB');
```
laf: This is already available in`$poll_device['sysObjectID']`

includes/polling/os/asyncos.inc.php
```diff
@@ -20,3 +20,22 @@
...
+
+# Get stats only if device is web proxy
+if (strcmp($sysobjectid, '.1.3.6.1.4.1.15497.1.2') == 0) {
+    $connections = snmp_get($device, 'tcpCurrEstab.0', '-OQv', 'TCP-MIB');
+
+    if (is_numeric($connections)) {
+        $rrd_def = 'DS:connections:GAUGE:2000:0:U';
```
laf: We do this slightly different now, please check any other poller file for examples.

PR #6754, opened 2017-05-30
Added SQL procedure for fixing validation errors.

cron410:
If you see any issues here, we will need to work together to fix them.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Support/Install Validation.md
+++ b/doc/Support/Install Validation.md
@@ -31,3 +31,10 @@ OK - This is a good thing, you can skip over these :)
 WARN - You probably want to check this out.
 
 FAIL - This is going to need your attention!
+
+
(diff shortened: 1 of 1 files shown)


laf:
We've actually avoided telling people to run these queries themselves or automating to prevent the code recommending something it shouldn't. I'm happy with it though - @murrant ?

cron410:
Maybe we should add a huge DANGER DRAGONS AHEAD disclaimer?

murrant:
I'd rather add a way to apply the fixes after user input rather than docs instructing them to run the sql.

murrant:
I think it is good to hear the feedback and the reports should die down over time.

murrant:
We could probably do something here, but I don't think the documentation is the right place.  Also, perhaps not the right time yet.  Closing for now.

PR #6757, opened 2017-05-30
Irc host authentication

Olen:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This patch will allow you to add users/hosts to be pre-authenticated.
$config['irc_auth'][username1][] = "*!*ident@*host.tld"

### DIFF
--- a/LibreNMS/IRCBot.php
+++ b/LibreNMS/IRCBot.php
@@ -319,29 +319,8 @@ class IRCBot
         $this->command = str_replace(':.', '', $this->command);
         $tmp           = explode(':.'.$this->command.' ', $this->data);
         $this->user    = $this->getAuthdUser();
-        if (isset($this->config['irc_auth'])) {
-            foreach ($this->config['irc_auth'] as $nms_user => $hosts) {
-                foreach ($hosts as $host) {
-                    $host = preg_replace("/\*/", ".*", $host);
-                    if (preg_match("/$host/", $this->getUserHost($this->data))) {
-                        $user_id = get_userid(mres($nms_user));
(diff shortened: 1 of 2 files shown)


deutor:
Actually it pre-authenticates you every time you send a command. 
I think that either moving that feature to _auth (so .auth admin authenticates you without token) or  pre-authenticate only if not $this->isAuthd() would save some CPU ticks, though those "wasted" ticks do not really matter nowadays (pre-authentication on first command if user is not authorized would still require changing _auth so one could refresh user['devices'] and user['ports'] with pre-authentication). 
Overall I like this change as it will save me some work required to authenticate to 20 bots :)

Anyway $this->user['expire']  should honor  config['irc_authtime'] setting so it should be:
`$this->user['expire'] = (time() + ($this->config['irc_authtime'] * 3600));`
instead of:
`$this->user['expire'] = (time() + 3600);`

Olen:
I thought of using _auth, but that means you still need to type ".auth" at leas once, which was something I wanted to avoid.  But I agree that if we move the check to _auth, it makes sense to honor irc_authtime as well.
But I agreee that checking isAuthd() first is a good thing.  I will add that.  I will look into checking devices and ports.
I will also add the same test to external commands.  This might lead to some refactoring of the code.

laf:
@deutor if you're happy with these changes we'll merge in.

deutor:
I'm very happy with these changes :)

laf:
Needs docs adding before I forget: http://docs.librenms.org/Extensions/IRC-Bot/#configuration-defaults

PR #6758, opened 2017-05-30
Irc templates

Olen:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This will change the behaviour of the irc bot, so it will display info based on the alert-temaplates.
Be aware that this might lead to excessive flooding if the templates are big.
Use {if transport == irc} in the templates to format them for irc

### DIFF
--- a/LibreNMS/IRCBot.php
+++ b/LibreNMS/IRCBot.php
@@ -253,10 +253,19 @@ class IRCBot
 
             if ($this->config['irc_alert_chan']) {
                 foreach ($this->config['irc_alert_chan'] as $chan) {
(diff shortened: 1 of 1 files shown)


laf:
@Olen you should just remove the old code you commented out, git maintains history so we can always go back to it if needs be.

I'd also say we should notify people of this change, we have quite a few users of irc alerting.  Please edit https://github.com/librenms/librenms.github.io/blob/master/notifications.rss (you can see previous changes to that file to see what to change). Set the date one week in advance and we will merge this on that date.

laf:
Needs merging on the 8th of June.

PR #6763, opened 2017-05-31
C.H.I.P. power monitor

drid:
Power readings for C.H.I.P
Adds soltage and current readings for battery, ACIN and VBUS.
Adds battery charge percentage.
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Why aren't the new files just in the linux file? The issue with doing it the way you have is that if we add an os called chip then it will conflict.

laf:
Last thing for me, if `$index` conflicts with previous code then the sensors won't be discovered. As this is using numerical OIDs then use double index for the discover_sensor() code (`116.2`) as an example.

Issue #4604, opened 2016-09-26
LibreNMS viewing issue regardless of browser


allano337:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ ] Please include all of the information between the `====================================` section of `./validate.php`.
- [ ] If you would like us to add a new device then please provide the information asked for [here](http://docs.librenms.org/Support/FAQ/#faq20)
- [x] Please provide as much detail as possible.
in full screen the lists bar or any other is not showing, if you reduce viewable size you see it once the LibreNMS logo is removed. see attached. On Version	8b68429 - Sun Sep 25 2016 08:01:58 GMT-0700
![librenms header](https://cloud.githubusercontent.com/assets/20955418/18847439/2051e7ec-83df-11e6-9619-cf88477d63b5.jpg)
![librenms header not full](https://cloud.githubusercontent.com/assets/20955418/18847445/237f41c6-83df-11e6-93a6-a3ebd8c56020.jpg)

allano337:
to further clarify the issue, opening the web browser the links etc. can't be seen unless the screen is reduced, then you see what's being hidden. Thus first screen shot and then the second, you'll see that the LibreNMS logo is gone and everything then fits in the row and you see what should be there.

Seems like something was added to the top row and it moved user settings, global setting and the search bar down and made the header wider

crcro:
i think that then it was a PR with changes from glyphicons to fontawsome and i think they have different padding/spacing which make your resolution not big enough to fit the entire top menu

vitalisator:
same issue with 1366x768 display on firefox (ver. 48.0) browser.

laf:
Is your install up to date?

vitalisator:
yes,
==========================================================
LibreNMS Version: daf2cf7a243bde3920e4d431e32921c8a31476d3
DB Schema: 142
PHP: 5.6.24-0+deb8u1
MySQL: 5.5.52-0+deb8u1
RRDTool: 1.4.8
SNMP: NET-SNMP 5.7.2.1
==========================================================

vitalisator:
can confirm that the issue is gone after my browser cache is cleared.
@allano337, please update to latest version and then clear your browser cache.

setiseta:
I think this ist still present.
[...]

PR #6770, opened 2017-06-02
Fix Navibar Linebreaks

setiseta:
On width between 1199px and 1280px there is a navibar linebreak, and on
some lists there are controlles hidden.
This is related to the closed ( but in my case unfixed) issue #4604


setiseta:
Actually it removes the brand on the given width, but it should also be possible to shrink it to remove the linebreak.
tell me what you favor.

PR #6772, opened 2017-06-02
Add service alert rules

florianbeer:
https://community.librenms.org/t/configuring-alerting-via-nagios-plugins/1389/5?u=florianbeer

- [x] followed [code guidelines](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers


laf:
Thanks @florianbeer - just had to correct an extra `"` but all good aside from that.

PR #6774, opened 2017-06-03
TP Link ER-6020

czilian:
This small branch VPN router has very poor SNMP support. However, its standard switch attributes can be recognised. To give it at least the right OS logo suggesting to have it recognised with the existing TP Link OS. Since it does not give back a sysObjectId, included the sysDescr in the existing Yaml.
Then also tweaked the logo a bit by adding the 'TP Link' underneath the graphics. Should help because not everyone can recognise brands just by graphics.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
You need to also add the sysObjectID to the test file. Can you not use the sysObjectID for the detection, we much prefer that than sysDescr checks.

laf:
@czilian Any further update?

laf:
Feel free to comment back with more information and we can re-open this.

Issue #6372, opened 2017-04-06
Cisco 3650 NX-OS spiky graphs since ./daily.sh update  on 2017-03-23 00:07


runekausland:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Since the release my server updated to by daily.sh on 2017-03-23 00:07 I am having issues with spikes and incorrect graphs in my graphs on IOS-XE Cisco 3650 switches. Visually it looks like they are using 32-bit counters on all interfaces, but I do not think they are. SNMPv1 is disabled on the switches. Intermapper SNMP polls the same switches without issues. I do not see this issue on my IOS-XE Cisco 4500x switches. No interfaces show more than 100Mbps load. Overnight from 3.5Gbps load to a spiky 64Mbs average.

Component | Version
--------- | -------
LibreNMS  | 699e99e7954948f869f48018e8fa7ccef2c3c0c7
DB Schema | 184
PHP       | 7.0.17
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2

runekausland:
![graph](https://cloud.githubusercontent.com/assets/21310408/24748583/6587d2f4-1ac0-11e7-8f87-b8e38e6068dc.png)

mrhoribu:
[...]

Issue #6257, opened 2017-03-23
ifSpeed missing on device that doesn't have ifHighSpeed


murrant:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

./poller.php -h HOSTNAME t -m ports -d
https://p.libren.ms/view/0cc810bf

snmpwalk:
https://p.libren.ms/view/c9f4ee69

laf:
We could add ifSpeed to be obtained as well, it would marginally increase poll times.

Someone else has reported an issue which because the first index is missing for ifHCInOctets then we skip to non-hc. Things like this and that strike me as where we should say please speak to the vendor to resolve this - surely if you return data for ifXEntry you should return it all.

An alternative would be a more intelligent system that overtime works out what stats a device has and we start to only poll for that.

murrant:
Learning poller, interesting idea.  Yeah, boo for broken snmp, but that is our reality.

PR #6777, opened 2017-06-03
fix: Check for ifHCInOctets and ifHighSpeed before falling back to if…

laf:
…Entry

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6372

### DIFF
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -199,7 +199,8 @@ if ($device['os'] === 'f5' && (version_compare($device['version'], '11.2.0', '>=
             $port_stats = snmpwalk_cache_oid($device, 'ifXEntry', $port_stats, 'IF-MIB');
         }
         $hc_test = array_slice($port_stats, 0, 1);
-        if (!isset($hc_test[0]['ifHCInOctets']) && !is_numeric($hc_test[0]['ifHCInOctets'])) {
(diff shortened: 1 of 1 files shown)


murrant:
#6257

laf:
Anything specific from that issue? This won't fix the issue you see there though.

murrant:
It seems like it will.  Sorry, the title is misleading on that issue.

Basically, the device doesn't have ifHighSpeed, but ifSpeed was never walked by LibreNMS.

laf:
So his fixes if we don't have both ifHSpeed and ifHCInOctets so if that's the case then yes it will.

includes/polling/ports.inc.php
```diff
@@ -199,7 +199,8 @@
             $port_stats = snmpwalk_cache_oid($device, 'ifXEntry', $port_stats, 'IF-MIB');
         }
         $hc_test = array_slice($port_stats, 0, 1);
-        if (!isset($hc_test[0]['ifHCInOctets']) && !is_numeric($hc_test[0]['ifHCInOctets'])) {
+        if ((!isset($hc_test[0]['ifHCInOctets']) && !is_numeric($hc_test[0]['ifHCInOctets'])) &&
```
murrant: Are you sure you don't want `||` here?  When you use or, it fixes devices missing ifHighSpeed...

laf: We can switch to `||`.

laf: Pushed

PR #6785, opened 2017-06-06
Syslog hooks for Oxidized integration (and more)

deepseth:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Oxidized.md
+++ b/doc/Extensions/Oxidized.md
@@ -121,3 +121,10 @@ $config['oxidized']['ignore_types'] = array('server');
 $config['oxidized']['ignore_os'] = array('linux');
 ```
 
+### Trigger configuration downloads
+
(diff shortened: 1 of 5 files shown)


deepseth:
This is super rough, and before I spend more time tidying it up, want to make sure I'm doing it inline with how it would be liked to be done, or if something needs refactoring (e.g. the oxidized function needs moving to a different file?)

Absolutely do not expect this current version to be accepted as is, as it's terribad, but it does work :-)

laf:
I think it makes more sense to try and abstract this away from being directly in the code and instead allow users to create their own hooks / search strings that are compared.

deepseth:
Example entry in config.php:
`$config['os']['ios']['syslog_hook'][] = Array('regex' => '/%SYS-(SW[0-9]+-)?5-CONFIG_I/', 'script' => '/opt/librenms/scripts/syslog-notify-oxidized.php');
`

deepseth:
Added matches for NXOS and IOSXR as well, running this in prod myself at the moment and it appears to be working well.

laf:
That looks a lot better to me. You need to add docs to cover the use and I'd also have a global `$config[]` option to enable the hooks as well as per OS.

deepseth:
I've updated the docs to cover the use, and added a global $config[] option to enable the hooks.

laf:
@deepseth LGTM.

@murrant - thoughts?

laf:
Oh and @deepseth - You need to sign the CLA.

PR #6802, opened 2017-06-08
Fix: Retrieval of hardware and version from AirOS 8.x

corny:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I don't have old AirOS devices to test against.

### DIFF
--- a/includes/polling/os/airos.inc.php
+++ b/includes/polling/os/airos.inc.php
@@ -3,7 +3,9 @@
 foreach (array(10,5) as $i) {
     $result = snmp_get_multi_oid($device, "dot11manufacturerProductName.$i dot11manufacturerProductVersion.$i", '-OQUs', 'IEEE802dot11-MIB');
 
-    if (is_array($result)) {
+    // If invalid, $result contains one empty element.
+    // So we have to verify it contains exactly two elements.
+    if (count($result) == 2) {
         $hardware = 'Ubiquiti ' . $result["dot11manufacturerProductName.$i"];
         $version  = $result["dot11manufacturerProductVersion.$i"];
         list(, $version) = preg_split('/\.v/', $version);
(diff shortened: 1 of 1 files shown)


laf:
How sure are you that the OID you've removed isn't used in older / other firmware versions?

It's probably worth converting this to multi_gets and querying both OIDs.

corny:
As I wrote I don't have any possibility to test it against old devices. Thanks for mentioning `multi_gets`. I will adjust my code to query both OIDs.

laf:
Give us a shout if you need help.

`includes/polling/os/tpdin.inc.php` has an example of use.

corny:
Thanks a lot!

corny:
```php
snmp_get_multi_oid($device,"dot11manufacturerProductName.5 dot11manufacturerProductName.10", '-OQUs', 'IEEE802dot11-MIB');
```
always returns an empty result. Now I query product name and version with `10` and then with `5`.

includes/polling/os/airos.inc.php
```diff
@@ -1,8 +1,16 @@
 <?php
 
-$hardware = 'Ubiquiti '.trim(snmp_get($device, 'dot11manufacturerProductName.5', '-Ovq', 'IEEE802dot11-MIB'));
+foreach (array(10,5) as $i) {
+    $result = snmp_get_multi_oid($device, "dot11manufacturerProductName.$i dot11manufacturerProductVersion.$i", '-OQUs', 'IEEE802dot11-MIB');
 
-$version         = trim(snmp_get($device, 'dot11manufacturerProductVersion.5', '-Ovq', 'IEEE802dot11-MIB'));
-list(, $version) = preg_split('/\.v/', $version);
+    if (is_array($result)) {
```
laf: I updated this to is_array(), please check this still works.

corny:
`$result` is always an array and it is never empty. Did I ever say I dislike PHP?
I used `print_r()` to inspect the value of `result`.

Invalid result:
```
Array
(
    [] => 
)
```
Valid result:
```
Array
(
    [dot11manufacturerProductName.10] => NanoBeam 5AC 19
    [dot11manufacturerProductVersion.10] => XC.qca955x.v8.1.4.34481.170608.1728
)
```
That means `$result` has to contain exactly two elements.

laf:
Brill, thanks for the PR and sorry I messed it up a little :)

corny:
My first attempt didn't work either. That's just PHP ^^

PR #6808, opened 2017-06-08
Emit tab field to elasticsearch to allow for data-driven URLs.

pblasquez:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/alerts/transport.elasticsearch.php
+++ b/includes/alerts/transport.elasticsearch.php
@@ -76,6 +76,7 @@ $data = array(
     "severity" => $severity,
     "first_occurrence" => $obj['timestamp'],
     "entity_type" => "device",
+    "entity_tab" => "overview",
     "entity_id" => $obj['device_id'],
     "entity_name" => $obj['hostname'],
     "entity_descr" => $obj['sysDescr'],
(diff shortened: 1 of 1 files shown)


includes/alerts/transport.elasticsearch.php
```diff
@@ -88,42 +89,49 @@
...
                 $data['entity_descr'] = $v['ifAlias'];
                 break;
             case (array_key_exists('sensor_id', $v)):
-                $data['entity_type'] = 'sensor';
+                $data['entity_type'] = $v['sensor_class'];
+                $data['entity_tab'] = 'health';
                 $data['entity_id'] = $v['sensor_id'];
-                $data['entity_name'] = $v['sensor_class'];
-                $data['entity_descr'] = $v['sensor_descr'];
+                $data['entity_name'] = $v['sensor_descr'];
```
laf: Did you mean to swap these around?

pblasquez: I did, the way sensors are arranged it makes the most sense, for two reasons:

1. It allows a single template on the far side to link back for any entity using this format: http://192.168.1.1/device/device=1/tab=<entity_tab>/<entity_type>=<entity_id>/
2. For sensors, sensor_name is typically a one word designator like 'junos', whereas sensor_descr is more what would typically be a 'name' on another entity, like '#1 Fan Speed'.  While still keeping the sensor_name value in 'entity_descr' in case it is useful.

PR #6813, opened 2017-06-09
newdevice: Improved Cisco Unified CM detection and renamed to ucos

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This relies on https://github.com/librenms/librenms.github.io/pull/90

ucos seems the better os name: http://www.cisco.com/c/en/us/td/docs/voice_ip_comm/cucm/cucos/7_1_2/cucos/osg_712_cm/iptpch2.html

Why not cucos, well we don't have cnxos or cios yet these are called Cisco NX-OS and Cisco IOS....


murrant:
@laf this looks good if you want to re-date the notification and send that out.

laf:
Done.

PR #6817, opened 2017-06-10
newdevice: Added basic detection for Extreme Wireless Controllers

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/OS/Ewc.php
+++ b/LibreNMS/OS/Ewc.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * Ewc.php
+ *
+ * Extreme Wireless Controller
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


LibreNMS/OS/Ewc.php
```diff
@@ -0,0 +1,54 @@
...
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessDistanceDiscovery;
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessFrequencyDiscovery;
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessNoiseFloorDiscovery;
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessPowerDiscovery;
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessQualityDiscovery;
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessRateDiscovery;
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessRssiDiscovery;
+use LibreNMS\OS;
+
+class Airos extends OS implements
```
murrant: @laf  Class name is wrong here, should be Ewc
Also, It does not implement any interfaces since there are no methods.

laf: It was done as a template for someone else so they could just run github-apply and carry on.

PR #6819, opened 2017-06-11
First pass at implementing ExtremeWireless support

trs80:
Includes global AP and user count, and per-AP user count

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I haven't run `./scripts/pre-commit.php` as I don't have composer set up, I mostly want to see what people and the CI bots think of this.


LibreNMS/OS/Ewc.php
```diff
@@ -0,0 +1,108 @@
...
+ * @link       http://librenms.org
+ * @copyright  2017 James Andrewartha
+ * @author     James Andrewartha <trs80@ucc.asn.au>
+ */
+
+namespace LibreNMS\OS;
+
+use LibreNMS\Device\WirelessSensor;
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessApCountDiscovery;
+#use LibreNMS\Interfaces\Discovery\Sensors\WirelessCcqDiscovery;
```
laf: You can just delete all the lines you've commented out.

LibreNMS/OS/Ewc.php
```diff
@@ -0,0 +1,108 @@
...
+    #WirelessRssiDiscovery
+{
+    /**
+     * Returns an array of LibreNMS\Device\Sensor objects
+     *
+     * @return array Sensors
+     */
+    public function discoverWirelessClients()
+    {
+        $apstats =  $this->getCacheByIndex('apStatsMuCounts', "HIPATH-WIRELESS-HWC-MIB");
```
murrant: The getCacheByIndex() function should only be used if the data is required for multiple sensors.  Just use the normal snmp functions.

LibreNMS/OS/Ewc.php
```diff
@@ -0,0 +1,108 @@
...
+        $apnames =  $this->getCacheByIndex('apName', "HIPATH-WIRELESS-HWC-MIB");
+        $sensors = array(
```
murrant: Ap count doesn't belong here, please implement `WirelessApCountDiscovery`

LibreNMS/OS/Ewc.php
```diff
@@ -0,0 +1,108 @@
...
+            new WirelessSensor(
```
murrant: This sensor discovery doesn't belong here and is in fact duplicated in the place it should be.

LibreNMS/OS/Ewc.php
```diff
@@ -0,0 +1,108 @@
...
+                $index,
+                "Clients ($name)",
+                $apStatsMuCounts
+            );
+        }
+        return $sensors;
+    }
+
+    /**
+     * Discover wireless capacity. Type is capacity.
```
murrant: This phpdoc description is for the wrong sensor.  If you aren't using something that will do it for you, copy the phpdoc from the interface file.

LibreNMS/OS/Ewc.php
```diff
@@ -0,0 +1,108 @@
...
+     * Returns an array of LibreNMS\Device\Sensor objects that have been discovered
+     *
+     * @return array Sensors
+     */
+    public function discoverWirelessApCount()
+    {
+        $apactive =  $this->getCacheByIndex('apActiveCount', "HIPATH-WIRELESS-HWC-MIB");
```
murrant: Remove this, it is not needed.  You specify a single OID when you create the sensor, so just let the discovery take care of it.

LibreNMS/OS/Ewc.php
```diff
@@ -0,0 +1,108 @@
...
+        return array(
+            new WirelessSensor(
+                'ap-count',
+                $this->getDeviceId(),
+                '.1.3.6.1.4.1.4329.15.3.5.2.1.0',
+                'ewc',
+                0,
+                'Connected APs',
+                $apactive[0]
```
murrant: Remove this.

LibreNMS/OS/Ewc.php
trs80: Yeah, I was leaving them to remind me to implement them. I'll remove before my final commit.

LibreNMS/OS/Ewc.php
trs80: What are the normal snmp functions? snmpwalk_cache_oid?

LibreNMS/OS/Ewc.php
trs80: I think I meant to change this to mobileUnitCount for the total number of associated APs.

LibreNMS/OS/Ewc.php
trs80: What about https://github.com/librenms/librenms/blob/master/LibreNMS/OS/Ciscowlc.php#L88 ? I've found the OID for license capacity so I'll be adding that.

LibreNMS/OS/Ewc.php
trs80: The interface has the same description in https://github.com/librenms/librenms/blob/master/LibreNMS/Interfaces/Discovery/Sensors/WirelessApCountDiscovery.php#L31 and I copied it from https://github.com/librenms/librenms/blob/master/LibreNMS/OS/Ciscowlc.php#L81 (although I removed the percentage bit).

LibreNMS/OS/Ewc.php
murrant: You still don't need to fetch it if all devices of this OS support that oid.

Adding max capacity, you do need to fetch that, but yeah the cisciwlc code is probably close to what you want here.

LibreNMS/OS/Ewc.php
murrant: Dooooooh, copy paste error :D. I'll fix that.

LibreNMS/OS/Ewc.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Returns an array of LibreNMS\Device\Sensor objects
+     *
+     * @return array Sensors
+     */
+    public function discoverWirelessClients()
+    {
+        $sensors = array(
+            new WirelessSensor(
+                'client',
```
murrant: The type is wrong here, should be "clients"

trs80: Cheers, I was wondering why it wasn't showing up.

trs80:
OK, I'm happy with this to be committed, I'll work on the other sensors (frequency, SNR etc.) over time.

LibreNMS/OS/Ewc.php
```diff
@@ -0,0 +1,125 @@
...
+                0,
+                'Connected APs'
+            ),
+            new WirelessSensor(
+                'ap-count',
+                $this->getDeviceId(),
+                array('1.3.6.1.4.1.4329.15.3.2.10.1.7.0', '1.3.6.1.4.1.4329.15.3.2.10.1.8.0'),
+                'ewc',
+                1,
+                'Total AP licenses'
```
murrant: Wouldn't it be better to set this as the max for Connected APs?

trs80: Yeah, I had it that way to start, but it didn't put a line on the graph (which is perhaps a more general feature request) and it's complicated anyway since the licenses are shared between two controllers in HA. Having said that I found 1.3.6.1.4.1.4329.15.3.5.1.1 which lists the total number of APs registered (including those on the HA controller) which I will add as a separate counter with the total AP licenses as the max, and remove the licenses remaining sensor as it's now redundant.

laf: @trs80 Do you need anything further from us to get this working?

trs80:
Any further thoughts?

laf:
We were just waiting on you before that last commit.

murrant:
Looks good to me.

PR #6828, opened 2017-06-13
fix: Check HC counters individually

murrant:
Vendors are very inconsistent returning ifXTable values.
Some only implement some of them (seemingly randomly).
Others return them, but they are always 0.

This attempts to fix that by checking each value individually.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -443,7 +443,7 @@ foreach ($ports as $port) {
         }
 
         if ($device['os'] === 'airos-af' && $port['ifAlias'] === 'eth0') {
-            $airos_stats = snmpwalk_cache_oid($device, '.1.3.6.1.4.1.41112.1.3.3.1', $airos_stats, 'UBNT-AirFIBER-MIB');
+            $airos_stats = snmpwalk_cache_oid($device, '.1.3.6.1.4.1.41112.1.3.3.1', array(), 'UBNT-AirFIBER-MIB');
             $this_port['ifInOctets'] = $airos_stats[1]['rxOctetsOK'];
             $this_port['ifOutOctets'] = $airos_stats[1]['txOctetsOK'];
             $this_port['ifInErrors'] = $airos_stats[1]['rxErroredFrames'];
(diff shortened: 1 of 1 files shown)


sorano:
Tested this and its looking good for my device.

includes/polling/ports.inc.php
```diff
@@ -470,16 +472,18 @@
...
 
-        if (is_numeric($this_port['ifHCInBroadcastPkts']) && is_numeric($this_port['ifHCOutBroadcastPkts']) && is_numeric($this_port['ifHCInMulticastPkts']) && is_numeric($this_port['ifHCOutMulticastPkts']) && $device['os'] !== 'ciscosb') {
-            echo 'HC ';
-            $this_port['ifInBroadcastPkts']  = $this_port['ifHCInBroadcastPkts'];
-            $this_port['ifOutBroadcastPkts'] = $this_port['ifHCOutBroadcastPkts'];
-            $this_port['ifInMulticastPkts']  = $this_port['ifHCInMulticastPkts'];
-            $this_port['ifOutMulticastPkts'] = $this_port['ifHCOutMulticastPkts'];
+        // use HC values if they are available
+        foreach ($hc_mappings as $hc_oid => $if_oid) {
+            if (isset($this_port[$hc_oid]) && $this_port[$hc_oid]) {
```
laf: Should we do is_numeric() here?

Also we excluded ciscosb before which is now dropped.

murrant: is_numeric() doesn't seem worthwhile.  `$this_port[$hc_oid] == true` (the == true is implied in the code) covers empty strings and 0 values. is_numeric() only covers empty strings.

Excluding ciscosb was because the hc oids are present, but at least some of them don't work and only return 0. When we check each stat, we can exclude 0 values because it only affects that individual stat.

Treating the HC stats as a group really bit us because of inconsistent vendor implementations.

laf:
I guess the only thing here still is we may not have looked up the non HC counters still but it's certainly a better check than before.

PR #6833, opened 2017-06-13
newdevice: Chatsworth PDU (legacy old pdus not sure model number)

EnzoZafra:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6831


includes/definitions/chatsworth.yaml
```diff
@@ -0,0 +1,8 @@
+os: chatsworth
```
laf: This probable should be chatsworth-pdu unless a better name can be found.

includes/definitions/chatsworth.yaml
```diff
@@ -0,0 +1,8 @@
...
+text: 'Chatsworth PDU'
+type: power
+icon: chatsworth
+nobulk: 1
+discovery:
+    - sysObjectId:
+        - .0.10.43.6.1.4.1
```
laf: Is that genuinely the sysObjectID, if so it's probably worth raising to them as that's just, erm, whacky.

EnzoZafra: Unfortunately, yes.. I will try to contact them about it but I'd be surprised if they actually do anything about it

PR #6836, opened 2017-06-14
Update Ubuntu 16.04 nginx install doc with a better nginx config

ndum:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Installation/Installation-CentOS-7-Nginx.md
+++ b/doc/Installation/Installation-CentOS-7-Nginx.md
@@ -101,10 +101,14 @@ server {
  index       index.php;
  access_log  /opt/librenms/logs/access_log;
  error_log   /opt/librenms/logs/error_log;
+ charset utf-8;
  gzip on;
  gzip_types text/css application/javascript text/javascript application/x-javascript image/svg+xml text/plain text/xsd text/xsl text/xml image/x-icon;
  location / {
-  try_files $uri $uri/ @librenms;
+  try_files $uri $uri/ /index.php?$query_string;
+ }
+ location /api/v0 {
+  try_files $uri $uri/ /api_v0.php?$query_string;
  }
  location ~ \.php {
   include fastcgi.conf;
@@ -113,10 +117,6 @@ server {
  }
  location ~ /\.ht {
   deny all;
- }
- location @librenms {
-  rewrite api/v0(.*)$ /api_v0.php/$1 last;
-  rewrite ^(.+)$ /index.php/$1 last;
  }
 }
 ```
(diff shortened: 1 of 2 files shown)


f0o:
Although the API rule is not inherently wrong, we chose namer location for a reason. Our API is versioned and in due time we need to support /v0 /v1 etc. This is why we had a rewrite with named location instead of matching the actual directoryand try_files

ndum:
Hi,

Please correct me, but is v0 actually not hardcoded also?

`rewrite api/v0(.*)$ /api_v0.php/$1 last;`

if u need v0 / v1 just add a second location? (as example)

```
location /api/v0 {
 try_files $uri $uri/ /api_v0.php?$query_string;
}

location /api/v1 {
 try_files $uri $uri/ /api_v1.php?$query_string;
}
```

This named location does not works in all situation, that is why i created this MR. In addition it's actually match both rules every time add this is bad for the performance.

laf:
Does your change support /api/v0 though, it looks like it will pass v0 as a parameter to api_v0.php. Maybe just be explicit with the v0 in the rule?

ndum:
> Does your change support /api/v0 though, it looks like it will pass v0 as a parameter to api_v0.php. Maybe just be explicit with the v0 in the rule?


Yes, if you send a request (example: xxx.xxx.org/api/v0/devices) the "devices" will be passed as a parameter. 

$query_string is just the last input parameter, so it will send "devices" to api_v0.php. 
example: api_v0.php?devices
`
Like i said before, if you want another API location (v1 / v2 or what ever), just add this as a new location.

laf:
Imho I think you should change it to `location /api/v0 {` now so that it's clear.

ndum:
I've changed it now to /api/v0

laf:
Brilliant thanks.

Will fix when we work out what Travis changed :/

ndum:
Perfect Thanks!

laf:
This breaks access to /api-access/ for me.

laf:
@ndum Please see my last comment.

ndum:
Hi, 
I'm looking on it, 
On 2-3 installs at my environment it works fine.Also the  /api-access/
I will check this out.

laf:
Sorry no I didn't see the change.

Works fine. Can you please update the CentOS install docs as well, should be the same changes but we need to ensure uniformed info.

laf:
Oh and please rebase to fix the travis issue.

ndum:
I have updated CentOS documentation now also.

Thanks ;)

laf:
Thanks very much :)

PR #6838, opened 2017-06-14
Fix for syslog-messages from zywall (USG series)

Olen:
I am not sure this works for ALL ZyWalls, but it seems like host os is used before, so I just assumed Zyxel have some kind of consistency in their bugs...
This patch makes nicely formatted Syslog entries from my USG60 at least.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Can you create a test (or tests) for this in `tests/SyslogTest.php` ?

Issue #6848, opened 2017-06-15
OS's with configuration in config.php no longer being discovered


deepseth:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://opencollective.com/librenms/donate

Looks like changes made in #6555 (specifically in `functions.php`) have had unexpected consequences.

Reading the contents of `$config['os'][$os]` and ignoring the contents of ./includes/definitions/$os.yaml results in $tmp['discovery'] never being set, and that OS not being detected.

Instead, we should read the file first, and then merge with what's in $config['os'][$os]. I think.

```
root@netmon-lon:/opt/librenms# ./validate.php 
==========================================================
Component | Version
--------- | -------
LibreNMS  | c66f6c9704e17c862ff691230a26df5ccfb6d469
DB Schema | 195
PHP       | 7.0.18-0ubuntu0.16.04.1
MySQL     | 10.0.29-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3
==========================================================

[WARN]  Your install is out of date, last update: Wed, 14 Jun 2017 19:32:56 +0000
[OK]    Database connection successful
[OK]    Database schema correct
[WARN]  Your local git contains modified files, this could prevent automatic updates.
Modified files:
	 doc/Extensions/Oxidized.md
	 doc/Extensions/Syslog.md
	 includes/syslog.php
	 tests/SyslogTest.php

```

```
[...]

PR #6849, opened 2017-06-15
Pull data from YAML before overriding from $config

deepseth:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes issue #6848

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -100,13 +100,12 @@ function getHostOS($device)
     // check yaml files
     $pattern = $config['install_dir'] . '/includes/definitions/*.yaml';
     foreach (glob($pattern) as $file) {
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -100,13 +100,12 @@ function getHostOS($device)
     // check yaml files
     $pattern = $config['install_dir'] . '/includes/definitions/*.yaml';
     foreach (glob($pattern) as $file) {
+        $tmp = Symfony\Component\Yaml\Yaml::parse(
```
murrant: Parsing the yaml is slow, so this is why the check was added in the first place.
My thoughts was check `$config['os'][$os]['os']` which is unlikely to be set by the user in config.php, that does seem a little hacky though.

deepseth:
@murrant has written #6850 instead.

PR #6854, opened 2017-06-16
Test template

Olen:
This patch allows you to specify "state" on the command line.
Nice for testing templates with conditionals for "up" and "down"

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
You actually can't be guaranteed that the state selected will have existed. Why not just overwrite the returned state value from the db with the one passed on the argument to ensure it's what the user expects?

Olen:
I tried that first, but the result was not as expected.  
The title was always set to the title from the latest state, which was quite confusing when there was a mismatch between the title and the content.
If the state you are testing has never happened, you will not have any real data to display, so the test would not make much sense anyway.  You would need to populate it with fake data somehow.
As the patch is now, if the state does not exist, you will get the latest state (like the current behaviour).

PR #6858, opened 2017-06-17
Change locks to use flock, instead of pid checking

keeperofdakeys:
Change locks to use flock, as pid checking is not sufficient when PID Namespaces are involved.

On my system I use pid namespaces for cron jobs. So when `discovery.php -h new` runs, it always has the pid of 3 - which prevents the pid based locks from working. Using flock instead gets around this issue.

An alternative to this would be to use mysql advisory locks (the `GET_LOCK()` function).

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/FileLock.php
+++ b/LibreNMS/FileLock.php
@@ -1,0 +1,95 @@
+<?php
+
+/**
+ * LibreNMS
+ *
+ *   This file is part of LibreNMS.
+ *
+ * @package    LibreNMS
+ * @subpackage FileLock
+ * @copyright  (C) 2017
+ *
+ */
+
+namespace LibreNMS;
+
+class FileLock
+{
+    private $name = "";
+    private $file = "";
+    /**
+     * @var resource | false
+     */
+    private $handle = false;
+    private $acquired = false;
+
+    private function __construct($lock_name)
+    {
+        global $config;
+
+        $this->name = $lock_name;
+        $this->file = "$config[install_dir]/.$lock_name.lock";
+        $this->handle = fopen($this->file, "w+");
+    }
+
+    public function __destruct()
+    {
+        $this->release();
+    }
+
+    /**
+     * Release the lock.
+     */
+    public function release()
+    {
+        if (!$this->acquired) {
+            return;
+        }
+
+        if ($this->handle !== false) {
+            flock($this->handle, LOCK_UN);
+            fclose($this->handle);
+        }
+        if (file_exists($this->file)) {
+            unlink($this->file);
+        }
+    }
+
+    /**
+     * Given a lock name, try to acquire the lock.
+     * On success return a FileLock object, or on failure return false.
+     * @param string $lock_name Name of lock
+     * @return mixed
+     */
+    public static function lock($lock_name)
+    {
+        $lock = new self($lock_name);
+        if ($lock->handle === false) {
+            return false;
+        }
+
+        if (flock($lock->handle, LOCK_EX | LOCK_NB)) {
+            $lock->acquired = true;
+            return $lock;
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * Given a lock name, try to acquire the lock, exiting on failure.
+     * On success return a FileLock object.
+     * @param string $lock_name Name of lock
+     * @return FileLock
+     */
+    public static function lockOrDie($lock_name)
+    {
+        $lock = self::lock($lock_name);
+
+        if ($lock === false) {
+            echo "Failed to acquire lock $lock_name, exiting\n";
+            exit(1);
+        }
+        return $lock;
+    }
+}
--- a/alerts.php
+++ b/alerts.php
@@ -30,7 +30,7 @@ require __DIR__ . '/includes/init.php';
 
 $options = getopt('d::');
 
-set_lock('alerts');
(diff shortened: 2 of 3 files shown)


laf:
Thanks for submitting this. Will run it on my dev install to test it out.

laf:
If you want this PR merged sooner, please rebase to include the fix for Travis breaking on PHP 5.6 tests

murrant:
Not a fan of the global $lock_files variable, but if the is no good way to avoid it, it's ok as-is.

keeperofdakeys:
Personally I don't like global variables either, but I did that to preserve the api. If you don't mind it changing, we can return an `array("filename" =>  "/path/to/blah", "handle" => $handle)`, or make it a Class (first seems simpler). Due to the nature of flock, the open handle needs to be kept somewhere.

laf:
Feel free to change how it works :)

keeperofdakeys:
Is there any good reason why `discovery.php` tries to update the sql schema? It's hard to check "is a file flocked" without obtaining the lock, which is the logic used in `discovery.php` to determine whether to run the schema updater.

So we can either rip this out of `discovery.php`, or replace it with a `file_exists` check.

laf:
discovery.php runs the schema updates so that anyone forgetting to run them if they've updated manually it will catch that. We shouldn't run in poller as poller is time sensitive.

What do you want to do with `file_exists`?

keeperofdakeys:
I was thinking we could just check if the lock file exists, if it doesn't then another schema update isn't occurring at that moment (or it had died).

After thinking about it though, we could get around this if `update.php` was refactored into a function that didn't die if it couldn't get a lock. Then `discovery.php` could just call it, and the other four invocations of `update.php` could themselves die if the update couldn't happen (ie. update_sql() returns false).

I'll start working on it and see how it goes.

murrant:
@keeperofdakeys the class looks good :)

Unfortunately, I am working on some updates that conflict with your schema update changes.  Would be happy to merge my changes in later, but I think your code needs some changes.  If you want you can roll back your schema update changes and I will merge this and update my PR to use FileLock.

Functions and code should not be mixed in the same file.
Some of your locks are acquired after the DB update process is run.

keeperofdakeys:
I accidentally pushed this a bit early. In light of what you're doing, I'll undo all my update schema changes, and just introduce the new FileLock. I'll need to leave the old locks for now so that the old schema update code will still work however.

keeperofdakeys:
Unfortunately I'm not well versed in modern PHP, and scrutinizer is giving me this error:
```It seems like fopen($this->file, 'w+') of type resource is incompatible with the declared type boolean of property $handle.```

Should this be declared as NULL instead, or is there a documentation/attribute that I need?

**Comment by keeperofdakeys** [comment]
* Fixing some linting issues on PHP7.

File: LibreNMS/FileLock.php [hunk 1]
```diff
@@ -0,0 +1,92 @@
...
+ *
+ */
+
+namespace LibreNMS;
+
+class FileLock
+{
+    private $name = "";
+    private $file = false;
+    private $handle = false;
```
murrant: You can leave these without a value (they will be null)

murrant:
You can add a phpdoc comment for the property to declare the type.
`/** $handle resource|false */`

Or something like that, on mobile right now.

You can also ignore scrutinizer.  It is there to warn you about potential issues.

laf:
Thanks @keeperofdakeys, tested and looks good to me.

murrant:
LGTM :)
```
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
SNMP: Get[0/0.00s] Walk [0/0.00s]
MySQL: Cell[1/0.00s] Row[0/-0.00s] Rows[4/0.03s] Column[0/0.00s] Update[0/0.00s] Insert[0/0.00s] Delete[0/0.00s]
LibreNMS Discovery
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
LibreNMS Discovery
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
Failed to acquire lock new-discovery, exiting
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
SNMP: Get[0/0.00s] Walk [0/0.00s]
MySQL: Cell[1/0.00s] Row[0/-0.00s] Rows[4/0.03s] Column[0/0.00s] Update[0/0.00s] Insert[0/0.00s] Delete[0/0.00s]
LibreNMS Discovery
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
LibreNMS Discovery
LibreNMS Discovery
Failed to acquire lock new-discovery, exiting
```

laf:
Thanks for contributing @keeperofdakeys :D

image

PR #6865, opened 2017-06-19
feature: New snmpwalk_group() function

murrant:
Intended to replace several other snmp functions, but may not behave exactly as they did.
Properly handles multi-line values.
Adds additional index data after the grouping so it isn't lost. Perhaps might make optional.
Might add $skip variable to support fully numeric oid indexes. Thoughts?

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/snmp.inc.php
+++ b/includes/snmp.inc.php
@@ -516,6 +516,58 @@ function snmpwalk_cache_triple_oid($device, $oid, $array, $mib = null, $mibdir =
 }//end snmpwalk_cache_triple_oid()
 
 
+/**
+ * Walk an snmp mib oid and group items together based on the index.
+ * This is intended to be used with a string based oid.
+ * Any extra index data past $depth will be added after the oidName to keep grouping consistent.
+ *
+ * Example:
+ * snmpwalk_group($device, 'ifTable', 'IF-MIB');
+ * [
+ *   1 => [ 'ifIndex' => '1', 'ifDescr' => 'lo', 'ifMtu' => '65536', ...],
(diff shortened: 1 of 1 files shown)


includes/snmp.inc.php
```diff
@@ -516,6 +516,58 @@ function snmpwalk_cache_triple_oid($device, $oid, $array, $mib = null, $mibdir =
...
+ * @param array $device Target device
+ * @param string $oid The string based oid to walk
+ * @param string $mib The MIB to use
+ * @param int $depth how many indexes to group
+ * @param array $array optionally insert the entries into an existing array (helpful for grouping multiple walks)
+ * @return array grouped array of data
+ */
+function snmpwalk_group($device, $oid, $mib = '', $depth = 1, $array = array())
+{
+    $cmd = gen_snmpwalk_cmd($device, $oid, '-OQUsb', $mib);
```
laf: Setting the flags statically can be a problem. I set `OeQUs` in the sensors yaml pr as we need `e` mainly to print the data unparsed by the mib. this is used for state translations so you get back 4 instead of noError.

murrant: The snmpwalk_cache_oid functions don't allow options.

I could allow options but I would need to filter them.

Or we could come up with a really good default and force the use of other snmpwalk if custom options are needed.

laf: It does:

`snmpwalk_cache_oid($device, $oid, $array, $mib = null, $mibdir = null, $snmpflags = '-OQUs')`

Maybe if you just add e to that default list if you don't want to accept the options.

Issue #6866, opened 2017-06-19
Secure cookies not enabled


rzig:
Hi. After running a vulnerability scan on my LibreNMS server, I was informed that it did not have secure cookies enabled, which can lead to session hijacking attacks. As of the current version of LibreNMS, secure cookies cannot be enabled in config.php or any other LibreNMS-specific configuration files. To enable secure cookies, the following needs to be added to php.ini:
```
session.cookie_secure = 1
```
This can also be set by running:
```
ini_set('session.cookie_secure', 1);
```

I think it would be good if this was an automatic behavior, and I'd be happy to work on getting an option for secure cookies into config.php if you want.

Thanks for all the hard work!

laf:
I guess you could detect if the site was access over https and enable it automatically. The alternative is as like you've said, to provide a config.php option. If you can submit a PR for this that would be ace.

N

rzig:
I'll work on that. As a temporary patch, adding `ini_set('session.cookie_secure', 1);` to config.php will fix the issue.

rzig:
OK. So, adding `ini_set('session.cookie_secure', 1);` to `config.php` only secures the session identifier cookie. I'll attach a PR that takes care of securing the other cookies, but the only way to get the session cookie secured is by modifying `config.php`, or `php.ini`.

murrant:
rzig: I reviewed `session.cookie_secure` and it appears that it does not enable secure cookies.  It just disables insecure (http) cookies. Am I wrong?

rzig:
That is correct. `session.cookie_secure` will only transmit session cookies over HTTPS, in addition to making the `PHPSESSID` cookie a "secure" cookie. In the [pull request](https://github.com/librenms/librenms/pull/6868), I made sure all cookies would be secure when `$config['secure_cookies']` is `true`.

murrant:
Perhaps.  In my environment, I have nginx configured to redirect http to https and have HSTS enabled.  HSTS is a much more robust way to ensure all your site is only accessed over https.

rzig:
I'm not sure that HSTS forces the cookies to be secure, as I believe that is a setting that can only be enabled in the code. I may be wrong, though.

murrant:
Ok, a little more research there is a bit that instructs the browser to only transmit the data over https.

I'm still wondering if there is a way to avoid having an option and just set it in the correct situations.

rzig:
[...]

PR #6868, opened 2017-06-20
Enable secure cookies

rzig:
This will enable secure cookies automatically over HTTPS and has a line in `config.default.php` that can be uncommented to enable secure session cookies.

Would you like me to add any documentation on the line in `config.default.php`?
#### Testers

fixes: #6866


config.php.default
```diff
@@ -49,3 +49,6 @@ $config['enable_billing'] = 1;
 
 # Enable the in-built services support (Nagios plugins)
 $config['show_services'] = 1;
+
+# Uncomment the following line if you're using HTTPS and want secure session cookies
+# ini_set('session.cookie_secure', 1);
```
murrant: Don't put code in this file, only config definitions.
You can move this to init.php

rzig:
The Travis CI build failed, but I don't think it is as a result of my code as the lines I added were commented out and the error related to a syslog test that appears to be a result of the previous merge.

html/includes/authentication/functions.php
```diff
@@ -142,9 +142,12 @@ function set_remember_me()
         dbInsert($db_entry, 'session');
     }
 
-    setcookie('sess_id', $sess_id, $expiration, '/', null, false, true);
-    setcookie('token', $token_id, $expiration, '/', null, false, true);
-    setcookie('auth', $auth, $expiration, '/', null, false, true);
+    // Params: name, value, expire, path, domain, secure, httponly
```
laf: you can get rid of all of these comments.

html/includes/authentication/functions.php
```diff
@@ -142,9 +142,12 @@ function set_remember_me()
...
+    // We set the secure param to $_SERVER["HTTPS"], which will return true
+    // and make it a secure cookie if HTTPS is being used
+    setcookie('sess_id', $sess_id, $expiration, '/', null, $_SERVER["HTTPS"], true);
```
laf: I'm not really sure we should just be basing it on if the user is connecting over https or not as surely that's what we have now.

Would it not be better to have the user set in config.php if the connection should be secure and use that variable here?

includes/init.php
```diff
@@ -90,6 +90,9 @@
 require $install_dir . '/includes/definitions.inc.php';
 include $install_dir . '/config.php';
 
+# Uncomment the following line if you're using HTTPS and want secure session cookies
```
laf: Please use `//` instead of `#` just to be consistent.

We should use the same config variable we would use in setcookie() to also run this ini_set() which would be better placed in `includes/process_config.inc.php`

html/includes/authentication/functions.php
rzig: Currently secure cookies are turned off regardlesss of whether the user is connecting over HTTP or HTTPS. I figured having it done automatically would be the best solution, but can change it to be in `config.php` if you want.

murrant:
I've pushed a fix for the travis failures, you'll have to merge it to fix your build failure.

rzig:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Forgot to include this earlier, sorry.

config.php.default
```diff
@@ -49,3 +49,6 @@ $config['enable_billing'] = 1;
 
 # Enable the in-built services support (Nagios plugins)
 $config['show_services'] = 1;
+
+# Whether to use secure cookies. Should be true if you're using HTTPS
+$config['secure_cookies'] = false;
```
murrant: Lets move this to `includes/defaults.inc.php` and then this looks good to go.

Thanks for discussing in depth :)

rzig:
There's a bit of an issue. When I try and apply the patch it says `error: patch failed: includes/process_config.inc.php:36`. I'm thinking it has something to do with whitespace, but I don't see anything out of the ordinary. Any ideas?

rzig:
The git issue is fixed - it still complains about whitespace, but it applies the patches.

laf:
LGTM

PR #6869, opened 2017-06-20
docs: Reorganised alerting docs + added some clarifications

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
I'm fine with breaking these into pages, but it  we need better navigation.  Like a link list on the left side.

laf:
I don't think it's possible to have a custom TOC per page. we might be able to fudge it with headings in each doc.

murrant:
I'll give it a look.

murrant:
My main problem with breaking things up into smaller pages is it is difficult to move between the the pages.  Like there is not go to next page.

laf:
That is a problem but is it a worse problem to have than a huge wall of text?

murrant:
Wish we had a little more control...

PR #6873, opened 2017-06-21
Added "system name" for the "Services list"

algebur:
service_01

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/services.inc.php
+++ b/includes/services.inc.php
@@ -129,7 +129,7 @@ function poll_service($service)
 
     // If we do not have a cmd from the check script, build one.
     if ($check_cmd == "") {
-        $check_cmd = $config['nagios_plugins'] . "/check_" . $service['service_type'] . " -H " . ($service['service_ip'] ? $service['service_ip'] : $service['hostname']);
+        $check_cmd = $config['nagios_plugins'] . "/check_" . $service['service_type'];
         $check_cmd .= " " . $service['service_param'];
     }
 
--- a/includes/services.inc.php
+++ b/includes/services.inc.php
@@ -129,7 +129,7 @@ function poll_service($service)
 
     // If we do not have a cmd from the check script, build one.
     if ($check_cmd == "") {
-        $check_cmd = $config['nagios_plugins'] . "/check_" . $service['service_type']; . " -H " . ($service['service_ip'] ? $service['service_ip'] : $service['hostname']);
+        $check_cmd = $config['nagios_plugins'] . "/check_" . $service['service_type'] . " -H " . ($service['service_ip'] ? $service['service_ip'] : $service['hostname']);
         $check_cmd .= " " . $service['service_param'];
     }
 
--- a/includes/services.inc.php
+++ b/includes/services.inc.php
@@ -129,7 +129,7 @@ function poll_service($service)
 
     // If we do not have a cmd from the check script, build one.
     if ($check_cmd == "") {
-        $check_cmd = $config['nagios_plugins'] . "/check_" . $service['service_type'];
+        $check_cmd = $config['nagios_plugins'] . "/check_" . $service['service_type']; . " -H " . ($service['service_ip'] ? $service['service_ip'] : $service['hostname']);
         $check_cmd .= " " . $service['service_param'];
     }
 
--- a/html/pages/devices.inc.php
+++ b/html/pages/devices.inc.php
@@ -1,6 +1,6 @@
 <?php
 
-// Set Defaults here algebur
+// Set Defaults here
 
 if (!isset($vars['format'])) {
     $vars['format'] = "list_detail";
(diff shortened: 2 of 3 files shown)


includes/services.inc.php
```diff
@@ -129,7 +129,7 @@ function poll_service($service)
 
     // If we do not have a cmd from the check script, build one.
     if ($check_cmd == "") {
-        $check_cmd = $config['nagios_plugins'] . "/check_" . $service['service_type'] . " -H " . ($service['service_ip'] ? $service['service_ip'] : $service['hostname']);
```
laf: Why are you removing this? It's needed for the checks

algebur: It's not needed for the checks, because it makes errors for plugins (nagios and others). Some plugins work, some don't. 

For example, if you use it 
"/check_" . $service['service_type'] . " -H " . ($service['service_ip'] ? $service['service_ip'] : $service['hostname']);

It runs "-H" by default. But some plugins don't need "-H" in command line, and you get "syntax error"

1. Check APC
apc_plugin_01
apc_plugin_02

2. Check Mitel
mitel_plugin
mitel_plugin_02

3. Another strange thing is when you want to use "ping" from "Nagios plugins".
the normal command for ping is: 
-H 172.20.1.56 -w 1000.0,20% -c 2000.0,60%
but if you run it, it looks like:
**-H** -H 172.20.1.56 -w 1000.0,20% -c 2000.0,60%
and result will be:
ping_plugin

But what is it? Is it normal result? It's garbage!!!

If you use "/check_" . $service['service_type'] ;  , I mean without "-H"
And type normal Parameters (which usually we use from http://nagios-plugins.org/doc/man/check_ping.html):
ping_plugin_02

you will get true result, which must be:
ping_plugin_03

If you need to run service, please, type full command from plugin documentation.
I hope you understand what I mean!

laf: You're using them wrong.

If the device you want to do the check against is the hostname you've added the device as then don't set any parameters or the IP. If you want to use a different hostname or IP then set it in the IP box. You don't need to use -H in the parameters box at all. Removing the code you have done means you will break existing service checks for most people.

If a service check doesn't need the -H then you create a custom service check file like we already have, see `includes/services/check_procs.inc.php`

algebur: Thanks for explaining! 

> If a service check doesn't need the -H then you create a custom service check file like we already have, see includes/services/check_procs.inc.php

Where I should create custom service check file? How to use it?

murrant: https://github.com/librenms/librenms/tree/master/includes/services

murrant: Should we block `-H` in the extra parameters?

algebur: I created new check file /includes/services/check_ping_custom.inc.php

`<?php

$check_cmd = $config['nagios_plugins'] . "/check_ping ".$service['service_param'];
`
How to use it in LibreNMS? I don't see it in the services list

laf: The web ui shows the nagios checks that are located in your plugin dir, you are using check_ping here so you'd need a check_ping.inc.php. What params are you passing as you mush provide a -H value to check_ping anyway which is already taken care of.

algebur: Ok! I got it. It works fine. Thanks a lot! 
But any way it is better to add SysName near the Hostname. Please do it.
Also my question how to add SysName to Alert Widget window?
sysname
Because it is too hard to understand what is host in huge list

murrant: Yes, adding sysName is good change, remove the other changes and we can merge that change :)

murrant: @algebur Two suggestions I see this all the time with people wanting to have devices show up as names but add them as an IP.  There are two options.

1. Set up a DNS server, even if it is just a local one on the LibreNMS box, but then you won't get the full benefit of the work to set it up.
2. Use this setting `$config['force_ip_to_sysname'] = true;`  If you see a spot that should respect that setting, but it does not, I suggest adding support.

P.S. https://github.com/librenms/librenms/blob/master/html/includes/table/alerts.inc.php

algebur: Hello!
I added 
$config['force_ip_to_sysname'] = true;
to file
/master/html/includes/table/alerts.inc.php
It works fine! 
Thanks!

algebur: It is Great Idea to add
$config['force_ip_to_sysname'] = true;
Can we add same code for emails?
I still receive emails with IP hostnames. It is hard to recognize what is that. 
here is example of my email. 
sysname_001

laf: You need to deal with that by using templates and setting what you want to see.

laf:
Please can you sign the CLA, we can't merge until this is done.

algebur:
Ok. I did it

PR #6874, opened 2017-06-22
refactor: Add an option to force transport for the snmp auto discovery

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---------------------------

- When TCP scan is running, the auto discovery with snmp scan is really slow
- Add an option to choose to use only tcp, udp or both

Thanks @schubling for the code.

### DIFF
--- a/doc/Extensions/Auto-Discovery.md
+++ b/doc/Extensions/Auto-Discovery.md
@@ -106,13 +106,19 @@ Enabled by default.
 This module is invoked from bgp-peers discovery module.
 
 #### SNMP Scan
-This isn't actually an auto- mechanism but manually invoked.
+This isn't actually an auto-mechanism but manually invoked.
 
 It's designed to scan through all of the subnets in your config or what you have manually specified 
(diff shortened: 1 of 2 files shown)


murrant:
I don't think you need the word force here.  Just call it transport.  -t (Select transport, tcp or udp.  The default is to try udp then tcp)

FTBZ:
OK changed and added information into the documentation.

laf:
What do people think of just defaulting to udp only and people have to set tcp or both? I think it's rare to see people using tcp for snmp so let's not slow things down.

murrant:
@laf tcp is only tried if udp fails and only if a device is pingable.

laf:
I know but people are complaining that snmp scan is slow, if it's because of tcp then why bother trying by default.

murrant:
@laf I'm working on speeding up snmp-scan and I don't think it will matter much.

FTBZ:
#6889

PR #6882, opened 2017-06-24
add OID for HP VC Flex-10/10D Module

bcenker:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6881


tests/snmpsim/hpvc.snmprec
```diff
@@ -1 +1,2 @@
 1.3.6.1.2.1.1.2.0|6|1.3.6.1.4.1.11.5.7.5.1
+1.3.6.1.2.1.1.2.0|6|1.3.6.1.4.1.11.5.7.5.4
```
laf: You need to put this in a new snmprec file. It also needs to include the sysDescr.

murrant:
Also, there seems to be another VirtualConnect device `.1.3.6.1.4.1.11.5.7.5.6| HP VC FlexFabric-20/40 F8 Module 4.45 2015-07-21T00:33:55Z`

I would suggest just changing the deteciton sysObjectId to `.1.3.6.1.4.1.11.5.7.5.`

murrant:
Thanks!

PR #6883, opened 2017-06-24
refactor: finish logic and definition separation

murrant:
Clean up rewrites to only have function definitions
Move authentication initialization into a function

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
I have no way to test anything other than mysql and http-auth - both of which are minimal changes and work fine :)

murrant:
Honestly, there is not much substantial change here.

murrant:
I tested it with AD auth

PR #6889, opened 2017-06-25
feature: parallel snmp-scan.py

murrant:
Reduces scan time of a /24 from 5 minutes to 14 seconds
Work is done by addhost.php

Just tries to addhost.php hostname/ip right now
Might need some more complexity added there, but I wasn't sure what.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
```
-bash-4.2$ ./snmp-scan.py -l
usage: snmp-scan.py [-h] [-r NETWORK] [-t THREADS] [-l] [-v]
snmp-scan.py: error: $config['nets'] is not set in config.php, you must specify a network to scan with -r
-bash-4.2$ ./snmp-scan.py --legend
usage: snmp-scan.py [-h] [-r NETWORK] [-t THREADS] [-l] [-v]
snmp-scan.py: error: $config['nets'] is not set in config.php, you must specify a network to scan with -r
```
Got this trying to list legend ^^

I do wonder if we scrap the php version, I don't think it's worth having two that do the same thing one just not as good as the other.

murrant:
Yes, scrap the php version, but not confident enough to do that yet ;)

That output seems appropriate laf, we don't have any networks to scan.  Perhaps the wording needs to be improved.

legend isn't a one-off option, it is printed at the top of the output and then the scan commences.

laf:
Oh it's a one off in the php version. I don't care either way but do you want to mimic the php version more closely?

murrant:
I changed it so it may not mimic the php version, but it can be used as a dropin by changing snmp-scan.php to snmp-scan.py

laf:
I definitely think we should ditch the php in this pull request.

Btw I think scan time depends on other factors, a /24 scan for me took 90+ seconds - still loads better than the php version.

FTBZ:
FYI, when scanning a /24 range with TCP not enabled it took hours for me not only 5 minutes.

laf:
Using this python version?

FTBZ:
Nope, I'm saying the old snmp-scan.php. I did't test the new one because it requiring python3.

murrant:
5 minutes is the python script without threading it does not try tcp at this time...

FTBZ:
It works immediately better... 

> Scanned 254 IPs: 30 known devices, added 0 devices, failed to add 37 devices
Runtime: 74.45 seconds

murrant:
@FTBZ Try with this setting enabled in config.php `$config['addhost_alwayscheckip'] === true`

murrant:
I think this is ready to go.  I'd say merge this.  Make an announcement about moving from snmp-scan.php to snmp-scan.py and then dump the snmp-scan.php.  I think some people run this in cron...

sorano:
Tested this and seeing really nice speed improvements:

```
librenms@nms:~$ /opt/librenms/snmp-scan.py 172.31.5.0/24
Scanning IPs:
*******..............************..*.*..***.*.*...*...............----.**.....-.------..----------..-..-....-..**--..----.....................*-*...**************........****..----..**********.................-------.....------------**********-*--

Scanned 247 IPs: 72 known devices, added 0 devices, failed to add 57 devices, 7 ips excluded by config
Runtime: 102.01 seconds
librenms@nms:~$ time /opt/librenms/snmp-scan.php -r 172.31.5.0/24
Range: 172.31.5.0/24
|||||||..**.*****.........****.****.*****-*******.--..---..---.....-----..----.-......--.-.--------....-**.****........--------............--.-**-...*****........................*********************-----....----..--..---*******.................****.....
Scanned 254 IPs, Already known 72 Devices, Added 0 Devices, Failed to add 57 Devices.
Runtime: 723.93255090714 secs

real    12m3.964s
user    0m12.884s
sys     0m8.000s


with this setting enabled in config.php $config['addhost_alwayscheckip'] = true;

librenms@nms:~$ /opt/librenms/snmp-scan.py 172.31.5.0/24
Scanning IPs:
*******................***.*****..************....................----*-*------...-.-...--------.--...-.....-.**-----....................-.....****************......**.**...*..----***..*******.................-------.....------------*********-**--

Scanned 247 IPs: 73 known devices, added 0 devices, failed to add 57 devices, 7 ips excluded by config
Runtime: 101.61 seconds

```

Issue #6894, opened 2017-06-27
API Service State Filter not filtering as expected


Dubby85:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
==========================================================
Component | Version
--------- | -------
LibreNMS  | 5e32474d26cc30c44bad82fc3e926c0183e8e020
DB Schema | 196
PHP       | 7.0.19-1
MySQL     | 10.1.23-MariaDB-9+deb9u1
RRDTool   | 1.6.0
SNMP      | NET-SNMP 5.7.3
==========================================================

- [X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
This link just takes me to an unrelated section of the FAQ
- [X] Please provide as much detail as possible.

"/api/v0/services?state=<1|2>" I would expect to just return any services in a warning/critical state, however if a host has 1 service in a warning/critical state all other services on that host that are in a 0 state also get returned. For Example: https://pastebin.com/iZnZyueW only 1 of 3 services on arcee is bad but all services on the host are being returned.

- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://opencollective.com/librenms/donate

PR #6895, opened 2017-06-27
Fix for API services bug #6894

Dubby85:
Fixed an issue with the output including all services from a host that had a single matched service  regardless of whether they matched the filter.

Bug report: https://github.com/librenms/librenms/issues/6894

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixes: #6894

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -1512,31 +1512,34 @@ function list_services()
     $code     = 200;
     $message  = '';
     $services = array();
-    $where    = '';
+    $where    = array();
+    $params   = array();
 
-    // Filter BY STATE
+    // Filter by State
     if (isset($_GET['state'])) {
-        $where .= (!empty($where)) ? ' AND' : ' WHERE';
-        $where .= " S.service_status= ".$_GET['state']." AND S.service_disabled='0' AND S.service_ignore='0'";
+        $where[] = '`service_status`=?';
+        $params[] = $_GET['state'];
+        $where[] = "`service_disabled`='0'";
+        $where[] = "`service_ignore`='0'";
+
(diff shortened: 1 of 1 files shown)


murrant:
It might be nicer to build the query like this (partial untested example)
```php
$where = array();
$params = array();

if (isset($_GET['state'])) {
    $where[] ='`service_status`=?';
    $params[] = $_GET['state'];
    $where[] = "`service_disabled`='0'";
    $where[] = "`service_ignore`='0'";
}

$query = 'SELECT * FROM `services`';
if(!empty($where)) {
    $query .= ' WHERE ' . implode(' AND ', $where);
}
$query .= ' GROUP BY S.service_ip ORDER BY S.service_ip';

$services = dbFetchRows($query, $params);
```

Try to use ? for any supplied variables.

Dubby85:
Thanks @murrant it worked out of the box, I have committed an updated version based on your suggestion.

html/includes/api_functions.inc.php
```diff
@@ -1510,53 +1511,50 @@ function list_services()
...
-            $devicewhere  = " AND `service_type` LIKE ?";
-            $sql_param[1] = $_GET['type'];
-        }
-
-        $services[] = dbFetchRows("SELECT * FROM `services` WHERE `device_id` = ?".$devicewhere, $sql_param);
+    if (!empty($where)) {
+        $query .= ' WHERE ' . implode(' AND ', $where);
     }
+    $query .= ' ORDER BY service_ip';
+    $services[] = dbFetchRows($query, $params);
```
murrant: Odd that $services is a double nested array.

murrant:
Thanks for the fix, much better now.

Issue #6820, opened 2017-06-11
Add support for HPE Comware CPU and Memory and Temp


kkrumm1:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the 
==========================================================
Component | Version
--------- | -------
LibreNMS  | 51ccec94c519b32fd098bce27904a8612e224641
DB Schema | 194
PHP       | 7.0.19
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
==========================================================
[OK]    Database connection successful
[OK]    Database schema correct
` section of `./validate.php` which you can run from the cli.
- [X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Discovery: https://gist.github.com/kkrumm1/46978bcf3a393fbce4253207299e6e19
Poller: https://gist.github.com/kkrumm1/9a7e22fa3f151a6120eb621719ede834
SNMP Walk: https://gist.github.com/kkrumm1/40e23cf478896efa37d91ac35aa1bf36

Object: hh3cEntityExtCpuUsage (% Percent)
OID: 1.3.6.1.4.1.25506.2.6.1.1.1.1.6

Object: hh3cEntityExtMemUsage (% Percent)
OID: 1.3.6.1.4.1.25506.2.6.1.1.1.1.8

Object: h3cEntityExtTemperature (Celsius)
OID"  1.3.6.1.4.1.25506.2.6.1.1.1.1.12

PR #6896, opened 2017-06-27
newdevice: Added memory detection, fixed cpu polling

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

CPU: Was already supported but imho broken, it wouldn't detect any CPUs as entPhysicalName contained more than just Board so that's now fixed.

MEM: Added support. I've used entPhysical from the DB like the processor code does and we did in fdb although a switch out to walking entPhysical is possible (I actually did it like that at first so I can switch back easy enough).

TEMP: Already supported.

Fixes: #6820


kkrumm1:
Thank you - laf,  I appreciate it.

PR #6902, opened 2017-06-28
feature: support fdb table on generic devices

murrant:
That implement Q-BRIDGE-MIB or BRIDGE-MIB

refactored code to use snmpwalk_group()
modified snmpwalk_group() so it works as intended
refactored ajax table code to use joins and reduce extra queries
add indexes for device_id and port_id, drop primary index
Make all columns sortable
Fix a few other small issues

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


kkrumm1:
Fancy!

lundgrenolof:
I tested this and it found FDB tables on my HP Procurve devices. It did not find anything on zyxel gs2200, openwrt, generic linux, esxi or ubiquiti router.

For how long are record saved?

murrant:
@lundgrenolof the records are updated each discovery.  So they will stay that long.

Devices need to return fdb data via snmp to be supported.  There is a standard MIB for that Q-BRIDGE-MIB and BRIDGE-MIB.  If you have a device you know returns data, open a new issue and it will likely get added.

PR #6926, opened 2017-06-30
newdevice: Add Calix AXOS/E5-16F Detection

djamp42:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Does AXOS use the same mibs as E7?

Also, this should probably be more generic like `.1.3.6.1.4.1.6321.1.2.4.1.4`
And you need to add an additional snmprec file: http://docs.librenms.org/Developing/os/Test-Units/

djamp42:
I'm trying to get the MIBs, but I believe they are different, as AXOS is the new software Calix is going to. 
I'm not sure what you mean by more generic, as that OID is the same one i have. 
For the snmpprec file, the sysDescr returns the model number but also includes the version number of the AXOS software installed, so i'm not sure how you handle this as E5-16F could have different versions of the AXOS software on it. 

1.3.6.1.2.1.1.1.0|4|Calix, Inc E5-16F AXOS-R2.2.0
1.3.6.1.2.1.1.2.0|6|1.3.6.1.4.1.6321.1.2.4.1.4

murrant:
I mean we match by the start of the sysObectID string, so 1.3.6.1.4.1.6321.1.2.4. will likely match every AXOS device, not just E5-16F.

We probably need to start a new OS for AXOS if they have different MIBs. I suggest calix-axos

djamp42:
I just checked and it uses the same mibs as the e7. It actually uses less.

laf:
So should we do this as a new OS axos.yaml if that's what this device runs?

murrant:
@laf I say no at this time.  We could split it off later.
AXOS is basically the SDN (They like to call it SDA) platform for Calix.

laf:
You confused me then as the email notification said no but this says yes :)

@djamp42 if you can split this off into a new os called axos that would be ace, if you don't have time let us know and one of us will sort it out.

murrant:
@laf changed my mind and edited post.

djamp42:
@laf I would rather you do it, just because I'm not a programmer by trade. If you need any other device info let me know.

laf:
@djamp42 I've update this now, please test the update to make sure it does what you need.

I've removed what appears to be non-relevant discovery for models in the calix range so this is limited. If you find it's missing something that was there when it was calix then let me know.

djamp42:
Thanks, Works fine, just added another E5 and was detected as Calix with this pull request.

includes/polling/os/axos.inc.php
```diff
@@ -0,0 +1,20 @@
...
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+
+// Device might not have a card 1 (or even card2 if it is an E7-20)
+$version = strtok(snmp_walk($device, "e7CardSoftwareVersion.1", "-OQv", "E7-Calix-MIB"), PHP_EOL);
```
murrant: I'm not sure this is relevant to the hardware.
@djamp42 can you snmpwalk these on your device and let us know if they are valid for the hardware?

djamp42: E5-16F doesn't have slots, it's a single chassis of 16 gfast ports. My understanding is the E7 will support NG-PON2 and these new cards will run AXOS. They haven't been released yet.

murrant: @djamp42 That seems to muddle things more...  I'll do some research with my Calix contacts in a couple of days.

laf:
Thanks for the pull request @djamp42 :)

image

PR #6931, opened 2017-07-03
Use the new locks for schema updates

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

No code changes inside the block (try ?w=1 on the github url)

Thanks @keeperofdakeys


laf:
@keeperofdakeys  @murrant I've removed the discovery one as it's not needed as we already do the check in sql-schema/update.php

murrant:
@laf I don't think that is what the discovery lock was for.  I remember someone having long running new device discovery.

On this PR, do you think we should make it wait before continuing or just continue straight away like it is now?

Also, will this work for distributed pollers?

laf:
@murrant The long running issue was multiple stacked new discoveries which https://github.com/murrant/librenms/blob/c8c6bd0f1f410b1a5c75d81f34071f24080f2da8/discovery.php#L36 fixes.

I don't think waiting is the best option really

keeperofdakeys:
If you're concerned about distributed pollers, you could use mysql named locks (`GET_LOCK()`) for schema updates.

murrant:
@laf Oh, I didn't even know that was there.  Read your comment on mobile and didn't notice the change. 👍 

@keeperofdakeys That's what we would have to do, but the question is are we concerned about distributed pollers?

murrant:
As for the waiting, I don't think it is the best thing either.  But do realize I mean wait with a timeout and periodically while waiting check the lock.

laf:
I guess we have 4:59m to play with before a new discovery process is kicked off so go for it. 30 seconds maybe?

murrant:
@laf we could use this to do per-device discovery/polling locking like the python poller service does.

laf:
@murrant The mysql locking here, how does that work for replicated mysql or mysql clusters?

murrant:
@laf From the MySQL docs:
```
This function is unsafe for statement-based replication. A warning is logged if you use this function when binlog_format is set to STATEMENT.
```
So, assuming that is not the case, things should work.  I don't have access to either of those setups at this time.

laf:
@f0o Any thoughts?

keeperofdakeys:
For flock, I'd remove the `LOCK_NB` bitmask if you wanted to wait forever.

laf:
@murrant Ready for another test?

murrant:
Yeah, passed my tests.  With the 30s wait they all run, but not concurrently.  I set it to 0 to check that again.

laf:
@f0o Any comments on this?

@murrant Is the MySQL lock actually used here?

murrant:
@laf not anymore

laf:
What's the need for `LibreNMS/MysqlLock.php` then?

murrant:
None, I can dump it but it works fine as long as there is not statement based replication or galera

laf:
Testing overnight.

Just needs the mysqllock stuff removing from test unit. I can do that tomorrow if you don't get round to it.

murrant:
Thanks @laf

PR #6932, opened 2017-07-03
Added HTTPS config instructions

tslytsly:
Added new section explaining how to enable SSL on Ubuntu and Nginx using a LetsEncrypt SSL cert.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/General/Security.md
+++ b/doc/General/Security.md
@@ -19,7 +19,7 @@ hijacking attacks. Set ``$config['secure_cookies'] = true`` in ``config.php`` to
 Please ensure you keep your install [up to date](Updating.md).
 
 ### Enabling HTTPS
-This example is specifically for Ubuntu 16.04 eith Nginx and uses an SSL certificate from Let's Encrypt.
+This example is specifically for Ubuntu 16.04 with Nginx and uses an SSL certificate from Let's Encrypt.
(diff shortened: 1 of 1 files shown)


laf:
I think these belong somewhere else. I'll have a think.

tslytsly:
Yea, I wasn't sure either.
I thought about as a sub-heading of http://docs.librenms.org/Installation/Installation-Ubuntu-1604-Nginx/#web-server, but then I thought that SSL is not an install requirement. (Even if it is a good idea)

murrant:
I think I would prefer to have the certbot renew cron in it's own file.  This way if someone copies the standard cron from our repo over the one they have they don't delete it.
Also, any reason you run certbot as librenms, but then have it sudo?

tslytsly:
> I think I would prefer to have the certbot renew cron in it's own file. This way if someone copies the standard cron from our repo over the one they have they don't delete it.

Yea, good point. I can change and test that on my system.

> Also, any reason you run certbot as librenms, but then have it sudo?

I originally ran it as root without sudo, but then I was getting ownership warnings for my Nginx access and error logs when I ran validate.php.
So I switched it to run as librenms with sudo and, so far, I've not got the same error.

laf:
I think the best place for this is under Support -> SSL Configuration

This requires creating a new page which is more than just a new file so please ensure you read: http://docs.librenms.org/Developing/Creating-Documentation/

As for the permissions on the logs, this is fixed in validate.php now so you can just drop the sudo from librenms in cron and split that cron entry out into it's own.

PR #6935, opened 2017-07-03
Fix bug get_graph_by_port_hostname() only hostname

bitency:
From the documentation:

Function: get_graph_by_port_hostname top
Get a graph of a port for a particular device.

hostname can be either the device hostname or id

But the function is only works with a hostname and not with a ID.
I have modified a few things that solve the problem.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/api_v0.php
+++ b/html/api_v0.php
@@ -12,167 +12,1559 @@
  * the source code distribution for details.
  */
 
-$init_modules = array('web', 'alerts');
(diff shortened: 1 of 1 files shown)


murrant:
@bitency you seem to have replaced the api_v0.php file with the api_functions.inc.php file's contents.  Can you please fix that.  Let us know if you need help with git.

bitency:
Please close this, wrong file.

PR #6936, opened 2017-07-03
Fix bug get_graph_by_port_hostname() only hostname

bitency:
From the documentation:

Function: get_graph_by_port_hostname top
Get a graph of a port for a particular device.

hostname can be either the device hostname or id

But the function is only works with a hostname and not with a ID.
I have modified a few things that solve the problem.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -51,7 +51,7 @@ function get_graph_by_port_hostname()
     $app          = \Slim\Slim::getInstance();
     $router       = $app->router()->getCurrentRoute()->getParams();
     $hostname     = $router['hostname'];
-	$device_id    = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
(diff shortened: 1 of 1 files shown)


html/includes/api_functions.inc.php
```diff
@@ -51,6 +51,7 @@ function get_graph_by_port_hostname()
     $app          = \Slim\Slim::getInstance();
     $router       = $app->router()->getCurrentRoute()->getParams();
     $hostname     = $router['hostname'];
+	$device_id    = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
```
laf: Thanks, looks like the indentation here needs two more spaces.

murrant: I'm not sure that it matters, but this does an additional sql query.
You could also modify the query below.

murrant: It's a tab instead of spaces :)

laf: :)

PR #6942, opened 2017-07-03
fix: define standard ups-mib divisors properly

murrant:
Set default divisors for all devices and remove per-os overrides for them.
We should only use per-os overrides for non-standard behavior.

High possibility for breakage here if devices not listed here have non-standard behavior.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -959,49 +959,28 @@ function get_device_divisor($device, $os_version, $sensor_type, $oid)
             } else {
                 return 1;
             }
-        } elseif ($sensor_type == 'voltage') {
-            return 1;
         }
-    } elseif (($device['os'] == 'huaweiups') && ($sensor_type == 'frequency')) {
-        return 100;
-    } elseif (($device['os'] == 'netmanplus') && ($sensor_type == 'load')) {
-        return 1;
-    } elseif (($device['os'] == 'netmanplus') && ($sensor_type == 'voltage')) {
-        if (preg_match('/.1.3.6.1.2.1.33.1.2.5./', $oid)) {
-            return 10;
-        }
-        return 1;
-    } elseif ($device['os'] == 'generex-ups') {
-        if ($sensor_type == 'load') {
-            return 1;
-        } elseif ($sensor_type == 'voltage' && !starts_with($oid, '.1.3.6.1.2.1.33.1.2.5.')) {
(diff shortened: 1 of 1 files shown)


laf:
How's this going to use the standards from the mibs? I thought the reason we had this function in place is that we already know devices don't honour what the standard UPS-MIB divisor is?

murrant:
@laf Already discussed a bit on IRC.  But the divisor for everything in UPS-MIB isn't just a flat 10 as suggested by the previous code.

Load and Voltage (except upsBatteryVoltage) have a divisor of 1.

This is why we keep having to define those for specific devices as bug reports come in.

laf:
This is upsBatteryVoltage:

```
/opt/librenms-snmpsim/captures/aeg-ups.snmpwalk:.1.3.6.1.2.1.33.1.2.5.0 = INTEGER: 547
/opt/librenms-snmpsim/captures/b6907a29:.1.3.6.1.2.1.33.1.2.5.0 = INTEGER: 2054
/opt/librenms-snmpsim/captures/cs141.snmpwalk:.1.3.6.1.2.1.33.1.2.5.0 = INTEGER: 3394
/opt/librenms-snmpsim/captures/dellups-3750w.snmpwalk:.1.3.6.1.2.1.33.1.2.5.0 = INTEGER: 2080
/opt/librenms-snmpsim/captures/eaton-powerxpert.snmprec:1.3.6.1.2.1.33.1.2.5.0|2|5444
/opt/librenms-snmpsim/captures/eatonups-ex2200.snmpwalk:.1.3.6.1.2.1.33.1.2.5.0 = INTEGER: 820
/opt/librenms-snmpsim/captures/powerwalker-2.snmprec:1.3.6.1.2.1.33.1.2.5.0|2|2054
/opt/librenms-snmpsim/captures/powerwalker.snmpwalk:.1.3.6.1.2.1.33.1.2.5.0 = INTEGER: 409
/opt/librenms-snmpsim/captures/tripplite.snmprec:1.3.6.1.2.1.33.1.2.5.0|2|542
```

@murrant That seems to suggest different divisors :(

murrant:
@laf I'm not sure we use rfc1628 for any of those that report four digit battery voltages

As far as I can tell this should not change the behavior of any devices we know about and should limit the number of "workarounds" in this function to only devices that have non-standard behavior.

Issue #6933, opened 2017-07-03
Powerwalker UPS data incorrect


tslytsly:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for  bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).
If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

```
==========================================================
Component | Version
--------- | -------
LibreNMS  | 701fbbc29b7a25e215032a6435fac72c06094976
DB Schema | 198
PHP       | 7.0.18-0ubuntu0.16.04.1
MySQL     | 10.0.29-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3
==========================================================
```
Pastebin links: 
Disco: https://p.libren.ms/view/4d50fdc1
Poll: https://p.libren.ms/view/99adf121
SNMPWalk: https://p.libren.ms/view/b6907a29

**Details**
LibreNMS seems to report battery voltage and frequency slightly wrong for Powerwalker UPS.

Ours is showing Batt volts of 205.2v and Output Freq as 50.0Hz on its inbuilt webpage, but in LNMS it's showing as 2052v and 500Hz.

This seems to be because it reports the data without decimal places. 
Input and Output voltage get reported correctly, so it's showing Output voltage of 230.0 on LNMS, even though the MIB reports 2300.

laf:
What version of software are you running on these? Data from another powerwalker shows it's fine so we need to work out of if this is a difference on the software version.

tslytsly:
Web manager reports firmware is 1.4.0.3

[...]

PR #6944, opened 2017-07-03
newdevice: Moved powerwalker over to using rfc compat for sensors

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

In some way fixes: #6933 

Also done a notification. https://github.com/librenms/librenms.github.io/pull/91 for the 11th which shouldn't be merged until tomorrow.


murrant:
Looks good, except it looks like we need to skip power sensors that return < 0

image

laf:
I thought that for something else recently and someone made a valid point that if those change then it will be picked up on the next poll rather than discovery.

murrant:
Yeah, no idea on how the device behaves...

PR #6958, opened 2017-07-04
newdevice: Updated powerwalker sensors to use custom mib + new feature of sensor discovery for options

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This will need a notification sending out.

So moved powerwalker to use it's own mib rather than the bodge at rfc compat.

Added support to sensor discovery in yaml to provide group options for sensor classes to avoid code duplication.


laf:
@murrant If this gets approved, I'd like to take the code changes and put them into a new PR so we can make use of them in other current PRs rather than waiting for this to be merged after a notification goes out.

doc/Developing/os/Health-Information.md
```diff
@@ -40,23 +40,28 @@ mib: NETBOTZV2-MIB
...
         airflow:
-            -
-                oid: airFlowSensorTable
-                value: airFlowSensorValue
-                divisor: 10
-                num_oid: .1.3.6.1.4.1.5528.100.4.1.5.1.2.
-                descr: airFlowSensorLabel
-                index: 'airFlowSensorValue.{{ $index }}'
+            data:
+                options:
```
murrant: options is in the wrong place here.

murrant: This options is in the wrong place.

murrant:
Looks good overall!  How much of a gain is it to have the group options vs just requiring them in each sensor?

laf:
This pr would have meant having them in 90% of the sensors so just seems daft not to make it more global. Divisor is the same for a lot of devices so I think it's a good enough saving.

murrant:
Sounds good to me.  Go ahead and split that off (fixing the documentation with invalid yaml)

PR #6967, opened 2017-07-05
feature: add timeout to AD auth, default is 5s

murrant:
Can be configured by $config['auth_ad_timeout']

One of my ad servers stop responding to ldap requests and LibreNMS never tried to use the second ad server, just gave a http timeout error.

Adding a timeout allows ldap to give up and try additional servers, or at least actually return some html (which will print an connection failed error).

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Authentication.md
+++ b/doc/Extensions/Authentication.md
@@ -156,12 +156,13 @@ If you set ```$config['auth_ad_require_groupmembership']``` to 1, the authentica
 ##### Sample configuration
 
 ```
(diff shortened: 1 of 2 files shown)


laf:
Naughty powerwalker @murrant :)

laf:
Can you add the config option to docs?

Is it easy enough to add a reconnect in so if the connection has timed out it will just auto connect rather than kick the user our (just spit balling)

PR #6971, opened 2017-07-06
fix: Change .htaccess to compensate for Apache bug

lasseleegaard:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
This was reported as being broken in Apache 2.4.25. I've tested this change on a slightly earlier version and it's fine.

I don't know why it breaks in the first place as that's a valid rule but I did find a link that indicates it might be a bug in apache: https://serverfault.com/questions/847312/rewriting-urls-while-using-mod-proxy-fcgi-and-path-info-ends-up-with-error-404/847750#847750

This has been reported before so maybe it's about time we changed it.

laf:

murrant:
Seems ok, but I don't have any apache to test on.

paulgear:
I've applied this to my system running Apache 2.4.18-2ubuntu3.3 and it didn't seem to break anything, so I think we're safe to merge.

PR #6973, opened 2017-07-06
validate improvements

murrant:
Always output header (may be missing mysql data)
Improve version display, now looks like '1.28-129-g74e6c3edf' <tag>-<commits since tag>-<shortag>
Add mysqli to the list of required extensions, check extensions before init
Add dbIsConnected() function, used to check if we have db in version_info()
Do not die on IPv4 and IPv6 so validate can print errors.
Move git checks so they are output together
fix rrdtool 1.7.0 reports version as 1.7.01.7.0

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/about.inc.php
+++ b/html/pages/about.inc.php
@@ -155,7 +155,7 @@ $mysql_version   = $versions['mysql_ver'];
 $netsnmp_version = $versions['netsnmp_ver'];
 $rrdtool_version = $versions['rrdtool_ver'];
 $schema_version  = $versions['db_schema'];
-$version         = `git rev-parse --short HEAD`;
+$version         = $versions['local_ver'];
 $version_date    = $versions['local_date'];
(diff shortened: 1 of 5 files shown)


laf:
@murrant `LibreNMS  | 1.28-130-gae4e66999` - what does that even mean :)

murrant:
@laf I put it in the post but it got stripped.   It means tag-commits since tag-short commit hash

laf:
So 1.28 = Tag

1.29 is the latest tag though:

```bash
-bash-4.2$ git describe --tags $(git rev-list --tags --max-count=1)
1.29
-bash-4.2$ ./validate.php
Component | Version
--------- | -------
LibreNMS  | 1.28-130-gae4e66999
```

murrant:
I'll check it

PR #6976, opened 2017-07-07
refactor: drop ifHighSpeed from the database it is unused

murrant:
We still use it to calculate ifSpeed

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
There's more mentions of ifHighSpeed in ports polling which may need to be looked at. We also mention it in alerting and api docs.

Which then brings us on to, whilst we don't use it, I don't think we can now be sure others don't. We export that data to InfluxDB + others now and also to the API so this could lead to a breaking change which at minimum needs to be communicated but imho we should just leave this in place.

murrant:
The other references in the ports are to update ifSpeed based on ifHighSpeed.  Many of the entries are null, I can't see how that would be useful.

I'll look at it again.  We could emulate ifHighSpeed, but I don't see the point.  I'm fine with leaving it but it caused me some issues when troubleshooting a problem.

PR #6985, opened 2017-07-08
feature: Added ability specify options for sensors yaml discovery

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This is from #6958 so that we don't have to wait for a week to merge it whilst we notify on the powerwalker changes.

@murrant Should be the same code / doc as you reviewed in #6958 but with the fix in the docs you mentioned.

### DIFF
--- a/doc/Developing/os/Health-Information.md
+++ b/doc/Developing/os/Health-Information.md
@@ -40,23 +40,28 @@ mib: NETBOTZV2-MIB
 modules:
     sensors:
         airflow:
-            -
-                oid: airFlowSensorTable
-                value: airFlowSensorValue
-                divisor: 10
-                num_oid: .1.3.6.1.4.1.5528.100.4.1.5.1.2.
-                descr: airFlowSensorLabel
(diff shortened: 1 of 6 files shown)


doc/Developing/os/Health-Information.md
```diff
@@ -67,6 +72,16 @@ The only sensor we have defined here is airflow. The available options are as fo
   - `descr` (required): The visible label for this sensor. It can be a key with in the table or a static string, optionally using `{{ index }}`
   - `index` (optional): This is the index value we use to uniquely identify this sensor. `{{ $index }}` will be replaced by the `index` from the snmp walk.
   - `skip_values` (optional): This is an array of values we should skip over.
+  - `skip_values_lt` (optional): If sensor value is less than this, skip the discovery.
```
murrant: Should this just be `skip_value_lt`?  Without the **s**.

laf: makes sense, i've updated to reflect

PR #6996, opened 2017-07-10
device: Improved HP ILO4 support

TomEvin:
Added HP ILO4 temperature (HDD, motherboard), power, state (powersupply, HDD, Fans, memory slots) sensors support.
Tested with HP ProLiant ML350p Gen8-Gen9

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@TomEvin Can you delete CPQHLTH-MIB from the base mib directory?

TomEvin:
@murrant Done. I hope it's not cause any problem in other part of the codes. That's why I didn't wanted to move it from the original place.
I would move all the CPQ* mib files to the HP folder, because all of them HP mibs.
(PS: I also added CPU state check now.)

murrant:
@TomEvin You were right I did a quick search and found
`includes/discovery/sensors/temperature/linux.inc.php` will be incorrect with the moved mib.

Can you change line 28 to:
```
$descr = snmp_get($device, $descr_oid, '-Oqnv', 'CPQHLTH-MIB', 'hp');
```
 
To fix that.

Moving the other CQP*-MIB files shouldn't cause an issue based on my search.

laf:
Also, we've recently switched to be able to discover sensors using a YAML file, http://docs.librenms.org/Developing/os/Health-Information/#simple-health-discovery

Could you convert to using that? If you don't have time then can you provide the output of `snmpbulkwalk -OUneb -v2c -c COMMUNITY HOSTNAME .` and I'll try.

TomEvin:
@murrant I have made the changes. Please check it

@laf I'm glad I could reached this far with php. If you could do a yml base I could extend that in the future. If you need more info, let me know.
[snmpbulk.txt](https://github.com/librenms/librenms/files/1138057/snmpbulk.txt)

laf:
@TomEvin That doesn't contain all the data needed. Can you re-run the snmpwalk with ` .1.3.6.1.4.1.232.6` at the end instead of `.`

TomEvin:
You are right, only the important part was not in it.
[snmpbulk2.txt](https://github.com/librenms/librenms/files/1153397/snmpbulk2.txt)

laf:
@TomEvin Ok so I've made the updates and i'm sure I've got it working pretty much like for like with yaml (needed some core code changes to work).

As you submitted this from your master branch I'm wary of submitting this back to it. You can test my branch: https://github.com/laf/librenms/tree/tomevin-master -
 to test: http://docs.librenms.org/Support/FAQ/#how-can-i-test-another-users-branch

You might notice some slight differences in descriptions (like memory states) as we can't run another snmp query to just get that value.

If that works ok, I'll push to your master branch - but FYI you'll be in for some pain with your git after :(

TomEvin:
@laf I ran the first test. The CPQHLTH-MIB is a html file, not a valid mib in your branch! Beside this it looks fine. I could not test everything, but I trust in your experties. If there is an error, I'll find it. ;) Also I'm planning add additional ILO functions anyway.
My questions: Should I add any new devices with yaml only in the future? inc.php solution is obsolote?

laf:
@TomEvin I've just fixed that now, I'll push to your branch and we can then merge.

We'd prefer to use the yaml way but it's not always possible. It's a little easier to add new sensors using the yaml file and requires less code.

laf:
Ok, I can't push to your master. I think you should clean your master back to default. Then take my branch and use that for your own new PR unless you want me to just create the PR? You will still need to fix your master branch though.

TomEvin:
@laf I think you cando it.
Because I use Git for the first time, "fix ma master branch" means delete my branch, and create new branch from librenms?

laf:
@TomEvin if you have librenms/librenms setup as a remote then skip this next line:

`git remote add upstream https://github.com/librenms/librenms.git && git remote update upstream`

Now run:

```
git fetch upstream
git reset --hard upstream/master
```

laf:
I've created #7053

You should have a read of http://docs.librenms.org/Developing/Using-Git/

TomEvin:
@laf Could you try it out now? If still not good, I'll use the "hard" way. :)

laf:
I don't want to push to your master :)

If you've fixed it, do `git push origin master` if it complains saying do a force push then try `git push origin +master`

laf:
Closing in favour of #7053

PR #7002, opened 2017-07-10
fix all mib errors in base mib directory

murrant:
Move cabletron mibs into enterasys directory
move CISCO-LWAPP mibs into cisco directory and fix errors
remove MPLS-LDP mibs as I couldn't reconcile them and they seem unused
move watchguard IPSEC-ISAKMP-IKE-DOI-TC into watchguard directory
update several cisco mib files to remove errors
Fix typo in tripplite mib
Fix RADWIN-MIB-WINLINK1000 based on pdf I found online, some lines were missing, added the bare minimum to make it valid

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Out of interest how did you work this out? Wondering if we can add some form of unit test.

murrant:
Running snmptranslate with +ALL.  We could probably make a test for it.

Only test each directory individually, otherwise it hits the max TCs.

Issue #7001, opened 2017-07-10
Installation bug: PHP module `mysqli` not checked in `/install.php` before use


sTywin:
Steps to reproduce:

1. Fresh installation (i.e. no config.php)
2. No `mysqli` extension enabled/installed for php-fpm
3. Browse to `/` or `/install.php`

Expected outcome:

- Dependencies table shown with `mysqli` in red

Actual outcome:

- Receive the following PHP error:
`Fatal error: Uncaught Error: Call to undefined function mysqli_query() in /opt/librenms/includes/dbFacile.php:94 Stack trace: #0 /opt/librenms/includes/dbFacile.php(290): dbQuery('SELECT DISTINCT...', Array) #1 /opt/librenms/includes/dbFacile.php(323): dbFetchRows('SELECT DISTINCT...', Array, false) #2 /opt/librenms/includes/dbFacile.php(401): dbFetch('SELECT DISTINCT...', Array, false) #3 /opt/librenms/includes/common.php(1605): dbFetchColumn('SELECT DISTINCT...') #4 /opt/librenms/includes/init.php(165): load_all_os(true) #5 /opt/librenms/html/install.php(22): require('/opt/librenms/i...') #6 {main} thrown in /opt/librenms/includes/dbFacile.php on line 94`

Addenda:

- For context, I am working on a set of Docker images. I am aware of `jarischaefer/docker-librenms`, however in my implementation, each service (cron, php-fpm, mariadb, nginx, and eventually rrdcached and memcached) are in separate containers.
- It's still obvious that it's `mysqli` that's missing, it would just be nice if the user saw the nice table (as they do if e.g. `gd` and `mcrypt` are missing).

Ticking of boxes:

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      - Yes, commit 6110393
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
      `[FAIL]  config.php does not exist, please copy config.php.default to config.php`
- [ ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
      - I didn't see anything relevant to this particular issue, which is an install-time issue only.
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

PR #7003, opened 2017-07-10
fix: Stop loading all oses when we have no db connection

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Stop trying to load all oses if we have no db connection (install process).

Fixes: #7001


murrant:
@laf you sure this won't have any side effects?

We could force it to load all os (which doesn't require the db) if nodb is specified.

laf:
Not 100% no but we call it with $existing = true so the DB is attempted to be used. 

I can change it to pass false if nodb is set?

murrant:
@laf Yeah, that is what I was thinking, take a performance penalty, but might cause less unexpected edge cases...

laf:
@murrant Updated, tested and it now shows the module not being installed in install.php if mysqli isn't available.

PR #7011, opened 2017-07-11
fix: snmp-scan.py now runs on python 2.7 as well as python 3

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/snmp-scan.py
+++ b/snmp-scan.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 """
 Scan networks for snmp hosts and add them to LibreNMS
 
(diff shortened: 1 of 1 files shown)


kedearian:
Tested fix, sensors reappeared.

laf:
Slight documentation issue in that it says to specify -r for a network to scan but that's not needed:

```
-bash-4.2$ ./snmp-scan.py
usage: snmp-scan.py [-h] [-t THREADS] [-l] [-v] [network [network ...]]
snmp-scan.py: error: $config['nets'] is not set in config.php, you must specify a network to scan with -r
-bash-4.2$ ./snmp-scan.py -r 192.168.10.0/24
usage: snmp-scan.py [-h] [-t THREADS] [-l] [-v] [network [network ...]]
snmp-scan.py: error: $config['nets'] is not set in config.php, you must specify a network to scan with -r
-bash-4.2$ ./snmp-scan.py 192.168.10.0/24
Scanning IPs:
```

murrant:
@laf you found a bug :)  the network argument over-writes the -r argument, but with action=append we get a list in a list.

We'll preserve the backwards compatible options still

PR #7014, opened 2017-07-12
fix: Fix for Rittal LCP graphs divisors

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---------------------

- Disabled unnecessary discovery modules
- Corrected wrong divisors

Can't find a way to disable the wrong over graphs or to change them by the ones included in the definitions file.

### DIFF
--- a/includes/definitions/rittal-lcp.yaml
+++ b/includes/definitions/rittal-lcp.yaml
@@ -4,16 +4,16 @@ type: environment
 icon: rittal
 mib_dir:
     - rittal
+discovery_modules:
+    storage: 0
+    mempools: 0
+    processors: 0
 discovery:
     -
         sysObjectId:
             - .1.3.6.1.4.1.2606.7
     - sysDescr:
         - 'Rittal LCP'
-discovery_modules:
-    storage: 0
-    mempools: 0
-    processors: 0
 over:
     - { graph: device_temperature, text: 'Temperatures' }
     - { graph: device_power, text: 'Cooling capacity' }
--- a/includes/definitions/rittal-lcp.yaml
+++ b/includes/definitions/rittal-lcp.yaml
@@ -2,6 +2,9 @@ os: rittal-lcp
 text: 'Rittal LCP'
 type: environment
 icon: rittal
+over:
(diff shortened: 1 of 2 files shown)


includes/definitions/rittal-lcp.yaml
```diff
@@ -4,14 +4,16 @@ type: environment
...
             - .1.3.6.1.4.1.2606.7
     - sysDescr:
         - 'Rittal LCP'
-        
\ No newline at end of file
+discovery_modules:
+    storage: 0
+    mempools: 0
+    processors: 0
+over:
```
laf: Can you move this back above discovery, just to keep things standard.

includes/discovery/sensors/entity-sensor.inc.php
```diff
@@ -109,6 +109,21 @@
                 } $descr = preg_replace('/[T|t]emperature[|s]/', '', $descr);
             }
 
+            if ($device['os'] == 'rittal-lcp') {
```
laf: Is this 100% applicable to all Rittal-lcp devices or based on the firmware?

FTBZ: Probably, because the LCP is only one product of the line at this time.

includes/definitions/rittal-lcp.yaml
FTBZ: I removed this because the LCP is an appliance Cooling Unit and shows information about memory, storage and CPU with this enabled. We have no interaction with this information and we can't change anything on this side. So it's not a good information to show that, in my opinion. In my case on our two LCP the memory is saturated even when the unit is not running. So it's probably not important to show this information. 🤔

FTBZ:
@laf An idea how to change the default over graph that can't be edited for devices using the Entity MIB?

laf:
You can change the over graph to any graphs as far as I know, just use the type= part of the graph url

murrant:
@FTBZ might have ran into the OS defs cache.   I might rip that out :)

FTBZ:
@murrant You're right, this morning it's working and I changed nothing. Thanks guys.

PR #6962, opened 2017-07-05
Add field 'co' in databases for OpenTSDB 

yac01:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

PR #7022, opened 2017-07-13
feature: Send collected data to OpenTSDB

yac01:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


build.sql
```diff
@@ -10,7 +10,7 @@ CREATE TABLE IF NOT EXISTS `bill_ports` ( `bill_id` int(11) NOT NULL, `port_id`
...
...` varchar(16) COLLATE utf8_unicode_ci NOT NULL DEFAULT 'udp', `timeout` int(11) DEFAULT NULL, `retries` int(11) DEFAULT NULL, `bgpLocalAs` varchar(16) CHARACTER SET latin1 DEFAULT NULL, `sysDescr` text CHARACTER SET latin1, `sysContact` text CHARACTER SET latin1, `version` text CHARACTER SET latin1, `hardware` text CHARACTER SET latin1, `features` text CHARACTER SET latin1, `location` text COLLATE utf8_unicode_ci, `os` varchar(32) COLLATE utf8_unicode_ci DEFAULT NULL, `status` tinyint(4) NOT NULL DEFAULT '0', `ignore` tinyint(4) NOT NULL DEFAULT '0', `disabled` tinyint(1) NOT NULL DEFAULT '0', `uptime` bigint(20) DEFAULT NULL, `agent_uptime` int(11) NOT NULL, `last_polled` timestamp NULL DEFAULT NULL, `last_polled_timetaken` double(5,2) DEFAULT NULL, `last_discovered_timetaken` double(5,2) DEFAULT NULL, `last_discovered` timestamp NULL DEFAULT NULL, `purpose` varchar(64) COLLATE utf8_unicode_ci DEFAULT NULL, `type` varchar(20) COLLATE utf8_unicode_ci NOT NULL, `serial` text COLLATE utf8_unicode_ci, PRIMARY KEY (`device_id`), KEY `status` (`status`), KEY `hostname` (`hostname`), KEY `sysName` (`sysName`), KEY `os` (`os`) ) ENGINE=InnoDB  DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
```
laf: You've still got changes here :(

laf:
Tagging #6962 for future reference.

build.sql
yac01: yes it me who add the field co. like this when we build shema databases they will create 
What is the problem here carrying I have run validate.php and it's correct 
corr

murrant:
@yac01 Perhaps you should split the customer identification off into a separate pull request?  I think there are more changes needed there, but the opentsdb support is close to ready.

yac01:
@murrant  
You are right that is why you could merge my request with the master and I will think about how to add the customer identité field in the shema sql and I will make another request tomorrow to add it, 
Just one last question about the configuration of opentsdb, should I put it as default in default.inc.php and put all a false or leave it like that and the one who want to use it will have that add this configuration in config.php As explained in the doc ?? thank you

laf:
I'm ok with this as well now, I'll edit the 'co' stuff out of docs for now though.

laf:
You can leave the defaults out, people will add this to config.php

PR #7028, opened 2017-07-13
Update  Allow discovery of IAP radios on Aruba Virtual Controller - Arubaos.php

kkrumm1:
Allow discovery of IAP radios on Aruba Virtual Controller -

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/OS/Arubaos.php
+++ b/LibreNMS/OS/Arubaos.php
@@ -101,23 +101,27 @@ class Arubaos extends OS implements
     {
         $data = snmpwalk_cache_oid_num($this->getDevice(), $oid, array(), 'AI-AP-MIB');
 
-        @@ -105,6 +105,7@@ class Arubaos extends OS implements
-         foreach ($data as $oid => $entry) {
-             $oid_parts = explode('.', $oid);
-             $index = end($oid_parts);
-+            $tmp_index = "$oid.$index";
- 
-             if ($type == 'frequency') {
-                 $current = WirelessSensor::channelToFrequency($this->decodeChannel($entry[$oid]));
-@@ -117,7 +118,7 @@ class Arubaos extends OS implements
-                 $this->getDeviceId(),
-                 $oid,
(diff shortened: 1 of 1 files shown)


murrant:
@kkrumm1 looks like you pasted the diff in instead of applying it :)

murrant:
@kkrumm1 I fixed up your changes I think , please check it

Honestly when you get into submitting code changes, it is just easier to use the command line (to me at least).  I suggest you give this a read: http://docs.librenms.org/Developing/Using-Git/

:)   We're happy to answer questions.

kkrumm1:
Thanks murrant, will do..i got some reading to do. 
I just did pull test and got this message "<stdin>:44: trailing whitespace."  not sure if something to be concerned about or not. 
The Pull request did work tho.  👍

murrant:
If you want to for practice, you can discard this and try again :)

kkrumm1:
so if i wanted to add the ability to see how many wireless clients are connected.... looking the Aurba SNMP mib guide it says it should be this OID. 1.3.6.1.4.1.14823.2.3.3.1.2.4. 

Could i edit it here like this?

laf:
The file already has a `discoverWirelessClients()`. Is this not doing what you need?

kkrumm1:
It's not pulling that data. :(

laf:
The current one looks valid as well. The one you've mentioned, I don't see that in our MIB, do you have an up to date MIB?

kkrumm1:
i think they are both vailed one is for the normal AP and mine is for IAP. 
here is the Aurba IAP mib updated this month. https://gist.github.com/kkrumm1/df47e0869969194843c0de2df70dab86#file-arubaiapmib

i can also get you the updated Aurba OS mib.

laf:
Ah ok, I guess that can be added. We can merge this for now and then add further support in another PR.

kkrumm1:
ok sounds good 👍

PR #7030, opened 2017-07-14
tests: Tests for errors in mib files

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@laf didn't make an attempt to fix any of the issues.  What do you think?

murrant:
@laf `phpunit --group mibs`

laf:
You've got that excluded though in phpunit.xml?

murrant:
Yes, it takes awhile to run and has lots of failures.

I was thinking of it more as an optional tool, we could remove it from excluded if it is ever all cleaned up.

murrant:
@laf  I changed my mind.  Removing duplicates is not safe.  I spot checked pbn + bdcom :(

laf:
@murrant how come?

murrant:
NMS-PROCESS-MIB might not be found  and it can't find a module it is importing because it is in a subdirctory.  Which isn't defined in all calls.  I'm just a little unsure.

laf:
That mib only imports NMS-TC and NMS-SMI which are both in the root as well.

murrant:
NMS-SMI imports PBN-ROOT, which is in pbn folder.

laf:
Ok. How useful is this then :(

I'd really like to tidy it up. Maybe if we just do the rename for now?

murrant:
I would like to merge the cleanup, just in a separate PR.

This is basically a tool to check mins for us, maybe we enabled it by default.

murrant:
But I have some ideas to clean this up a little more.

laf:
OK, merge away when you're ready.

PR #7033, opened 2017-07-14
added agent setup to mkdocs

tslytsly:
noticed that the Agent Setup doc is not listed under Extensions, but can be found by searching. Added to mkdocs.yml

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -42,6 +42,7 @@ pages:
     - Extensions/Applications.md
     - Extensions/Authentication.md
     - Extensions/Auto-Discovery.md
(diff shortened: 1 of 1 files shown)


murrant:
I kind of hid it on purpose so it was only linked from the applications page.  But I guess it does add info like processes...

tslytsly:
fair enough, if there is a good reason to hide it.

laf:
At the moment the Application doc isn't clear on linking to the Agent setup so I'm also ok with adding this for now until a more complete tidy up can be done.

We also probably need to standardise on some of the wording in this.

PR #7047, opened 2017-07-17
feat: ADD co field for Opentsdb

yac01:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/sensors/fanspeed/nos.inc.php
+++ b/includes/discovery/sensors/fanspeed/nos.inc.php
@@ -15,15 +15,18 @@ $oids = snmp_walk($device, '.1.3.6.1.4.1.1588.2.1.1.1.1.22.1.2', '-Osqn');
 $oids = trim($oids);
 foreach (explode("\n", $oids) as $data) {
     $data = trim($data);
-    if ($data and $data[37] == "2") {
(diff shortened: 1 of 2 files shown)


murrant:
Now, what does this `co` field do and why is it on the device table?

yac01:
@murrant  I am updating the doc and  I explain what the field co do

yac01:
hello @laf  @murrant  the field co is for the cutomer identité for who wante to suffix her metrics by this.
and you can to add for eche device her customer identité because when you use grafana Librenms genérate metrics with long queries and it is a problem in some cases because grafana loses time to look for the metric in the database opentsdb and sometimes it put error request so to avoid it is better to suffix by an object code Or the name of customer like this we win the speed and avoid error request

murrant:
As mentioned on IRC, this needs a way to modify the value from the webui.

Would also be good if it applied to other data stores, but that is not required. ;)

yac01:
for the moment i can't do this but when i will have time i will  do it

laf:
@yac01 Any update to this?

yac01:
hello @laf  yes i have started one page that by what you can modify all database but i have not finishe this page because I was obliged to work on the migration from observium to librenms and because that as like all igration i had to resolve some probleme for some device and réhabilitated somme plugin of observium to librenms but i think i will continue that this week

PR #7050, opened 2017-07-18
feature: Added script to test alerts

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Added a cli script to force send an alert. The alert must be currently active but this gives you a forced cli approach to being able to debug.


murrant:
Trying to test this in my dev box, which may not have alerting properly set up.

When running the script `./scripts/test-alert.php -r 1 -h 1` I get no output.

laf:
It needs to be an active alert that is triggering right now.

murrant:
Oh, I probably should have read the docs first :)

I made a quick change, seem reasonable?

laf:
Perfectly :)

PR #7058, opened 2017-07-19
newdevice: Add state sensor for HPE MSL

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/discovery/hpe-msl.yaml
+++ b/includes/definitions/discovery/hpe-msl.yaml
@@ -1,0 +1,19 @@
+mib: SEMI-MIB
+modules:
+    sensors:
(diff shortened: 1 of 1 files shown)


includes/definitions/discovery/hpe-msl.yaml
```diff
@@ -0,0 +1,19 @@
+mib: SEMI-MIB
+modules:
+    sensors:
+        state:
+            data:
+                -
+                    oid: hpHttpMgHealth
+                    value: hpHttpMgHealth
```
laf: You should be able to delete value and state_name as it's the same as oid.

PR #7059, opened 2017-07-20
Added log events for jira transport

aldemira:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/alerts/transport.jira.php
+++ b/includes/alerts/transport.jira.php
@@ -26,6 +26,7 @@ if($obj['status'] == 0 && $obj['msg'] != 'This is a test alert') {
     return True;
 }
 
+$device = device_by_id_cache($obj['device_id']); // for event logging
 
 $username = $opts['username'];
 $password = $opts['password'];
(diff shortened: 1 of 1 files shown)


includes/alerts/transport.jira.php
```diff
@@ -58,7 +59,10 @@
 $ret  = curl_exec($curl);
 $code = curl_getinfo($curl, CURLINFO_HTTP_CODE);
 if( $code == 200 ) {
-	return true;
+    $jiraout = json_decode($ret, true);
+    log_event("Created jira issue ".$jiraout['key']." for ".$device, $device, 'poller', 5);
```
laf: We already record alerts by transport. I don't think we should have individual log_events in here.

You could convert them to d_echo() for debug or we'd need to find a way to allow the return of some data to be used in the central log_event() for alerts.

aldemira: Never mind then, I'll just delete it until we find something.

aldemira: I'm sorry, but I'll have to reopen this. There's bug in the transport file. And I'm too lazy to open another pr (as long as you're ok with it)

PR #7062, opened 2017-07-20
Loading speed improvement when viewing syslogs for specific device

EnzoZafra:
When trying to view the syslogs for a specific device 
(for example http://[hostname]/device/device=[device_id]/tab=logs/section=syslog/) 
the dropdown for 'priorities' and 'programs' are not filtered to search only for that specific device_id. This causes the SQL query to be extremely slow and sometimes causes the browser to crash when there are lots of syslog entries in the database.

Also, since the syslog entries in the table themselves are filtered by device_id, I think that the 'programs' and 'priorities' dropdown should be as well; to reflect the entries in the syslog table.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/syslog.inc.php
+++ b/html/pages/syslog.inc.php
@@ -49,7 +49,12 @@ print_optionbar_start();
                         <select name="program" id="program" class="form-control input-sm">
                             <option value="">All Programs</option>
                                 <?php
-                                foreach (dbFetchRows('SELECT DISTINCT `program` FROM `syslog` ORDER BY `program`') as $data) {
+                                $sqlstatement = 'SELECT DISTINCT `program` FROM `syslog`';
+                                if (is_numeric($vars['device'])) {
+                                    $sqlstatement = $sqlstatement . ' WHERE device_id=' . $vars['device'];
+                                }
+                                $sqlstatement = $sqlstatement .' ORDER BY `program`';
+                                foreach (dbFetchRows($sqlstatement) as $data) {
                                     echo '"<option value="'.mres($data['program']).'"';
                                     if ($data['program'] == $vars['program']) {
                                         echo ' selected';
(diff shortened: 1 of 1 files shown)


florianbeer:
Nice! I am going to try this out right now.

florianbeer:
Hmm, can't really say it's faster, but it is certainly much cleaner.
Especially having all of the programs in the dropdown, even the ones that are not relevant to this device, always bothered me a lot. Now it is all neat and pristine 👍

Definitely a vote from me to merge this in!

html/pages/syslog.inc.php
```diff
@@ -49,7 +49,12 @@
                         <select name="program" id="program" class="form-control input-sm">
                             <option value="">All Programs</option>
                                 <?php
-                                foreach (dbFetchRows('SELECT DISTINCT `program` FROM `syslog` ORDER BY `program`') as $data) {
+                                $sqlstatement = 'SELECT DISTINCT `program` FROM `syslog`';
+                                if (is_numeric($vars['device'])) {
+                                    $sqlstatement = $sqlstatement . ' WHERE device_id=' . $vars['device'];
```
laf: You need to pass variables to be used as prepared statements.

Like: https://github.com/EnzoZafra/librenms/blob/69f4e92142efe96f09cfc25722a26aac45e9ac95/html/includes/table/ports.inc.php#L140

You basically pass the variable into an array and pass that to dbFetchRows()

EnzoZafra: Ah yes. That was a stupid mistake.

EnzoZafra:
@florianbeer Yeah, it only helps with speed when there are thousands of programs that aren't related to the device. We found that even though there are only say 5 syslog entries for the device it loads as slow as the overall syslog page.

florianbeer:
> Yeah, it only helps with speed when there are thousands of programs that aren't related to the device. We found that even though there are only say 5 syslog entries for the device it loads as slow as the overall syslog page

Yea, that makes sense. Gets me thinking though: why do you have thousand of different programs? Maybe we can improve the detection and organisation of those strings?

EnzoZafra:
Look at this table: https://pastebin.com/WApYR69S
It looks like all of the devices with > 1000 distinct programs has an 'ios' OS, so it probably has something to do with how its parsed.

When I searched through the syslog for device_id = 1224 (device with 165623 distinct programs) this is a snippet of what I found: https://pastebin.com/p0p2jwrE
Now if we fix the issue in our switches that creates these syslog entries, those programs wouldn't be created in the first place.
The first part of the line is the 'program'. I'm not sure if this is intended but it looks like its creating an index for each entry and therefore creating thousands of distinct programs..

florianbeer:
That is definitely not intended. I suggest you open a new issue, so we can get to the bottom of this. Shouldn't be to hard to fix the parser.

html/pages/syslog.inc.php
```diff
@@ -42,14 +43,20 @@
                         <?php
                         } else {
                             echo '<input type="hidden" name="device" id="device" value="' . $vars['device'] . '">';
+                            $params[] = $vars['device'];
```
laf: It makes much more sense to put this under the `$sqlstatement = $sqlstatement . ' WHERE device_id=?';` lines so it's obvious.

Issue #6993, opened 2017-07-10
Core port not shown in overview 


raymonvdm:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.

```
Component | Version
--------- | -------
LibreNMS  | 61103938dc691bb7faa0207551c4924e54923bad
DB Schema | 199
PHP       | 7.0.18-0ubuntu0.16.04.1
MySQL     | 5.7.18-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3
```

- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).

https://p.libren.ms/view/4e2ac96a
https://p.libren.ms/view/299195f3
https://p.libren.ms/view/4739e2d1

- [x] Please provide as much detail as possible.

The interface description is set to this (which works for observium)

`interface GigabitEthernet0/23
 description "Core: Raymon[1Gbit](Uplink Raymon) {O000002}"`

But the ports are not shown under Ports -> Core 
The Customers are also not shown Ports -> Customers

- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

laf:
Please test this patch:

https://p.libren.ms/view/raw/2b6bc3e2

raymonvdm:
This issue is solved for me now!

murrant:
laf: the \" is because the " is in the string inputed by the user.

Look at this one: `\"Cust: Raymon[1Gbit](crs2.dro2.raqxs.net Eth3 - Trunk) {O000002}`

PR #7063, opened 2017-07-21
fix: Strip " and / from snmpwalk_cache_oid()

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #6993 

I've not done any other snmp functions yet but they probably should be done to save trim() being used everywhere.

### DIFF
--- a/includes/snmp.inc.php
+++ b/includes/snmp.inc.php
@@ -369,6 +369,7 @@ function snmpwalk_cache_oid($device, $oid, $array, $mib = null, $mibdir = null, 
         list($oid,$value)  = explode('=', $entry, 2);
         $oid               = trim($oid);
         $value             = trim($value);
+        $value             = trim($value, '\"');
         list($oid, $index) = explode('.', $oid, 2);
         if (!strstr($value, 'at this OID') && isset($oid) && isset($index)) {
             $array[$index][$oid] = $value;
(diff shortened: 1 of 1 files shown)


murrant:
@laf you could also just modify the previous trim.

snmpwalk_group() trims quotes, but not / why is that needed?

laf:
@murrant I left the old trim in place as that trims the defaults off, if you specify what to trim that doesn't happen anymore does it? I thought it didn't hence adding the new trim.

The / is in the snmpwalk output so seems to be sent back by the device. I've seen this in other walks as well.

murrant:
You can include the defaults.

This is what I have snmpwalk_group()
```
trim($value, "\" \n\r")
```

murrant:
I couldn't find an example of `/` in the linked bug.  I did however see an example of `""/""` where this would cause it to return `""` instead of `"/"`

laf:
The poller output in the bug report shows what is returned now:

`\Core: Raymon[1Gbit](Uplink Raymon) {O000002}\`

Seeing as we also have to trim `"` I've added them both in, I can update and add what you use for `snmpwalk_group` as well.

murrant:
laf: the \" is because the " is in the string inputed by the user and the snmp daemon escapes it...

Look at this one: `\"Cust: Raymon[1Gbit](crs2.dro2.raqxs.net Eth3 - Trunk) {O000002}`

laf:
Which of those are you concerned about as it looks like we should strip them all.

murrant:
laf mostly, the `\` because it should only be trimmed if it precedes `"`

It is likely that particular edge case will never cause us any issues.

laf:
It's a bit of a rabbit whole this.

This is what I've come up with so far but I feel less comfortable about this solution then I do my original:

https://p.libren.ms/view/raw/b6afe033

murrant:
Your paste isn't loading.  What do you think about just trimming: `"\" \\\n\r"` and seeing how it goes?

laf:
Updated

PR #7066, opened 2017-07-21
feature: Config helper to simplify config access

murrant:
Saves from having to check isset() and all that junk.  Allows specifying a default if it isn't set.
Uses dot separated config paths like the database.

Usage:
  Config::get('install_dir')
  Config::get('memcached.enable', false)
  Config::getDeviceSetting($device, 'retries', 'snmp', 5)
  Config::getOsSetting($device, 'nobulk')
  Config::set('poller_modules', array())

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Config.php
+++ b/LibreNMS/Config.php
@@ -1,0 +1,142 @@
+<?php
+/**
+ * Config.php
+ *
+ * Config convenience class to access and set config variables.
+ *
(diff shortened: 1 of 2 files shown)


murrant:
I coded this for the snmp rewrite and have had a few times where I wished it existed in mainline.

Issue #7057, opened 2017-07-19
Raspberry pi sensors not added 


Jo0Lz:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

Component | Version
--------- | -------
LibreNMS  | 1.29-95-g5c46ad1a7
DB Schema | 199
PHP       | 7.1.6-2ubuntu1
MySQL     | 10.1.24-MariaDB-5
RRDTool   | 1.6.0
SNMP      | NET-SNMP 5.7.3

Followed [this](https://community.librenms.org/clicks/track?url=http%3A%2F%2Fdocs.librenms.org%2FExtensions%2FApplications%2F%23raspberry-pi&post_id=956&topic_id=430) tutorial. The raspberry.sh script works correctly, and the information is being "discovered", however sensors are not added to the LibreNMS database.

./raspberry.sh 
![image](https://user-images.githubusercontent.com/19903665/28361930-c579ed9e-6c7a-11e7-84f3-a7d18e989eb4.png)

Proof that the values are discovered with an SNMPwalk:
![image](https://user-images.githubusercontent.com/19903665/28363114-7a78455c-6c7f-11e7-9367-c6f66e945433.png)

`jo0lz@domotica:~ $ snmpwalk -Os -v1 -c public domotica .1.3.6.1.4.1.8072.1.3.2.4.1.2.9.114.97.115.112.98.101.114.114.121
[...]

PR #7068, opened 2017-07-21
fix: Fixed Raspberry Pi sensors support

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7057 

Before we did an snmp query to check if it was a rPi. We still only do one query in each of these and if it's not valid we break out or carry on.

### DIFF
--- a/includes/discovery/sensors/frequency/linux.inc.php
+++ b/includes/discovery/sensors/frequency/linux.inc.php
@@ -3,24 +3,22 @@
  * raspberry pi frequencies
  * requires snmp extend agent script from librenms-agent
  */
-$raspberry = snmp_get($device, 'HOST-RESOURCES-MIB::hrSystemInitialLoadParameters.0', '-Osqnv');
+$sensor_type = "raspberry_freq";
+$oid = '.1.3.6.1.4.1.8072.1.3.2.4.1.2.9.114.97.115.112.98.101.114.114.121.';
 
-if (preg_match("/(bcm).+(boardrev)/", $raspberry)) {
-    $sensor_type = "raspberry_freq";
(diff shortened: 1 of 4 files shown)


includes/discovery/sensors/frequency/linux.inc.php
```diff
@@ -3,24 +3,22 @@
  * raspberry pi frequencies
  * requires snmp extend agent script from librenms-agent
  */
-$raspberry = snmp_get($device, 'HOST-RESOURCES-MIB::hrSystemInitialLoadParameters.0', '-Osqnv');
+$sensor_type = "raspberry_freq";
+$oid = '.1.3.6.1.4.1.8072.1.3.2.4.1.2.9.114.97.115.112.98.101.114.114.121.';
```
murrant: Why wasn't this walked instead of doing for with several gets?

What if .6 is not a valid data point, but .7 is?

laf: That would mean walking the entire table all the time when you only need a couple of values.

I can try and get this walk into pre-cache but it's not a named OID so array returned using numerical OIDs are horrible.

Jo0Lz:
So, I tried this patch...

OSMC:
image

Raspbian:
image

I see codec sensors added, on the OSMC one it added every codec twice. No voltages etc. yet.
The double temperature got added when trying to resolve the issue manually by trying things out [this topic](https://community.librenms.org/t/raspberry-pi-temperature-voltage-freq/)

laf:
@murrant Updated now. One thing with this though, I doubt the state worked before either as the output from the snmp query is a string which is never converted as we don't use a MIB here.

So the raspberrypi.sh script needs updating to output numerical info.

laf:
https://github.com/librenms/librenms-agent/pull/140

Could be done better but I have little time or care for the application support right now :)

PR #7077, opened 2017-07-25
Added mempool support for Avaya ERS 3500

tslytsly:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/mempools/avaya-ers.inc.php
+++ b/includes/discovery/mempools/avaya-ers.inc.php
@@ -3,15 +3,16 @@
 $OID = '.1.3.6.1.4.1.45.1.6.3.8.1.1.12';
 
 if ($device['os'] == 'avaya-ers') {
-    // Memory information only known to work with 5500 and 5600 switches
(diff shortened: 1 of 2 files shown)


murrant:
Hmm, why are we using a regex against sysDescr here?  Odd.  Also, for polling it should check the DB.

Your PR is fine, but this code is troubling.

laf:
@murrant The regex is already done and looks like it's to limit the devices that this support is for.

tslytsly:
@laf and @murrant 
laf is correct, although I don't see a need to limit it? We use a lot of Avaya products, and as far as I know that is the correct OID for all ERS devices.

laf:
@tslytsly remove the if check then :)

tslytsly:
@laf ok then, I'll close this PR and open a new one then.
The MIBs are from well before Avaya bought Nortel, before Nortel bought Baystack I think.
So I doubt there are many switches out there in the wild that don't support these MIBs.
I'll do the same for CPU and Temp, all part of the same MIBs.

PR #7079, opened 2017-07-25
Add if statement for ROS os when querying remote port id for LLDP map

EnzoZafra:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Siemens Ruggedcom switches does not have the functionality to give ports/interfaces descriptions. When you do an snmpwalk on oid `ifDescr' for each port, it is empty and therefore inserts a whitespace into the 'ifDescr' field in the ports table.

When you do an snmpwalk on oid 'lldpRemoteSystemsData', you get the following: https://pastebin.com/LbechkBG

The `discovery-protocols` module will then query the database for a remote_port_id using this `ifDescr` which will always return the first port (port 1) because all of the `ifDescr` for each port on that device is a whitespace. I created an if statement that will search the `ifName` only instead of `ifDescr` OR `ifName`.

Instead of an if-statement, another solution I suggest is re-arranging the query from this:

`$remote_port_id = dbFetchCell('SELECT \`port_id\` FROM \`ports\` WHERE (\`ifDescr\` = ? OR \`ifName\` = ? OR \`ifDescr\` = ? OR \`ifName\` = ? OR \`ifPhysAddress\` = ?) AND \`device_id\` = ?', array($if, $if, $id, $id, $remote_port_mac_address, $remote_device_id));`

to this:

`$remote_port_id = dbFetchCell('SELECT \`port_id\` FROM \`ports\` WHERE (\`ifName\` = ? OR \`ifDescr\` = ? OR \`ifDescr\` = ? OR \`ifName\` = ? OR \`ifPhysAddress\` = ?) AND \`device_id\` = ?', array($id, $id, $if, $if, $remote_port_mac_address, $remote_device_id));`

but I'm not sure how this will break the map for other OS'. 

We can also store the `ifName` as the `ifDescr` in the `ports` module if the `ifDescr` oid is empty. That way, it will fix any other issues caused by `ifDescr` being empty. Let me know what is preferred

### DIFF
--- a/includes/discovery/discovery-protocols.inc.php
+++ b/includes/discovery/discovery-protocols.inc.php
@@ -181,7 +181,11 @@ if ($device['os'] == 'pbn' && $config['autodiscovery']['xdp'] === true) {
                     if ($remote_device_id) {
                         $if             = $lldp['lldpRemPortDesc'];
                         $id             = $lldp['lldpRemPortId'];
-                        $remote_port_id = dbFetchCell('SELECT `port_id` FROM `ports` WHERE (`ifDescr` = ? OR `ifName` = ? OR `ifDescr` = ? OR `ifName` = ? OR `ifPhysAddress` = ?) AND `device_id` = ?', array($if, $if, $id, $id, $remote_port_mac_address, $remote_device_id));
+                        if ($device['os'] == 'ros') {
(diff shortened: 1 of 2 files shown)


laf:
But they return ifName and / or ifAlias? I reckon it might be better to fix this in ports polling.

I don't think changing the order is the right thing to do, ifDescr is/should be a standard value returned so this is an edge case rather than the norm

EnzoZafra:
Yup, there is an ifName and/or an ifAlias per port. My idea is to set the ifDescr as the ifName if it's empty. Should I do this only for the 'ROS' OS or just do it for all?

laf:
I would do it for all. We do the same for ifAlias / ifName already.

PR #7080, opened 2017-07-25
Update to ports polling

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I'll put inline comments in place.

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7080)

### DIFF
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -66,20 +66,6 @@ $stat_oids_db_extended = array(
     'ifOutMulticastPkts',
 );
 
-$cisco_oids = array(
-    'locIfHardType',
-    'locIfInRunts',
-    'locIfInGiants',
-    'locIfInCRC',
-    'locIfInFrame',
-    'locIfInOverrun',
-    'locIfInIgnored',
-    'locIfInAbort',
-    'locIfCollisions',
-    'locIfInputQueueDrops',
-    'locIfOutputQueueDrops',
-);
-
 $pagp_oids = array(
     'pagpOperationMode',
 );
@@ -172,6 +158,7 @@ $shared_oids = array(
 echo 'Caching Oids: ';
 $port_stats = array();
 $data       = array();
+$polled     = null;
 
 if ($device['os'] === 'f5' && (version_compare($device['version'], '11.2.0', '>=') && version_compare($device['version'], '11.7', '<'))) {
     require_once 'ports/f5.inc.php';
@@ -201,6 +188,7 @@ if ($device['os'] === 'f5' && (version_compare($device['version'], '11.2.0', '>=
                         $port_stats[$i] = $data[$i];
                     }
                     $port_stats = snmp_get_multi($device, $oids, '-OQUst', 'IF-MIB', null, $port_stats);
+                    $port_stats[$i]['polled'] = time();
                 }
             }
         }
@@ -219,6 +207,7 @@ if ($device['os'] === 'f5' && (version_compare($device['version'], '11.2.0', '>=
                 $port_stats = snmpwalk_cache_oid($device, $oid, $port_stats, 'IF-MIB', null, '-OQUst');
             }
         }
+        $polled = time();
     }
 }
 
@@ -284,8 +273,6 @@ if ($device['os_group'] == 'cisco' && $device['os'] != 'asa') {
 } elseif ($device['os'] != 'asa') {
     $port_stats = snmpwalk_cache_oid($device, 'dot1qPvid', $port_stats, 'Q-BRIDGE-MIB');
 }//end if
-
-$polled = time();
 
 // End Building SNMP Cache Array
 d_echo($port_stats);
@@ -430,7 +417,8 @@ foreach ($ports as $port) {
             $this_port['ifDescr'] = $matches[1];
         }
 
-        $polled_period = ($polled - $port['poll_time']);
+        $tmp_polled = $this_port['polled'] ?: $polled;
+        $polled_period = ($tmp_polled - $port['poll_time']);
 
         $port['update'] = array();
         $port['update_extended'] = array();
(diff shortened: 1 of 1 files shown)


includes/polling/ports.inc.php
```diff
@@ -66,20 +66,6 @@
     'ifOutMulticastPkts',
 );
 
-$cisco_oids = array(
```
laf: Not used at all

includes/polling/ports.inc.php
```diff
@@ -201,6 +188,7 @@
                         $port_stats[$i] = $data[$i];
                     }
                     $port_stats = snmp_get_multi($device, $oids, '-OQUst', 'IF-MIB', null, $port_stats);
+                    $port_stats[$i]['polled'] = time();
```
laf: Before hand, we obtained the polled time for the ports way after all the snmpwalks when we actually only want to do it after ifEntry/ifXEntry. So we now set this either on a per port basis for selected ports polling or right after the if(X)Entry runs so we get as close to accurate time as possible.

includes/polling/ports.inc.php
```diff
@@ -472,14 +460,22 @@
...
-        // use HC values if they are available
-        foreach ($hc_mappings as $hc_oid => $if_oid) {
-            if (isset($this_port[$hc_oid]) && $this_port[$hc_oid]) {
-                d_echo("$hc_oid ");
-                $this_port[$if_oid] = $this_port[$hc_oid];
-            } else {
-                d_echo("$if_oid ");
+        $hc_enabled = false;
+        // use HC values if they are available and not disabled
+        if ($config['os'][$device['os']]['no_hc'] != true) {
```
laf: Added the option to be able to skip using hc counters. I've helped someone recently where the hc values are utterly broken.

includes/polling/ports.inc.php
```diff
@@ -641,6 +631,20 @@
 
             $oid_prev = $oid.'_prev';
             if (isset($port[$oid])) {
+                if ($device['uptime'] < $config['rrd']['step']) {
```
laf: This block attempts to address counter rollover for ports in 32 bit mode.

If the device hasn't recently rebooted and the current Octet value is < then the previous something has gone wrong. So we subtract the previous Octet value from the max possible for 64bit and then add the current one onto that value.

laf:
I might switch the rolled calculation to this one: https://blog.chasefox.net/calculating-ifinoctets/

Anyone got any comments on that?

kkrumm1:
is this only for cisco devices? or all devices?

laf:
@mathias-combelle It's an option per OS, so you can set:

`$config['os']['merakimx']['no_hc'] = true;` in config.php. Feel free to test this pull request and report back.

laf:
@kkrumm1 the general changes are for all devices. The one to skip using HC counters is user configurable per OS.

mathias-combelle:
@laf : Is it in the update or should I test it manually ?
I'm on DEV auto-update mode and do not see the new code.

Thanks,

murrant:
@mathias-combelle This is code that is being reviewed for inclusion.

To test it, follow the instructions at the top of the ticket:
```
If you would like to test this pull request then please run: ./scripts/github-apply <pr_id>, i.e ./scripts/github-apply 7080
```

raphbap:
Hi, I tried it but unfortunately it stopped the polling of Meraki devices (got the notifications + no more data).

When I disabled the option in config.php with this command: 
`$config['os']['merakimx']['no_hc'] = false;`, I kept getting the following pop-up message (although Meraki devices were polled again): 

> It appears as though you have some devices that haven't completed polling within the last 15 minutes, you may want to check that out :)

laf:
please post the output on pastebin of `./poller.php -h HOSTNAME -d -m ports`

raphbap:
Here it is :) : https://pastebin.com/QQ7gKN9F

laf:
Remove the patch, you don't have a module installed (which I thought was part of php core). `./scripts/github-remove -d`

laf:
Updated to remove dependancy on gmp_sub()

includes/polling/ports.inc.php
```diff
@@ -641,6 +631,25 @@
...
+                        $max_64 = '18446744073709551616';//Unused at the moment.
+                        $max_32 = '4294967296';
+                        if ($hc_enabled !== true) {
+                            if (function_exists('gmp_sub')) {
+                                // We store the new rolled version for use in rrd. DO NOT store in the DB as we need
+                                // to calculate the value again on the next poll.
+                                $this_port[$oid . '_rolled'] = (gmp_sub($max_32, $port[$oid])) + $this_port[$oid];
+                            } else {
+                                // No GMP http://uk1.php.net/manual/en/book.gmp.php function
+                                $this_port[$oid . '_rolled'] = ($max_32 - $port[$oid]) - $this_port[$oid];
```
murrant: Shouldn't the second `-` be `+`?

includes/polling/ports.inc.php
murrant: You could use the new config class here:
`Config::getOsSetting($device, 'no_hc', false)`

murrant:
I thought RRD automatically handles rolled over counters.

laf:
As far as I'm aware no, if you push more than 125Mbps on v1 snmp (or non-HC counters on any version) then you get a flat capped graph or drops in data.

includes/polling/ports.inc.php
laf: Corrected.

I actually had 64Bit in there originally as well when testing for another fix but it produced crazy numbers. I think that was due to the device itself so we could possibly add rollover for 64Bit and at which point GMP is better for precision.

I can drop for now if you like.

laf:
bump

mathias-combelle:
I tried : 
/scripts/github-apply 7080
But got the same issue (looks like it still search for this missing gmp_sub() module).

Then, after a "./scripts/github-remove -d" I tried with : 
./scripts/github-apply 5926
But got an error applying the patch :/

"
error: patch failed: scripts/github-remove:**_xx_**
error: scripts/github-remove: patch does not apply
"
I will wait for a new update

laf:
> But got the same issue (looks like it still search for this missing gmp_sub() module).

`if (function_exists('gmp_sub')) {`

Nope, we check first. Seems like you've not managed to re-apply the patch. Use github-remove then do `git status` to make sure your branch is clean.

mathias-combelle:
"On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working tree clean"

Then /github-apply 5926 -> Fail

laf:
The branch is clean against the current master so either your install is too old or another reason.

You could just download the file and copy it over.

mathias-combelle:
Just run validate.php, made some changes and force the update. Now it has applied the changes for sure :)

I try again and let you know, thanks for your support and tips

laf:
@librenms/reviewers Any comments to this or I'll be merging in in 48 hours.

murrant:
The code itself looks good.  I am just unsure about the usage of gmp.  But not enough to not merge, so +1

Rosiak:
I'll see if I can test this tomorrow.

laf:
Don't bother for now @Rosiak.

mathias-bap-git:
Hi veryone, has this feature been pushed ?
It's still indicating "Merge in 48hours" for a while. 
I still have gaps in all my graphs (Meraki...) so I was really looking forward to this solution.

Thanks in advance,

f0o:
It's been closed rather than merged, chances are that there was a blocker somewhere.

I haven't followed this PR, best to ask laf directly in discord or IRC :)

mathias-bap-git:
Thanks, that's what I was affraid of. 
I will ask him on IRC, thanks for the reply !

murrant:
@mathias-bap-git I thought your issue was 32bit counters on the Meraki, this is for 64bit counter overflows, which would not help for 32bit counters.

PR #7081, opened 2017-07-26
Removed check for switch model or firmware version for Avaya ERS switches

tslytsly:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/mempools/avaya-ers.inc.php
+++ b/includes/discovery/mempools/avaya-ers.inc.php
@@ -3,15 +3,16 @@
 $OID = '.1.3.6.1.4.1.45.1.6.3.8.1.1.12';
 
 if ($device['os'] == 'avaya-ers') {
-    // Memory information only known to work with 5500 and 5600 switches
-    if (preg_match('/5[56][0-9][0-9]/', $device['sysDescr'])) {
+    // Memory information only known to work with 3500, 5500 and 5600 switches
+    if (preg_match('/[35][56][0-9][0-9]/', $device['sysDescr'])) {
         // Get major version number of running firmware
         $fw_major_version = null;
         preg_match('/[0-9]\.[0-9]/', $device['version'], $fw_major_version);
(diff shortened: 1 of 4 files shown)


tslytsly:
There is no reason to check the model or firmware.
These MIBs should be present in all ERS switches.
The MIB has been around since before Baystack were bought out.

includes/discovery/mempools/avaya-ers.inc.php
```diff
@@ -2,25 +2,12 @@
 
 $OID = '.1.3.6.1.4.1.45.1.6.3.8.1.1.12';
 
-if ($device['os'] == 'avaya-ers') {
```
laf: You still need to leave that check in place otherwise this runs for every OS.

includes/discovery/processors/avaya-ers.inc.php
```diff
@@ -1,15 +1,9 @@
 <?php
 
-if ($device['os'] == 'avaya-ers') {
-    // Processor information only know to work with 3500, 5500 and 5600 switches
-    if (preg_match('/[35][56][0-9][0-9]/', $device['sysDescr'])) {
-        $procs = snmp_walk($device, '.1.3.6.1.4.1.45.1.6.3.8.1.1.6', '-Osqn');
```
laf: You've removed this from your update so it won't run :)

And you also need the `if ($device['os'] == 'avaya-ers') {` still.

tslytsly: Oh yea... doh!

tslytsly:
@laf I think that's right.

I thought that the script only ran if the device was an ERS. 
Do these scripts get run against every device?

laf:
Just the missing snmpwalk still in `includes/discovery/processors/avaya-ers.inc.php`.

But yes, you need to have the checks for OS in as all files are run against all devices, old legacy bad bad code :(

PR #7085, opened 2017-07-27
fix: Add device status reason to up messages.

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Isn't that what already happens?

murrant:
No, $response['status_reason'] is empty when the check succeeds.

laf:
If it's empty and you aren't setting it to something non-empty then why is it now going to contain something.

murrant:
@laf I am using $device['status_reason'], which will be the previous status reason ;)

PR #7086, opened 2017-07-27
fix: Correction of the echo of the opentsdb function to debug

yac01:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/opentsdb.inc.php
```diff
@@ -18,9 +18,12 @@ function opentsdb_update($device, $measurement, $tags, $fields)
     if ($config['opentsdb']['enable'] == true) {
         if ($opentsdb != true) {
             $opentsdb = fsockopen($config['opentsdb']['host'], $config['opentsdb']['port']);
-            d_echo("Connection made to OpenTSDB");
+            d_echo("Connection made to OpenTSDB\n");
+        }
+        if ($opentsdb == true) {
```
laf: Isn't this now going to print out the connection is made twice, the first as it establishes the connection above and again because the variable is now set.

laf: Probably easier to do `} else {` from the `if ($opentsdb != true) {` and do the echo.

yac01:
what do you think about like this  
because if ai do like what do you say 
imagine opentsdb == false it will go in the loop 
{ $opentsdb = fsockopen($config['opentsdb']['host'], $config['opentsdb']['port']);
    d_echo("Connection made to OpenTSDB"); }

and try to do the connection and after print the connection has mad, imagine when  it do the line 
$opentsdb = fsockopen($config['opentsdb']['host'], $config['opentsdb']['port']); and that has field it will do the next line and print connection made to  opentsdb this is why we need to do an ather test after the line 
$opentsdb = fsockopen($config['opentsdb']['host'], $config['opentsdb']['port'] 

but you have right it will write twice connection has made this why i delete the first and remade the test after with the second test loop  

I hope to be clear otherwise I re-explain

laf:
It's already been merged now so I hope the last commit you did was what you've proposed - it looks good to me.

PR #7084, opened 2017-07-26
Update IRC-Bot.md

allano337:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

In the doc's doesn't it say to run ./irc.php to start the IRC-Bot, I have added it to the doc's

Also one of our engineers made a very basic systemd start up script in /etc/systemd/system/librenms-irc.service set to start at boot but it can be stopped or started just like any other systemd script such as systemctl start librenms-irc.service

librenms-irc.service
[Unit]
Description=Librenms IRC bot

[Service]
ExecStart=/opt/librenms/irc.php
Restart=always

[Install]
WantedBy=default.target

laf suggested that we should contribute this back to the code base by submitting a pull request.

Thanks

PR #7087, opened 2017-07-27
feature: Create librenms-irc.service

allano337:
basic systemd start up script that goes in /etc/systemd/system/librenms-irc.service it is set to start at boot, but it can be stopped or started just like any other systemd script ie. systemctl start librenms-irc.service

I will update doc's on how to implement for systemd on pull request #7084

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/misc/librenms-irc.service
+++ b/misc/librenms-irc.service
@@ -1,8 +1,8 @@
 # /etc/systemd/system/librenms-irc.service 
 
-librenms-irc.service
 [Unit]
 Description=Librenms IRC bot
+After=network.target
 
 [Service]
 ExecStart=/opt/librenms/irc.php
--- a/misc/librenms-irc.service
+++ b/misc/librenms-irc.service
@@ -1,0 +1,12 @@
+# /etc/systemd/system/librenms-irc.service 
+
+librenms-irc.service
+[Unit]
+Description=Librenms IRC bot
+
+[Service]
+ExecStart=/opt/librenms/irc.php
+Restart=always
+
+[Install]
+WantedBy=default.target


misc/librenms-irc.service
```diff
@@ -0,0 +1,12 @@
+# /etc/systemd/system/librenms-irc.service 
+
+librenms-irc.service
+[Unit]
+Description=Librenms IRC bot
+
+[Service]
+ExecStart=/opt/librenms/irc.php
```
murrant: Can you extract `/opt/librenms` to this:
Environment=BASE_DIR=/opt/librenms

Might be a good idea to set this to allow central configuration:
EnvironmentFile=/etc/sysconfig/librenms

But, I can do that later to unify the systemd scripts.

laf: @murrant I think it's worth doing before this is merged in. It only means people have to re-deploy this when it is updated which isn't part of our updates. I know it's not _needed_ but it would be good to start it right.

aldemira: Just my 2 cents, on Debian based systems central config directory is /etc/default. Also I'd like to prepare sysv init scripts for any daemons as well, so are there any more daemons?

murrant: `/etc/sysconfig/librenms` was just a guess.  Is there a more distro neutral location?

aldemira: Your guess is right for rpm based systems. Having a config file for not letting people to fiddle with init files (or unit files whatever you call it) is the right way, however, I think we should leave this to package maintainers, there's no distro neutral location that I know of.

laf: /etc/default/ exists in centos so we can choose that for now.

PR #7088, opened 2017-07-28
Create alteonos.yaml

sfini:
Add Support for radware/Alteon application Switch

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
You'll need to add a snmprec file for each entry in your discovery array.

laf:
@sfini Please also remove the comments from the yaml file.

sfini:
@murrant 
Is snmpwalk ok ?

laf:
@sfini We don't need snmpwalk, you just need to create a test unit file as per: http://docs.librenms.org/Developing/os/Test-Units/

sfini:
@murrant 
Hi Tony .. Added radware.snmprec ...  Is it ok ?

sfini:
@murrant 

https://github.com/librenms/librenms/blob/aff2bb084f0008664bbadd2a14e6bd27a7de9a97/tests/snmpsim/radware.snmprec

laf:
You've not pushed the file into this branch but into a new one :)

In https://github.com/sfini/librenms - Click on the master branch in the top left:

image

Select patch-1 branch, now add the file to that and it should push it here.

sfini:
@murrant 

All Done ... Thank you..

includes/definitions/radware.yaml
```diff
@@ -0,0 +1,13 @@
+os: radware
+text: RADWARE
+type: network
+icon: radware
```
laf: Sorry, one last thing I've noticed is that the icon radware doesn't exist. You will also need to add this in:

http://docs.librenms.org/Developing/os/Initial-Detection/#icon-and-logo

sfini:
@laf 
Done...

includes/definitions/radware.yaml
sfini: Done :+1:

laf:
One last question, is radware the right name for the os? That seems to be the vendor name which we don't usually use. I.e Cisco have Cisco IOS so the os name is ios. What's the case here?

sfini:
@laf 
Hi Neil .. You're right .. The correct OS for this type is AlteonOS ...
There are another branches of RADWARE co. , that have other OS (like APSoluteOS, etc.)
So, i think is better to have different type of yaml files as you have for cisco ....
I have renamed radware.yaml in alteonos.yaml, and made all the changes.
I also pushed the alteonos.rb to the oxidized's github.

sfini:
@laf 
Do i have to upload mib too ?

laf:
If you aren't making use of the mib (you aren't so far) then no.

PR #7095, opened 2017-07-31
fix: add html line breaks for html emails

aldemira:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -891,9 +891,11 @@ function send_mail($emails, $subject, $message, $html = false)
         $mail->XMailer = $config['project_name_version'];
         $mail->CharSet = 'utf-8';
         $mail->WordWrap = 76;
-        $mail->Body = $message;
         if ($html) {
             $mail->isHTML(true);
+            $mail->Body = nl2br($message);
+        } else {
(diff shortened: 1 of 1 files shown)


laf:
I'm not sure about this change, this will add additional line breaks for users who've already used `<br />` but also formatted the templates for readability by using newlines.

Imho, users should format the template as expected for html rather than rely on us working out what should be formatted and what shouldn't.

aldemira:
I see your point, sure. Default template looks even uglier with html turned on, that was my starting point.

laf:
@aldemira That makes sense. You can fix that in https://github.com/librenms/librenms/blob/master/includes/alerts.inc.php#L449

Do the check for html enabled emails after that and if enabled run through `nl2br()`. That will get you the same result without breaking anything.

aldemira:
What about putting the default template in alert templates page, obviously non-deletable and non-attachable?

laf:
You're more than welcome to :) You'll need to also ensure it can be reverted to default if/when someone makes breaking changes as well.

imho, a simple nl2br() in the code seems the easier and less problematic approach.

PR #7103, opened 2017-08-02
fix: check session directory is writable before install.php

murrant:
Drop pear check.
A few html/bootstrap cleanups too.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/install.php
+++ b/html/install.php
@@ -92,12 +92,11 @@ $complete = 1;
 
 ?>
 <!DOCTYPE HTML>
-<html>
+<html lang="en">
 <head>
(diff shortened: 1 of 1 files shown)


laf:
How come the pear check is dropped? Using it is part of the install guide which users should have already completed but also may not have.

murrant:
Well, I might have another PR...

sTywin:
On my default installation, this fails since `session_save_path()` returns an empty string, and thus `is_writable("")` returns `0`. In this default condition, PHP will use `/tmp` as the session save path, which is indeed writable.

PR #7106, opened 2017-08-02
feature: New IP parsing classes.  Removes usage of Pear Net_IPv4 and Net_IPv6.

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/table/address-search.inc.php
+++ b/html/includes/table/address-search.inc.php
@@ -1,5 +1,7 @@
 <?php
 
+use LibreNMS\Util\IP;
+
 $param = array();
 
 if (is_admin() === false && is_read() === false) {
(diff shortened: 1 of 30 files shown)


murrant:
Ready for review, sorry, I wasn't too happy with the initial implementation.  Feeling happy with it now.

laf:
```
Array ( [0] => 1 [1] => Call to undefined function is_ip() [2] => /opt/librenms/html/includes/dev-overview-data.inc.php [3] => 35 )
```

That's device overview page.

html/includes/table/address-search.inc.php
```diff
@@ -81,14 +83,12 @@
     $speed = humanspeed($interface['ifSpeed']);
     $type  = humanmedia($interface['ifType']);
 
-    if ($_POST['search_type'] == 'ipv6') {
-        list($prefix, $length) = explode('/', $interface['ipv6_network']);
-        $address               = Net_IPv6::compress($interface['ipv6_address']).'/'.$length;
+    if ($_POST['search_type'] == 'mac') {
```
laf: Think this is muddled up. Two checks for mac and no ipv4_network now.

laf:
Going through a full discovery now, tested all the webui elements changed + api which are all fine except the one above.

Will see how polling is going after 15 minutes.

laf:
```
Checking BGP peer [IPV6 Address here] PHP Fatal error:  Call to undefined method LibreNMS\Util\IPv6::compress() in /opt/librenms/includes/functions.php on line 787

Fatal error: Call to undefined method LibreNMS\Util\IPv6::compress() in /opt/librenms/includes/functions.php on line 787
```

laf:
Discovery ran through ok. So just that poller error and that's it.

html/includes/table/address-search.inc.php
murrant: shit, yeah, I started to change that, then changed it back (apparently only partially)

murrant:
@laf I only was able to fully test discovery modules ipv4-addresses and ipv6-addresses.

Can you (or anyone) test bgp-peers, cisco-sla, loadbalancers/f5-ltm, ntp/cisco

laf:
I could only test arista bgp. The rest are kit I don't have in my test lab.

murrant:
@laf: I think the parts of bgp I touch are in a common code area.

laf:
Running another full disco to test. After that I can't test anything more.

laf:
Discovery completed.

Issue #7005, opened 2017-07-11
Initial-Detection OS NEC-SV9300PBX


kkrumm1:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the 
==========================================================
Component | Version
--------- | -------
LibreNMS  | 431d15aad9a9115806abba85678efa0b9d46ae66
DB Schema | 199
PHP       | 7.0.19
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
==========================================================
 section of `./validate.php` which you can run from the cli.
- [X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

NEC OID: 1.3.6.1.4.1.119.2
Discovery Output: https://gist.github.com/kkrumm1/5e5c91f58de5394644297216aa75afa7#file-nec-sv9300-pbx-discovery
PollerOutput:https://gist.github.com/kkrumm1/9a03d8b71440f604b0dec2a4818bdce3#file-nec-sv9300-pbx-poller
SNMPWalk: https://gist.github.com/kkrumm1/37490f68483798021d878e60dea5cc3f#file-nec-sv9300-pbx-snmpwalk

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

PR #7108, opened 2017-08-02
newdevice: Added detection for NEC Univerge devices

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7005 

Basic detection as I can't find anything specific to add.


laf:
> dont need the VFE part that was the name of the campus. lol

It's only test data so not to worry.



> doesn't seem to be showing the sysname in the field.

The data you gave me doesn't have sysName so the device isn't returning it to display I'm afraid.

PR #7111, opened 2017-08-03
first and last value where always U since \" wasn't stripped from input

funkestefan:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7111)


laf:
@murrant. I think we should apply the trim() from the previous PR to `snmp_get()`. We end up just adding `trim('"', $value)` all over the show right now. What do you reckon?

murrant:
@laf, probably.  Again, I'm not that confident about the `\`

laf:
@murrant then we just switch to trim('"')

murrant:
@laf merging this for now, we can address the other change you suggested separately if you like.

PR #3177, opened 2016-03-07
Corrected aggregate graph on smokeping page

geordish:

PR #7119, opened 2017-08-04
feature: Add context name to the SNMPv3 options.

KlaasT:
feature: Add the context name option for SNMPv3 to the device settings.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7119)


laf:
Can you give some background for this change please?

KlaasT:
Sure. It is about this discussion https://community.librenms.org/t/add-virtual-device-using-snmp-context/1818/4

If you want to monitor the virtual systems on a Checkpoint VSX gateway you will need to use SNMPv3 and make use of -n to specify a context name. With the context name you will be able to see the SNMP data belonging to that virtual system.

ciscoqid:
On 8/8/2017 12:54 PM, Klaas Tammling wrote:
>
> Sure. It is about this discussion
> https://community.librenms.org/t/add-virtual-device-using-snmp-context/1818/4
>
> If you want to monitor the virtual systems on a Checkpoint VSX gateway you will need to
> use SNMPv3 and make use of -n to specify a context name. With the context name you will
> be able to see the SNMP data belonging to that virtual system.
>

FWIW, with v3 context is also useful for monitoring certain MIBs, like the BRIDGE-MIB, for
different VLANs (BRIDGE-MIB was spec'ed before VLANs existed).  The equivalent in
SNMPv1/2c is to index via a '@<vlan>' community suffix, at least for Cisco.  This allows
you to get useful info about spanning tree, like TCN events, per VLAN.  This is true for
Cisco, presumably others:

    ftp://ftp.cisco.com/pub/mibs/supportlists/wsc5000/wsc5000-communityIndexing.html

But that context would not be programmed into the system like this one :).

Mark

-- 
Mark D. Nagel, CCIE #3177 Emeritus <mnagel@willingminds.com>
Principal Consultant, Willing Minds LLC (http://www.willingminds.com)
cell: 949-279-5817, desk: 714-495-4001, fax: 714-844-4698

** For faster support response time, please
** email support@willingminds.com or call 714-495-4000

KlaasT:
I forgot to add that the code inside LibreNMS already seems to be prepared for SNMPv3 contexts. I just had to add the database field and the option for the SNMP settings and was already able to poll the virtual systems.

network-guy:
I'm worried this change could adversely affect things like FDB polling that rely on iterating through contexts with SNMPv3. I think that should be vetted before this gets merged.

With the VSX, is it possible to get a list of contexts via SNMP that can be iterated for polling?

laf:
> I'm worried this change could adversely affect things like FDB polling that rely on iterating through contexts with SNMPv3. I think that should be vetted before this gets merged.

I agree here. I'm not sure this is a fix for Contexts overall.

KlaasT:
I couldn't find an OID yet which will show all virtual systems. However I am still searching.

On FDB polling I can say that our FDB table is still filled with quite a huge amount of entries. However I am not sure how the code for the FDB polling actually works.

network-guy:
@KlaasT The code for FDB polling, at least on Cisco, polls a list of VLANs, then polls those VLANs as contexts. So VLAN 1 would poll as context vlan-1. If you add a device by context (maybe a VRF attached to a context) that is also capable of doing FDB polling, then it could have unintended consequences. It will probably just fail, but it needs to be vetted. I'm sure there is also other code that relies on context discovery.

I'm not saying this is a bad idea, I'm just saying that you need to make sure this won't affect existing features, even if used in a method that is not intended or expected.

KlaasT:
I've had a look at the complete SNMP output of the VSX gateway however I couldn't find any table which gives me a hint on which virtual systems are being deployed on the server. So I don't know if there is a list where we can iterate through (though I think that would be a better idea). However we still have an open case at Checkpoint regarding this topic maybe we can get some answers there.

laf:
@KlaasT Like @network-guy says, this is better dealt with by iterating over the relevant contexts. I think this will cause more conflicts and isn't the way to approach it anyway.

Proposing that we close this.

KlaasT:
The problem is just that at the moment there doesn't seem to be a way to iterate over the contexts.

I agree that this might not be the best way (if there is a list of all available context names) however the context switch on SNMPv3 is normally a command-line option which can be set. Why shouldn't it be possible to set this option for every device like we can set every other option?

laf:
It's not strictly about just using -n flag for setting the context. We already have existing code that uses `context_name` as a variable name within the `$device` array. I can't say for certain what's going to happen without going through the code and checking which I've just not had time to do so far.

I do know it at least is completely overwrote in some of the discovery files so anyone setting that value in the WebUI will have it altered during discovery.

laf:
I'm proposing to close this. No further discussions have carried on and it still doesn't seem to be the way forward for proper context support.

@librenms/reviewers ?

murrant:
Yes, this will need a more robust solution.  Please do resubmit if you think of something.  Thanks.

murrant:

https://community.librenms.org/t/use-of-snmpv3-contexts-for-device-discovery/9486/2

PR #7121, opened 2017-08-04
feature: ability to edit default alert template

aldemira:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Wow you went for the full thing :)

I probably can't test till Sunday.

aldemira:
Good thing that you didn't test it, I forgot to push the sql file.

includes/alerts.inc.php
```diff
@@ -446,6 +446,9 @@ function DescribeAlert($alert)
     $i           = 0;
     $device      = dbFetchRow('SELECT hostname, sysName, sysDescr, hardware, version, location, purpose, notes, uptime FROM devices WHERE device_id = ?', array($alert['device_id']));
     $tpl         = dbFetchRow('SELECT `template`,`title`,`title_rec` FROM `alert_templates` JOIN `alert_template_map` ON `alert_template_map`.`alert_templates_id`=`alert_templates`.`id` WHERE `alert_template_map`.`alert_rule_id`=?', array($alert['rule_id']));
+    if (!$tpl) {
+        $tpl = dbFetchRow("SELECT `template`,`title`,`title_rec` FROM `alert_templates`  WHERE `name`= 'Default Alert Template'");
+    }
     $default_tpl = "%title\r\nSeverity: %severity\r\n{if %state == 0}Time elapsed: %elapsed\r\n{/if}Timestamp: %timestamp\r\nUnique-ID: %uid\r\nRule: {if %name}%name{else}%rule{/if}\r\n{if %faults}Faults:\r\n{foreach %faults}  #%key: %value.string\r\n{/foreach}{/if}Alert sent to: {foreach %contacts}%value <%key> {/foreach}";
```
laf: You should get rid of these references now that it should never match.

laf:
I also think it's worth adding an FAQ along the lines of, My 'Default Alert Template' no longer exists and tell people to run `mysql -u librenms -p < sql-schema/202.sql` maybe?

aldemira:
Do you think any other documentation is needed?

doc/Support/FAQ.md
```diff
@@ -28,6 +28,7 @@ source: Support/FAQ.md
  - [What is the Difference between Disable Device and Ignore a Device?](#faq28)
  - [Why can't Normal and Global View users see Oxidized?](#faq29)
  - [What is the Demo User for?](#faq30)
+ - [Why does modifying 'Default Alert Template' fail?](#faq31)
```
laf: I'd say it needs to be why does the default alert template not appear in the templates list.

If it's not in the DB you won't see it to even modify it.

laf:
I think the doc needs clarification more.

Also, you seem to have removed the update to alerts.inc.php so the new DB template isn't used if no others exist, it will just fall back to the inbuilt template.

doc/Support/FAQ.md
aldemira: Actually, the listing on the on alert templates page is static. So it'll appear whether the record is in the db or not.

aldemira:
Apparently I didn't quite understand your first change req. Anyway what happens with the latest commit is:
alert.inc.php checks for any associated templates.
If not found retrieve the default one from the db. 
If we do not have that, just the built-in one.

PR #7123, opened 2017-08-04
Attempt to fix repeated sql issue we come across

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This is a repeat offender. We have to attempt to drop the primary key first for those who already have it as you can't attempt to add it back in if it exists, generates an error which means it will be flagged in travis, web ui and console installs as such.


murrant:
Might work...

laf:
It seems to work ok. At the end of the day it's only the queries people are running manually with the addition of dropping the primary key to stop errors being flagged. 

As for the root cause. Difficult to say but we've now made so many changes to older schema files to try and fix the install process that I expect we catch people between two states now, one being that they've already run an older pre-modified schema change and then the second trying to run a later schema change which also fixed what we modified in older schemas but they weren't compatible. 

Repeat occurrences of posts to the schema thread on community should be wrapped up and fixed. It's just noise we don't need.

murrant:
Agreed, just want to make sure we still don't have issues with new installs.  If it is people that have borked schema from schema updates that have since been fixed, great :)

laf:
I don't think anyone has a true idea of why we see the differences however....

1. We merge this and it fixes the issue and reduces the number of reports
2. We merge this and it doesn't fix it and we continue to get reports.
3. We don't merge and guarantee to get reports and have people manually fix them

I prefer #1 and hope we don't see #2

PR #7141, opened 2017-08-11
feature : add link one hour for display  metrics 

yac01:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


yac01:
I added this link because I think users of librenms when they have an incidents or just to see the status of their devices they look more the last moment than the last 2 years that's what I think with this Links users will win more time

laf:
This is my opinion. As is, I don't think we should merge this. We already have 9 graphs at the top adding another one as well just makes it more excessive and will slow page load speeds slightly.

Whilst I'm not a fan of config options for the sake of them, I do think it fits here nicely. Should be a relatively minor change and will give users complete flexibility over the number of mini graphs for time frames.

network-guy:
@yac01 @laf I agree with Laf. It's too cluttered. Personally I would like to remove the 48 hour and 2 week graphs as I rarely use them. If there's a config option to specify graph time defaults, or just enable/disable a list of them, that should be beneficial to everyone.

laf:
@network-guy I've pushed an update to here to allow you to specify the graphs you want + documentation. Fancy giving it a try?

laf:
@yac01 You've added a commit with the name "YACINE BENAMSILI" which isn't in your github account so the CLA isn't showing as signed. Please either remove that commit or squash it out.

network-guy:
@laf Sure. I'll try it out later today.

network-guy:
@laf Seems to work as expected. You may want to add the available options to the documentation though, since you can't use arbitrary time values.

What's the intended difference between ['mini']['widescreen'] and ['row']['normal']?

yac01:
lose

PR #7142, opened 2017-08-11
Added discovery support for Trango Apex Lynx OS

metavrs:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/trango.yaml
```diff
@@ -0,0 +1,9 @@
+os: trango
```
laf: As discussed, looks like this should be apex-lynx or just lynx.

includes/polling/os/trango.inc.php
```diff
@@ -0,0 +1,3 @@
+<?php
+$version = trim(snmp_get($device, ".1.3.6.1.4.1.5454.1.80.1.1.2.0", "-Oqv", "GIGA-PLUS-MIB"), '"');
```
laf: You are using numerical OIDs here but referencing the MIB, you can either change to using the named OID or drop the MIB name as it won't be used.

Same for the below.

includes/polling/os/trango.inc.php
```diff
@@ -0,0 +1,3 @@
...
+$hardware = "Trango " . trim(snmp_get($device, ".1.3.6.1.2.1.1.1.0", "-Oqv", "GIGA-PLUS-MIB"), '"');
```
laf: This variable is used to populate some search fields so I think dropping 'Trango ' should be done.

mibs/trango/TRANGO-OrionLynx-MIB-COMPACT.mib
```diff
@@ -0,0 +1,618 @@
+GIGA-PLUS-MIB DEFINITIONS ::= BEGIN
```
laf: You can rename this mib file to be `GIGA-PLUS-MIB DEFINITIONS` so it's obvious what it is.

mibs/trango/GIGA-PLUS-MIB-DEFINITIONS.mib
```diff
@@ -0,0 +1,618 @@
+GIGA-PLUS-MIB DEFINITIONS ::= BEGIN
```
laf: Just need to drop the .mib from the end of the filename and that's it :)

metavrs: done!

metavrs:
Thank you, sir!  My first OS merge :)

Issue #7116, opened 2017-08-04
NEW device ets8801  Expert Transfer Switch 8801


kristoferus75:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
Component | Version
--------- | -------
LibreNMS  | 1.30-31-g0ff847a
DB Schema | 200
PHP       | 5.4.16
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2

- [ x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
https://p.libren.ms/view/e0fdc96c
https://p.libren.ms/view/97ca27a8
https://p.libren.ms/view/ca8f6a25
- [x ] Please provide as much detail as possible.

New device Expert Transfer Switch 8801-1   ets8801
https://www.gude.info/power-distribution/transfer-switch/expert-transfer-switch-8801-1.html

The device have 2 Power connections which should be not zero:

Only this two values are nesssacery to know -> available ore not available:

"not zero if primary Power available"
"not zero if secondary Power available"

```
snmpget -v2c -mALL -c public 10.254.104.2 etsPrimPowAvail.0
GUDEADS-ETS-MIB::etsPrimPowAvail.0 = INTEGER: 1
snmpget -v2c -mALL -c public 10.254.104.2 etsSecPowAvail.0
GUDEADS-ETS-MIB::etsSecPowAvail.0 = INTEGER: 1

```
- [x ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

https://p.libren.ms/view/27b7a2df

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

PR #7145, opened 2017-08-12
newdevice: Added support for Gude ETS devices

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7116


laf:
This has vendor junk because I switched from murrant/composer-install and had to run composer install.

All the updates seem safe so I'm leaving them here as I have no idea how to revert this as those files don't show changed in my local install.

kristoferus75:
Hi !

I have added one ats device but i see no states -> which modules i need to activate ? -> sensors activated !

also the sysname seems wrong value !

kind regards

laf:
Works for me

image


sysName is sent from the device so if it's wrong in LibreNMS then it's wrong on the device.

Can you please stop replying to closed pull requests, if you need help please use irc or community forum.

kristoferus75:
the hostname is set on the device  ok i will use the irc or community forum

kristoferus75:
Hi !

The primary power shows not the icon available it shows in red 1100# and the secondary power icon is ok !

Recent Events:
State sensor Primary power has changed from (1200) to (1100)

thanks
kristoferus75

PR #7160, opened 2017-08-15
NewDevice - Hitachi Data Systems SAN

EnzoZafra:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Looks good :)

Can you move the mib into a folder called hitachi, they are per vendor rather than per os.

includes/definitions/svos.yaml
```diff
@@ -0,0 +1,9 @@
+os: svos
+text: 'Hitachi Storage Virtualization Operating System (SVOS)'
+type: storage
+icon: hds
+mib_dir:
+    - hitachi
+discovery:
+    - sysObjectId:
+        - .1.3.6.1.4.1.116.3.11.4.1.1
```
laf: Sorry one last thing, this is quite specific sysObjectId. We don't have anything else registered in `1.3.6.1.4.1.116.` so you can probably use that, or be safe and do `1.3.6.1.4.1.116.3.` for now.

EnzoZafra: I set it to `.1.3.6.1.4.1.116.3.` just to be safe. Hitachi has quite a bit of other product that probably does not run svos

PR #7162, opened 2017-08-15
fix: Fixed dump_db_schema / validate to work with newer versions of MariaDB

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes validate.php issues when running MariaDB 10.2.7 or later.

Removed two unused functions.

Tested on both MariaDB 10.2.7 and 5.5.50


PdeNiet:
I tested it on MariaDB 10.2.7 as promised, works fine with this patch.
Thanks for looking into it Laf.

PR #7175, opened 2017-08-17
feature: Active Directory user in nested groups

MrDragon:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


html/includes/authentication/active_directory.inc.php
```diff
@@ -37,33 +37,43 @@ function authenticate($username, $password)
...
                 if (isset($config['auth_ad_debug']) && $config['auth_ad_debug']) {
                     if ($entries['count'] == 0) {
-                        throw new AuthenticationException('No groups found for user, check base dn');
-                    } else {
                         throw new AuthenticationException('User is not in one of the required groups');
                     }
                 }
 
-                throw new AuthenticationException();
+                return false;
```
MrDragon: Sorry, was my lack of understanding. Reverting.

html/includes/authentication/active_directory.inc.php
```diff
@@ -37,33 +37,43 @@ function authenticate($username, $password)
...
+                        // user is in the current group
                         return true;
                     }
                 }
 
                 // failed to find user
                 if (isset($config['auth_ad_debug']) && $config['auth_ad_debug']) {
                     if ($entries['count'] == 0) {
-                        throw new AuthenticationException('No groups found for user, check base dn');
-                    } else {
```
MrDragon: I'm not searching what group(s) a user directly belongs to, but look at the defined groups and check if the user is a member of the listed groups or subgroups of that group. As far as I know "count" is always 0 or positive. On that search, if it's positive, the user is a member of that group and we've already returned "true", which only leaves "0" as a possibility.
Maybe the message should be more precise, like "User is not in one of the required groups, or the defined groups are outside the base dn" ?

html/includes/authentication/active_directory.inc.php
```diff
@@ -104,6 +90,38 @@ function reauthenticate($sess_id, $token)
...
+    $search = ldap_search(
+        $ldap_connection,
+        $config['auth_ad_base_dn'],
+        // add 'LDAP_MATCHING_RULE_IN_CHAIN to the user filter to search for $username in nested $group_dn
+        // limiting to "DN" for shorter array
+        "(&" . get_auth_ad_user_filter($username) . "(memberOf:1.2.840.113556.1.4.1941:=$group_dn))",
+        array("DN")
+    );
+    $entries = ldap_get_entries($ldap_connection, $search);
+    unset($entries[0]['memberof']['count']); //remove the annoying count
```
murrant: The unset doesn't seem to be needed.

html/includes/authentication/active_directory.inc.php
```diff
@@ -104,6 +90,38 @@ function reauthenticate($sess_id, $token)
...
 
 
+function user_in_group($username, $groupname)
+{
+    // check if user is member of the given group or nested groups
+
+    global $config, $ldap_connection;
+
+    // get DN for auth_ad_group
+    $filter = "(&(objectClass=group)(cn=$groupname))";
```
murrant: $filter is unused

murrant:
I know the current code does not use it, but if you could switch to the Config class that would be great (not doing so won't block the PR).

Soo, `$config['auth_ad_groups'][$group]['level']` -> `Config::get("auth_ad_groups.$group.level")`

You can also specify a default value in the second argument to return when the variable is unset.

MrDragon:
Ok, fixed issues (I hope), and changed that line to Config::get.
I didn't change other "$config" references, because, well, tbh, I don't know what I'm doing. Hadn't used classes before, haven't used 'git' before. I'm just an old-school procedural language, "patch -p0" kinda guy, good enough to mess with code until it does what I need it to do ;)

I appreciate your patience and input!

MrDragon:
Sorry if I put in too many pull requests - I think I may have finally figured out where to see if I already requested that or not. Not quite sure though...

murrant:
Nice work @MrDragon :)

laf:
@MrDragon We had to revert this PR due to reported issues of AD auth problems: #7254

murrant:
Part of the issue is Config::get(), because some people have periods in groups and we use the group name as the key in the config.

Sorry about that.  Just using $config[] will fix that.  I'm not sure if we have other issues though.

MrDragon:
Thank you for re-fixing that.

PR #7179, opened 2017-08-17
Updated pi-hole graphs for better grouping

mbwall:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


mbwall:
The intention here is to separate out graphs with different scales.  

Unfortunately its all fairly useless with the current values coming out of pi-hole (24hr rolling sum).  Looking into a fix for that, and open to ideas.

laf:
From the pi-hole script it looks like some of the data is per day only so not sure you can fix that. Others look like they are rolling values so should produce ok results now (assuming the data changes enough).

PR #7180, opened 2017-08-18
multiple fixes under agentStpSwitchConfigGroup in EdgeSwitch-SWITCHING-MIB

elfchief:
I was unable to find anywhere to submit these changes upstream, so I'm just
submitting them to you so that you can have them (since you were nice enough
to have the MIB for me to start with).

Corrections here have been verified both in the firmware (which, if you peel
it apart, contains both the OIDs and associated names), and by pulling the
associated data from switches and comparing the results to the expected
results gleaned from the switch's human-targeted interfaces.

Specific corrections:
  agentStpPortTable should be at agentStpSwitchConfigGroup.9, not .7
  agentStpCstConfigGroup should be at .10, not .8
  agentStpCstPortTable should be at .11, not .9
  agentStpMstTable should be at .12, not .10
  agentStpMstPortTable should be at .13, not .11
  agentStpMstVlanTable should be at .14, not .12
  agentStpBpduGuardMode should be at .7, not .13
  agentStpBpduFilterDefault should be at .8, not .14

  ...and agentStpPortHelloTime should exist at agentStpPortEntry.10

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/mibs/edgeswitch/EdgeSwitch-SWITCHING-MIB
+++ b/mibs/edgeswitch/EdgeSwitch-SWITCHING-MIB
@@ -6798,7 +6798,7 @@ agentPortMirrorTypeTable OBJECT-TYPE
          STATUS      current
          DESCRIPTION
                      "STP port table."
-         ::= { agentStpSwitchConfigGroup 7 }
+         ::= { agentStpSwitchConfigGroup 9 }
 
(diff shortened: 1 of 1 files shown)


laf:
So you've stopped by to submit a fix to the mib rather than being a LibreNMS user? If so, wow, thanks :)

PR #7182, opened 2017-08-18
Cambium ePMP CPU reporting fix

neg2led:
See https://community.librenms.org/t/cambium-epmp-cpu-usage-not-reporting-correctly/1877 for more details, and the comment below.

Pre-commit output;

> librenms@orwell:~$ ./scripts/pre-commit.php
> Running unit tests... success
> Running style check... success
> Running lint check... success
> Tests ok, submit away :)

This is a very small change; one number changed from 100 to 10, and a division removed.

* Remove division by 10

Removes unnecessary division by 10 from poller file; this is handled by processor_precision in the main include.

* Correct processor_precision

Adjusted processor precision divider from 100 to 10 (as sysCPUUsage.0 reports from 0-1000 as 0-100.0%)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/discovery/processors/cambium.inc.php
```diff
@@ -21,7 +21,7 @@
             '0',
             'cambium-cpu',
             $descr,
-            '100',
+            '10',
```
murrant: This should be 1 shouldn't it?

neg2led: This value is the "processor_precision" value in the database. includes/polling/processors.inc.php divides whatever the CPU-specific polling module returns by this number.

The raw value presented by the OID ranges from 0-1000 - a value of 50 is 5% - so the raw value needs to be divided by 10.

Currently, includes/polling/processors/cambium-cpu.inc.php divides the raw value by 10, and passes that value back to includes/polling/processors.inc.php;

https://github.com/librenms/librenms/blob/149857f246fef4d83f5f37a7e5a87fdedce113f2/includes/polling/processors/cambium-cpu.inc.php#L14-L16

processors.inc.php then divides the return value by processor_precision;

https://github.com/librenms/librenms/blob/149857f246fef4d83f5f37a7e5a87fdedce113f2/includes/polling/processors.inc.php#L27

Giving us a total divisor of 1000, which we see in a poller run;

```
Processor Processor... Cambium CPU UsageSNMP[/usr/bin/snmpget -v2c -c COMMUNITY -OvQ -M /opt/librenms/mibs:/opt/librenms/mibs/cambium:/opt/librenms/mibs/cambium udp:HOSTNAME:161 CAMBIUM-PMP80211-MIB::sysCPUUsage.0]
50
 
0.05%
```

(0.05 being 1/1000th of 50)

This gives us two options;

1. Remove the division from cambium-cpu.inc.php, and set processor_precision to 10
2. Leave the division in cambium-cpu.inc.php, and set processor_precision to 1

I've opted for option 1 - a search indicates processor_precision is only used by processors.inc.php, and only for dividing the raw value returned from a sub-module, so it seems a bit redundant to move it out to the device-specific sub-module when functionality is already in place.

If there's a compelling reason to choose option 2, please let me know - the only reason I can think of is that it's modifying two files instead of one, which doesn't seem like a great reason.

laf: I do sometimes worry about changes like this as it could have been ok for earlier firmware for the device and now Cambium have changed it so this would break things.

@pheinrichs are you able to comment on this to check as you added the original support?

@neg2led if this is approved by pheinrichs then I think you can actually now get rid of the polling file as it shouldn't be needed.

pheinrichs: This looks good to me, option 1 seems like the best way to go here. I've noticed ePMP haven't been reporting right but was to busy to look into. Happy to see a fix.

laf: Thanks for confirming. @neg2led can you remove the polling file as I think it's not needed.

neg2led: Just confirmed this on my own install, removing the file entirely works great. Will update this :)

neg2led:
Removed the cambium-cpu.inc.php file after confirming it's no longer required at all. And hey, the automated checks didn't fail with an unrelated error this time! Always a plus.

murrant:
Thanks for the clarification, I thought the oid returned 0-100.

neg2led:
Originally it was specified as 0-100 - but an implementation bug made it report 0-1000. When the bug was found, it was decided to change the spec rather than fix the implementation, as the additional precision was considered a plus (though not really necessary)

More info here; http://community.cambiumnetworks.com/t5/ePMP-2000-and-1000/High-CPU-Usage/td-p/42240

laf:
Thanks for the fix @neg2led :D

PR #7183, opened 2017-08-18
This change allows discovery/polling of all sensors (fans and tempera…

netmax79:
…tures) from Brocade VDX devices

instead of only the first few. This is tested with VDX 6740, 6740T, 8770-4, 8770-8.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
@netmax79 Does this leave the `index` value the same for existing sensors or is that changed?

netmax79:
Values of existing sensors are keept, new ones are added.

PR #7185, opened 2017-08-19
docs: Docs alerting

benwa:
Mostly tabelize the entities document for readability. Also cleanup some markdown


murrant:
Super job @benwa :)

PR #7188, opened 2017-08-20
Feature: Notify about failed updates, block detectable bad updates

murrant:
Ability to post notifications when the update fails.
Detect and roll back updates that will cause broken installs. (Needs testing)
Add severity to notifications, critical (2) notifications will display a toast.

This will be used for removing in-tree dependencies and raising the minimum php version.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Does this just post one failed update notice or would it keep posting them each day?

f0o:
It looks like it would create a new notification for each failure

murrant:
@f0o The new_notification function does not allow new notifications with the same checksum.  So if the title and message are the same, it will not be created.

murrant:
FYI, I would like to get this merged in 1.31 because it is a pre-requisite for other work.

laf:
I'm unsure why travis hasn't picked this up:

```
MariaDB [librenms]> ALTER TABLE `notifications` ADD `severity` INT DEFAULT 0 NULL COMMENT '0=ok,1=warning,2=critical' AFTER `body`;
ERROR 1292 (22007): Incorrect datetime value: '0000-00-00' for column 'datetime' at row 13
```

That's after setting the sql mode to what we run in tests (I run that in my dev anyway)

laf:
Ok updated, pretty sure as well the mysql checks were never running before so that's now fixed.

PR #7193, opened 2017-08-21
feat:add links for librenms graph

yac01:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Support/Configuration.md
+++ b/doc/Support/Configuration.md
@@ -210,6 +210,33 @@ $config['int_l2tp']                = 0;  # Enable L2TP Port Types
 ```
 Enable / disable certain menus from being shown in the WebUI.
 
+You are able to adjust the number and time frames of the quick select time options for graphs and the mini graphs shown per row.
+
+Quick select:
+```php
+$config['graphs']['mini']['normal'] = array(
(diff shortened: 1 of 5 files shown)


yac01:
@laf excuse me because This is the second time I have committed the same thing 
i think i have problems in my count  git ub because in each time when i make some modification in github i have the probleme that don't stop "ontinuous-integration/travis-ci/pr" but now I would not make any other changes on github i would do it only in the terminal

murrant:
The docs doesn't seem to mention `$config['graphs']['mini']['widescreen']` and I'm not sure how the three fit together.

laf:
`$config['graphs']['mini']['widescreen']` = Widescreen setup for the pages like a direct port page.

`$config['graphs']['mini']['normal']` = Normal screen size of the above.

`$config['graphs']['row']['normal']` = The row of graphs you use to select the time period for a graph you want to see.

We can't use mini normal and row normal as this would change the number of graphs people see by default so this setup replicates the existing setup with the ability to custom either of those above scenarios.

murrant:
Thanks, @laf we just need that in the docs and I think this is ready.

laf:
@murrant doc update pushed.

PR #7206, opened 2017-08-22
fix: Fix Comware Processor Discovery && Hardware Info

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Tested on MSR, VSR, 5800, 5900, 5940, 10500.

Processor:
Discovery broke if processor usage at discovery point was 0, hence we could risk having the processor added/removed during each run on devices with no load. Secondly, we can't rely on the class "module" as other entities are also branded as this hence adding bogus processors. This will not break graphs as we continue to use the same indexes. I have changed the descr as we should just present what HPE publish, even if the devices is IRF'ed(this now matches the descr in mempools).

Hardware:
The current MIB available doesn't support the newer models, also the model number listed in the sysDescr is a much better representation.

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7206)

### DIFF
--- a/includes/discovery/processors/comware.inc.php
+++ b/includes/discovery/processors/comware.inc.php
@@ -13,23 +13,16 @@
 
 if ($device['os'] == 'comware') {
     echo 'HPE Comware ';
+    $entity_data = snmpwalk_cache_oid($device, 'entPhysicalName', array(), 'ENTITY-MIB');
+    $procdata = snmpwalk_cache_oid($device, 'hh3cEntityExtCpuUsage', array(), 'HH3C-ENTITY-EXT-MIB');
 
-    $entity_data = snmpwalk_cache_oid($device, 'entPhysicalClass', array(), 'ENTITY-MIB');
-    $indexes = array_keys(array_filter($entity_data, function ($entry) {
-        return $entry['entPhysicalClass'] == 'module';
-    }));
-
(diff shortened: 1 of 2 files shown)


kkrumm1:
need any testers?

laf:
@kkrumm1 let me know when you've tested so we can merge.

kkrumm1:
I tested it and on HPE Comware Switch model 1910 and the CPU used to show but now its not discovering it. 
On the HPE Comware 10508 model, the CPU is showing but the name is "virtue Board"

discovery output 1910: https://gist.github.com/kkrumm1/6714b821f24f796f721771a70072fd6b
discovery output 10508: https://gist.github.com/kkrumm1/72cdd77978e4d7d0274bdffad56ea481

Rosiak:
Just a few quick thoughts:

As for the 1910 I'm not quite sure how to fix that since it seems that I cannot match on the entPhysicalName. I would have to match ```Level 1 Virtual Module #1```, but what makes ```Level 2 Virtual Module #0``` invalid in this case?

For the 10508 we are now just using the what's provided by HPE, even though this might be a bit crappy in some cases :(

kkrumm1:
Okay, no worries on the 1910's they are junk switches. :)

PR #7208, opened 2017-08-22
refactor: Updated validate.php to only warn users the install is out of date if > 24 hours

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

General update. We shouldn't warn users that the install is only just out of date.


murrant:
This will only trigger if it is two days out of date.  I'll fix that up.

PR #7212, opened 2017-08-23
Copy redhat.svg to redhatenterpriseserver.svg

rockyluke:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
@rockyluke What's the reason for this as I don't see anywhere we would make use of it?

rockyluke:
I see for my RedHat Enterprise : 

> Operating System | Linux 2.6.32-573.18.1.el6.x86_64 (**RedHatEnterpriseServer** 6.7)

laf:
Does it now work as I thought you needed to use the distro script to get specific os logos past the standard Tux one.

rockyluke:
With the distro script I get two more RedHat identifiers that's why I need new svg : 
- RedHatEnterpriseServer
- RedHatEnterpriseES

(sorry it's only a cosmetic issue...)

murrant:
Perhaps change the distro script to return something more consistent?

rockyluke:
I can do that if you prefer :)

laf:
I think it's definitely better to update distro than keep adding logos.

laf:
Closing, if you can update the distro script that would be better.

rockyluke:
It's done : https://github.com/librenms/librenms-agent/pull/145

PR #7215, opened 2017-08-23
Greatly reduces application JavaScript memory leak

rmedlyn:
Removed clone statement from jquery-bootgrid destroy function that was causing anything in a bootgrid table to persist and not get garbage collected. Also added some code to make sure a table is emptied of elements and event listeners on widget refresh. If those are too much, calling destroy on its own after the fix still significantly reduced the leak. I also realize that my minification might be a bit different than the original. I can take a look at that if it's a concern.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7215)


murrant:
I see there is a PR for this upstream.  Perhaps it is best to get it fixed there as someone may just update the bootgrid files later, removing your fix.

At least you should comment on that PR to help it get merged.

https://github.com/rstaib/jquery-bootgrid/issues/216

laf:
This library hasn't had an update in 2 years. We could just fork it and maintain this patch locally or merge in expect that no further updates are to be provided upstream.

rmedlyn:
At this point, I think that maintaining a fork seems to be the best long-term option short of porting to a more up-to-date library like DataTables.

murrant:
Yeah, unfortunate, bootgrid does look unmaintained.

Looks good after you fix the conflict.  It looks like your branch might be out of date with master or something.

laf:
@rmedlyn Are you able to fix the conflict?

rmedlyn:
Should be fixed now. Sorry for the delay.

rmedlyn:
@laf Does everything look okay now?

laf:
Testing

PR #7216, opened 2017-08-23
fix: Comware dBm Limits && Comware Sensor Descr

Rosiak:
1. Make the interface sensors descriptions match the same "style" as Cisco.
2. Do not detect interfaces which have not set hh3cTransceiverDiagnostic, this solves an issue where the MGMT interface on a device is wrongfully monitored.
3. After yesterdays dBm thresholds update I discovered that we need to round off the values we insert into the DB, otherwise we end in a discovery loop where we will try to update the thresholds due to the discovered value is larger than what we can store in the DB, hence this will differ on each discovery.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


kkrumm1:
This fixed my issues. 👍

PR #7223, opened 2017-08-24
feature: Improve install process

murrant:
Fix incorrectly updating session with build-ok before start of schema update
Set a timeout for progress on the schema build 40s (lock wait time is 30s, so must be more than that).  Allow the user to restart the process if this timeout is reached.
Animate the progress bar while waiting for the schema update. Stop animation on failure or success.
Properly destroy the session after install.  This allows the user to restart if they need to without any tricks. Only need to make sure config.php does not exist.
Move next step buttons to the right.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/ajax_output.php
+++ b/html/ajax_output.php
@@ -14,7 +14,6 @@
 
 session_start();
 if (isset($_SESSION['stage']) && $_SESSION['stage'] == 2) {
-    $_SESSION['build-ok'] = true;
     $init_modules = array('web', 'nodb');
     require realpath(__DIR__ . '/..') . '/includes/init.php';
 } else {
--- a/html/includes/output/db-update.inc.php
+++ b/html/includes/output/db-update.inc.php
@@ -61,7 +61,11 @@ if (($fp = popen($cmd . ' 2>&1', "r"))) {
     if (pclose($fp) === 0) {
         echo "Database is up to date!";
         $_SESSION['build-ok'] = true;
+    } else {
+        echo "Database schema update failed!";
     }
 }
 
+ob_end_flush();
+flush();
 session_write_close();
--- a/html/install.php
+++ b/html/install.php
@@ -75,12 +75,17 @@ if ($stage == 4) {
         $msg = "You haven't entered enough information to add the user account, please check below and re-try";
     }
 } elseif ($stage == 6) {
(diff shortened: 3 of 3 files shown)


html/ajax_output.php
```diff
@@ -14,7 +14,6 @@
 
 session_start();
 if (isset($_SESSION['stage']) && $_SESSION['stage'] == 2) {
-    $_SESSION['build-ok'] = true;
```
murrant: This was a bad error here :(  Did not allow users to restart schema build if it failed.

laf:
Nice work @murrant :)

Install works as expected, trying to simulate a failure I restarted mysql during build-base, any schema updates attempted during the restart show an error and the whole process completes and thinks it's done still. Not sure if you'd expect that to happen or not?

murrant:
Thanks, I'll try making that failure happen.

murrant:
@laf Try with `./build-base.php; echo $?` that should always return non-zero during a fatal error.

But If the database build gets interrupted, the results are unpredictable.  I'm not sure how to handle that.

murrant:
I think that should be fixed in another PR.  If build-base.php returns the correct return code, then the install.php code does the right thing.

laf:
```bash
-- Done
1
-bash-4.2$
```

That's repeatable - If the restart happens during:

```bash
-- Creating base database structure
Step #0 ...
Step #1 ...
Step #2 ...
Step #3 ...
Step #4 ...
Step #5 ...
Step #6 ...
```

Then we get `1`

If it happens after in:

```bash
-- Updating database schema
000 -> 001 ... done (0 errors).
001 -> 002 ... done (0 errors).
002 -> 003 ... done (2 errors).
003 -> 004 ... done (24 errors).
004 -> 005 ... done (11 errors).
005 -> 006 ... done (5 errors).
```

We get `0`.

laf:
Does it not just need a `$return = 1;` here: https://github.com/librenms/librenms/blob/master/includes/sql-schema/update.php#L89

murrant:
More complex than that, if things happen in the middle of stuff it will still end up broken.

I added breaks.  Looking at things I discovered create table does an implicit commit.

PR #7230, opened 2017-08-27
docs: Updated Services.md include chmod +x

kkrumm1:
DO NOT DELETE THIS TEXT

Added the command to allow execution to the directory for Nagios plugins 

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Services.md
+++ b/doc/Extensions/Services.md
@@ -25,7 +25,8 @@ $config['nagios_plugins']   = "/usr/lib/nagios/plugins";
 ```
 
 This will point LibreNMS at the location of the nagios plugins - please ensure that any plugins you use are set to executable.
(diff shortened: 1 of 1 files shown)


doc/Extensions/Services.md
```diff
@@ -25,6 +25,7 @@ $config['nagios_plugins']   = "/usr/lib/nagios/plugins";
 ```
 
 This will point LibreNMS at the location of the nagios plugins - please ensure that any plugins you use are set to executable.
+Usaully: chmod +x /usr/lib/nagios/plugins
```
murrant: Change usually to "For example" an put this in a code block.

Also, wouldn't you want `chmod +x /usr/lib/nagios/plugins/*`?

kkrumm1: oops yes, I will change that.

PR #7237, opened 2017-08-28
Do not Initialize auth for cli commands, as they do not use auth anyway

jthurman42:
Currently, every new poller will initiate authentication, but does not use authentication in any way.  This adds unnecessary time to the poller process, and creates a ton of unused connections from polling workers to whatever auth source is being used.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7237)

### DIFF
--- a/includes/init.php
+++ b/includes/init.php
@@ -149,7 +149,9 @@ if (!module_selected('nodb', $init_modules)) {
 if (file_exists($config['install_dir'] . '/html/includes/authentication/'.$config['auth_mechanism'].'.inc.php')) {
     require_once $install_dir . '/html/includes/authentication/functions.php';
     require_once $config['install_dir'] . '/html/includes/authentication/'.$config['auth_mechanism'].'.inc.php';
-    init_auth();
+    if (! isCli()) {
+        init_auth();
+    }
 } else {
     print_error('ERROR: no valid auth_mechanism defined!');
     exit();


laf:
We'll need to check this in more detail, I have a feeling auth is needed for some alerting. I'll have to refresh my memory.

murrant:
Yes, I'm pretty sure auth is used by alerting.

@jthurman42 can you give more details about the issue you are trying to fix? What auth type you are using for example.

jthurman42:
The specific issue ran into was with `poller.php` requiring external authentication (in this case LDAP but any non-local auth) to be present, which has proven to be sub-optimal with a large number of pollers / devices.  The attempt is to remove this requirement to increase polling efficiency and remove unnecessary dependancies.

I'm open to suggestions on how we could either exclude the auth initiation from just polling, or add it back into those items that need it on the CLI.  Having the auth loaded inside of `init.php` makes it a bit tricky to disable from my quick reading through the code.

laf:
So we run `get_userlist()` for alerts.php which doesn't have any dependancy on poller. I actually can't see where we might call something in the auth files for polling right now.

@jthurman42 - Are you using alerting and do you see any issues?

murrant:
I think you are correct that the poller does not need authentication.  But I think the fix isn't correct.  Perhaps, reorganizing the init a bit.  We could even save a bit of ram by not including the authentication function files.

jthurman42:
@laf I am evaluating LibreNMS, and this was a performance hit that was found for our environment.  I have not evaluated the alerting functions yet, so I didn't run into any issues.

@murrant Agreed that this isn't the right solution since there are downstream impacts.  I'll see if I can find a more elegant way to meet the requirement.

laf:
Is checking `if (module_selected('auth', $init_modules) || module_selected('alerts', $init_modules)) {` not all we need before calling `init_auth()`

murrant:
@laf I believe so, but the application of the auth module may not be consistent since this is a change in behavior.  We will need to audit all init calls.

laf:
> @laf I believe so, but the application of the auth module may not be consistent since this is a change in behavior. We will need to audit all init calls.

Confused. Aren't we only now interested in running auth init for alerts and http and if so, the module_selected checks will do that and run the function like it does now?

murrant:
Look at the files that call init.php.   Some include 'auth' and 'http', others only include 'http'.

Also, if we aren't going to call ini_auth(), we probably shouldn't include the auth functions, just a waste of memory and the many of the auth functions rely on init_auth() having been called.  The other option is a late init, but that requires a check each function call.

laf:
> Look at the files that call init.php. Some include 'auth' and 'http', others only include 'http'.

Then maybe just call it when we want auth then. It's probably not needed for web based init.

laf:
My suggested patch: https://p.libren.ms/view/raw/9a5e3adf

```bash
diff --git a/includes/init.php b/includes/init.php
index 38e9c9e53..9bc552830 100644
--- a/includes/init.php
+++ b/includes/init.php
@@ -146,13 +146,15 @@ if (!module_selected('nodb', $init_modules)) {
     require $install_dir . '/includes/process_config.inc.php';
 }

-if (file_exists($install_dir . '/html/includes/authentication/'.$config['auth_mechanism'].'.inc.php')) {
-    require_once $install_dir . '/html/includes/authentication/functions.php';
-    require_once $install_dir . '/html/includes/authentication/'.$config['auth_mechanism'].'.inc.php';
-    init_auth();
-} else {
-    print_error('ERROR: no valid auth_mechanism defined!');
-    exit();
+if (module_selected('auth', $init_modules)) {
+    if (file_exists($install_dir . '/html/includes/authentication/' . $config['auth_mechanism'] . '.inc.php')) {
+        require_once $install_dir . '/html/includes/authentication/functions.php';
+        require_once $install_dir . '/html/includes/authentication/' . $config['auth_mechanism'] . '.inc.php';
+        init_auth();
+    } else {
+        print_error('ERROR: no valid auth_mechanism defined!');
+        exit();
+    }
 }

 if (module_selected('web', $init_modules)) {
```

murrant:
That seems fairly reasonable to me.

murrant:
I can think of a few better solutions that would take time and testing, but this is a good quick fix.

laf:
@jthurman42 Would you like to update this PR using the patch above for now?

laf:
@jthurman42 if you get back to this please reply and we will re-open

PR #7245, opened 2017-08-28
feature: Add error checking/exceptions for AD auth group lookups

ekoyle:
NB: this could break existing installations if they are using bad group names (either non-existent or non-unique)

Use `ldap_get_entries` instead of `ldap_get_first_entry` to be able to tell the difference between no matches and an LDAP query failure.  Also check for a non-unique match, as this could be extremely hard to debug otherwise and may lead to unexpected results.

`ldap_get_first_entry` has the same return value (`false`) when there is an error or when there are no results

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/authentication/active_directory.inc.php
+++ b/html/includes/authentication/active_directory.inc.php
@@ -97,15 +97,27 @@ function user_in_group($username, $groupname)
 
     global $ldap_connection;
 
+    $search_filter = "(&(objectClass=group)(cn=$groupname))";
+
     // get DN for auth_ad_group
     $search = ldap_search(
         $ldap_connection,
         Config::get('auth_ad_base_dn'),
-        "(&(objectClass=group)(cn=$groupname))",
+        $search_filter,
         array("cn")
     );
-    $result = ldap_first_entry($ldap_connection, $search);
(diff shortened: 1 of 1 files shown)


murrant:
@ekoyle Do you mind if I merge this into PR #7254?

ekoyle:
@murrant: Feel free.

murrant:
Merged.  Thanks.

laf:
I don't use AD but as we've just had issues with changes here in the past, are we sure none of this is going to break existing users installs?

murrant:
@laf this is just more specific error messages.  Should have been in the original nested groups PR.

I did make a comment of the removal of some of the message detail, but didn't think about emitting those in  user_in_group() function.

We might need to check where get_userlevel is called so we don't have uncaught exceptions.

ekoyle:
@murrant: it may change behavior in the case of a non-unique group name or an invalid group name among other valid group names (ie. both bad, but they may currently work anyway)

laf:
I think we should have asked for testers on this and other auth changes, it seems no setups are quite the same.

PR #7252, opened 2017-08-29
doc: Add SELinux context using cron and services

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

----------------------

check-services.php can't write the RRD file through cron when SELinux is enabled, adding the cron context solve the issue.


FTBZ:
⚠️ Wait before adding this to the documentation, I need to check more.

FTBZ:
I don't know why, but some of my installs have the correct SELinux context and other none. Searching on the Internet, it can be linked to a bug that was corrected on CentOS and probably not affecting what that was created during the bug.

Removing the cron file and adding it again has solved my issues with SELinux without changing the context.

PR #7255, opened 2017-08-29
Add a warning about the cookie configuration

mikeSimonson:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7255)

### DIFF
--- a/validate.php
+++ b/validate.php
@@ -334,6 +334,12 @@ if (empty($ini_tz)) {
     print_fail("You have a different system timezone ($sh_tz) specified to the php configured timezone ($php_tz), please correct this.");
 }
 
+// check php cookie config
+$cookieParams = session_get_cookie_params();
+if ($cookieParams['httponly'] !== false) {
+    print_fail("Disable httponly as it will at least break the install");
+}
+
 // Test transports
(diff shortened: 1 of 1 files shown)


laf:
What's the reason for this? We actually setting as part of the authentication cookies. It's not used in install.php though.

mikeSimonson:
If the httponly setting is set to on in the php configuration the ajax request made in the step of the install where the database is updated brakes.

murrant:
@mikeSimonson Interesting, could setting that explicitly in install.php fix the issue too?

laf:
@mikeSimonson Any comment?

mikeSimonson:
@murrant Maybe it could, I am not really sure that you can taught.

laf:
@mikeSimonson We set it in https://github.com/librenms/librenms/blob/master/html/includes/authenticate.inc.php#L7 already so seems possible to toggle it on / off in php. Not sure if that overwrites what the user sets in php.ini but it's worth a test.

laf:
@mikeSimonson bump

laf:
@mikeSimonson Feel free to comment on this still and we can re-open.

PR #7259, opened 2017-08-30
feature: AD support nested groups (resubmit #7175)

murrant:
Don't let groups with periods in the name cause issues.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7259)

### DIFF
--- a/LibreNMS/Exceptions/AuthenticationException.php
+++ b/LibreNMS/Exceptions/AuthenticationException.php
@@ -28,8 +28,12 @@ namespace LibreNMS\Exceptions;
 class AuthenticationException extends \Exception
 {
     // Redefine the exception so message defaults to a generic 'Invalid credentials'
-    public function __construct($message = 'Invalid credentials', $code = 0, \Exception $previous = null)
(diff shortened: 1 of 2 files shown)


murrant:
As far as I can tell, this is ready.

1. One user had periods in his group names, which caused them to be parsed incorrectly.  This is fixed in this version.
2. Another user had an issue where his groups were not in his base_dn, which should be the case but wasn't required by the previous code.  I don't think this needs to be fixed in the code.  We could create a notification to make sure users have a correct base_dn setting.

laf:
> Another user had an issue where his groups were not in his base_dn, which should be the case but wasn't required by the previous code. I don't think this needs to be fixed in the code. We could create a notification to make sure users have a correct base_dn setting.

We should definitely at least document it somewhere, even if it's just a community post.

murrant:
from the docs:
```
$config['auth_ad_base_dn']                 = "<dc=your-domain,dc=com>";  // groups and users must be under this dn
```

PR #7264, opened 2017-08-30
Add Support for DragonWave Horizon

pheinrichs:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/horizon.yaml
```diff
@@ -0,0 +1,10 @@
+os: horizon
+text: 'Dragonwave Horizon Compact'
+type: wireless
+icon: dragonwave
+over:
+    - { graph: device_wireless_snr, text: 'SNR' }
+    - { graph: device_wireless_power, text: 'Tx Power' }
+discovery:
+    - sysObjectId:
+        - .1.3.6.1.4.1.7262.2.2
```
laf: Looks like you can change that to just `1.3.6.1.4.1.7262.2.` to pick up more Horizon devices.

includes/definitions/horizon.yaml
```diff
@@ -0,0 +1,10 @@
+os: horizon
```
laf: Is this definitely the best os name? Is the firmware name?

includes/polling/os/horizon.inc.php
```diff
@@ -0,0 +1,13 @@
...
+ * LibreNMS
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+$version = trim(snmp_get($device, ".1.3.6.1.4.1.7262.2.2.1.3.2.1.0", "-Ovq"), '" ');
```
laf: No mib available?

includes/definitions/horizon.yaml
pheinrichs: Horizon would be the best of name IMO.

includes/polling/os/horizon.inc.php
pheinrichs: I have yet to find an easy source of getting MIBS for dragonwave. I'll keep digging to see if I can get a source on them to add them.

nosmowned:
Dragonwave Harmony Enhanced MC

[DW_HAR_MC_MIB.zip](https://github.com/librenms/librenms/files/1335125/DW_HAR_MC_MIB.zip)

PR #7271, opened 2017-08-31
add the multiple ldap servers in auth and ntp-client alert

thunder0666:
Add the multiple server into the ldap auth
example in config.php:
$config['auth_ldap_server'] = ["ldaps://ldap1.example.net","ldaps://ldap2.example.net","ldap://ldap3.example.net"];
the auth ldap chek the uri or not in server name.

in second the ntp-client
now we can alert if we have a offset +- value . 
example in config.php:
$config['apps']['ntp-client']['warning']   = 25; // 25 ms of offset in +-
$config['apps']['ntp-client']['critical']  = 50; // 50 ms of offset in +-

like nagios the alert have 3 values :
0 - mean OK
1 - mean warning
2 - mean Critical

and create alert like:
applications.app_type = "ntp-client" && applications.app_status = "2"

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/authentication/ldap-authorization.inc.php
+++ b/html/includes/authentication/ldap-authorization.inc.php
@@ -55,11 +55,11 @@ function init_auth()
 
     /* need if we use multiples ldap server */
     if (gettype($config['auth_ldap_server']) == 'array') {
-        foreach($config['auth_ldap_server'] as $ldap_server) {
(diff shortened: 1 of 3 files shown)


laf:
Thanks for the PR but you'll need to split these out into two PR's.

PR #7273, opened 2017-08-31
1.31.x

murrant:
Do not squash!


laf:
Is this merging 1.31.01 and 1.31.02 into 1.31?

laf:
Seen the comment in the other issue.

This is being merged in to 1.31 though, shouldn't we be creating 1.31.03 as 1.31.02 is the latest tag so people won't get this fix.

murrant:
@laf This is the change I mentioned we should be merging fixes into the 1.31 branch and not creating new 1.31.x branches.  Also, we cannot squash as that will change the hashes.

The tags are what determine if the users update or not   I also am planning on creating a new 1.31.03 tag.

PR #7274, opened 2017-08-31
Add multiples servers into the ldap auth

thunder0666:
Add multiples servers into the ldap auth
example in config.php:
$config['auth_ldap_server'] = ["ldaps://ldap1.example.net","ldaps://ldap2.example.net","ldap://ldap3.example.net"];
the auth ldap chek the uri or not in server name.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7274)

### DIFF
--- a/html/includes/authentication/ldap-authorization.inc.php
+++ b/html/includes/authentication/ldap-authorization.inc.php
@@ -47,15 +47,39 @@ function init_auth()
     if (! isset($_SESSION['username'])) {
         $_SESSION['username'] = '';
     }
-
-    /**
-     * Set up connection to LDAP server
-     */
-    $ldap_connection = @ldap_connect($config['auth_ldap_server'], $config['auth_ldap_port']);
-    if (! $ldap_connection) {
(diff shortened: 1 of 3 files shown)


murrant:
Can't you just use `$config['auth_ldap_server'] = "ldaps://ldap1.example.net daps://ldap2.example.net";`?

thunder0666:
actually i test if the $config['auth_ldap_server'] is a array or not. and i test the server into a loop to find how is alive.
if i use a char like a separator ( usualy coma or space or ???? ) it's a bit more complicated to split a string with a char and create them into a array .
but i you absolutly need that :) .....

laf:
@thunder0666 @murrant is referring to that `ldap_connect()` accepts multiple servers separated by strings so you don't need to do this. I think he does the same in the active_directory auth module:

`You can also provide multiple LDAP-URIs separated by a space as one string`

laf:
@thunder0666 Any comment?

laf:
Feel free to reply here and we can re-open this.

thunder0666:
laf your rigth i delete my repo.
thanks for your attention.

PR #7275, opened 2017-08-31
Fixed IPv6 host renaming

Zmegolaz:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Renaming a host which contains an IPv6 address fails, as it tries to move a directory with the exact hostname. The actual directory name has some characters escaped.
This commit updates the renamehost() function to trim [] and replace : with _, in the same way as rrd_name() in includes/rrdtool.inc.php does.

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7275)


murrant:
Thanks for the fix.

I just checked and there are other areas of the code that would be affected by the sanitized ipv6 address directories.

I think the best fix would be to implement a new funciton `get_rrd_dir($hostname)` or something.

Zmegolaz:
I was afraid of that. Just to be clear, the directories are already sanitized in the current code when first adding a device, this patch makes it possible to rename to/from that.

Would you like me to try to find all other places? I'm very new to LibreNMS, but I can give it a shot.

murrant:
That would be great.

Basically, you are looking for `$config['rrd_dir'] . $hostname` the second variable could be a lot of things but should be the hostname.

Zmegolaz:
I think I've found all places now. I haven't been able to test every thing I changed, since I don't have any place to add all applications. But all automatic tests passed at least. :)

Please let me know if you find anything weird.

murrant:
Looks great, thanks @Zmegolaz

PR #7278, opened 2017-08-31
fix: Clear instead of set alert if updates are disabled

ekoyle:
Rename argument in set_notification to be clearer about what it does.  The `-r` (remove) argument to `daily.sh` needs to evaluate to true to clear the alert or false to set it.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7278)


laf:
I'm not sure there is a need to rename the arg but the other change is already live:

https://github.com/librenms/librenms/blob/master/daily.sh#L154

ekoyle:
Apparently I missed that PR.  The idea behind changing the name is to prevent the same confusion in the future.

laf:
It's not a message that's being passed though is it, it's just a 0 or 1 (or > potentially)?

ekoyle:
Yes, then php is casting it to a boolean.

The name sounds like it is expecting the return from an executed command,
in which 0 is success and 1 is failure.  I think that is the real reason
this bug was introduced.

laf:
I think @murrant agreed the logic was reversed by accident.

But as mentioned, this fix is already in and I'm not sure the variable name change is suitable

ekoyle:
I believe that the logic was reversed because the name is inaccurate.  It is called `arg_result` in [daily.sh](https://github.com/librenms/librenms/blob/master/daily.sh#L113), but `-r` means `remove` in [daily.php](https://github.com/librenms/librenms/blob/master/daily.php#L106).

Renaming `arg_result` to `arg_remove` may prevent confusion in the future.  If you don't agree, feel free to close this PR.

laf:
@murrant ?

murrant:
I admit the naming is confusing.  Assume this pseudo-code function.

    set_notification($type, $result)

$result is a boolean represented numerically, not a return code, this is where I tripped myself up and added the bug.  (the fix is in master already)

I intend to add another type of notification in the future about the php version.

Fully renaming this function and it's arguments would be welcome if you can come up with better names, keeping the above in mind.

ekoyle:
@murrant This PR just renamed $arg_result to $arg_remove_message in
daily.sh to reflect what the argument actually does (with this change, it
now resembles the language in daily.php).

I would have made the same mistake with the current naming.

-- 
Eldon

On Sep 6, 2017 11:26 PM, "Tony Murray" <notifications@github.com> wrote:

I admit the naming is confusing. Assume this pseudo-code function.

set_notification($type, $result)

$result is a boolean represented numerically, not a return code, this is
where I tripped myself up and added the bug. (the fix is in master already)

I intend to add another type of notification in the future about the php
version.

Fully renaming this function and it's arguments would be welcome if you can
come up with better names, keeping the above in mind.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/librenms/librenms/pull/7278#issuecomment-327690167>,
or mute
the thread

murrant:
@ekoyle But that is still confusing.  If you set remove_message to false, does it create the message or do nothing?

ekoyle:
Still confusing, but unlikely for a user to get the logic reversed.

Another option could be to split it into two functions, one set_ and one
clear_ which set the appropriate values for -r.

murrant:
I like two separate functions better much less ambiguous.

ekoyle:
@murrant: I have not tested this last commit, but is this more of what you were thinking?

laf:
I actually think `local arg_remove_message=$2;` isn't right, it's not a message nor is it specific to removal. I still think the original variable name sufficed.
`
Rest of it gets a +1

ekoyle:
It is a boolean flag, if true the message is deleted if false it is added.
I don't see what that has to do with a "result", and it doesn't say result
anywhere in daily.php (daily.php uses "remove" iirc)...

Is there a better name that could be used both places?  Maybe success or
something?  I still think "result" sounds like it wants an exit status
passed (especially in a shell script), in which case 0 would indicate a
successful operation (delete the message) and nonzero would indicate
failure (the opposite of what it does now).

laf:
> It is a boolean flag, if true the message is deleted if false it is added.
> I don't see what that has to do with a "result", and it doesn't say result
> anywhere in daily.php (daily.php uses "remove" iirc)...

`arg_remove_message` so for false then that variable name is wrong. The original isn't the best either but no point swapping one variable for another if it doesn't clear things up.

Unless someone can think of a good name for the variable then I suggest we just leave it as is.

murrant:
@ekoyle You are thinking of the code as instructing daily.php what to do like a command.

I and maybe laf too are think of it as event.  So daily.sh informs daily.php that an update event happened and what the result was.

Not sure what is better, but mixing the two is only adding confusion.

ekoyle:
The existing `daily.sh` and `daily.php` seem to disagree on the meaning of `-r` (the arguments to `daily.php` are undocumented, so I am inferring the meaning there).  I would assume that `daily.php` is the authoritative source, since it defines the action taken on `-r`.  This is why I renamed the variable in `daily.sh`.

Maybe the best option would be to pick a word that starts with a different letter for the mnemonic (`-s` for `successful`?) and update both files.

Here is the language used in `daily.php` (which makes me believe `-r` means `remove_notification`):

```php
/* daily.php:105 */
        if ($options['r']) {
            remove_notification($title);
        } else {
            new_notification(
                $title,
                'The daily update script (daily.sh) has failed. Please check output by hand. If you need assistance, '
                . 'visit the <a href="https://www.librenms.org/#support">LibreNMS Website</a> to find out how.',
                2,
                'daily.sh'
            );
```

Here is the language from `daily.sh`, which leads me to believe that `-r` means `result`:

```sh
# daily.sh:110
set_notification() {
    local args="$@";
    local arg_type=$1;
    local arg_result=$2;

    php "${LIBRENMS_DIR}/daily.php" -f set_notification -t ${arg_type} -r ${arg_result};
}
```

murrant:
Yeah, definitely.  Let me add some comments here.
```
        if ($options['r']) {
            // the result was success, remove any standing notifications in case they exist
            remove_notification($title);
        } else {
            // the result was failure, post a notification of this failure
```

murrant:
Not sure where we came out here.

I reworked things in #7418.  Let me know if anyone wants to re-open this PR.

PR #7280, opened 2017-09-01
Syslog future entries cleanup

deepseth:
When switches reload, often we end up with entries in the database that show timestamps from the future. These entries then remain until the future date gets reached, which is undesirable.

This patch deletes any entries that are more than 7 days into the future. This accounts for the majority of clock skew, where devices can be hours or days ahead "legitimately" while solving for others.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7280)

### DIFF
--- a/daily.php
+++ b/daily.php
@@ -57,8 +57,23 @@ if ($options['f'] === 'syslog') {
                 break;
             }
         }
-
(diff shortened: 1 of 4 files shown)


deepseth:
I screwed this up, will fix and recreate.

PR #7282, opened 2017-09-01
Syslog future entries cleanup (take 2)

deepseth:
When switches reload, often we end up with entries in the database that show timestamps from the future. These entries then remain until the future date gets reached, which is undesirable.

This patch deletes any entries that are more than 7 days into the future. This accounts for the majority of clock skew, where devices can be hours or days ahead "legitimately" while solving for others.

 DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7282)

### DIFF
--- a/daily.php
+++ b/daily.php
@@ -67,15 +67,15 @@ if ($options['f'] === 'syslog') {
                 break;
             }
 
-            if (dbDelete('syslog', 'seq >= ? AND seq < ? AND timestamp > DATE_SUB(NOW(), INTERVAL 7 DAY)', array($rows, $limit)) > 0) {
+            if (dbDelete('syslog', 'seq >= ? AND seq < ? AND timestamp > DATE_SUB(NOW(), INTERVAL ? DAY)', array($rows, $limit, $config['syslog_future_purge'])) > 0) {
                 $rows = $limit;
-                echo "Syslog cleared for entries from more than 7 days in the future 1000 limit\n";
+                echo "Syslog cleared for entries from more than ".$config['syslog_future_purge']." days in the future 1000 limit\n";
             } else {
                 break;
             }
         }
 
-        dbDelete('syslog', 'seq >= ? AND timestamp > DATE_SUB(NOW(), INTERVAL 7 DAY)', array($rows));
+        dbDelete('syslog', 'seq >= ? AND timestamp > DATE_SUB(NOW(), INTERVAL ? DAY)', array($rows, $config['syslog_future_purge']));
     }
 }
 
--- a/daily.php
+++ b/daily.php
@@ -57,8 +57,23 @@ if ($options['f'] === 'syslog') {
                 break;
             }
         }
-
         dbDelete('syslog', 'seq >= ? AND timestamp < DATE_SUB(NOW(), INTERVAL ? DAY)', array($rows, $config['syslog_purge']));
+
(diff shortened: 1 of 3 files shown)


deepseth:
I'd actually like to change this to be a configurable setting, rather than forcing an arbitrary 7 day value on everyone. Can be defaulted to 7 days, but should have a $config var. Will update and commit when ready.

laf:
You need to doc this as well.

murrant:
Any reason not to bound incoming syslog entries by the current time instead of allowing times in the future.   I think that approach could hide issues with timezones and incorrect clocks, but is that something we care about?

deepseth:
I personally don't care, and would rather have all timestamps normalised to the host's local time, but that may not be true of everyone. I'm happy to dump this and just make this a simple documentation change where the vars being used by syslog server are altered instead.

kkrumm1:
I agree with both sides I think should have both options. 
One: to have the config options to purge future syslog entries. (this pull request)
Second: Bond syslog entries as they come into the syslog server time.  (also keep in some users may not know how to configure that in syslog) 
But we should give the users the options. 
Just my two cents. 👍

murrant:
I wouldn't mind both options with one config setting.

Like: `$config['future_syslog_handling'] = 'delete';` or 'rewrite' or 'none'.

Variable name could be better.

kkrumm1:
that would be awesome :)

laf:
So where are we up to with this?

kkrumm1:
I think it's gonna be change to the syslog.conf adding and documenting the change to docs.
adding the R_ to force syslog to use the host time when recoved and not the device time recived. 
` program("/opt/librenms/syslog.php" template ("$HOST||$FACILITY||$PRIORITY||$LEVEL||$TAG||$R_YEAR-$R_MONTH-$R_DAY $R_HOUR:$R_MIN:$R_SEC||$MSG||$PROGRAM\n") template-escape(yes));
`

deepseth:
Syslog-ng and rsyslog both now use received time rather than reported time
(rsyslog had conflicting examples to do either, depending on which block of text users opted to copy)
I'm happy to strip out the need for future deleting stuff as that's no longer really necessary

laf:
I just thought, why not just have a config option which stops the syslog entry being processed in the first place if the data is > by X days threshold.

laf:
@deepseth Any thoughts to my last comment?

murrant:
I'd say, just send the doc changes and call it good.  We can point people to update their configs if they have any issues.

laf:
Seems odd that we'd let data be stored that we know should be binned off. I'll leave you to merge if you like but seems more sensible to drop the data first to me.

murrant:
Yeah, but the syslog message did happen, just has the incorrect date.

laf:
I know, but we get rid of them on the next daily run anyway

deepseth:
Sorry, my intention was to remove the 'future delete' stuff, and just use the documentation change to have timestamps replaced on ingress at syslog-ng / rsyslog, since that essentially resolves the problem, and doesn't result in messages getting lost due to timestamp issues.

murrant:
@deepseth Are planning on updating this PR?

murrant:
Resubmitted in #7519

PR #7285, opened 2017-09-04
fix: Updated graph_types table so graph_subtype has no default value

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This should fix the validate issues. We insert a value into this column so it makes sense to just have it set as NOT NULL with no default value.

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7285)


murrant:
There was one more place I think.  Let me see if I can find it.

murrant:
Thanks laf, been meaning to get this fixed.

Issue #7218, opened 2017-08-24
Procurve Router 7102dl and  ProCurve Access Point 530 NA J8986A recognizes as a generic device


crc540:
LibreNMS recognizes these devices as a generic device and not a hp network device.

![untitled1](https://user-images.githubusercontent.com/31289840/29647164-c8caa6ae-8855-11e7-8ab6-c4617869d9bd.png)

![untitled](https://user-images.githubusercontent.com/31289840/29646907-415a418a-8854-11e7-8047-d4a576557954.png)        
Component | Version
--------- | -------
LibreNMS  | 1.30.01-100-g30c1b79
DB Schema | 204
PHP       | 7.0.22-0ubuntu0.16.04.1
MySQL     | 10.0.31-MariaDB-0ubuntu0.16.04.2
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3
====================================
[OK]    Database connection successful
[OK]    Database schema correct

DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

laf:
Thanks for the bug report or new device request, unfortunately you have not provided the information asked for in point 3 when you created this issue.

Please provide pastebin links to the information we need and we will re-open this issue.

crc540:
https://p.libren.ms/view/5a695e5a
https://p.libren.ms/view/e2326e56
https://p.libren.ms/view/0b21734f#L39  

I have gave you the info for one I have about 15 of the exact device having the same problem

[...]

PR #7286, opened 2017-09-04
newdevice: Added detection for HP Procurve Access Points

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7218

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7286)


includes/definitions/procurve-wa.yaml
```diff
@@ -0,0 +1,11 @@
+os: procurve-wa
+text: 'HP ProCurve Access Point'
```
murrant: I a bit confused by the name, this doesn't appear to be an access point at all.

ftp://ftp.hp.com/pub/networking/software/ProCurve-SR-7102dl-Quick-Start-Guide.pdf

kkrumm1: the original post was for two devices but I don't he supplied info for both.....
"Procurve Router 7102dl and ProCurve Access Point 530" 
https://github.com/librenms/librenms/issues/7218

laf: Good point. He's supplied data for a router and I looked at the AP side of the title. I'll tell the OP to create two separate requests with proper data.

murrant: It looks like this PR was mostly correct for the router.

PR #7289, opened 2017-09-05
added basic os for EricssonLG ES switches

tslytsly:
Used new-os script to add basic support for EricssonLG iPECS ES network switch.

There was already basic support for EricssonLG iPECS UCP PBX, but the files were all called `ericsson-es`.
So I renamed them to `ericsson-ucp` so that I could use `ericsson-es` for the ES switches.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7289)


includes/definitions/ericsson-es.yaml
```diff
@@ -1,9 +1,12 @@
 os: ericsson-es
-text: Ericsson iPECS ES
+text: 'EricssonLG SwitchOS'
 type: network
 icon: ericsson
+group: ericssonlg
```
laf: Needs changing to ericsson

laf:
Thanks, few changes needed but should be good after that :)

tslytsly:
> Is ericsson-es the right name now, should we be calling it ericsson-lg or ericsson-switchos?

Not sure, technically both the PBX and network switches are Ericsson LG iPECs devices. the PBX is the UCP and the switch is the ES. As far as I know ELG don't have an official name for the OS, so I just called it switchos.
Not sure what naming convention you want to follow?

> You'll need to re-add the test file for ericsson-es (or whatever it's new name).

Yep, when I pushed the commit I didn't realise I needed to do this, but I looked in the docs and will make sure both are correct.

tslytsly:
Hmm, I added test files and got these [errors.](https://p.libren.ms/view/02441d02)

Not sure how to read them.

Here are both test files:
ericsson-ucp:
```
1.3.6.1.2.1.1.1.0|4|Ericsson-LG,UCP100,UCP-R2.1.42&Boot Version-1.0Ea MAR/17
1.3.6.1.2.1.1.2.0|6|1.3.6.1.4.1.572.16838
```
ericsson-es:
```
1.3.6.1.2.1.1.1.0|4|ES-2026P Advanced Smart FE POE Switch
1.3.6.1.2.1.1.2.0|6|1.3.6.1.4.1.572.17389.107
```

laf:
> Not sure, technically both the PBX and network switches are Ericsson LG iPECs devices. the PBX is the UCP and the switch is the ES. As far as I know ELG don't have an official name for the OS, so I just called it switchos.

If they use the same OS then we should potentially keep them under on os discovery for us. Any reason why we wouldn't do that?



> Hmm, I added test files and got these errors.

It's not matching the data properly at the moment. Let's work on getting the OS names right and fix that later.

tslytsly:
> If they use the same OS then we should potentially keep them under on os discovery for us. Any reason why we wouldn't do that?

Well that's the thing. They are both branded as iPECS. But the PBX is a different beast from the switch. They do not share anything.
In fact, I suspect that the network switch is a re-badged netgear. 😄 

I think they just call them both iPECS to make them easier to sell as a package.

laf:
I'm fine with the name change, however I've just noticed why this isn't working. We already have an OS called ipecs which is one you are trying to add here.

tslytsly:
> I'm fine with the name change, however I've just noticed why this isn't working. We already have an OS called ipecs which is one you are trying to add here.

Ah, that will be it then.
I've pushed another commit to try and tidy this up some more. it's clear that it's not that clear. 😄

laf:
Unless you've got a genuine need to rename the current OS names then I'd suggest not doing this, we will have to notify users of the change for those that might reference these in device groups or alerting.

tslytsly:
>Unless you've got a genuine need to rename the current OS names then I'd suggest not doing this, we will have to notify users of the change for those that might reference these in device groups or alerting.

I think there is a need, the old names were not clear. In fact you could say that they were wrong.
I think that this follows the LibreNMS naming convention.
But I'm happy to just adjust the existing ericsson-es file to include the 2000 series switches and abandon the name change. 😄

laf:
If you really want to go down a name change then you'll need to update the notifications.rss feed [here](https://github.com/librenms/librenms.github.io/blob/master/notifications.rss) - see past commits to the file for an idea on what to put. Give users 5 days notice saying we will be changing the os names from X -> Y and that they will need to adjust alert rules / device groups.

murrant:
Also, os names cannot have an `_` in them, just change it to `-`.

tslytsly:
sorry for all the extra commits, still getting the hang of this git stuff. 😄

includes/polling/os/elg-ipecs-es.inc.php
```diff
@@ -1,8 +1,8 @@
 <?php
 /**
- * ericsson-es.inc.php
+ * ericsson-ucp.inc.php
```
laf: This and the line two below need updating to reference the new file name

tslytsly: is [this](https://github.com/librenms/librenms/pull/7289/commits/e4dbdb73e9bb548fe5d1e11c4dbcf86fecea6a90) right?

laf:
Actually you are missing the core test unit file for elg-ipecs-ucp, basically you need a default test file called elg-ipecs-ucp.snmprec so just rename one of the two to that and it should pass checks.

tslytsly:
@laf it failed again: https://p.libren.ms/view/a60a9727

Sorry, I don't really understand what these test files are doing, so I'm not sure what is wrong.
It seems to be still referencing the old os names with the _ in them, but not sure where from.

laf:
`git mv tests/snmpsim/elg-ipecs-ucp100.snmprec tests/snmpsim/elg-ipecs-ucp.snmprec`

tslytsly:
thanks @laf that sorted one of the errors, but I'm still getting this:
```
There were 2 failures:

1) LibreNMS\Tests\OSDiscoveryTest::testOS with data set #111 ('elg-ipecs_es')
No snmprec files found for elg-ipecs_es!

/opt/librenms/tests/OSDiscoveryTest.php:75 

2) LibreNMS\Tests\OSDiscoveryTest::testOS with data set #112 ('elg-ipecs_ucp')
No snmprec files found for elg-ipecs_ucp!

/opt/librenms/tests/OSDiscoveryTest.php:75
```
I cannot find where it is getting those references to the old os names from...

laf:
That must be local as it's fine in travis. Remove cache/os_defs.cache and it will rebuild.

tslytsly:
thanks, that worked.

laf:
> If these two OS use the same or mostly mibs, likely they should be the same OS.

I recently did this for a PR and did two OS' as the differences were enough.

I don't know this vendor / os enough to make that call here. Thoughts @murrant?

tslytsly:
> I recently did this for a PR and did two OS' as the differences were enough.
Just FYI:
iPECS-UCP is a PBX, IP phone system and iPECS-ES are L2 and L3 smart network switches.

laf:
I'm fine with the split. @murrant?

laf:
Ready for merging now on the 25th of Sept

Issue #3527, opened 2016-05-16
Cannot Create API tokens /w Active Directory Authentication


compuguy:
Currently using the latest release with Active Directory, but the authenticated users are unable to create api tokens. This may be related to the bug  #3069 and merge #3178. The create api key dialog box does not list any users.

compuguy:
I'm closing the ticket, because @Ultra2D mentioned the frontend is broken for AD/LDAP, and will be worked on in the upcoming UI.

PR #3178, opened 2016-03-07
API access when using LDAP authentication

Ultra2D:
Partly fixes #3069, but might break API access for authentication mechanisms that do store username and user_id in the database, but don't have the `get_user` function. Do those exist?

The `get_user` function I added is not pretty, but should do for now.

Issue #3069, opened 2016-02-23
API-tokens with LDAP authentication


Ultra2D:
The API part of the WebUI does not work with LDAP authentication at the moment. I have fixed it for myself for now, as this probably will be changed as part of the rewrite (#2965).

The id of the user creating a token is saved in the database. To display the tokens, function `get_user($user_id)` is used. Will that stay the same after the rewrite? 
I ask, because that function is not supported by LDAP authentication at the moment, but can be added quite easily.

awlx:
Same for Active Directory support :(.

TJuberg:
@Ultra2D Would you mind sharing your fix for the interim?

laf:
@TJuberg just insert a token straight into the tokens table in the DB.

TJuberg:
Thank you.

Just getting started with LibreNMS and don't know my way around yet :)

laf:
No worries, that's a bodge at the moment :(

Ultra2D:
@laf  get_user() is missing from LDAP auth. Should it be added?

compuguy:
I think this hasn't been fixed even with merge #3178.

Ultra2D:
Correct, this is caused by looking for users in the database, which does not work for LDAP users  ( https://github.com/librenms/librenms/blob/5600bf3995024b11bb49e2294d2b4d4bc37b9dde/html/pages/api-access.inc.php#L57 ). But I think this should be fixed during/after the rewrite (also see https://github.com/librenms/librenms/issues/3128#issuecomment-194494246 ).

f0rkz:
Is there any update to this issue?

networkpadawan:
any update or workaround?

Ultra2D:
This issue is closed, so you might have to create a new one if it does not work for you. For me, creating an access token using the web interface does work at the moment. Otherwise you can always add an entry to table 'api_tokens' manually .

dneto82:
HI. I created the token manually in database, but still getting.

"message": "API Token is missing or invalid; please supply a valid token"

murrant:
Does LDAP support using a bind user?  Otherwise, it cannot perform ldap actions that require authentication in the api.

zlesnr:
@laf I don't believe this to be fixed.  Would you mind reopening this please?  We've just brought up LibreNMS in our environment and we'd like to use the API, but strictly use LDAP for authentication.

PR #7291, opened 2017-09-05
Add config option "api_auth_local" to bypass auth_mechanism for API-requests

janegilv:
Adds an option (api_auth_local) to the config. This can be set to true to force use of local mysql database (token- and user-info) for API-requests. Done in our case to not have it check LDAP-database for API-requests (which was broken). 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7291)

### DIFF
--- a/doc/Extensions/Authentication.md
+++ b/doc/Extensions/Authentication.md
@@ -232,3 +232,20 @@ $config['auth_ad_binduser']     = "ad_binduser";
 $config['auth_ad_bindpassword'] = "ad_bindpassword";
 $config['auth_ldap_cache_ttl']  = 300;
 ```
+
+### API access 
+
+Config option: `auth_api_local`
+
+The API uses the authentication extension set with `$config['auth_mechanism']` by default. It is possible to have API calls authenticated against the mysql-database with the following config (default value is false):
+
+```php
+$config['auth_api_local'] = true;
+```
+
+It is not possible to add local tokens from the web-interface if `auth_mechanism` is something else than mysql. 
+To add API-tokens/users you will have to either:
+
+- manually alter the database
+
+- temporarily set `$config['auth_mechanism'] = mysql;` and add them from the webinterface
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -12,14 +12,14 @@
  * the source code distribution for details.
  */
 
+use LibreNMS\Config;
+
 function authToken(\Slim\Route $route)
 {
-    global $config;
-
     $app   = \Slim\Slim::getInstance();
     $token = $app->request->headers->get('X-Auth-Token');
     if (isset($token) && !empty($token)) {
-        if (!function_exists('get_user') || $config['api_auth_local'] === true) {
+        if (!function_exists('get_user') || Config::get('api_auth_local', true)) {
             $username = dbFetchCell('SELECT `U`.`username` FROM `api_tokens` AS AT JOIN `users` AS U ON `AT`.`user_id`=`U`.`user_id` WHERE `AT`.`token_hash`=?', array($token));
         } else {
             $username = get_user(dbFetchCell('SELECT `AT`.`user_id` FROM `api_tokens` AS AT WHERE `AT`.`token_hash`=?', array($token)));
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -14,10 +14,12 @@
 
 function authToken(\Slim\Route $route)
 {
+    global $config;
+
     $app   = \Slim\Slim::getInstance();
     $token = $app->request->headers->get('X-Auth-Token');
     if (isset($token) && !empty($token)) {
-        if (!function_exists('get_user')) {
+        if (!function_exists('get_user') || $config['api_auth_local'] === true) {
             $username = dbFetchCell('SELECT `U`.`username` FROM `api_tokens` AS AT JOIN `users` AS U ON `AT`.`user_id`=`U`.`user_id` WHERE `AT`.`token_hash`=?', array($token));
         } else {
             $username = get_user(dbFetchCell('SELECT `AT`.`user_id` FROM `api_tokens` AS AT WHERE `AT`.`token_hash`=?', array($token)));
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -19,7 +19,7 @@ function authToken(\Slim\Route $route)
     $app   = \Slim\Slim::getInstance();
     $token = $app->request->headers->get('X-Auth-Token');
     if (isset($token) && !empty($token)) {
-        if (!function_exists('get_user') || Config::get('api_auth_local', true)) {
+        if (!function_exists('get_user') || Config::get('api_auth_local', false)) {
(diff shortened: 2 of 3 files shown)


jallakim:
Previous relevant issues and/or PRs; #3069 + #3527 + #3178.

html/includes/api_functions.inc.php
```diff
@@ -14,10 +14,12 @@
 
 function authToken(\Slim\Route $route)
 {
+    global $config;
+
     $app   = \Slim\Slim::getInstance();
     $token = $app->request->headers->get('X-Auth-Token');
     if (isset($token) && !empty($token)) {
-        if (!function_exists('get_user')) {
+        if (!function_exists('get_user') || $config['api_auth_local'] === true) {
```
laf: Can you switch to using `Config::get()` for this:

https://github.com/murrant/librenms/blob/5f1a1fc1e7da177374cce16802706ccfcba03e36/html/includes/authentication/active_directory.inc.php#L39

html/includes/api_functions.inc.php
```diff
@@ -12,12 +12,14 @@
...
 
+use LibreNMS\Config;
+
 function authToken(\Slim\Route $route)
 {
     $app   = \Slim\Slim::getInstance();
     $token = $app->request->headers->get('X-Auth-Token');
     if (isset($token) && !empty($token)) {
-        if (!function_exists('get_user')) {
+        if (!function_exists('get_user') || Config::get('api_auth_local', true)) {
```
jallakim: Shouldn't this be `Config::get('api_auth_local', false)`, to be uniform with `$config['api_auth_local'] = false;` specified in the `defaults.inc.php` file?

jallakim: Shouldn't this be `Config::get('api_auth_local', false)`, to be uniform with `$config['api_auth_local'] = false;` specified in the `defaults.inc.php` file?

janegilv: Absolutely correct mister

laf:
Thanks for this, unfortunately one of the commits you did was using `Jan Egil Vestbø` which hasn't signed the CLA. You'll probably need to squash these commits and force push as your github user to fix.

murrant:
Out of curiosity, would it be possible to fix the ldap auth module?  API authentication works with the active directory auth module.

jallakim:
@murrant, this option would still be useful even if we fix the LDAP auth module to work with the API (in our case we don't want/need API-users to be in LDAP, but we still want LDAP for normal user logon).

murrant:
@jallakim Hmm, makes sense.  Basically, you want dedicated API users.

But how will you manage these users?  They won't show up in the user management (or anywhere in the webui) if you have ldap enabled.  Essentially, they are hidden users.

jallakim:
@laf, this should be good to go now, yes?

@murrant, that's a discussion for another time, as there is many ways to solve this;

- Hybrid between local and external auth (you can use LDAP/AD, and still have/manage local users). If there are overlapping users, how should one handle this?
- Assign API-keys to externally authenticated users (LDAP/AD/whatnot)
- Create full-permission API-keys that are not "assigned" to a specific user, and as such, is independent of whatever authentication mechanism you use (local or external).

Neither of these sticks out as a "clear winner", and some, if not all, of them could also be combined in some way.

Hence, the override option in this PR is a simple way to at least provide the basic functionality, albeit at the cost of having to manage it manually directly via the database.

laf:
@jallakim Option 2 is what murrant was referring to, he's done the same for AD that using bind users means API keys work for AD auth so is this something similar we can do for ldap?

jallakim:
@laf, of course it's possible, but outside the scope for this PR? Even if we implement that, there is still a use case for having non-LDAP API-keys, even when you're doing LDAP-authentication.

Can we just close/merge this PR? If-so-be we can also create a new issue addressing the need to fix API-auth via LDAP, similar to what's been done for AD.

laf:
We don't just merge PRs if we don't feel like it's the right direction to go :)

jallakim:
@laf, but, as I've stated multiple times now, there is still a use case for this option, even if/when we fix LDAP API-tokens. They are not mutually exclusive. I feel like you've been dodging this part of my previous messages.

laf:
> @laf, but, as I've stated multiple times now, there is still a use case for this option, even if/when we fix LDAP API-tokens. They are not mutually exclusive. I feel like you've been dodging this part of my previous messages.

No, I'm responding to your comments you keep making. I already said in my first reply this looks a sensible way forward.

> (In my use case, I'd like to have LDAP authentication for normal login, but still local API-tokens for the API -- even when/if we fix LDAP API-tokens; I have no interest in having API-tokens for scripts based on LDAP-users).

And my responses previously are still valid, we don't merge PRs because it's best for one user, we have to and try to take other considerations into account. Just because a PR fixes a problem doesn't mean it's the right fix.

To try and answer your potential next response, my responses are not specific to this PR. I'm trying to explain to you why PRs may or may not get merged in. It's not personal, it's for the good of the project. Now having said that, my thoughts on this are that we should merge it. We may decide to remove it later on as this is in essence a bypass but who knows.

jallakim:
> No, I'm responding to your comments you keep making. I already said in my first reply this looks a sensible way forward.

My bad; I interpreted it as if you were reversing your previous "this looks fine", and replaced it with "please fix it like this".

> To try and answer your potential next response, my responses are not specific to this PR. I'm trying to explain to you why PRs may or may not get merged in. It's not personal, it's for the good of the project. Now having said that, my thoughts on this are that we should merge it. We may decide to remove it later on as this is in essence a bypass but who knows.

I understand this -- it was the assumption that you reversed "this looks fine" that put me in that mode (-:

doc/Extensions/Authentication.md
```diff
@@ -232,3 +232,20 @@ $config['auth_ad_binduser']     = "ad_binduser";
...
+
+Config option: `auth_api_local`
+
+The API uses the authentication extension set with `$config['auth_mechanism']` by default. It is possible to have API calls authenticated against the mysql-database with the following config (default value is false):
+
+```php
+$config['auth_api_local'] = true;
+```
+
+It is not possible to add local tokens from the web-interface if `auth_mechanism` is something else than mysql.
```
murrant: This sentence is incorrect.  It works properly for AD auth.

murrant:
Often, I will ask questions about a PR to make sure we think it through.

I think this PR is probably fine, but `api_auth_local` feels a bit ambiguous to me, perhaps `api_auth_mysql`.  I noted an issue with the documentation.

laf:
@jallakim Can you test @murrant's new PR: https://github.com/librenms/librenms/pull/7335?

laf:
@jallakim If you have time to respond to the last comment, please reply and we can re-open this.

jallakim:
It would would be @janegilv who would have to test it.

jallakim:
@laf, so, as @janegilv wrote in the PR; it works. However, that still doesn't solve the use case for having API-keys locally (i.e. not dependent on external authentication sources in order for them to work), which is more or less what this PR was created for. I'm not saying this PR should be approved in it's current state, and I have no strong opinion on how this should be implemented.

Issue #7249, opened 2017-08-29
New Device - SAF Integra (non W)


kmpanilla:
SAF has another model in the Integra line-up. It's not the W line.. just called simply the "Integra" as far as I can tell.  I'd like to submit the following so someone can add it to the available devices.

SAF Integra
DISCOVERY: 	https://p.libren.ms/view/1b74f454
POLLER:    	https://p.libren.ms/view/1d017251
SNMPBULKWALK:   https://p.libren.ms/view/e5d4c2a5
MIBS:
 SAF-INTEGRAB:   https://p.libren.ms/view/9fdc752b
 SAF-INTEGRA:    https://p.libren.ms/view/abb7a6db
 SAF-ENTERPRISE: https://p.libren.ms/view/254291e8

Proposed OS Name: SAF Integra

There are three vendor supplied MIBs.  The SAF-INTEGRA and SAF-ENTERPRISE are also required for the SAF-INTEGRAW host which I also requested.

Please let me know if I can provide a live device for testing against, or any other information.

Thanks.

laf:
I'm not sure it's worth splitting these out. I'll add the support to the other PR based on the W model.

laf:
Change of plan, I split them out but in the same PR.

Issue #7248, opened 2017-08-29
New Device - SAF Integra-W


kmpanilla:
I'd like to submit details for a new SAF device. The SAF Integra-W.

DISCOVERY:	https://p.libren.ms/view/d3f771d5
POLLER:		https://p.libren.ms/view/495d7128
SNMPBULKWALK:   https://p.libren.ms/view/06f53ceb	
MIBS:
 SAF-INTEGRAW:   https://p.libren.ms/view/5dd8a013 
 SAF-INTEGRA:    https://p.libren.ms/view/abb7a6db
 SAF-ENTERPRISE: https://p.libren.ms/view/254291e8

Proposed OS Name: SAF Integra-W

This is unique to the SAF Lumina that is already supported. Supplied are links to the three MIBs supplied by SAF.

I tried to code it in manually, but have been running into time constraints and issues, so I thought I'd provide the details needed to let someone more skilled than I add it.

Please let me know if I can provide other details or access to a live device. We have a number of them.

PR #7292, opened 2017-09-05
newdevice: Added support for Saf Integra Access points

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7248 
Fixes: #7249

image

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7292)


laf:
This includes two OS'.

I did originally have them in one but it seemed silly, the code is sort of duplicated but each has it's own MIB and oids so separate seems sensible.

includes/definitions/saf-integra-b.yaml
```diff
@@ -0,0 +1,14 @@
...
+group: saf
+over:
+    - { graph: device_bits, text: 'Device Traffic' }
+    - { graph: device_processor, text: 'CPU Usage' }
+    - { graph: device_mempool, text: 'Memory Usage' }
+discovery:
+    - sysObjectId:
+        - .1.3.6.1.4.1.7571.100.1.1.7.1
+poller_modules:
+    wifi: 1
```
murrant: You don't need to enable the wifi polling module, that is only legacy.

PR #7303, opened 2017-09-08
Proposal: ignore loopback interfaces by default

florianbeer:
Adding server & network device loopbacks to default ignored ports.

Monitoring loopback doesn't add any value IMHO and could even trigger confusing alerts for new users: https://community.librenms.org/t/port-utilisation-over-threshold-for-lo-interface/2041

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7303)

### DIFF
--- a/includes/defaults.inc.php
+++ b/includes/defaults.inc.php
@@ -560,6 +560,7 @@ $config['bad_iftype'][] = 'aal5';
 $config['bad_iftype'][] = 'shdsl';
 $config['bad_iftype'][] = 'mpls';
 
+$config['bad_if_regexp'][] = '/^lo[0-9]*.*/';
 $config['bad_if_regexp'][] = '/^ng[0-9]+$/';
 $config['bad_if_regexp'][] = '/^sl[0-9]/';
 


murrant:
@florianbeer I will be submitting a change to allow the interface ignores to be defined at the os level.  Would that help with this?

florianbeer:
@murrant os-level ignores are definitely a welcome addition, but I'm thinking more of a "sane defaults" way here.

To discuss: Is there any merit in ever monitoring the default loopback of a device?

If not, this PR aims to reduce clutter and possible confusion for new users by completely removing it from the web interface.

zombah:
Hello,
Loopback usualy host ip addresses for device management, dynamic routing and source ip's for various protocols.
I think its fine not to poll their ifSpeed but ignoring them in web interface is not
good idea imho.

murrant:
@florianbeer I do have some crappy devices that report the statistics for the ethernet interface on the lo interface :(

florianbeer:
See that's why I needed your crowd intelligence :)

I'm a systems engineer so I deal with servers and typically out of band management is mapped to a private subnet or firewalled public IP so I never have anything I want to monitor on my loopbacks. But seeing that there are use cases I haven't come across, I'm going to close this now.

It's easy enough to add the proposed line to config.php and ignore them for individual installs.

murrant:
@florianbeer I would feel much more comfortable adding this into the os yaml files for specific OS ;)

florianbeer:
Might make sense, at least on my servers (Linux) it could be a good addition.

PR #7313, opened 2017-09-10
Update to use DataTables instead of BootGrid

geordish:
As mentioned on a previous issue, bootgrid is no longer maintained. I've taken a look to switching to DataTables instead.

This patch just moves the ports page over as a test.
Before:
image

After:
image

Currently column sorting isn't working, but the rest of it is functional.

If the consensus is that this is worthwhile, I will continue to change all places where bootgrid is used is switched over.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7313)


laf:
No issues from me :)

What's needed to get column sorting fixed?

murrant:
There is a bootstrap plugin for datatables,  would be best to use that.

geordish:
@murrant: bootstrap plugin:
image

@laf Shouldn't take much. Just didn't want to invest too much time up front.

geordish:
This should now be working exactly the same as before, but with DataTables.

florianbeer:
In my opinion the visual aspect is quite important too and the bootgrid version was nicer to look at in my eyes. Less clutter and generally more styled.

Great job on replacing the outdated library though, thanks!

geordish:
Reopened on #7318

PR #7323, opened 2017-09-12
Support for up/down detection of ping only devices.

Zmegolaz:
This allows the user to disable all SNMP checks and only ping a device. No extra service checks are needed and up/down status and alerting works as expected.
I have a few points I'd like to ask your opinion about:

1. Should we add a "Disable SNMP" button to the "Add device" page too? snmp-scan.py (despite the name)?
2. The OS is set to "Ping only" and the user can specify hardware. Should it be called something else? Should the user be able to specify more details?
3. The disable SNMP setting is currently saved in a separate column and the code checks that value. However we could instead only set the OS of the device and use that field to decide if SNMP checks should be performed. It's not as clear and probably not as future proof, but it'd be a column less in the database.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7323)


laf:
Definitely a good starting point. I'll comment further when I've got more time.

Zmegolaz:
The Scrutinizer inspection mentions two bugs, a variable "does not seem to be defined for all execution paths leading up to this point", but it's wrong, that point can never be reached without the variable defined. It's very trivial to fix, but maybe a bit unnecessary. Up to you. :)

murrant:
@Zmegolaz what if instead of a separate column, we just set the OS to "ping"?  Not sure if that is better or worse.

Zmegolaz:
@murrant, Yeah, that's what I was thinking about in point 3. I can't make up my mind about what I think is the best way.

Yes! Icon. I'll get one.

laf:
So my thought on this is that if a user adds a device which is only ICMP then we should at that stage allow them to select the OS it is - we will at that stage do nothing more with it other than make use of the logo and type, description etc.

It does need a few more areas:

  - Adding new hosts via CLI and WebUI need the option to allow just ICMP only.
  - A lot of graphs are going to be attempted to be shown including a lot of the usual tabs. We should hide those in the best way we can but I'm not sure what that is right now.

I've not checked code still yet, I think we should discuss how this PR should go before doing that.

laf:
> Letting the user select OS can surely be done. A dropdown might be too long, but maybe an autocomplete field?

Yes an auto complete will be fine. We already have some code in the graph widget that does this and can be re-used.


> Actually, very few invalid graphs and tabs are attempted. The only ones we need to hide are the Poller Modules Performance graph and the Maps tab.

The popups in the top right will need to be hidden. Not sure what else.

murrant:
@laf I think I would rather keep the os as "ping" that way it is more clear what data we have and we can customize the "ping" os to hide as much snmp stuff as possible.

However, we do need the ability for the user to customize the display of this a bit.  Like setting an icon.  So maybe, we let the use select an "os" but only set icon and some descriptive text, not the actual os value.

laf:
> However, we do need the ability for the user to customize the display of this a bit. Like setting an icon. So maybe, we let the use select an "os" but only set icon and some descriptive text, not the actual os value.

That's all dealt with by the OS so I really don't see the point in using a single OS. It limits the ability to do device groups and alerting on the correct OS all to save hiding some graphing info.

Gorian:
@murrant - I'd like to piggyback off of this code to add support for AWS services as OSes, so ideally you would add the code to disable SNMP checks, and other OSes could use that without having to be a "ping" OS...

murrant:
@laf I'm not sure which would be better.

We could also go the route of separating ping devices entirely in the ui to keep them from making things crappy.

Might need to try out various things and see which works the best.

Zmegolaz:
I did a proof of concept for allowing the user to specify OS to see it in action. Personally, I think it turned out nice, it was relatively clean.

laf:
Still no code review right now but testing it at present. Some more feedback:

  - Can you apply the select box UI we have (see misc)
  - We need addhost.php and Add Device in the UI to support adding non-snmp devices.
  - Not a fault of this PR but worth fixing, if ICMP is disabled then remove the ping response graph: `/graphs/type=device_ping_perf/device=342/to=1506979800/from=1506893400/`

I'm happy with how it works though. It's not quite right if you change an existing device over to be non-snmp due to all the data that did exist but I'm not sure that's one for this PR.

laf:
Adding a device via WebUI, the OS search isn't great. Search for Arista and you don't get any results. You actually don't seem to get anything other than a default list.

Zmegolaz:
Whops, I broke it when fixing a Scrutinizer issue. Fixed again.

FTBZ:
Small question, "ping only" device will be listed like others? It will be possible to add services to this device? If yes, it will be really nice and will simplify understanding for services management.

laf:
@Zmegolaz Is this complete now as far as you're concerned so I can code review?

Zmegolaz:
@laf Yes! Everything should be implemented now.

murrant:
Alright got around to testing this.

The OS field should be clearly marked as optional.  I think calling this OS is misleading here, but I'm not that worried about it.

I think the edit menu should be filtered.  Things to remove: Port Settings, Applications, Modules, Storage, Processors, Memory, Components  Not sure: IPMI, Misc

I haven't looked at the code yet.  I was thinking about icons a bit and whipped one up.  Not sure about it but here:

ping svg

laf:
> The OS field should be clearly marked as optional. I think calling this OS is misleading here, but I'm not that worried about it.

As in choosing the OS for the device? Why would we have that as optional?

> I haven't looked at the code yet. I was thinking about icons a bit and whipped one up. Not sure about it.

What's the icon for? If it's for the ping.yaml then imho that should be binned off unless you want it as just another OS option people can choose for ping only devices.

addhost.php
```diff
@@ -16,7 +16,7 @@
 $init_modules = array();
 require __DIR__ . '/includes/init.php';
 
-$options = getopt('g:p:f::');
+$options = getopt('Pg:p:f::');
```
laf: Not sure we should use upper case P here. Maybe i = icmp?

addhost.php
```diff
@@ -181,6 +196,7 @@
     c_echo(
         "\n".$config['project_name_version'].' Add Host Tool
 
+    Usage (ICMP)     : ./addhost.php [-g <poller group>] [-f] -P <%Whostname%n> [os] [hardware]
```
laf: Move the example for this below SNMP so as to not cause confusion. Change `Usage (ICMP)` to `Usage (ICMP Only)`.

html/ajax_ossuggest.php
```diff
@@ -0,0 +1,70 @@
...
+        $ret["$lev"] = $elem;
+    }
+
+    ksort($ret);
+    return $ret;
+}
+
+header('Content-type: application/json');
+if (isset($_GET['term'])) {
+    $_GET['term'] = mres($_GET['term']);
```
laf: Switch `mres()` for `clean()`

html/pages/addhost.inc.php
```diff
@@ -260,4 +291,62 @@ function changeForm() {
...
+        });
+
+    $("#os").on("typeahead:selected typeahead:autocompleted", function(e,datum) {
+        $("#os_id").val(datum.os);
+        $("#os").html('<mark>' + datum.text + '</mark>');
+    });
+
+    $("[name='snmp']").bootstrapSwitch('offColor','danger');
+<?php
+if ($_POST['hostname'] && $_POST['snmp'] != 'on') {
```
laf: Wrap `$_POST` access with `clean()`

html/pages/device/edit/snmp.inc.php
```diff
@@ -2,113 +2,137 @@
...
-        $max_oid      = mres($_POST['max_oid']);
-        $v3           = array(
-            'authlevel'  => mres($_POST['authlevel']),
-            'authname'   => mres($_POST['authname']),
-            'authpass'   => mres($_POST['authpass']),
-            'authalgo'   => mres($_POST['authalgo']),
-            'cryptopass' => mres($_POST['cryptopass']),
-            'cryptoalgo' => mres($_POST['cryptoalgo']),
-        );
+        if ($_POST['snmp'] == 'on') {
```
laf: Wrap this in `clean()`

laf: Same for all other _POST or _GET access.

snmp-scan.py
```diff
@@ -147,6 +147,7 @@ def scan_host(ip):
     Example: 192.168.0.0/24
     Example: 192.168.0.0/31 will be treated as an RFC3021 p-t-p network with two addresses, 192.168.0.0 and 192.168.0.1
     Example: 192.168.0.1/32 will be treated as a single host address""")
+    parser.add_argument('-P', '--ping', action='store_const', const="-P", default="", help="Only try to ping the device.")
```
laf: Maybe switch to -i / --icmp. Does this also need an example adding?

addhost.php
Zmegolaz: I chose -P here and in snmp-scan.py to get the same letter. -i is already used in snmp-scan.py, and -p is used here in addhost.php. I can still change it to -p and -i if you'd prefer.

laf: Maybe leave it then. Just easy to confuse people that's all.

laf:
@librenms/reviewers I'd like to merge this tomorrow - any further comments?

includes/polling/functions.inc.php
```diff
@@ -259,18 +259,22 @@ function poll_device($device, $options)
...
-        include 'includes/polling/core.inc.php';
-
         $force_module = false;
-        if ($options['m']) {
-            $config['poller_modules'] = array();
-            foreach (explode(',', $options['m']) as $module) {
-                if (is_file('includes/polling/'.$module.'.inc.php')) {
-                    $config['poller_modules'][$module] = 1;
-                    $force_module = true;
+        if ($device['snmp_disable'] != '1') {
```
murrant: Probably should just be `!$device['snmp_disable']`

snmp-scan.py
```diff
@@ -120,7 +121,7 @@ def scan_host(ip):
             pass
 
         try:
-            add_output = check_output(['/usr/bin/env', 'php', 'addhost.php', hostname or ip])
```
murrant: I feel link snmp-scan.py option should be allow ping only devices to be added, this way it tries snmp first.

includes/functions.php
```diff
@@ -2052,7 +2058,10 @@ function device_is_up($device, $record_perf = false)
     $response              = array();
     $response['ping_time'] = $ping_response['last_ping_timetaken'];
     if ($ping_response['result']) {
-        if (isSNMPable($device)) {
+        if ($device['snmp_disable']) {
```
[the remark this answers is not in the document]
Zmegolaz: It's a good idea to add a check that makes isSNMPable() return true if SNMP is disabled, but it would make the name of the function a bit misleading. Is that fine or should we rename it?

snmp-scan.py
Zmegolaz: I think I prefer to keep it as ping only. If someone really want to do a combined scan they could start with the SNMP scan to add all SNMP devices and then do a ping scan to add the rest. Otherwise we would lose the ability to add devices as ping only if they happen to reply to SNMP.

It would also take longer time to scan it, but I don't think that matter much.

includes/functions.php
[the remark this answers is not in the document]
murrant: Ignore that, but perhaps merge these two code blocks:
```
        if ($device['snmp_disable'] || isSNMPable($device)) {
            $response['status']        = '1';
            $response['status_reason'] = '';
```

PR #7331, opened 2017-09-13
RouterOS support for monitoring a USB connected UPS

troykelly:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7331)

### DIFF
--- a/includes/definitions/discovery/routeros.yaml
+++ b/includes/definitions/discovery/routeros.yaml
@@ -1,4 +1,4 @@
-mib: MIKROTIK-MIB:UPS-MIB
+mib: MIKROTIK-MIB
 modules:
     sensors:
         temperature:
@@ -9,11 +9,6 @@ modules:
                     num_oid: .1.3.6.1.4.1.14988.1.1.19.1.1.6.
                     descr: mtxrOpticalName
                     index: 'mtxrOpticalTemperature.{{ $index }}'
-                -
-                    oid: upsBatteryTemperature
-                    num_oid: .1.3.6.1.2.1.33.1.2.5.
-                    descr: UPS Battery Temperature
-                    index: 'upsBatteryTemperature.{{ $index }}'
         voltage:
             data:
                 -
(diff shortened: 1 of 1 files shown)


murrant:
Nice work @troykelly.  But, does adding `rfc1628_compat: 1` to includes/definitions/routeros.yaml work too?

laf:
From a check of the first few OIDs it does look like it would.

troykelly:
@murrant I'm really not sure - if there is a better way to do it - I'm happy to - I'm just struggling a little with how to add things properly.
Is there somewhere I should be looking outside of [this doc](https://docs.librenms.org/Developing/os/Health-Information/)?

troykelly:
@murrant @laf When I added the `rfc1628_compat: 1` to the file - most things appear to still work - except for the standby time. RouterOS reports it in seconds - so it would need a divisor.
Assuming that the preference is to use this method - is there a way to specify the divisor specifically for that one MIB?

murrant:
@troykelly rfc1628_compat: 1 is to save duplication.  If not for that, your could would ave been absolutely correct.

To fix the divisor you can modify the get_device_divisor() function.  This is were we account for quirks:
https://github.com/librenms/librenms/blob/master/includes/discovery/functions.inc.php#L943

I'm planning on adding more documentation for things like rfc1628_compat, right now the only documentation is the code.

troykelly:
Apologies @murrant - any advice on the below - it doesn't seem to be affecting the displayed value. [Line 969](https://github.com/librenms/librenms/blob/master/includes/discovery/functions.inc.php#L969) [Code](https://github.com/reallyreally/librenms/blob/master/includes/discovery/functions.inc.php#L969)
```
    } elseif ($device['os'] == 'routeros') {
        if ($sensor_type == 'runtime') {
            return 60;
        }
    }
```

murrant:
@troykelly looks correct, did you run discovery again? also, sometimes the os cache can give issues.  ` rm cache/os_defs.cache`

troykelly:
I've tried removing the cache, and always do discovery before polling when testing `./discovery.php -h 999 && ./poller.php -h 999`

I've put an example discovery & poll in the [community post](https://community.librenms.org/t/mikrotik-ups-snmp/2054/5) if that provides any insight?

murrant:
Actually, I looked at the runtime code in `includes/discovery/sensors/runtime/rfc1628.inc.php`
60 is already the divisor there, so the code in get_device_divisor() is not needed.

Perhaps there is a bug somewhere, I'll give it a look.

murrant:
I found the issue,rfc1628 runtime doesn't use get_device_divisor().  I've sent PR #7340, which fixes that.

troykelly:
@laf @murrant I've been kicked off the forum because multiple users reported my posts as spam. No point contributing where it's not wanted.
Cheers for the help.

laf:
I think you need to relax a little @troykelly. No one has kicked you off or even reported you, discourse the forum we use has flagged your account automatically as it triggered some of it's anti-spam thresholds:

image

With regards to your comment about this community being no better than Observium, that is absolutely not true. I can understand why you might have thought that from being blocked but it's an automated system and no one's maliciously trying to get you banned.

I'll leave it up to you if you want to re-open this. We welcome all contributions here.

troykelly:
Could I gently suggest then that the wording of the message I received (three times) be changed?
screen shot 2017-09-15 at 4 58 06 pm
This could have easily been avoided had I known it was an automated spam trap. The wording made it clear (although incorrectly) that my posts had been specifically flagged by several other users.

laf:
That is the default wording from the discourse forum and not something we've programmed or added in, I think we can change it so we'll take a look.

murrant:
@troykelly I apologize for the block and the inappropriately worded email. `githbub.com` and `docs.librenms.org` have been added to the list of domains new users are allowed to repeatedly link to.  We will also be changing the wording of the email.  Thanks for letting us know :)

PR #7332, opened 2017-09-13
feature: Host dependencies

aldemira:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I felt obliged to produce this code :) . It obviously needs vigorous testing, I'll ask it on the forum thread. I really wouldn't want to break the alerts!


kkrumm1:
let us know when it's ready for testing. I can test it out on one of my dev servers.

aldemira:
I think it is ready for testing now.

laf:
Wow, awesome work :)

No code review yet but a quick play around:

  - Delete parent host button not working for me due to mysql strict (we should always aim to support this with new queries: MySQL Error: Incorrect integer value: '' for column 'parent_id' at row 1 (UPDATE `devices` set `parent_id` ='' WHERE `device_id` =  '173'))
  - We should definitely add the ability to show the bulk host dependancies.
  - Make the devices in the /host-dependencies/ section clickable.
  - Whilst the /host-dependencies/ page has been bootgrid enabled (someone is actually converting these to Datatables which should then be done here, it does however need server side processing done. I have 4k+ devices so this page won't work well for me otherwise :)
  - Do we need to index the new column?

aldemira:
@laf 
Yes, I'm not a big fan of bootgrid, so to do the conversion, I'd like to wait until datatable pr is merged, and that will save me from any possible redundant work. 
For the db index, I'm using join to get the parent hostname from the devices table, so I thought it would make things faster. 
I'm working on the rest meanwhile

murrant:
@aldemira You can prepend `EXPLAIN ` to your sql queries to check the execution plan and see if it uses the correct indexes.

aldemira:
Yes I checked it, and it's not being used. So I'm removing the index line.

barryodonovan:
Hi,

Following code (includes/alerts.inc.php lines 889-892), $result will always be falsy since 'status' == 0 when 'status_reason' = 'icmp' therefore function will never return true



+    $result = dbFetchCell("SELECT `status` from `devices` WHERE device_id = ? and status_reason = 'icmp'", array($parent_id));
+    if ($result) {
+        return true;
+    }

barryodonovan:
Hi,

Unfortunately the query is still incorrect.

It needs to be 
```php
    $result = dbFetchCell("SELECT `device_id` from `devices` WHERE device_id = ? and status_reason = 'icmp'", array($parent_id));
    if ($result) {
        return true;
    }
```

Also there appears to be some issue with the diff generated as a result of changing 207.sql to 208.sql
I had to remove that section of the diff and apply the sql by hand.

On a positive note, recursive parent dependencies seem to work really well, I went to 4 levels on my dev box and only got one notice (once everything settled down after a full poller cycle)

Barry

murrant:
If you don't care about the result, you can just do:
```
"SELECT 1 from `devices` WHERE device_id = ? and status_reason = 'icmp'"
```

aldemira:
Is datatables pr withdrawn? If so, I'll fix the SQL problems and we'll be good to go.

laf:
@aldemira It was, no idea why so will have to stick with what we have for now.

laf:
@aldemira You all done ready for a code review?

aldemira:
Yes please!

doc/Alerting/Hostdependencies.md
```diff
@@ -0,0 +1,7 @@
+source: Alerting/Hostdepencies.md
```
laf: This file won't get generated until you add it mkdocs.yaml. As it's a hidden one (part of another heading) then add it to hidden at the bottom and give it a good title.

html/includes/forms/delete-host-dependency.inc.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Copyright (c) 2017 Aldemir Akpinar <https://github.com/aldemira/>
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+header('Content-type: text/plain');
```
laf: We should definitely switch this to output json (some of the earlier form code I did was plain text and it was wrong / bad so don't base it on that).  See this PR for examples: https://github.com/librenms/librenms/pull/7489/files

html/includes/forms/get-host-dependencies.inc.php
```diff
@@ -0,0 +1,45 @@
...
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+header('Content-type: application/json');
+
+if (is_admin() === false) {
+    die('ERROR: You need to be admin');
+}
+
+syslog(LOG_INFO, 'alo');
```
laf: Is this meant to be here, I'd say not.

html/includes/forms/save-host-dependency.inc.php
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Copyright (c) 2017 Aldemir Akpinar <https://github.com/aldemira/>
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+header('Content-type: text/plain');
```
laf: Same goes here for json output.

html/includes/forms/save-host-dependency.inc.php
```diff
@@ -0,0 +1,47 @@
...
+        echo 'A device cannot depend itself';
+        exit;
+    }
+    $device_arr[] = $dev;
+}
+$clause = implode(',', array_fill(0, count($device_arr), '?'));
+
+// So mysql_string should be something like IN('?','?','?') and device_arr is an array of devices_ids.
+// Mysql query should be like this: UPDATE devices set parent_id = $_POST['parent_id'] WHERE device_id IN ('<int>','<int>','<int>'); 
+if (dbQuery('UPDATE `devices` set parent_id = '.$_POST['parent_id'].' WHERE `device_id` IN('.$clause.')', $device_arr)) {
```
laf: How come you can't use `dbUpdate()` here?

html/includes/modal/delete_host_dependency.inc.php
```diff
@@ -0,0 +1,70 @@
...
+    $.ajax({
+        type: 'POST',
+        url: 'ajax_form.php',
+        data: { type: "delete-host-dependency", device_id: device_id },
+        dataType: "html",
+        success: function(msg) {
+            $("#message").html('<div class="alert alert-info">'+msg+'</div>');
+            $("#confirm-delete").modal('hide');
+            setTimeout(function() {
+               location.reload(1);
```
laf: Rather than reloading it should remove the data from the table.

html/includes/modal/delete_host_dependency.inc.php
```diff
@@ -0,0 +1,70 @@
...
+            }, 1000);
+        },
+        error: function() {
+            $("#message").html('<div class="alert alert-info">The host dependency could not be deleted.</div>');
+            $("#confirm-delete").modal('hide');
+            setTimeout(function() {
```
laf: Same here.

html/includes/modal/edit_host_dependency.inc.php
```diff
@@ -0,0 +1,118 @@
...
+$('#edit-dependency').on('show.bs.modal', function() {
+    var device_id = $("#edit-device_id").val();
+
+    $.ajax({
+        type: 'POST',
+        url: 'ajax_form.php',
+        data: { type: "get-host-dependencies", "device_id": device_id },
+        dataType: "json",
+        success: function(output) {
+            console.log($('#edit-parent_id').val());
```
laf: Left over debug?

html/includes/modal/edit_host_dependency.inc.php
```diff
@@ -0,0 +1,118 @@
...
+            if(msg.indexOf("ERROR:") <= -1) {
+                $("#message").html('<div class="alert alert-info"><button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>'+msg+'</div>');
+                $("#edit-dependency").modal('hide');
+                $('#availableparents')
+                    .find('option')
+                    .remove()
+                    .end()
+                    .append('<option value="0">None</option>')
+                    .val('0');
+               setTimeout(function() {
```
laf: Same again

html/includes/modal/manage_host_dependencies.inc.php
```diff
@@ -0,0 +1,144 @@
...
+        url: 'ajax_form.php',
+        data: { type: "get-host-dependencies", "parent_id": parent_id, "viewtype": "fromparent" },
+        dataType: "json",
+        success: function(output) {
+            $.each(output, function (i, elem) {
+                $('#manalldevices option[value="'+ elem.device_id + '"').prop("selected", true);
+            });
+        },
+        error: function(output) {
+            console.log(output);
```
laf: Left over debug?

html/includes/modal/manage_host_dependencies.inc.php
```diff
@@ -0,0 +1,144 @@
...
+
+    $.ajax({
+        type: 'POST',
+        url: 'ajax_form.php',
+        data: { type: "save-host-dependency", device_ids: device_ids, parent_id: parent_id },
+        dataType: "html",
+        success: function(msg) {
+            $("#message").html('<div class="alert alert-info"><button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>'+msg+'</div>');
+            $("#manage-dependencies").modal('hide');
+            setTimeout(function() {
```
laf: Same again

html/includes/modal/manage_host_dependencies.inc.php
```diff
@@ -0,0 +1,144 @@
...
+               location.reload(1);
+            }, 1000);
+        },
+        error: function() {
+            $("#message").html('<div class="alert alert-info"><button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span><button>The host dependency could not be saved.</div>');
+            $("#manage-dependencies").modal('hide');
+            setTimeout(function() {
```
laf: Same again

includes/alerts.inc.php
```diff
@@ -802,6 +802,10 @@ function RunAlerts()
             $noacc = false;
         }
 
+        if (IsParentDown($alert['device_id'])) {
+            $noiss = true;
```
laf: Do we need to add anything to eventlog here to say alert is skipped?

doc/Alerting/Hostdependencies.md
aldemira: Is mkdocs.yaml in another repo? I couldn't find it.

html/includes/forms/save-host-dependency.inc.php
aldemira: @laf If I remember correctly, 'IN' keyword didn't work correctly with dbUpdate()

doc/Alerting/Hostdependencies.md
laf: Nope, should be in the root of this one.

aldemira:
@laf hopefully I'll push the changes you've requested in a few hours. But I just wanted to comment about having multiple parents for a device; I'm not sure, I see your point of course, but if we have more than one parent, what will be the case for not to alert for downtime. Having any of them down, or both? Also how common is this scenario? Btw, I'm just trying to start a conversation, not oppose you or something. Maybe we can start a thread on the forum.

lowinger42:
Multiple parents are useful if the network is redundant. For example a switch may have dual uplinks, if both go down we should supress alarms. If one or two are up the network works.

Detecting that one switch is down in a redundant setup is done separately.

Nagios & co has this exact functionality.

barryodonovan:
I would agree with Anders that multi parent is useful. Most of our wireless network is built in rings to provide redundancy.


However, initially think that the current implementation is worth pushing, as some dependency is better than none.


If one of my radios has two parents, then the loss of one will not cause it to go down, therefore no alert will be generated anyway.


I currently have spur segments whereby the loss of one link can lead to >100 alerts!, That's overwhelming the usefulness of an alert 😉


Once we have dependency in place, the need or otherwise for multi parent dependency will become clearer.


My 2c.


aldemira:
@lowinger42 yeah you're right. With the current code, there could be the case when one of the parents and the child device go down at the same time, you won't be alerted (for the child) even though there is access to the network behind the parents and everything else. So a potential nuisance there.

html/includes/forms/save-host-dependency.inc.php
murrant: @aldemira check: dbGenPlaceholders()

murrant:
Do we need to use the dependency map to determine the order we query devices?

laf:
That's not a bad idea but could lead to complexity right now, distributed pollers will be an interesting one for that.

laf:
@librenms/reviewers anyone else want to review?

barryodonovan:
I don't know if it's of any benefit (everything I thought I understood about LibreNMS seems to become obsolete in the 6 months since I last put any major time into it) but this all looks reasonable to me.

barryodonovan:
@aldemira Any idea when this might be ready to go? I'd really like to get it into production. Thanks.

laf:
@aldemira We'd love to get this ready to be in the 1.34 release due at the end of the month.

However to do that we need this to be ready by mid November otherwise we don't have enough time for testing in master. Are you able to rebase this and action any other requested changes?

aldemira:
@laf Sure, I got time make the changes. Just let me know what needs to be done!

laf:
So a rebase is needed to fix the files mentioned below this comment.

Some in line comments need reviewing as well, hopefully others will put some feedback in here as well.

aldemira:
Hi there, I think that's all. I don't know why github isn't picking up modifications for alerts.inc.php though.

kkrumm1:
so when I go to Manage Host Dependencies I don't have any options.

image

barryodonovan:
@kkrumm1 It's working for me, fresh pull 10 minutes ago
host-dep

kkrumm1:
Im not sure why its not working i think @aldemira is working on it tho.

Maybe since having all my device in by IP address?

aldemira:
@kkrumm1 My installation is mostly IP addresses as well, I'm going to do a fresh install and see what happens.

kkrumm1:
Its could be me that box I was using is testing boxed with allot git patches... heh

murrant:
@aldemira Have you thought of pulling in data from the links table to assist with selection?

laf:
> so when I go to Manage Host Dependencies I don't have any options.

Schema 216.sql from this hasn't applied at a guess so the query is failing. Check logs/librenms.log to see if it complains of parent_id

laf:
@aldemira I'm working on some changes to the html / json side of things. I'll submit a PR to your branch so you can review.

laf:
Actually a diff will have to do:

https://p.libren.ms/view/raw/81752222

My last comments after this are:

  - Can you link the two hostnames in the /host-dependencies/ page to the actual devices to make it easier for users to click around.
  - Can you convert that same page to full bootgrid with ajax backend. Installs with a lot of devices may experience issues loading all devices in one go.

image

Once those are sorted then I'm happy to merge this which I would love to do this coming week, preferably early Mon/Tue?

laf:
Actually, 1 more thing. The IsParentDown() check is restricted to icmp, this is fine for devices which check both icmp and snmp but if icmp checks are disabled then we should just be checking if the device itself is marked as down.

aldemira:
@laf Today is a bit hectic, but I'll work on your points tomorrow, shouldn't take much time to finish.  Thanks!

aldemira:
@laf I've just committed your diff, thank you, code looks a lot cleaner now :) However for html/includes/modal/manage_host_dependencies.inc.php after this line the diff was a bit tangled, I tried to fix it. 

@@ -129,22 +141,26 @@ $('#manhostdep-save').click('', function(event) {

laf:
> @laf I've just committed your diff, thank you, code looks a lot cleaner now :) However for html/includes/modal/manage_host_dependencies.inc.php after this line the diff was a bit tangled, I tried to fix it.
> 
> @@ -129,22 +141,26 @@ $('#manhostdep-save').click('', function(event) {

Do you need me to take a look at it?

laf:
> @laf I've just committed your diff, thank you, code looks a lot cleaner now :) However for html/includes/modal/manage_host_dependencies.inc.php after this line the diff was a bit tangled, I tried to fix it.
> 
> @@ -129,22 +141,26 @@ $('#manhostdep-save').click('', function(event) {

Do you need me to take a look at it?

aldemira:
@laf I think I got it thanks. One thing left and that is the device down checks.

laf:
Awesome - thanks so much for this :)

kkrumm1:
👍

murrant:
I'm ok with merging this as is, if it continues to get some refinement.

Notes:

- I feel like we should be using select2 or something as the drop down selector can get a little unusable.
- Some of the select dropdowns are not styled in bootstrap style.
- What happens when someone wants multiple parents for one device?
- I feel like there should be some message in the alerts window indicating that this affects alerts, but I could be totally wrong here.

aldemira:
Even though this PR is open for ages, I'm quite happy to not to rush it and add anything you guys like. 
I'll try to handle the dropdowns today. For multiple parents, yes, that is something that need to be implemented sooner rather than later. Shall we bump the merge for another release and I can code it? 
When we move from single parent to multiple parents, we'll need to change the DB column to varchar, maybe it's better to do it right now and and store the data as json or csv.

aldemira:
@laf Is SELECT 1 from `devices` WHERE device_id = ? and status = 0 ok for a down detection? It used to be SELECT 1 from `devices` WHERE device_id = ? and status_reason = 'icmp'  

Sorry I couldn't see the values in the docs. Thanks.

barryodonovan:
@aldemira I'm pretty sure `status = 0` will cause alerts on both ICMP and SNMP down status.

laf:
> @laf Is SELECT 1 from devices WHERE device_id = ? and status = 0 ok for a down detection? It used to be SELECT 1 from devices WHERE device_id = ? and status_reason = 'icmp'

Little more complicated than that. `SELECT 1 from devices WHERE device_id = ? and status_reason = 'icmp'` is fine for most devices but doesn't include devices that have icmp disabled and so only do snmp checks. That info is most likely in the devices_attribs table.

laf:
@aldemira maybe: `SELECT hostname from devices LEFT JOIN devices_attribs ON devices.device_id=devices_attribs.device_id WHERE status = 0 AND (status_reason = 'icmp' OR (devices_attribs.attrib_type='override_icmp_disable' AND devices_attribs.attrib_value=true));`

aldemira:
@laf That's perfect thanks!

aldemira:
Hi everyone, I'm almost finished with multiple parents code but I need some input from you guys. Now with the single parent code, on the manage host dependencies modal, when you chose a parent it would pull up and select the child hosts in the multiple select box below, and with the multiple parents scenario we can do a similar thing, but what happens when I add a second parent and this parent has a different set of children? This doesn't need to a real world scenario but could be human error during the configuration. Or maybe we should configure this a wizard style window, I dunno. My intention for this button/modal was to quickly setup the children for a parent host rather than going through the devices one by one.

murrant:
For selecting children of a device, merge with existing parents.

I wonder if you need pivot table to make this work nicely.

aldemira:
Hey guys, just pushed the multi-parent code. Your inputs are most valuable!

murrant:
@aldemira Did you consider moving the parent child relationships to their own table to make it easier/faster to manage/query?

A quick mock up of the table:
```
CREATE TABLE device_relationships
(
    parent_device_id INT,
    child_device_id INT,
    CONSTRAINT device_children_device_id_child_device_id_pk PRIMARY KEY (parent_device_id, child_device_id),
    CONSTRAINT device_children_devices_device_id_device_id_fk FOREIGN KEY (parent_device_id, child_device_id) REFERENCES devices (device_id, device_id) ON DELETE CASCADE
);
```

aldemira:
@murrant Oh no, you told me about this before and I forgot :(, Sure I'll look into it

aldemira:
After umpteenth messages and commits I think we're ready for testing.

laf:
:)

Will give this a whirl tonight hopefully.

sql-schema/223.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE device_relationships
```
laf: FYI, this needs to be on one line

laf:
Ooh that's slick now:

image

I've checked out the branch, will run it overnight then test alerting dependencies.

html/includes/forms/get-host-dependencies.inc.php
```diff
@@ -0,0 +1,102 @@
...
+                    $rows = $_POST['rowCount'];
+                    $current = $_POST['current'];
+                    $deps_query .= " LIMIT ".$rows * ($current - 1).", ".$rows;
+                }
+            } else {
+                $deps_query .= " ORDER BY a.hostname";
+            }
+
+            if (isset($_POST['format']) && !empty($_POST['searchPhrase'])) {
+                $searchphrase = '%'.$_POST['searchPhrase'].'%';
```
murrant: You should call mres() on any user input like `$_POST['searchPhrase']`

laf: Just use `$vars[$name]` instead of `$_POST[$name]` or `$_GET[$name]` as that runs it through `clean()` which does `mres()`.

laf:
I'll test in more detail tonight but a quick test shows that this still alerts on child nodes even when the parent is marked as down.

aldemira:
@laf thanks, I'll have a look into it too.

includes/alerts.inc.php
```diff
@@ -879,3 +884,28 @@ function ExtTransports($obj)
...
 }//end ExtTransports()
+
+/**
+ * Check if a device's parent is down
+ * Returns true if parent is down
+ * @param int $device Device-ID
+ * @return bool
+ */
+function IsParentDown($device)
+{
```
murrant: This whole function could be one sql query, but I'm fine with it like this.

murrant: Figured out the SQL, also, we need to take disable and ignore into account here I think.
```sql
SELECT 1 from devices as d
  LEFT JOIN devices_attribs as a USING (device_id)
  LEFT JOIN device_relationships as r ON d.device_id=r.parent_device_id
WHERE d.status = 0
  AND d.ignore=0
  AND d.disabled=0
  AND r.child_device_id=?
  AND (d.status_reason='icmp' OR (a.attrib_type='override_icmp_disable' AND a.attrib_value=true))
```

misc/db_schema.yaml
```diff
@@ -441,6 +441,7 @@ devices:
     - { Field: override_sysLocation, Type: tinyint(1), 'Null': true, Extra: '', Default: '0' }
     - { Field: notes, Type: text, 'Null': true, Extra: '' }
     - { Field: port_association_mode, Type: int(11), 'Null': false, Extra: '', Default: '1' }
+    - { Field: parent_id, Type: text, 'Null': true, Extra: ''}
```
murrant: db_schema.yaml needs to be updated to the new schema.

html/index.php
```diff
@@ -160,6 +161,7 @@
     <?php
     }
     ?>
+  <script src="js/select2.min.js">
```
murrant: Missing closing </script> here, breaks other scripts.

includes/alerts.inc.php
aldemira: With this query we still need to check how many parents current device has, and compare the result with count(1) or something. For example, if our child device has 3 parents and only 2 of them are down, we still need to run issueAlert(), don't you think?

murrant: @aldemera Yeah, we could do counts and compare them.   I'll think about it more.

murrant: Yeah, count parents and count down parents.

2 queries is much better than 1+n

laf:
@aldemira Sorry but it's still sending alerts to both devices even though laf2 has a dependency of laf:

image

I'm on ce689eeda4383cfac4128e512d402c7ce24fb945 commit as reference.

laf:
Also, when editing host dependencies and you click save, the page should refresh (or at least the table should).

includes/alerts.inc.php
```diff
@@ -879,3 +884,25 @@ function ExtTransports($obj)
...
+/**
+ * Check if a device's all parent are down
+ * Returns true if all parents are down
+ * @param int $device Device-ID
+ * @return bool
+ */
+function IsParentDown($device)
+{
+    $parent_count = dbFetchCell("SELECT count(`parent_device_id`) from `device_relationships` WHERE `child_device_id` = ?", array($device));
+    if (!$parent_id || $parent_id == null || $parent_id == 0) {
```
laf: Might be this that's causing it. Not checking $parent_count here but $parent_id.

includes/alerts.inc.php
```diff
@@ -879,3 +884,25 @@ function ExtTransports($obj)
...
+        return false;
+    }
+
+
+    $down_parent_count = dbFetchCell("SELECT count(1) from devices as LEFT JOIN devices_attribs as a USING (device_id) LEFT JOIN device_relationships as r ON d.device_id=r.parent_device_id WHERE d.status=0 AND d.ignore=0 AND d.disabled=0 AND r.child_device_id=? AND (d.status_reason='icmp' OR (a.attrib_type='override_icmp_disable' AND a.attrib_value=true))", array($device));
```
laf: This query doesn't work for me.  This does:

```
SELECT count(1) from devices as d LEFT JOIN devices_attribs as a ON d.device_id=a.device_id LEFT JOIN device_relationships as r ON d.device_id=r.parent_device_id WHERE d.status=0 AND d.ignore=0 AND d.disabled=0 AND r.child_device_id=? AND (d.status_reason='icmp' OR (a.attrib_type='override_icmp_disable' AND a.attrib_value=true))
```

laf:
I think this fixes it for me:

https://p.libren.ms/view/raw/b0bbe5db

includes/alerts.inc.php
murrant: You can even skip the second query if parents == 0 ;)

includes/alerts.inc.php
```diff
@@ -893,20 +893,16 @@ function ExtTransports($obj)
  */
 function IsParentDown($device)
 {
-    $parent_id = dbFetchRows("SELECT `parent_device_id` from `device_relationships` WHERE `child_device_id` = ?", array($device));
+    $parent_count = dbFetchCell("SELECT count(`parent_device_id`) from `device_relationships` WHERE `child_device_id` = ?", array($device));
```
murrant: counts should almost always be `COUNT(*)`  instead of a specific column

includes/alerts.inc.php
murrant: `count(1)` ??? this also should be `COUNT(*)`

includes/alerts.inc.php
murrant: Yeah, should be:
```
if (!$parent_count) {
    return false;
}
```
that covers 0 and null btw.

html/includes/forms/save-host-dependency.inc.php
```diff
@@ -0,0 +1,51 @@
...
+        if (!is_numeric($dev)) {
+            $status = array('status' => 1, 'message' => 'Device ID must be an integer!');
+            break;
+        } elseif (in_array($dev, $_POST['parent_ids'])) {
+            $status = array('status' => 1, 'message' => 'A device cannot depend itself');
+            break;
+        }
+        foreach ($_POST['parent_ids'] as $parent) {
+            if (is_numeric($parent) && $parent != 0) {
+                dbInsert(array('parent_device_id' => $parent, 'child_device_id' => $dev), 'device_relationships');
```
laf: This doesn't take into account the removed parents. You might need to delete all parents associated first then re-populate them. You can also use dbBulkInsert to save on the loop.

aldemira: @laf OK, but we'll still need to validate the input before deleting anything, don't we? and that requires a for loop :(

laf: If you build up for the bulk insert you can then delete once you know you have valid data

laf:
I've done an intro video for this: https://www.youtube.com/watch?v=KMAarVS9QQ8

When starting out doing it, I realised that 'host' isn't standardised and we use device more (we still have some use of 'host') so I updated the references to this for the video. Patch is here: https://p.libren.ms/view/raw/c6a16990

Also moved: `git mv html/pages/host-dependencies.inc.php html/pages/device-dependencies.inc.php`

aldemira:
@laf added your patch and the bulkinsert. Great video btw.

murrant:
Do we need to use device dependencies to determine poller order?

laf:
> Do we need to use device dependencies to determine poller order? I could see a situation where we poll a child (especially a distant child) and the parent hasn't been polled yet. Alerts will not behave as expected here.

It would be helpful but can we easily work that out, how many levels down do you go before saying enough. I.e child device has 4 parents, one of those parents has 4 parents and so on. Also if someone accidentally sets the last parent to have a parent which is the first child. 

I had thought that maybe in alerts we do an icmp check against the parents but would that slow it down too much. What about a daemon that ran doing icmp status checks all the time? We may be able to leverage fping for this.

For now I'm not sure this matters, the alerts will be suppressed once the parent host is down. Might just be worth a warning in the doc.

Rosiak:
A note in the docs is sufficient for now I think.
We should aim to have this PR merged soon-ish, it's been alive for 3 months and a-lot of work from @aldemira .

laf:
I might just drop the travis requirements to merge this in. If @aldemira hasn't rebased tomorrow I'll do that tomorrow night.

brettmarl:
we just started using this feature on our deployment - super excited. @aldemira  - is there any way to get the dependencies list back as a property via the /devices/ API?

aldemira:
@brettmarl Sure I'll have a look!

olds463:
@aldemira  anyway to reflect in alert template when parent is down what the child devices of the parent is?

also when clicking "manage host dependencies" maybe adding option to clear child list? When you have a lot of devices its very time consuming to add relationships.

aldemira:
@olds463 Do you mind posting this on https://community.librenms.org ? Let's discuss it in there, thanks.

brettmarl:
Fantastic - thank you guys!

PR #7334, opened 2017-09-13
Support for device credentials and passing to Oxidized via API

eabeukes:
Additional config to enable LibreNMS to add/edit and show username, password & enable-password via API call on /api/v0/oxidized for Oxidized

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7334)

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -1234,7 +1234,7 @@ function list_oxidized()
     $devices = array();
     $device_types = "'".implode("','", $config['oxidized']['ignore_types'])."'";
     $device_os    = "'".implode("','", $config['oxidized']['ignore_os'])."'";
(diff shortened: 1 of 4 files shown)


laf:
Thanks for submitting this.

Aside from IPMI details we've avoided storing sensitive information like this before. I can't personally decide if I'm comfortable in doing so either. Do you not use tacacs to simplify your oxidized config?

@librenms/reviewers thoughts?

eabeukes:
Thanks Neil - I understand the reasons behind not wanting to hold 
passwords etc. I can see about amending the page so its 
default-disabled, and on enabling display a warning about storing 
sensitive information unencrypted etc? This would ensure home/sme users 
without these requirements can use the new page and more "mature" users 
dont need to use it if they have an alterntive solution already in place.

Neil Lathwood wrote:

laf:
I'm actually wondering if we use something like:

https://github.com/defuse/php-encryption

to store sensitive info. Yes if someone compromises the server where the php code is stored then they can decrypt everything. However this would prevent the credentials being sniffed on the wire in MySQL queries and other places at lease.

PR #7335, opened 2017-09-13
feature: LDAP auth update: alerts, api, remember me

murrant:
Add ability to use a bind account if the server does not allow anonymous bind.
If the server does allow anonymous bind, no config change is needed.
Defer ldap connection until it is needed (saves connections from pollers)
Use Config class

FYI, I have no way to test this.

TODO: update/validate docs

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7335)

Fixes: #3069 

#7291, #7237

### DIFF
--- a/doc/Extensions/Authentication.md
+++ b/doc/Extensions/Authentication.md
@@ -137,7 +137,7 @@ $config['auth_ldap_port']   = 389;
 $config['auth_ldap_prefix'] = "uid=";
 $config['auth_ldap_suffix'] = ",ou=People,dc=example,dc=com";
 $config['auth_ldap_group']  = "cn=groupname,ou=groups,dc=example,dc=com";
-$config['auth_ldap_groupbase'] = "ou=group,dc=example,dc=com";
+$config['auth_ldap_groupbase'] = "ou=groups,dc=example,dc=com";
 $config['auth_ldap_groups']['admin']['level'] = 10;
 $config['auth_ldap_groups']['pfy']['level'] = 7;
 $config['auth_ldap_groups']['support']['level'] = 1;
@@ -146,6 +146,14 @@ $config['auth_ldap_uid_attribute'] = 'uidnumber';
 ```
 
 Typically auth_ldap_suffix, auth_ldap_group, auth_ldap_groupbase, auth_ldap_groups are what's required to be configured.
+
+### LDAP bind user (optional)
+If your ldap server does not allow anonymous bind, it is highly suggested to create a bind user, otherwise "remember me", alerting users, and the API will not work.
+```php
+$config['auth_ldap_binduser'] = 'ldapbind'; // will use auth_ldap_prefix and auth_ldap_suffix
+#$config['auth_ldap_binddn'] = 'CN=John.Smith,CN=Users,DC=MyDomain,DC=com'; // overrides binduser
+$config['auth_ldap_bindpassword'] = 'password';
+```
 
 ### LDAP server redundancy
 
--- a/html/includes/authentication/ldap.inc.php
+++ b/html/includes/authentication/ldap.inc.php
@@ -1,46 +1,29 @@
 <?php
 
+use LibreNMS\Config;
 use LibreNMS\Exceptions\AuthenticationException;
 
 function init_auth()
 {
-    global $config, $ldap_connection;
-
-    $ldap_connection = @ldap_connect($config['auth_ldap_server'], $config['auth_ldap_port']);
-
-    if ($config['auth_ldap_starttls'] && ($config['auth_ldap_starttls'] == 'optional' || $config['auth_ldap_starttls'] == 'require')) {
-        $tls = ldap_start_tls($ldap_connection);
-        if ($config['auth_ldap_starttls'] == 'require' && $tls === false) {
-            echo '<h2>Fatal error: LDAP TLS required but not successfully negotiated:'.ldap_error($ldap_connection).'</h2>';
-            exit;
-        }
-    }
+    // deferred
 }
 
 
 function authenticate($username, $password)
 {
-    global $config, $ldap_connection;
-
-    if (!$ldap_connection) {
-        throw new AuthenticationException('Unable to connect to ldap server');
-    }
+    $connection = get_ldap_connection(true);
 
     if ($username) {
-        if ($config['auth_ldap_version']) {
-            ldap_set_option($ldap_connection, LDAP_OPT_PROTOCOL_VERSION, $config['auth_ldap_version']);
-        }
-
-        if ($password && ldap_bind($ldap_connection, $config['auth_ldap_prefix'].$username.$config['auth_ldap_suffix'], $password)) {
-            if (!$config['auth_ldap_group']) {
+        if ($password && ldap_bind($connection, get_full_dn($username), $password)) {
+            if (!Config::has('auth_ldap_group')) {
                 return true;
             } else {
                 $ldap_groups = get_group_list();
                 foreach ($ldap_groups as $ldap_group) {
                     $ldap_comparison = ldap_compare(
-                        $ldap_connection,
+                        $connection,
                         $ldap_group,
-                        $config['auth_ldap_groupmemberattr'],
+                        Config::get('auth_ldap_groupmemberattr', 'memberUid'),
                         get_membername($username)
                     );
                     if ($ldap_comparison === true) {
@@ -54,7 +37,7 @@ function authenticate($username, $password)
             throw new AuthenticationException('A password is required');
         }
 
-        throw new AuthenticationException(ldap_error($ldap_connection));
+        throw new AuthenticationException(ldap_error($connection));
     }
 
     throw new AuthenticationException();
@@ -63,7 +46,16 @@ function authenticate($username, $password)
 
 function reauthenticate($sess_id, $token)
 {
-    return false;
+    $sess_id = clean($sess_id);
+    $token = clean($token);
+
+    list($username, $hash) = explode('|', $token);
+
+    if (!user_exists($username, true)) {
+        throw new AuthenticationException();
+    }
+
+    return check_remember_me($sess_id, $token);
 }
 
 
@@ -92,15 +84,23 @@ function adduser($username, $password, $level, $email = '', $realname = '', $can
 }
 
 
-function user_exists($username)
+function user_exists($username, $throw_exception = false)
 {
-    global $config, $ldap_connection;
+    try {
+        $connection = get_ldap_connection();
 
-    $filter  = '('.$config['auth_ldap_prefix'].$username.')';
-    $search  = ldap_search($ldap_connection, trim($config['auth_ldap_suffix'], ','), $filter);
-    $entries = ldap_get_entries($ldap_connection, $search);
-    if ($entries['count']) {
-        return 1;
+        $filter = '(' . Config::get('auth_ldap_prefix') . $username . ')';
+        $search = ldap_search($connection, trim(Config::get('auth_ldap_suffix'), ','), $filter);
+        $entries = ldap_get_entries($connection, $search);
+        if ($entries['count']) {
+            return 1;
+        }
+    } catch (AuthenticationException $e) {
+        if ($throw_exception) {
+            throw $e;
+        } else {
+            echo $e->getMessage() . PHP_EOL;
+        }
     }
 
     return 0;
@@ -109,21 +109,26 @@ function user_exists($username)
 
 function get_userlevel($username)
 {
-    global $config, $ldap_connection;
-
     $userlevel = 0;
 
-    // Find all defined groups $username is in
-    $filter  = '(&(|(cn='.join(')(cn=', array_keys($config['auth_ldap_groups'])).'))('.$config['auth_ldap_groupmemberattr'].'='.get_membername($username).'))';
-    $search  = ldap_search($ldap_connection, $config['auth_ldap_groupbase'], $filter);
-    $entries = ldap_get_entries($ldap_connection, $search);
+    try {
+        $connection = get_ldap_connection();
+        $groups = Config::get('auth_ldap_groups');
 
-    // Loop the list and find the highest level
-    foreach ($entries as $entry) {
-        $groupname = $entry['cn'][0];
-        if ($config['auth_ldap_groups'][$groupname]['level'] > $userlevel) {
-            $userlevel = $config['auth_ldap_groups'][$groupname]['level'];
+        // Find all defined groups $username is in
+        $filter = '(&(|(cn=' . join(')(cn=', array_keys($groups)) . '))(' . Config::get('auth_ldap_groupmemberattr', 'memberUid') . '=' . get_membername($username) . '))';
+        $search = ldap_search($connection, Config::get('auth_ldap_groupbase'), $filter);
+        $entries = ldap_get_entries($connection, $search);
+
+        // Loop the list and find the highest level
+        foreach ($entries as $entry) {
+            $groupname = $entry['cn'][0];
+            if ($groups[$groupname]['level'] > $userlevel) {
+                $userlevel = $groups[$groupname]['level'];
+            }
         }
+    } catch (AuthenticationException $e) {
+        echo $e->getMessage() . PHP_EOL;
     }
 
     return $userlevel;
@@ -132,14 +137,18 @@ function get_userlevel($username)
 
 function get_userid($username)
 {
-    global $config, $ldap_connection;
+    try {
+        $connection = get_ldap_connection();
 
-    $filter  = '('.$config['auth_ldap_prefix'].$username.')';
-    $search  = ldap_search($ldap_connection, trim($config['auth_ldap_suffix'], ','), $filter);
-    $entries = ldap_get_entries($ldap_connection, $search);
+        $filter = '(' . Config::get('auth_ldap_prefix') . $username . ')';
+        $search = ldap_search($connection, trim(Config::get('auth_ldap_suffix'), ','), $filter);
+        $entries = ldap_get_entries($connection, $search);
 
-    if ($entries['count']) {
-        return $entries[0][$config['auth_ldap_uid_attribute']][0];
+        if ($entries['count']) {
+            return $entries[0][Config::get('auth_ldap_uid_attribute' . 'uidnumber')][0];
+        }
+    } catch (AuthenticationException $e) {
+        echo $e->getMessage() . PHP_EOL;
     }
 
     return -1;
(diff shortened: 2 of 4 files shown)


laf:
https://twitter.com/KingDaveRa/status/908080940911644672

Feedback from one user.

zlesnr:
Any update on this?  We'd like to see this get in, as currently we are unable to issue API keys.

michelbragaguimaraes:
@zlesnr I am facing the same issue. I'd like to configure oxydized and I use active directory to auth.

laf:
@zlesnr have you tested this?

@michelbragaguimaraes You should be using active directory auth then which with a bind account supports the API already.

michelbragaguimaraes:
@laf That was it. I just set the binduser and it worked.
Awesome... Thank Ya!

zlesnr:
@laf We're not in a position to test at this time.

laf:
Ok. Without people testing then I'm not sure when this will get merged in.

janegilv:
I tested this. Our current configuration stopped working. We are not allowing anonymous access to our ldap, perhaps the previous implementation used the user-bound ldap-connection. After setting a bind user (needed to do a small fix to set binduser with full dn), authentication worked. I can also confirm that group levels seems to be working.

I am not sure how things are supposed to work. But:
 - The user list gets populated, but there are duplicate users. We have several $config['auth_ldap_groups']['...'] in our configuration, where users can be part of different groups. Users are listed once pr every group they are a member of.
- scripts/auth_test.php shows empty username, realname, user_id and email. This is working in the master branch (I dont know if this is a false alarm if the script also needs to be updated).
- Pressing "Edit Permissions"/"Edit User" does not lead anywhere. Something happens, and the page is reloaded. I wouldn't expect it to be possible to do something here, but thought I could check if the user information was present.
- Trying to create an API token gives the error message: "ERROR: error with data, please ensure a valid user and token have been specified.". The user list is populated with duplicates as described earlier.

---

Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[html/includes/authentication/ldap.inc.php, line 92 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYOKCBrQwqpK1idyMw:-KvYOKCBrQwqpK1idyMx:bdg1ik7) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L92)):*
> ```PHP
>         $connection = get_ldap_connection();
> 
>         $filter = '(' . Config::get('auth_ldap_prefix') . $username . ')';
> ```

Might be a bit late to change this now, but: More normal naming convention also in line with the group-configuration would be auth_ldap_suffix "auth_ldap_userbase" for auth_ldap_suffix, and "auth_ldap_userattr" for auth_ldap_prefix.

---

*[html/includes/authentication/ldap.inc.php, line 148 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYNmdvSG7ElHQlU12u:-KvYNmdvSG7ElHQlU12v:bopbvm1) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L148)):*
> ```PHP
> 
>         if ($entries['count']) {
>             return $entries[0][Config::get('auth_ldap_uid_attribute' . 'uidnumber')][0];
> ```

Looking at our directory this is "uidNumber".

---

*[html/includes/authentication/ldap.inc.php, line 180 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYOAvbieo6ZCW6Wwmd:-KvYOAvbieo6ZCW6Wwme:b-t36y6d) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L180)):*
> ```PHP
>                 $username = $entry['uid'][0];
>                 $realname = $entry['cn'][0];
>                 $user_id = $entry[Config::get('auth_ldap_uid_attribute' . 'uidnumber')][0];
> ```

Looking at our directory this is uidNumber.

---

*[html/includes/authentication/ldap.inc.php, line 329 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYLm_6OkjdpD3Pbd7X:-KvYLm_6OkjdpD3Pbd7Y:btvsjrb) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L329)):*
> ```PHP
>         if (ldap_bind(
>             $ldap_connection,
>             get_full_dn(Config::get('auth_ldap_binduser')),
> ```

This forces the binduser to be mixed with the users. Suggesting to only use "Config::get('auth_ldap_binduser')", the full DN of the bind user must then be specified in the config file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librenms/librenms/7335#-:-KvYYORRY6VZH1rQhZta:b-lrcnxu)*

murrant:
Fixed duplicates in userlist.  Can you update and pastebin the outptu of auth_test.php  Perhaps soemthing isn't working.

I suspect we have some sort of issue with the user id.

---

Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[html/includes/authentication/ldap.inc.php, line 92 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYOKCBrQwqpK1idyMw:-Kv_8jjxYLIzobCtdzPp:b-5ut2en) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L92)):*
<details><summary><i>Previously, janegilv wrote…</i></summary><blockquote>

Might be a bit late to change this now, but: More normal naming convention also in line with the group-configuration would be auth_ldap_suffix "auth_ldap_userbase" for auth_ldap_suffix, and "auth_ldap_userattr" for auth_ldap_prefix.
</blockquote></details>

This is inherited and would break everyone's ldap auth to change it now. I agree the names are less than optimal, but not worth the pain to change it at this time.

---

*[html/includes/authentication/ldap.inc.php, line 148 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYNmdvSG7ElHQlU12u:-Kv_8z63o7FbTAcs1mea:b-f6sm9n) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L148)):*
<details><summary><i>Previously, janegilv wrote…</i></summary><blockquote>

Looking at our directory this is "uidNumber".
</blockquote></details>

That is what is being used unless you set auth_ldap_uid_attribute

---

*[html/includes/authentication/ldap.inc.php, line 180 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYOAvbieo6ZCW6Wwmd:-Kv_93s9FheETbEiuVI1:b-2f48mx) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L180)):*
> That is what is being used unless you set auth_ldap_uid_attribute

---

*[html/includes/authentication/ldap.inc.php, line 329 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYLm_6OkjdpD3Pbd7X:-Kv_97tlsN88oI-3zdOE:b-npswua) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L329)):*
<details><summary><i>Previously, janegilv wrote…</i></summary><blockquote>

This forces the binduser to be mixed with the users. Suggesting to only use "Config::get('auth_ldap_binduser')", the full DN of the bind user must then be specified in the config file.
</blockquote></details>

the default auth_ldap_prefix and auth_ldap_suffix is '', so if you don't set that, you can specify auth_ldap_binduser as full dn.  I'm not really sure if that is adequate or not.  Again, the only experience I have with ldap is AD.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librenms/librenms/7335#-:-Kv_CKJnFYyOzBJZ0lZu:b-bxkiey)*

janegilv:
Duplicates in userlist is no longer shown.

Output from auth_test.php (with some information sensored):
```
sudo -iu librenms scripts/auth_test.php  -u user1 -d -v
Authentication Method: ldap
Password: 
Authenticate user user1: 
AUTH SUCCESS

User ():
  username => test
  realname => testfirst testlast
  user_id => 
  email => test@test.net
Groups: ...
```
As you are saying, the user_id is missing. Also, this is the information for another user ("test") which the script has been run for earlier. Could the ldap-connection be cached/not closed with the other user logged in? I tried going as far as restarting the server, but the issue still persists.

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[html/includes/authentication/ldap.inc.php, line 148 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYNmdvSG7ElHQlU12u:-KvbtBbF12JH0Fv7_ERg:b-g68eql) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L148)):*
<details><summary><i>Previously, murrant (Tony Murray) wrote…</i></summary><blockquote>

That is what is being used unless you set auth_ldap_uid_attribute
</blockquote></details>

Notice the upper-case "N"

---

*[html/includes/authentication/ldap.inc.php, line 180 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYOAvbieo6ZCW6Wwmd:-KvbtLtMGASZ0xsZE7ss:b-g68eql) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L180)):*
<details><summary><i>Previously, murrant (Tony Murray) wrote…</i></summary><blockquote>

> That is what is being used unless you set auth_ldap_uid_attribute
</blockquote></details>

Notice the upper-case "N"

---

*[html/includes/authentication/ldap.inc.php, line 329 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYLm_6OkjdpD3Pbd7X:-KvbtP-Ga2v73u-ldX5A:b-oa3fsn) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L329)):*
<details><summary><i>Previously, murrant (Tony Murray) wrote…</i></summary><blockquote>

the default auth_ldap_prefix and auth_ldap_suffix is '', so if you don't set that, you can specify auth_ldap_binduser as full dn.  I'm not really sure if that is adequate or not.  Again, the only experience I have with ldap is AD.
</blockquote></details>

So we use auth_ldap_prefix and auth_ldap_suffix to tell librenms where to search for users. E.g. auth_ldap_prefix="uid=", and auth_ldap_suffix="ou=people,dc=domain,dc=net". Now our bind-users "serviceaccounts" are separated. e.g. "cn=librenms,ou=serviceaccount,dc=domain,dc=net". This makes us unable to filter where in the tree, and on which type the user search should happen.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librenms/librenms/7335#-:-KvbrqqKkom569sRhl96:b-6ypu44)*

janegilv:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[html/includes/authentication/ldap.inc.php, line 329 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYLm_6OkjdpD3Pbd7X:-KvbvezrTg1uSQcMli_6:b-c5a75k) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L329)):*
<details><summary><i>Previously, janegilv wrote…</i></summary><blockquote>

So we use auth_ldap_prefix and auth_ldap_suffix to tell librenms where to search for users. E.g. auth_ldap_prefix="uid=", and auth_ldap_suffix="ou=people,dc=domain,dc=net". Now our bind-users "serviceaccounts" are separated. e.g. "cn=librenms,ou=serviceaccount,dc=domain,dc=net". This makes us unable to filter where in the tree, and on which type the user search should happen.
</blockquote></details>

So if this is unclear, what I am proposing is to not use ```get_full_dn(Config::get...)``` in this line, only ```Config::get...```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librenms/librenms/7335)*

murrant:
Did you try setting auth_ldap_uid_attribute?

There is no cache in LibreNMS, perhaps OpenLDAP?  Also, are you sure the info is incorrect?

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[html/includes/authentication/ldap.inc.php, line 148 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYNmdvSG7ElHQlU12u:-KvecMJqEEy9EZMfJQMe:bxdc6r4) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L148)):*
<details><summary><i>Previously, janegilv wrote…</i></summary><blockquote>

Notice the upper-case "N"
</blockquote></details>

Yeah, I noticed the N.  Your ldap implementation is case sensitive for this?  I think it would be ok to change the default to uidNumber, but I just kept what had been there in the past (after looking it has been this way since the fork).

If you set it to uidNumber does that work?

---

*[html/includes/authentication/ldap.inc.php, line 329 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYLm_6OkjdpD3Pbd7X:-Kvee_c92jpHfAUuhxdu:bj830w1) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L329)):*
<details><summary><i>Previously, janegilv wrote…</i></summary><blockquote>

So if this is unclear, what I am proposing is to not use ```get_full_dn(Config::get...)``` in this line, only ```Config::get...```
</blockquote></details>

That is clear, but doesn't seem like a good solution.  Most users just want to put in the username, not the full dn.  Any better ideas on how to allow both?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librenms/librenms/7335#-:-Kvec6AQ4vt08JTHeb7S:b-ob9fvi)*

murrant:
@janegilv I just found a bad type regarding auth_ldap_uid_attribute.  Please test with the new code, thanks.

janegilv:
So I tested the last version, but still the same output (with or without the config option). If I mess around with the username of our test-user, the returned information sometimes changes to the information for a different user. Perhaps there is a search with multiple hits, and the wrong one is selected somewhere?

Is there an easy way of getting debug output for logins?

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[html/includes/authentication/ldap.inc.php, line 329 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYLm_6OkjdpD3Pbd7X:-KvhYZgjPPrNGmDcyp4v:bhyiy9b) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L329)):*
<details><summary><i>Previously, murrant (Tony Murray) wrote…</i></summary><blockquote>

That is clear, but doesn't seem like a good solution.  Most users just want to put in the username, not the full dn.  Any better ideas on how to allow both?
</blockquote></details>

In all other software I have used the full DN is used, I guess this is what users would expect. In AD the tree-structure is probably more defined and homogenous across environments than ldap.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librenms/librenms/7335#-:-KvhXrZUH5vepuakoOj6:b-j23xym)*

janegilv:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[html/includes/authentication/ldap.inc.php, line 148 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYNmdvSG7ElHQlU12u:-Kvh_49yAPSKtnUOffEN:blfm995) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L148)):*
<details><summary><i>Previously, murrant (Tony Murray) wrote…</i></summary><blockquote>

Yeah, I noticed the N.  Your ldap implementation is case sensitive for this?  I think it would be ok to change the default to uidNumber, but I just kept what had been there in the past (after looking it has been this way since the fork).

If you set it to uidNumber does that work?
</blockquote></details>

I dont know if the implementation is case-sensitive or not. Just noticed it differed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librenms/librenms/7335)*

janegilv:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[html/includes/authentication/ldap.inc.php, line 148 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYNmdvSG7ElHQlU12u:-KvhhtZBOi0ObkK_eFx0:bgi30na) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L148)):*
<details><summary><i>Previously, janegilv wrote…</i></summary><blockquote>

I dont know if the implementation is case-sensitive or not. Just noticed it differed.
</blockquote></details>

The problem with the userid seems to be here. If I replace ```Config::get('auth_ldap_uid_attribute' . 'uidnumber')``` with only ```uidnumber```, this returns the uid. The output then becomes:

```
sudo -iu librenms scripts/auth_test.php  -u test -v
Authentication Method: ldap
Password: 
Authenticate user test: 
AUTH SUCCESS

User (1xxx):
Groups: ...
```

uidnumber is the correct one for the right users. Rest of the information about the user disappears.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librenms/librenms/7335)*

janegilv:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[html/includes/authentication/ldap.inc.php, line 148 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYNmdvSG7ElHQlU12u:-Kvhj__5uL69aojVKNpF:b-9ngdbv) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L148)):*
<details><summary><i>Previously, janegilv wrote…</i></summary><blockquote>

The problem with the userid seems to be here. If I replace ```Config::get('auth_ldap_uid_attribute' . 'uidnumber')``` with only ```uidnumber```, this returns the uid. The output then becomes:

```
sudo -iu librenms scripts/auth_test.php  -u test -v
Authentication Method: ldap
Password: 
Authenticate user test: 
AUTH SUCCESS

User (1xxx):
Groups: ...
```

uidnumber is the correct one for the right users. Rest of the information about the user disappears.
</blockquote></details>

Doing the same in all spots where ```Config::get...``` is used to reference an array makes auth_test behave as expected. I dont know PHP well enough to know the proper way of referencing here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librenms/librenms/7335)*

janegilv:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[html/includes/authentication/ldap.inc.php, line 148 at r1](https://reviewable.io:443/reviews/librenms/librenms/7335#-KvYNmdvSG7ElHQlU12u:-Kvhm9BvUtBCNqtfoqus:b-cpbgbp) ([raw file](https://github.com/librenms/librenms/blob/50062bb163802fccf9b99de38305bd87df6ff333/html/includes/authentication/ldap.inc.php#L148)):*
Also uidnumber has to be lowercase. php_get_entries API:
>The structure of the array is as follows. The attribute index is converted to lowercase. (Attributes are case-insensitive for directory servers, but not when used as array indices.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/librenms/librenms/7335)*

murrant:
@janegilv I'm not sure if you've tested the most recent code or not.  Also, lets not use reviewable, it was getting a little confusing.

janegilv:
Sorry about the wall of text up there.

uidNumber has to be all lowercase. All the ldap-attribute are converted to lowercase in the array-keys. Maybe it would make sense to convert the attributes from Configuration::get to lowercase in order to remove confusion when configuring. If users configure with a mixed-case attribute as seen in their ldap, this will not work.

In every other application I have used, the bind-user is specified with a full dn in the configuration. I've had to make manual changes for our current setup to work now. One option would be to rename the configuration option: `auth_ldap_binduser -> auth_ldap_binddn`.

The last commit seems to be working fine with these changes. I can add API-tokens and connect them to users, edit user rights and so on.

murrant:
Ok, thank you that makes sense as to why the code was originally lowercase.

murrant:
@janegilv Ok, one final test?  I think this is looking good.

Other testers would be great too.

janegilv:
Seems to be working well.

PR #7336, opened 2017-09-13
webui: Show how long a device has been down if it is down

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I've only done it on the overview page as I felt showing it in the device list will be confusing as the column is titled uptime.

https://community.librenms.org/t/down-devices-shows-up-time/2073

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7336)

### DIFF
--- a/html/includes/dev-overview-data.inc.php
+++ b/html/includes/dev-overview-data.inc.php
@@ -16,7 +16,13 @@ if ($config['overview_show_sysDescr']) {
 echo '</div>
       <table class="table table-hover table-condensed table-striped">';
 
-$uptime = $device['uptime'];
+$uptime = formatUptime($device['uptime']);
+$uptime_text = 'Uptime';
+if ($device['status'] == 0) {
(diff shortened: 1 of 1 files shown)


html/includes/dev-overview-data.inc.php
```diff
@@ -16,7 +16,13 @@
 echo '</div>
       <table class="table table-hover table-condensed table-striped">';
 
-$uptime = $device['uptime'];
+$uptime = formatUptime($device['uptime']);
```
murrant: Minor, but shouldn't this be in an else?

laf: Doesn't need to be. We set it to what it was before and overwrite it if the device is down using another value.

laf:
bump @librenms/reviewers

kkrumm1:
seems legit laf. 
image

PR #7337, opened 2017-09-14
refactor: rewrite is_valid_port()

murrant:
Convert is_valid_port to Config
Several Config optimizations
Update documentation

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7337)

### DIFF
--- a/LibreNMS/Config.php
+++ b/LibreNMS/Config.php
@@ -37,6 +37,15 @@ class Config
     public static function get($key, $default = null)
     {
         global $config;
+
+        if (isset($config[$key])) {
+            return $config[$key];
+        }
+
+        if (!str_contains($key, '.')) {
+            return $default;
+        }
+
         $keys = explode('.', $key);
 
         $curr = &$config;
@@ -90,9 +99,18 @@ class Config
      */
     public static function getOsSetting($os, $key, $default = null)
     {
-        if ($os) {
-            $os_key = "os.$os.$key";
+        global $config;
 
+        if ($os) {
+            if (isset($config['os'][$os][$key])) {
+                return $config['os'][$os][$key];
+            }
+
+            if (!str_contains($key, '.')) {
+                return self::get($key, $default);
+            }
+
+            $os_key = "os.$os.$key";
             if (self::has($os_key)) {
                 return self::get($os_key);
             }
@@ -113,6 +131,21 @@ class Config
      */
     public static function getCombined($os, $key, $default = array())
     {
+        global $config;
+
+        if (!self::has($key)) {
+            return self::get("os.$os.$key", $default);
+        }
+
+        if (!isset($config['os'][$os][$key])) {
+            if (!str_contains($key, '.')) {
+                return $default;
+            }
+            if (!self::has("os.$os.$key")) {
+                return self::get($key, $default);
+            }
+        }
+
         return array_unique(array_merge(
             (array)self::get($key, $default),
             (array)self::getOsSetting($os, $key, $default)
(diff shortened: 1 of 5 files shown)


doc/Developing/os/Settings.md
```diff
@@ -11,6 +11,28 @@ For example, to set an alternate icon for ios:
...
+### Ignoring Interfaces
+See also: [Global Ignoring Interfaces Config](../../Support/Configuration.md#interfaces-to-be-ignored)
+
+> These settings are merged with the global settings, so you can only undo global ones with good_if
+
+```yaml
+empty_ifdescr: false # allow empty ifDescr
+bad_if: # ifDescr (substring, case insensitive)
+    - lp0
+bad_if_regexp: # ifDescr (regex, case insensitive)
```
laf: Is it not case insensitive depending on the regex? Same for the other two _regexp ones

laf:
Testing now

laf:
After the docs confirmation then this LGTM

doc/Developing/os/Settings.md
murrant: Previous code appended i on the end of the regex, so I left it this way.

peelman:
This, or something akin to it, seems to be affecting our production config.  After tonight's Daily.sh run, we began discovering all interfaces, including ones explicitly ignored in config.php.

Specifically: 

    $config['bad_ifname_regexp'][] = '/^(Ont|ont) \d{1,3}.*/';
    $config['bad_ifname_regexp'][] = '/^Dslport.*/';

So I just went from a few thousand ports to a like 20,000.

LibreNMS/Config.php
```diff
@@ -113,6 +131,21 @@ public static function getOsSetting($os, $key, $default = null)
...
+        global $config;
+
+        if (!self::has($key)) {
+            return self::get("os.$os.$key", $default);
+        }
+
+        if (!isset($config['os'][$os][$key])) {
+            if (!str_contains($key, '.')) {
+                return $default;
+            }
```
network-guy: This is causing issues. It's failing validation for the $key value and not returning it. Commenting out lines 141-143 fixes the issue. Is there a reason it's looking for a period in the string, since str_contains doesn't do regex checking?

network-guy:
@murrant @laf 

Added a comment to the functions.php code. the following code is the culprit for @peelman's comment.

`            if (!str_contains($key, '.')) {
                 return $default;
             }`

Once I commented it out my alerts have normalized.

laf:
Thanks. I've reverted it as it's got at least another place that's breaking it depending on user config

LibreNMS/Config.php
murrant: Yes, if the string doesn't contain a period, we don't need to try to split it apart and walk the config array.

Can you give an example of what setting(s) are set that is causing the issues?

laf: It's anything for the is_port_valid(). I tried `$config['bad_ifname_regexp'][] = '/^Dslport.*/';`

It's because it gets the str_contains bit int he first place. That config above will always match the !isset and so move on to returning the default.

peelman: For me it was anything.  Even the default ignored interfaces were not being ignored.  I had a ton of `ATM` and VMware vSwitch interfaces show up that didn't exist before 1am Sunday.

The two I really cared most about though were: 

    $config['bad_ifname_regexp'][] = '/^(Ont|ont) \d{1,3}.*/';
    $config['bad_ifname_regexp'][] = '/^Dslport.*/';

As we have a lot of Calix equipment and our Libre box started shouting alerts into Slack for ONT ports and DSL ports being down, (as people's equipment was rebooting, etc), polling was taking 15-20 minutes per machine, and the load on the server was in the triple digits.

so yeah...

I've taken us out of the daily master update and moved us to the release cycle.  Libre is becoming too critical for us to stay on the bleeding edge I guess.

PR #7341, opened 2017-09-14
refactor: rfc1628 tidy up

murrant:
reduce the amount of snmp calls

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/discovery/sensors/charge/rfc1628.inc.php
```diff
@@ -3,22 +3,23 @@
...
-
-        discover_sensor($valid['sensor'], 'charge', $device, $current_oid, $index, $type, $descr, 1, 1, 15, 50, null, 101, $current);
-    }
+if (is_numeric($value)) {
+        discover_sensor(
+            $valid['sensor'],
+            'charge',
+            $device,
+            '.1.3.6.1.2.1.33.1.2.4.0',
+            500,
```
laf: This will lose people historical data as the old index was 500 + $index.

includes/discovery/sensors/current/rfc1628.inc.php
```diff
@@ -2,80 +2,109 @@
...
+if (is_numeric($battery_current)) {
+    $oid = '.1.3.6.1.2.1.33.1.2.6.0';
+    $divisor = get_device_divisor($device, $pre_cache['poweralert_serial'], 'current', $oid);
+
+    discover_sensor(
+        $valid['sensor'],
+        'current',
+        $device,
+        $oid,
+        500,
```
laf: Same here.

includes/discovery/sensors/load/rfc1628.inc.php
```diff
@@ -2,6 +2,8 @@
 
 echo 'RFC1628 ';
 
+$data = snmpwalk_group($device, 'upsOutputPercentLoad', 'UPS-MIB');
```
laf: This is overwritten in the foreach below.

includes/discovery/sensors/charge/rfc1628.inc.php
murrant: Nope, the only valid index is 0.  This is not a table, just a single oid.

murrant:
I updated the rfc1628 load discovery, but have not tested it yet. I will do later when I'm on my other computer.

murrant:
Went back and forth between master and this with three different ups models (tripplite, powerwalker, and eaton)

Only two differences:
power: bypass had wrong oid **fixed**
current: the old code was incorrectly adding a couple of sensors even when they didn't exist.  So, the new code fixed one bug :D

laf:
Using snmpsimd, I'm seeing some duplicates:

Powerwalker:

image

image

image

image

Eaton:

image

image

image

I can provide the snmprec files if you need.

murrant:
Crap, I accidentally included my testing changes to the yaml files.  Eatonups and powerwalker don't actually use rfc1628 but they provide data, so I used the to test.

Issue #1011, opened 2015-05-14
Pull in UPS info from Mikrotik


laf:
Please see:

https://groups.google.com/forum/#!topic/librenms-project/8SF0LrTRvNQ

Lonecrowe:
Add the fan speeds and voltages and temps too while you are at it :) I think its under /system/health

Lonecrowe:
Also add 1.3.6.1.2.1.4.24.3.0 please.  It counts how many active routes there are.  Handy for monitoring OSPF.

Lonecrowe:
Mikrotik health voltage fans temps etc  handy here 
mtxrHealth!@#.iso.org.dod.internet.private.enterprises.mikrotik.mikrotikExperimentalModule.mtXRouterOs.mtxrHealth

OID	 .1.3.6.1.4.1.14988.1.1.3

Also very handy is  DHCP LEASE COUNT

mtxrDHCPLeaseCount!@#.iso.org.dod.internet.private.enterprises.mikrotik.mikrotikExperimentalModule.mtXRouterOs.mtxrDHCP.mtxrDHCPLeaseCount

OID	 .1.3.6.1.4.1.14988.1.1.6.1

paulgear:
@Lonecrowe, it would be good to have separate issues created for the features you want to see - then we can track each feature separately. :+1: on those ideas, though. :-)

acidsas:
I also need this feature! I can provide access to few mikrotiks with APC UPSes.

PR #7352, opened 2017-09-16
newdevice: Added support for routeros to pull UPS info

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #1011
Replaces: #7331

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7352)

### DIFF
--- a/includes/definitions/routeros.yaml
+++ b/includes/definitions/routeros.yaml
@@ -4,6 +4,7 @@ type: network
 nobulk: 1
 mib_dir:
     - mikrotik
+rfc1628_compat: 1
 over:
     - { graph: device_bits, text: 'Device Traffic' }
     - { graph: device_processor, text: 'CPU Usage' }


murrant:
Wasn't there a quirk for runtime?

```
    } elseif ($device['os'] == 'routeros') {
        if ($sensor_type == 'runtime') {
            return 60;
        }
```

laf:
I thought you fixed that in your rfc fix?

murrant:
I fixed the ability for that particular piece of code to work.  Before it wouldn't work as the runtime/rfc1628.inc.php did not call get_device_divisor()

laf:
So `get_device_divisor()` already returns 60 for runtime when the OID starts with `.1.3.6.1.2.1.33.1.2.2` which one of the runtimes does, the other is `.1.3.6.1.2.1.33.1.2.3` and that returns divisor 1.

I don't have access to data to test if we still need to return 60 for `.1.3.6.1.2.1.33.1.2.3`

murrant:
Yeah, I thought they were both 60...

PR #7360, opened 2017-09-18
refactor: rewrite is_valid_port()

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Take two.  Fixed the error and added tests.
fix: `return $default;` -> `return self::get($key, $default);`

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7360)

### DIFF
--- a/LibreNMS/Config.php
+++ b/LibreNMS/Config.php
@@ -37,6 +37,15 @@ class Config
     public static function get($key, $default = null)
     {
         global $config;
+
+        if (isset($config[$key])) {
+            return $config[$key];
+        }
+
+        if (!str_contains($key, '.')) {
+            return $default;
+        }
+
         $keys = explode('.', $key);
 
         $curr = &$config;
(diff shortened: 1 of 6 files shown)


laf:
Tested fine for me.

Is it worth adding something like: https://p.libren.ms/view/raw/b62397e4 (`tests/PortsTest.php`)

murrant:
TLDR: yes

Higher level tests are almost universally better.  They test more and the right things.  Also allow for refactoring without re-writing tests.

We can't go all the way to the top level with librenms tests right now, so, we build up from the bottom.

network-guy:
LGTM

laf:
So adding it in or not for now @murrant as the PR itself is good to go.

murrant:
@laf This is good to merge for me.  We can wait to add the higher level tests I talked about separately.

PR #7368, opened 2017-09-19
Added primary key to perf_times to improve DELETE performance on repl…

brybry192:
…icas using ROW based replication.

We're operating with two replicas in our environment and have been running into significant replication lag (~2 hours) during/following the daily purge on librenms.perf_times. We're using row based replication and with about 88k deletes occurring daill, the lack of a primary key is generating a full table scan for each delete event on the replicas when processed. To resolve the lag we added a primary key to the table. A search through the repo didn't turn anything up, but please be aware that this changes the table layout, so if there are any unstable queries that use select * statements against this table that don't map columns carefully, they'll now return more data with the primary key included. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7368)


laf:
thanks for the PR.

We actually don't even make use of `perf_times` right now :( So, it's probably worth doing a `TRUNCATE perf_times;` in the first line then update the table so that it's done as quick as possible.

Can you sign the CLA please.

Also you need to run `./scripts/build-schema.php` to update the schema file we use for validation.

laf:
@brybry192 

Thanks. You still need to:

  - Run `./scripts/build-schema.php` to generate the schema file
  - Squash the commits to one using your account so the CLA stops looking for `Bryant Vinisky`

laf:
@brybry192 Any update on this?

laf:
I've submitted #7493

PR #7377, opened 2017-09-21
Applied correct divisors to voltage for Alpha UPSs

noaheroufus:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7377)

### DIFF
--- a/includes/definitions/discovery/fxm.yaml
+++ b/includes/definitions/discovery/fxm.yaml
@@ -2,6 +2,8 @@ mib: mibs/alpha/Argus-Power-System-MIB
 modules:
     sensors:
         voltage:
+            options:
+              divisor: 10
             data:
                 -
                     oid: upsOutputTable
@@ -32,16 +34,17 @@ modules:
                     oid: upsOutputFrequency
                     num_oid: .1.3.6.1.4.1.7309.6.1.4.2.
                     index: 'upsOutputFrequency.{{ $index }}'
-                    divisor: 10
                     descr: Output
+                    divisor: 10
         current:
+            options:
+              divisor: 10
             data:
                 -
                     oid: upsOutputTable
                     value: upsOutputCurrent
                     num_oid: .1.3.6.1.4.1.7309.6.1.4.4.1.3.
                     index: 'upsOutputCurrent.{{ $index }}'
-                    divisor: 10
                     descr: 'Output #{{ $index }}'
         power:
             data:


includes/definitions/discovery/fxm.yaml
```diff
@@ -32,16 +34,17 @@ modules:
                     oid: upsOutputFrequency
                     num_oid: .1.3.6.1.4.1.7309.6.1.4.2.
                     index: 'upsOutputFrequency.{{ $index }}'
-                    divisor: 10
                     descr: Output
+                    divisor: 10
         current:
+            options:
+              divisor: 10
```
laf: This doesn't change the divisor. You've just moved it to be applied to all currents detected.

includes/definitions/discovery/fxm.yaml
```diff
@@ -32,16 +34,17 @@ modules:
                     oid: upsOutputFrequency
                     num_oid: .1.3.6.1.4.1.7309.6.1.4.2.
                     index: 'upsOutputFrequency.{{ $index }}'
-                    divisor: 10
                     descr: Output
+                    divisor: 10
```
laf: This has just moved a couple of lines, any reason?

includes/definitions/discovery/fxm.yaml
```diff
@@ -2,6 +2,8 @@ mib: mibs/alpha/Argus-Power-System-MIB
 modules:
     sensors:
         voltage:
+            options:
+              divisor: 10
```
laf: So this is the only thing affecting graphs?

includes/definitions/discovery/fxm.yaml
noaheroufus: I just noticed the apc.yaml had the divisors after the descr line.

includes/definitions/discovery/fxm.yaml
noaheroufus: Yes, all three appear to need the divisor.

includes/definitions/discovery/fxm.yaml
laf: The order doesn't matter.

includes/definitions/discovery/fxm.yaml
laf: Ok but the PR states it fixes current, voltage and frequency but I can only see voltage as being the one that is actually changed.

includes/definitions/discovery/fxm.yaml
noaheroufus: Ignore that

includes/definitions/discovery/fxm.yaml
noaheroufus: Changed the title to correspond with *just* voltage being changed.

laf:
Thanks for the fix, quick responses and congrats on your first PR :)

image

Issue #7350, opened 2017-09-16
SAF Tehnika Platform


kmpanilla:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.

Component | Version
--------- | -------
LibreNMS  | 1.31.03-47-g529660b
DB Schema | 206
PHP       | 7.0.22-0ubuntu0.16.04.1
MySQL     | 5.6.36-82.0-log
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3

The current SAF Tehnika OS has its platform identifying non descriptively. I have four different "CFIP Luminas" radios with different "HW" numbers and only the number appears in the drop down for platform.  For instance, I have 3, 11, 13, 14, 15.  See below for sysDescr outputs for these each radios that are matching the SAF Tehnika OS type (each with a different platform type):

 SAF microwave radio;CFIP Lumina FODU v2.70.07;Model:2;HW:15;SN: 369250501839;PC: I11HJT05LB;IDU PCB: I0BMDB05_R07
 SAF microwave radio;CFIP Lumina FODU v1.70.07;Model:2;HW:11;SN: 320370100484;PC: I18GHT03H;IDU PCB: I0BMDB05_R03
 SAF microwave radio;CFIP Lumina FODU v1.70.07;Model:2;HW:13;SN: 321590100555;PC: I24GHT02L;IDU PCB: I0BMDB05_R05
 SAF microwave radio;CFIP Lumina FODU v2.70.07;Model:2;HW:14;SN: 368780100640;PC: I23NHT02HB;IDU PCB: I0BMDB05_R06
 SAF microwave radio;SAF FreeMile v1.70.14;Model:4;HW:3;SN: 325560101070;PC: Z24FEE01H;IDU PCB: Z0BMDB01_02

[...]

PR #7378, opened 2017-09-21
Proper hardware and version identification for SAF

noaheroufus:
Fixes Issue #7350 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7378)

### DIFF
--- a/includes/polling/os/saf-integra-b.inc.php
+++ b/includes/polling/os/saf-integra-b.inc.php
@@ -26,3 +26,5 @@
 $saf_integra = preg_split('/;/', $poll_device['sysDescr']);
 
 $serial = str_replace('S/N: ', '', $saf_integra[8]);
+$version = str_replace('Vers: ', '', $saf_integra[1]);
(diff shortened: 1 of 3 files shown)


murrant:
I know some of the existing code is like this, but this looks extremely fragile.

Any way you can make it nicer with some regex or something?

kmpanilla:
@murrant which lines are looking fragile?  All of it?

murrant:
Say saf adds a semi-colon at some-point?  Then it is all broken.

murrant:
For example of the sysDescr's listed in the ticket here is a regex that would work:

`radio;(CFIP |SAF )?(.*) v([0-9.]*).*SN: ([0-9]+)`  https://regex101.com/r/33ICT5/3

Then the first group is throw away (CFIP and SAF), the second group is the product name, the third group is the version, and the fourth group is the serial number.

Issue #7310, opened 2017-09-09
New Device: IgniteNet FusionSwitch Fiber


MHammett:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.

1.31.03-27-g455c721 - Fri Sep 08 2017 15:22:03 GMT-0500

- [X ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
====================================
Component | Version
--------- | -------
LibreNMS  | 1.31.03-27-g455c721
DB Schema | 206
PHP       | 5.6.30-0+deb8u1
MySQL     | 10.0.32-MariaDB-0+deb8u1
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2.1
====================================

- [X ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
https://p.libren.ms/view/38fde8a0
https://p.libren.ms/view/85586277
http://indy-librenms.midwest-ix.com/ignitenetwalk.txt  (going to the pastebin crashed)
- [ X] Please provide as much detail as possible.
https://www.ignitenet.com/products/fusionswitch-fiber/

- [X ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

laf:
I've submitted a PR for detection. If you have a MIB / can find one we can probably add more info

MHammett:
I just poked IgniteNet. I'll update when I have more.

MHammett:
MIB for switch:  https://www.dropbox.com/s/kw0ig09vgkjyw5z/ES4552BH2.mib?dl=0

MIB (work in progress) for IgniteNet wireless: https://www.dropbox.com/s/eiaixnklv6t25lt/IGNITENET-MIB.mib?dl=0

laf:
[...]

PR #7384, opened 2017-09-24
newdevice: Added more detection for IgniteNet FusionSwitch

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7310 

image

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7384)

### DIFF
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -1097,8 +1097,15 @@ function discovery_process(&$valid, $device, $sensor_type, $pre_cache)
             foreach ($raw_data as $index => $snmp_data) {
                 // get the value for this sensor, check 'value' and 'oid', if state string, translate to a number
                 $data_name = isset($data['value']) ? $data['value'] : $data['oid'];  // fallback to oid if value is not set
-                if (is_numeric($snmp_data[$data_name])) {
-                    $value = $snmp_data[$data_name];
+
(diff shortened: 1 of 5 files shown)


includes/discovery/functions.inc.php
```diff
@@ -1097,8 +1097,15 @@ function discovery_process(&$valid, $device, $sensor_type, $pre_cache)
             foreach ($raw_data as $index => $snmp_data) {
                 // get the value for this sensor, check 'value' and 'oid', if state string, translate to a number
                 $data_name = isset($data['value']) ? $data['value'] : $data['oid'];  // fallback to oid if value is not set
-                if (is_numeric($snmp_data[$data_name])) {
-                    $value = $snmp_data[$data_name];
+
+                $tmp_value = $snmp_data[$data_name];
+                preg_match('/[\d\.\-]+/', $tmp_value, $temp_response);
```
murrant: I'm guessing something returns a string containing a value and text?

I'd prefer the is_numeric() test first so we can avoid the preg_match for most things.

murrant: Also, perhaps this would be a better regex: `-?\d*\.?\d+`

It only allows a single - at the start and a single decimal point, possibly at the start

laf: Thanks, pushing an update now.

laf:
bump

Issue #7319, opened 2017-09-11
Add support for APC In Row RD CRACs


Blorpy:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
`Component | Version
--------- | -------
LibreNMS  | 1.31.03-30-g9cafcb6
DB Schema | 206
PHP       | 5.4.16
MySQL     | 5.5.52-MariaDB
RRDTool   | 1.5.6
SNMP      | NET-SNMP 5.7.2`
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

Looking to have support for various sensors for APCs In Row coolers [http://www.apc.com/shop/ca/en/products/InRow-RD-300mm-Fluid-Cooled-208-230V-60Hz/P-ACRD200](http://www.apc.com/shop/ca/en/products/InRow-RD-300mm-Fluid-Cooled-208-230V-60Hz/P-ACRD200). Right now the device is detected as "APC Management Module", and there isn't any sensors shown, only traffic for the NIC. 

Based on digging in the MIB and using snmpwalk, the following is primarily what seems useful, but of course someone might argue:

.1.3.6.1.4.1.318.1.1.13.4.5.2.1.2 = Cooling output (kW)
.1.3.6.1.4.1.318.1.1.13.4.5.2.1.3 = Cooling demand (kW)
.1.3.6.1.4.1.318.1.1.13.4.5.2.1.5 = Airflow (metric, liter/s)
.1.3.6.1.4.1.318.1.1.13.4.5.2.1.7 = Rack inlet temperature (metric, C)
[...]

PR #7385, opened 2017-09-24
newdevice: Added sensor detection for APC In Row RD devices

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7319 

Adds pressure (kPa) and cooling (W)

image

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7385)


Blorpy:
Maybe the issue is that there should be a decimal point in the return value then. 

For example when 24 is returned, this implies 2.4kW, but is displayed as 24000W, which is 24kW.

PR #7386, opened 2017-09-25
allow displaying special character in description

tomarch:
I need to put some html link in this description field

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7386)

### DIFF
--- a/html/includes/dev-overview-data.inc.php
+++ b/html/includes/dev-overview-data.inc.php
@@ -47,7 +47,7 @@ if (!empty($device['ip'])) {
 if ($device['purpose']) {
     echo '<tr>
         <td>Description</td>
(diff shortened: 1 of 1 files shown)


murrant:
Could you use the procedure field?

tomarch:
No, procedure url is not what i'm looking.
I have one wiki page for all my device, and i need a link to it.

murrant:
Oh yeah, procedures are for alert rules.

I'm very cautious when it comes to displaying user editable data as html, for security reasons.  I'm not sure of how safe this would be.

laf:
Unfortunately this allows for xss:

image

As an example of how to allow certain html code see: `html/includes/common/notes.inc.php`

PR #7387, opened 2017-09-25
Update influx-db dependencies

murrant:
Remove dev-dependencies

Note: looks like we were using the symfony/yaml from this instead of /vendor due to autoload order, but yaml is just a dev dependency here.

I have only tested this on PHP 7.1 because I don't have 5.5 or 5.6 on my dev machine where I have influxdb set up.  I took care to force the php version to 5.5 when updating packages so we don't get too new of packages.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7387)


laf:
Tested on 5.6.31 by running netcat.

lgtm

PR #7392, opened 2017-09-26
Add OpsGenie transport.

celalemre:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7392)


includes/alerts/transport.opsgenie.php
```diff
@@ -0,0 +1,45 @@
...
+curl_setopt($curl, CURLOPT_URL, $url );
+curl_setopt($curl, CURLOPT_CUSTOMREQUEST, "POST"); 
+curl_setopt($curl, CURLOPT_RETURNTRANSFER, 1);
+curl_setopt($curl, CURLOPT_HTTPHEADER, array('Content-Type: application/json'));
+curl_setopt($curl, CURLOPT_POSTFIELDS, json_encode($obj));
+
+$ret = curl_exec($curl);
+$code = curl_getinfo($curl, CURLINFO_HTTP_CODE);
+
+var_dump("Response from OpsGenie: " . $ret); //FIXME: proper debugging
```
laf: Is this left in by accident?

celalemre: Do you mean the FIXME part? If so, no. I left it there, because after inspecting other parts of the code, I saw these kind of FIXMEs next to all the var_dump statement, and thought if a more proper debugging mechanism is introduced in the future, then someone can apply it to my code, too. If you mean the var_dump statement itself not being in the if statement below it, that's also on purpose. We would like to be able to have the response from OpsGenie even if the request is made successfully. But, I can move it into the if statement to have the response from OpsGenie only when an error occurs, if it's more acceptable that way.

laf: I meant the FIXME one. That's going to be output no matter what so should be removed imho. The one in the != 200 if statement is fine though.

celalemre: I see. I'll move it into the if statement below, then. Thank you!

laf:
Ok last thing. You can append your sql change in to an existing file, you need to create a new one like `208.sql`

celalemre:
Sure thing, I'll push a new commit for it.

laf:
image

Thanks for the pull request @celalemre :)

chasgames:
Hi

Is this implemented? I can't get this to work,

After putting in the opsgenie API URL, it says ERROR: "No Config Item"

I tried hard-setting it in the config.php and restarting, but no success.
`$config['alert']['transports']['opsgenie']['url'] = 'https://URL';`

Thanks

celalemre:
Hello @chasgames, is the problem still happening? I'll try to reproduce it by installing from scratch and will get back to you with the results.

celalemre:
@chasgames I installed LibreNMS from scratch in a new environment, configured OpsGenie transport and didn't face any problems. If you're still facing the problem, please reach us via support.

https://www.opsgenie.com/contact-us

murrant:
@chasgames It is best to set the url in the webui instead of the config.php file.

chasgames:
@celalemre @murrant  oddly, even though OPSgenie URL was empty in the WebUI - it did actually connect in the end to the OPSgenie (I just didn't get any alarms for a few hours!) , I can only imagine it was the hardsetting in the config file which did it. I'll take a video clip tomorrow to show you the behaviour in the WebUI in my environment.

chasgames:
Ok - tested today, working fine via WebGUI. I did upgrade to the latest version yesterday, that might have been it. Thanks guys

celalemre:
@chasgames Glad to hear that the problem is solved :)

PR #7418, opened 2017-10-02
feature: issue warning notification if php version is less than 5.6.4

murrant:
rename set_notification function in daily.sh to set_notifiable_result
print output when a daily.sh process fails

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7418)

### DIFF
--- a/daily.php
+++ b/daily.php
@@ -30,6 +30,14 @@ if ($options['f'] === 'update') {
     exit(0);
 }
 
+if ($options['f'] === 'check_php_ver') {
+    $min_version = '5.6.4';
+    if (version_compare(PHP_VERSION, $min_version, '>=')) {
+        exit(0);
+    }
+    exit(1);
+}
+
 if ($options['f'] === 'rrd_purge') {
     if (is_numeric($config['rrd_purge']) && $config['rrd_purge'] > 0) {
         $cmd = "find ".$config['rrd_dir']." -type f -mtime +".$config['rrd_purge']." -print -exec rm -f {} +";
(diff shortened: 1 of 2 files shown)


laf:
I can't help but get annoyed by 'notifiable' for some reason!

So this will introduce the min requirement as 5.6.4. When do you think this should be merged as we need to give people enough notice as we had 15% of the survey users say they run < 5.6 and 25% said we should still support < 5.6.

murrant:
I think I'd like to target the December release for the hard minimum date.  Which means I'd like to merge this this release cycle or the next.  Giving people plenty of time to update their infrastructure.  If we have a hard date, we should update the message here to indicate that.

This does not do anything other than issue a warning.  Once we set a date, we can change this to block updates if the PHP version is not new enough.  At that time, more code needs to be added to roll back and issue a critical notification instead of a warning.

laf:
Ok then.

I don't think this should be the first warning we give people. What about a Community post which we can run social media updates on to warn people of the soft date (this PR to be merged) then the hard date?

murrant:
I opened this to start the discussion.

What does everyone think of these dates:
Soft: Oct. 25
Hard: Nov. 25 or Dec. 30

murrant:
As for notifiable.  Still struggling to find an optimal name for that function :D

laf:
Happy with those timelines aside from Dec 30. We should avoid anything that late - or push it into Jan.

murrant:
Yeah, I was dubious about the end of December, too much other stuff going on.  But I wasn't sure if November was too early.

laf:
Maybe we should say that it won't be any earlier than mid November but we will only merge the blocking of < 5.6 when we need to push an update that can no longer supporting it rather than just block updates based on a set date?

murrant:
I don't think that works.  The number of things being held back by the php version will only grow and we can keep holding them back but I don't think that is a healthy thing.

Things I can think of off the top of my head:

- remove array_column library
- update dependencies to current versions instead of the last version that supports php 5.3
- Laravel
- Some sort of ORM, possibly just Eloquent from Laravel.
- Convert array format from awful php 5.3 syntax to 5.4 syntax.

laf:
Fine by me then. Maybe we go with the cut off in the new year then?

murrant:
Ok, January it is. Probably should be enough before the release so we get testing.

murrant:
Added hard cut off date.  This should be merged mid-November.

murrant:
Merge Mid November.

laf:
Only issue I've found @murrant is that if you have auto updates disabled then this check is skipped.

It would need moving lower down in daily.sh to `post-pull` possibly?

PR #7420, opened 2017-10-03
newdevice: technicolor TG650S and TG670S 

tomarch:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7420)


tests/snmpsim/technicolor_tg650s.snmprec
```diff
@@ -0,0 +1,3 @@
+1.3.6.1.2.1.1.1.0|4|TG650s
```
laf: Is this the full sysDescr and sysObjectID from the devices?

includes/polling/os/technicolor.inc.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+use LibreNMS\RRD\RrdDefinition;
```
laf: You don't need this line.

Please also add a copyright info https://docs.librenms.org/#General/Contributing/#copyright

includes/definitions/technicolor.yaml
```diff
@@ -0,0 +1,13 @@
+os: technicolor
+text: 'technicolor'
```
laf: text: Should describe the device in a little more detail. Like `Cisco IOS`

includes/definitions/technicolor.yaml
```diff
@@ -0,0 +1,13 @@
+os: technicolor
```
laf: The os name should be called after the firmware the device runs like IOS for Cisco IOS. Using the vendor name isn't really the best option, what can be used instead?

includes/polling/os/technicolor.inc.php
tomarch: This was add by the new-os.php script.
Fixed

tests/snmpsim/technicolor_tg650s.snmprec
tomarch: sysDescr is OK
sysObjectID is 1.3.6.1.4.1.2863.405.1

Fixed by new commit

includes/definitions/technicolor.yaml
tomarch: I don't find many information about it.
I've only this:
_COMPANY_NAME=Technicolor
_COMPANY_URL=http://www.technicolor.com
_PROD_NAME=Technicolor TG
_BRAND_NAME=MediaAccess


Maybe we can call it "tg-os" or "mediaaccess-os".

includes/polling/os/technicolor.inc.php
laf: I'll fix that in the script

laf: In fact, it's not. We don't create that file.

includes/definitions/technicolor.yaml
laf: tgos would be ok if the TG devices run the same firmware

html/images/os/technicolor.svg
```diff
@@ -0,0 +1,103 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
laf: Can you compress this svg by using: https://jakearchibald.github.io/svgomg/

includes/definitions/tgos.yaml
```diff
@@ -0,0 +1,13 @@
...
+type: network
+icon: technicolor
+over:
+    - { graph: device_bits, text: 'Device Traffic' }
+    - { graph: device_processor, text: 'CPU Usage' }
+    - { graph: device_mempool, text: 'Memory Usage' }
+discovery:
+    - sysObjectId:
+        - .1.3.6.1.4.1.2863.405.1
+    - sysDescr_regex:
```
laf: You can remove this check. Using the sysObjectId is better and preferred.

tomarch: You speak about the "sysDescr_regex" part, right ?

laf: Correct (and obviously the line below it).

laf:
Thanks again :D

PR #7422, opened 2017-10-03
Add more Procera interfaces

pheinrichs:
I've added more ports related to the procera mibs. Please review and let me know if you think this could be done any better or if you have any suggestions. 

Thanks :)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7422)


includes/definitions/procera.yaml
```diff
@@ -11,3 +11,5 @@ processor_stacked: 1
 discovery:
    - sysObjectId:
        - .1.3.6.1.4.1.15397.2
+mib_dir:
```
murrant: MIB dirs matching the OS name are added by default.

pheinrichs: Well that's convenient. I'll remove it, thanks Murrant.

includes/polling/ports/procera.inc.php
```diff
@@ -0,0 +1,64 @@
...
+* along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*
+* @package    LibreNMS
+* @link       http://librenms.org
+* @copyright  2017 Paul Heinrichs
+* @author     Paul Heinrichs <pdheinrichs@gmail.com>
+*/
+
+$packetlogic_stats = snmpwalk_cache_oid($device, 'netDeviceTable', array(), 'PACKETLOGIC-CHANNEL-MIB');
+$packetlogic_info = snmpwalk_cache_oid($device, 'channelInfoTable', array(), 'PACKETLOGIC-CHANNEL-MIB');
```
murrant: If you put both these in the same array it will make things easier for you below.  Just send the array from the first walk to the second walk instead of array().

snmpwalk_group() is a newer function, but either is fine here.

murrant:
The approach seems reasonable to me.

includes/polling/ports/procera.inc.php
laf: I think I'd like to see the use of `snmpwalk_group()` if possible.

pheinrichs: I attempted the snmpwalk_group(), I don't know if the oddly decimal labeled indexes are causing issues but here's what happened, also created some odd loop that created 6 more interfaces :confused: 

https://pastebin.com/eADLHYij

Did I use it wrong?

murrant: That is odd @pheinrichs can you do that same thing with debug set so I can see the raw snmpwalk output?

pheinrichs: Adding the depth at 2 seems to be kind of working but still giving the array a sub array with an index of [.0] 

https://pastebin.com/0gSBzhwq

murrant: Yeah, the snmp engine isn't returning it right.

It is returning `channelInternalTxAborted[11].0` when it should be `channelInternalTxAborted[11][0]`

Maybe I can update the code to handle this as it might be common when walking tables of single values.

murrant: #7427 should fix this

includes/polling/ports/procera.inc.php
```diff
@@ -0,0 +1,64 @@
...
+    foreach ($channelTypes as $cType) {
+        foreach ($required as $ifEntry => $IfxStat) {
+            $procera_port[$ifEntry] = $packetlogic_stats[$index][0][$cType['type'].$IfxStat];
+        }
+        $procera_port['ifName'] = $packetlogic_stats[$index][0]['channelName']. ' '.$cType['name'];
+        $procera_port['ifDescr'] = $packetlogic_stats[$index][0]['channelName']. ' '.$cType['name'];
+        $procera_port['ifConnectorPresent'] = ($packetlogic_stats[$index][0]['NegotiatedMedia'] != 'linkdown' ? "true" : "false");
+        $procera_port['ifOperStatus'] = ($packetlogic_stats[$index][0]['channelActive'] === 'active' ? "up" : "down");
+        $procera_port['ifType'] = 'ethernetCsmacd';
+        array_push($port_stats, $procera_port);
```
laf: This will add it to the existing array, are these ports in addition to? If so, do we need anything like speed and things to be registered?

includes/polling/ports/procera.inc.php
```diff
@@ -0,0 +1,64 @@
...
+* You should have received a copy of the GNU General Public License
+* along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*
+* @package    LibreNMS
+* @link       http://librenms.org
+* @copyright  2017 Paul Heinrichs
+* @author     Paul Heinrichs <pdheinrichs@gmail.com>
+*/
+
+$packetlogic_stats = snmpwalk_group($device, 'netDeviceTable', 'PACKETLOGIC-CHANNEL-MIB', 2, array());
```
murrant: depth should be 1, so you can omit it and array() in the first call.

includes/polling/ports/procera.inc.php
murrant: channelInternalNegotiatedMedia should allow speed/duplex to be filled in.

But yes, please try to fill in as many fields from ifTable as possible.

murrant:
@pheinrichs You might want to merge or rebase this against master to pull in the snmpwalk_group() changes.

pheinrichs:
@murrant @laf  I've added IfSpeed / duplex and also changed the depth to 1 in the snmpgroup_walk. I also revisited the PACKETLOGIC-CHANNEL-MIB to make sure I didn't miss any extra fields. Let me know what you think! :)

includes/polling/ports/procera.inc.php
```diff
@@ -46,19 +46,33 @@
...
+            $procera_port[$ifEntry] = $packetlogic_stats[$index][$cType['type'].$IfxStat];
         }
-        $procera_port['ifName'] = $packetlogic_stats[$index][0]['channelName']. ' '.$cType['name'];
-        $procera_port['ifDescr'] = $packetlogic_stats[$index][0]['channelName']. ' '.$cType['name'];
-        $procera_port['ifConnectorPresent'] = ($packetlogic_stats[$index][0]['NegotiatedMedia'] != 'linkdown' ? "true" : "false");
-        $procera_port['ifOperStatus'] = ($packetlogic_stats[$index][0]['channelActive'] === 'active' ? "up" : "down");
+        $procera_port['ifName'] = $packetlogic_stats[$index]['channelName']. ' '.$cType['name'];
+        $procera_port['ifDescr'] = $packetlogic_stats[$index]['channelName']. ' '.$cType['name'];
+        $procera_port['ifConnectorPresent'] = ($packetlogic_stats[$index][$cType['type']."NegotiatedMedia"] != '0' ? "true" : "false");
+        $procera_port['ifOperStatus'] = ($packetlogic_stats[$index]['channelActive'] == true ? "up" : "down");
```
murrant: `$packetlogic_stats[$index]['channelActive'] == 1` might be a little more clear.  This oid returns 1 or 0.

includes/polling/ports/procera.inc.php
```diff
@@ -46,19 +46,33 @@
...
-            $procera_port[$ifEntry] = $packetlogic_stats[$index][0][$cType['type'].$IfxStat];
+            $procera_port[$ifEntry] = $packetlogic_stats[$index][$cType['type'].$IfxStat];
         }
-        $procera_port['ifName'] = $packetlogic_stats[$index][0]['channelName']. ' '.$cType['name'];
-        $procera_port['ifDescr'] = $packetlogic_stats[$index][0]['channelName']. ' '.$cType['name'];
-        $procera_port['ifConnectorPresent'] = ($packetlogic_stats[$index][0]['NegotiatedMedia'] != 'linkdown' ? "true" : "false");
-        $procera_port['ifOperStatus'] = ($packetlogic_stats[$index][0]['channelActive'] === 'active' ? "up" : "down");
+        $procera_port['ifName'] = $packetlogic_stats[$index]['channelName']. ' '.$cType['name'];
+        $procera_port['ifDescr'] = $packetlogic_stats[$index]['channelName']. ' '.$cType['name'];
+        $procera_port['ifConnectorPresent'] = ($packetlogic_stats[$index][$cType['type']."NegotiatedMedia"] != '0' ? "true" : "false");
```
murrant: NegotiatedMedia is used 3 times, a temporary variable might be in order.

murrant:
@pheinrichs I'm testing this on our PL7810.

The ports from ifMib are mostly useless and duplicates, do you think we should remove them?  They do contain valid MAC info though...

pheinrichs:
@murrant Them containing macs is what makes me hesitant. I'd vote for leave them?

murrant:
I'm saying you should merge them together.

pheinrichs:
@murrant sorry for the delay. I don't think i understand what you would like done. Is there an example i could look off of.

murrant:
um, I think the airfiber devices do it.  

basically, instead of presenting 16 ports to the user (when there is physically 8), present 8 ports to the user, taking the correct parts of data from if-mib and the procera mibs and combining them.

Basically what you are doing here, but also pull in the if-mib data.

pheinrichs:
I've been comparing the two, and they don't really match up. the IF-MIB generates about 10 ports, but there are only about 3 virtual ports(3 internal/ 3 external) that get discovered from my PR. Is this still the desired way to go?

murrant:
I looked at it and you're right it doesn't line up.

I've thought about this and think the best solution is to add this to the yaml file:
```
bad_if_regexp: '/^pl[0-9]+$/'
```

PR #7426, opened 2017-10-04
bug: fail isPingable if fping errors

f0o:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This will respect the exitcode of fping and fail in situations where hostname is not reachable or if other issues while execution appear.


laf:
Can someone else test this on an install. It stops my poller processes completing:

image

murrant:
I just tested this and didn't see any problems.  Are you seeing that happen consistently laf?

laf:
Checked out twice and it then just builds up old processes. This is my dev env so it heavily relies on snmpsimd which is where the issue could be.

laf:
I've removed most of my devices and this runs ok now.

PR #7427, opened 2017-10-04
fix: snmpwalk_group tables not using entries.

murrant:
When walking an arbitrary oid, it may not be grouped properly if the address ends with .0

For example: channelInternalTxAborted[11].0

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


pheinrichs:
@murrant Looks like this did the trick. :+1:

PR #7437, opened 2017-10-05
fix: issues with recent discovery refactor

murrant:
fix find_port_id() deleting all variables when device_id was given
improve find_port_id() to find ports where only a number is given
fix orphan deletion count
fix insert and update of links with strings where int are required

This should fix map issues, because previously it would never have remote_port_id.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixes: #7439

### DIFF
--- a/includes/discovery/discovery-protocols.inc.php
+++ b/includes/discovery/discovery-protocols.inc.php
@@ -235,7 +235,8 @@ foreach (dbFetchRows($sql, array($device['device_id'])) as $test) {
 }
 
 // remove orphaned links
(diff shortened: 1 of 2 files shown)


kkrumm1:
@kkrumm1 remind to test kk

kkrumm1:
running test.

kkrumm1:
https://pastebin.com/GAqJB4XM

discovery ran faster that time.... but idk how though it was this round.

murrant:
@kkrumm1 this is not a speed improvement, but bugfixes.

To improve speed addHost() will need to be changed.

kkrumm1:
sounds good ... I don't see any obvious bugs than. :)

HomeServ:
After applying this fix my network map was back.

PR #7441, opened 2017-10-06
DEVICE: zyxelnwa, add storage, mempools and wireless

tomarch:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/zyxelnwa.inc.php
```diff
@@ -1,3 +1,6 @@
 <?php
 
 $hardware = $poll_device['sysDescr'];
+$version = explode("ITS", trim(snmp_get($device, '.1.3.6.1.4.1.890.1.15.3.1.6.0', '-Osqv'), '"'), 2);
```
laf: can you not just do `list($version,) = explode....` here?

includes/polling/os/zyxelnwa.inc.php
```diff
@@ -1,3 +1,6 @@
...
+$version = $version[0];
+$serial = trim(snmp_get($device, '.1.3.6.1.4.1.890.1.15.3.1.12.0', '-Oqv'), '"');
```
laf: Do you not have the MIB to use here? You can then do an snmp multi get.

laf:
Thanks @tomarch. the only last thing, sorry to be a pain is to rename the MIBs. The filenames should be that of the actual MIB name to make it easier to find. I.e: `mibs/zyxelnwa/ZYXEL-ES-COMMON.MIB` should be `mibs/zyxelnwa/ZYXEL-ES-COMMON`. The name is from the line: `ZYXEL-ES-COMMON DEFINITIONS ::= BEGIN` (line 13) of the mib.

tomarch:
No problem ;)

laf:
image

Issue #7370, opened 2017-09-20
LLDP of BDCOM S2524D switches is not detected


Pentium100MHz:
Hello, I have a few BDCOM S2524D switches and LibreNMS does not detect their links using LLDP. I have found out that it uses the wrong MIB (LLDP-MIB instead of NMS-LLDP-MIB). Here is how I fixed it in my system, hopefully this will help fix the bug for everyone.

diff discovery-protocols.inc.php /opt/librenms/includes/discovery/discovery-protocols.inc.php 
<pre>
92c92
< if ($device['os'] == 'pbn' && Config::get('autodiscovery.xdp') === true) {
---
> if (($device['os'] == 'pbn' || $device['os'] == 'bdcom')  && Config::get('autodiscovery.xdp') === true) {
94c94
<     $lldp_array  = snmpwalk_cache_oid($device, 'lldpRemoteSystemsData', array(), 'NMS-LLDP-MIB', 'pbn');
---
>     $lldp_array  = snmpwalk_cache_oid($device, 'lldpRemoteSystemsData', array(), 'NMS-LLDP-MIB', $device['os']);</pre>

After the change, links from BDCOM switches are detected correctly.

murrant:
The second line is unnecessary, the OS's directory is always included.

Perhaps you should send a pull request for this change.  https://docs.librenms.org/Developing/Using-Git/

PR #7443, opened 2017-10-06
fix: lldp discovery change local port resolution

murrant:
dot1dBasePortIfIndex seems to skip some ports in certain configurations on Cisco devices, use lldpLocPortId instead.
Add quirk for Calix devices that return different port labels than they actually are.
This needs to be tested.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixes: #7370

### DIFF
--- a/includes/discovery/discovery-protocols.inc.php
+++ b/includes/discovery/discovery-protocols.inc.php
@@ -81,7 +81,7 @@ if (Config::get('autodiscovery.xdp') === true) {
     echo PHP_EOL;
 }//end if
 
-if ($device['os'] == 'pbn' && Config::get('autodiscovery.xdp') === true) {
+if (($device['os'] == 'pbn' || $device['os'] == 'bdcom') && Config::get('autodiscovery.xdp') === true) {
     echo ' NMS-LLDP-MIB: ';
     $lldp_array  = snmpwalk_group($device, 'lldpRemoteSystemsData', 'NMS-LLDP-MIB');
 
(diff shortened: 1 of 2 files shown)


includes/discovery/discovery-protocols.inc.php
```diff
@@ -117,17 +117,14 @@
     echo ' LLDP-MIB: ';
     $lldp_array  = snmpwalk_group($device, 'lldpRemTable', 'LLDP-MIB', 3);
     if (!empty($lldp_array)) {
-        $dot1d_array = snmpwalk_group($device, 'dot1dBasePortIfIndex', 'BRIDGE-MIB');
```
laf: How do you know that this won't break things?

murrant: I don't, but that data doesn't include all ports, so I couldn't use it on the Cisco device I was trying to fix.  This needs testing against several devices.  On my devices, this worked out a lot better.

We could walk both and then we would have more data.

laf: Walking both would be better imho. I can't test this and we may not see many people testing it to be sure

laf:
Does this fix #7370?

kkrumm1:
I can run a test on this but I don't have any cisco or Calix devices... just HP. 
Would the testing still be valid? @murrant

murrant:
@kkrumm1 definitely

kkrumm1:
testing

kkrumm1:
not seeing any issues so far at work.

khobbits:
I've been running this for 3 days without any noticeable issues.
My LLDP discovery is behaving better than it has been, but I know there have been other recent changes.

murrant:
Thanks for the great feedback @khobbits .  I'll see if I can fix that up.

murrant:
@khobbits reported this working on Discord.

Issue #7374, opened 2017-09-21
Clickatell Alert API issue


utelisysadmin:
Seems there is some divergence with regards to clickatell API communication for users registered after November 2016. While the old API continues to work for users registered prior that date, new users have to hit a different endpoint:

**Pre Nov. 2016**:
https://api.clickatell.com/http/sendmsg?user=xxxx&password=xxxx&api_id=xxxx&to=xxxx&text=xxxx

**Post Nov. 2016**:
https://platform.clickatell.com/messages/http/send?apiKey=xxxxxxxxxxxxxxxx==&to=xxxxxxxxxxx&content=Test+message+text

While I can probably re-write the transport.clickatell.php file, I am not sure how community wants to handle this with regards to older installations / users.

DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [ ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ ] Please provide as much detail as possible.
- [ ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

florianbeer:
Maybe just add it to https://docs.librenms.org/Alerting/Transports/#clickatell

Or if the new API works for everyone (old & new users), make a PR to change the file itself.

laf:
@utelisysadmin Please submit this as a pull request and we can discuss it there. It's a lot easier :)

PR #7446, opened 2017-10-07
fix: Fixed clickatell alert transport

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7374 

Not sure how this was supposed to work before, variables that we hadn't documented or were available in the webui were expected. Update based on code from that issue + reference from https://www.clickatell.com/developers/api-documentation/http-api-send-message/


laf:
Small update. Removed from as we don't have support for setting that.

Also updated two config columns to resolve this not working in strict mode.

murrant:
Looks good to merge after fixing db_schema.yaml

I can't test it of course.

laf:
Fixed db_schema

PR #7448, opened 2017-10-09
newdevice: EATON-ATS

tomarch:
add definition, voltage, frequency and state

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Thanks @tomarch 

The sensors look like they could be done from yaml, have you tried that? Need any help in doing that?

includes/definitions/eaton-ats.yaml
```diff
@@ -0,0 +1,14 @@
...
+type: power
+icon: eaton
+mib_dir:
+    - eaton
+over:
+    - { graph: device_current, text: Current }
+    - { graph: device_voltage, text: Voltage }
+discovery:
+    - sysObjectId:
+        - .1.3.6.1.4.1.705.1
```
laf: This will detect devices as eaton-mgeups, you will need to just detect on sysDescr_regex like you've used below and exclude that same regex from the mgeups OS, use `sysDescr_regex_except.

laf: https://travis-ci.org/librenms/librenms/jobs/285547619

tomarch:
I haven't try with yaml. I will take a look. Is there any performance difference between php?

includes/definitions/eaton-ats.yaml
```diff
@@ -0,0 +1,14 @@
...
+text: 'Eaton ATS'
+type: power
+icon: eaton
+mib_dir:
+    - eaton
+over:
+    - { graph: device_current, text: Current }
+    - { graph: device_voltage, text: Voltage }
+discovery:
+    - sysObjectId:
```
laf: You still need to remove this.

laf:
No performance difference but all discovery is being moved to yaml so if it's possible to do it in that (which it appears to) then we won't accept the php side of it.

tomarch:
Ok, i've try with yaml but it can't work. The index return by the device is a string ("source1" and "source2"), so i can't have a numerical OID  : 
NMP[/usr/bin/snmpget -v1 -c COMMUNITY -OUQnt -M /opt/librenms/mibs:/opt/librenms/mibs/eaton udp:HOSTNAME:161 .1.3.6.1.4.1.534.10.2.2.2.1.2.source1 .1.3.6.1.4.1.534.10.2.2.2.1.2.source2 .1.3.6.1.4.1.2682.2.2.3.1.00]
.1.3.6.1.4.1.534.10.2.2.2.1.2.source1: Unknown Object Identifier (Sub-id not found: enterprises -> source1)
.1.3.6.1.4.1.534.10.2.2.2.1.2.source2: Unknown Object Identifier (Sub-id not found: enterprises -> source2)

Unless there's a trick to obtain numerical index with yaml syntax, we need php file.

laf:
It will be easier if you pastebin your yaml, we can then advise.

tomarch:
https://pastebin.com/DAzmf2en

Issue #7452, opened 2017-10-09
shorthost Target Length is hardcoded


jof:
# Boilerplate
Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [ x ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ x ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.

```
==========================================================
Component | Version
--------- | -------
LibreNMS  | 9009633392f219e01173bd26f1c01b7cf55e6b34
DB Schema | 209
PHP       | 7.0.18-0ubuntu0.16.04.1
MySQL     | 10.0.29-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3
==========================================================
```

# Summary
The `shorthost` function defined in `includes/common.php` enables a variable target length of the short hostname. However, very few to no calls actually set different target lengths.
And importantly, the target length is not a setting exposed to end users.

In our environment, we have some names like router1.long-ish-site-name2.OUR.TLD
While we have many sites with different "router1" hosts. This kind of breaks some of the port_bits graphs, as they all show "router1" as the title.
For some of our longer site names, we would like to get these included in the graph title to make things clear.

# Proposal
We define a new configuration parameter in the global $config array which sets a user-desired target `shorthost` length.
Then, inside of `function shorthost()` we check for this desired length being set, as if so, use that as the target length.

laf:
PR open so closing this.

PR #7453, opened 2017-10-09
Create a target shorthost length.

jof:
This is meant to address #7452 
```
jof@spacecuke ~/src/librenms % ./scripts/pre-commit.php
Running unit tests... success
Running style check... success
Running lint check... success
Tests ok, submit away :) 
```

#### Testers

### DIFF
--- a/includes/common.php
+++ b/includes/common.php
@@ -86,6 +86,7 @@ function shorthost($hostname, $len = 12)
     if (filter_var($hostname, FILTER_VALIDATE_IP)) {
         return $hostname;
     }
+    $len = Config::get('shorthost_target_length', $len);
 
     $parts = explode(".", $hostname);
     $shorthost = $parts[0];
--- a/includes/defaults.inc.php
+++ b/includes/defaults.inc.php
@@ -946,5 +946,3 @@ $config['api']['cors']['origin'] = '*';
 $config['api']['cors']['maxage'] = '86400';
 $config['api']['cors']['allowmethods'] = array('POST', 'GET', 'PUT', 'DELETE', 'PATCH');
 $config['api']['cors']['allowheaders'] = array('Origin', 'X-Requested-With', 'Content-Type', 'Accept', 'X-Auth-Token');
-
-// $config['shorthost_target_length'] = 12;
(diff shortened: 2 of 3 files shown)


murrant:
Would just removing  the mydomain setting from the hostname work?

jof:
@murrant: Looking at the `shorthost()` function, it doesn't take into consideration anything to do with mydomain.
It splits up the `hostname` into DNS labels (split by ".") and concatenates the left-most labels until the maximum length is met or exceeded: https://github.com/librenms/librenms/blob/46ee1b377cf262022c311926b879b7649973cc98/includes/common.php#L83-L98

murrant:
I am just trying to avoid adding a new setting that could cause unexpected results.

Many RRD layouts can be broken by having too long of string.

My suggestion was that removing the common part of the domain might be a better way to shorten.  Then to fix you particular issue, you would need to increase the default length.  Then go through the uses and set some back to 12.

includes/common.php
```diff
@@ -86,6 +86,10 @@ function shorthost($hostname, $len = 12)
     if (filter_var($hostname, FILTER_VALIDATE_IP)) {
         return $hostname;
     }
+    global $config;
```
murrant: This can just be:
```
   $len = Config::get('shorthost_target_length', $len);
```

includes/defaults.inc.php
```diff
@@ -947,3 +947,6 @@
 $config['api']['cors']['maxage'] = '86400';
 $config['api']['cors']['allowmethods'] = array('POST', 'GET', 'PUT', 'DELETE', 'PATCH');
 $config['api']['cors']['allowheaders'] = array('Origin', 'X-Requested-With', 'Content-Type', 'Accept', 'X-Auth-Token');
+
+// The target maximum hostname length when applying the shorthost() function.
+$config['shorthost_target_length'] = 12;
```
murrant: Please comment this out.

Did you check if setting it higher messes up any pages/graphs?

jof: I am running this in my environment with strings up to 25 chars long with no problem.
I will comment this out and leave a bit more description

includes/defaults.inc.php
```diff
@@ -947,3 +947,11 @@
 $config['api']['cors']['maxage'] = '86400';
 $config['api']['cors']['allowmethods'] = array('POST', 'GET', 'PUT', 'DELETE', 'PATCH');
 $config['api']['cors']['allowheaders'] = array('Origin', 'X-Requested-With', 'Content-Type', 'Accept', 'X-Auth-Token');
+
+// The target maximum hostname length when applying the shorthost() function.
```
laf: We need docs adding as well. If you can move all of these comments to https://github.com/librenms/librenms/blob/master/doc/Support/Configuration.md#webui-settings somewhere in that section that would be good.

After that, 👍 from me.

laf:
@jof Any further update on this?

jof:
Sorry for the delay, I've just been busy with other stuff.

That's a much better location for documentation. Thanks for the pointer!
I pushed an updated branch.

PR #7456, opened 2017-10-10
fix: snmp_get_multi returns no data if the oid doesn't contain a period

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/snmp.inc.php
+++ b/includes/snmp.inc.php
@@ -202,8 +202,16 @@ function snmp_get_multi($device, $oids, $options = '-OQUs', $mib = null, $mibdir
         $oid               = trim($oid);
         $value             = trim($value, "\" \n\r");
         list($oid, $index) = explode('.', $oid, 2);
-        if (!strstr($value, 'at this OID') && isset($oid) && isset($index)) {
-            $array[$index][$oid] = $value;
+
+        if (!str_contains($value, 'at this OID')) {
+            if (is_null($index)) {
+                if (empty($oid)) {
+                    continue; // no index or oid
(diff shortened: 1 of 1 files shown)


murrant:
The other fix is to simple add:
```
if (is_null($index)) {
    $index = 0;
}
```

erotel:
HI im test and for me work this code:

   

```
if (!str_contains($value, 'at this OID')) { 
           if (is_null($index)) {
                $index = 0;
                if (empty($oid)) {
                    continue; // no index or oid
                }
            }
       $array[$index][$oid] = $value;
}
```

laf:
How does this help when a device is asked for sysUptime.0 and replies with sysUptime = X as in the code we will expect .0 to be there or am I misunderstanding why this patch is proposed?

murrant:
It doesn't work around the device bugs, just makes it possible to.

So, to work around the device bug, something like this is also needed.:
```
if (isset($snmpdata[0])) {
   $poll_device = $snmpdata[0];
} else {
     $poll_device = $snmpdata;
}
```

laf:
But doesn't this still then need custom code in the poller / core module to make use of the returned data that doesn't have the array index of 0?

erotel:
Podle mě není nic víc měnit. Tahle oprava jen přidá k oid index 0,pokud tam není žádný index. Pokud oid index má, nic se nemění.

murrant:
@laf This does NOT fix the issue @erotel is having with his device.  I just fixed a bug that he uncovered.

PR #7460, opened 2017-10-10
Update Callback-Stats-and-Privacy.md

gitwuethrich:
Hi again
1. deleted 1 "total" before BGP (i hope this is what was meant) and switched to "total devices" accordingly, so it isn't once "... total" and then "total ..."
2. bgp -> BGP (doesn't really matter, but abbreviations are usually uppercase)
3. added a "we"... at first it sounded like "you shouldn't ask us" or something like that, with the "we" it's clearer in my opinion

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/General/Callback-Stats-and-Privacy.md
+++ b/doc/General/Callback-Stats-and-Privacy.md
@@ -11,12 +11,12 @@ Now onto the bit you're interested in, what is submitted and what we do with tha
 
 #### What is submitted ####
 - All data is anonymous.
-- Generic statistics are taken from the database, these include things like device total, type and os, port types, speeds and total, total bgp peers. Take a look at the code for full details.
+- Generic statistics are taken from the database, these include things like total devices, type and os, port types, speeds and total BGP peers. Take a look at the code for full details.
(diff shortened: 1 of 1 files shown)


doc/General/Callback-Stats-and-Privacy.md
```diff
@@ -11,12 +11,12 @@ Now onto the bit you're interested in, what is submitted and what we do with tha
 
 #### What is submitted ####
 - All data is anonymous.
-- Generic statistics are taken from the database, these include things like device total, type and os, port types, speeds and total, total bgp peers. Take a look at the code for full details.
```
murrant: Oops lost something here.  This meant port types, port speeds, and port counts **AND** BGP peer count.  Now it doesn't mention port count.

Should replace total with count in my opinion.

gitwuethrich: @murrant Ah I see. I changed it again, would be glad if you could take a look at it (also I hope I didn't tag you too many times, the comment on the other request was a mistake...)

PR #7466, opened 2017-10-10
New os ray

erotel:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Thanks for the PR.

The two sensors need to be moved to yaml (see our docs). If you need help with this then let me know.

erotel:
Move sensor to yaml not work.

laf:
Why?

erotel:
Ray have broken snmp.Not return oid index. return etc. sysuptime =X not sysuptime.0 = X.

laf:
At some stage we won't be using php to do discovery so you will need to raise this to the vendor to get fixed,

murrant:
@erotel laf means the temperature and voltage sensors.

PR #7470, opened 2017-10-10
added poller modules to ceraos.yaml

barryodonovan:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Disabled most poller modules in ceraos.yaml definition file.
None of the disabled modules apply to Ceragon radios and on some firmwares the repeated extra SNMP queries can cause the SNMP daemon to hang requiring a power cycle to restore functionality.
The polling time reduces from >150 seconds in many cases and >300 in some to ~40 seconds.

### DIFF
--- a/includes/definitions/ceraos.yaml
+++ b/includes/definitions/ceraos.yaml
@@ -10,22 +10,71 @@ discovery:
     - sysObjectId:
         - .1.3.6.1.4.1.2281
 poller_modules:
-    applications: 1
+    applications: 0
+    aruba-controller: 0
+    bgp-peers: 0
+    cisco-ipsec-flow-monitor: 0
+    cisco-remote-access-monitor: 0
+    cisco-cef: 0
+    cisco-sla: 0
+    cisco-mac-accounting: 0
+    cipsec-tunnels: 0
+    cisco-ace-loadbalancer : 0
+    cisco-ace-serverfarms: 0
+    cisco-asa-firewall: 0
+    cisco-voice: 0
+    cisco-cbqos: 0
+    cisco-otv: 0
+    cisco-vpdn: 0
+    entity-physical: 0
     hr-mib: 0
     ipSystemStats: 0
     ipmi: 0
+    junose-atm-vp: 0
+    loadbalancers: 0
+    mef: 0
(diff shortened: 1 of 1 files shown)


laf:
You only need to disable or enable the ones not done by default so a lot of these can be removed.

Worth adding the discovery ones in as well?

barryodonovan:
@laf 
I've disabled all those which are globally enabled by default. As a result of doing this I've realised that I had disabled many modules globally when I first built my system but didn't understand the correct approach to take. I am going to go over the other devices I use and i may open a similar PR for modules I am sure are not applicable though in my experience the ceragon are most sensitive to SNMP crashing due to excessive polling.

laf:
@barryodonovan Thanks for the update. However a lot of these are already disabled in defaults so they don't need to be specified here. Especially cisco- based ones.

murrant:
@barryodonovan go to the webui and you can see what is enabled/disabled easily.  Then change only the ones needed.

barryodonovan:
@murrant, @laf,

I'll go over it again, I based the list on the docs rather than the GUI as I had already changed lots of them on my install.

Is there a file which specifies the defaults rather than me having to do a clean install to see them?

Barry

laf:
@barryodonovan Maybe the docs are wrong :( But you can find all defaults in `includes/defaults.inc.php`

barryodonovan:
@laf Thanks,

I'll update the docs too.

includes/definitions/ceraos.yaml
```diff
@@ -10,4 +10,37 @@ discovery:
     - sysObjectId:
         - .1.3.6.1.4.1.2281
 poller_modules:
-    wifi: 1
```
laf: How come wifi has been disabled?

laf:
One question, are these now all disabled as they aren't supported or is this personal choice?

barryodonovan:
There's nothing in WiFi relating to ceragon. Everything is in the MiB except the sensors.

barryodonovan:
None of the disabled modules apply to ceragon, at best they return no such OID.

There are certain things which I'm not using like STP which can apply so I have left them enabled.

Issue #7469, opened 2017-10-10
Poller cannot connect to MySQL via Unix socket


ZedTheYeti:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
Component | Version
--------- | -------
LibreNMS  | 1.32-26-g3381b40
DB Schema | 210
PHP       | 7.0.22-0ubuntu0.16.04.1
MySQL     | 10.0.31-MariaDB-0ubuntu0.16.04.2
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3
- [ ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ ] Please provide as much detail as possible.
- [ ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

I have set up a brand new LibreNMS instance on Ubuntu Server 16.04.3 LTS with MariaDB on the same host using Unix sockets for the two to communicate (with the below config)
```
$config['db_host'] = 'localhost';
$config['db_port'] = '3306';
$config['db_name'] = 'librenms';
$config['db_socket'] = '/var/run/mysqld/mysqld.sock';
```

validate.php connects to the database fine and so does everything else except poller-wrapper.py. The error information given by poller-wrapper.py is simply "ERROR: Could not connect to MySQL database!". However, editing poller-wrapper.py to actually print the exception info reveals the following traceback and more helpful error message "Unknown MySQL server host '/var/run/mysqld/mysqld.sock'".
```
Traceback (most recent call last):
  File "/opt/librenms/poller-wrapper.py", line 207, in <module>
    db = db_open()
  File "/opt/librenms/poller-wrapper.py", line 97, in db_open
    db = MySQLdb.connect(host=db_server, user=db_username, passwd=db_password, db=db_dbname)
[...]

PR #7471, opened 2017-10-10
Supply MySQL database hostname when using Unix sockets to connect

ZedTheYeti:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes issue #7469 

### DIFF
--- a/html/install.php
+++ b/html/install.php
@@ -39,7 +39,7 @@ $config['db_port']=$dbport;
 $config['db_socket']=$dbsocket;
 
 if (!empty($config['db_socket'])) {
-    $config['db_host'] = '';
+    $config['db_host'] = 'localhost';
     $config['db_port'] = null;
 } else {
     $config['db_socket'] = null;
--- a/poller-wrapper.py
+++ b/poller-wrapper.py
@@ -79,12 +79,9 @@ db_username = config['db_user']
 db_password = config['db_pass']
 db_port = int(config['db_port'])
 
-if config['db_host'][:5].lower() == 'unix:':
+if config['db_socket']:
     db_server = config['db_host']
-    db_port = 0
-elif config['db_socket']:
-    db_server = config['db_host']
-    db_port = 0
+    db_socket = config['db_socket']
 else:
     db_server = config['db_host']
 
@@ -93,8 +90,8 @@ db_dbname = config['db_name']
 
 def db_open():
     try:
-        if db_port == 0:
-            db = MySQLdb.connect(host=db_server, user=db_username, passwd=db_password, db=db_dbname)
+        if db_socket:
+            db = MySQLdb.connect(host=db_server, unix_socket=db_socket, user=db_username, passwd=db_password, db=db_dbname)
         else:
             db = MySQLdb.connect(host=db_server, port=db_port, user=db_username, passwd=db_password, db=db_dbname)
         return db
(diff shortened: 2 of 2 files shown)


murrant:
@ZedTheYeti sorry to move conversation here, but I put the first comment in the wrong place. :(

Ok, I see db_socket should really only be set when the socket is in a non-standard location (of course, users won't follow that). 

Another way to express your change would be to remove the `elif config['db_socket']:` section.

I'm fine with merging this for now, as long as it is understood that this ignores db_socket for now.

laf:
Is the better fix not to set db_host as localhost when socket is passed?

ZedTheYeti:
@murrant We need some sort of check as it sets the `db_port` to 0 which the `db_open()` method then checks to determine what parameters to pass to MySQLdb.connect()

@laf Doing so leaves out the case that @murrant mentioned where someone is using a non-standard Unix socket. When using sockets we do need to set `db_host` to localhost and that will work so long as the MySQL socket is in, what MySQL thinks is, the default location for that distro.

laf:
> @laf Doing so leaves out the case that @murrant mentioned where someone is using a non-standard Unix socket. When using sockets we do need to set db_host to localhost and that will work so long as the MySQL socket is in, what MySQL thinks is, the default location for that distro. In instances of non-standard socket location, I believe, we need to set unix_socket to the socket path.

So we set `db_host` to localhost and `db_socket` to the socket provided, then pass `db_socket` to connect? If so that would be good to just update to do that. If you can't then I'll edit the PR to do it.

ZedTheYeti:
@laf I'm happy to make the change I just haven't had the chance yet. I might tonight. The other question is if the config logic that checks if `db_host` starts with `unix:` should be kept.
```
if config['db_host'][:5].lower() == 'unix:':
    db_server = config['db_host']
    db_port = 0
```
So far as I can tell setting the db_host to a Unix socket breaks other parts of LibreNMS so it doesn't seem that it's supported and therefore we don't need to check for it in `poller-wrapper.py` and can just check for the condition that db_socket is defined. Can you confirm that?

murrant:
I've written some other python code that connects to the db.  First I build a set of arguments, then if socket is set, I add that to the arguments, finally, I pass the arguments to the connect function.

https://github.com/murrant/librenms/blob/new-python-service/LibreNMS/__init__.py#L33

ZedTheYeti:
I've added two commits, the first change is to `poller-wrapper.py` and it sets a `db_socket` variable when `config['db_socket']` is set and later passes that variable to the `unix_socket` parameter of `MySQLdb.connect()`. This should hopefully fix cases where a non-standard socket location is used.

The second change is to fix an error that occurs when specifying a Unix socket during install. I noticed that the installer errors out if a Unix socket is given, the error is caused by `install.php` setting `db_host` to an empty string when `db_socket` is set, which then causes an error when the installer tries to lookup network information for the `db_host`. A quick look at the [mysqli docs](http://php.net/manual/en/mysqli.construct.php) says setting the host to `localhost` when specifying a socket is allowed, so that's what `install.php` does now.

html/install.php
```diff
@@ -39,7 +39,7 @@
 $config['db_socket']=$dbsocket;
 
 if (!empty($config['db_socket'])) {
-    $config['db_host'] = '';
```
murrant: This bug is actually in dbConnect in my opinion.

Inserting this in line 60 of dbFacile.php should fix it without breaking anyone.
```
    if (empty($host)) {
        $host = 'localhost';  //needed because of persistent p: below
    }
```

murrant:
Gave this some testing.  Looks good for the most part.

If db_host isn't set, it have an error.  However, the php code will error out if db_host is unset too.  I'm not concerned about that.

PR #7474, opened 2017-10-11
feature: Web validation

murrant:
Allow validations to be run and viewed from the web page.
Pre-check cannot be run on the web, because we won't get a web page if those don't work.
Make validations modular and hopefully easier to manage/write.

Complete, looking for testers to verify.

Note, must be updated for db schema changes because it alters the format of the schema file slightly.

TODO:

- [x] fix certain validations that fail when run under the webui
- [x] clickable links
- [x] Review new console output
- [x] go through and pull out or create fixes text
- [x] clean up code and standardize names

Probably future:
make fixes actionable (like a fix button)
make async instead of running on page load

image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/sql-schema/212.sql
+++ b/sql-schema/212.sql
@@ -1,0 +1,1 @@
+DELETE FROM `pollers` WHERE `poller_name` LIKE '%\n';
--- a/LibreNMS/IRCBot.php
+++ b/LibreNMS/IRCBot.php
@@ -171,7 +171,7 @@ class IRCBot
             if ($this->config['irc_alert']) {
                 $this->alertData();
             }
-            
+
             if ($this->config['irc_conn_timeout']) {
                 $inactive_seconds = time() - $this->last_activity;
                 $max_inactive = $this->config['irc_conn_timeout'];
@@ -677,7 +677,7 @@ class IRCBot
 
     private function _version($params)
     {
-        $versions       = version_info(false);
+        $versions       = version_info();
         $schema_version = $versions['db_schema'];
         $version        = substr($versions['local_sha'], 0, 7);
 
--- a/LibreNMS/Interfaces/ValidationGroup.php
+++ b/LibreNMS/Interfaces/ValidationGroup.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * ValidationGroup.php
+ *
+ * An interface for validation groups, this will be run by Validator.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2017 Tony Murray
+ * @author     Tony Murray <murraytony@gmail.com>
+ */
+
+namespace LibreNMS\Interfaces;
+
+use LibreNMS\Validator;
+
+interface ValidationGroup
+{
+    /**
+     * Validate this module.
+     * To return ValidationResults, call ok, warn, fail, or result methods on the $validator
+     *
+     * @param Validator $validator
+     */
+    public function validate(Validator $validator);
+
+    /**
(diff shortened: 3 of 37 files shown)


laf:
Great idea. Let us know when it's ready for proper testing / code review.

laf:
`PHP Fatal error:  Cannot use LibreNMS\Config as Config because the name is already in use in /opt/librenms/LibreNMS/Validations/Database.php on line 28` - From `./validate.php`

laf:
image

laf:
image


```
-bash-4.2$ ll .pki/
total 4
drwxrw---- 2 librenms librenms 4096 Aug 22  2016 nssdb
```

laf:
Seems like a few bugs for me. I run mysql strict mode on as an fyi.

murrant:
I can't reproduce this:
```
PHP Fatal error: Cannot use LibreNMS\Config as Config because the name is already in use in /opt/librenms/LibreNMS/Validations/Database.php on line 28 - From ./validate.php
```
or the permission one, what are the permissions for the parent directory?

Fixed the output error, made some changes to cli output without checking web after.

laf:
> I can't reproduce this:
> 
> PHP Fatal error: Cannot use LibreNMS\Config as Config because the name is already in use in /opt/librenms/LibreNMS/Validations/Database.php on line 28 - From ./validate.php

Still happens even pulling your latest changes in.

> or the permission one, what are the permissions for the parent directory?

`drwxrw----    3 librenms librenms    4096 Aug 22  2016 .pki`

> Fixed the output error, made some changes to cli output without checking web after.

Better now.

The UI looks cluttered, more an observation at the moment as I don't have a suggestion to improve.

murrant:
Yeah, just working on presenting the data right now.  We could collapse all the groups initially, but we would need to add some extra visual cues that they can be expanded.

laf:
Not a fault of this PR but having just removed and re-installed php I got:

```php
Array ( [0] => 1 [1] => Uncaught exception 'Exception' with message 'DateTime::__construct(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone.' in /opt/librenms/LibreNMS/Validations/Updates.php:51 Stack trace: #0 /opt/librenms/LibreNMS/Validations/Updates.php(51): DateTime->__construct() #1 /opt/librenms/LibreNMS/Validator.php(69): LibreNMS\Validations\Updates->validate(Object(LibreNMS\Validator)) #2 /opt/librenms/html/pages/validate.inc.php(47): LibreNMS\Validator->validate() #3 /opt/librenms/html/index.php(215): require('/opt/librenms/h...') #4 {main} thrown [2] => /opt/librenms/LibreNMS/Validations/Updates.php [3] => 51 )
```

As I hadn't set the timezone. this  leads to a totally blank page.

laf:
Possible suggestion for the display:

https://p.libren.ms/view/raw/c523c7e8

image

murrant:
I made that change plus added a text label.  Honestly, I'm not that focused on the display, because I feel like it will be wrong and we'll need to tweak it anyway.

laf:
The webui change I proposed frames each error a bit better so you can distinguish between where one finishes and the other starts. Would you mind if I / you changed it to use it?

murrant:
@laf go ahead and push whatever you think to my branch.

sql-schema/212.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM `pollers` WHERE `poller_name` LIKE '%\n';
```
laf: I was just about to merge this until I saw this.

I'm not sure of the implications with this so I'm going to take a look at my work install tomorrow.

murrant: This PR makes a change to pollers where poller_name is just the hostname.  The old would have a line return at the end of the hostname.  This now doesn't.  It results in multiple entries for the same poller in the pollers table.  Which isn't really used anywhere except validation and the poller display.  

I wasn't sure how to clean those up, so I just did this since they will be repopulated next poller run.

murrant: perhaps I could edit poller wrapper to ignore the line return when checking if it exists and update that record with the new name.

murrant: @laf I reviewed options and came back to this.
1. Update poller-wrapper.py and librenms-service.py to update the a name if not needed, but that adds unnecessary code in both spots.
2. Do an update to trim the poller_name field.  This could result in duplicate entries in the pollers table.
3. Delete all entries that poler_name ends in a line return, they will be re-added next poller run with the correct poller_name.

3 seems the best to me.

laf: Looking at the where it's used, 3 does seem fine

PR #7476, opened 2017-10-11
New os Alcoma

erotel:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Thanks @erotel. All looks good, just a question on the os name. It's usually not a good idea to base it on the vendor name. Looking at the firmware for this device it doesn't appear to be called anything but to safeguard the vendor building a new product line should we use `alcoma-al`?

erotel:
this device does not support ssh or telnet, so I do not know the system name.
The firmware is the same for both the AL series and the MP series.
the older ALxxMP / ME series (MPS, ME, ME32, MP90) does not support snmp queries wirelessly.
Can I renaming it to "Alcoma_almp"?

laf:
@erotel seems fine with me, `alcoma-almp` though to be consistent.

LibreNMS/OS/AlcomaAlmp.php
```diff
@@ -0,0 +1,97 @@
+<?php
+/**
+ * Alcoma.php
```
laf: For completeness may as well change this

PR #7477, opened 2017-10-11
Fix Moxa cpu polling problem, also add support for EDS-G508E, and pow…

aldemira:
…er graphs

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Could you try moving the sensors to yaml?

It may be possible to specify the `oid:` value as `oid: MOXA-EDSG508E-MIB:: powerConsumption`. Be interested to see if it works.

aldemira:
@laf Sure, I'll try that, but can we use different MIB names for specific sysDescr? The problem with moxa guys is that, they have MIBs for each model, which is more or less the same, but still it needs to be specified.

laf:
> @laf Sure, I'll try that, but can we use different MIB names for specific sysDescr? The problem with moxa guys is that, they have MIBs for each model, which is more or less the same, but still it needs to be specified.

No you can't but you can just put both as sensors so each will be tried.

aldemira:
@laf I've tried with MIB_MODULE_NAME:powerConsumption and snmpbulkwalk successfully retrieves the data, however detection fails on line 1090 in  includes/discovery/functions.inc.php, because discovery_process() expects powerconsumption, not MIB_MODULE:powerconsumption :((
Any suggestions?

laf:
Can you show the yaml you've tried?

aldemira:
modules:
    sensors:
        power:
            data:
                -
                    oid: MOXA-IKS6726A-MIB:powerConsumption
                    num_oid: .1.3.6.1.4.1.8691.7.116.1.10.8.
                    desc: 'Power Consumption'
                    index: 'powerConsumption.{{ $index }}'
        power:
            data:
                -
                    oid: MOXA-EDSG508E-MIB:powerConsumption
                    num_oid: .1.3.6.1.4.1.8691.7.69.1.10.8.
                    desc: 'Power Consumption'
                    index: 'powerConsumption.{{ $index }}'

laf:
Should it not be `MOXA-IKS6726A-MIB;:powerConsumption`?

To fix the issue though you can specify `value: powerConsumption`

aldemira:
@laf Many thanks! Works like a charm!

includes/definitions/discovery/moxa-etherdevice.yaml
```diff
@@ -0,0 +1,16 @@
+modules:
+    sensors:
+        power:
+            data:
+                -
+                    oid: MOXA-IKS6726A-MIB:powerConsumption
```
laf: Shouldn't this be `MOXA-IKS6726A-MIB::powerConsumption` I.e double `::`. Same for the other?

includes/definitions/discovery/moxa-etherdevice.yaml
```diff
@@ -0,0 +1,16 @@
...
+                    num_oid: .1.3.6.1.4.1.8691.7.116.1.10.8.
+                    descr: 'Power Consumption'
+                    index: 'powerConsumption.{{ $index }}'
+                    value: 'powerConsumption'
```
laf: You can drop the quotes from this, so it's just `value: powerConsumption`. Not strictly needed but all others are like that.

includes/polling/mempools/moxa-etherdevice-mem.inc.php
```diff
@@ -20,6 +20,13 @@
...
-$mempool['used'] = snmp_get($device, "usedMemory.0", '-OvQ', 'MOXA-IKS6726A-MIB');
-$mempool['free'] = snmp_get($device, "freeMemory.0", '-OvQ', 'MOXA-IKS6726A-MIB');
+// Moxa people enjoy creating similar MIBs for each model!
+if ($device['sysDescr'] == 'IKS-6726A-2GTXSFP-T') {
+    $mibmod = 'MOXA-IKS6726A-MIB';
+} else if ($device['sysDescr'] == 'EDS-G508E-T') {
+    $mibmod = 'MOXA-EDSG508E-MIB';
+}
+
+$mempool['total'] = snmp_get($device, "totalMemory.0", '-OvQ', $mibmod);
```
laf: Can you convert all these to `snmp_get_multi_oid` to save snmp calls.

includes/discovery/mempools/moxa-etherdevice.inc.php
```diff
@@ -14,8 +14,15 @@
...
-    $total = snmp_get($device, "totalMemory.0", '-OvQ', 'MOXA-IKS6726A-MIB');
-    $avail = snmp_get($device, "freeMemory.0", '-OvQ', 'MOXA-IKS6726A-MIB');
+    // Moxa people enjoy creating similar MIBs for each model!
+    if ($device['sysDescr'] == 'IKS-6726A-2GTXSFP-T') {
+        $mibmod = 'MOXA-IKS6726A-MIB';
+    } else if ($device['sysDescr'] == 'EDS-G508E-T') {
+        $mibmod = 'MOXA-EDSG508E-MIB';
+    }
+
+    $total = snmp_get($device, "totalMemory.0", '-OvQ', $mibmod);
```
laf: Can you convert these two `snmp_get_multi_oid`

includes/definitions/moxa-etherdevice.yaml
```diff
@@ -11,3 +11,4 @@ mib_dir:
 discovery:
     - sysObjectId:
         - .1.3.6.1.4.1.8691.7.116
+        - .1.3.6.1.4.1.8691.7.69
```
laf: You will need to add tests for this as well: https://docs.librenms.org/#Developing/os/Test-Units/

laf:
@aldemira Just the last snmp_get() conversion to snmp_get_multi_oid() needed and then it's good for a merge.

includes/discovery/mempools/moxa-etherdevice.inc.php
```diff
@@ -12,12 +12,20 @@
...
+        $mibmod = 'MOXA-EDSG508E-MIB';
+    }
+
+    $mem_res = snmp_get_multi_oid($device, 'totalMemory.0 freeMemory.0', '-OQUs', $mibmod);
+    $total = $mem_res['totalMemory.0'];
+    $avail = $mem_res['freeMemory.0'];
 
     if ((is_numeric($total)) && (is_numeric($avail))) {
-        discover_mempool($valid_mempool, $device, 0, 'moxa-etherdevice-mem', 'Memory', '1', null, null);
+        discover_mempool($valid_mempool, $device, 0, 'moxa-etherdevice', 'Memory', '1', null, null);
```
laf: The removal of `-mem` will lead to loss in data so you'll need to leave this as it was,

aldemira: Yeah, I didn't want to have a silly file name under polling dir. And that is totally my mistake, I'm sorry.

laf: It's cool, It should be how you changed it to but 'legacy code' :(

PR #7481, opened 2017-10-12
Allow sysname to be used as the hostname in oxidized API call

petracvv:
We are monitoring our network with Librenms and are evaluating oxidized as a possible config management tool. We do not currently have DNS names for our network devices but we have good sysnames in their SNMP configurations. It would be great if we could use those sysnames instead of the hostnames (IPs) when defining regular expressions to use with the librenms oxidized group mapping settings.

This pull request adds a config option `$config['oxidized']['sysname_as_hostname']` that lets the Oxidized API endpoint return the sysname as the hostname for devices when set to `true`.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


petracvv:
Never mind. This does not seem to solve my problem.

PR #7482, opened 2017-10-13
Features: Add state sensors for Palo Alto Networks firewall

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

----------------------------

- Add state sensors for the HA
- Add an alert rule to the collection
- ~~Used PHP and not YAML because I need to convert string to integer~~
- ~~Not sure where to store the MIB file using PHP~~


FTBZ:
Test files are coming.

laf:
As this works then it means the mib is converting the response to a text string. Why not try using a numerical oid in the yaml, I've never actually tried it before but it might work ok.

FTBZ:
Tested without success with some variante of this code:

`        state:
            data:
                -
                    oid: 1.3.6.1.4.1.25461.2.1.2.1.13.0
                    num_oid: .1.3.6.1.4.1.25461.2.1.2.1.13.
                    index: 'panSysHAMode.{{ $index }}'
                    descr: High Availability Mode
                    states:
                        - { descr: Active-Passive, graph: 1, value: 1, generic: 0 }
                        - { descr: Active-Active, graph: 1, value: 2, generic: 0 }`

Idea?

laf:
Try `oid: .1.3.6.1.4.1.25461.2.1.2.1.13`

FTBZ:
No change, this is always the string and not the integer output.

includes/discovery/sensors/state/panos.inc.php
```diff
@@ -0,0 +1,139 @@
+<?php
+
+$temp = snmp_get($device, "panSysHAMode.0", "-Ovqe", "PAN-COMMON-MIB");
+$cur_oid = '.1.3.6.1.4.1.25461.2.1.2.1.13.0';
+$index = '0';
+
+echo $temp;
```
laf: The only thing left I think is to change all these `echo` to `d_echo()`

FTBZ: I can remove, this is especially useful for testing but not for the merge perhaps.

laf: Remove it then if you like :)

FTBZ:
I'm checking if everything is working fine. Wait before merging, don't want to break something more :p

murrant:
@laf if the mib is loaded it will use the textual convention even for numeric oids.

includes/discovery/sensors/state/panos.inc.php
```diff
@@ -0,0 +1,127 @@
...
+    $state_name = 'panSysHAMode';
+    $state_index_id = create_state_index($state_name);
+
+    //Create State Translation
+    if ($state_index_id !== null) {
+        $states = array(
+            array($state_index_id,'Active-Passive',0,1,0) ,
+            array($state_index_id,'Active-Active',0,2,0)
+        );
+        foreach ($states as $value) {
```
murrant: @FTBZ This isn't the most recent way to manage states in php.  If you can't get it to work in yaml, can you update to match the example here: https://docs.librenms.org/#Developing/Sensor-State-Support/

murrant:
@FTBZ I thought the state sensor code always converts strings to int.

includes/discovery/sensors/state/panos.inc.php
```diff
@@ -0,0 +1,127 @@
+<?php
+
+$temp = snmp_get($device, "panSysHAMode.0", "-Ovqe", "PAN-COMMON-MIB");
+$cur_oid = '.1.3.6.1.4.1.25461.2.1.2.1.13.0';
+$index = '0';
+
+#Convert string to integer
+if ($temp == 'active-passive') {
+    $temp = 1;
+} elseif ($temp == 'active-active') {
```
murrant: Did you try lowercase strings in the yaml?

murrant:
@FTBZ I made this work in yaml with two changes:

First, the value gets overwritten by the walk returning multiple items with the same oid but bad values.
```
diff --git a/includes/snmp.inc.php b/includes/snmp.inc.php
index 66ec195a6..ec9a03b34 100644
--- a/includes/snmp.inc.php
+++ b/includes/snmp.inc.php
@@ -378,7 +378,9 @@ function snmpwalk_cache_oid($device, $oid, $array, $mib = null, $mibdir = null,
         $oid               = trim($oid);
         $value             = trim($value, "\" \\\n\r");
         list($oid, $index) = explode('.', $oid, 2);
-        if (!strstr($value, 'at this OID') && isset($oid) && isset($index)) {
+
+        $bad_strings = array('at this OID', 'past the end of the MIB tree');
+        if (!str_contains($value, $bad_strings) && isset($oid) && isset($index)) {
             $array[$index][$oid] = $value;
         }
     }
```

Second, change the descriptions to lowercase:
```
mib: PAN-COMMON-MIB
modules:
    sensors:
        state:
            data:
                -
                    oid: panSysHAMode
                    num_oid: .1.3.6.1.4.1.25461.2.1.2.1.13.
                    index: 'panSysHAMode.{{ $index }}'
                    descr: High Availability Mode
                    states:
                        - { descr: active-passive, graph: 1, value: 1, generic: 0 }
                        - { descr: active-active, graph: 1, value: 2, generic: 0 }
```

FTBZ:
Thanks for the help, someone can double check?

murrant:
@FTBZ just to verify, you want all states for these to show as green?

FTBZ:
@murrant Yes, I think it's better because a device can be active OR passive during the initial configuration. Depending the primary status it's not a bad thing.

murrant:
You could set it to "unknown" which is gray for passive.

FTBZ:
As you wish...

laf:
@FTBZ are you updating this as murrant suggested or leaving as is?

laf:
@FTBZ prod :)

FTBZ:
Ok, changed to unknown/grey for passive device.

screen shot 2017-10-31 at 07 19 00
screen shot 2017-10-31 at 07 18 19

PR #7487, opened 2017-10-14
feature: Improve Device Neighbour WebUI

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

- Creates new device menu called Neighbours, which includes list(a table) and the map
- Cleans up old neighbours UI code


murrant:
Great @Rosiak.

Also might be worthwhile to remove the SQL that requires remote device to be in LibreNMS.

Rosiak:
The current code used in the neighbours file doesn't require the remote device to be in libre, in order to display the relation. That's only the actual map afaik.

murrant:
Like this, there might be some more things we can put under the neighbor tab.
A per-port neighbor entry might not be bad either.

PR #7489, opened 2017-10-14
webui: Add ability to mark all notifications as read

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixes:
https://community.librenms.org/t/ability-to-mark-all-notifications-as-re
ad/2368/1

However it's done in a slightly ugly way with the window.location.reload() as my JS skills are lacking, and it's late. So please give some input :)

### DIFF
--- a/html/pages/notifications.inc.php
+++ b/html/pages/notifications.inc.php
@@ -30,8 +30,19 @@ $notifications = new ObjectCache('notifications');
   <div class="row">
     <div class="col-md-12">
       <h1><a href="/notifications">Notifications</a></h1>
-      <h4><strong class="count-notif"><?php echo $notifications['count']; ?></strong> Unread Notifications <?php echo ($_SESSION['userlevel'] == 10 ? '<button class="btn btn-success pull-right fa fa-plus new-notif" data-toggle="tooltip" data-placement="bottom" title="Create new notification" style="margin-top:-10px;"></button>' : ''); ?><?php echo ($_SESSION['userlevel'] == 10 ? '<button class="btn btn-success pull-right fa fa-eye read-all-notif" data-toggle="tooltip" data-placement="bottom" title="Mark all as Read" style="margin-top:-10px;"></button>' : ''); ?>
-        </h4>
+      <h4>
+<?php
+echo '<strong class="count-notif">' . $notifications['count'] . '</strong> Unread Notifications ';
+
+if (is_admin()) {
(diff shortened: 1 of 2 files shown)


html/pages/notifications.inc.php
```diff
@@ -30,7 +30,8 @@
   <div class="row">
     <div class="col-md-12">
       <h1><a href="/notifications">Notifications</a></h1>
-      <h4><strong class="count-notif"><?php echo $notifications['count']; ?></strong> Unread Notifications <?php echo ($_SESSION['userlevel'] == 10 ? '<button class="btn btn-success pull-right new-notif" style="margin-top:-10px;">New</button>' : ''); ?></h4>
+      <h4><strong class="count-notif"><?php echo $notifications['count']; ?></strong> Unread Notifications <?php echo ($_SESSION['userlevel'] == 10 ? '<button class="btn btn-success pull-right fa fa-plus new-notif" data-toggle="tooltip" data-placement="bottom" title="Create new notification" style="margin-top:-10px;"></button>' : ''); ?><?php echo ($_SESSION['userlevel'] == 10 ? '<button class="btn btn-success pull-right fa fa-eye read-all-notif" data-toggle="tooltip" data-placement="bottom" title="Mark all as Read" style="margin-top:-10px;"></button>' : ''); ?>
```
murrant: Can you use is_admin() instead of checking $_SESSION directly

murrant:
Found an issue, if you keep clicking the button, it keeps adding notifications_attribs entries.

Also, why is this admin only?

laf:
@Rosiak ?

Rosiak:
Seems like this should be possible for every user, I agree on that.
That said I haven't had the time to look further into it - I'll try to see if I have a change in the coming week or the week after.

html/pages/notifications.inc.php
```diff
@@ -232,6 +233,24 @@
...
+  $(document).on( "click", ".read-all-notif", function() {
+    $(this).attr("disabled", true);
+    $.ajax({
+      type: 'POST',
+      url: 'ajax_form.php',
+      data: {type: 'notifications', action: 'read-all-notif'},
+      dataType: "json",
+      success: function (data) {
+        if( data.status == "ok" ) {
+          window.location.reload()
```
murrant: Should be able to hide all unread notifications and update the notification counts in jquery.

html/pages/notifications.inc.php
murrant: Mark All as read should not be displayed on the notifications page.

PR #7490, opened 2017-10-15
feature: Improve APC rPDU2 Sensor Discovery

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixes:
https://community.librenms.org/t/apc-pdu-environment-sensor/2360

Also removed some unused variables in the state code.
In general our APC code needs cleanup, but I won't head up that task right now..


laf:
How come this hasn't been done in yaml, I've only had a quick look and it seems all do-able.

laf:
@Rosiak ?

Rosiak:
```
if ($data['rPDU2SensorTempHumidityStatusType'] != 'notInstalled') 
```
Is possible to do something like this in the yaml discovery?

laf:
Not at present but you can do checks based on the returned value, is that helpful?

includes/discovery/sensors/state/apc.inc.php
```diff
@@ -230,11 +227,74 @@
...
         create_sensor_to_state_index($device, $state_name, $state_name . '.' . $index);
     }
 }
+
+foreach ($pre_cache['r2pdu_humidity_temp_table'] as $index => $data) {
+    if ($data['rPDU2SensorTempHumidityStatusType'] != 'notInstalled') {
+        if ($data['rPDU2SensorTempHumidityStatusTempStatus']) {
+            $cur_oid        = '.1.3.6.1.4.1.318.1.1.26.10.2.2.1.9.' . $index;
+            $state_name     = $data['rPDU2SensorTempHumidityStatusName'] . ' Temperature Probe Status';
+            $state_index_id = create_state_index($state_name);
```
murrant: This is the old style for creating state indexes.  Check here: https://docs.librenms.org/#Developing/Sensor-State-Support/

murrant:
We could do some arbitrary value checking like this:
valid: '{{ $rPDU2SensorTempHumidityStatusType }} != notInstalled`

laf:
> if ($data['rPDU2SensorTempHumidityStatusType'] != 'notInstalled')

Should now be possible. See hwg-ste / hwg-ste2 files for how to use it.

PR #7497, opened 2017-10-16
Refactored authorizers to classes

mcq8:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

The idea of this PR is to refactor all the authentication to classes and have less global methods.
It should now be more clear what to implement if you want to add an authoriser.


mcq8:
I forgot about php 5.3, I'm just to used to shortened array syntax

murrant:
Thanks for working on this.  Didn't have time to give this a full look yet, but here are some initial thoughts.

You can avoid the global variable $authorizer by making a proxy with static methods for the singleton authorizer.  You could also put your factory in that class and use the autoloader instead of sourcing a php file.

Please use Config::get() instead of direct access to $config, we are trying to remove global variables :)

Can move the interface to `LibreNMS\Interfaces\`?

html/pages/edituser.inc.php
```diff
@@ -249,10 +251,10 @@
                     $vars['can_modify_passwd'] = '1';
                 }
 
-                $authorizer->updateUser($vars['user_id'], $vars['new_realname'], $vars['new_level'], $vars['can_modify_passwd'], $vars['new_email']);
+                AuthorizerFactory::get()->updateUser($vars['user_id'], $vars['new_realname'], $vars['new_level'], $vars['can_modify_passwd'], $vars['new_email']);
                 print_message('User has been updated');
-                if (!empty($vars['new_pass1']) && $vars['new_pass1'] == $vars['new_pass2'] && $authorizer->passwordscanchange($vars['cur_username'])) {
-                    if ($authorizer->changepassword($vars['cur_username'], $vars['new_pass1']) == 1) {
+                if (!empty($vars['new_pass1']) && $vars['new_pass1'] == $vars['new_pass2'] && \LibreNMS\Authentication\AuthorizerFactory::get()->passwordscanchange($vars['cur_username'])) {
```
murrant: oops unneeded full namespace here.

LibreNMS/Authentication/AuthorizerFactory.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace LibreNMS\Authentication;
+
+use LibreNMS\Config;
+use LibreNMS\Interfaces\Authentication\Authorizer;
+
+class AuthorizerFactory
+{
+    protected static $_instace;
```
murrant: Probably a typo here `$_instace` -> `$_instance`

murrant:
Looks pretty good, just needs testing.  Too bad we can't use dependency injection eh?

What auth methods did you test?

mcq8:
Still doing the Config.
Only the mysql so far I don't have an AD.

murrant:
Good, I can test AD

LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
```diff
@@ -134,8 +134,8 @@ public function getUserlevel($username)
         // Loop the list and find the highest level
         foreach ($entries as $entry) {
             $groupname = $entry['cn'][0];
-            if (Config::get('auth_ldap_groups')[$groupname]['level'] > $userlevel) {
-                $userlevel = Config::get('auth_ldap_groups')[$groupname]['level'];
+            if (Config::get('auth_ldap_groups.' . $groupname . '.level') > $userlevel) {
```
murrant: You can't do this because some group names have periods in the names :)

mcq8: Seems like there are a couple of more I need to fix, php 5.3 does not like the array acces directly on the return of a function.
https://3v4l.org/hcOHo#output

mcq8: @murrant is it ok to just make a temp var for it right now?
```
            $authLdapGroups = Config::get('auth_ldap_groups');
            if ($authLdapGroups[$groupname]['level'] > $userlevel) {
                $userlevel = $authLdapGroups[$groupname]['level'];
            }
```
Not sure I want to drag new methods into those classes as part of this refactoring.
The cleanup of the ldap and AD duplication can be for an other refactoring.

murrant: Yes, that is the best way to fix for now.

laf:
@mcq8 Sorry needs a rebase now due to another merged PR.

murrant:
@mcq8 we merged an update to ldap authentication.  Are you still planning on updating this?

mcq8:
yes I'll rebase

LibreNMS/Authentication/ADAuthorizationAuthorizer.php
```diff
@@ -0,0 +1,354 @@
...
+    {
+        $userlist = array();
+        $userhash = array();
+
+        $ldap_groups = $this->getGroupList();
+
+        foreach ($ldap_groups as $ldap_group) {
+            $search_filter = "(memberOf=$ldap_group)";
+            if (Config::get('auth_ad_user_filter')) {
+                $search_filter = "(&{Config::get('auth_ad_user_filter')}$search_filter)";
```
murrant: Typo here causes the Config::get to be in the filter instead of the intended output.

murrant:
AD alerting tested.

murrant:
Looks good to me.  Would be good to have ldap and radius tests. But they looked ok to me.

papatango71:
The update to LibreNMS/Authentication/LdapAuthorizationAuthorizer.php  breaks my /overview pages.  I am able to log in and and see everything else except my main overview dashboard.  Below is the error that I got when running /overview/debug=yes

Array ( [0] => 1 [1] => Uncaught Error: Call to undefined method LibreNMS\Authentication\LdapAuthorizer::ldap_compare() in /opt/librenms/LibreNMS/Authentication/LdapAuthorizer.php:152 Stack trace: #0 /opt/librenms/LibreNMS/Authentication/LdapAuthorizer.php(178): LibreNMS\Authentication\LdapAuthorizer->getUserlist() #1 /opt/librenms/html/includes/functions.inc.php(1606): LibreNMS\Authentication\LdapAuthorizer->getUser('4') #2 /opt/librenms/html/pages/front/tiles.php(23): get_dashboards() #3 /opt/librenms/html/index.php(218): require('/opt/librenms/h...') #4 {main} thrown [2] => /opt/librenms/LibreNMS/Authentication/LdapAuthorizer.php [3] => 152 )

PR #7498, opened 2017-10-17
Show only authorized services in availability map

florianbeer:
Up until now, normal users could see all configured services in availability map. This PR checks for permissions first, before displaying the items. I haven't looked at the dashboard widget yet, maybe it needs similar changes.

#### Testers

### DIFF
--- a/html/includes/common/availability-map.inc.php
+++ b/html/includes/common/availability-map.inc.php
@@ -256,7 +256,7 @@ if (defined('SHOW_SETTINGS')) {
     }
 
     if (($mode == 1 || $mode == 2) && ($config['show_services'] != 0)) {
-        if ($_SESSION['userlevel'] >= '5') {
+        if (is_normal_user() === false) {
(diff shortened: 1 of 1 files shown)


html/includes/common/availability-map.inc.php
```diff
@@ -256,8 +256,14 @@
     }
 
     if (($mode == 1 || $mode == 2) && ($config['show_services'] != 0)) {
-        $service_query = 'select `S`.`service_type`, `S`.`service_id`, `S`.`service_desc`, `S`.`service_status`, `D`.`hostname`, `D`.`sysName`, `D`.`device_id`, `D`.`os`, `D`.`icon` from services S, devices D where `S`.`device_id` = `D`.`device_id` ORDER BY '.$serviceOrderBy.';';
-        $services = dbFetchRows($service_query);
+        if ($_SESSION['userlevel'] >= '5') {
```
laf: You can change this to `if (is_normal_user() === false) {` rather than using sessions direct.

florianbeer: Neat, thanks!

Issue #7431, opened 2017-10-04
Old OSPF interfaces in MySQL database not clearing out when device is rediscovered.


netnem:
I've got an alert rule set up in LibreNMS as such:

%ospf_ports.ospfIfIpAddress ~ "10.4."   && %ospf_ports.ospfIfState = "down"

It reports back a Brocade device_id (63) with ospfIfIpAddress of 10.4.0.72 and shows down.  However, this OSPF interface has been completely removed from device_id (63).  There is no OSPF configuration on the device at all for this subnet or the port that the OSPF adjacency used to be on. 

I run a new discovery on the device (./discovery.php -h 63) and the old interface still shows up in the MYSQL database (as well as the alert notifications).  

I can manually delete the row ( DELETE FROM 'librenms','ospf_ports' WHERE 'device_id'='63' and 'ospf_port_id'='10.4.0.72' ) and then run a rediscovery on the device and the OSPF port does not repopulate.

In addition, I often test OSPF configurations on a router that is at my desk, and then deploy the router in a different location for official deployment.  The OSPF interface when the new router is deployed (off-site) will remain as showing as part of the router that was at my desk.

laf:
It could be pretty simple fix but i'd like to get some data to see: Can you post the output of `./poller.php -h HOSTNAME -d -r -f -m ospf`

netnem:
Sure, I'll have to pull a new example using the 144.30.251 address range:

#verified ip address config does not exist on router
device15#show run | inc 144.30.251.70
device15#

#verified in SQL database:
'15', '2737', '144.30.251.70.0', '144.30.251.70', '0', '0.0.0.160', 'broadcast', 'enabled', '1', '1', '5', '10', '40', 
'120', 'down', '0.0.0.0', '0.0.0.0', '347', NULL, 'active', 'multicast', 'false', 'none', ''

./poller.php -h 15 -d -r -f -m ospf 
https://pastebin.com/duhcX8im

laf:
PR submitted @netnem - can you test please.

netnem:
Confirmed working.

==Details==

I was able to remove ospf config, rerun poller.php and the alert disappeared.  I also moved the ospf config from one router to another router, and the alert moved to where it was supposed to be (previously it would stay on the old router.)  I then moved the ospf connection back to the original location and the alert moved where it should be.  Fix is performing as expected.

murrant:
Thanks for testing @netnem !

PR #7503, opened 2017-10-17
fix: Fixed ospf polling not removing stale data

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7431 

Tidied a fair bit up.

Fixed debug output

Fixed inserts to insert all the data we have, ensures mysql strict compliance as those columns don't allow nulls.

Fixed deletions as we were trying to access an array like it was one dimensional when it was multi due to the use of context name.

### DIFF
--- a/includes/polling/ospf.inc.php
+++ b/includes/polling/ospf.inc.php
@@ -98,24 +98,31 @@ foreach ($vrfs_lite_cisco as $vrf_lite) {
     foreach ($ospf_instances_poll as $ospf_instance_id => $ospf_entry) {
         // If the entry doesn't already exist in the prebuilt array, insert into the database and put into the array
         if (empty($ospf_instances_db[$ospf_instance_id][$device['context_name']])) {
-            dbInsert(array('device_id' => $device['device_id'], 'ospf_instance_id' => $ospf_instance_id, 'context_name' => $device['context_name']), 'ospf_instances');
+            $tmp_insert = array(
+                    'device_id' => $device['device_id'],
+                    'ospf_instance_id' => $ospf_instance_id,
+                    'context_name' => $device['context_name'],
+            );
+            foreach ($ospf_oids_db as $oid_db) {
+                $tmp_insert[$oid_db] = $ospf_entry[$oid_db];
+            }
+            dbInsert($tmp_insert, 'ospf_instances');
+            unset($tmp_insert);
             echo '+';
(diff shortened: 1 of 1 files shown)


murrant:
@laf This always updates all entries for me, the data seems to be correct.
```
#### Load poller module ospf ####
Processes: U Areas: UU Ports: UUU Neighbours: UUUUUUUUUUU
```

laf:
Found and fixed - thanks @murrant

includes/polling/ospf.inc.php
```diff
@@ -305,30 +307,33 @@
...
+    d_echo($ospf_nbrs_db);
 
     // Loop array of entries and update
     if (is_array($ospf_nbrs_db)) {
         foreach ($ospf_nbrs_db as $ospf_nbr_id => $ospf_nbr_db) {
+            $ospf_nbr_db = array_shift($ospf_nbr_db);
             if (is_array($ospf_nbrs_poll[$ospf_nbr_id])) {
                 $ospf_nbr_poll = $ospf_nbrs_poll[$ospf_nbr_id];
 
                 $ospf_nbr_poll['port_id'] = @dbFetchCell('SELECT A.`port_id` FROM ipv4_addresses AS A, ospf_nbrs AS I WHERE A.ipv4_address = ? AND I.port_id = A.port_id AND I.device_id = ? AND A.context_name = ?', array($ospf_nbr_poll['ospfNbrIpAddr'], $device['device_id'], $device['context_name']));
```
murrant: Change this to:
```
$ospf_nbr_poll['port_id'] = (int)dbFetchCell('SELECT ...
```

If there is no port_id returned, this sql query returns "0" and `"0" != NULL`

includes/polling/ospf.inc.php
```diff
@@ -305,30 +307,33 @@
...
-    }
+    d_echo("\nPolled: ");
+    d_echo($ospf_nbrs_poll);
+    d_echo('Database: ');
+    d_echo($ospf_nbrs_db);
 
     // Loop array of entries and update
     if (is_array($ospf_nbrs_db)) {
         foreach ($ospf_nbrs_db as $ospf_nbr_id => $ospf_nbr_db) {
+            $ospf_nbr_db = array_shift($ospf_nbr_db);
```
murrant: This fixes the update issues, but I'm unsure if it breaks users with contexts.  If we are only polling one context at a time, I assume it will be just fine.

laf: We poll one context at a time.

murrant:
Another pre-existing bug neighbor ids all have `.0` appended to them.  Breaks a few things.

PR #7510, opened 2017-10-18
fix: Errors in the documentations for CentOS 7

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

-------------

- Removed composer from the packages (can't work like this due to dependencies)
- Added the correct XML package
- Added the directories creation (was removed from the documentation ?)
- Separated steps for SELinux to avoid copy/past errors

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,10 +22,8 @@ html/plugins/*
 !html/plugins/Test/
 includes/custom/*
 junk
-logs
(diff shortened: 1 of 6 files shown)


FTBZ:
⚠️ The fping section is missing too, why isn't this include in the default configuration file?

PR #7522, opened 2017-10-20
Added temperature sensor to datacom switches.

alangregory:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/discovery/datacom.yaml
+++ b/includes/definitions/discovery/datacom.yaml
@@ -1,0 +1,13 @@
+mib: DMswitch-MIB
+modules:
+    sensors:
+        temperature:
+            data:
+                -
+                    oid: SwitchInfoEntry
+                    value: swTemperature 
+                    num_oid: .1.3.6.1.4.1.3709.3.5.201.1.1.2.1.12.
+                    descr: 'Temperature {{ $index }}'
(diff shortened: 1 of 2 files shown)


includes/definitions/discovery/datacom.yaml
```diff
@@ -0,0 +1,13 @@
...
+    sensors:
+        temperature:
+            data:
+                -
+                    oid: SwitchInfoEntry
+                    value: swTemperature 
+                    num_oid: .1.3.6.1.4.1.3709.3.5.201.1.1.2.1.12.
+                    descr: 'Temperature {{ $index }}'
+                    index: 'swTemperature.{{ $index }}'
+                    low_limit: 1
```
laf: How come you are hard coding the low / high limits? Is it not worth leaving it to be autocalculated?

alangregory: These are values supplied by the manufacturer. Where can i check how values are autocalculated ?

laf: Supplied in the docs but not via snmp? If it's not from snmp then leave it as is. The auto calc is done on the current value so can vary.

alangregory: Yep, no info on mibs.

PR #7524, opened 2017-10-20
Added support for logmaster(ups vendors) os/devices

alangregory:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


alangregory:
It seems that sysObjectId is used on hardware from differen ups sellers, is it possible to force the checking of sysDescr and sysObjectId?

laf:
zynos os does what you need.

Also, logo in png format should be 32x32

alangregory:
@laf ./scripts/pre-commit.php still shows errors on local dev. Is there some kind of cache for snmpsim ?

laf:
You might need to remove cache/os-defs.cache

alangregory:
XPPC-MIB ia a mib used by many vendors. Os is called logmaster.

laf:
Thanks for the updates. Last question, do these devices not support the rfc1628 code we already have?

alangregory:
Looks like 70% works with rfc1628 , so i'm only adding aditional sensors. Is it possible to overide  realtionship from some states ? Like:
https://github.com/librenms/librenms/blob/738067e6223391fcf3b9925707e48057eb1b2287/includes/discovery/sensors/state/rfc1628.inc.php#L51-L59
logmaster returns a diferente association of id and status.

laf:
It's not possible right now. We probably should work out some way to make that happen though.

alangregory:
Ok, added the state that matters.

Issue #4929, opened 2016-11-03
Some Windows hosts are not showing disk utilization / properties


gilrim:
DO NOT DELETE THIS INFORMATION.

```
==========================================================
LibreNMS Version: 6bed6cbea187b1a59b3cd049a3aa29ca6081723d
DB Schema: 148
PHP: 7.0.8-0ubuntu0.16.04.3
MySQL: 10.0.27-MariaDB-0ubuntu0.16.04.1
RRDTool: 1.5.5
SNMP: NET-SNMP 5.7.3
==========================================================
```

I've got a bunch of windows boxes added, and for most of them things are working smoothly. But I've also got a few five units that doesn't update its storage details for the C:\ drive. I'm including screenshots and discover/poller/snmp outputs - are there anything else I could provide to help identify what's happening?

From just looking at the snmp responses it seems that windows is providing the data, but librenms is messing up parsing it?

![2016_11_03_ _selection_001](https://cloud.githubusercontent.com/assets/434698/19962219/efe8904c-a1b6-11e6-8eeb-40f2ba1aca43.png)
![screenshot at 2016-11-03 11 13 23](https://cloud.githubusercontent.com/assets/434698/19962226/f5e8c264-a1b6-11e6-9193-6e6c34a01f02.png)
[poller-notrofl.txt](https://github.com/librenms/librenms/files/568651/poller-notrofl.txt)
[discovery-notrofl.txt](https://github.com/librenms/librenms/files/568650/discovery-notrofl.txt)

laf:
This looks to be an issue with the device, it's not returning data:

```
    [storage_id] => 872
    [device_id] => 147
    [storage_mib] => hrstorage
    [storage_index] => 2
    [storage_type] => hrStorageFixedDisk
    [storage_descr] => C:\\ Label:  Serial Number 4c65395d
    [storage_size] => 0
    [storage_units] => 0
    [storage_used] => 0
    [storage_free] => 0
    [storage_perc] => 0
    [storage_perc_warn] => 60
    [storage_deleted] => 0
    [units] =>
    [used] => 0
    [size] => 0
    [free] => 0
```

Used and size shouldn't be 0.

Not sure this is something we can fix.

gilrim:
I'm trying to make sense of the poller-log, but while I do find the numbers your referring to on lines 241-257, I also find it on the lines 321-327. There the numbers haven't been calculated yet, but both disksize and untilization is listed..
[...]

PR #7531, opened 2017-10-23
Don't init storage_cache array

tomarch:
This array has already been init in the mempools polling.
Be deleting this init, we use the $storage_cache array  populated by mempools polling, so we bypass a new snmp call.
This not just a optimisation, this also resolve issue with some WINDOWS (it timeout the second time we call hrstorage in a few second) like in #4929

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
That code is a nightmare :)

PR #7532, opened 2017-10-23
updated default list of modules in poller and discovery support docs.

barryodonovan:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

The list of en/disabled poller and discovery modules in the support docs was out of date.
They are now current as per ./include/defaults.inc.php


doc/Support/Poller Support.md
```diff
@@ -42,38 +42,49 @@ These are the default poller config items. You can globally disable a module by
...
+$config['poller_modules']['netstats']                    = 1;
+$config['poller_modules']['hr-mib']                      = 1;
+$config['poller_modules']['ucd-mib']                     = 1;
+$config['poller_modules']['ipSystemStats']               = 1;
+$config['poller_modules']['ports']                       = 1;
+$config['poller_modules']['bgp-peers']                   = 1;
+$config['poller_modules']['junose-atm-vp']               = 0;
+$config['poller_modules']['toner']                       = 0;
+$config['poller_modules']['ucd-diskio']                  = 1;
+$config['poller_modules']['wifi']                        = 0;
```
murrant: Can you remove wifi?  It is deprecated.

barryodonovan: It is still in the defaults file, should it not remain consistent until wifi is obsolete?

laf: It's still doing stuff though so should be kept until it is deprecated.

doc/Support/Poller Support.md
```diff
@@ -42,43 +42,54 @@ These are the default poller config items. You can globally disable a module by
...
+$config['poller_modules']['aruba-controller']            = 0;
+$config['poller_modules']['entity-physical']             = 1;
+$config['poller_modules']['applications']                = 1;
+$config['poller_modules']['mib']                         = 0;
+$config['poller_modules']['stp']                         = 1;
+$config['poller_modules']['ntp']                         = 1;
+$config['poller_modules']['services']                    = 1;
+$config['poller_modules']['loadbalancers']               = 0;
+$config['poller_modules']['mef']                         = 0;
+$config['poller_modules']['tnms-nbi']                    = 0;
```
laf: I think this line can go, tnms-nbi isn't a poller module.

barryodonovan:
OK, it is in the defaults file though, I just C&P from the file. I was trying to think of an easy way to just include the section of the file into the docs but more hassle than it's worth.

laf:
Can be deleted from the defaults as well then :D

doc/Support/Poller Support.md
murrant: No, each OS that needs it has it explicitly enabled in their OS definition.

It is set to false globally by default and there is no reason for a user to change that.

PR #7533, opened 2017-10-23
Support a wider range of link speeds in network maps

khobbits:
Add support for links at 25Gbps, 56Gbps, and those faster than 100Gbps.
Will likely need adjusted down as links higher than 100Gbps become common, as 20 width looks a bit silly.

If people would prefer, I could swap those matched to be <= matches, to catch odd link speeds.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/print-map.inc.php
+++ b/html/includes/print-map.inc.php
@@ -158,18 +158,10 @@ foreach ($list as $items) {
     }
 
     $speed = $items['local_ifspeed']/1000/1000;
-    if ($speed == 100) {
-        $width = 3;
-    } elseif ($speed == 1000) {
-        $width = 5;
-    } elseif ($speed == 10000) {
-        $width = 10;
-    } elseif ($speed == 40000) {
-        $width = 15;
-    } elseif ($speed == 100000) {
+    if ($speed > 500000) {
         $width = 20;
     } else {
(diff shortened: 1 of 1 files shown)


murrant:
Great, thanks for looking into this.  Might I suggest a function so we can handle all speeds instead of just explicitly defined ones:
```php
    if ($speed > 200000) {
        $width = 20;
    } else {
        // a nice curve
        $width = round(0.7412 * pow($speed, 0.2708));
    }
```

khobbits:
I was trying to keep changes to a minimum!
Sure, I'll have a test and resubmit.

laf:
> $width = round(0.75 * pow($val, 0.242));

This.

khobbits:
After playing with it a bit I was thinking:

```
$width = round(0.77 * pow($val, 0.25));

10	1
100	2
1000	4
10000	8
25000	10
40000	11
50000	12
56000	12
100000	14
200000	16
400000	19
500000	20
```

With the larger scale, the difference between 1gig and 10gig is less pronounced, and there is no difference between 40gig and 50gig.

500Gbps is still larger than most connections, and the scale could be adjusted in a few years, if 100mbps starts to fade, and faster connections are more common.

PR #7539, opened 2017-10-24
Disable modules enabled by default to improve discovery/poller speed.

alangregory:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Do these devices genuinely not support these modules or is this personal choice?

alangregory:
No support, except the port module, but it is unusable at this time.
Each Onu returns they interfaces without data, and then we receive an interface table entry of 6k to 10k. I am in contact with the manufacturer, when it is fixed, I will implement the correct reading.
Example:
https://gist.github.com/alangregory/c9409815fb1f3d52f9986eb88a157cb0

PR #7543, opened 2017-10-25
newdevice: initial proxmox detection

tomarch:
Dectect PROXMOX host

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/proxmox.yaml
```diff
@@ -0,0 +1,25 @@
...
+discovery_modules:
+    applications: 1
+    bgp-peers: 0
+    stp: 0
+    vmware-vminfo: 0
+    libvirt-vminfo: 0
+processor_stacked: 1
+discovery:
+    - sysDescr_regex:
+        - '/PVE/'
```
murrant: This doesn't need to be a regex.  You are just searching for PVE anywhere in the sysDescr.
```
 - sysDescr: PVE
```

murrant:
The os icon should be squarish.  Move the current svg you have to the logos directory and extract the left symbol for this os icon.

includes/definitions/proxmox.yaml
```diff
@@ -21,5 +21,7 @@ discovery_modules:
     libvirt-vminfo: 0
 processor_stacked: 1
 discovery:
-    - sysDescr_regex:
-        - '/PVE/'
+    - sysObjectId:
```
murrant: This is wrong, it means match either sysObjectID or sysDescr.   You need to have them in the same array (-).

includes/definitions/proxmox.yaml
```diff
@@ -0,0 +1,27 @@
...
+    applications: 1
+    bgp-peers: 0
+    stp: 0
+    vmware-vminfo: 0
+    libvirt-vminfo: 0
+processor_stacked: 1
+discovery:
+    - sysObjectId:
+            - .1.3.6.1.4.1.8072.3.2.10
+    - sysDescr:
```
murrant: This is wrong, it means match either sysObjectID or sysDescr. You need to have them in the same array (-).
Reply…

tomarch: Sorry. Should be fixed now.

includes/definitions/proxmox.yaml
```diff
@@ -0,0 +1,28 @@
...
+    stp: 0
+    vmware-vminfo: 0
+    libvirt-vminfo: 0
+processor_stacked: 1
+discovery:
+    -
+      sysObjectId:
+            - .1.3.6.1.4.1.8072.3.2.10
+      sysDescr:
+            - PVE
```
laf: I think to be 100% safe here it might be better to do:

```
sysDescr_regex:
    - '/ PVE /'
```

murrant: @laf just a note, this didn't need to be a regex `- ' PVE '` works too

laf: Good point

PR #7552, opened 2017-10-26
Eltek enexus

barryodonovan:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/eltek.yaml
```diff
@@ -0,0 +1,55 @@
...
+type: power
+icon: eltek
+#sysObjectID: .1.3.6.1.4.1.12148
+nobulk: 1
+mib_dir:
+    - eltek
+over:
+    - { graph: device_bits, text: 'Device Traffic' }
+discovery:
+    - .1.3.6.1.4.1.12148.10.2.4.0
```
murrant: This isn't valid.  remove the discovery section.

barryodonovan: Gone, but now I don't understand how it is doing discovery!

murrant: check #7558

murrant:
You might be able to use #7558 if/when it is merged.

includes/polling/os/eltek.inc.php
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2017 Tony Murray
+ * @author     Tony Murray <murraytony@gmail.com>
+ */
+
+$hardware = snmp_get($device, 'powerSystem.powerSystemModel.0', '-Ovqa', 'SP2-MIB') . ' ' . snmp_get($device, 'powerSystem.powerSystemNominalVoltage.0', '-Ovqa', 'SP2-MIB'). 'v Nominal voltage with ' . snmp_get($device, 'rectifiers.rectifiersNumberOfRectifiers.0', '-Ovqa', 'SP2-MIB') . ' rectifiers of type ' . snmp_get($device, 'rectifiers.rectifierTable.rectifierEntry.rectifierType.1', '-Ovqa', 'SP2-MIB');
```
laf: Please convert all snmp_gets into a single snmp_get_multi_oid(). However I'm not sure that hardware value is right. This is used to distinguish the hardware at a basic level (used in searches and the dropdown on device pages) so something this long is ott imho.

includes/polling/os/eltek.inc.php
```diff
@@ -0,0 +1,28 @@
...
+$version = snmp_get($device, 'controlSystem.controlSystemInventory.controlUnitTable.controlUnitEntry.controlUnitSwVersion.1', '-Ovqa', 'SP2-MIB');
+//$features = implode(', ', explode(PHP_EOL, snmp_walk($device, 'vpwrModuleOID', '-Oqvs', 'EltekEnexusPowersystem_branch10_V37')));
```
laf: Delete this if it's of no use.

tests/snmpsim/eltek.snmprec
```diff
@@ -0,0 +1 @@
+.1.3.6.1.4.1.12148.10.2.4.0|4|Eltek
```
laf: Is this the one that doesn't have a sysObjectId or sysDescr?

barryodonovan: Yes indeed.

barryodonovan:
@laf @murrant 
I'm quite happy to change it, it only covers a subset of the Eltek-Valere family.
Minicom is the model I have, eNexus is the product family, SP2 is the MIB, I'm not sure which is better? There is no obvious firmware name.

Perhaps, Eltek-eNexus? That's what I've called the branch for the PR as I think it is most accurate?

laf:
I'd go for just `enexus`

barryodonovan:
I'll push it tomorrow morning

barryodonovan:
Back to failing on the pre-commit script again.

Can't see why, last time it was the cache invalidation issue but not now.

laf:
Most likely because we had to remove the snmp_get ability in discovery, it's back in again now so just rebase and it should pass.

laf:
:( This has ended up with loads of commits we've already merged :(

barryodonovan:
Hopefully this rebase is a bit cleaner!

PR #7554, opened 2017-10-26
docs: Update network map.md

kkrumm1:
Updated the Network Map doc steps and example pictures.

#### Please note

#### Testers


doc/Extensions/Network-Map.md
```diff
@@ -12,55 +14,59 @@ $config['network_map_items'] = array('mac','xdp');
 ```
 
 Either remove mac or xdp depending on which you want.
+XDP is based on FDP, CDP and LLDP support based on the device type.
 
 A global map will be drawn from the information in the database, it is worth noting that this could lead to a large network map.
 Network maps for individual devices are available showing the relationship with other devices.
+Also you can Build Device Groups and those Device Groups can be drawn with Network Map.
```
laf: Can you?

kkrumm1: Yes -  Overview -> Maps -> Device Groups Maps

image

laf: I did not know that :)

PR #7558, opened 2017-10-26
feature: Allow snmpget in os discovery yaml

murrant:
Convert all remaining os except airos.

Affected OS: asuswrt-merlin, ddnons, dsm, extrahop, huaweiups, ibmtl, pcoweb, pktj, qnap, remoteeye4, sentry3, sentry4, tomato

There should be not change in detection.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
The main downside here is each snmpget adds a few seconds to OS discovery.  They were previously tried after all other OS, but now they will be tried alphabetically like others.  Perhaps we could add an optimization to defer all OS using snmpget till last, but I don't think it is that bad right now.

laf:
Awesome. Mixing up using mib and MIB:: in places but I don't see that as an issue.

PR #7564, opened 2017-10-27
fix: lldp find_port_id does not the correct port

murrant:
Order statements to give some results precedence over others.
Union the results but limit 1 so we can stop when we get the first result.
Checking if variables doesn't work very well without the device_id

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
I won't be able to get this tested before saturday, not sure we should be merging this into 1.33 though due to how late it is.

murrant:
It is a bug in the existing code because it will find ALL ports that match any criteria then return the one with the lowest port_id.  On some devices, this could map all links to the same port.

PR #7567, opened 2017-10-27
fix: fix a couple of OS cache invalidation issues

murrant:
Fixes issues where we might be using outdated os definitions.

Core issue was we had already loaded the os definitions from the cache before we call load_all_os().
We need to remove the previous definitions, (but keep user settings).

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Ok, happy with this now.

PR #7570, opened 2017-10-28
Set ifIndex to null if there is no dot1dBasePortIfIndex

ParthKolekar:
The ifIndex here is only being used to query for $local_port_id, and when there is no dot1dBasePortIfIndex, the default value is set to the index of the array. This is problematic because the index for interfaces in SNMP for devices do not start from 1. 

ifIndex = 1 is set for VLAN1,
ifIndex = 2 is set for VLAN2, and so forth.....

This results in librenms wrongly attaching the local_port_id to VLan1. 

Setting this to null should not be a problem. The reason for this is that $local_port_id should be totally resolvable from lldpLocPortId.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/discovery-protocols.inc.php
+++ b/includes/discovery/discovery-protocols.inc.php
@@ -126,7 +126,7 @@ if (($device['os'] == 'pbn' || $device['os'] == 'bdcom') && Config::get('autodis
             if (is_numeric($dot1d_array[$entry_key]['dot1dBasePortIfIndex'])) {
                 $ifIndex = $dot1d_array[$entry_key]['dot1dBasePortIfIndex'];
             } else {
-                $ifIndex = $entry_key;
+                $ifIndex = null;
             }
 
             $local_port_id = find_port_id($lldp_ports[$entry_key]['lldpLocPortId'], $ifIndex, $device['device_id']);


includes/discovery/discovery-protocols.inc.php
```diff
@@ -126,7 +126,7 @@
             if (is_numeric($dot1d_array[$entry_key]['dot1dBasePortIfIndex'])) {
                 $ifIndex = $dot1d_array[$entry_key]['dot1dBasePortIfIndex'];
             } else {
-                $ifIndex = $entry_key;
+                $ifIndex = null;
```
murrant: This will break other devices.

ParthKolekar: This sets an ifIndex = 1, 2, 3, 4, 5..... 

According to my knowledge, the physical ports are not at that index in SNMP. At least for my devices, the only things at those indices are VLAN1, VLAN2, VLAN3, VLAN4..... 


Also... as I mentioned in commit.... the ifIndex is only used for local_port finding. The LLDP should give the interface name or mac or something better. 


I do not know many devices, so this might be working. But this seems like a hack to assume that if the ifIndex is not dot1dbaseportifindex, it should be set to the index of the array.

ParthKolekar: Setting it to null might be better, because then it ignores arbritary lookups.

ParthKolekar: @murrant https://github.com/librenms/librenms/pull/7564 actually solves my problems, So I am ok if you don't want to apply this.

murrant:
@ParthKolekar Some devices don't use ifIndex and just return the port number for lldp.  This is how we match ports on those devices.

ParthKolekar:
Yes, I think it is ok to do. Since #7564 actually solves my issue.

PR #7572, opened 2017-10-28
Update RFC1123 Hostname Check

ParthKolekar:
The regex was wrong. It failed to match devices like `cisco-3750x` even though it is a valid hostname. 
I picked this regex from the RFC1123 documentation, and updated it.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/tests/CommonFunctionsTest.php
+++ b/tests/CommonFunctionsTest.php
@@ -122,12 +122,15 @@ class CommonFunctionsTest extends \PHPUnit_Framework_TestCase
     public function testIsValidHostname()
     {
         $this->assertTrue(is_valid_hostname('a'), 'a');
+        $this->assertTrue(is_valid_hostname('a.'), 'a.');
         $this->assertTrue(is_valid_hostname('0'), '0');
         $this->assertTrue(is_valid_hostname('a.b'), 'a.b');
         $this->assertTrue(is_valid_hostname('localhost'), 'localhost');
         $this->assertTrue(is_valid_hostname('google.com'), 'google.com');
         $this->assertTrue(is_valid_hostname('news.google.co.uk'), 'news.google.co.uk');
         $this->assertTrue(is_valid_hostname('xn--fsqu00a.xn--0zwm56d'), 'xn--fsqu00a.xn--0zwm56d');
+        $this->assertTrue(is_valid_hostname('www.averylargedomainthatdoesnotreallyexist.com'), 'www.averylargedomainthatdoesnotreallyexist.com');
+        $this->assertTrue(is_valid_hostname('cont-ains.h-yph-en-s.com'), 'cont-ains.h-yph-en-s.com');
         $this->assertTrue(is_valid_hostname('cisco-3750x'), 'cisco-3750x');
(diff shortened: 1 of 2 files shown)


tests/CommonFunctionsTest.php
```diff
@@ -122,13 +122,14 @@ public function testStringToClass()
     public function testIsValidHostname()
     {
         $this->assertTrue(is_valid_hostname('a'), 'a');
-        $this->assertTrue(is_valid_hostname('a.'), 'a.');
```
ParthKolekar: This is not a valid domain name anyway, according to me.

f0o: It is a valid hostname however.

`example.net` and `example.net.` are both the same and valid.
The latter of this example is a `true` FQDN as it denominates the absolute root whereas the first can still imply a search-domain.

ParthKolekar:
@f0o , I added support for domains ending in a dot. Can you review again?

ParthKolekar:
Not sure how I can reduce the complexity of the regex. It is as it is. 😅

tests/CommonFunctionsTest.php
```diff
@@ -129,11 +129,16 @@ public function testIsValidHostname()
         $this->assertTrue(is_valid_hostname('google.com'), 'google.com');
         $this->assertTrue(is_valid_hostname('news.google.co.uk'), 'news.google.co.uk');
         $this->assertTrue(is_valid_hostname('xn--fsqu00a.xn--0zwm56d'), 'xn--fsqu00a.xn--0zwm56d');
+        $this->assertTrue(is_valid_hostname('www.averylargedomainthatdoesnotreallyexist.com'), 'www.averylargedomainthatdoesnotreallyexist.com');
+        $this->assertTrue(is_valid_hostname('cont-ains.h-yph-en-s.com'), 'cont-ains.h-yph-en-s.com');
+        $this->assertTrue(is_valid_hostname('cisco-3750x'), 'cisco-3750x');
+        $this->assertFalse(is_valid_hostname('cisco_3750x'), 'cisco_3750x');
```
murrant: This is a valid hostname?

murrant:
I'm confused by this.  It doesn't look the the regex was changed really, just the syntax.

I added all the test cases including a couple of new ones and it passes with the old regex.

ParthKolekar:
I could have sworn that `cisco-3750x` failed. But now that I re-check it, it does not fail. Seems like I have fat fingers, and should check things. 😅

ParthKolekar:
Since this adds un-necessary complexity to the hostname regex, please feel free to revert this.

laf:
I've reverted this PR. It would be good still if you can add some additional tests to cover your use case.

ParthKolekar:
A dumb github question... But will it still track things if I add them to this branch? I.E. will adding my tests back require a separate PR?

laf:
Yes it will need a separate PR. I'd delete this branch now and do a new one.

murrant:
Yeah the histories have diverged, it will be much easier to start a new branch.

PR #7579, opened 2017-10-29
Re-add tests from #7572

ParthKolekar:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


ParthKolekar:
Just found this interesting undocumented feature. https://github.com/php/php-src/pull/826

This does a pretty good job of validating domains in PHP library.

murrant:
Unfortunately, that seems to be PHP7.

ParthKolekar:
True. But when we do drop support, we'll have a good library function to depend on for this.

PR #7582, opened 2017-10-30
Fix: Better sanity check of hostname when adding device.

Zmegolaz:
Fixes a bug where you can add devices without hostname.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/addhost.inc.php
+++ b/html/pages/addhost.inc.php
@@ -11,10 +11,14 @@ if ($_SESSION['userlevel'] < 10) {
     exit;
 }
 
-$hostname = (isset($_POST['hostname']) && clean($_POST['hostname']) != "") ? clean($_POST['hostname']) : false;
+if (isset($_POST['hostname']) && clean($_POST['hostname']) != "") {
+    $hostname = clean($_POST['hostname']);
+} else {
+    $hostname = false;
+}
 if ($hostname !== false && !is_valid_hostname($hostname) && !IP::isValid($hostname)) {
     $hostname = false;
-    print_error('Invalid hostname.');
+    print_error('Invalid hostname or IP.');
 }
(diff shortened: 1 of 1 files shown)


murrant:
You forgot to check for IPs.

`IP::isValid($hostname)` might be handy ;)

Zmegolaz:
is_valid_hostname() works for IPs too, since it allows dots and `[a-z\d]` mixed. Or is there a risk it will be modified to do a more strict check later on?

laf:
Should we not be assigning $hostname = clean($_POST['hostname']); before all the checks then checking $hostname otherwise we may have a case where is_valid_hostname() is true before clean() is applied?

murrant:
@Zmegolaz Does it work for IPv6 addresses?

Zmegolaz:
Very good points, both of you. I've updated it.

html/pages/addhost.inc.php
```diff
@@ -10,7 +11,11 @@
     exit;
 }
 
-$hostname = isset($_POST['hostname']) ? clean($_POST['hostname']) : false;
+$hostname = (isset($_POST['hostname']) && clean($_POST['hostname']) != "") ? clean($_POST['hostname']) : false;
```
murrant: Can you change this from a ternary to a regular if else?  That would make it more readable.

html/pages/addhost.inc.php
```diff
@@ -10,7 +11,11 @@
...
+if ($hostname !== false && !is_valid_hostname($hostname) && !IP::isValid($hostname)) {
+    $hostname = false;
+    print_error('Invalid hostname.');
```
murrant: The text should probably be `Invalid hostname or IP.`

PR #7583, opened 2017-10-30
Fix Services container alignment

florianbeer:
This has been bugging me for a while... ;)
The services container wasn't properly aligned with the rest of the boxes (e.g. Events, Syslog etc.). This small HTML change fixes that.

#### Testers


murrant:
Nice, and valid Boostrap now too :)

PR #7585, opened 2017-10-30
Fail isPingable() if fping fails, take 2

Zmegolaz:
Another take on https://github.com/librenms/librenms/pull/7426.
If the command fping fails, the function isPingable() should fail too, for example for invalid hostnames.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
I feel like it would be a bit cleaner to leave the return array as a single level array and either unset exitcode before updating the db or array_filter() the array for valid keys for the database.

Third option is to add exitcode field to the database, but I don't think that would be worth storing.

PR #7586, opened 2017-10-30
fix: improve poller validation

murrant:
remove failures when there are duplicate poller entries
add warning that there are duplicate poller entries
link to the poller log in the webui
small fix to getBaseURL() to avoid exception when run from cli.
extract poller validations to private methods for better readability

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Validations/Poller.php
+++ b/LibreNMS/Validations/Poller.php
@@ -53,35 +53,81 @@ class Poller implements ValidationGroup
             return; // can't check poller/discovery if there are no devices.
         }
 
-        // FIXME pollers table is only updated by poller-wrapper.py
-        // check poller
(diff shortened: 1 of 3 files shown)


laf:
Worth applying 212.sql again as another schema update?

murrant:
@laf I dunno maybe, I think it was a timing issue.  This at least reports the correct issues even if there are duplicates.

laf:
I think we should, no harm in doing so and it will save the expected questions we will get asked or at least stop users having to manually fix.

murrant:
Want it in this PR or a separate one?

laf:
This one makes the most sense. Happy to merge straight after that.

PR #7566, opened 2017-10-27
refactor: use one snmpget during os discovery

murrant:
Use only one snmpget to fetch sysObjectId and sysDescr
fix multiline data in snmp_get_multi_oid()
add snmp_get_multi_oid() to mock.snmp.inc.php

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [ ] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

PR #7587, opened 2017-10-30
feature: Allow snmpget in os discovery yaml

murrant:
Re-submit after release of: #7558, #7561, #7566 
Minor change to trim quotes from snmpget...

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/functions.php
```diff
@@ -173,6 +182,21 @@ function checkDiscovery($array, $sysObjectId, $sysDescr)
...
+        } elseif ($key == 'snmpget') {
+            $options = isset($value['options']) ? $value['options'] : '-Oqv';
+            $mib = isset($value['mib']) ? $value['mib'] : null;
+            $mib_dir = isset($value['mibdir']) ? $value['mibdir'] : null;
+            $op = isset($value['op']) ? $value['op'] : 'contains';
+
+            $get_value = snmp_get($device, $value['oid'], $options, $mib, $mib_dir);
+
+            if (is_string($get_value)) {
+                $get_value = trim($get_value, '"');
```
laf: We really need to move this check into snmp_get() imho. Trimming " is littered everywhere within the code and I just can't see where doing it would cause us issues.

murrant:
The cache invalidation should be run pretty early in daily.sh after the code update, but there is still a window that discovery could start in to have invalid cache.

laf:
But the cache invalidation won't run until the next time daily.sh does so if someone updates from before that PR to this or later then the cache issue still persists.

includes/functions.php
laf: Ok, the PR that trims that from snmp_get is in so you should be able to remove this line now. Feel free to merge straight after.

murrant:
We'll merge this sometime tomorrow.

PR #7593, opened 2017-10-31
Fix: discovery.php -h all stops working after ping only device

Zmegolaz:
If a ping only device is discovered before any other device, the discovery_modules config variable was emptied for the rest of them.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
> I would prefer if snmp_disable, return at the beginning of this.

Out of interest why?

PR #7600, opened 2017-11-01
Add runtime divisor to fxm

pheinrichs:
Runtime was returning in seconds for the Alpha FXM so a 60 divisor was added to reduce to minutes

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
I've just seen that the first line is `mib: mibs/alpha/Argus-Power-System-MIB` when it should be `mib: Argus-Power-System-MIB` - can you update please?

PR #7601, opened 2017-11-01
Single Sign-On Authentication Mechanism

TheMysteriousX:
This is an authentication module that can be used to integrate LibreNMS with single sign-on providers.

The authentication code is imperative rather than object oriented; this has meant that I've had to make more changes than I'd like.

There are tests included, but they must be executed standalone and with config.php manually set (hence the .disabled extension) - the authentication code being imperative means that once a mechanism is loaded, it cannot be unloaded or replaced, breaking other tests.

I've also stopped http-auth being special cased in authenticate.php by tweaking it to use the two methods I've added.

It's been a few years since I've written any PHP, so feedback is gratefully received.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Authentication/AuthorizerBase.php
+++ b/LibreNMS/Authentication/AuthorizerBase.php
@@ -33,6 +33,7 @@ abstract class AuthorizerBase implements Authorizer
 {
     protected static $HAS_AUTH_USERMANAGEMENT = 0;
     protected static $CAN_UPDATE_USER = 0;
+    protected static $AUTH_IS_EXTERNAL = 0;
 
     /**
      * Log out the user, unset cookies, destroy the session
@@ -243,5 +244,15 @@ abstract class AuthorizerBase implements Authorizer
     {
         //not supported by default
         return 0;
+    }
+
+    public function authIsExternal()
+    {
+        return static::$AUTH_IS_EXTERNAL;
+    }
+
+    public function getExternalUsername()
+    {
+        return null;
     }
 }
--- a/LibreNMS/Authentication/Auth.php
+++ b/LibreNMS/Authentication/Auth.php
@@ -25,6 +25,7 @@ class Auth
                 'http-auth' => 'LibreNMS\Authentication\HttpAuthAuthorizer',
                 'ad-authorization' => 'LibreNMS\Authentication\ADAuthorizationAuthorizer',
                 'ldap-authorization' => 'LibreNMS\Authentication\LdapAuthorizationAuthorizer',
+                'sso' => 'LibreNMS\Authentication\SSOAuthorizer',
             );
 
             $auth_mechanism = Config::get('auth_mechanism');
--- a/LibreNMS/Authentication/Auth.php
+++ b/LibreNMS/Authentication/Auth.php
@@ -37,4 +37,15 @@ class Auth
         }
         return static::$_instance;
     }
+
+    /**
+     * Destroy the existing instance and get a new one - required for tests.
+     *
+     * @return Authorizer
+     */
+    public static function reset()
+    {
+        static::$_instance = null;
+        return static::get();
+    }
 }
--- a/LibreNMS/Authentication/HttpAuthAuthorizer.php
+++ b/LibreNMS/Authentication/HttpAuthAuthorizer.php
@@ -10,6 +10,7 @@ class HttpAuthAuthorizer extends AuthorizerBase
 {
     protected static $HAS_AUTH_USERMANAGEMENT = 1;
     protected static $CAN_UPDATE_USER = 1;
+    protected static $AUTH_IS_EXTERNAL = 1;
 
     public function authenticate($username, $password)
     {
@@ -102,5 +103,14 @@ class HttpAuthAuthorizer extends AuthorizerBase
     public function updateUser($user_id, $realname, $level, $can_modify_passwd, $email)
     {
         dbUpdate(array('realname' => $realname, 'level' => $level, 'can_modify_passwd' => $can_modify_passwd, 'email' => $email), 'users', '`user_id` = ?', array($user_id));
+    }
+
+    public function getExternalUsername()
+    {
+        if (isset($_SERVER['REMOTE_USER'])) {
+            return clean($_SERVER['REMOTE_USER']);
+        } elseif (isset($_SERVER['PHP_AUTH_USER'])) {
+            return clean($_SERVER['PHP_AUTH_USER']);
+        }
     }
 }
--- a/LibreNMS/Authentication/SSOAuthorizer.php
+++ b/LibreNMS/Authentication/SSOAuthorizer.php
@@ -1,0 +1,297 @@
+<?php
+/**
+ * SSOAuthorizer.php
+ *
+ * -Description-
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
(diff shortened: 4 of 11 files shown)


TheMysteriousX:
Looks like there are a few compatibility problems to fix up - I'll sort those momentarily.

TheMysteriousX:
Code now passes Travis' checks under PHP 5.3.
I can't do much about the Scrutinizer failures; what it's flagging up is 100% correct, but is due to the auth mechanisms having repeated code and identical functions in each file.

laf:
Thanks for contributing this.

I've not gone through the code in detail yet and have no real way to test this but it seems like it would be better to put the config into another level within the config array like radius: https://docs.librenms.org/#Extensions/Authentication/#radius-authentication

I.e `$config['sso']['sso_group_strategy']`

TheMysteriousX:
Just pushed a new version with the configuration moved down a level.

I've not tested it this way, but it should be possible to simulate a single sign-on environment using SetEnv - something like this in your Apache config:

```
SetEnv REMOTE_USER ssouser
SetEnv mail ssouser@example.org
SetEnv displayName "SSO User"
SetEnv level 10
```

Then a basic config like:

```
$config['sso']['mode'] = 'env';
$config['sso']['realname_attr'] = 'displayName';
$config['sso']['email_attr'] = 'mail';
$config['sso']['level_attr'] = 'level';
$config['sso']['create_users'] = true;
$config['sso']['update_users'] = true;
$config['sso']['group_strategy'] = 'attribute';
```

Which should create an entry in the db for 'ssouser', and allow access at level 10. Might be a few rough edges in this configuration, but it should work.

html/includes/authentication/sso.inc.php
```diff
@@ -0,0 +1,336 @@
...
+ * @author     Adam Bishop <adam@omega.org.uk>
+ */
+
+use LibreNMS\Config;
+use LibreNMS\Util\IP;
+use LibreNMS\Exceptions\AuthenticationException;
+use LibreNMS\Exceptions\InvalidIpException;
+
+/**
+ * Some functionality in this mechanism is inspired by confluence_http_authenticator (@chauth) and graylog-plugin-auth-sso (@Graylog)
```
laf: Please add both of these packages to the Acknowledgements doc: https://github.com/librenms/librenms/blob/master/doc/General/Acknowledgement.md

html/includes/authentication/sso.inc.php
```diff
@@ -0,0 +1,336 @@
...
+
+function init_auth()
+{
+    // No initialisation is required
+}
+
+function authenticate($username, $password)
+{
+    if (empty($username)) {
+        throw new AuthenticationException('sso.user_attr was not found or was empty');
```
laf: Users don't see config values like that so you will need to `$config['sso']['user_attr']` in this output

html/includes/authentication/sso.inc.php
```diff
@@ -0,0 +1,336 @@
...
+    // Build the user's details from attributes
+    $email = auth_sso_get_attr(Config::get('sso.email_attr'));
+    $realname = auth_sso_get_attr(Config::get('sso.realname_attr'));
+    $description = auth_sso_get_attr(Config::get('sso.descr_attr'));
+    $can_modify_passwd = 0;
+
+    $level = auth_sso_calculate_level();
+
+    // User has already been approved by the authenicator so if automatic user create/update is enabled, do it
+    if (Config::get('sso.create_users') && !user_exists($username)) {
```
laf: Is this block going to be called on every page and every graph load as we run `authenticate()` each time.

html/includes/print-menubar.php
```diff
@@ -639,7 +639,7 @@
 
 if ($_SESSION['authenticated']) {
     echo('
-           <li><a href="logout/"><i class="fa fa-sign-out fa-fw fa-lg" aria-hidden="true"></i> Logout</a></li>');
+           <li><a href="' . $config['auth_logout_handler']  . '"><i class="fa fa-sign-out fa-fw fa-lg" aria-hidden="true"></i> Logout</a></li>');
```
laf: Should switch to `Config::get()` here as well.

html/includes/authentication/sso.inc.php
TheMysteriousX: Done.

html/includes/authentication/sso.inc.php
TheMysteriousX: I don't believe that's the case - I threw in a print statement and authenticate() only seems to be called on the first request. It gets short circuited if there's an active session.

If it did get called repeatedly (user with broken cookie settings) it's idempotent, and if a database cache is present, the impact is just an additional sql query per page load.

html/includes/print-menubar.php
TheMysteriousX: Done, by changing how this is done. The template isn't really the right place to handle authentication stuff as it's prone to being altered. I implemented the same functionality by making the URL the user is sent to after log out configurable.

html/includes/authentication/sso.inc.php
TheMysteriousX: Done.

html/includes/authentication/sso.inc.php
laf: Actually I've just remember we switched to using sessions once validated so you're right on this.

laf:
LGTM.

laf:
one thing to note actually. Someone has got a PR in the move all of the auth modules to classes #7497 

As that's been in for a while, it would be unfair to ask that user to do the work to convert this as well. Would you be able to migrate this once 7497 is merged?

TheMysteriousX:
Sure that's fine - I have a couple of questions about the new structure, I'll pop them in the other pull request.

TheMysteriousX:
Answered my questions with a bit more reading.

laf:
@TheMysteriousX The other PR is now merged in.

TheMysteriousX:
Naïve rebase, do not merge as it probably doesn't work (yet).

TheMysteriousX:
I think this is now ported. I'll give it another once over with fresher eyes tomorrow.

Summary of changes:
* Migrated a couple of methods into the interface/base class
* Converted imperative mechanism to class based
* Enabled tests as authentication is now mutable
* Added Auth::reset to allow a new mechanism to be loaded at runtime

The CI check that is failing is PHP5.3 - I've used the wrong array syntax to support this.

I can go back and change the syntax, but I think PHP5.3 was being dropped?

laf:
It's not being dropped until January so this will need to be resolved before then.

LibreNMS/Authentication/AuthorizerBase.php
```diff
@@ -244,4 +245,14 @@ public function updateUser($user_id, $realname, $level, $can_modify_passwd, $ema
         //not supported by default
         return 0;
     }
+
+    public function authExternal()
```
murrant: This would probably be better named `authIsExternal()`

murrant:
:+1: for implementing unit tests :)

TheMysteriousX:
Great, thanks for reviewing.

TheMysteriousX:
Thanks for merging - the company I work for use this mechanism on their instance, so feel free to @ me if there are any issues.

murrant:
@TheMysteriousX I'm looking to refactor this a bit to inheret from MysqlAuthorizer.

I can't figure out why you ignore the given $username in functions and use $this->getExternalUsername() instead?  Would this cause errors if we query for another user besides the logged in one?

TheMysteriousX:
That would be fine.

The username passed by functions should always be taken by from getExternalUsername, so by taking it from there it eliminates the possibility that the username could have been modified (accidentally or maliciously).

murrant:
But those functions return info about the requested user, not the logged in user. 

For example, say you are viewing a shared dashboard with your code all shared dashboards would show the currently logged in user as the owner.

TheMysteriousX:
Ok, I misunderstood their purpose then - I assumed getUserid/getUserLevel/userExists were intended as internal methods as they're not called outside of the auth code (except for scripts that do weird things, like the IRC bot).

I've checked and it doesn't seem to affect functionality - dashboards are all correctly owned, modifying users doesn't result in incorrect values being assigned etc.

I can send a pull to change the values if you like, or have you already changed it in your branch?

murrant:
The plan is to change the class to extend MysqlAuth (to remove code duplication) and remove extraneous functions.  I would prefer if you make the change as I can only test with the unit tests.

PR #7603, opened 2017-11-01
fix: ping fails on servers that don't set PATH in cron

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Config.php
+++ b/LibreNMS/Config.php
@@ -157,10 +157,27 @@ class Config
      *
      * @param mixed $key period separated config variable name
      * @param mixed $value
+     * @param bool $persist set the setting in the database so it persists across runs
      */
-    public static function set($key, $value)
+    public static function set($key, $value, $persist = false)
     {
         global $config;
+
+        if ($persist) {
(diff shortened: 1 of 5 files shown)


LibreNMS/Config.php
```diff
@@ -157,10 +157,27 @@ public static function getCombined($os, $key, $default = array())
...
+
+        if ($persist) {
+            $res = dbUpdate(array('config_value' => $value), 'config', '`config_name`=?', array($key));
+            if (!$res && !dbFetchCell('SELECT 1 FROM `config` WHERE `config_name`=?', array($key))) {
+                $insert = array(
+                    'config_name' => $key,
+                    'config_value' => $value,
+                    'config_default' => '',
+                    'config_descr' => '',
+                    'config_group' => '',
```
laf: For rrdtool binary in the config table we have 'external' set as group and sub_group - do the same here? May as well also set the default the first time.

murrant: I'm not sure what that would look like, perhaps optional arguments on the Config::set command?

laf: It's just a DB thing, it's so we could add them to the settings page.

PR #7605, opened 2017-11-01
Fix: Only add args.ping if it's set.

Zmegolaz:
snmp-scan.py froze when not using -P for ping only. `check_output()` apparently doesn't ignore empty arguments.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/snmp-scan.py
+++ b/snmp-scan.py
@@ -122,7 +122,7 @@ def scan_host(ip):
 
         try:
             arguments = ['/usr/bin/env', 'php', 'addhost.php', hostname or ip]
-            if args.ping != '':
+            if args.ping:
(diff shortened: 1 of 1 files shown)


snmp-scan.py
```diff
@@ -121,7 +121,10 @@ def scan_host(ip):
             pass
 
         try:
-            add_output = check_output(['/usr/bin/env', 'php', 'addhost.php', args.ping, hostname or ip])
+            arguments = ['/usr/bin/env', 'php', 'addhost.php', hostname or ip]
+            if args.ping != '':
```
murrant: `if args.ping:` is all that is needed here.

PR #7608, opened 2017-11-02
check % disabled ports before per-port-polling

barryodonovan:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Support/Performance.md
+++ b/doc/Support/Performance.md
@@ -15,7 +15,7 @@ We absolutely recommend running this, it will save on IO load. [RRDCached](http:
 It's advisable after 24 hours of running MySQL that you run [MySQL Tuner](https://raw.githubusercontent.com/major/MySQLTuner-perl/master/mysqltuner.pl)
 which will make suggestions on things you can change specific to your setup.
 
-One recommendation we can make is that you set the following in my.cnf under a [mysqld] group: 
+One recommendation we can make is that you set the following in my.cnf under a [mysqld] group:
 
 ```bash
 innodb_flush_log_at_trx_commit = 0
@@ -39,12 +39,12 @@ You can disable modules globally then re-enable the module per device or the opp
 
 #### SNMP Max Repeaters
 
-We have support for SNMP Max repeaters which can be handy on devices where we poll a lot of ports or bgp sessions for instance and 
+We have support for SNMP Max repeaters which can be handy on devices where we poll a lot of ports or bgp sessions for instance and
 where snmpwalk or snmpbulkwalk is used. This needs to be enabled on a per device basis under edit device -> snmp -> Max repeaters.
 
 You can also set this globally with the config option `$config['snmp']['max_repeaters'] = X;`.
 
-It's advisable to test the time taken to snmpwalk IF-MIB or something similar to work out what the best value is. To do this run the following 
+It's advisable to test the time taken to snmpwalk IF-MIB or something similar to work out what the best value is. To do this run the following
 but replace <REPEATERS> with varying numbers from 10 upto around 50. You will also need to set the correct snmp version, hostname and community string:
 
 `time snmpbulkwalk -v2c -cpublic HOSTNAME -Cr<REPEATERS> -M /opt/librenms/mibs -m IF-MIB IfEntry`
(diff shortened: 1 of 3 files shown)


doc/Support/Performance.md
```diff
@@ -62,30 +62,32 @@ You can also set this globally with the config option `$config['snmp']['max_oid'
...
-The default 16 threads that `poller-wrapper.py` runs as isn't necessarily the optimal number. A general rule of thumb is 
+The default 16 threads that `poller-wrapper.py` runs as isn't necessarily the optimal number. A general rule of thumb is
 2 threads per core but we suggest that you play around with lowering / increasing the number until you get the optimal value.
 
 This can be changed by going to the cron job for librenms. Usually in /etc/cron.d/librenms and changing the "16"
 
 */5  *    * * *   librenms    /opt/librenms/cronic /opt/librenms/poller-wrapper.py 16
 
-KEEP in MIND that this dosnt always help, it depnds on your system and CPU. So Be careful. 
+KEEP in MIND that this dosnt always help, it depnds on your system and CPU. So Be careful.
```
murrant: dosnt -> does not

includes/polling/ports.inc.php
```diff
@@ -176,7 +176,7 @@
 if ($device['os'] === 'f5' && (version_compare($device['version'], '11.2.0', '>=') && version_compare($device['version'], '11.7', '<'))) {
     require_once 'ports/f5.inc.php';
 } else {
-    if ($config['polling']['selected_ports'] === true || $device['attribs']['selected_ports'] == 'true') {
+    if ($config['polling']['selected_ports'] === true || $device['attribs']['selected_ports'] == 'true' && dbFetchRow("SELECT (count(*) / (SELECT count(*) from `ports` WHERE `device_id` = ?))> 0.33 FROM `ports` WHERE `device_id`= ? AND `deleted` = 1 HAVING count(*) > 10", array($device['device_id'],$device['device_id']))) {
```
murrant: Use dbFetchCell() to get a single result.

murrant:
I think you should take ifAdminStatus and disabled into account too.

Perhaps extract the variables to settings so they are tunable.

I think we need to collect some data.

I came up with this query to show port stats, but I think we need the time it takes on each device for the full poll and per-port poll:

```
SELECT
  `op`.`device_id`,
  (SELECT `os` FROM `devices` WHERE `device_id` = `op`.`device_id`) AS `os`,
  count(*) AS `total`,
  ((SELECT count(*) from `ports`
     WHERE `device_id` = `op`.`device_id` AND
     (`deleted` = 1 OR `ifAdminStatus` != 'up' OR `disabled` = 1)
   ) / count(*)) AS `ratio`
FROM `ports` AS `op`
GROUP BY `op`.`device_id` ORDER BY `ratio` DESC;
```

doc/Support/Performance.md
barryodonovan: Also a typo in `depnds` but I don't know how to pull your changes back into mine...

barryodonovan:
image
@murrant 
This is one device, a mikrotik router with ~60 pppoe sessions and 13 ports. You can see the point at which I turned on per-port

murrant:
I made script to collect stats, useful?  #7626

includes/polling/ports.inc.php
```diff
@@ -176,7 +176,7 @@
 if ($device['os'] === 'f5' && (version_compare($device['version'], '11.2.0', '>=') && version_compare($device['version'], '11.7', '<'))) {
     require_once 'ports/f5.inc.php';
 } else {
-    if ($config['polling']['selected_ports'] === true || $device['attribs']['selected_ports'] == 'true') {
+    if ($config['polling']['selected_ports'] === true || $device['attribs']['selected_ports'] == 'true' && dbFetchCell("SELECT (count(*) / (SELECT count(*) from `ports` WHERE `device_id` = ?))> 0.33 FROM `ports` WHERE `device_id`= ? AND `deleted` = 1 HAVING count(*) > 10", array($device['device_id'],$device['device_id']))) {
```
Zmegolaz: Could you move the SQL outside of the if(), to make it easier to read?

laf:
I'm proposing that we close this.

 Overall I think it's definitely the right approach that we try and work out the best options automatically but I think we need to do it in a structure way so we can re-use the code for more than just selected port polling. @librenms/reviewers

murrant:
Sounds good to me.  Let's revisit this at a later time.

PR #7609, opened 2017-11-02
Fix: Correct sensor ID when removing device.

Zmegolaz:
Correct variable name when removing sensors.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
You've done this commit from the wrong user so the CLA needs signing as that user.

Issue #3606, opened 2016-06-06
Add support for HWg-STE Ethernet thermometer


MatGit5:
Please add support for HWg-STE Ethernet thermometer

http://www.hw-group.com/products/HWg-STE/STE_ip_temperature_sensor_en.html

[snmpwalk.txt](https://github.com/librenms/librenms/files/300389/snmpwalk.txt)
[discovery.txt](https://github.com/librenms/librenms/files/300388/discovery.txt)
[poller.txt](https://github.com/librenms/librenms/files/300387/poller.txt)

[HWg-STE_OID.txt](https://github.com/librenms/librenms/files/300393/HWg-STE_OID.txt)
[HWg-STE.mib.txt](https://github.com/librenms/librenms/files/300400/HWg-STE.mib.txt)

Rosiak:
@MatGit5 Could you please try to see if it's possible to get more out of the SNMPWALK part?
Pretty sure it holds more information than that.

MatGit5:
snmpwalk.
2 temperature sensors attached,
Temperature reported in Celsius, once in string value and then in Integer x 10.
[snmpwalk-sensors.txt](https://github.com/librenms/librenms/files/310673/snmpwalk-sensors.txt)

Rosiak:
This still doesn't look like complete output.
Can you please post the output of this command?

```
snmpbulkwalk -On -v2c -c COMMUNITY HOSTNAME .
```

MatGit5:
Actually, seems like this device does not support/respond to snmp bulk... maybe it will with a newer firmware.
It has two inputs that accept temperature and humidity sensors.
It is possible that other variants of HWg devices provide more info.
Please check out the demo here: http://ste.hwg.cz

laf:
If you add the device as v1 then it won't use bulkwalk.

PR #7610, opened 2017-11-02
device: update hwg-ste and add hwg-ste-plus support

tomarch:
Update HWG-STE support:
-disable bulk
-disable discovery and poller that don't work (the poller time goes from 50s to 2sec now)
-remove bits over graph
-add temperature over graph
-update STE-MIB
-remove php file for temperature and sensors (replace by yaml)

Add HWG-STE-PLUS (Same as HWG-STE but with 2 Dry Contacts)
-use same definition as HWG-STE
-dry contact value (open or closed)
-dry contact alarm status (normal or alarm)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
I'm assuming people lose the old data with this update?

tomarch:
I suppose nobody have data unless they have disable bulkwalk manually (this device doesn't support bulk).
How to not lose old data (except by create a new os) ?

laf:
The only way to not lose old data would be to match up the indexes.

Are you 100% all of these devices don't work with bulkwalk or just some do some don't?

tomarch:
Sure the hwg-ste-plus don't work with bulkwalk.
For the hwg-ste it's seem to be the same (as seen in #3606) and it's the same os.

PR #7614, opened 2017-11-02
newdevice: support Avtech RoomAlert 32E/W and RoomAlert 11E

murrant:
Move other avtech sensors to yaml
This required more flexibility when checking pre-cached data

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixes: #7530


murrant:
@laf no there are mibs for different devices, that have the same names, but different oids.

I did not want to add an os for each device.  Also there are no tables anyway so it wouldn't be much different.

I took care to maintain the rrd name of the temperature sensors, but I think I couldn't make it match for the state sensors.

PR #7617, opened 2017-11-03
Fix: Better sanity checks of fping options.

Zmegolaz:
If someone specifies too aggressive fping options, the fping command fails and all devices fail the ping check.

Should we mention this in the docs?

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -661,16 +661,28 @@ function isPingable($hostname, $address_family = AF_INET, $attribs = array())
     $response = array();
     if (can_ping_device($attribs) === true) {
         $fping_params = '';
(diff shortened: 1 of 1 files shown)


murrant:
Default settings show this warning, can you fix that too?

```
fping -e -q  -r 3 -t 500 -c 3 -p 200 localhost
fping: warning: timeout (-t) value larger than period (-p) produces unexpected results
localhost : xmt/rcv/%loss = 3/3/0%, min/avg/max = 0.03/0.03/0.04

```

Zmegolaz:
Updated! These checks/modifications will currently be done for each host, should we instead move it to some more central place?

PR #7618, opened 2017-11-03
 fix: devices detected as ibmtl or generic

murrant:
Guess at the sysObjectID, please report if you have devices that should be ibmtl and are no longer detected.
Add output of os in discovery.
Change the way we clear the os def cache during daily update remove it (it will be regenerated on next discovery run)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/ibmtl.yaml
```diff
@@ -5,6 +5,8 @@ icon: generic
 over:
     - { graph: device_bits, text: Traffic }
 discovery:
-    - snmpget:
+    -
+      sysObjectId: .1.3.6.1.4.1.14851
```
laf: http://www.snmplink.org/cgi-bin/nd/m/Ent/I/IBM%20Corp/IBM/ibmProd/TotalStorage%20TS%20(Tape%20Storage)%20Systems/TS-3500%20-%20Tape%20Library/3584%20TotalStorage%20UltraScalable%20Tape%20Library/2009__09__23/SNIA-SML-rev1-21.mib

Indicates that's probably right.

murrant: Close enough :)

joelsdc:
Hi, please let me know if you want a separate issue, but I just found today that we are facing the same problem with ServerTech PDUs (Sentry3).
 
Until Friday all was OK, today all our PDUs are reported as Generic devices, thus no power/current graphs.

Discovery says:

OS: Generic Device (generic)

Please advice.

Thanks,
Joel.

Issue #7621, opened 2017-11-03
Add support for Arista power supply state in sensors table


shimamizu:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ x Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

###############

LibreNMS is not currently collecting *any* sensor_class = state alarms for Arista devices. I have spoken to Arista and gotten the MIB's you need to poll in order to get the "Status" column from the CLI output you would see in "show environment power" to be able to alert on Power Supply state. 
Please add this logic into Arista as this will then allow easy creation of Alerts for knowing when a power supply goes bad / loses power on an Arista switch.

Here's the email I got back from Arista TAC requesting information on the MIB to grab this value: 

> Unfortunately we don't have a straightforward way to poll the operating status of specifically PSUs at this time. We do have RFE210545 opened to provided a consolidated PSU MIB, and I have added you to the requesters list for the functionality. The RFE does not currently have a targeted release date, in the mean time you can use the following method which uses each PSUs SNMP index to poll it's operating status with the Entity MIB:
> 
> Find the SNMP Index of the PSUs:
> sq514...16:11:10#show snmp mib walk entPhysicalDesc | egrep "PowerSupply.$"
> ENTITY-MIB::entPhysicalDescr[100711000] = STRING: PowerSupply1
> ENTITY-MIB::entPhysicalDescr[100721000] = STRING: PowerSupply2
> 
[...]

PR #7625, opened 2017-11-04
feature: Entity State polling

murrant:
Display entity state on the Inventory page.
Allows for alerting based on states.

image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixes: #7621


laf:
So this isn't working for me, the sensors are added and then removed immediately.

However, is this not better to be added as yaml as it seems supported and then we can have a similar flag like rfc1628_compat in the os definition to load entity-sensor for OS we need it on?

murrant:
Yeah, the valid was the wrong array or something.  This was just an exploration to see if we wanted this.

laf:
I definitely say we should have it :)

shimamizu:
screen shot 2017-11-13 at 4 34 10 pm
This looks promising :) Example on an Arista 7508 chassis.

shimamizu:
Let me know what a simple alert rule would be to alert on say if PWR-3KT-AC-RED were to go down (aka if Power Supply in Slot 1 randomly died) based on where you're storing this in the database.

includes/polling/entity-state.inc.php
```diff
@@ -0,0 +1,106 @@
...
+                $state_data[$index]['entPhysical_id'] = $id;
+                $state_data[$index]['entStateLastChanged'] = $state['entStateLastChanged'];
+                d_echo("Inserting entity state:: ", '+');
+                d_echo($state);
+            }
+        }
+    }
+
+    if (!empty($state_data)) {
+        var_dump($state_data);
```
laf: Debug data left in.

laf:
One big issue :(

image

Ignore the earlier data but the clean data from this PR is from 1pm onwards until it drops when I switch back from master. A change from 250 seconds to 50 seconds. I fear this will kill some devices :(

murrant:
@laf yeah, any ideas on how to make that faster?  Like I said earlier, there is a lot of data here.
5 points of data for every piece of equipment in the entitymib (which includes ports usually)

laf:
> @laf yeah, any ideas on how to make that faster? Like I said earlier, there is a lot of data here.
> 5 points of data for every piece of equipment in the entitymib (which includes ports usually)

Just do it how the current sensors table does it, multi gets until we've iterated through all the data. That means we need to either store the OIDs or be able to build them from the data we do store. If it's the same 5 OIDs with the index then you probably store that now.

murrant:
I would think multi-getting the same data as the walk would be less efficient.

I guess there needs to be a way to disable the states for some entities?

laf:
Some options:

  - walk entStateLastChanged, compare to the DB and for those that have changed, multi get the data you want.

  - do like we do for selected ports polling and walk entStateOper, if it's enabled multi get the data you want.

Also, is it worth storing the _prev values or are you using the StateLastChanged value to alert on or something else?

laf:
For some reason this has stopped working for me. It's probably my local install so I'll keep digging.

murrant:
Not sure about alerting.  I need to come up with an alert rule or two.  I'm not sure what a failed state looks like as I don't want to make my production test device fail ;)

kkrumm1:
any word on this?

murrant:
Ha sent gotten back to it.  Need to move existing code to discovery and write a more efficient poller.

murrant:
@laf see if that is better for polling time

laf:
It is, it still increases polling time from 35/40 seconds to 85 seconds.

I'd definitely be concerned about this for people though. That's enough to tip pollers over the 5 minute mark. What about setting this as disabled by default?

murrant:
40s is a lot better that 200s.  Perhaps disabled by default is a good thing.

laf:
> 40s is a lot better that 200s. Perhaps disabled by default is a good thing.

It is yes but it's (imho) too high for a default enable.

PR #7626, opened 2017-11-04
Script to collect port polling data and compare full walk vs selectiv…

murrant:
…e port polling.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Works for me. Some comments:

Add support for specifying hostnames / wildcards?

Show what the values are in (seconds) for the time column.

Change color in difference column to show if the change would be a bonus (or indicate that a - in that column is a good thing).

Flag to auto enable selected port polling for devices it's worth it for?

Added to performance docs.

I know it's a lot to ask ^^ but this quickly shows where you can get some good savings.

scripts/collect-port-polling.php
```diff
@@ -0,0 +1,107 @@
...
+foreach ($devices as $index => $device) {
+    $count = dbFetchCell('SELECT COUNT(*) FROM `ports` WHERE `device_id`=?', array($device['device_id']));
+    $inactive = dbFetchCell(
+        "SELECT COUNT(*) FROM `ports` WHERE `device_id`=? AND ($inactive_sql)",
+        array($device['device_id'])
+    );
+
+    $devices[$index]['port_count'] = $count;
+    $devices[$index]['inactive_ratio'] = ($inactive == 0 ? 0 : ($inactive / $count));
+    $devices[$index]['difference'] = $device['part_time'] - $device['full_time'];
```
Zmegolaz: `part_time` should be `selective_time`

Zmegolaz:
Very nice script, I like it. I too have a feature request, to add a column with the difference in percent in addition to just seconds. I can help implement the features if you merge it, or help me on how commit to this PR. :P

laf:
You should be able to either push to murrants branch if you add his repo as a remote or actually just click edit on the file here and edit it directly.

murrant:
I'm glad you guys are excited :) I just threw this together as a quick way to get data so we could make a better decision on automatically enabling per-port polling.

@Zmegolaz 
```
git remote add murrant git@github.com:murrant/librenms.git
git fetch murrant
git checkout --track murrant/collect-port-polling 
```
Off the top of my head, so might be slightly wrong :D

Zmegolaz:
I managed! I think I'm starting to understand git now.

I added all the features in our comments, but I'm not very happy with all the hoops I had to jump through to get the printf work correctly with colors. There are also docs and CLI help to be added.

But I'll have to leave that for now, will work on it more on Sunday or Monday. I have a party to plan and host. :)

Zmegolaz:
Script done, docs left. Any comments?

laf:
Will take a look as soon as I can.

Thanks for doing the updates :)

murrant:
Hmm, I see thoughts on including hostname?  Since we are now intending people use this to set up per-port polling should we include hostname in the output?

I originally excluded it because I just wanted people to send us the output so we could make an informed decision about enabling per-port polling automatically.

murrant:
I'm not sure one pass is enough to make a good decision about enabling/disabling per-port polling.

Possibly rename the script too.

laf:
I think we can definitely allow people to make more use of this with the extra features. Trying to use it to gauge if we should turn it on globally may not give us enough data to make that decision.

Would be better off for us to start tracking some of the data in polling so we can record max time for a full port walk vs selected port polling and switch from one to the other as we need to maintain performance.

murrant:
Perhaps store this data on stats?  With either an opt-out or opt-in option for that.

laf:
> Perhaps store this data on stats? With either an opt-out or opt-in option for that.

I was more meaning locally. So that the local install can make the decisions (once support is added) and then make adjustments.

Zmegolaz:
Yeah, let's merge this, we can always try to figure out the best way forward later. This is a good first step.

PR #7628, opened 2017-11-05
webui: Improve Services Page Look

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Before:
before_r
After:
after_r

### DIFF
--- a/html/css/styles.css
+++ b/html/css/styles.css
@@ -2180,3 +2180,11 @@ label {
 .severity-unknown {
   color: darkgrey;
 }
+
+.service-status {
+    width:10px;
+    height:50px;
+    display: block;
+    margin: 0;
+    border-radius: 0;
+}
--- a/html/pages/services.inc.php
+++ b/html/pages/services.inc.php
@@ -6,33 +6,35 @@ $pagetitle[] = 'Services';
 <div class="container-fluid">
   <div class="row">
 <?php
-
 print_optionbar_start();
-
 require_once 'includes/modal/new_service.inc.php';
+
 require_once 'includes/modal/delete_service.inc.php';
 
 echo "<span style='font-weight: bold;'>Services</span> &#187; ";
-
 $menu_options = array(
-    'basic'   => 'Basic',
+    'basic' => 'Basic',
 );
+
 if (!$vars['view']) {
     $vars['view'] = 'basic';
 }
 
 $status_options = array(
-    'all'       => 'All',
-    'ok'        => 'Ok',
-    'warning'   => 'Warning',
-    'critical'  => 'Critical',
+    'all' => 'All',
+    'ok' => 'Ok',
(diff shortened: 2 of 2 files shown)


laf:
I like the tidying up of things, however the status check boxes now go away from how the devices page is - imho we should remain consistent as much as possible throughout the webui.

FTBZ:
I've a lot of services, I think that a small refactor to save space will be nice. Now this is a long list to scroll. This PR is not affecting the space, only the readability?

florianbeer:
Actually, I don't really think this does anything to tidy up things or make it more readable. It only moves the status color from the service name background to a strip on the left side and shifts around the other fields (e.g. description to the right, why?). Having the edit/delete buttons on top of each other also looks a bit weird to me, to be honest. At least we should have a bit of padding in between. The width of the "Last Changed" column was intentionally chosen to accommodate everything in one line.

Here's a screenshot of how the current services page looks on my installation:
screen shot 2017-11-06 at 21 10 00

FTBZ:
Didn't notice that the description column was gone. This is important for me too like @florianbeer. This is where I add the detail about the test because sometimes you're using the same plugin for different tests. Like the http for checking https and certificat.

Rosiak:
I believe the screenshots I included was bad, as it was taken on a rather small screen which could cause some confusion. Also I did include the descr column, I just forgot to name it.

Please take a look:
screen shot 2017-11-07 at 08 06 07
screen shot 2017-11-07 at 08 05 37

FTBZ:
Nice shot! Can the green label be reduced to keeping it on one line and save space on the screen?

laf:
I still don't think we should be moving the colour to the side, it's different from the devices page and we really should have some consistency.

Rosiak:
@laf 
Perfectly understand that, I could do the same to the other pages if people think it's OK.
Otherwise I'll just close this PR.

laf:
Well in all honesty and I don't know if this is the same still but that's how obse.... device page looks (or at least similar)

Rosiak:
Ok.
I dunno how obs.. looks, as I got the inspiration from Incinga2.
I guess if people are not up for it, I'll close it.

laf:
It does look nicer, but we would need all tables that do it the other way to be updated.

Rosiak:
So, should I proceed with updating the other ones or just close this PR?

laf:
I'm happy for the other ones to be updated. Just not happy about the disjointed look if we merged this one PR :)

FTBZ:
I think we should start somewhere, no? The padding can be reduce IMHO.

laf:
imho no, if we merged the alerts PR we would have 3 totally separate looks for tables like that. Sometimes our webui is the worst part of the software because it's changed by many people to look how they want, we need some kind of standards / consistency.

PR #7629, opened 2017-11-05
fix: Do not add invalid Cisco senors seen in IOS 15.6(3)M1

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Cisco introduced invalid sensors into IOS 15.6(3)M1, this fix makes sure we do not add these sensors.

### DIFF
--- a/includes/discovery/sensors/state/cisco.inc.php
+++ b/includes/discovery/sensors/state/cisco.inc.php
@@ -28,7 +28,6 @@ $tables = array(
 foreach ($tables as $tablevalue) {
     $temp = snmpwalk_cache_multi_oid($device, $tablevalue[0], array(), $tablevalue[4]);
     $cur_oid = $tablevalue[1];
-    $process_sensor = true;
 
     if (is_array($temp)) {
         if ($temp[0][$tablevalue[2]] == 'nonRedundant' || $temp[0]['cswMaxSwitchNum'] == '1') {
@@ -137,9 +136,8 @@ foreach ($tables as $tablevalue) {
 
         foreach ($temp as $index => $entry) {
             if ($tablevalue[2] == 'ciscoEnvMonTemperatureState' && (empty($temp[$index][$tablevalue[3]]))) {
-                $process_sensor = false;
-            }
-            if ($process_sensor == true) {
+                d_echo('Invalid sensor, skipping..');
+            } else {
                 //Discover Sensors
                 $descr = ucwords($temp[$index][$tablevalue[3]]);
                 if ($state_name == 'cRFStatusUnitState' || $state_name == 'cRFStatusPeerUnitState' || $state_name == 'cRFCfgRedundancyOperMode' || $state_name == 'cswRingRedundant') {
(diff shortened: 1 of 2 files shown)


includes/discovery/sensors/state/cisco.inc.php
```diff
@@ -135,26 +136,31 @@
...
-                $descr = $tablevalue[3] . $swrolenumber;
-            } elseif ($state_name == 'cswSwitchState') {
-                $swstatenumber++;
-                $descr = $tablevalue[3] . $swstatenumber;
-            } elseif ($state_name == 'cswStackPortOperStatus') {
-                $stack_port_descr = get_port_by_index_cache($device, $index);
-                $descr = $tablevalue[3] . $stack_port_descr['ifDescr'];
-            } elseif ($state_name == 'cefcFRUPowerOperStatus') {
-                $descr = snmp_get($device, 'entPhysicalName.'.$index, '-Oqv', 'ENTITY-MIB');
+            if ($tablevalue[2] == 'ciscoEnvMonTemperatureState' && (is_null($temp[$index][$tablevalue[3]]) or empty($temp[$index][$tablevalue[3]]))) {
```
laf: `or` should be a `||` to be consistent.

Is `empty()` not enough here or is `is_null()` needed as well?

includes/discovery/sensors/state/cisco.inc.php
```diff
@@ -135,26 +136,31 @@
...
+                $process_sensor = false;
             }
-            discover_sensor($valid['sensor'], 'state', $device, $cur_oid.$index, $index, $state_name, $descr, '1', '1', null, null, null, null, $temp[$index][$tablevalue[2]], 'snmp', $index);
+            if ($process_sensor == true) {
```
laf: Why not just leave this to the previous `if()` check that I commented on, seems pointless setting a variable.

includes/discovery/sensors/state/cisco.inc.php
murrant: Yeah, just empty should be fine: http://php.net/manual/en/types.comparisons.php

includes/discovery/sensors/state/cisco.inc.php
Rosiak: Not sure what you mean here.
What are you suggesting?

laf: Litterally just change it from:

```php
}
if ($process_sensor == true) {
```

To:

```php
} else {
```

No point in assigning a variable in one if statement only to open a new one using that variable.

laf:
Does an empty description definitely mean bad sensor?

includes/discovery/sensors/state/cisco.inc.php
Rosiak: Now I understand what you mean, you are totally right.

Rosiak:
@laf I won't say I'm 100% sure, but I've only seen it empty in this software version so far.

PR #7633, opened 2017-11-05
Adding manage_bills.php script to scripts directory

rucarrol:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/scripts/manage_bills.php
+++ b/scripts/manage_bills.php
@@ -28,7 +28,7 @@ function list_bills($bill_name)
 }
 
 // This will get an array of devices we are interested in from the CLI glob
-function get_devices($host_glob,$nameType)
+function get_devices($host_glob, $nameType)
 {
     return dbFetchRows("SELECT `device_id`,`hostname`,`sysName` FROM `devices` WHERE `".$nameType."` LIKE ?", array("%$host_glob%"));
 }
@@ -79,19 +79,18 @@ function create_bill($devs, $intf_glob, $id)
 
 $options = getopt('l:c:s:h:f');
 
-if (empty($options['s']) && empty($options['h'])) { 
+if (empty($options['s']) && empty($options['h'])) {
     echo "Please set -s or -h\n";
     exit(1);
-}
-else if (!empty($options['s']) && !empty($options['h'])) { 
+} else if (!empty($options['s']) && !empty($options['h'])) {
     echo "Please set either -s or -h, not both\n";
     exit(1);
 }
-if (!empty($options['s'])) { 
+if (!empty($options['s'])) {
     $host_glob = str_replace('*', '%', mres($options['s']));
     $nameType = "sysName";
 }
-if (!empty($options['h'])) { 
+if (!empty($options['h'])) {
     $host_glob = str_replace('*', '%', mres($options['h']));
     $nameType = "hostname";
 }
@@ -143,4 +142,3 @@ if ($flush) {
 }
 
 $ret = create_bill($devices, $intf_glob, $id);
-
--- a/scripts/manage_bills.php
+++ b/scripts/manage_bills.php
@@ -16,9 +16,10 @@ require realpath(__DIR__ . '/..') . '/includes/init.php';
 // Find the correct bill, exit if we get anything other than 1 result.
 function list_bills($bill_name)
 {
-    $bill = dbFetchRows("SELECT `bill_id`,`bill_name` FROM `bills` WHERE `bill_name` LIKE ?", array("%$bill_name%"));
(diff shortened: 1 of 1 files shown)


laf:
Thanks for this.

I've made a few changes with a git diff if you want to take a look: https://p.libren.ms/view/397b1866

Any reason you don't check hostname rather than sysName?

Can you reverse the default of `-r` as it could lead to confusion if it wipes out existing bills.

rucarrol:
> I've made a few changes with a git diff if you want to take a look: https://p.libren.ms/view/397b1866

Nice, thank you! 

> Any reason you don't check hostname rather than sysName?

Mainly as the set of devices I'm working on have a `hostname` which is not the same as `sysName`. I have no real opinion either way, but my reasoning was mostly that sysName would be what the device things it is, and `hostname` would be what DNS thinks it is (Which could change). Happy to pick `hostname` if that's the established way of doing it. 

> Can you reverse the default of -r as it could lead to confusion if it wipes out existing bills.

Sure - should I change the name to flush, so it becomes a bit more obvious?

laf:
> Mainly as the set of devices I'm working on have a hostname which is not the same as sysName. I have no real opinion either way, but my reasoning was mostly that sysName would be what the device things it is, and hostname would be what DNS thinks it is (Which could change). Happy to pick hostname if that's the established way of doing it.

Hostname would be better as it's unique, sysName isn't always depending on the devices :(

You could have two flags so people can pass either hostname or sysname?

> Sure - should I change the name to flush, so it becomes a bit more obvious?

Flush seems good.

rucarrol:
> You could have two flags so people can pass either hostname or sysname?

Will do, tomorrow. 

> Flush seems good.

Please see [adfa1e965ac3e928b69411c6ae7f15ebf671957ahttps://github.com/librenms/librenms/pull/7633/commits/adfa1e965ac3e928b69411c6ae7f15ebf671957a).

laf:
Looks good to me :)

laf:
@rucarrol Any joy on the last few changes?

rucarrol:
Hey @laf , just committed the diff :)

scripts/manage_bills.php
```diff
@@ -0,0 +1,144 @@
...
+if (!empty($options['h'])) {
+    $host_glob = str_replace('*', '%', mres($options['h']));
+    $nameType = "hostname";
+}
+
+
+$bill_name = str_replace('*', '%', mres($options['l']));
+$intf_glob = str_replace('*', '%', mres($options['c']));
+
+if (empty($bill_name)) {
```
laf: This should also check that no -h and -s have been specified so that running the script with no args will give you this help.

scripts/manage_bills.php
```diff
@@ -0,0 +1,144 @@
...
+    echo "Usage:\n";
+    echo "-l <bill name glob>   Bill name to match\n";
```
laf: `-b` instead maybe for bill name?

scripts/manage_bills.php
```diff
@@ -0,0 +1,144 @@
...
+    echo "-s <sysName glob>     sysName to match (Cannot be used with -h)\n";
+    echo "-h <hostname glob>    Hostname to match (Cannot be used with -s)\n";
+    echo "-c <Interface description glob>   Interface description to match\n";
```
laf: `-i` for interface description maybe?

scripts/manage_bills.php
```diff
@@ -0,0 +1,144 @@
...
+    echo "If I wanted to add all interfaces containing the description Telia to a bill called 'My Lovely Transit Provider'\n";
+    echo "php manage_bills.php -l 'My Lovely Transit Provider' -d all -c Telia";
+    echo "\n";
+    exit;
+}
+
+if ($bill_name == 'all') {
+    $bill_name = '%';
+}
+if ($intf_glob == 'all') {
```
laf: Seeing as you can specify all interfaces I think we should bail the script if no `-c` (or `-i`) is set otherwise interfaces are added again

scripts/manage_bills.php
```diff
@@ -0,0 +1,144 @@
...
+
+// This will flush bill ports if -r is set on cli
+function flush_bill($id)
+{
+    echo("Removing ports from bill ID $id\n");
+    return dbDelete('bill_ports', '`bill_id` = ?', array($id));
+}
+
+
+function create_bill($devs, $intf_glob, $id)
```
laf: I'd rename this to `add_ports` so it clarifies what it does.

rucarrol:
Hey @laf 

Code review was much appreciated, thank you for taking the time to do this.

I changed the flow logic around a bit for detecting -h/-s, so that it'll print warnings, then help. 

Not setting -i will also cause the script to bail out. 

I also renamed the create_bill function to add_ports_to_bill (So I would not clobber the add_ports functioning already existing). 

Cheers + thanks against for assisting.

PR #7646, opened 2017-11-06
newdevice : Stormshield

tomarch:
Initial support for Stormshiel firewall:

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/stormshield.inc.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+use LibreNMS\RRD\RrdDefinition;
```
laf: You don't need this line :)

You should also add a copyright.

includes/polling/os/stormshield.inc.php
```diff
@@ -0,0 +1,8 @@
...
+
+$hardware = trim(snmp_get($device, '.1.3.6.1.4.1.11256.1.0.1.0', '-Osqv'), '"');
```
laf: Can you not use the MIB here to use named OIDs?

laf: You should also update all of these to a single `snmp_get_multi_oid` to avoid multiple lookups

mibs/stormshield/STORMSHIELD-ALARM-MIB.txt
```diff
@@ -0,0 +1,282 @@
+STORMSHIELD-ALARM-MIB DEFINITIONS ::= BEGIN
```
laf: Please drop the `.txt` from the filename of all mibs.

tests/snmpsim/stormshield.snmprec
```diff
@@ -0,0 +1,3 @@
+1.3.6.1.2.1.1.1.0|4|NS-BSD
```
laf: Is that 100% the entire sysDescr?

includes/definitions/stormshield.yaml
```diff
@@ -0,0 +1,16 @@
+os: stormshield
```
laf: The company name isn't usually the best name to use for the OS. Does the firmware / software have a name you can use? If not, the product genre should be used.

tomarch:
Thanks for the review.
I think i've made all the needed changes.

includes/polling/os/ns-bsd.inc.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2017 Thomas GAGNIERE
+ * @author     Thomas GAGNIERE <tgagniere@reseau-concept.com>
+ */
+
+$data = snmp_get_multi_oid($device, 'snsModel.0 snsVersion.0 snsSerialNumber.0 snsSystemName.0', '-OUQs', 'STORMSHIELD-PROPERTY-MIB');
+
+$hardware = trim($data['snsModel.0'], '"');
```
laf: Is trim needed here as we do it within `snmp_get_multi_oid()`

PR #7655, opened 2017-11-06
Add OS detection and temperature sensor support for Asentria SiteBoss

thecityofguanyu:
…Initial commit. OS definition/polling, temp sensor discovery/polling, MIBs, logo, snmprec test

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/siteboss.yaml
```diff
@@ -0,0 +1,12 @@
+os: siteboss
```
laf: Is this the OS /firmware name or does it not have one and so the product name is used?

includes/definitions/siteboss.yaml
```diff
@@ -0,0 +1,12 @@
...
+type: server
+icon: siteboss
+over:
+    - { graph: device_bits, text: 'Device Traffic' }
+    - { graph: device_processor, text: 'CPU Usage' }
+    - { graph: device_mempool, text: 'Memory Usage' }
+mib_dir: asentria
+discovery:
+    - sysObjectId:
+        - .1.3.6.1.4.1.3052.12.3.2
```
laf: Is this worth changing to `.1.3.6.1.4.1.3052.12.3.` or `.1.3.6.1.4.1.3052.12.`. Our stats show no other devices on either of these.

includes/discovery/sensors/temperature/siteboss.inc.php
```diff
@@ -0,0 +1,34 @@
...
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+echo 'Asentria SiteBoss ';
+
+$tempoids = snmp_walk($device, 'esIndexPC', '-Osqn', 'SITEBOSS-530-STD-MIB');
```
laf: Can this not be converted to yaml?

includes/discovery/sensors/temperature/siteboss.inc.php
```diff
@@ -0,0 +1,34 @@
...
+foreach (explode("\n", $tempoids) as $data) {
+    $data = trim($data);
+    $data = explode(" ", $data);
+    if ($data[1] == '1') {
+        $oid = str_replace('.1.3.6.1.4.1.3052.12.1.1.1.1.2.', '', $data[0]);
+        $temperature_oid = ".1.3.6.1.4.1.3052.12.1.1.1.1.6.$oid";
+        $descr_oid       = ".1.3.6.1.4.1.3052.12.1.1.1.1.4.$oid";
+        $descr           = snmp_get($device, $descr_oid, '-Oqv', '');
+        $temperature     = str_replace('"', '', snmp_get($device, $temperature_oid, '-Oqv', ''));
+        $temperature     = fahrenheit_to_celsius($temperature);
```
laf: You can actually just register this as a callback in the `discover_sensor()` function. See includes/discovery/sensors/temperature/geist-watchdog.inc.php

includes/definitions/siteboss.yaml
thecityofguanyu: The latter. Did not have one (that I could find) and so the product name was used.

includes/discovery/sensors/temperature/siteboss.inc.php
thecityofguanyu: I looked at the YAML sensor docs and certainly considered it, but wasn't as confident in my ability to make it work as I was with the PHP files.

includes/definitions/siteboss.yaml
thecityofguanyu: I'll look through some MIBs and see if I can find any obvious reason not to.

includes/discovery/sensors/temperature/siteboss.inc.php
thecityofguanyu: @laf I guess I don't fully understand what you mean there. I'm probably missing something here, but I don't *see* a function callback in `discover_sensor()` in that geist file.

Do you mean something to this effect?
```php
discover_sensor($valid['sensor'], 'temperature', $device, $temperature_oid, $oid, 'siteboss', $descr, '1', '1', null, null, null, null, fahrenheit_to_celsius($temperature));
```

includes/discovery/sensors/temperature/siteboss.inc.php
```diff
@@ -0,0 +1,32 @@
...
+    $data = trim($data);
+    $data = explode(" ", $data);
+    if ($data[1] == '1') {
+        $oid = str_replace('.1.3.6.1.4.1.3052.12.1.1.1.1.2.', '', $data[0]);
+        $temperature_oid = ".1.3.6.1.4.1.3052.12.1.1.1.1.6.$oid";
+        $descr_oid       = ".1.3.6.1.4.1.3052.12.1.1.1.1.4.$oid";
+        $descr           = snmp_get($device, $descr_oid, '-Oqv', '');
+        $temperature     = str_replace('"', '', snmp_get($device, $temperature_oid, '-Oqv', ''));
+
+        discover_sensor($valid['sensor'], 'temperature', $device, $temperature_oid, $oid, 'siteboss', $descr, '1', '1', null, null, null, null, fahrenheit_to_celsius($temperature));
```
laf: should be:

`discover_sensor($valid['sensor'], 'temperature', $device, $temperature_oid, $oid, 'siteboss', $descr, '1', '1', null, null, null, null, 'fahrenheit_to_celsius');`

includes/polling/sensors/temperature/siteboss.inc.php
```diff
@@ -0,0 +1,15 @@
...
+ * © 2017 Chris A. Evans <thecityofguanyu@outlook.com>
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+$sensor_value = fahrenheit_to_celsius($sensor_value);
```
laf: With the register of the callback this file shouldn't be needed.

thecityofguanyu:
@laf By all means. See paste linked below for SNMPwalk output:

https://p.libren.ms/view/d9f6db7d

The reason I'm unsure of how to make it work as YAML is because of the conditional that I included in the PHP here:
```php
 if ($data[1] == '1') {
```

In that snippet,  `$data[1]` is value content of an SNMPwalk over `esIndexPC`. That value has to be equal `1` for it to be a temperature sensor. If you just took all `esIndexPC` values without the conditional, then you would end up with non-temperature indexes that you don't want to consider.

laf:
> @laf By all means. See paste linked below for SNMPwalk output:

Needs to be with `.` rather than an OID as I need the whole thing to simulate the device.

thecityofguanyu:
@laf Sorry. See below:

https://p.libren.ms/view/fee7497f

laf:
@thecityofguanyu I've pushed to your branch so pull in the changes and test.

@murrant This includes core changes to sensor discovery - thoughts?

includes/definitions/discovery/siteboss.yaml
```diff
@@ -0,0 +1,15 @@
...
+modules:
+    sensors:
+        temperature:
+            data:
+                -
+                    oid: esPointTable
+                    value: esPointValueInt
+                    num_oid: .1.3.6.1.4.1.3052.12.1.1.1.1.6.
+                    descr: esPointName
+                    skip_values:
```
murrant: Hmm, I thought about having the format like this:
`{{ $esIndexPC }} != 1`  and parsing it out, but I'm not sure which is better.

includes/discovery/functions.inc.php
```diff
@@ -916,12 +916,22 @@ function ignore_storage($os, $descr)
  * @param $value
  * @param $data
  * @param $group
+ * @param null $index
+ * @param array $pre_cache
  * @return bool
  */
-function can_skip_sensor($value, $data, $group)
+function can_skip_sensor($value, $data, $group, $index = null, $pre_cache = array())
```
murrant: Don't pass the index, just pass the data for this one sensor (as in call the function with `$pre_cache[$index]`)

includes/definitions/discovery/siteboss.yaml
laf: Yeah I saw your comment in the other PR but that relies on good formatting and a good parser. I chose to make things specific which should lead to less issues with a negligible increase in yaml size.

includes/discovery/functions.inc.php
laf: Easy enough to do. Will update if we agree this is the way forward.

thecityofguanyu:
@laf Just pulled down your changes. All tests good for me.

includes/discovery/functions.inc.php
laf: Updated.

laf: Actually updated :/

laf:
Should be good to go now @librenms/reviewers

thecityofguanyu:
@laf 

This does not appear to convert F values to celsius after your YAML conversions? I must have not tested the pull closely enough, so my bad there.

From discovery:
```shell
Discover sensor: .1.3.6.1.4.1.3052.12.1.1.1.1.6.16.1.1, 16.1.1, siteboss, Sperry Remote, snmp, 1, 1, , 71
SQL[SELECT COUNT(sensor_id) FROM `sensors` WHERE `poller_type`= 'snmp' AND `sensor_class` = 'temperature' AND `device_id` = '283' AND sensor_type = 'siteboss' AND `sensor_index` = '16.1.1'] 
SQL[SELECT * FROM `sensors` WHERE `sensor_class` = 'temperature' AND `device_id` = '283' AND `sensor_type` = 'siteboss' AND `sensor_index` = '16.1.1'] 
.Final sensor value: 0
Final sensor value: 0
Array
(
    [siteboss] => Array
        (
            [16.1.1] => 1
        )

)
```

From poller:
```shell
Modules status: Global+ OS  Device  
#### Load poller module sensors ####
SQL[SELECT `sensor_class` FROM `sensors` WHERE `device_id` = '283' GROUP BY `sensor_class`] 
SQL[SELECT * FROM `sensors` WHERE `sensor_class` = 'temperature' AND `device_id` = '283'] 
SNMP[/usr/bin/snmpget -v1 -c 'public' -OUQnt -M /opt/librenms/mibs:/opt/librenms/mibs/asentria udp:asentria_spe00:161 .1.3.6.1.4.1.3052.12.1.1.1.1.6.16.1.1]
.1.3.6.1.4.1.3052.12.1.1.1.1.6.16.1.1 = 70

Checking (snmp) temperature Sperry Remote... 
70 
RRD[update asentria_spe00/sensor-temperature-siteboss-16.1.1.rrd N:70 --daemon unix:/var/run/rrdcached.sock]
RRDtool Output: OK u:0.00 s:0.00 r:9.13
Sending asentria_spe00.sensor.sensor.temperature.siteboss.16_1_1.sensor 70 1510688002
SQL[UPDATE `sensors` set `sensor_current` ='70',`sensor_prev` ='69',`lastupdate` =NOW() WHERE `sensor_class` = 'Temperature' AND `sensor_id` = '4161'] 
```

Issue #7391, opened 2017-09-26
Check Point Appliance 1430 Device Support


rebortg:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
Component | Version
--------- | -------
LibreNMS  | 1.31.03-69-g0713963
DB Schema | 207
PHP       | 7.0.22-0ubuntu0.16.04.1
MySQL     | 10.0.31-MariaDB-0ubuntu0.16.04.2
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3

- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
https://p.libren.ms/view/7b05aee2
https://p.libren.ms/view/ecdd9c2b
https://p.libren.ms/view/caa38104

- [x] Please provide as much detail as possible.
Check Point Small Business Appliance 1430
This is Check Point's 1430 Appliance R77.20.60 - Build 654
https://www.checkpoint.com/products/1400-security-appliances/

- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

PR #7656, opened 2017-11-06
newdevice: Updated Checkpoint Gaia detection and added storage

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7391


laf:
I've updated the index column in storage to match sensors. Not all devices return a single int for indexes :(

includes/definitions/gaia.yaml
```diff
@@ -3,10 +3,17 @@ text: 'Check Point GAiA'
...
+mib_dir:
+    - checkpoint
 over:
     - { graph: device_bits, text: 'Device Traffic' }
     - { graph: device_processor, text: 'Processor Usage' }
     - { graph: device_mempool, text: 'Memory Usage' }
 discovery:
     - sysObjectId:
         - .1.3.6.1.4.1.2620.1.6.123.1
+    - snmpget:
```
murrant: Just FYI, this snmp get will be queried for every OS after this one (alphabetically)

It would be great if you could run a check before any snmpget to narrow it down a bit.

laf: Why will it be run for every os after this? Surely if it makes it this far it will just do it once then move on to the next OS?

I don't see any other way of identifying this OS though, it's using standard Linux sysDescr and sysObjectID :(

murrant: No, it only does it once, but I'm saying if the target system is anything past G alphabetically, this snmp_get will be run.

In the snmprec file, you have sysObjectId and sysDescr, why not use those to only have the get run on linux systems?

laf: Can you even do that, I thought if you said check those two + this snmpget it will run the snmpget or will it return false once the first check against say sysObjectID fails if it's not Linux?

includes/definitions/gaia.yaml
```diff
@@ -3,10 +3,17 @@ text: 'Check Point GAiA'
...
+        oid: CHECKPOINT-MIB::osName.0
+        mibdir: checkpoint
```
murrant: mibdir -> mib_dir

murrant:
@laf 
```
discovery:
    - sysObjectId:
        - .1.3.6.1.4.1.2620.1.6.123.1
    -
      sysObjectId: .1.3.6.1.4.1.8072.3.2.10
      snmpget:
        oid: CHECKPOINT-MIB::osName.0
        mib_dir: checkpoint
        op: 'contains'
        value: Gaia
```

Here's what happens for gaia_1430.snmprec:
1. Check sysObjectId against .1.3.6.1.4.1.2620.1.6.123.1 **fail**
2. Check sysObjectId against .1.3.6.1.4.1.8072.3.2.10 **success**
3. Check snmpget **success**

The first level in the array is OR.  The second level is AND.  The third level is OR.

laf:
Updated and pushed - cheers.

PR #7661, opened 2017-11-07
refactor: Switch to using discovery-wrapper.py by default

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Seems sensible to run like this now. Won't effect new installs.

### DIFF
--- a/doc/Extensions/Auto-Discovery.md
+++ b/doc/Extensions/Auto-Discovery.md
@@ -6,7 +6,7 @@ source: Extensions/Auto-Discovery.md
 LibreNMS provides the ability to automatically add devices on your network, we can do this via
 a few methods which will be explained below and also indicate if they are enabled by default.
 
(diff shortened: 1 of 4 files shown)


murrant:
Great, but I think the default thread count should be 2 or 4 instead of 1.

laf:
1 means it’s a direct replacement for the php version and should guarantee that we don’t overload installs, changing it to another value won’t guarantee that :(

Zmegolaz:
I've been running this with 4 threads for a day now and tried adding/removing devices, seems to work.

But this will _only_ affect new installs, won't it? In that case nothing will break for existing users if it's set to more threads.

laf:
> But this will only affect new installs, won't it? In that case nothing will break for existing users if it's set to more threads.

Correct but we've had no feedback so imho, I'd prefer to start out with how it is now and as people start reporting positive changes then we start increasing the limit.

I actually plan to remove the thread argument and allow people to just set in config.

PR #7666, opened 2017-11-07
fix: errors with missing posix extension

murrant:
It should be an optional dependency

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Validations/Php.php
+++ b/LibreNMS/Validations/Php.php
@@ -63,6 +63,13 @@ class Php implements ValidationGroup
                 $validator->fail("Missing PHP extension: $extension", "Please install $extension");
             }
         }
+
+        $suggested_extensions = array('posix' => 'php-process');
+        foreach ($suggested_extensions as $extension => $packages) {
+            if (!extension_loaded($extension)) {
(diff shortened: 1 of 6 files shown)


LibreNMS/Validations/Php.php
```diff
@@ -63,6 +63,13 @@ private function checkExtensions(Validator $validator)
                 $validator->fail("Missing PHP extension: $extension", "Please install $extension");
             }
         }
+
+        $suggested_extensions = array('posix' => 'php-process');
+        foreach ($suggested_extensions as $extension => $packages) {
+            if (!extension_loaded($extension)) {
+                $validator->warn("Missing optional PHP extension: $extension", "It is suggested you install $packages or the one that matches your php version");
```
laf: Will this not just lead to people asking about install this and should we not just have it as part of the install docs?

murrant: Yeah, probably would be worth adding it to the install docs.  No change is necessary on Ubuntu.

laf: As in it's already installed?

murrant: As in for Debian/Ubuntu, it is part of php core.

PR #7677, opened 2017-11-08
newdevice: initial detection for Arris Cable Modem devices

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/arris-cm.yaml
+++ b/includes/definitions/arris-cm.yaml
@@ -1,0 +1,10 @@
+os: arris-cm
+text: ARRIS DOCSIS
+type: network
+icon: arris
+discovery:
+    -
+      sysObjectId:
+        - .1.3.6.1.4.1.4115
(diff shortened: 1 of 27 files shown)


includes/definitions/arris-cm.yaml
```diff
@@ -0,0 +1,10 @@
+os: arris-cm
```
laf: Definitely the right os name to use? would arris-docsis not fit better? Do they have a name for the firmware?

murrant: I started out as arris-docsis, then I saw the name of their MIB was arris-cm...  Perhaps I should include the mibs even though they are not used?

laf: Up to you, I used to tell people not to when I could see they weren't used but they can be useful later on.

Let me know what you decide or merge when you feel it's ready.

PR #7682, opened 2017-11-08
fix: fping6 validation when fping6 doesn't support IPv6

murrant:
Is the warning appropriate or should I remove it?

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Validations/Programs.php
+++ b/LibreNMS/Validations/Programs.php
@@ -40,7 +40,7 @@ class Programs implements ValidationGroup
     public function validate(Validator $validator)
     {
         // Check programs
-        $bins = array('fping', 'fping6', 'rrdtool', 'snmpwalk', 'snmpget', 'snmpbulkwalk');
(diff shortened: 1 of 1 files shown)


laf:
Is it certain versions of fping6 that don't support it and if so, maybe list those in the warning?

murrant:
It was a centos 6 package on rpmforge.  Just a bad package.  But could be others out there or self-complied fping.

These two for sure:
https://centos.pkgs.org/6/repoforge-x86_64/fping-3.9-1.el6.rf.x86_64.rpm.html
https://centos.pkgs.org/6/repoforge-x86_64/fping-3.10-1.el6.rf.x86_64.rpm.html

laf:
Leave the message for now and let's see what people say in reaction to it.

murrant:
@setiseta update the validation to show warning only when IPv6 is disabled.  Then send a pull request so we can merge it for everyone.

PR #7684, opened 2017-11-08
* Health/sensor view for specific device/ports/port

bergroth:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/device/port.inc.php
+++ b/html/pages/device/port.inc.php
@@ -76,6 +76,10 @@ $menu_options['fdb']    = 'FDB Table';
 $menu_options['events'] = 'Eventlog';
 $menu_options['notes'] = 'Notes';
 
+if (dbFetchCell("SELECT COUNT(*) FROM `sensors` WHERE `device_id` = ? AND `entPhysicalIndex` = ? ", array($device['device_id'],$port['ifIndex']))) {
+    $menu_options['sensors'] = 'Health';
+}
+
 if (dbFetchCell("SELECT COUNT(*) FROM `ports_adsl` WHERE `port_id` = '".$port['port_id']."'")) {
     $menu_options['adsl'] = 'ADSL';
 }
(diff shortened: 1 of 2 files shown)


laf:
What devices does this work for?

bergroth:
I've only tested it with Juniper.

html/pages/device/port.inc.php
```diff
@@ -76,6 +76,10 @@
 $menu_options['events'] = 'Eventlog';
 $menu_options['notes'] = 'Notes';
 
+if (dbFetchCell("SELECT COUNT(*) FROM `sensors` WHERE `device_id` = ? AND `entPhysicalIndex` = ? ", array($device['device_id'],$port['ifIndex']))) {
```
laf: I think this query also needs: `entPhysicalIndex_measured = 'ports'`

laf:
Thanks, this needs an additional column to be checked then. See my inline comment.

bergroth:
Added and still works with Juniper devices

murrant:
Hmm, this doesn't work on my cisco device.  I don't think these fields are consistently set accross all our health sensors:
| sensor_id | sensor_descr             | entPhysicalIndex | entPhysicalIndex_measured | entity_link_type | entity_link_index |
|-----------|--------------------------|------------------|---------------------------|------------------|-------------------|
|      9543 | Gi2/0/25 Transmit Power  | 2041             |                           | port             |             10625 |
|      9544 | Gi2/0/25 Receive Power   | 2042             |                           | port             |             10625 |
|      9548 | Gi2/0/26 Transmit Power  | 2047             |                           | port             |             10626 |
|      9549 | Gi2/0/26 Receive Power   | 2048             |                           | port             |             10626 |

laf:
I added those two columns in not realising we had some support, just needs updating to use the old columns

PR #7691, opened 2017-11-09
edgeos bad_if_regexp

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

added bad_if_regexp to filter out some interfaces that are always down in edgeos hardware.
tested on er 8 pro with 1.9.7 fix 4

### DIFF
--- a/includes/definitions/edgeos.yaml
+++ b/includes/definitions/edgeos.yaml
@@ -11,3 +11,8 @@ discovery:
     - sysDescr_regex:
         - '/^EdgeOS/'
         - '/^EdgeRouter/'
+bad_if_regexp:
+    - '/^npi[0-9]/'
+    - '/^loop[0-9]/'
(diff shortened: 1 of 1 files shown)


includes/definitions/edgeos.yaml
```diff
@@ -11,3 +11,8 @@ discovery:
     - sysDescr_regex:
         - '/^EdgeOS/'
         - '/^EdgeRouter/'
+bad_if_regexp:
+    - '/^npi[0-9]/'
```
laf: The others make sense but what type of port is this?

crcro: from ubnt forums: 

Those are additional devices created by the Ethernet driver (only on the ERPro-8 and ER-8) and are not used by the system, so they can be ignored.

PR #7699, opened 2017-11-09
webui: Expose pollers link to show total poll time for all devices per poller

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/print-menubar.php
+++ b/html/includes/print-menubar.php
@@ -671,11 +671,11 @@ if ($_SESSION['userlevel'] >= '10') {
            <li class="dropdown-submenu">
                <a href="#"><i class="fa fa-th-large fa-fw fa-lg" aria-hidden="true"></i> Pollers</a>
                <ul class="dropdown-menu scrollable-menu">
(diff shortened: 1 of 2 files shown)


murrant:
@laf The groups tab is exposed once you go into the pollers section.  That gave me a thought.

Instead of requiring `distributed_poller` be set on the main node, perhaps we just count the number of poller groups before exposing the poller group setting in the add and edit device dialogs?

laf:
> @laf The groups tab is exposed once you go into the pollers section. That gave me a thought.

Sorry I don't know what you mean here.

> Instead of requiring distributed_poller be set on the main node, perhaps we just count the number of poller groups before exposing the poller group setting in the add and edit device dialogs?

Up to you but an additional sql query to show the link, worth it?

murrant:
Creating poller groups is only relevant for people using distributed polling.

The sql query would be very quick, counting the rows on a table with probably single digit entries.

murrant:
After some looking at the poller-wrapper problem, I vote for 1

PR #7700, opened 2017-11-10
mib_dir consistency

murrant:
also, all the snmpget mib_dir lines were unneeded.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Travis failure might be because the mib_dir values can or are arrays.

You could edit `function mibdir()` to handle that and leave them as arrays.

PR #7703, opened 2017-11-10
Update Cambium devices to the new wireless sensors

pheinrichs:
I would like to open this issue to start the talk for cambium refactoring.
I've started a branch for concept on my thoughts on the process and what would all need to take place.

https://github.com/pheinrichs/librenms/tree/cambium-refactor

My thoughts on the process is to break down the cambium os to MIB specific groups, if the hardware uses a separate MIB create a new OS for that cambium device. A lot hardware are using separate mibs to handle similar tasks, this is where the messy code comes in currently with the large amount of conditionals.

Where now:
os Cambium = epmp(CAMBIUM-PMP80211-MIB),
pmp(WHISP-BOX-MIBV2-MIB),
ptp650(CAMBIUM-PTP650-MIB,
ptp250(CAMBIUM-PTP250-MIB)

Refactored would be
os = epmp
os = pmp
os = ptp650
os = ptp250

Doing so, would then make using the refactored wireless section @murrant created much easier and reduce the amount of clutter created by the 'cambium' monster I Created.


laf:
Much better :D

murrant:
:+1: From me!

includes/polling/os/pmp.inc.php
```diff
@@ -0,0 +1,50 @@
...
+    'OFDM'      => 'PMP 430',
+    'AP'        => 'PMP 100'
+);
+
+foreach ($models as $desc => $model) {
+    if (strstr($cambium_type, $desc)) {
+
+        $hardware = $model;
+
+        if (strstr($model, 'PTP')) {
```
murrant: There probably could be a lot of improvement in this code, but I'm willing to leave that for a separate PR.

For example, strstr() can be replaced with our  str_contains(),

pheinrichs:
Early next week i'll start adding the Wireless Sensors and hopefully get those up to par where they are at currently.

murrant:
You need to fix up the copyright/headers for some of your files.

For your wireless sensors, might I suggest one base os class and then an os class for each device os that extends that base os.  Makes for easy code sharing.

pheinrichs:
@murrant Ahh right good call. I'll update the headers. 

I'll look into doing a base OS! Thanks

includes/polling/os/ptp250.inc.php
```diff
@@ -9,7 +9,34 @@
  * option) any later version.  Please see LICENSE.txt at the top level of
  * the source code distribution for details.
 */
+use LibreNMS\RRD\RrdDefinition;
 
 $version = $poll_device['sysDescr'];
 $masterSlaveMode = ucfirst(snmp_get($device, 'masterSlaveMode.0', '-Oqv', "CAMBIUM-PTP250-MIB"));
 $hardware = 'PTP 250 '. $masterSlaveMode;
+
+$txModulation = snmp_get($device, ".1.3.6.1.4.1.17713.250.5.9.0", "-Ovqn", "");
```
murrant: I think modulation is usually a rate sensor.

includes/polling/os/pmp.inc.php
```diff
@@ -61,3 +61,192 @@
...
+    $fields = array(
+        'last' => $lastLevel,
+    );
+    $tags = compact('rrd_def');
+    data_update($device, 'canopy-generic-450-powerlevel', $tags, $fields);
+    $graphs['canopy_generic_450_powerlevel'] = true;
+    unset($lastLevel);
+}
+
+$ssr = str_replace('"', "", snmp_get($device, "linkSignalStrengthRatio.2", "-Ovqn", "WHISP-APS-MIB"));
```
murrant: Is SSR a form of SNR?

pheinrichs: Here's the desc as per the mib: 
 Signal Strength Ratio in dB is the power received by the vertical antenna input (dB) -
power received by the horizontal antenna input (dB).
MIMO radios only.

I'm not sure, it might be a bit of a stretch but could potentially work putting it in snr.

includes/polling/os/ptp250.inc.php
pheinrichs: Good call, I'll merge that in. Thanks

includes/polling/os/pmp.inc.php
```diff
@@ -43,7 +43,7 @@
 );
 
 foreach ($models as $desc => $model) {
-    if (str_contains($cambium_type, $desc)) {
+    if (strstr($PMP, $desc)) {
```
pheinrichs: Whoops, this was for debugging purposes. I'll Switch this back on the next push

pheinrichs:
@murrant I'd love to hear your opinion on the latest commit. I had to make further adjustments to make sure the hardware type is being identified correctly.

murrant:
@pheinrichs seems overly complex.  Perhaps just reverting and initializing $hardware as 'PMP-100'

includes/polling/os/pmp.inc.php
barryodonovan: It's not an SNR, it's a measure of the imbalance in the chains. It's actually calculable and useful for all dual polarity radios, and presented directly in many devices. A change in the SSR is indicative of movement of an antenna in one plane only, or of degredation in cables or antenna affecting one polarity only. Because of the opaque nature of the calculations done on multi chain radios to generate an SNR figure, a degredation in one chain only can be masked or completely hidden in the headline figure.

This highlights one of the issues I see with the new wireless sensors paradigm, enterprise grade radios have a lot of metrics which are either not relevant to or not easily measureable on, consumer WiFi products. With regard to cambium, the new Medusa radios have further metrics relating to the active phased array antenna. Similar metrics will also be present on the Radwin Jet and probably future ubiquiti and Mikrotik radios.

barryodonovan:
I would not agree with having the CMM in this OS (even though it shares a base OID). The CMM is not a wireless device, it is a power / sync injector.

pheinrichs:
@barryodonovan Yeah, I felt the same way, I gave it, it's own OS a few commits ago and just threw in a regex to check for CMM

includes/polling/os/pmp.inc.php
murrant: @barryodonovan I have no problem with adding metrics.

pheinrichs: I'll throw in a new metric!

pheinrichs:
@murrant Another thing to note, I've been removing the duplicate graphs after they've been migrated to sensors. Is this the desired approach as this would cause loss of data from the previous ones?

murrant:
The old and new rrd names will be different so, yes historical data will go away.  These devices are older so I'm fine with this but we will need to send user notification.

pheinrichs:
@murrant Will do, i'm pretty close i think. One thing i'm noticing is the 'CMM' hardware, on first snmp discovery (or just snmp walk) the sysObjectID is as expected. On second discovery (snmp walk) the sysObjectID is a completely random. Looks like a bug in hardware, but i'd love to account for this bug to avoid issues. Is it completely out of question to snmpget(.1.3.6.1.6.3.1.1.4.3.0) which seems to return the expected result without the bug of it changing?( .1.3.6.1.4.1.161.19.250.256)

On a side note, there is no real support for cmm's currently so this has no impact on legacy things.

murrant:
Yeah, we can do snmpgets in discovery, but we like to qualify them (aka check sysDescr or sysObjectID first)

murrant:
You might want to add an extra snmprec file to simulate the misbehavior so it doesn't get changed later.

pheinrichs:
@murrant I think this should be ready for some testing

murrant:
We will also need to send notification of the changes.  Put it at least a week out.

murrant:
Merge on Dec 6

PR #7704, opened 2017-11-10
feature: Added support for setting php memory_limit in config.php

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Users can now change this dynamically rather than having to reload the web service / php-fpm.

### DIFF
--- a/doc/Support/Configuration.md
+++ b/doc/Support/Configuration.md
@@ -39,6 +39,14 @@ $config['db_port']   = NULL;
 $config['db_socket'] = '/run/mysqld/mysqld.sock';
 ```
 
+### Core
+
+#### PHP Settings
+
+You can change the memory limits for php within `config.php`. The value is in Megabytes and should just be an int value:
+
+`$config['php_memory_limit'] = 128;`
+
 ### Programs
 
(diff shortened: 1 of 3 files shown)


murrant:
Seems reasonable, I would be much happier if we could just display the error to the user.

laf:
> Seems reasonable, I would be much happier if we could just display the error to the user.

Registering our own custom error handler has limitations afaik so we don't get to trap all errors that occur.

murrant:
I mean setting display errors to a different level.  To many errors would be printed at this time :/

laf:
Seems like a different discussion / PR. Regardless of showing this to the user, allowing people to change it in config.php makes it easier for users to fix.

murrant:
I was concerned we might not be able to set that in some situations, but seems ok.  http://php.net/manual/en/ini.list.php

Where you've placed this in the config it doesn't allow the setting to be set via mysql.  Ok with this?

laf:
> Where you've placed this in the init.php it doesn't allow the setting to be set via mysql. Ok with this?

I decided not to bother with the webui ability as most likely you'll want to edit this when you can't access it.

PR #7710, opened 2017-11-12
Add Permission checks to API

richardlawley:
The API functions do not contain any authorisation checks - as long as a valid API key is given, all functions are allowed, bypassing device/port/bill permissions.  Non-admin users cannot create their own key, but an admin can create a key for them.

This PR adds basic authorisation checks in.  Initially by adding fake $_SESSION variables, which mean that the standard is_admin/device_permitted/port_permitted/bill_permitted functions all work as expected.

I've also been through the API functions adding in the authorisation checks.  Anything I wasn't 100% sure about now has a is_admin requirement, as does any action function (add/edit/delete).  List devices/ports/bills respect user permissions.

Also added shortcut function for returning an error and finishing - api_error.

I've got no way to test alternate authentication mechanisms, but any auth code has been based  on includes/authentication.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -119,7 +119,7 @@ function check_is_admin()
 
 function check_is_read()
 {
-    if (!is_admin && !is_read()) {
+    if (!is_admin() && !is_read()) {
         api_error(403, 'Insufficient privileges');
     }
 }
@@ -1122,15 +1122,15 @@ function list_bills()
         $sql    .= '`bill_ref` = ?';
         $param[] = $bill_ref;
     } elseif (is_numeric($bill_id)) {
-        $sql    .= '`bills`.`bill_id` = ?';
+        $sql    .= '`bill_id` = ?';
         $param[] = $bill_id;
     }
     if (!is_admin() && !is_read()) {
-        $sql    .= ' AND `bills`.`bill_id` IN (SELECT `bill_id` FROM `bills_perms` WHERE `user_id` = ?)';
+        $sql    .= ' AND `bill_id` IN (SELECT `bill_id` FROM `bill_perms` WHERE `user_id` = ?)';
         $param[] = $_SESSION['user_id'];
     }
 
-    foreach (dbFetchRows("SELECT `bills`.*,COUNT(port_id) AS `ports_total` FROM `bills` LEFT JOIN `bill_ports` ON `bill_ports`.`bill_id`=`bills`.`bill_id` WHERE $sql GROUP BY `bill_name`,`bill_ref` ORDER BY `bill_name`", $param) as $bill) {
+    foreach (dbFetchRows("SELECT * FROM `bills` WHERE $sql ORDER BY `bill_name`", $param) as $bill) {
         $rate_data    = $bill;
         $allowed = '';
         $used = '';
@@ -1206,7 +1206,6 @@ function update_device()
 
 function get_device_groups()
 {
-    check_is_read();
     $app = \Slim\Slim::getInstance();
     $router = $app->router()->getCurrentRoute()->getParams();
     $status   = 'error';
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -135,6 +135,7 @@ function check_not_demo()
 function get_graph_by_port_hostname()
 {
     // This will return a graph for a given port by the ifName
+    global $config;
     $app          = \Slim\Slim::getInstance();
     $router       = $app->router()->getCurrentRoute()->getParams();
     $hostname     = $router['hostname'];
@@ -207,6 +208,7 @@ function get_port_stats_by_port_hostname()
 function get_graph_generic_by_hostname()
 {
     // This will return a graph type given a device id.
+    global $config;
     $app          = \Slim\Slim::getInstance();
     $router       = $app->router()->getCurrentRoute()->getParams();
     $hostname     = $router['hostname'];
@@ -531,6 +533,7 @@ function list_ospf()
 function get_graph_by_portgroup()
 {
     check_is_read();
+    global $config;
     $app    = \Slim\Slim::getInstance();
     $router = $app->router()->getCurrentRoute()->getParams();
     $group  = $router['group'] ?: '';
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -509,9 +509,6 @@ function list_ospf()
     check_is_read();
 
     $app        = \Slim\Slim::getInstance();
-    $code       = 500;
-    $status     = 'error';
-    $message    = 'Error retrieving ospf_nbrs';
     $sql        = '';
     $sql_params = array();
     $hostname   = $_GET['hostname'];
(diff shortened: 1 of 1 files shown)


html/includes/api_functions.inc.php
```diff
@@ -1319,33 +1404,35 @@ function list_bills()
...
-            $used    = format_bytes_billing($rate_data['total_data']);
-            $percent = round(($rate_data['total_data'] / ($bill['bill_quota'])) * 100, 2);
-            $overuse = $rate_data['total_data'] - $bill['bill_quota'];
-            $overuse = (($overuse <= 0) ? "-" : format_bytes_billing($overuse));
-        }
-        $bill['allowed'] = $allowed;
-        $bill['used'] = $used;
-        $bill['percent'] = $percent;
-        $bill['overuse'] = $overuse;
+        if (bill_permitted($bill['bill_id'])) {
```
Zmegolaz: This could be "if !bill_permitted then continue" instead, to avoid a bit of indentation.

Zmegolaz:
While I think authorization is great, I'm a bit worried about existing installs which (unknowingly?) rely on their API tokens being all powerful. I'm not sure how to best fix this, any idea?

Also, I guess there's now no problem in letting users create their own tokens for only their own user, might be a nice feature.

laf:
> While I think authorization is great, I'm a bit worried about existing installs which (unknowingly?) rely on their API tokens being all powerful. I'm not sure how to best fix this, any idea?

We can send out a notification for that. It's not by design they can do this and merely a mistake that the webui allows them to.

html/includes/api_functions.inc.php
richardlawley: Updated, but the style rules require this is still inside braces.

laf:
@richardlawley Can you rebase this and I'll start testing

richardlawley:
Rebased.

I've got another PR in progress to standardise the output of the API functions (discussed on discord - some functions return err-msg and some return message).  I kept this separate as it's not really related to the permissions issue, but if you're planning on testing each of the functions for this anyway, it might be less work if I add it into this too, so there's only one run through needed?  Happy to keep this separate if that's preferred.

laf:
Stick it in the same as we'll have to notify users for both.

richardlawley:
I have mixed opinions about the loop issue - the existing HTML views (devices etc) all use this technique which is why I used it (to be consistent).  The biggest problem with putting the restriction in the query is that the permission check code is duplicated - some relies on device_permitted, some relies on the SQL filter.  If you were to change device_permitted at some point in the future, code relying on SQL would also need to be updated.

Having said that, I did use the SQL for a couple of functions, get_all_ports and get_port_graphs, because I was going for minimal rewriting of the functions and they simply returned the query result.

laf:
`html/includes/table/devices.inc.php` uses the raw query to list devices. If you mean single device page views then the permitted check is probably easier all round as it's only a query for a single device at that stage.

richardlawley:
I was looking at `html/pages/devices.inc.php`, which is triggered when the device view is set to graph, but fair point.

laf:
I'm happy to help go through the sql queries if you need?

I'm also going to add support for sending json unpretified by default and the ability to send ?pretty=true. Just to save some data, we make use of the API quite a bit and I'd like to trim the response sizes down a little.

richardlawley:
I've added the SQL queries and helper functions to make the return values from the API consistent.  This took some time as it involved going through each of the functions.  Now instead of returning JSON from every API function, you can just call api_success or api_error as appropriate.

As you just mentioned about changing to non-prettified output by default, this is also easier now that the output all goes through these functions instead.

html/includes/api_functions.inc.php
```diff
@@ -112,19 +200,13 @@ function get_port_stats_by_port_hostname()
...
-    $app->response->headers->set('Content-Type', 'application/json');
-    echo _json_encode($output);
+    api_success($port, 'port');
 }
 
 
 function get_graph_generic_by_hostname()
 {
     // This will return a graph type given a device id.
-    global $config;
```
laf: This is still needed for the pages that return graphs as we use `$config` within the included files.

laf:
One issue with the removal of `$config` for graphs. Once that's fixed I'll carry on testing.

laf:
I've tagged this for 1.35 release but would like to get it out as soon as possible after this sundays release as an fyi

html/includes/api_functions.inc.php
```diff
@@ -34,17 +44,93 @@ function authToken(\Slim\Route $route)
...
+function check_is_admin()
+{
+    if (!is_admin()) {
+        api_error(403, 'Insufficient privileges');
+    }
+}
+
+function check_is_read()
+{
+    if (!is_admin && !is_read()) {
```
laf: Should be `is_admin()`

html/includes/api_functions.inc.php
```diff
@@ -1360,75 +1159,54 @@ function list_bills()
...
-        'message' => $message,
-    );
-    $app->response->setStatus($code);
-    $app->response->headers->set('Content-Type', 'application/json');
-    echo _json_encode($output);
 }
 
 function get_device_groups()
 {
+    check_is_read();
```
laf: Probably drop this. No admin/read users can see groups in the webui

html/includes/api_functions.inc.php
```diff
@@ -1305,33 +1108,29 @@ function list_bills()
...
-        $sql   = '';
-        $param = array();
+        $sql    .= '`bills`.`bill_id` = ?';
+        $param[] = $bill_id;
     }
-
-    if (count($param) >= 1) {
-        $sql = "WHERE $sql";
+    if (!is_admin() && !is_read()) {
+        $sql    .= ' AND `bills`.`bill_id` IN (SELECT `bill_id` FROM `bills_perms` WHERE `user_id` = ?)';
```
laf: Table is `bill_perms`

html/includes/api_functions.inc.php
```diff
@@ -1305,33 +1108,29 @@ function list_bills()
...
+        $param[] = $_SESSION['user_id'];
     }
 
-    foreach (dbFetchRows("SELECT `bills`.*,COUNT(port_id) AS `ports_total` FROM `bills` LEFT JOIN `bill_ports` ON `bill_ports`.`bill_id`=`bills`.`bill_id` $sql GROUP BY `bill_name`,`bill_ref` ORDER BY `bill_name`", $param) as $bill) {
+    foreach (dbFetchRows("SELECT `bills`.*,COUNT(port_id) AS `ports_total` FROM `bills` LEFT JOIN `bill_ports` ON `bill_ports`.`bill_id`=`bills`.`bill_id` WHERE $sql GROUP BY `bill_name`,`bill_ref` ORDER BY `bill_name`", $param) as $bill) {
```
laf: This old query is bad. Use:

`foreach (dbFetchRows("SELECT `bills`.* FROM `bills` WHERE $sql ORDER BY `bill_name`", $param) as $bill) {`

laf:
Some more changes, after that. Works for me.

html/includes/api_functions.inc.php
richardlawley: I did see this, along with the query inside the loop to populate ports, but as this had been approved as an addition before I left it.

laf:
Notification sent out, will merge in Wed.

laf:
@richardlawley amazing work - thanks for doing that :)

PR #7714, opened 2017-11-13
Junos dwdm interface sensor support

bergroth:
four new sensors
    Chromatic Dispersion ps/nm
    Delay s (seconds)
    Q_factor (dB)
    preFEC_BER (ratio)

  reuse of
    SNR, dbm, temperature, current

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Thanks.
It seems like the sensors need to be a little more generic.
q-factor -> quality_factor or just quality
prefec_ber -> ber (for junos, you can just set the description to PreFEC BER)

includes/discovery/sensors/frequency/junos.inc.php
```diff
@@ -0,0 +1,41 @@
...
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ *tis*/
+
+echo 'JunOS ';
+
+$multiplier = 1000000;
```
murrant: Instead of multiplying by 1000000 here, you should probably just store the value in MHz.
This also avoids having to change the sensor limits.

includes/polling/functions.inc.php
```diff
@@ -166,7 +166,8 @@ function record_sensor_data($device, $all_sensors)
         }
 
         $rrd_name = get_sensor_rrd_name($device, $sensor);
-        $rrd_def = RrdDefinition::make()->addDataset('sensor', 'GAUGE', -20000, 20000);
+
```
murrant: You probably don't need this, see comment about frequency.

bergroth:
Storing Frequency sensor as MHz and not multiplying shows the limits as -3.0kHz
and 3.0kHz when it should be -+3.0GHz. Looks a bit odd to me. Are there any way
do set the unit (MHz, KHz ...) in the discovery function?

includes/discovery/sensors/frequency/junos.inc.php
murrant: Ignore this. :)  The value should be stored in Hz.  (Wireless sensors are stored in MHz)

includes/polling/functions.inc.php
```diff
@@ -166,7 +166,8 @@ function record_sensor_data($device, $all_sensors)
...
+        $rrd_def = RrdDefinition::make()->addDataset('sensor', 'GAUGE', -2000000000, 2000000000);
```
murrant: Can you just remove the limits here, thanks.  No reason to limit that I can think of.

doc/Developing/os/Health-Information.md
```diff
@@ -25,6 +25,10 @@ Currently we have support for the following health metrics along with the values
 | state                           | #                           |
 | temperature                     | C                           |
 | voltage                         | V                           |
+| delay                           | s                           |
+| quality_factor                  | dB                          |
```
laf: Any links to show that dB is the measurement of quality factor?

doc/Developing/os/Health-Information.md
```diff
@@ -25,6 +25,10 @@ Currently we have support for the following health metrics along with the values
...
+| chromatic_disperision           | ps/nm                       |
```
laf: Should this be ps/nm km?

doc/Developing/os/Health-Information.md
```diff
@@ -25,6 +25,10 @@ Currently we have support for the following health metrics along with the values
 | state                           | #                           |
 | temperature                     | C                           |
 | voltage                         | V                           |
+| delay                           | s                           |
```
laf: Still not sure if ms is a better value to store

html/includes/print-menubar.php
```diff
@@ -392,6 +392,11 @@
     'pressure' => 'thermometer-empty',
     'cooling' => 'thermometer-full',
     'airflow' => 'angle-double-right',
+    'delay' => 'clock-o',
+    'chromatic_dispersion' => 'indent',
+    'ber' => 'sort-amount-desc',
+    'quality_factor' => 'arrows',
```
laf: `area-chart` maybe?

bergroth:
On 2017-11-15 21:31, Neil Lathwood wrote:
>
> *@laf* commented on this pull request.
>
> Some queries around the measurement values for the new sensors.
>
> I think we can convert all of this to yaml with a little extra code. Would you
> be able to provide us with snmpwalk output for this device so we can simulate
> it and do the conversion?
>
sure what is the best way do deliver the file?

I had a couple of things I couldn't get to work with yaml.
1. To set description to  IF-MIB::ifDescr.index
2. to multiply two values to get BER: exponent and mantisa
3. the index for max an min has a different index compare to the interface index.

> --------------------------------------------------------------------------------
>
> In doc/Developing/os/Health-Information.md
> <https://github.com/librenms/librenms/pull/7714#discussion_r151241570>:
>
> > @@ -25,6 +25,10 @@ Currently we have support for the following health metrics along with the values
>  | state                           | #                           |
>  | temperature                     | C                           |
>  | voltage                         | V                           |
> +| delay                           | s                           |
> +| quality_factor                  | dB                          |
>
> Any links to show that dB is the measurement of quality factor?
>
https://www.juniper.net/documentation/en_US/junos/topics/concept/otn-signal-degrade-monitoring-understanding.html
http://mapyourtech.com/entries/general/what-is-q-factor-and-what-is-its-importance-

> --------------------------------------------------------------------------------
>
> In doc/Developing/os/Health-Information.md
> <https://github.com/librenms/librenms/pull/7714#discussion_r151241835>:
>
> > @@ -25,6 +25,10 @@ Currently we have support for the following health metrics along with the values
>  | state                           | #                           |
>  | temperature                     | C                           |
>  | voltage                         | V                           |
> +| delay                           | s                           |
> +| quality_factor                  | dB                          |
> +| chromatic_disperision           | ps/nm                       |
>
> Should this be ps/nm km?
>
ps/nm/km is the specification for the fiber. When you measure over the fiber
length it becomes ps/nm.
> --------------------------------------------------------------------------------
>
> In doc/Developing/os/Health-Information.md
> <https://github.com/librenms/librenms/pull/7714#discussion_r151241951>:
>
> > @@ -25,6 +25,10 @@ Currently we have support for the following health metrics along with the values
>  | state                           | #                           |
>  | temperature                     | C                           |
>  | voltage                         | V                           |
> +| delay                           | s                           |
>
> Still not sure if ms is a better value to store
>
I would be nice to have ms, us, ns and ps.
> --------------------------------------------------------------------------------
>
> In html/includes/print-menubar.php
> <https://github.com/librenms/librenms/pull/7714#discussion_r151242581>:
>
> > @@ -392,6 +392,11 @@
>      'pressure' => 'thermometer-empty',
>      'cooling' => 'thermometer-full',
>      'airflow' => 'angle-double-right',
> +    'delay' => 'clock-o',
> +    'chromatic_dispersion' => 'indent',
> +    'ber' => 'sort-amount-desc',
> +    'quality_factor' => 'arrows',
>
> |area-chart| maybe?
>
sure.

bergroth:
On 2017-11-15 21:31, Neil Lathwood wrote:
>
> *@laf* commented on this pull request.
>
> Some queries around the measurement values for the new sensors.
>
> I think we can convert all of this to yaml with a little extra code. Would you
> be able to provide us with snmpwalk output for this device so we can simulate
> it and do the conversion?
>

https://pastebin.com/Ac8c0Fke

laf:
> 1. To set description to  IF-MIB::ifDescr.index 2. to multiply two values to get BER: exponent and mantisa 3. the index for max an min has a different index compare to the interface index.

1. We can make that available by caching the data.

2. If you put that work into a function we can add support for passing a function in yaml.

3. This could be trickier, with the data we can see.

> juniper.net/documentation/en_US/junos/topics/concept/otn-signal-degrade-monitoring-understanding.html http://mapyourtech.com/entries/general/what-is-q-factor-and-what-is-its-importance-

Juniper link seems to suggest dBQ?

> I would be nice to have ms, us, ns and ps.

Yeah we can't do that :(

laf:
@bergroth we need the snmpwalk in this format: `snmpbulkwalk -OUneb -v2c -c COMMUNITY HOSTNAME .`

bergroth:
- Updated with -OUneb https://pastebin.com/Ac8c0Fke

- Q-factor according to the MIB are in unit dB
   jnxPMCurQ OBJECT-TYPE
        SYNTAX          Integer32
        UNITS           "0.1 dB"
        MAX-ACCESS      read-only
        STATUS          current
        DESCRIPTION
            "'Q' factor estimated at Rx Transceiver port "
        ::= { jnxOpticsPMCurrentEntry 5 }

and Cisco says the same
https://www.cisco.com/c/en/us/td/docs/optical/metroplanner/metroplanner_9_2_1/operations/guide/454mp921_opsguide/454mp921_interfacemodel.html


- Librenms handles the IS units nicely. Stored as seconds and displayed by RRD and Librenms as ps.
screen shot 2017-11-20 at 07 42 15

laf:
Thanks but can you provide the output as is from `snmpbulkwalk -OUneb -v2c -c COMMUNITY HOSTNAME .` - notice the trailing dot. Don't use any other OIDs. We need things like the interface data to match all of this up.

bergroth:
These are routers with full internet table, full snmpwalk going to be huge and
take a long time. Is the IF-MIB ok or do you need more? I might be able to get
one card up in the lab but it might take some time.

laf:
IF-MIB and Sensors might be ok. I need the data to replicate this PR basically.

I've got base junos data so that should be ok.

bergroth:
Included the if-mib as well

laf:
Got a link to the data?

bergroth:
same pastebin as before
https://pastebin.com/Ac8c0Fke

laf:
Can you provide the data as one output in the format required. We don't need the data in both formats and split out like that,

laf:
This won't work without some back end changes I've made but this yaml for chromatic disperson works as expected:

```yaml
mib: JUNIPER-IFOPTICS-MIB:JNX-OPT-IF-EXT-MIB:IF-MIB
modules:
    sensors:
        pre-cache:
            data:
                - oid:
                    - ifDescr
        chromatic_dispersion:
            data:
                -
                    oid: jnxOpticsPMCurrentTable
                    value: jnxPMCurChromaticDispersion
                    num_oid: .1.3.6.1.4.1.2636.3.71.1.2.1.1.1.
                    entPhysicalIndex: '{{ $index }}'
                    entPhysicalIndex_measured: 'ports'
                    descr: '{{ $ifDescr }} CD'
                    index: 'jnxPMCurChromaticDispersion.{{ $index }}'
```

laf:
@bergroth how you getting on with this?

bergroth:
I got all but the BER working with YAML.

BER

has two problems
1. it's two OIDs, exponent and mantisa that needs to calulated into a new value.
(using user_func)
2. the BER index is "ifindex.0". .0 needs to be stripped  for entPhysicalIndex.
Got it working with new code.

I can commit it tomorrow

<<M

laf:
Have to keep ber in php for now then.

murrant:
@bergroth yeah, keep BER in php for now.

FYI, the wireless sensor code supports multiple oids and "aggregating" them.  But that code is very different from the sensor code, for now.

bergroth:
hope I got it right.

build.sql
```diff
@@ -41,7 +41,7 @@ CREATE TABLE IF NOT EXISTS `port_in_measurements` ( `port_id` int(11) NOT NULL,
...
+CREATE TABLE IF NOT EXISTS `sensors` ( `sensor_id` int(11) NOT NULL AUTO_INCREMENT, `sensor_class` varchar(64) CHARACTER SET latin1 NOT NULL, `device_id` int(11) NOT NULL DEFAULT '0', `poller_type` varchar(16) COLLATE utf8_unicode_ci NOT NULL DEFAULT 'snmp', `sensor_oid` varchar(64) CHARACTER SET latin1 NOT NULL, `sensor_index` varchar(10) COLLATE utf8_unicode_ci NOT NULL, `sensor_type` varchar(255) CHARACTER SET latin1 NOT NULL, `sensor_descr` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL, `sensor_divisor` bigint(20) NOT NULL DEFAULT '1', `sensor_multiplier` int(11) NOT NULL DEFAULT '1', `sensor_current` float DEFAULT NULL, `sensor_limit` float DEFAULT NULL, `sensor_limit_warn` float DEFAULT NULL, `sensor_limit_low` float DEFAULT NULL, `sensor_limit_low_warn` float DEFAULT NULL, `entPhysicalIndex` varchar(16) CHARACTER SET latin1 DEFAULT NULL, `entPhysicalIndex_measured` varchar(16) CHARACTER SET latin1 DEFAULT NULL, PRIMARY KEY (`sensor_id`), KEY `sensor_host` (`device_id`), KEY `sensor_class` (`sensor_class`), KEY `sensor_type` (`sensor_type`) ) ENGINE=InnoDB  DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
```
murrant: You cannot change history.  Please add this to your new schema sql file.

bergroth:
I reverted that file, @laf said that I couldn't change that file. corrected
db_schema.yaml

includes/discovery/sensors/current/junos.inc.php
```diff
@@ -17,14 +17,14 @@
...
+        $descr = dbFetchCell('SELECT `ifDescr` FROM `ports` WHERE `ifIndex`= ? AND `device_id` = ?', array($index, $device['device_id'])) . ' Tx Current';
         $limit_low = $entry['jnxDomCurrentTxLaserBiasCurrentLowAlarmThreshold'] / $divisor;
         $warn_limit_low = $entry['jnxDomCurrentTxLaserBiasCurrentLowWarningThreshold'] / $divisor;
         $limit = $entry['jnxDomCurrentTxLaserBiasCurrentHighAlarmThreshold'] / $divisor;
         $warn_limit = $entry['jnxDomCurrentTxLaserBiasCurrentHighWarningThreshold'] / $divisor;
         $current = $entry['jnxDomCurrentTxLaserBiasCurrent'];
         $entPhysicalIndex = $index;
         $entPhysicalIndex_measured = 'ports';
-        discover_sensor($valid['sensor'], 'current', $device, $oid, 'rx-' . $index, 'junos', $descr, $divisor, $multiplier, $limit_low, $warn_limit_low, $warn_limit, $limit, $current, 'snmp', $entPhysicalIndex, $entPhysicalIndex_measured);
+        discover_sensor($valid['sensor'], 'current', $device, $oid, 'tx-' . $index, 'junos', $descr, $divisor, $multiplier, $limit_low, $warn_limit_low, $warn_limit, $limit, $current, 'snmp', $entPhysicalIndex, $entPhysicalIndex_measured);
```
laf: Annoyingly, we can't change this from rx-. It will mean data loss for these sensors :(

sql-schema/223.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `sensors` MODIFY `sensor_divisor` BIGINT(20) NOT NULL DEFAULT '1';
```
laf: we have the potential for another schema change for sensors here. I'd like to get that into this PR so we only do one overall change to the table. Just waiting on confirmation that it fixes another issue.

PR #7717, opened 2017-11-13
API - Validate columns parameter against fields in table

richardlawley:
API functions get_all_ports and get_port_graphs had an SQL injection vulnerability as they take a list of columns to include and do not validate this.  This PR adds that validation in.

Example URL: `/api/v0/devices/1/ports?columns=VERSION()--`

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -818,6 +818,7 @@ function get_port_graphs()
     } else {
         $columns = 'ifName';
     }
+    validate_column_list($columns, 'ports');
 
     // use hostname as device_id if it's all digits
     $device_id   = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
@@ -890,6 +891,7 @@ function get_all_ports()
     } else {
         $columns = 'ifName';
     }
+    validate_column_list($columns, 'ports');
     $ports = dbFetchRows("SELECT $columns FROM `ports` WHERE `deleted` = 0");
 
     $output = array(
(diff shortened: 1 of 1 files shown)


laf:
Thanks for this PR. As per my comment on discord, we actually store all table info in `misc/db_schema.yaml`.

`$db_schema = Yaml::parse(file_get_contents($schema_file));`

you can then iterate through that rather than running a DB query.

richardlawley:
I did look at that, but it looked a much more heavyweight option than doing a quick query to the DB to get the list for that table only.  The file is 100K+ of text to be parsed and then iterated over, rather than a single SQL query which is likely held in the query cache after the first execution anyway.

laf:
It's a yaml file so you just load it into an array. It should be simpler than the DB query.

richardlawley:
Ok.  I think the DB method was more efficient, but I've changed it to load from the yaml.

html/includes/api_functions.inc.php
```diff
@@ -1693,3 +1695,34 @@ function list_logs()
...
+
+function validate_column_list($columns, $tableName)
+{
+    global $config;
+    $app = \Slim\Slim::getInstance();
+
+    $column_names = explode(',', $columns);
+    $db_schema = Symfony\Component\Yaml\Yaml::parse(file_get_contents($config['install_dir'] . '/misc/db_schema.yaml'));
+    $valid_columns = array();
+    foreach ($db_schema[$tableName]['Columns'] as $col) {
```
murrant: You can use array_column()  (we have a library adding it for 5.3 and soon will be able to use the native implementation)

html/includes/api_functions.inc.php
```diff
@@ -1693,3 +1695,34 @@ function list_logs()
...
+        $valid_columns[] = $col['Field'];
+    }
+    $invalid_columns = array();
+
+    foreach ($column_names as $col) {
```
murrant: Similarly, this could be replaced with
```
$invalid_columns = array_diff(array_map('trim', $column_names), $valid_columns);
```

html/includes/api_functions.inc.php
```diff
@@ -1693,3 +1695,34 @@ function list_logs()
     $app->response->headers->set('Content-Type', 'application/json');
     echo _json_encode($output);
 }
+
+function validate_column_list($columns, $tableName)
+{
+    global $config;
+    $app = \Slim\Slim::getInstance();
```
murrant: Probably only need to get an instance of slim if there is an error, so you could move this into that if block.

murrant:
Looks good, just a few code style suggestions.

richardlawley:
Fair points, wasn't aware of those functions (not primarily a PHP developer).

PR #7718, opened 2017-11-13
fix: Searching for the user in LDAP is much faster than dumping all the us…

davidediger:
…ers and iterating server-side.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/authentication/ldap.inc.php
+++ b/html/includes/authentication/ldap.inc.php
@@ -217,6 +217,7 @@ function can_update_users()
 
 function get_user($user_id)
 {
+    $user = array();
     try {
         $connection = get_ldap_connection();
 
@@ -240,7 +241,7 @@ function get_user($user_id)
                         get_membername($username)
                     );
                     if (!Config::has('auth_ldap_group') || $ldap_comparison === true) {
-                        $userlist[$username] = array(
+                        $user[] = array(
                             'username' => $username,
                             'realname' => $realname,
                             'user_id' => $user_id,
--- a/html/includes/authentication/ldap.inc.php
+++ b/html/includes/authentication/ldap.inc.php
@@ -217,10 +217,42 @@ function can_update_users()
 
 function get_user($user_id)
 {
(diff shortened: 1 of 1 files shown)


html/includes/authentication/ldap.inc.php
```diff
@@ -217,10 +217,42 @@ function can_update_users()
...
+                        $userlist[$username] = array(
+                            'username' => $username,
+                            'realname' => $realname,
+                            'user_id' => $user_id,
+                            'email' => $email,
+                        );
+                    }
+                }
+            }
             return $user;
```
murrant: $user doesn't seem to be set?  Also, the return should probably be inside the foreach loop.

Zmegolaz:
This is almost identical to the `get_userlist()` function, how about breaking it out to a separate function and reuse that in both? Or put `$filter` as an optional argument to either `get_user()` or `get_userlist()` and call that one from the other.

murrant:
@Zmegolaz It would need to be a third function.  You can't change the signature of get_user() and get_userlist().

html/includes/authentication/ldap.inc.php
davidediger: Fixed the return in the latest commit.  The LDAP query should only return 1 result, so inside or outside the foreach shouldn't matter.  I'm not sure what we should do if the LDAP query returns more than one result.

davidediger:
get_userlist() and get_user() are substantially similar except for a) the LDAP filter and b) returning one user vs an array of users.  a) is relatively easy to abstract.  b) is going to require additional engineering.  Agree we should avoid code duplication.  I just wanted to push what we're running in production to get a conversation started.

murrant:
@davidediger The authentication refactor has been merged, can you update this to the new classes?

laf:
@davidediger Any update on this?

laf:
No response on this so far. Closing until we hear back from @davidediger

PR #7724, opened 2017-11-13
feature: Added support for various logging levels

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

  - Removed the config for mysql logging. It wasn't documented so I doubt anyone will have configured it. The new default won't log that anyway now.
  - Added Webui support for specifying the logging level to use.
  - Defaults to info which is basically poller and disco logging. The idea is to minimise the amount of logging we do unless users want to debug something - even switch it off if they want.


includes/functions.php
```diff
@@ -76,13 +76,25 @@ function only_alphanumeric($string)
     return preg_replace('/[^a-zA-Z0-9]/', '', $string);
 }
 
-function logfile($string)
+function logfile($string, $level = 'error')
```
murrant: You could use the PHP constants here instead of a string.
http://php.net/manual/en/function.syslog.php

includes/functions.php
```diff
@@ -76,13 +76,25 @@ function only_alphanumeric($string)
...
+        'error' => 7,
+        'debug' => 9,
+    );
+    // Replace human readable values with numeric
+    $level = $levels[$level];
 
-    $fd = fopen($config['log_file'], 'a');
-    fputs($fd, $string . "\n");
-    fclose($fd);
+    $log_level = Config::get('logging.level', '0');
```
murrant: '0' should probably be just 0 to avoid a type cast here.

includes/functions.php
```diff
@@ -76,13 +76,25 @@ function only_alphanumeric($string)
...
+    if ($log_level !== '0') {
```
murrant: same here.

murrant:
Good one, just a little tweak to the function arguments.

includes/functions.php
laf: Done

includes/functions.php
murrant: Do we need a none?

laf: Not necessarily but for us, we send our logs to /dev/null to save the disk io. I know that would be reduced if we choose info now but none seemed ok to have.

I did start to change it but imho I don't think the PHP constants fit for us. DEBUG is level 7 and INFO level 6 whilst ERR is 0. I could just skip having a debug so the check is still simple but it just doesn't seem right.

murrant: Not sure I follow here @laf  LOG_ERR is 3.  We could definitely have the default be 2 (LOG_CRIT)

murrant: Also have the bonus of log levels being familiar since they match syslog.

laf: What I mean is that the debug level is higher level then info. I.e the severity level is ascending until you get to DEBUG which then is higher than info. So we either need to skip the use of debug (we use this term in a lot of places so seems daft to drop here) or we make the check against the required level a little more complicated

murrant: I think you are trying to make the two lists match up.  I'm saying scrap your original list in this PR and just use the syslog compatible levels.

laf: I don't agree personally so I don't plan on making any changes, feel free to change it if it bothers you enough.

murrant: That's fine I just feel like you are possibly re-inventing the wheel here.  We could just use Monolog.

murrant:
Yeah, I think monolog is the way to go.

Issue #7409, opened 2017-09-29
Add config option to allow stacked bandwidth graphs


laf:
https://community.librenms.org/t/mrtg-style-graphs/1584

Add a config option like `$config['graphs']['port']['stacked']` default to false and allow true.

A central function most likely needs adding to `html/includes/functions.inc.php` that will just return 1 or -1 based on the config option and passed in to the places you see `-1,*` in `html/includes/graphs/generic_*`

PR #7725, opened 2017-11-14
feature: stacked graphs

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

as per: #7409 

before:
screen shot 2017-11-14 at 3 43 01 am

after: 
screen shot 2017-11-14 at 3 43 19 am


murrant:
You should be using Config instead of a custom function.  Like `Config::get('graphs.port.stacked')`.
You should allow this to be set in the webui, under the webui -> graph settings section.

Also, you probably shouldn't call it MRTG graphs, just stacked graphs is fine.

config.php.default
```diff
@@ -49,3 +49,6 @@ $config['enable_billing'] = 1;
 
 # Enable the in-built services support (Nagios plugins)
 $config['show_services'] = 1;
+
+# Enable the MRTG style graphs if set to true
+$config['graph']['stacked'] = false;
```
laf: Maybe `$config['graphs']['stacked'] = false; `

We already have graphs for a few other config options.

laf: In fact, I'd say remove it from here anyway as we want users to use the webui to config this

html/pages/settings/webui.inc.php
```diff
@@ -24,6 +24,10 @@
             'svg' => 'svg',
         ),
     ),
+    array('name' => 'webui.graph_stacked',
```
laf: You've used two variables now for this. I'd actually say `graphs.stacked` is better but I realise I used `webui.graph_type` for svg so I'd be fine with keeping this as well.

html/includes/graphs/generic_multi_seperated.inc.php
```diff
@@ -1,19 +1,41 @@
...
+ * @copyright  2017 LibreNMS
+ * @author     LibreNMS Contributors
+*/
+
+use LibreNMS\Config;
 
 require 'includes/graphs/common.inc.php';
 
+if (Config::get('graph.stacked') == true) {
+    $transparency = 95;
```
laf: What's the reason for the different transparency values in the files?

Overall we should put this block into a function to make it re-usable

crcro: this graph is used on device ports overview and lower values make it unpleasant to view (too much transparency).

laf: Is that not the same for other pages then?

laf:
image


image

crcro:
is better this way?
about the colors ... should change them?

laf:
Better but I would still use a key value return on the array, `return array('transparency' => x, 'conversion' => y);` or something so that using the variables within the graph code is obvious to someone else without having to look at what the function does.

The colours definitely need fixing, it's indistinguishable right now on what is what.

crcro:
screen shot 2017-11-16 at 1 17 32 pm
screen shot 2017-11-16 at 1 17 50 pm


is this acceptable @laf ?

laf:
It's definitely better but why can't we just have the original colours with some transparency?

crcro:
because if they overlay you will get that greyed tone (like in your picture where you have same in/out):

screen shot 2017-11-16 at 2 01 38 pm
screen shot 2017-11-16 at 2 01 49 pm

laf:
> because if they overlay you will get that greyed tone (like in your picture where you have same in/out):

Ah sorry, right so you've not changed the colours it's just how it looks because of the transparency.

I'm not the target audience for this really so I'll leave it up to others comment on if they think it looks ok..

kkrumm1:
are there other color options for stacked?

crcro:
@kkrumm1 ... the scope is to use the same colors for inverted and stacked graphs. currently libre uses greens and blues for bits graph but at @laf points i've come with greens and purples ...

kkrumm1:
Looks good to me then :) thank you @crcro

kkrumm1:
👍

kkrumm1:
When I try enable stacked graphs in the WebUI I get this error.
image

laf:
If this is on a test install you try different PRs on then it might be because your schema was already at or past this version. Run this in mysql:

```
INSERT INTO `config` (`config_name`, `config_value`, `config_default`, `config_descr`, `config_group`, `config_group_order`, `config_sub_group`, `config_sub_group_order`, `config_hidden`, `config_disabled`) VALUES('webui.graph_stacked', 'false', 'false', 'Display stacked graphs instead of inverted graphs', 'webui', 0, 'graph', 0, '0', '0');
```

kkrumm1:
That worked... Also, this was on my production server. I never applied this PR to it this was after daily updates.

PR #7729, opened 2017-11-14
device: Changing HWg STE2 to YAML and adding new logos

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

screen shot 2017-11-14 at 16 55 53


FTBZ:
Ping @crcro for information.

includes/definitions/hwg-ste2.yaml
```diff
@@ -4,9 +4,37 @@ type: environment
 icon: hwg
 over:
     - { graph: device_temperature, text: Temperature }
-    - { graph: device_humidity, text: Humidity }
```
murrant: Why was this overview graph removed?

FTBZ: It's not a standard shipped sensor, only the temperature sensor is shipped by default. The humidity optional, so it's better to remove this graph as over.

laf:
So I think this will lose historical data due to the index change. Whilst I prefer using more complete indexes now rather than single digits. In cases like this we should maintain historical data so just change to `index: '{{ $index }}'`

FTBZ:
@laf Done, indexes were changed. But on my installation these sensors were not working so I don't think that will impact people.

includes/definitions/discovery/hwg-ste2.yaml
```diff
@@ -0,0 +1,43 @@
...
+                            op: '!='
+                            value: 4
+        state:
+            data:
+                -
+                    oid: inpTable
+                    value: inpAlarmState
+                    num_oid: .1.3.6.1.4.1.21796.4.9.1.1.4.
+                    descr: 'Alarm: {{ $inpName }}'
+                    index: '{{ $index }}'
```
laf: Sorry you didn't need to do it here as that's a new sensor.

PR #7734, opened 2017-11-14
Implement dynamic temperature scale detection and value conversion for siteboss

thecityofguanyu:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Just added comment to the discovery file because I have no idea how to open an empty PR.

### DIFF
--- a/includes/definitions/discovery/siteboss.yaml
+++ b/includes/definitions/discovery/siteboss.yaml
@@ -1,3 +1,8 @@
+# Just adding comments because I have no idea how to open an empty PR
+# Action items: 1) Find way to call fahrenheit_to_celsius PHP function from YAML file to convert temp value to C from F
+# Action items: 2) Based on value of substring in esPointValueStr (F or C),  dynamically call that previous function 
+# Example output for string search: SITEBOSS-530-STD-MIB::esPointValueStr.16.1.1 = STRING: 69 F
+
 mib: SITEBOSS-530-STD-MIB
 modules:
     sensors:
--- a/includes/definitions/discovery/siteboss.yaml
+++ b/includes/definitions/discovery/siteboss.yaml
@@ -1,8 +1,3 @@
-# Just adding comments because I have no idea how to open an empty PR
-# Action items: 1) Find way to call fahrenheit_to_celsius PHP function from YAML file to convert temp value to C from F
-# Action items: 2) Based on value of substring in esPointValueStr (F or C),  dynamically call that previous function 
-# Example output for string search: SITEBOSS-530-STD-MIB::esPointValueStr.16.1.1 = STRING: 69 F
-
 mib: SITEBOSS-530-STD-MIB
 modules:
     sensors:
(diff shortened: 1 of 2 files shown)


murrant:
Why not use yaml to discover, but have a custom polling file?

laf:
@murrant we don't need to have a custom polling file if we register a user function (i.e f_to_c for temps).

What this needs is a check for the returned value within yaml for the string f or F and if found we add the function. I'll take a look at this.

laf:
@thecityofguanyu please try this patch:

https://p.libren.ms/view/raw/d22581fa

laf:
@thecityofguanyu All work ok for you?

thecityofguanyu:
@laf sorry, responded to you earlier on Discord.

All is good! Definitely being converted to celsius and graphing as such.

laf:
Good shout.

Poller does it before divisor/multiplier runs so we need to do it the same for disco. This means in discovery_process()

@thecityofguanyu Can you add this patch as well please:

```php
diff --git a/includes/discovery/functions.inc.php b/includes/discovery/functions.inc.php
index 0f2eef7a5..db2328377 100644
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -1038,6 +1038,11 @@ function discovery_process(&$valid, $device, $sensor_type, $pre_cache)
                     $high_limit = is_numeric($data['high_limit']) ? $data['high_limit'] : dynamic_discovery_get_value('high_limit', $index, $data, $pre_cache, 'null');

                     $sensor_name = $device['os'];
+
+                    if (isset($user_function) && function_exists($user_function)) {
+                        $value = $user_function($value);
+                    }
+
                     if ($sensor_type === 'state') {
                         $sensor_name = $data['state_name'] ?: $data['oid'];
                         create_state_index($sensor_name, $data['states']);

```

thecityofguanyu:
@murrant @laf second patch applied. Re-discovered a node and re-polled sensors -- still looks good to me.

```
SNMP[/usr/bin/snmpget -v1 -c 'public' -OUQnt -M /opt/librenms/mibs:/opt/librenms/mibs/asentria udp:asentria_kel00:161 .1.3.6.1.4.1.3052.12.1.1.1.1.7.16.1.1]
.1.3.6.1.4.1.3052.12.1.1.1.1.7.16.1.1 = "47 F"

Checking (snmp) temperature temp...
8.33
RRD[update asentria_kel00/sensor-temperature-siteboss-16.1.1.rrd N:8.33 --daemon unix:/var/run/rrdcached.sock]
RRDtool Output: OK u:0.00 s:0.00 r:9.15
Sending asentria_kel00.sensor.sensor.temperature.siteboss.16_1_1.sensor 8.33 1511360467
SQL[UPDATE `sensors` set `sensor_current` ='8.33',`sensor_prev` ='7.78',`lastupdate` =NOW() WHERE `sensor_class` = 'Temperature' AND `sensor_id` = '3885']
```

PR #7735, opened 2017-11-14
fix: use memcached to lock daily processes on Distributed Pollers

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/Proc.php
```diff
@@ -43,6 +43,11 @@ class Proc
      */
     private $_synchronous;
 
+    /**
+     * @var int|null hold the exit code, we can only get this on the first process_status after exit
+     */
+    private $_exitcode = null;
```
laf: Are these changes supposed to be in here? Seems totally different from the locking side.

laf:
This now requires distributed pollers to be running php-memcached module, I can see the check for the function exists but i don't see where this falls back gracefully?

murrant:
Hmm, that is a good point.  Thanks @laf

murrant:
The fallback is there.  I guess I just hadn't thought of it that way.

If memcached is missing it throws a lock exception, which is caught and skips all of the code that should be inside the lock.  This is the current behavior, none of the mysql processes are run when distributed poller is enabled.

laf:
> If memcached is missing it throws a lock exception, which is caught and skips all of the code that should be inside the lock. This is the current behavior, none of the mysql processes are run when distributed poller is enabled.

How is that current behavior? All the DB cleanup still run at the moment when you run ./daily.sh

murrant:
It is the current behavior for distributed polling.

For non-distributed polling, it will use the existing FileLock.

laf:
Thanks.

I'll merge this in but will exclude from the release this sunday.

PR #4999, opened 2016-11-15
refactor: Improve Cisco ASA Polling Performance

Rosiak:
- I went through the modules and disabled the ones not supported by the ASA platform.
- I disabled sensor discovery for context based devices as one usually poll the admin context for that purpose, also this output is buggy for contexts on some firmware versions.

Tested on various ASA platforms and on various firmware levels.
This should yield quite a performance improvement on some installs.

asa

PR #7745, opened 2017-11-16
fix: enexus do not disable the os discovery module

murrant:
and other modules it is pointless to disable

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/enexus.yaml
+++ b/includes/definitions/enexus.yaml
@@ -6,30 +6,17 @@ nobulk: 1
 mib_dir:
     - eltek
 discovery:
-    -
-      snmpget:
+    - snmpget:
           oid: .1.3.6.1.4.1.12148.10.2.6.0
-          value:
-            - 'Micropack System'
+          value: 'Micropack System'
 poller_modules:
-    applications: 0
-    bgp-peers: 0
     entity-physical: 0
     hr-mib: 0
     ipSystemStats: 0
-    ipmi: 0
-    mempools: 0
     netstats: 0
-    ntp: 0
     ospf: 0
-    processors: 0
-    services: 0
-    storage: 0
-    ucd-diskio: 0
     ucd-mib: 0
-    wireless: 0
     ports: 0
-    wifi: 0
     stp: 0
 discovery_modules:
     ports-stack: 0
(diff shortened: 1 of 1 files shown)


includes/definitions/enexus.yaml
```diff
@@ -6,30 +6,17 @@ nobulk: 1
...
-    -
-      snmpget:
+    - snmpget:
           oid: .1.3.6.1.4.1.12148.10.2.6.0
-          value:
-            - 'Micropack System'
+          value: 'Micropack System'
 poller_modules:
-    applications: 0
-    bgp-peers: 0
```
laf: How come these have been re-enabled. If the devices don't support them then no point in polling/discoverying them.

murrant: Because there is no point in disabling them, and I hate people copying things like this.

It saves one sql query to disable them.

murrant: Also, services isn't poller/discovery module.  The wireless module doesn't even do an sql query if there is no support.

laf:
> Because there is no point in disabling them, and I hate people copying things like this.
> 
> It saves one sql query to disable them.

But it genuinely makes a difference. https://github.com/librenms/librenms/pull/4999

This made a huge difference for us polling 2000+ Cisco ASAs. I agree some of them slipped in by accident but we shouldn't disco/poll modules which have no data for some OS'

murrant:
I manually checked each one I removed to make sure there were no snmp calls.

Honestly, I would like to go through all the poller modules and make sure they check the db before doing any snmp calls.  Since failing snmp calls take a substantial amount of time.

laf:
But that only works on polling not discovery and whilst we're quite tolerant of discovery times, we should still improve things where we can.

We also recommend people disable modules in the performance doc, no point in asking users to do this across all installs when we can just do it quite simply.

murrant:
Yes, if you look, I didn't remove any valid discovery modules here.  

Quite alright with users disabling whatever.  Yeah, if users want to go through and disable modules that take 0.001s that is fine and I'm sure happens.

I could actually see a future where there is no separate toggles for poller modules.

PR #7757, opened 2017-11-17
Only warn if there is ipv6 disabled

setiseta:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Validations/Programs.php
+++ b/LibreNMS/Validations/Programs.php
@@ -58,6 +58,12 @@ class Programs implements ValidationGroup
     {
         $target = ($bin == 'fping' ? '127.0.0.1' : '::1');
         $validator->execAsUser("$cmd $target 2>&1", $output, $return);
+
+        if ($output[count($output) -1] == '::1 is unreachable') {
+            $validator->warn("fping6 available, but ipv6 seems disabled");
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


LibreNMS/Validations/Programs.php
```diff
@@ -58,6 +58,12 @@ public function extraFpingChecks(Validator $validator, $bin, $cmd)
     {
         $target = ($bin == 'fping' ? '127.0.0.1' : '::1');
         $validator->execAsUser("$cmd $target 2>&1", $output, $return);
+
+        if ($output[count($output) -1] == '::1 is unreachable') {
```
murrant: Probably a little more clear to use this:
```
if (str_contains($output, '::1 is unreachable')) {
```

LibreNMS/Validations/Programs.php
```diff
@@ -58,6 +58,12 @@ public function extraFpingChecks(Validator $validator, $bin, $cmd)
...
+            $validator->warn("fping6 available, but ipv6 seems disabled");
```
murrant: This message should probably just be;
```
IPv6 is disabled on your server, you will not be able to add IPv6 devices.
```

murrant:
Thanks!

PR #7758, opened 2017-11-17
fix Cisco SG300 to use 64-bit counters

itvirta:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -210,8 +210,10 @@ if ($device['os'] === 'f5' && (version_compare($device['version'], '11.2.0', '>=
             $port_stats = snmpwalk_cache_oid($device, 'ifXEntry', $port_stats, 'IF-MIB');
         }
         $hc_test = array_slice($port_stats, 0, 1);
-        if ((!isset($hc_test[0]['ifHCInOctets']) && !is_numeric($hc_test[0]['ifHCInOctets'])) ||
-            ((!isset($hc_test[0]['ifHighSpeed']) && !is_numeric($hc_test[0]['ifHighSpeed'])))) {
+        if (!isset($hc_test[0]['ifHCInOctets'])    || !is_numeric($hc_test[0]['ifHCInOctets'])    ||
+            !isset($hc_test[0]['ifHighSpeed'])     || !is_numeric($hc_test[0]['ifHighSpeed'])     ||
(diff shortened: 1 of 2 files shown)


includes/polling/ports.inc.php
```diff
@@ -210,8 +210,10 @@
             $port_stats = snmpwalk_cache_oid($device, 'ifXEntry', $port_stats, 'IF-MIB');
         }
         $hc_test = array_slice($port_stats, 0, 1);
-        if ((!isset($hc_test[0]['ifHCInOctets']) && !is_numeric($hc_test[0]['ifHCInOctets'])) ||
-            ((!isset($hc_test[0]['ifHighSpeed']) && !is_numeric($hc_test[0]['ifHighSpeed'])))) {
+        if (!isset($hc_test[0]['ifHCInOctets'])    || !is_numeric($hc_test[0]['ifHCInOctets'])    ||
+            !isset($hc_test[0]['ifHighSpeed'])     || !is_numeric($hc_test[0]['ifHighSpeed'])     ||
+            !isset($hc_test[0]['ifHCInUcastPkts']) || !is_numeric($hc_test[0]['ifHCInUcastPkts']) ||
+            $hc_test[0]['ifHCInUcastPkts'] <= 0) {
```
laf: ifHCInUcastPkts could be legitimately zero so not sure `<=` is the best check.

itvirta: Hmm, yes, if it's zero it will be ignored, but ifEntry could still get loaded in full on devices that don't need it. I could see about making a per-model flag, but I couldn't know the exact list of devices that need it.

Of course, I suppose there could be a user-configurable per-device for option for skipping the optimized ifEntry loading, that would fix the problem on these.

laf: Is this issue not a firmware issue? I can't help but feel like we're putting more work arounds when this should be fixed (if it already isn't) by the vendor.

itvirta: @laf that's one way of looking at it. But currently libremns is explicitly made to read only the 32-bit counters, which doesn't cause just missing data, but broken data (because of the overflows). In my opinion, missing packet counters is better than broken octet counters, so the least-nuisance solution would be to just the read the HC counters as usual, and ignore the fact that the packet counters show zeroes. Getting zeroes should be rather obvious, and immediately noticeable, which overflowing counters might not be. So just remove the `bad_ifXEntry` then, and avoid throwing the baby out with the bathwater.

Issue #7210, opened 2017-08-22
Links for API Docs on home page 404


dfc:
The API docs links from https://docs.librenms.org/ are all broken:
```
            <a href="/API/API-Docs/">API Docs</a><br>
            <a href="/API/API-Docs/#api-structure">API Structure</a><br>
            <a href="/API/API-Docs/#api-endpoints">API Endpoints</a><br>
            <a href="/API/API-Docs/"><strong>More...</strong></a>
```

It looks like this should be

```
            <a href="/API/">API Docs</a><br>
            <a href="/API/#api-structure">API Structure</a><br>
            <a href="/API/#api-endpoints">API Endpoints</a><br>
            <a href="/API/"><strong>More...</strong></a>
```

laf:
Are you submitting a pull request to fix?

PR #7762, opened 2017-11-17
 docs: Continue to reorganize docs

murrant:
Make landing pages for /Installation, /API, etc.
Split API out into multiple docs
Properly use multi-level navigation
Add lower nav buttons (tops are bug so disable for now)
Add logo for fanciness
Refine index page a bit

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


doc/API/API-Docs.md
```diff
@@ -1 +0,0 @@
-<meta http-equiv="refresh" content="0; url=/API" />
```
laf: This is to preserve old links, why remove it?

doc/Support/FAQ.md
```diff
@@ -268,7 +262,7 @@ There is a known issue for net-snmp, which causes it to report incorrect disk si
 
 #### <a name="faq8"> How do I add support for a new OS?</a>
 
-Please see [Supporting a new OS](../Developing/Support-New-OS.md)
+Please see [Supporting a new OS](/#Support-for-a-New-OS)
```
laf: Links should stay pointing to markdown files, they should get converted correctly whilst ensuring if we change the theme again and a # isn't needed then it doesn't break.

doc/Support/Support-New-OS.md
```diff
@@ -1,2 +1,2 @@
 source: Support/Support-New-OS.md
-<meta http-equiv="refresh" content="0; url=/Developing/Support-New-OS/" />
+<meta http-equiv="refresh" content="0; url=/Developing/Support-New-OS/__" />
```
laf: What's the reason for the __?

doc/API/API-Docs.md
murrant: I thought it was unused in the docs, I'll look through the original mkdocs.yml again.

laf: It's from the v1 theme. The links are in older installs of LibreNMS and most probably linked to from reddit and other places in old posts.

murrant: This wasn't defined in mkdocs.yml, so it wasn't being built.  That is why I removed it.

doc/Support/Support-New-OS.md
murrant: probably attempting to type in another window and I missed it.  Thanks.

doc/API/API-Docs.md
murrant: If you can tell me the URL it was supposed to be, I can fix it.

laf: It was pre-mkdocs by the looks of it:
https://github.com/librenms/librenms/issues/7210

Shows that the old links were still in place.

laf: You've fixed it anyway.

PR #7765, opened 2017-11-19
webui: alerts table refresh

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

before:
screen shot 2017-11-19 at 3 23 30 am

after:
screen shot 2017-11-19 at 3 23 02 am


kkrumm1:
I like it just did a test pull on it. 👍 
idk if users will miss the severity status.

crcro:
it’s not missing, it just doesn’t have the title in table header, the icon is still there at the begining of each row

kkrumm1:
ah, I see. this is very nice.

laf:
Really like the idea behind this. However I have a few comments :)

1) The change to the status indicator goes away from how the devices and services page looks. Same as this PR: https://github.com/librenms/librenms/pull/7628.

2) The status seems like it's no longer relevant. It's not about if notifications are going out it's just an indicator that the alert is active. In essence the status will never be anything other than `alert`.

3) Can you get some colour in the ack icon.

4) Can you change the tooltip messages to be in normal case rather than all UPPERCASE. I.e: Uppercase.

5) The Procedure URL doesn't work for me.

crcro:
Is this better @laf ?

screen shot 2017-11-20 at 3 33 18 am

FTBZ:
My opinion about all the recent PR for the UI is that the paging is always too big in all interface. We lost a lot of place only for ... blank space. Same with the title on the dashboard.

So, everything that can make the interface more readable and tiny is welcome.

laf:
@crcro a lot better.

Now, you just need to work with @Rosiak to decide on the best way to display the left hand side block and then he can update the services and devices page so we can merge them at the same time :)

@FTBZ The table hasn't changed so I think this might be the status block on the left increasing the space between rows.

crcro:
after recent changes:

screen shot 2017-11-20 at 9 48 36 pm

laf:
Looks great.

crcro:
alerts page cleanup
before:
screen shot 2017-11-25 at 10 51 09 pm

after:
screen shot 2017-11-25 at 10 50 10 pm

crcro:
pushed the alert log table

before:
screen shot 2017-12-01 at 1 42 30 am

after:
screen shot 2017-12-01 at 1 42 12 am

laf:
Probably don't need the Status column here, the colours will show the state.

daynedrak:
So just want to leave a comment, as this changed while I was away for the holidays. While I get that you're trying to clean it up and make it more compact, I really hate not being able to see whether or not the alert is muted. Furthermore, the acknowledgement colors threw me. If I remember right, muted showed green. Regardless, the way it is now, seeing red when it's acknowledge and green when it's not acknowledged seems reversed. I see red and my instinct is something is wrong, and in an acknowledgement column, my instinct was that it was unacknowledged. Seeing green for unacknowledged alerts likewise seems backwards

laf:
I'm wondering if we should have orange as muted / acknowledged rather than green anyway.

Will submit a PR now.

laf:
Going for this:

image

PR #7773, opened 2017-11-20
fix: bgp-peers fails when ip parsing fails

murrant:
Don't try to call ->getFamily() if we didn't parse an IP
Stops the discovery after this is encountered.

We have a junos device returning '&hT' for a peer, tracking down the source, but fixing this is advised in the meantime.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/bgp-peers.inc.php
+++ b/includes/discovery/bgp-peers.inc.php
@@ -1,6 +1,7 @@
 <?php
 
 use LibreNMS\Config;
+use LibreNMS\Exceptions\InvalidIpException;
 use LibreNMS\Util\IP;
 
 if (Config::get('enable_bgp')) {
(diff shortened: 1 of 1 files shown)


laf:
Doesn't cause issues on Arista.

However should the lack of IP being returned not be handled in IP::fromHexString otherwise this check will have to be repeated anywhere else this is used.

murrant:
Most of the time we just convert it to a string afterwards.  So, null -> ''.  This tries to call a method on the object, which is null when it fails.  

Try catch is much better, otherwise you have to add in if not null else.

murrant:
Test by user that reported the issue.

murrant:
@laf there is a check inside IP::fromHexString, it throws an invalid IP exception.  However, we had enabled the ignore errors parameter on this call previously.

laf:
Maybe we shouldn't return null then and instead an empty string?

Ignoring errors is correct though, we should never break polling or discovery from completing.

murrant:
`''->addressFamily()` would still cause a fatal error ;)

BradHooper:
This fixed the lack of BGP prefix graphs on Juniper devices. Thank you.

PR #7775, opened 2017-11-21
fix: Use correct entity columns for sensors

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Shouldn't have created these two columns in the first place.

This fixes that, updates the config macro, updates the existing alert rules, updates the sensors tables and then drops the old columns.

Only port was used in this regards so these queries are safe and we don't lose any other data.

### DIFF
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -206,7 +206,7 @@ function discover_device(&$device, $options = null)
 //end discover_device()
 
 // Discover sensors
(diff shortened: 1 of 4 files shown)


asteen-nexcess:
@laf Looks good! Had to delete/re-add the alert rules for sensor under/over limit with linked port in order for it to alert properly, but is working as expected. Thanks again for the help.

murrant:
./daily.sh regenerates alert rules, so that should make the alert rules update right away.

PR #7778, opened 2017-11-22
docs: new SELinux docs

murrant:
I am not feeling the best about this doc's location/integration.

What do we think about splitting apart the install docs?  We could share some docs between the different docs...
Also would show up on the left toc and have the prev/next buttons.  I just don't want too many sections.

However it would would probably be nice to have the option to show the install docs all on one page too.  Not sure if we can do that.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Not sure I like directing users away from the install page to somewhere else just to get some lines for a copy and paste.

murrant:
I agree.  That is why I thought about splitting the install pages up.  Then it would be just part of the flow

murrant:
I'm fine with just embedding this in the install docs for now. Sound good?

laf:
Yup

laf:
Actually, I'm wondering if this should be a script that people run to set SELinux perms up.

murrant:
@laf not a bad thought.  But perhaps merge this first to get the info out there.

PR #7799, opened 2017-11-26
feature: timezone support for graylog

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

feature request from community site

before:
screen shot 2017-11-26 at 4 40 06 pm

after:
screen shot 2017-11-26 at 4 39 45 pm


murrant:
Not sure I like the way this is fixed.   I think they should just be displayed in the timezone of the browser accessing them.

murrant:
http://www.jquery-bootgrid.com/documentation#converters

murrant:
If you don't want to force it on the user, you should add a timezone setting to the user settings.  With browser timezone as the default.

That is a much larger scope.  Up to you.

crcro:
@murrant: so I made it like:

if no timezone is set in config, print default user timezone using moment.js.
if timezone is set then display using php date functions.

is this better?

laf:
Any reason why you've removed my name from the copyright?

crcro:
By mistake, copy/paste the entire copyright section, restored it. Sorry

murrant:
Yeah, seems a like a good compromise for now.  Thanks.

html/includes/table/graylog.inc.php
```diff
@@ -1,7 +1,6 @@
 <?php
-
 /*
- * LibreNMS
+ * Librenms
```
laf: Really petty but this should still be `LibreNMS`

PR #7805, opened 2017-11-27
Added info on using OIDs in skip_values and entity for sensors

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This adds a missing doc update about skip_values and also adds support for entPhysicalIndex which is needed for another PR (same patch will be in that PR but it should rebase nicely - just thought I'd add it here).


murrant:
Just need to rebase this and should be good to go.

Issue #7644, opened 2017-11-06
Support for devices NetPIng


YuliaKoroleva:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [ x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ x]Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

Hi!
There is a task to connect the device Uniping server solution v3 / SMS to LibreNMS to receive data from temperature, humidity sensors and so on.
Exchange is performed using SNMP v.2 protocol
Device - http://www.netpingdevice.com/products/uniping-server-solution-v3-sms
MIB - http://www.netping.ru/Pub/Firmwares/[Pub]%20DKSF%2070.6.6%20MB.mib

Component | Version
LibreNMS  | 1.33-39-g21495a8
DB Schema | 215
PHP       | 7.0.22-0ubuntu0.16.04.1
MySQL     | 10.0.31-MariaDB-0ubuntu0.16.04.2
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3

Discovery.php - https://p.libren.ms/view/6ec607fb
Poller.php - https://p.libren.ms/view/3745676d
snmpbulkwalk - https://p.libren.ms/view/4e3d03b5

I apologize for the inconvenience.

kkrumm1:
You didn't run` ./discovery.php HOSTNAME -d` correcly. 

You need to replace HOSTNAME with the name of your device or ip.

YuliaKoroleva:
Sorry.
Discovery - https://p.libren.ms/view/febfb2aa
[...]

PR #7807, opened 2017-11-27
device: Added support for NetPing UniPing devices

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7644 

I've excluded a lot of sensors based on value as it seems like they don't exist.

image


includes/definitions/discovery/uniping.yaml
```diff
@@ -0,0 +1,108 @@
...
+                -
+                    oid: npSmokeTable
+                    value: npSmokeStatus
+                    num_oid: .1.3.6.1.4.1.25728.8200.1.1.2.
+                    descr: 'Smoke {{ $index }}'
+                    skip_values: 5
+                    index: 'npSmokeStatus.{{ $index }}'
+                    state_name: npSmokeStatus
+                    states:
+                        - { descr: ok, graph: 0, value: 0, generic: 0 }
```
murrant: These are a little easier to read if you put descr last ;)

laf: Our docs still have that way: https://docs.librenms.org/#Developing/Sensor-State-Support/

murrant: haha, I didn't change all of them.  I only changed some.  I'll fix that up.

PR #7809, opened 2017-11-27
webui: devices table refresh

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

webui unification, in progress, started from @Rosiak PR #7628 and @laf suggestion.
notes: uses css class and js function from PR #7765

before:
screen shot 2017-11-28 at 12 51 08 am

after:
screen shot 2017-11-28 at 12 50 18 am

### DIFF
--- a/html/includes/table/devices.inc.php
+++ b/html/includes/table/devices.inc.php
@@ -135,10 +135,8 @@ foreach (dbFetchRows($sql, $param) as $device) {
 
     if ($device['status'] == '0') {
         $extra = 'label-danger';
-        $msg = $device['status_reason'];
     } else {
         $extra = 'label-success';
-        $msg = 'up';
     }
 
     if ($device['ignore'] == '1') {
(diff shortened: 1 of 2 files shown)


murrant:
You also disabled the links to the graphs view here.  I don't think there is another way to get to that view.

crcro:
@murrant ... it was there but not in the screen capture, nonetheless i've updated the layout atm:

screen shot 2017-12-05 at 12 45 52 am

laf:
The layout was a bit 'rubbish' before because of what we tried to fit on. That doesn't work quite so well now:

image

Have to say though, overall it looks a lot cleaner. Something needs to be done about the up / down / icmp / ping status column. That doesn't look nice now. I'm not sure what to suggest other than dropping the up / down and only having it say if it's icmp or ping (i.e if it is down) or wrap the text in something that makes it look a little slicker.

murrant:
Maybe putting up/down in a label like it was before (but not as big)

PR #7825, opened 2017-11-30
Feature: (API) Retrieve BGP sessions by ID

Lee303:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This PR adds an API endpoint for retrieving BGP sessions by ID.

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -523,8 +523,8 @@ function get_bgp()
     if (!is_numeric($bgp_session_count)) {
         api_error(500, 'Error retrieving BGP peer');
     }
-    if (count($bgp_session_count) == 0) {
-        api_error(404, 'BGP peer $bgpPeerId does not exist');
+    if ($bgp_session_count == 0) {
+        api_error(404, "BGP peer $bgpPeerId does not exist");
     }
 
     api_success($bgp_session, 'bgp_session');
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -511,8 +511,9 @@ function get_bgp()
     check_is_read();
 
     $app        = \Slim\Slim::getInstance();
+    $router = $app->router()->getCurrentRoute()->getParams();
 
-    $bgpPeerId        = $_GET['id']
+    $bgpPeerId        = $router['id'];
     if (!is_numeric($bgpPeerId)) {
         api_error(400, 'Invalid id has been provided');
     }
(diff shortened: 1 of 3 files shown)


html/includes/api_functions.inc.php
```diff
@@ -506,6 +506,31 @@ function list_bgp()
...
+
+    $app        = \Slim\Slim::getInstance();
+    $router = $app->router()->getCurrentRoute()->getParams();
+
+    $bgpPeerId        = $router['id'];
+    if (!is_numeric($bgpPeerId)) {
+        api_error(400, 'Invalid id has been provided');
+    }
+
+    $bgp_session       = dbFetchRows("SELECT `bgpPeers`.* FROM `bgpPeers` LEFT JOIN `devices` ON `bgpPeers`.`device_id` = `devices`.`device_id` WHERE `bgpPeerState` IS NOT NULL AND `bgpPeerState` != '' AND bgpPeer_id = ?", array($bgpPeerId));
```
murrant: Why are you joining the devices table here if you only return data from the bgpPeers table?

html/includes/api_functions.inc.php
```diff
@@ -506,6 +506,31 @@ function list_bgp()
...
 
 
+function get_bgp()
+{
+    check_is_read();
+
+    $app        = \Slim\Slim::getInstance();
+    $router = $app->router()->getCurrentRoute()->getParams();
+
+    $bgpPeerId        = $router['id'];
```
murrant: How do you figure out the bgpPeer_id via the API?

What if I want to look up by bgpPeerIdentifier ?

html/includes/api_functions.inc.php
Lee303: My bad, this was copied and adapted from the list_bgp() function. Will fix shortly.

html/includes/api_functions.inc.php
Lee303: This is returned from GET /api/v0/bgp (list_bgp()), and I chose this for get_bgp() as this is the primary key and uniquely identifies the BGP session. I believe bgpPeerIdentifier would be best suited to to an additional filter for list_bgp()? Apologies if I've missed the mark here, this is my first venturing into librenms :)

laf: > What if I want to look up by bgpPeerIdentifier ?

That's not a unique reference within the DB.

laf:
Thanks for doing this spotts, I'll give this a quick test later :)

html/includes/api_functions.inc.php
laf: Fixed for you :)

PR #7826, opened 2017-11-30
device: Rittal IT Chiller / Carel pCOweb card

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---------------------------------------
**Note that the MIB file is ugly, but I've finally found a solution that's compatible with the YAML during the discovery AND the polling.**

- Added support for the Rittal IT Chiller by the pCOweb card
- Added documentation about the pCOweb card
- Corrected a bad SQL schema file that I've injected in a past commit

screen shot 2017-11-30 at 15 12 26


murrant:
Two suggestions off the top:
1. E.E.R. should not be abbreviated in the graph heading
2. E.E.R. should probably be a new sensor type.  (see an example here: https://github.com/librenms/librenms/pull/7714/files)

FTBZ:
I've a problem with the colling graphs. The values appear in the device Overview (disco and poll seem to be working), but the Free Cooling Capacity and Total cooling Capacity graphs are not working. It's showing -nanW as value. Compressor Cooling Capacity is working fine. Idea?

laf:
Post the output of `./poller.php -h HOSTNAME -d -m sensors`

FTBZ:
https://p.libren.ms/view/cc7837da

laf:
That output seems to suggest the data is recording ok. In the device overview page, next to the graph for this sensor it also displays the text value, is this correct?

FTBZ:
The numbers in the overview are always correct. But the graph does strange things (only the graph for a few minutes):
screen shot 2017-12-06 at 08 20 48

laf:
We do only allow a range of -20000 - 20000 for sensors, I wonder if it's going beyond that.

FTBZ:
Correct, the value is 29300W.

FTBZ:
Nice, it's working.

laf:
I've made some changes but looks good to me.

Question though, what's eer and does it not belong as a sensor?

FTBZ:
EER -> Energy Efficiency Ratio for cooling systems. It tells you if you're efficient or not. Yes, it can be a sensor but not existing at this time.

laf:
Then we need to shift that into sensors, if you want me to do that give me a shout. We've got a few PRs recently that add this type of support.

FTBZ:
If you can take care of it, it would be nice. I watched the PR but I'm not comfortable with it (not merged at this time). Another thing that's missing it's a sensor for water flow rate in l/m. If you can create a PR with only one sensor, I can use it as an example to create one of the two.

laf:
We've got an other PR in adding more sensors, will do it after that to save merge conflicts.

murrant:
@FTBZ Other PR is now merged :)

FTBZ:
Can this PR be merged like this and I check the two new sensors in a separate one?

laf:
Sure, just needs a rebase to fix the merge conflict :)

laf:
One last question before merging, is this the best name for this OS? Seems very hardware specific.

FTBZ:
The name was chosen by @murrant to track all device of this kind.

PR #7828, opened 2017-11-30
feature: save application metrics to db for alerting

murrant:
However, alerting will not work because ResolveGlues() is broken.
Can add workaround after state_translations alerting is merged

Does not update all applications yet, not sure if that should be done here or in another PR. Most are simple, some are complex.

Introduces two handy functions dbDeleteOrphans() and array_by_column().  Will replace those in other locations after this is merged or separate them out if this is not merged.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Missing the schema file?

I can't help but feel we're wasting time with this :(

murrant:
Why do you feel it is a waste of time?

Yeah, forgot schema :(

Issue #7814, opened 2017-11-28
Additional sysContact email address validation


nrcbhermsen:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

```
====================================
Component | Version
--------- | -------
LibreNMS  | 1.33-169-ge1bb326
DB Schema | 217
PHP       | 7.0.24
MySQL     | 5.5.56-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
====================================

```
Need to provide additional validation of sysContact field to not only check if it exists but whether the value is a valid email address.  Currently if the field has a value in it but it is not an email address the email alerts will not fall back to the default contact.  Using the following example will result in a failure as the sysContact exists but is not a valid email address:

```
array(1) {
    ["Systems Engineering"]=>
    string(3) "NOC"
```

When the email alert attempts to fire off it does catch that the given contact is not an email address and fails:

`mail => ERROR: You must provide at least one recipient email address.
`

laf:
Can you give this patch a try please:

https://p.libren.ms/view/raw/04728aac

nrcbhermsen:
Patch appears to work beautifully.  The default contact got the email as expected.

laf:
@nrcbhermsen #7830 submitted. Small change in it but only to check an array isn't empty first. Thanks for testing.

PR #7830, opened 2017-11-30
fix: Validate email addresses used in alerting

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7814

### DIFF
--- a/includes/alerts.inc.php
+++ b/includes/alerts.inc.php
@@ -233,6 +233,7 @@ function RunRules($device)
 function GetContacts($results)
 {
     global $config, $authorizer;
+
     if (sizeof($results) == 0) {
         return array();
     }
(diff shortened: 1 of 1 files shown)


kkrumm1:
We put Asset Tags into the sys contact on our switches. This shouldn't break anything?

laf:
As in for the email address or the name?

We only check the something@domain.com part so if you have something@domain.com <asset tag> then it's fine.

If the asset tag is the email address but it's not a genuine email address then this will no longer attempt to send emails to that address (which is good) but if you have the correct options set, we fall back to the default contact you have defined.

kkrumm1:
okay it's just a string of numbers. Should be okay then. 👍 

image

laf:
So that would never have sent an email before anyway. This will mean that you still have a chance to fall back to a valid email so it's a good thing :)

PR #7844, opened 2017-12-03
Refactor alert transports to classes

mcq8:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Only moved the code to classes to have less ev[a|i]l code.
Next step could be to have the $obj as an actual object so integrators know what to expect from the var.
I'm not sure about the function name call. If you have any suggestions let me know.


mcq8:
Seems like the Slack alerting is already making the project incompatible with php < 5.4.
I'll move it to the old array style and fix the linting.

laf:
Running on my install for testing.

laf:
Testing on mail transport works. Will code review soon.

laf:
LGTM.

@librenms/reviewers

LibreNMS/Interfaces/Alert/Transport.php
```diff
@@ -0,0 +1,38 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2017 Tony Murray
```
murrant: Wrong copyright here, pretty sure I didn't right it :)

LibreNMS/Interfaces/Alert/Transport.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+namespace LibreNMS\Interfaces\Alert;
+
+interface Transport
+{
+    /**
+     * Gets called when an alert is sent
+     *
+     * @param $obj array An array created by DescribeAlert
```
murrant: Can you rename this from `$obj` to `$alert_data`, or something like that.

You don't have to update all the transports.

LibreNMS/Interfaces/Alert/Transport.php
```diff
@@ -0,0 +1,38 @@
...
+     * @param $opts array|true The options from $config['alert']['transports'][$transport]
+     * @return bool Returns if the call was successful
+     */
+    public function call($obj, $opts);
```
murrant: Not sure I like the function name here.
Options I can think of:
submitAlert()
deliverAlert()
sendMessage()

LibreNMS/Interfaces/Alert/Transport.php
murrant: It would also be great if you documented the fields of `$obj` here

murrant:
Great work @mcq8 Just a few cosmetic suggestions.

LibreNMS/Interfaces/Alert/Transport.php
mcq8: I would rather do the documentation trough making an object out of the alert, but I'll do that trough a separate pull request.

InsaneSplash:
I didn't test the old php version, but moving to php 7.1 resolved the problem.

PR #7848, opened 2017-12-03
fix: Some graphs broke due to stacked graphs, remove that change

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

@crcro can you check this please. I've removed the transparency on this as this doesn't work in rrdtool 1.4. I can't see a difference before and after this patch in 1.6 so not sure what the intention was with it.


crcro:
@laf ... it should be visible only if using stacked graphs (have no issues with it, but also runnning rrdtool 1.7)

screen shot 2017-12-04 at 7 51 55 am

screen shot 2017-12-04 at 7 52 23 am

laf:
It is only used when stacked graphs are enabled but it's not compatible with rrdtool 1.4.x which we need to support.

crcro:
@laf ... able to reproduce bug, this seems to fix it.

laf:
Thanks @crcro

Issue #7616, opened 2017-11-02
Add CPU and Memory detecion for Edge-Core switches


kamils85:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.

Component | Version
--------- | -------
LibreNMS  | 1.33-23-gfb45f00
DB Schema | 215
PHP       | 7.0.15-0ubuntu0.16.10.4
MySQL     | 10.0.29-MariaDB-0ubuntu0.16.10.1
RRDTool   | 1.6.0
SNMP      | NET-SNMP 5.7.3

- [ ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ ] Please provide as much detail as possible.
- [ ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

According to this: https://github.com/librenms/librenms/pull/5466 the CPU support was added but it does not work with ECS4120-28F switches as they use different OID for CPU polling.

Not sure if the memory polling was ever added but it's not working either. 
Can this be added please?

Below the OIDs for CPU/Mem
[edgecos-cpu-mem.txt](https://github.com/librenms/librenms/files/1439351/edgecos-cpu-mem.txt)

Here you can find snmpwalk and up-to-date MIB for ECS: https://github.com/librenms/librenms/issues/7140

Thanks,

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

laf:
Thanks for the bug report or new device request, unfortunately you have not provided the information asked for in point 3 when you created this issue.

Please provide pastebin links to the information we need and we will re-open this issue.

kamils85:
Please find all the requested below:

discovery: https://p.libren.ms/view/e60cee9f
poller: https://p.libren.ms/view/a40d0f86
[...]

PR #7850, opened 2017-12-04
Added cpu and mempolls EdgeCore switch

erotel:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

ECS4510
ECS3528
ECS4120 - NOT TESTED (I do not have this device)
ECS4210
ECS3510

Fixes: #7616


includes/discovery/mempools/edgecos.inc.php
```diff
@@ -0,0 +1,57 @@
...
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+if ($device['os'] == 'edgecos') {
+    d_echo('EdgeCore Memory:');
+
+    $temp_total = snmp_get($device, '.1.3.6.1.4.1.259.10.1.24.1.39.3.1.0', '-Ovqn');//ECS4510
+    if (!empty($temp_total)) {
+         $total = snmp_get($device, '.1.3.6.1.4.1.259.10.1.24.1.39.3.1.0', '-Ovqn');
```
laf: No point in running this snmp_get, you already have this data.

You then should check if `$total and $avail` are set for the rest of these so you don't continuously send snmp gets if you match early on.

Also, can you not do anything with sysObjectID to work out what the OIDs are?

includes/discovery/processors/edgecos.inc.php
```diff
@@ -13,10 +13,38 @@
 if ($device['os'] == 'edgecos') {
     echo 'Edgecos : ';
 
-    $oid = '.1.3.6.1.4.1.259.10.1.24.1.39.2.1.0';
-    $descr = 'Processor';
-    $usage = snmp_get($device, $oid, '-Ovqn');
+    $usage = snmp_get($device, '.1.3.6.1.4.1.259.10.1.24.1.39.2.1.0', '-Ovqn');//ECS4510
```
laf: Same goes for this file really, no point in running snmp_gets all the time, just check if `$oid` is set before you do.

Check you can't sysObjectID to work these out as well. Will save a lot of code and snmp queries.

includes/polling/mempools/edgecos.inc.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Copyright (c) 2017 Martin Zatloukal <slezi2@pvfree.net>
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+if (starts_with($device['sysObjectID'], 'enterprises.259.10.1.24.')) { //ECS4510
```
laf: Ok so you've answered my question above, you can use the sysObjectID - you need to port this to discovery.

Can you not used the named OIDs in all of this?

You should convert these to `snmp_get_multi_oid()` when running more than one get.

includes/discovery/mempools/edgecos.inc.php
erotel: The use of sysObjectID is not possible because it is not known yet in discovery

```
#### Load disco module processors ####
Array
(
    [device_id] => 39
    [hostname] => 192.168.224.254
    [sysName] => 192.168.224.254
    [ip] =>.
    [community] => public
    [authlevel] =>.
    [authname] =>.
    [authpass] =>.
    [authalgo] =>.
    [cryptopass] =>.
    [cryptoalgo] =>.
    [snmpver] => v2c
    [port] => 161
    [transport] => udp
    [timeout] =>.
    [retries] =>.
    [snmp_disable] => 0
    [bgpLocalAs] =>.
    [sysObjectID] =>.
    [sysDescr] =>.
    [sysContact] =>.
    [version] =>.
    [hardware] =>.
    [features] =>.
    [location] =>.
    [os] => edgecos
    [status] => 1
    [status_reason] =>.
    [ignore] => 0
    [disabled] => 0
    [uptime] =>.
    [agent_uptime] => 0
    [last_polled] =>.
    [last_poll_attempted] =>.
    [last_polled_timetaken] =>.
    [last_discovered_timetaken] =>.
    [last_discovered] =>.
    [last_ping] =>.
    [last_ping_timetaken] =>.
    [purpose] =>.
    [type] => network
    [serial] =>.
    [icon] => edge-core.png
    [poller_group] => 0
    [override_sysLocation] => 0
    [notes] =>.
    [port_association_mode] => 1
    [attribs] => Array
        (
        )

    [snmp_max_repeaters] =>.
)
```

laf: It should be and if it isn't / doesn't we should make sure it is.

You could however just do the sysObjectID lookup once anyway to still save all of this but I favor fixing disco.

laf: Should already be available in `$device['sysObjectID']`

erotel: When adding a new device, $ device ['sysObjectID'] is only available after the first polling

murrant:
@erotel here is how these things should work.  In discovery, you should figure out the numeric oid that holds the data, save that in the discover_x() call and then the poller will just use that to poll it without any tricky logic.

laf:
> Rather than depending on sysObjectID, you could snmpget an oid and if that doesn't return data, try another oid.

In this case though that's not the best course of action. Using sysObjectID to build the query is better as it then only needs one get or walk, we just need to make sure sysObjectID is available, which it should be.

murrant:
I'm a bit cynical as I've been looking though all the processor discovery files, many are broken to some degree.

Sure you can use sysObjectID to save a query, but I don't like the way it is used here.  Using it to select certain models leaves out unknown or future models.

erotel:
Maybe I misunderstood (my English is terrible). In my opinion sysObjectID snmpget in discovery saves the number of queries. sysObjectID in the poller is right for me. Why load snmp oid when the type can be determined by the sysObjectID we know. This saves a lot of snmp queries (up to 4 queries for the latest model).
As for the new or unknown model, support must be added to yaml and other files.

murrant:
Sure, but please support devices you have not explicitly defined.

Also, you can access sysObjectID here `$device['sysObjectID']`

erotel:
The oid is always for a particular line.
eg SNMPv2-SMI :: enterprises.259.10.1.24. is for the ECS4510 series of switches.
  tested on ECS4510-52T ECS4510-28F

When adding a new device, $ device ['sysObjectID'] is only available after the first polling

laf:
> Sure you can use sysObjectID to save a query, but I don't like the way it is used here. Using it to select certain models leaves out unknown or future models.

Which would have been the same with querying the OID anyway, the difference here is we aren't attempting to query what is possibly an unused OID until we find the one we want.

However, we can fix the concern here by just appending the last few octets of the OID to sysObjectID as it's only that bit which changes, that will then cover for any that we don't know about.

murrant:
> When adding a new device, $ device ['sysObjectID'] is only available after the first polling

I think that should be fixed :-). But it is a little tricky.  Please use that for now and let it take two discovers for new devices.

Seems reasonable @laf

laf:
> I think that should be fixed :-). But it is a little tricky. Please use that for now and let it take two discovers for new devices.

It definitely should. I'd like to try and fix it for this PR so leave it with me.

laf:
#7922 adds `$device['sysObjectID']` and `$device['sysDescr']`, when it's merged you can update this PR to use those (or update before hand ready).

laf:
This just needs to wait for #7922 now and then an update to be made to the code. `enterprises.259.10.1.24.` would now be the full numerical OID even in polling. The linked PR standardises the use of `$device['sysObjectID']` throughout the codebase.

Will leave the blocker label in place until ready.

laf:
Thanks @erotel. Just needs the other PR merging now before this can go in.

murrant:
@erotel The sysObjectID PR is now merged.  But, we also merged a functional testing framework.

Would you be willing to add some test data?

You would probably need a few variants to represent the different types you have here.

erotel:
I hope I understand what you need.
But the script tells me the error.

[error save-test-data.php](https://pastebin.com/XRTYzpMs)

murrant:
@erotel you need to install snmpsim to generate the json files.  If you want, I might be able to do that from the snmprec files you uploaded.

murrant:
Also, you need to rebase this to pull in the sysObjectID changes (and delete cache/os_defs.cache after you do that).

erotel:
@murrant  Please.
Somehow I can not install snmpsim.

murrant:
@erotel the processor oids are not include in your previously uploaded snmprec files because you had not rebased.  You'll need to rebase and re-capture the data.  Verify the processor oid is present in each snmprec ;) 

I forgot about mempools too.  So when yo urun the capture do `-m processors,mempools`

laf:
@erotel Still needs the data @murrant mentioned in his last comment :)

erotel:
I added new json files where it was used
`/opt/librenms/scripts/save-test-data.php -h 192.168.224.192 -m  processors,mempools`

laf:
I could be missing it but I don't see the data in the snmprec file though for mem/cpu.

laf:
Awesome work @erotel. Thanks for doing this

PR #7855, opened 2017-12-04
Leaflet upgrade

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Tested with my dev install, worked ok. Can't guarantee outside of that :)


murrant:
Why do we include all the source code?

laf:
> Why do we include all the source code?

That's how blahdeblah wanted it done in the past. I'd prefer to just include the files we need as we have a lot of bloat in the html/js and lib/ folders.

I'm merging this in though so we can fix for now and deal with later.

PR #7856, opened 2017-12-05
docs: update world map doc

kkrumm1:
Update the World Map Doc. It was very confusing and had lots of old info. 

#### Please note

#### Testers


laf:
Think it needs more than that. Afaik the doc currently covers two maps we have available.

Worldmap and Globemap are different.

kkrumm1:
I see okay. Maybe should break it down into sections for each map.

kkrumm1:
I know some parts have config but they are already on by default in the default config. Should we keep that info on the doc? It's just going to confuse users thinking they have to add that config.php

kkrumm1:
okay, this going take allot more than i thought. I need to see how each Map works. Right now it's still very confusing.

PR #7861, opened 2017-12-05
rules: default alert rules, updated the name for

kkrumm1:
Sensor Over/Under to help users understand where this alert is coming from. :)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/misc/alert_rules.json
+++ b/misc/alert_rules.json
@@ -35,12 +35,12 @@
   },
   {
     "rule": "%sensors.sensor_current > %sensors.sensor_limit && %sensors.sensor_alert = \"1\" && %macros.device_up = \"1\"",
-    "name": "Sensor over limit",
+    "name": "Sensor over limit - Check Device Health Settings",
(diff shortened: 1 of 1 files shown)


murrant:
Possibly, not the best place for this.  Will not update existing rules for example.

What about detecting if there are multiple eventlog entries and then adding a custom message to the next eventlog? or something like that.

kkrumm1:
@murrant that sounds like a way better idea. :)

murrant:
How about an edit link in the eventlog, I wonder if it is reasonable to do that.

kkrumm1:
Seems reasonable and helpful. idk about the coding part. :-)

laf:
> Possibly not the best place for this. Will not update existing rules for example.

I'm not sure we should bother about old rules people have created. If users want to update existing rules then they can.

I'm fine with this as is, adding links in to eventlogs is over complicating this PR but could be good idea.

PR #7865, opened 2017-12-06
docs: create doc for IP-MAC-Address-Search-Feature

kkrumm1:
I know some of this common networking knowledge, I just thought it would nice to have a doc on it. :)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
I'm wondering how best to do things like this, we've got a lot of links already and personally I was hoping we'd reduce those rather than add to them.

The doc itself is possibly useful.

kkrumm1:
I was thinking of a Feature Section?

laf:
I'm still not overly happy with our docs and the layout so I'm not sure how to approach this right now. Leave it open for people to have a think.

murrant:
How about a blog?

murrant:
I am pretty certain things like this should be a blog post.  Would be much appreciated if you could convert it to one @kkrumm1. :)

PR #7868, opened 2017-12-06
doc: create device sensor help doc.

kkrumm1:
this should help users that get confused when receiving alerts for sensors.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Support/Device-Sensors.md
+++ b/doc/Support/Device-Sensors.md
@@ -2,7 +2,8 @@ source: Support/Device-Sensors.md
 
 ## Device Sensors 
 
-LibreNMS has a standard for device sensors they are split into groups.
+LibreNMS has a standard for device sensors they are split into categories.
+This doc is to help users understand device sensors in general, if you need help with developing sensors for a device please see the Contributing + Developing section. 
 
 ### Health Sensors
 
--- a/doc/Support/Device-Sensors.md
+++ b/doc/Support/Device-Sensors.md
@@ -1,0 +1,47 @@
+source: Support/Device-Sensors.md
+
+## Device Sensors 
+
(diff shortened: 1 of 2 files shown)


doc/Support/Device-Sensors.md
```diff
@@ -0,0 +1,49 @@
...
+
+LibreNMS has a standard for device sensors they are split into categories.
+This doc is to help users understand device sensors in general, if you need help with developing sensors for a device please see the Contributing + Developing section. 
+
+### Health Sensors
+
+The High and Low values of these sensors can be edited in Web UI by going to the device settings -> Health. There you can set your own custom
+High and Low values. List of these sensors can be found here [Link](../Developing/os/Health-Information.md)
+
+**Note** Some values are defined by the manufactures and other calculated when you add the device into librenms and some are set to null leaving it up
```
laf: None should be null, any in the code that are null will be auto calculated so you only have device defined or calculated ones.

doc/Support/Device-Sensors.md
```diff
@@ -0,0 +1,49 @@
...
+the user define the values, every environment is different and may require user input. 
+
+
+### Wireless Sensors
+
+Some Wireless have  High and Low values of these sensors can be edited in Web UI by going to the device settings -> Health. There you can set your own custom
+High and Low values. List of these sensors can be found here [Link](../Developing/os/Wireless-Sensors.md)
```
laf: I think these auto calculate as well if the devices don't tell us what they are - is that right @murrant

kkrumm1:
@laf good to go?

laf:
Go for it :)

PR #7881, opened 2017-12-09
fix: remove faulty memcached code (not related to distributed polling)

murrant:
Seen this cause:
slow page loads
incorrect (out of date) data
massive amounts of memory usage

I can't think of  a single reason to enable it.  Varnish is probably much more useful.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Absolutely no issues with this from me, how about adding it to validate to tell people it's deprecated if we detect the config is set and maybe a notification to also let people know?

murrant:
The main problem I have is I don't want to confuse people that use distributed polling.   If we do a notification it will need to be worded carefully.

laf:
> The main problem I have is I don't want to confuse people that use distributed polling. If we do a notification it will need to be worded carefully.

That's actually a pretty good reason.

PR #7891, opened 2017-12-11
Agent tries to insert processes data that is too long

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Do we want to strip that field down or just make the column bigger / text type?

murrant:
Not sure, I defaulted to truncating the field since that is better than the sql (which inserts all processes at once) failing.

laf:
Some people don't know what they were missing until it appears.

Personally I'd move it to a text column as we shouldn't be manipulating data like that imho.

PR #7895, opened 2017-12-11
Add API method to rename devices

thomseddon:
Not sure on how you'd like this method to be presented? Let me know if you'd prefer the new hostname to be in the body (I just wanted some kind of differentiation).

#### Please note

#### Testers

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -1262,6 +1262,30 @@ function update_device()
     }
 }
 
+function rename_device()
+{
+    check_is_admin();
+    global $config;
+    $app = \Slim\Slim::getInstance();
+    $router = $app->router()->getCurrentRoute()->getParams();
+    $hostname = $router['hostname'];
+    $device_id = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
+    $new_hostname = $router['new_hostname'];
+    $new_device = getidbyname($new_hostname);
+
+    if (empty($new_hostname)) {
+        api_error(500, 'Missing new hostname');
+    } elseif ($new_device) {
+        api_error(500, 'Device failed to rename, new hostname already exists');
+    } else {
+        if (renamehost($device_id, $new_hostname, 'api') == '') {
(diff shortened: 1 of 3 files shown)


html/includes/api_functions.inc.php
```diff
@@ -1262,6 +1262,41 @@ function update_device()
...
+        $message = 'Missing new hostname';
+    } elseif ($new_device) {
+        $message = 'Device failed to rename, new hostname already exists.';
+    } else {
+        if (renamehost($device_id, $new_hostname, 'api') == '') {
+            $status = 'ok';
+            $code = 200;
+            $message = 'Device has been renamed';
+        } else {
+            $message = 'Device failed to be reanmed';
```
murrant: small typo here `reanmed` -> `renamed`

laf:
Thanks @thomseddon.

The API code went through a little refactor the other week so you'll need to update the code a bit to match (I can do that if you haven't got time?)

thomseddon:
Fixed the typo, thanks for spotting that.

I've also added `check_is_admin();` to match `update_device`?

@laf what branch should I rebase from? I've just checked and it seems up to date with master on here, just let me know what I'm missing, can get it fixed today :)

laf:
> @laf what branch should I rebase from? I've just checked and it seems up to date with master on here, just let me know what I'm missing, can get it fixed today :)

@thomseddon sorry for the delay, busy week as you know.

So sending an error back can be done like: https://github.com/librenms/librenms/blob/master/html/includes/api_functions.inc.php#L255

Sending the end response can be done like: https://github.com/librenms/librenms/blob/master/html/includes/api_functions.inc.php#L263

Happy to update if you don't have time.

thomseddon:
@laf ah sorry, I see it now - my mistake!

I've switched to the new functions to form the responses (much nicer :+1:) and tided up a couple of other things (erroneous new line, unnecessary full stop in response)

Happy to make any more amends :)

laf:
Thanks for doing this @thomseddon :)

PR #7896, opened 2017-12-11
Fix: Add user level to getUser for LDAP authentication

### LINKED ISSUE #7893: LDAP (and AD?) auth getUser does not return level -> API fails
The newly added fake session for API authentication assumes that getUser will also return the user level. It does not for LDAP authentication, and probably neither for AD authentication. I don't know if the API fakes session should be changed (see https://community.librenms.org/t/new-api-restrictions-and-ldap-userlevel-issue/2772/2), or the user level should be added to all getUser functions.

API:
https://github.com/librenms/librenms/blob/9d66d7c8976adc175d1ac9a18537cab4a73cf4d6/html/includes/api_functions.inc.php#L25-L32

LDAP:
https://github.com/librenms/librenms/blob/6b5dccc169d529204066b5174392c8b74bd23bc0/LibreNMS/Authentication/LdapAuthorizer.php#L159-L164

AD:
https://github.com/librenms/librenms/blob/6b5dccc169d529204066b5174392c8b74bd23bc0/LibreNMS/Authentication/ActiveDirectoryAuthorizer.php#L219

Ultra2D:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes #7893.

### DIFF
--- a/LibreNMS/Authentication/LdapAuthorizer.php
+++ b/LibreNMS/Authentication/LdapAuthorizer.php
@@ -161,7 +161,6 @@ class LdapAuthorizer extends AuthorizerBase
                                 'realname' => $realname,
                                 'user_id' => $user_id,
                                 'email' => $email,
-                                'level' => $this->getUserlevel($username),
                             );
                         }
                     }
@@ -178,6 +177,7 @@ class LdapAuthorizer extends AuthorizerBase
     {
         foreach ($this->getUserlist() as $user) {
             if ($user['user_id'] === $user_id) {
+                $user['level'] = $this->getUserlevel($user['username']);
                 return $user;
             }
         }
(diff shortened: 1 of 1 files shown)


LibreNMS/Authentication/LdapAuthorizer.php
```diff
@@ -161,6 +161,7 @@ public function getUserlist()
                                 'realname' => $realname,
                                 'user_id' => $user_id,
                                 'email' => $email,
+                                'level' => $this->getUserlevel($username),
```
murrant: This is going to make getUserlist() even slower. Can you please add it on the getUser() function and make mark a TODO here.

Ultra2D: I am not sure why a TODO should be added to getUserlist (and what to write as a description of the TODO).

Maybe this fix should be added to #7718?

laf:
@Ultra2D You will need to sign the CLA before we can merge this. Should just be a case of clicking the link in the comment from the bot above.

Issue #7894, opened 2017-12-11
 sysObjectID not updating for Cambium cnPilot


vinitkhandagle:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [ ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

After updating Cambium cnPilot shows as generic Device below is the 
`./discovery.php -h HOSTNAME -d ` output

[PasteBin url of `discovery.php`](https://pastebin.com/5NTtcEqC)

pheinrichs:
I'll try and add proper support for this today.

murrant:
@vinitkhandagle can you provide all info requested?

Especially the snmpwalk?

vinitkhandagle:
information:

discovery.php
https://p.libren.ms/view/0b5e1896

poller.php
https://p.libren.ms/view/d4e0d3e3

snmpbulkwalk
https://p.libren.ms/view/3a9e0a2a

PR #7898, opened 2017-12-11
device: Cambium cnPilot Support

pheinrichs:
The cnPilot routers(cnpilotr) don't have a lot of info coming from the mibs. The E model have much better mibs and I've included a few wireless sensors to start it off, will need testing though.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7894


includes/polling/os/cnpilotr.inc.php
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+
+$version = trim(snmp_get($device, '.1.3.6.1.4.1.41010.1.1.1.0', '-OQv'), '"');
```
murrant: You don't need to trim quotes anymore, snmp_get does that.

includes/definitions/cnpilotr.yaml
```diff
@@ -0,0 +1,12 @@
...
+text: Cambium cnPilot Router
+type: network
+icon: cambium
+mib_dir: cambium
+over:
+    - { graph: device_bits, text: 'Device Traffic' }
+group: cambium
+discovery:
+    -
+        sysDescr_regex:
```
murrant: For completeness I would add sysObjectId: .1.3.6.1.4.1.17713.22 here.

includes/definitions/cnpilote.yaml
```diff
@@ -0,0 +1,11 @@
...
+text: Cambium cnPilot
+type: wireless
+icon: cambium
+mib_dir: cambium
+over:
+    - { graph: device_bits, text: 'Device Traffic' }
+group: cambium
+discovery:
+    - sysObjectId:
+            - .1.3.6.1.4.1.17713.22
```
murrant: I would also add sysDescr_except: '/(cnPilot R[0-9][0-9][0-9])/i' here.

LibreNMS/OS/Cnpilote.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Discover wireless tx or rx power. This is in dBm. Type is power.
+     * Returns an array of LibreNMS\Device\Sensor objects that have been discovered
+     *
+     * @return array
+     */
+    public function discoverWirelessPower()
+    {
+        $oid = '.1.3.6.1.4.1.17713.22.1.2.1.8.0'; //CAMBIUM-MIB::cambiumRadioTransmitPower.0
```
murrant: When I looked a the MIB a bit earlier, I noticed these are table entries.  Could there be more than one point of data returned (multiple radios)?

Do you have a device to test or are you just adding based on the MIB @pheinrichs ?

tests/snmpsim/cnpilotr_1.snmprec
```diff
@@ -0,0 +1 @@
+1.3.6.1.2.1.1.1.0|4|cnPilot R201 4.3.3-R4
```
murrant: Please add both sysObjectID and sysDescr for snmprec files.

includes/definitions/cnpilotr.yaml
pheinrichs: This device's sysObjectId only returns  .1.3.6.1.4. which is basically useless otherwise i would have included it.

tests/snmpsim/cnpilotr_1.snmprec
pheinrichs: Same situation as above. SysObjectID is a bust...

LibreNMS/OS/Cnpilote.php
pheinrichs: I don't have any devices to test on the cnpilot-e devices. This is purely on the mib

includes/definitions/cnpilote.yaml
```diff
@@ -0,0 +1,11 @@
+os: cnpilote
+text: Cambium cnPilot
+type: wireless
+icon: cambium
+mib_dir: cambium
+over:
+    - { graph: device_bits, text: 'Device Traffic' }
```
murrant: I like to add the client count to wireless devices overview:
```
- { graph: device_wireless_clients, text: 'Number of Clients' }
```

LibreNMS/OS/Cnpilote.php
```diff
@@ -0,0 +1,98 @@
...
+namespace LibreNMS\OS;
+
+use LibreNMS\Device\WirelessSensor;
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessClientsDiscovery;
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessSnrDiscovery;
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessPowerDiscovery;
+use LibreNMS\Interfaces\Discovery\Sensors\WirelessNoiseFloorDiscovery;
+use LibreNMS\OS;
+
+class Pmp extends OS implements
```
murrant: Class name is wrong. Should be Cnpilote

murrant:
Mib files should be the name of the mib (without txt).  I noticed both have the same it doesn't look as if they are the same (man, vendors are bad at snmp).  Possibly we need to store them in different directories to prevent name collisions.

includes/polling/os/cnpilote.inc.php
```diff
@@ -0,0 +1,15 @@
...
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+
+$version = $poll_device['sysDescr'];
+$hardware = snmp_get($device, '.1.3.6.1.4.1.17713.22.1.1.1.10', '-OQv');
```
laf: Can we not use the MIBs here and then use `snmp_get_multi_oid()`?

includes/polling/os/cnpilotr.inc.php
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+
+$version = snmp_get($device, '.1.3.6.1.4.1.41010.1.1.1.0', '-OQv');
```
laf: Can we not use the MIBs here and then use `snmp_get_multi_oid()`?

laf:
> Mib files should be the name of the mib (without txt). I noticed both have the same it doesn't look as if they are the same (man, vendors are bad at snmp). Possibly we need to store them in different directories to prevent name collisions.

Can do that by putting them in sub folders and specifying both in the yaml.

Or we could just use two different names in the same folder, the filename is just to make it easier to reference - I'd vote with sub folders

tests/snmpsim/cnpilotr_1.snmprec
murrant: @pheinrichs We need to include them otherwise our testing will be worthless.
Even (especially) if it just returns an empty string.

PR #7900, opened 2017-12-11
Rewrite LibreNMS Processor support

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/common.php
```diff
@@ -1888,3 +1888,11 @@ function array_by_column($array, $column)
 {
     return array_combine(array_column($array, $column), $array);
 }
+
+/**
+ * Dump and die
+ */
+function dd() {
```
laf: Wrap this in a function exists check for when we switch to Laravel?

laf:
How come you've not used the yaml added that's commented out?

includes/common.php
murrant: @laf sorry, this was for someone to test.  Not ready for review fully yet :)

murrant:
That was for testing hrProcessors are discovered by default ;)

PR #7908, opened 2017-12-12
Fixed and added details presented to Pagerduty

timothyfranson:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---

The JSON generated for presentation to PagerDuty was not correct so the 'custom details' were not being parsed correctly by PD.  This method fixes that and additionally presents many more details to PagerDuty.  In the case of an interface alarm, things like interface stats, attributes, etc.

## The message as received by PD BEFORE:
```
{
  "client": "LibreNMS",
  "description": "some event title",
  "details": [
    "port_id => 6; ifDescr => GigabitEthernet1/0/4; "
  ],
  "event_type": "trigger",
  "incident_key": "12345",
  "service_key": "12345abcdef"
}
```

## Message as received by PD AFTER:
```
{
  "client": "LibreNMS",
  "description": "some event title",
  "details": {
    "port_id": "6",
    "ifDescr": "GigabitEthernet1/0/4",
    "port_descr_descr": "CORE-1_GI1/0/1",
    <TRUNCATED, MANY MORE DETAILS>
  },
  "event_type": "trigger",
  "incident_key": "12345",
  "service_key": "12345abcdef"
}
```

### DIFF
--- a/LibreNMS/Alert/Transport/Pagerduty.php
+++ b/LibreNMS/Alert/Transport/Pagerduty.php
@@ -42,15 +42,12 @@ class Pagerduty implements Transport
         } else {
             $protocol['event_type'] = 'trigger';
         }
+        $details = array();
         foreach ($obj['faults'] as $fault => $data) {
-            $rough_details = array_filter(explode('; ', $data['string']));
-            $details = array();
-            foreach ($rough_details as $r) {
-                $kv_pair = explode(' => ', $r);
-                $details[$kv_pair[0]] = $kv_pair[1];
-            }
-            $protocol['details'] = $details;
+            $details = $data;
+            unset($details['string']);
         }
+        $protocol['details'] = $details;
         $curl = curl_init();
         set_curl_proxy($curl);
         curl_setopt($curl, CURLOPT_URL, 'https://events.pagerduty.com/generic/2010-04-15/create_event.json');
--- a/LibreNMS/Alert/Transport/Pagerduty.php
+++ b/LibreNMS/Alert/Transport/Pagerduty.php
@@ -42,12 +42,8 @@ class Pagerduty implements Transport
         } else {
             $protocol['event_type'] = 'trigger';
         }
-        $details = array();
-        foreach ($obj['faults'] as $fault => $data) {
-            $details = $data;
-            unset($details['string']);
-        }
-        $protocol['details'] = $details;
+        $protocol['details'] = end($obj['faults']);
+        unset($protocol['details']['string']);
         $curl = curl_init();
         set_curl_proxy($curl);
         curl_setopt($curl, CURLOPT_URL, 'https://events.pagerduty.com/generic/2010-04-15/create_event.json');
--- a/LibreNMS/Alert/Transport/Pagerduty.php
+++ b/LibreNMS/Alert/Transport/Pagerduty.php
@@ -43,7 +43,13 @@ class Pagerduty implements Transport
             $protocol['event_type'] = 'trigger';
         }
         foreach ($obj['faults'] as $fault => $data) {
-            $protocol['details'][] = $data['string'];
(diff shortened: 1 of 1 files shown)


LibreNMS/Alert/Transport/Pagerduty.php
```diff
@@ -43,7 +43,13 @@ public function deliverAlert($obj, $opts)
             $protocol['event_type'] = 'trigger';
         }
         foreach ($obj['faults'] as $fault => $data) {
-            $protocol['details'][] = $data['string'];
+            $rough_details = array_filter(explode('; ', $data['string']));
```
murrant: I think there is an existing array version stored in the faults array so you don't have to parse the string.  Did you check that?

timothyfranson: I didn't see an equivalent 'array' nested element that could directly be copied.  Looking over it again though I think there is a slightly easier way to do this.  I can test and resubmit tomorrow.

```
$details = array()
foreach ($obj['faults'] as $fault => $data) {
    $details = $data;
    unset($details['string']);
}
$protocol['details'] = $details;
```

Reference lines around 497, 524
https://github.com/librenms/librenms/blob/master/includes/alerts.inc.php

murrant: That's exactly what I was thinking sorry if I explained it poorly.  I wasn't sure of the format.

LibreNMS/Alert/Transport/Pagerduty.php
```diff
@@ -42,9 +42,12 @@ public function deliverAlert($obj, $opts)
         } else {
             $protocol['event_type'] = 'trigger';
         }
+        $details = array();
         foreach ($obj['faults'] as $fault => $data) {
-            $protocol['details'][] = $data['string'];
+            $details = $data;
```
murrant: You overwrite $details each loop, is that intended or is the loop unnecessary?

timothyfranson: The loop is unnecessary I've only kept it because it was previously written that way.  This could further be shortened to
$protocol['details'] = $obj['faults'][0];
unset($protocol['details']['string']);

timothyfranson: Whew ok so looking at this a little more this morning I think the original foreach was there due to the way the foreach loops at lines 497 and 524 are written.  $i++; should be at the end of each of those loops rather than the front, yes?

So rather than point at a specific array index, we can just grab the last element.  That will make this work now and into the future should the previously mentioned loop ever get adjusted.  Recommitting...

LibreNMS/Alert/Transport/Pagerduty.php
```diff
@@ -42,9 +42,8 @@ public function deliverAlert($obj, $opts)
         } else {
             $protocol['event_type'] = 'trigger';
         }
-        foreach ($obj['faults'] as $fault => $data) {
-            $protocol['details'][] = $data['string'];
-        }
+        $protocol['details'] = end($obj['faults']);
```
laf: Won't this just store the last fault when it can have multiple?

timothyfranson: Whoops, you are right, I didn't realize that it would send out these alarms bundled by device/rule.

Well there's an easy fix for that with the current code, I'll let someone else write and pull it if they like.  It doesn't parse very nicely on the mobile app though, and thus gets very wall-of-texty with any more than a couple faults.

I'm just going to kill this PR for the time being and do a cleaner rewrite, possibly with PD events api v2.  May make a new PR later when that's ironed out.

timothyfranson:
Going to kill this PR, will rewrite and resubmit later to support PagerDuty's V2 API since V1 is al

PR #7917, opened 2017-12-15
Stop sending emails to "NOC" for default_email target

deepseth:
It's confusing when emails are going to an address other than noc@, but showing as being sent to NOC, in mail clients. Better to just display the email address rather than making a name up.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
It's probably easier just to set the name part to `''`. You just get the email then.

laf:
@deepseth You'll need to rebase as well to fix a travis issue.

PR #7918, opened 2017-12-15
Add Huawei VRP sensors to yaml and fix bug in functions.inc.php

erotel:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


erotel:
travis check have problem?

laf:
We've not had that issue before. Might be a transient issue, as you've got to make a change to this PR then we'll see if it goes through after.

So, index values. For migrating php code to yaml you need to match them otherwise you lose data. So most of these just need to be '{{ $index }}'.  The exception to this is .1.3.6.1.4.1.2011.5.25.31.1.1.3.1.9. which needs to be 'tx-{{ $index }}'

includes/definitions/discovery/vrp.yaml
```diff
@@ -12,23 +12,31 @@ modules:
...
         dbm:
             data:
                 -
                     oid: hwOpticalModuleInfoTable
                     value: hwEntityOpticalRxPower
                     num_oid: .1.3.6.1.4.1.2011.5.25.31.1.1.3.1.8.
-                    descr: '{{ $entPhysicalName }}'
-                    index: 'hwEntityOpticalRxPower.{{ $index }}'
+                    descr: '{{ $entPhysicalName }} Rx'
+                    index: 'rx-{{ $index }}'
```
laf: This one is just `{{ $index }}`

erotel: old file have `` 'rx-' . $index ``

laf: You're right, when I looked earlier I thought it didn't

PR #7922, opened 2017-12-19
refactor: Updated discovery/poller to use numeric sysObjectID

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Makes sysObjectID and sysDescr available straight away so devices now only need 1 disco to find all data.

Same principle as poller.


murrant:
- Any concern that the sysObjectID coming from the database will start with "enterprises"?
- I noticed you left the fetch of sysObjectID and sysDescr in the getHost() function... oversight or planned?
- You missed two calls powerconnect.inc.php in mempools and processors "mib-2.1.2.0" (which is sysObjectID)

includes/discovery/core.inc.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+$snmpdata = snmp_get_multi_oid($device, 'sysName.0 sysObjectID.0 sysDescr.0', '-OUQn', 'SNMPv2-MIB:HOST-RESOURCES-MIB:SNMP-FRAMEWORK-MIB');
```
murrant: Isn't `SNMPv2-MIB` the only mib needed here?

erotel:
Discovery new device `[sysObjectID] => .1.3.6.1.4.1.259.10.1.22.101`

Discovery device after first poll `[sysObjectID] => enterprises.259.10.1.42.101`

laf:
> Any concern that the sysObjectID coming from the database will start with "enterprises" and the one you set is fully numeric?

Not really, we overwrite what's in the db within the `$device` array in the core.inc.php file. I can update poller as well to use full numerical and store that.

> I noticed you left the fetch of sysObjectID and sysDescr in the getHost() function... oversight or planned?

Planned. I realise we re-run the queries but I didn't want to use global variables and rely on them throughout the rest of the code.

> You missed two calls powerconnect.inc.php in mempools and processors "mib-2.1.2.0" (which is sysObjectID)

Well spotted, updated.

murrant:
If you update it in the poller, make sure you update alert rules and other things that rely on the enterprises existing.  I think full numeric is probably my preferred over the other.

getHostOS() already accepts $device...

laf:
> If you update it in the poller, make sure you update alert rules and other things that rely on the enterprises existing. I think full numeric is probably my preferred over the other.

Updated.

Done alert rules as well.

> getHostOS() already accepts $device...

It does but it's only called from disco when the OS is generic and then os.inc.php which can be selected not to run.

laf:
This should also fix a few broken things in polling.

murrant:
I think you mis-understand me about getHostOS()
Delete these lines:
```
    $sysDescr    = snmp_get($device, "SNMPv2-MIB::sysDescr.0", "-Ovq");
    $sysObjectId = snmp_get($device, "SNMPv2-MIB::sysObjectID.0", "-Ovqn");
```

change getHostOS() and ubnt.inc.php to use `$device['sysObjectID']` and `$device['sysDescr']`.

You'll also need to update testing to pre-populate those I think.

murrant:
On, second thought, perhaps those changes should be a separate PR.  Don't want to make too many changes at once.

laf:
> I think you mis-understand me about getHostOS()
> Delete these lines:
> 
>     $sysDescr    = snmp_get($device, "SNMPv2-MIB::sysDescr.0", "-Ovq");
>     $sysObjectId = snmp_get($device, "SNMPv2-MIB::sysObjectID.0", "-Ovqn");

Those are used within getHostOS() though. And we haven't yet asked for them in core.inc.php.

> change getHostOS() and ubnt.inc.php to use $device['sysObjectID'] and $device['sysDescr'].

Updated ubnt.inc.php. Not done getHostOS as per the comment above.

> You'll also need to update testing to pre-populate those I think.

Testing uses `$sysDescr` and `$sysObjectID` so don't think anything needs updating.

laf:
> Testing uses $sysDescr and $sysObjectID so don't think anything needs updating.

Actually I see now that travis broke. I've updated to take that into account now. testing should work ok.

laf:
I've refactored everything from sysObjectId -> sysObjectID as that's what the actual OID + DB column is.

This means that anyone who doesn't clear the OS cache when doing a git pull will have issues. `./daily.sh` takes care of this so as always, people should be using that to update.

murrant:
Looks good, I'll do some testing after the December release.

murrant:
@laf can you rebase this so I can test it?

laf:
Rebased + fixed the remaining few sysObjectId -> sysObjectID that slipped in since this was done.

murrant:
Looks good so far, you missed the yaml generation in new-os.php ;)

laf:
After the resent merges, this is now up to date again.

murrant:
I've got this on my main install, working fine (after I remembered to delete the os cache).

laf:
> I've got this on my main install, working fine (after I remembered to delete the os cache).

Yeah which is the only issue with this, if someone updates manually :(

Maybe we should add the cache clearing to includes/sql-schema/update.php to catch those cases otherwise we have no way to resolve it for people updating manually. Anyone just doing a git pull is asking for trouble until discovery runs to populate the schema anyway.

murrant:
We could do a post-merge git hook that removes that file.

laf:
> We could do a post-merge git hook that removes that file.

I didn't think you could commit hooks to repos for them to be automatically run.

murrant:
I think you are right.

PR #7923, opened 2017-12-19
webui: Single Server Details widget

aldemira:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
You're going to open a can of worms with this :) People will be asking for these all over!

So couple of things, can you download the min version of this library and use that instead. As it's only usable for dashboards then you should include the min.js file within the widget to stop it being needed elsewhere.

We already have raphael included which is the same version. `html/js/raphael-min.js` so you can drop yours.

Looks ace btw :)

aldemira:
@laf Yeah ok but, I couldn't find the min version. There are old versions of it on different CDNs but I wanted to use the official one.

laf:
> @laf Yeah ok but, I couldn't find the min version. There are old versions of it on different CDNs but I wanted to use the official one.

Yeah actually can't see it myself. We might need to min it ourselves (as we should with some others). It does raise the question, is this library supported and do we need to pick another one if it's not?

aldemira:
I'm open to suggestions. This project hasn't been touched for a year. There's gauge.js which a not very well documented and not as nice as this, but I can give it a try if you want to.

laf:
Don't want to cause additional work so let's leave it for now. If we start to use it in anger more then I think we should take a look at the best one.

aldemira:
Now I know why there isn't a minified version of the library. I've used a few different sites to compress it, and minified file always breaks the output. I'm attaching the screenshots.
screen shot 2017-12-21 at 09 50 01
screen shot 2017-12-21 at 09 54 07

laf:
I've rebased this and also fixed the sql queries to stop the device list being shown to people who don't have access. The widget will also not load now if the user doesn't have access and display an error message.

Will merge once checks complete.

kkrumm1:
Found a small issue :( When you try to use another Single Server Widget in the same dashboard it adds the info to the first Single server widget, not the second server widget.

image

aldemira:
Thanks @kkrumm1 ! Just sent another PR to fix this.

PR #7924, opened 2017-12-20
refactor remove all old style snmp calls

murrant:
When calling snmpget (etc) this way it doesn't work with the save-tests-data.php script.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/cisco-mac-accounting.inc.php
+++ b/includes/discovery/cisco-mac-accounting.inc.php
@@ -1,9 +1,8 @@
 <?php
 
 if ($device['os_group'] == 'cisco') {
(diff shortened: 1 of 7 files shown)


laf:
Tested a full discovery and polling has been running ok since I've checked this out. most of my dev won't use this code though so limited test. Code looks sane though.

eimann:
IMHO this triggers a lot of BGP down false-prositives on Ubiquity EdgeOS and Cisco IOS.
After daily.sh ran in den middle of the night a dozen alarms popped up.

murrant:
well, not so much false positives as breaks bgp-peers polling.

PR #7925, opened 2017-12-20
Added a custom sensor oid option to LibreNMS

erotel:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/device/edit.inc.php
+++ b/html/pages/device/edit.inc.php
@@ -51,6 +51,8 @@ if ($_SESSION['userlevel'] < '7') {
 
     $panes['component'] = 'Components';
 
+    $panes['custom'] = 'Custom OID';
+
     print_optionbar_start();
 
     unset($sep);
--- a/html/pages/device/edit/custom.inc.php
+++ b/html/pages/device/edit/custom.inc.php
@@ -137,7 +137,7 @@ foreach ($custom_graph as $key => $val) {
         <td>
         <input type="hidden" name="remove_custom_index" value="'.$val['sensor_index'].'" />
         <input type="hidden" name="remove_custom_class" value="'.$val['sensor_class'].'" />
-        <button type="submit" class="btn btn-danger remove-custom" id="remove-custom" name="remove_custom_id" value="'.$val['sensor_id'].'">Dlelete</button>
+        <button type="submit" class="btn btn-danger remove-custom" id="remove-custom" name="remove_custom_id" value="'.$val['sensor_id'].'">Delete</button>
         </td>
         </tr>
         </form>
--- a/html/pages/device/edit/custom.inc.php
+++ b/html/pages/device/edit/custom.inc.php
@@ -1,0 +1,148 @@
+<?php
+include '../includes/discovery/functions.inc.php';
+
+if ($_POST['editing']) {
+    if ($_SESSION['userlevel'] > '7') {
+        $custom_oid = mres($_POST['custom_oid']);
+        $sensor_class = mres($_POST['sensor_class']);
+        $custom_descr = mres($_POST['custom_descr']);
+        $custom_diviz = mres($_POST['custom_diviz']);
+        $custom_multip = mres($_POST['custom_multip']);
+        if (!empty($custom_oid) && !empty($custom_descr)) {
+            if (substr($custom_oid, 0, 1) != '.') {
+                $custom_oid = "." . $custom_oid;
+            }
+            $index_tmp  = dbFetchCell("SELECT `sensor_index` FROM `sensors` WHERE `device_id` = ? AND `sensor_type` LIKE 'custom_oid' ORDER BY `sensor_type` ASC", array($device['device_id']));
+            $index_tmp ++;
+            discover_sensor($valid['sensor'], $sensor_class, $device, $custom_oid, $index_tmp, 'custom_oid', $custom_descr, $custom_diviz, $custom_multip, null, null, null, null, null);
+
+            $update_message = 'Custom graph added.';
+            $updated        = 1;
+        } else {
+            $update_message = 'Oid and Description must be set';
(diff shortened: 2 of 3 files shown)


murrant:
Interesting idea, but I'm unsure about this.  It only supports sensors, which are pretty easy to add and I think should always be added upstream so users don't have to add manually.

erotel:
On the contrary, I think it's a good thing. If I want to add eg the temperature sensor that I use only, so why do it to PR. If I admit that it can be beneficial for others, then I will do a PR.
I need to use arduino to monitor temperature in a rack, but arduino has little memory, so I have to use nonstandard snmp.
In librenms, it appears to me as a generic device (I do not mind), I add customized thermometers and alerting.

kkrumm1:
I think this would be helpful for custom SNMP OID's.

laf:
I think this is partly the way forward. I've always thought we should have a 'custom' section where users can query single OIDs and this along with the codebase from Wireless looks like a good way forward.

I'd like to leave this open into the new year to discuss further if that's ok with everyone.

tlkhorses:
I would like to see this. I just came from another product where I could add them in packs for a product. For instance I would add EVM, PER, PHY etc for Mimosa products or FCS errors for Mikrotiks.

BradHooper:
I think this is a great idea as well. I have a lot of single values I'd like to monitor however don't believe it would be beneficial to all LibreNMS users to have them merged into LibreNMS as a feature.

It would be ideal if devices self discovered these custom OIDs once entered globally.

murrant:
BTW, this doesn't allow random custom OIDs.  They need to be a sensor type that we already support.
https://docs.librenms.org/Developing/os/Health-Information/

It is already possible to add via yaml and even config.php. (Don't make local edits to yaml unless you intend to contribute upstream)

erotel:
My test arduino snmp in rack.

```root@speedtest:~# snmpwalk -c public -v1 10.133.67.100
iso.3.6.1.2.1.1.1.0 = STRING: "Agentuino, a light-weight SNMP Agent."
iso.3.6.1.2.1.1.2.0 = STRING: "1.3.6.1.3.2009.0"
iso.3.6.1.2.1.1.3.0 = Timeticks: (530000) 1:28:20.00
iso.3.6.1.2.1.1.4.0 = STRING: "slezi2"
iso.3.6.1.2.1.1.5.0 = STRING: "serverovna"
iso.3.6.1.2.1.1.6.0 = STRING: "Italska"
iso.3.6.1.2.1.1.7.0 = INTEGER: 6
End of MIB


root@speedtest:~# snmpwalk -c public -v1 10.133.67.100 1.3.6.1.4.1.36582.0.1.1
iso.3.6.1.4.1.36582.0.1.1 = INTEGER: 180
Error: OID not increasing: iso.3.6.1.4.1.36582.0.1.1
 >= iso.3.6.1.4.1.36582.0.1.1

root@speedtest:~# snmpwalk -c public -v1 10.133.67.100 1.3.6.1.4.1.36582.0.1.2
iso.3.6.1.4.1.36582.0.1.2 = INTEGER: 170
Error: OID not increasing: iso.3.6.1.4.1.36582.0.1.2
 >= iso.3.6.1.4.1.36582.0.1.2

```

arduino


arduino1

murrant:
A good use for this @erotel.  I think the use of this needs to be very clear in the webui.  If that is satisfied I'm fine with this.

I can think of a couple of devices that have modular sensors (using fixed oids of course) where this would be useful.

murrant:
@erotel Just getting back to this.

I think move this to a button under the health settings tab.  (You'll have to make the tab always show up).
Have a modal pop up showing the settings you have on the "Custom OID" tab.

laf:
I'd like to approach this slightly different but I've had no time to put my thoughts down. I'll try and get to it soon.

laf:
Sorry for the delay on this. So here's my thoughts:

I don't think adding it to the sensors table is the overall right approach purely because someone will ask next week that it's done for wireless, processors, etc, etc.

What I think we should do is have a generic table that is similar to the way wireless or sensors work which allows people to add arbitrary OIDs in, set things like a high / low, what type of data should be returned, the section it should appear under (not to have it appear in say processors but as a processor section within 'Custom graphs' or similar.

This way people can just throw stuff into here. Allow it so people can import/export those as well to share with others as the natural concern for us is that people don't share what they've done and so people duplicate the effort all over the show.

It would take a bit of work as I think it would mean standardising the graphing library all over to make it re-usable.

What do people think to that?

k-y:
+1 to the generic table idea

kalamchi75:
Hi Guys, 

Thank you for the effort made to add this feature. I have tested it with a couple of custom OID and it seems to poll and graph properly. 
However, I wonder if this script can be made more generalized and customized. For instance, the drop down menu to select the sensor class, can be more general to include things like BPS (bits per second), or a field can be added to input a sensor class.  
Also, I noticed that the dbm sensor can only be added once. i.e. if I have more than one sensor in the device that reads dBm values, I cannot add them all. I can only add one, and if i try to add more, it will only update one that was added the first time. Any chance this can be modified to accept multiple dBm sensors ?
What i'm trying to achieve is to add some Mikrotik wireless sensors that are not being discovered/added automatically.

murrant:
@kalamchi75 That kind of is not what LibreNMS is.  It is actually very easy to add those sensors (you don't even need to write code) and then share with everyone.  LibreNMS shy's away from forcing everyone to manually configure everything.  Why have 1000 people configure something when 1 person can configure it and 1000 get it without doing anything.

The reason the PR exists is some devices have attachable sensors and don't indicate what kind of sensor is attached via snmp.  I suspect the reason it hasn't been merged is there is a lot of confusion that is for general use.

murrant:
Closing in favor of #8738

PR #7936, opened 2017-12-21
Fail2ban polling update and dropping of the FW checking

VVelox:
Since fail2ban on FreeBSD has moved to using anchors, the previous
issue of it dreking all over its self is no longer a problem.

And AFAIK it has never been an issue on Linux.

This also makes this script more cross platform as well instead.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Dropping the firewall stuff as since Fail2ban as updated to using anchors(while esoteric as drek, are uber awesome it turns out) in PF instead of tables it means it is a lot more reliable on those systems and I can't seem to find any references to issues with IPFW.

https://github.com/librenms/librenms-agent/pull/155

That update is also needed for this. The polling is not really backwards compatible. The poller is also massively cleaned up and better documented.

Also RRDs are backwards compatible. Any one with the old totals RRD will now just have a RRD with a unused field in it.

### DIFF
--- a/includes/polling/applications/fail2ban.inc.php
+++ b/includes/polling/applications/fail2ban.inc.php
@@ -17,12 +17,15 @@ $bannedStuff = explode("\n", $f2b);
 
 $total_banned=$bannedStuff[0];
 
-$rrd_name = array('app', $name, $app_id, 'totalbanned');
+$rrd_name = array('app', $name, $app_id);
 $rrd_def = RrdDefinition::make()
-    ->addDataset('banned', 'GAUGE', 0);
+    ->addDataset('banned', 'GAUGE', 0)
+    ->addDataset('firewalled', 'GAUGE', 0);
+
 
 $fields = array(
     'banned' =>$total_banned,
+    'firewalled'=>'U',
 );
 $metrics['total'] = $fields;
 
--- a/includes/polling/applications/fail2ban.inc.php
+++ b/includes/polling/applications/fail2ban.inc.php
@@ -16,23 +16,20 @@ $metrics = array();
 $bannedStuff = explode("\n", $f2b);
 
 $total_banned=$bannedStuff[0];
-$firewalled=$bannedStuff[1];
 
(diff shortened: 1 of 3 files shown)


VVelox:
hmm... it turns out this is not totally backwards compatible... it did not like it only updating one value in the RRD.

Removing that old RRD fixes it. The jail specific ones are fine though.

VVelox:
Was just chatting with Murrant and am going to rename the graph stuff.

includes/polling/applications/fail2ban.inc.php
```diff
@@ -16,23 +16,20 @@
...
 
 $total_banned=$bannedStuff[0];
-$firewalled=$bannedStuff[1];
 
-$rrd_name = array('app', $name, $app_id);
+$rrd_name = array('app', $name, $app_id, 'totalbanned');
 $rrd_def = RrdDefinition::make()
-    ->addDataset('banned', 'GAUGE', 0)
-    ->addDataset('firewalled', 'GAUGE', 0);
+    ->addDataset('banned', 'GAUGE', 0);
```
laf: Does this break existing graphs as it would expect 2 entries for people who ran the code before this change?

VVelox: Talked to Murrant about this and ended up changing the name of the RRD as otherwise people would need to remove the old ones.

laf: Why not just leave the old file in place and update the firewalled value with 0 and remove from the webui so you never see it. Changing the filename results in people losing the historic data. @murrant - what do you think?

VVelox: Was worried about bloat with RRDs with pointless data in them. But then again it is very minor.

Decided to rework this a bit.

Just shoving U into the old unused value. I've tested it and it works. :)

VVelox:
Running it with the U now and it it has looked good for like the last hour or so of updates.

PR #7946, opened 2017-12-23
Moved routeros sensors to yaml + fixed divisors

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
@murrant I'm not sure what the issue is here unless genuinely the move to yaml is breaking something.

Trying to run `./scripts/save-test-data.php -h 399 -o routeros -m sensors` produces:

https://p.libren.ms/view/raw/592c13b7 which seems to contain some bad data. Any ideas without me trawling through the new code?

includes/definitions/discovery/routeros.yaml
```diff
@@ -151,17 +178,24 @@ modules:
...
                 -
                     oid: mtxrHlFanSpeed2
                     num_oid: .1.3.6.1.4.1.14988.1.1.3.18.
                     descr: 'Fan #2'
-                    index: 'mtxrHlFanSpeed2.{{ $index }}'
+                    index: '1'
+        signal:
+            data:
+                -
+                    oid: mtxrWlRtabStrength
```
murrant: This doesn't seem right.  It seems to be wireless client signal strength...  It is indexed by MAC, which caused several issues.

It is also similar to the MSA issue with the string based index.

I would suggest dropping this sensor.

murrant: And don't forget to update the test data.

laf: Updated back to php

murrant:
@laf this looks good to me.  Would you be willing to test the ./save-test-data.php script here?

Capture the data in master, then add here and hopefully everything passes.

laf:
This is all the test data I have :(

PR #7966, opened 2017-12-27
Use ignore_storage function for NetApp storage discovery

Sneddo:
The previous discovery was not honoring the ignore_mount_* configs for me, and whilst investigating I noticed the ignore_storage function. I've updated the discovery to use this instead and the ignore options now appear to be working as expected.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
I think ignore_mounts() should be inside discover_storage()...  Are you up to making that change?

Sneddo:
Yeah, good call. Still wrapping my head around the code structure :)

I'll raise another PR once I've done the change. Looks like only one other discovery uses ignore_storage outside of discover_storage anyway, so should be pretty simple.

PR #7970, opened 2017-12-28
docs: added ports purge to the Clean up Options.

kkrumm1:
Added ports purge to clean up options, to help end user find easily and use. 

#### Please note

#### Testers


murrant:
This is fine, but should we add an automated option (disabled by default)?

laf:
We should have this as part of clean up yes. I'll create a separate issue for this (and yes, disabled by default).

PR #7973, opened 2017-12-28
Move ignore_storage call to discover_storage

Sneddo:
Calling ignore_storage as part of discover_storage to ignore based on description. Also updated NetApp and ucd-dsktable to remove calls to ignore_storage in these discovery includes.

Tested the NetApp discovery and works as expected.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -577,37 +577,39 @@ function discover_link($local_port_id, $protocol, $remote_port_id, $remote_hostn
 
 function discover_storage(&$valid, $device, $index, $type, $mib, $descr, $size, $units, $used = null)
 {
-    d_echo("Discover Storage: $index, $type, $mib, $descr, $size, $units, $used\n");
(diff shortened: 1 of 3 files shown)


includes/discovery/functions.inc.php
```diff
@@ -577,37 +577,39 @@ function discover_link($local_port_id, $protocol, $remote_port_id, $remote_hostn
...
-                    'storage_index' => $index,
-                    'storage_mib' => $mib,
-                    'storage_type' => $type,
-                    'storage_units' => $units,
-                    'storage_size' => $size,
-                    'storage_used' => $used,
-                ),
-                'storage'
-            );
+    if (ignore_storage($device['os'], $descr) != 1) {
```
murrant: Can you change this to be like this?
```php
if (ignore_storage($device['os'], $descr)) {
    return;
}
```
It saves an indent and long if statement.

murrant:
Looks good, just a quick style fix.  Then I'll give it a test :)

PR #7974, opened 2017-12-28
Changed default disk utilization warning threshold to 95% from 60%

admiralspark:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

----

Hello, 

The default storage alarm threshold is 60%. This is ridiculously low for any sort of servers, and though it can be adjusted on a per-server basis it creates unneeded extra work to manually change it each time. On top of that, validation checks fail if the value is manually changed by the end user. 

This is to open discussion on the subject, and if 95% is considered a good cutoff point then no further work is needed.


laf:
This just changes the hard coded limit when we should be offering a user configurable value that overwrites the 60 when inserting the data into the table.

Is this something you want to tackle?

murrant:
This would mean updating the function discover_storage() to set storage_perc_warn.

admiralspark:
@laf @murrant Python is my strong suit, so I can take a look but it'll take a bit. 

Do you think it should be set at initial setup? Or have it be a General Setting option? How do you want to handle mariadb validation using the validate script after, just check that the value exists or verify it matches the set value? 

I'm willing to do the work, but not sure what the best end result would be. I think a user configurable option post-installation seems good, and I can update discover_storage() as he said.

laf:
So you can add a webui config option but it does need a new section setting up for this so it's probably better to leave it for now unless you want to try that as well? I can do that part for you.

The main part is to update `discover_storage()` and add `storage_perc_warn` to the insert using `Config::get('discovery.storage.perc_warn')` - Made that up but it seems a logical config name.

@murrant what do you think?

murrant:
Seems good to me, but perhaps we should add default to the name to indicate that changing it's value won't update existing records.

admiralspark:
Hmm. @laf @murrant I can kill this PR and start a new one with a modified discover_storage() function. I like the idea of using discovery.storage.perc_warn_default in the config for it instead of modifying the base 60% warning.

PR #7976, opened 2017-12-29
Fix bugs in bgp-peers poller 

murrant:
peer data in wrong format
array_diff doesn't work right, change to array_diff_assoc
astext is too short for some

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Works for me. I'll keep this in mind when cutting the december release. Might not include the original code that breaks it anyway so will see. Thanks for fixing @murrant. I'll get some arista test data submitted shortly.

PR #7978, opened 2017-12-29
Add Oxidized API URL protocol

Zmegolaz:
file_get_contents() need the protocol ("http://" or similar) when fetching remote files, add it to the oxidized API URL config if the user forgot it.

From: https://www.reddit.com/r/LibreNMS/comments/7mrufq/i_am_a_moron_oxidized_config/

I'm not sure this is the best place to add this check, if you have any suggestion, please let me know.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/css/styles.css
+++ b/html/css/styles.css
@@ -2174,3 +2174,8 @@ label {
   vertical-align: middle;
   padding: 15px;
 }
+
+.validation:invalid {
+    border-color: red;
+}
+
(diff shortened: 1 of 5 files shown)


laf:
Maybe just doing https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation for now?

Ideally longer term we'd define the config values that should exist in the db so we can validate they exist. We could also then validate the input.

Zmegolaz:
Something like this?

laf:
Close, config still gets updated.

This patch fixes that: https://p.libren.ms/view/raw/b0206dc8

laf:
Brilliant, thanks for that :) We should go back through at some point and do the same for some of the other inputs.

PR #7982, opened 2017-12-29
Added Hillstone StoneOs detection

alangregory:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/stoneos.yaml
```diff
@@ -0,0 +1,19 @@
...
+        text: Device Traffic
+    -
+        graph: device_processor
+        text: CPU Usage
+    -
+        graph: device_mempool
+        text: Memory Usage
+discovery:
+    -
+        sysObjectId: .1.3.6.1.4.1.28557.1
```
laf: Do you need sysDescr as well? Will sysObjectID not be unique enough?

includes/definitions/stoneos.yaml
```diff
@@ -0,0 +1,19 @@
+os: stoneos
+text: Hillstone StoneOs
+type: network
+icon: hillstone
+group: hillstone
+over:
```
laf: Can you format the over graphs like the other yaml files please.

includes/definitions/stoneos.yaml
alangregory: I think only Hillstone use this Oid, removed sysdescr match.

includes/definitions/stoneos.yaml
alangregory: Fixed, was my yaml format plugin that changed it.

murrant:
@alangregory You need to reduce your svg files.

You can use this site: https://jakearchibald.github.io/svgomg/

laf:
@alangregory Thanks for submitting this :)

Would you mind submitting some additional test data as per https://docs.librenms.org/#Developing/os/Test-Units/#example-workflow

Ensure you anonymise the data before you submit it though.

PR #7983, opened 2017-12-29
webui: Added Oxidized config validator

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I'm happy for this not to be part of LibreNMS and I'll just chuck it on my site for people to use

### DIFF
--- a/doc/Extensions/Oxidized.md
+++ b/doc/Extensions/Oxidized.md
@@ -129,3 +129,10 @@ Using the Oxidized REST API and [Syslog Hooks](/Extensions/Syslog/#external-hook
 ```bash
 next_adds_job: true
 ```
+
+### Validate Oxidized config
+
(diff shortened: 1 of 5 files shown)


murrant:
Interesting, I'll take a peak sometime.

murrant:
Works :)

Needs positive feedback when the config is ok.
The textarea is too tall, I have to scroll a lot to get do the submit button.

murrant:
Any thoughts of putting a button under the oxidized page instead of the menu?
I'm not convinced one is better than the other, just a thought.

laf:
Just pushing a commit that will reduce the box height and also adds feedback when the config is ok.

I'm not fussed either way on where the link actually is, possibly would be more hidden if it was within the other page though.

laf:
That's a good point. I'll move it tomorrow.

laf:
Slight rejig of location for the link.

image

Also added small bit of blurb in docs for it.

PR #7985, opened 2017-12-30
Don't run tests on PHP 5.4, 5.5, 7.0.  Add PHP 7.2

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,17 +5,13 @@ services:
 matrix:
   fast_finish: true
   include:
-  - php: 7.1
+  - php: 7.2
     env: EXECUTE_BUILD_SCHEMA=true
-  - php: 7.0
+  - php: 7.1
     env: SKIP_STYLE_CHECK=1
   - php: 5.3
     sudo: required
     dist: precise
-    env: SKIP_STYLE_CHECK=1
-  - php: 5.4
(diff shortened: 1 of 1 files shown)


laf:
Is this for when we drop < 5.6 support? If so then I think we should hold out until that date.

murrant:
No, it just seems like a waste to test these middle versions.  Definitely open for discussion.

laf:
We're only 8 days away from dropping the support for these anyway so we may as well just wait until then?

murrant:
What about PHP 7.0 and maybe 7.1 after 7.2 is more established? Do we still want to run tests for them?

I'll admit the main reason for this PR is I don't like that we can't run all of the jobs at once in Travis.

laf:
https://docs.travis-ci.com/user/speeding-up-the-build/#PHP-optimizations to help speed things up a little maybe.

The thing with excluding versions is if functions get dropped or are added between them, it means we can break X version because no one can know which functions are always available so for the time difference, I'd prefer to keep them.

It would be good to skip all remaining builds if the first php version we check fails. They also have some other info on parallel builds using matrix, ours look to run sequentially so maybe we could speed things up there.

PR #7992, opened 2017-12-31
Test docs before deploying

murrant:
fix issue with docs

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


scripts/deploy-docs.sh
```diff
@@ -24,10 +29,16 @@ cd ../
...
-cd out
-
-touch .
-git add -A .
-git commit -m "GH-Pages update by travis after $TRAVIS_COMMIT"
-git push -q origin master 
+build_result=$?
+
+# Only deploy after merging to master
+if [ "$TRAVIS_PULL_REQUEST" == "false" -a "$TRAVIS_BRANCH" == "master" ]; then
```
laf: Never thought about doing it this way, nice work :)

Would it be a good idea to check the build_result first and if it's not 0 then exit before checking travis variables?

murrant: probably, no point in pushing broken docs.

PR #7993, opened 2018-01-01
webui: fix services display

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

fixed display bug when only one host had services

before:
screen shot 2018-01-01 at 4 35 35 am

after:
screen shot 2018-01-01 at 4 35 01 am

### DIFF
--- a/html/pages/services.inc.php
+++ b/html/pages/services.inc.php
@@ -145,10 +145,10 @@ require_once 'includes/modal/delete_service.inc.php';
                     $header = true;
                     $footer = false;
 
-                    $svc = 0;
-                    $svcs = dbFetchRows("SELECT * FROM `services` WHERE `device_id` = ? $where ORDER BY service_type", $sql_param);
-                    $svct = count($svcs);
-                    foreach ($svcs as $service) {
+                    $service_iteration = 0;
+                    $services = dbFetchRows("SELECT * FROM `services` WHERE `device_id` = ? $where ORDER BY service_type", $sql_param);
+                    $services_count = count($services);
+                    foreach ($services as $service) {
                         if ($service['service_status'] == '2') {
                             $label = 'label-danger';
                             $title = 'CRITICAL';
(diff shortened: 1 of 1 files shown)


html/pages/services.inc.php
```diff
@@ -35,12 +35,7 @@
                     $vars['view'] = 'basic';
                 }
 
-                $status_options = array(
```
laf: I think this is an unnecessary change. It's a bit more readable as it was .

html/pages/services.inc.php
```diff
@@ -141,9 +138,14 @@
                     } else {
                         $sql_param[0] = $device_id;
                     }
-                    $head = true;
 
-                    foreach (dbFetchRows("SELECT * FROM `services` WHERE `device_id` = ? $where ORDER BY service_type", $sql_param) as $service) {
+                    $header = true;
+                    $footer = false;
+
+                    $svc = 0;
```
laf: I'd like to see better variable names that describe what data they contain. Same for the two below.

laf:
I must be missing something here, what does this actually change as visually I can't see a difference between the old and the new with this on my test install? I can see the difference on your screenshots though. 

The current demo.librenms.org also doesn't show how it looks in your old screenshot.

crcro:
it only showed like that when services where critical and for one host only ...

laf:
I still don't see the difference with 1 host or 4.

This also breaks the dynamic removal of the service, it was fixed again recently so you'll need to rebase and make sure that these fixes aren't removed: https://github.com/librenms/librenms/pull/8030

crcro:
already rebased and included #8030 fix

laf:
@crcro Can you show a broken screenshot as I'm not convinced this is changing anything visually.

crcro:
@laf this is from a up2date librenms install

screen shot 2018-01-25 at 11 02 18 pm

murrant:
@crcro How it looks in master for me:

image

crcro:
@murrant how does it look if you click on "critical" only? on all, it shows good for me also

PR #7994, opened 2018-01-01
webui: apps lineup

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

enabled apps overview (atm no otherway than entering the url manually)
minor display changes to match latest changes.

screen shot 2018-01-01 at 5 20 13 am

screen shot 2018-01-01 at 5 14 19 am

screen shot 2018-01-01 at 5 14 37 am

### DIFF
--- a/html/pages/apps.inc.php
+++ b/html/pages/apps.inc.php
@@ -1,4 +1,5 @@
 <?php
+$pagetitle[] = 'Apps';
 
 $graphs['apache']    = array(
     'bits',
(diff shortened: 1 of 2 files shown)


html/pages/apps.inc.php
```diff
@@ -1,4 +1,26 @@
 <?php
+/*
```
laf: I've been thinking about this copyright you've added to files you've changed. I'm not 100% sure on the position but I would say that we shouldn't be adding a copyright that seems to claim it's done by LibreNMS contributors. Old historic files weren't written by us so I think you should remove this / these.

laf:
Overall I'm not 100% sure on the panel in panel approach but as we already have this for other pages then I'll leave it for now.

kkrumm1:
@crcro I know this is off topic but anyway you could maybe look into building Overview page for Device groups?

PR #8000, opened 2018-01-02
Added CheckPoint SecurePlatform support

ricardcc:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/splat.inc.php
```diff
@@ -0,0 +1,30 @@
+<?php
+/**
+ * SPLAT.inc.php
+ *
+ * NTTCOM MS poller module for Check Point SECUREPLATFORM
+**/
+
+use LibreNMS\RRD\RrdDefinition;
+
+$tmp_SPLAT = snmp_get_multi_oid($device, 'svnVersion.0 svnApplianceProductName.0 svnApplianceSerialNumber.0', '-OUQs', 'CHECKPOINT-MIB');
```
laf: Can you lowercase this variable please.

includes/definitions/splat.yaml
```diff
@@ -0,0 +1,30 @@
...
+discovery_modules:
+    ports-stack: 0
+    fdb-table: 0
+    stp: 0
+    wireless: 0
+    vlans: 0
+    bgp-peers: 0
+    ntp: 0
+    services: 0
+    ip6-addresses: 0
```
laf: This should be `ipv6-addresses` but does the device not support ipv6?

html/includes/graphs/device/splat_actsessions.inc.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+/*
+ * LibreNMS
+ *
+ * Copyright (c) 2014 Neil Lathwood <https://github.com/laf/ fua http://www.lathwood.co.uk>
```
laf: You need to put your own copyright here.

includes/definitions/splat.yaml
```diff
@@ -0,0 +1,30 @@
+os: splat
```
laf: This should be secureplatform rather than splat (unless it's really called splat?). I've seen references to Gaia for the newer gen - what about that?

ricardcc: Hi,
in network & security environments, checkpoint secureplatform is called splat (it's an acronism). I used this name becase it's how is known this device. But if you think is better change to secureplatform there is no problem.

I saw about CheckPoint Gaia, but for the moment I've not improved it. I saw that it has basic tests, but my company doesn't monitorize any gaia yet. I think in a mid future i will be able to improve it.
There are differences regarding to monitor these devices. Gaia contains OIDs which are not available in Gaia, for this reason I though to use a new os.

includes/definitions/splat.yaml
ricardcc: Ipv6 is supported for this platform, but it's not used. Anyway, I'll change it in order to support it

laf:
I'm still not 100% on the os name but I'm not going to block it because of that. Anyone else @librenms/reviewers care to take a look?

murrant:
I found this: 

image

Seems to indicate the OS should be secureplatform.

ricardcc:
I'm going to change the os name.

laf:
> I found this:
> 
> image
> 
> Seems to indicate the OS should be secureplatform.

They did say splat is the well known name for it but secureplatform is more standard for us.

Merging in now as we have some test data and all checks pass.

PR #8003, opened 2018-01-03
fix: Add back os_group junos

zombah:
This fixes some existing conditions, like cisco-vrf discovery module 
 
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/junos.yaml
+++ b/includes/definitions/junos.yaml
@@ -1,4 +1,5 @@
 os: junos
+group: junos
 text: 'Juniper JunOS'
 type: network
 over:


murrant:
I just noticed there is only one check for this setting in cisco-vrf.

Perhaps that check should just be if `$config['os'] = 'junos'`?

Either that or we should probably add the junos group to the other junos os.

zombah:
JunosE is only other mpls capable target for vrf support. I don't have such devices to test but will wait some time maybe someone will vote for it, if no, i'll drop group condition.

zombah:
Drop for now

PR #8009, opened 2018-01-03
fix: Cisco-vrf discovery for device os junos

zombah:
Fix vrf discovery for junos devices because there is no
os_group junos right now and no need to create it as
other junos variant probably doesn't support mpls l3vpn.
This PR superseeds #8003 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Thanks for researching :)

PR #8010, opened 2018-01-03
alert rules: added Dell iDRAC alert rules

kkrumm1:
to the alerts collection.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/misc/alert_rules.json
+++ b/misc/alert_rules.json
@@ -326,27 +326,27 @@
   },
   { 
     "rule": "%sensors.sensor_current ~ \"[3|4]\" && %sensors.sensor_oid = \".1.3.6.1.4.1.674.10892.5.5.1.20.140.1.1.4\"",
-    "name": "Dell idrac Virtual Disk Failed/Degraded"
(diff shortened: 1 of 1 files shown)


Rosiak:
Awesome stuff!
My OCD hits me, could you please change idrac to iDRAC? :D

kkrumm1:
@Rosiak Sure will do. :)

murrant:
@kkrumm1 are these types of rules needed anymore now that we can alert on state sensor generic values?

kkrumm1:
@murrant Yes, some users and I still like to use the specific rules instead of the state sensor generic values

murrant:
@kkrumm1 what could be improved so you can use the generic one?  a specific template built in?

kkrumm1:
The generic alert works it's just when you receive the alert you have scroll down just to see what sensor is broken and read the details in the alert. I like to use the sensor specific alerts so I can see right away what the issue is.

laf:
I don't have an issues with more specific alert rules as they cover all bases and are unlikely to change.

PR #8016, opened 2018-01-03
refactor: Move more state sensors to yaml

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Will try to go through more the next coming days.

Disclaimer:
I have not tested this, this is just converted manually.


laf:
Wahoo the @Rosiak is back :)

I've not compared the yaml to the php but do any of these change the state name or index so historic data is lost?

murrant:
@Rosiak If you have access to the devices, you could generate test data from before this PR then test it after.  

Rosiak:
I'll see what I can figure out, sorry for the delay, but time..

laf:
I've added test data for 3 of 5 of these devices. @Rosiak would you have time to add the other two if you have data?

If not, I think we just merge in for now.

murrant:
Remove the arp-data section from the json files. That module doesn't work right for testing yet.

murrant:
arp-tables fixed in #8119

PR #8017, opened 2018-01-03
api: Added API calls for routing and resources management

zombah:
This PR includes API calls:
VRF listing and optional filter by hostname or vrfname
Get VRF by vrf id
Cbgp listing and optional filter by hostname
All VLANs listing
All IP Addresses listing
All IP Networks listing
Get IP Addresses by network id

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -239,6 +239,24 @@ function get_graph_generic_by_hostname()
     rrdtool_close();
 }
 
+
+function list_locations()
+{
+    check_is_read();
+
+    $app           = \Slim\Slim::getInstance();
+    $router        = $app->router()->getCurrentRoute()->getParams();
+
+    $locations   = dbFetchRows("SELECT `locations`.* FROM `locations` WHERE `locations`.`location` IS NOT NULL");
+    $total_locations = count($locations);
+    if ($total_locations == 0) {
+        api_error(404, 'Locations does not exist');
+    }
+
+    api_success($locations, 'locations');
+}
+
+
 function get_device()
 {
     // return details of a single device
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -549,7 +549,7 @@ function list_cbgp()
         $device['ip'] = inet6_ntop($device['ip']);
         if (is_numeric($host_id)) {
             $device['parent_id'] = $host_id;
-         }
+        }
         $bgp_counters[] = $bgp_counter;
     }
     $total_bgp_counters = count($bgp_counters);
@@ -1417,7 +1417,7 @@ function list_vrf()
         $device['ip'] = inet6_ntop($device['ip']);
         if (is_numeric($host_id)) {
             $device['parent_id'] = $host_id;
-         }
+        }
         $vrfs[] = $vrf;
     }
     $total_vrfs = count($vrfs);
@@ -1493,7 +1493,7 @@ function list_vlans()
         $device['ip'] = inet6_ntop($device['ip']);
         if (is_numeric($host_id)) {
             $device['parent_id'] = $host_id;
-         }
+        }
         $vlans[] = $vlan;
     }
     $vlans_count = count($vlans);
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -1330,8 +1330,6 @@ function get_devices_by_group()
 
 function list_vrf()
 {
-    check_is_read();
-
     $app        = \Slim\Slim::getInstance();
     $sql        = '';
     $sql_params = array();
@@ -1346,6 +1344,7 @@ function list_vrf()
         $sql        = ' AND `vrfs`.`vrf_name`=?';
         $sql_params = array($vrfname);
     }
+    check_device_permission($device_id);
 
     $vrfs       = dbFetchRows("SELECT `vrfs`.* FROM `vrfs` WHERE `vrfs`.`vrf_name` IS NOT NULL $sql", $sql_params);
     $total_vrfs = count($vrfs);
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -553,8 +553,8 @@ function list_cbgp()
         $bgp_counters[] = $bgp_counter;
     }
     $total_bgp_counters = count($bgp_counters);
-    if (!is_numeric($total_bgp_counters)) {
-        api_error(500, 'Error retrieving bgp counters');
+    if ($total_bgp_counters == 0) {
+        api_error(404, 'BGP counters does not exist');
     }
 
     api_success($bgp_counters, 'bgp_counters');
@@ -858,9 +858,6 @@ function get_ip_addresses()
         $ipv4   = dbFetchRows("SELECT `ipv4_addresses`.* FROM `ipv4_addresses` JOIN `ports` ON `ports`.`port_id`=`ipv4_addresses`.`port_id` WHERE `ports`.`device_id` = ? AND `deleted` = 0", array($device_id));
         $ipv6   = dbFetchRows("SELECT `ipv6_addresses`.* FROM `ipv6_addresses` JOIN `ports` ON `ports`.`port_id`=`ipv6_addresses`.`port_id` WHERE `ports`.`device_id` = ? AND `deleted` = 0", array($device_id));
         $ip_addresses_count = count(array_merge($ipv4, $ipv6));
-        if (!is_numeric($ip_addresses_count)) {
-            api_error(500, 'Error retrieving IP Addresses');
-        }
         if ($ip_addresses_count == 0) {
             api_error(404, "Device $device_id does not have IP Addresses");
         }
@@ -870,9 +867,6 @@ function get_ip_addresses()
         $ipv4   = dbFetchRows("SELECT * FROM `ipv4_addresses` WHERE `port_id` = ?", array($port_id));
         $ipv6   = dbFetchRows("SELECT * FROM `ipv6_addresses` WHERE `port_id` = ?", array($port_id));
         $ip_addresses_count = count(array_merge($ipv4, $ipv6));
-        if (!is_numeric($ip_addresses_count)) {
-            api_error(500, 'Error retrieving IP Addresses');
-        }
         if ($ip_addresses_count == 0) {
             api_error(404, "Port $port_id does not have IP Addresses");
         }
@@ -882,9 +876,6 @@ function get_ip_addresses()
         $ipv4   = dbFetchRows("SELECT * FROM `ipv4_addresses` WHERE `ipv4_network_id` = ?", array($network_id));
         $ipv6   = dbFetchRows("SELECT * FROM `ipv6_addresses` WHERE `ipv6_network_id` = ?", array($network_id));
         $ip_addresses_count = count(array_merge($ipv4, $ipv6));
-        if (!is_numeric($ip_addresses_count)) {
-            api_error(500, 'Error retrieving IP Addresses');
-        }
         if ($ip_addresses_count == 0) {
             api_error(404, "IP Network $network_id does not exist or empty");
         }
@@ -1421,8 +1412,8 @@ function list_vrf()
         $vrfs[] = $vrf;
     }
     $total_vrfs = count($vrfs);
-    if (!is_numeric($total_vrfs)) {
-        api_error(500, 'Error retrieving vrfs');
+    if ($total_vrfs == 0) {
+        api_error(404, 'VRFs does not exist');
     }
 
     api_success($vrfs, 'vrfs');
@@ -1442,9 +1433,6 @@ function get_vrf()
 
     $vrf       = dbFetchRows("SELECT * FROM `vrfs` WHERE `vrf_id` IS NOT NULL AND `vrf_id` = ?", array($vrfId));
     $vrf_count = count($vrf);
-    if (!is_numeric($vrf_count)) {
-        api_error(500, 'Error retrieving VRF');
-    }
     if ($vrf_count == 0) {
         api_error(404, "VRF $vrfId does not exist");
     }
@@ -1497,8 +1485,8 @@ function list_vlans()
         $vlans[] = $vlan;
     }
     $vlans_count = count($vlans);
-    if (!is_numeric($vlans_count)) {
-        api_error(500, 'Error retrieving VLANs');
+    if ($vlans_count == 0) {
+        api_error(404, 'VLANs does not exist');
     }
 
     api_success($vlans, 'vlans');
@@ -1517,9 +1505,6 @@ function list_ip_addresses()
     $ipv4_addresses   = dbFetchRows("SELECT * FROM `ipv4_addresses`");
     $ipv6_addresses   = dbFetchRows("SELECT * FROM `ipv6_addresses`");
     $ip_addresses_count = count(array_merge($ipv4_addresses, $ipv6_addresses));
-    if (!is_numeric($ip_addresses_count)) {
-        api_error(500, 'Error retrieving IP Addresses');
-    }
     if ($ip_addresses_count == 0) {
         api_error(404, 'IP Addresses does not exist');
     }
@@ -1540,9 +1525,6 @@ function list_ip_networks()
     $ipv4_networks   = dbFetchRows("SELECT * FROM `ipv4_networks`");
     $ipv6_networks   = dbFetchRows("SELECT * FROM `ipv6_networks`");
     $ip_networks_count = count(array_merge($ipv4_networks, $ipv6_networks));
-    if (!is_numeric($ip_networks_count)) {
-        api_error(500, 'Error retrieving IP Networks');
-    }
     if ($ip_networks_count == 0) {
         api_error(404, 'IP Networks does not exist');
     }
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -250,7 +250,7 @@ function list_locations()
     $locations   = dbFetchRows("SELECT `locations`.* FROM `locations` WHERE `locations`.`location` IS NOT NULL");
     $total_locations = count($locations);
     if ($total_locations == 0) {
-        api_error(404, 'Locations does not exist');
+        api_error(404, 'Locations do not exist');
     }
 
     api_success($locations, 'locations');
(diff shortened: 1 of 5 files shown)


laf:
Thanks for this. Why not add checks to see if a user has perms to see those devices so you don't have to restrict this to read only and above?

zombah:
Changed list_vrf to use device permissions. For get_vrf it wont work because this function has no per device capabilities.

zombah:
I will add couple other new api calls i have to this PR

html/includes/api_functions.inc.php
```diff
@@ -526,6 +526,29 @@ function get_bgp()
...
 
 
+function list_cbgp()
+{
+    $app        = \Slim\Slim::getInstance();
+    $sql        = '';
+    $sql_params = array();
+    $hostname   = $_GET['hostname'] ?: '';
+    $device_id  = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
+    if (is_numeric($device_id)) {
```
laf: We should error if this doesn't return true.

html/includes/api_functions.inc.php
```diff
@@ -822,11 +845,37 @@ function get_ip_addresses()
         check_device_permission($device_id);
         $ipv4   = dbFetchRows("SELECT `ipv4_addresses`.* FROM `ipv4_addresses` JOIN `ports` ON `ports`.`port_id`=`ipv4_addresses`.`port_id` WHERE `ports`.`device_id` = ? AND `deleted` = 0", array($device_id));
         $ipv6   = dbFetchRows("SELECT `ipv6_addresses`.* FROM `ipv6_addresses` JOIN `ports` ON `ports`.`port_id`=`ipv6_addresses`.`port_id` WHERE `ports`.`device_id` = ? AND `deleted` = 0", array($device_id));
+        $ip_addresses_count = count(array_merge($ipv4, $ipv6));
+        if (!is_numeric($ip_addresses_count)) {
+            api_error(500, 'Error retrieving IP Addresses');
+        }
+        if ($ip_addresses_count == 0) {
```
laf: This just means it might have no IPs. I wouldn't indicate the device doesn't exist here.

html/includes/api_functions.inc.php
```diff
@@ -822,11 +845,37 @@ function get_ip_addresses()
...
     } elseif (isset($router['portid'])) {
         $port_id = urldecode($router['portid']);
         check_port_permission($port_id, null);
         $ipv4   = dbFetchRows("SELECT * FROM `ipv4_addresses` WHERE `port_id` = ?", array($port_id));
         $ipv6   = dbFetchRows("SELECT * FROM `ipv6_addresses` WHERE `port_id` = ?", array($port_id));
+        $ip_addresses_count = count(array_merge($ipv4, $ipv6));
+        if (!is_numeric($ip_addresses_count)) {
+            api_error(500, 'Error retrieving IP Addresses');
+        }
+        if ($ip_addresses_count == 0) {
```
laf: Same goes for this.

html/includes/api_functions.inc.php
```diff
@@ -1327,6 +1376,59 @@ function get_devices_by_group()
...
+    $device_id  = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
+    if (is_numeric($device_id)) {
+        $sql        = ' AND `vrfs`.`device_id`=?';
+        $sql_params = array($device_id);
+    }
+    if (!empty($vrfname)) {
+        $sql        = ' AND `vrfs`.`vrf_name`=?';
+        $sql_params = array($vrfname);
+    }
+    check_device_permission($device_id);
```
laf: This needs to be moved into the is_numeric check above otherwise we don't have a device_id potentially. Should fall back to is_read() if `if (is_numeric($device_id)) {` is false.

html/includes/api_functions.inc.php
```diff
@@ -1343,6 +1445,64 @@ function list_ipsec()
     api_success($ipsec, 'ipsec');
 }
 
+
+function list_vlans()
```
laf: Should also allow to pass hostname / device_id to limit the response.

html/includes/api_functions.inc.php
```diff
@@ -1343,6 +1445,64 @@ function list_ipsec()
...
+    $vlans_count = count($vlans);
+    if (!is_numeric($vlans_count)) {
+        api_error(500, 'Error retrieving VLANs');
+    }
+
+    api_success($vlans, 'vlans');
+}
+
+
+function list_ip_addresses()
```
laf: Should also allow to pass hostname / device_id to limit the response.

html/includes/api_functions.inc.php
```diff
@@ -1343,6 +1445,64 @@ function list_ipsec()
...
+    $ip_addresses_count = count(array_merge($ipv4_addresses, $ipv6_addresses));
+    if (!is_numeric($ip_addresses_count)) {
+        api_error(500, 'Error retrieving IP Addresses');
+    }
+
+    api_success(array_merge($ipv4_addresses, $ipv6_addresses), 'ip_addresses');
+}
+
+
+function list_ip_networks()
```
laf: Should also allow to pass hostname / device_id to limit the response.

html/includes/api_functions.inc.php
zombah: This will duplicate functionality of existing get_vlans function. I can add full listing into get_vlans, but i thought more simple functions better than fewer but with more nested conditions.

html/includes/api_functions.inc.php
zombah: Same as get_vlans above but with get_ip_addresses function.

html/includes/api_functions.inc.php
zombah: Networks tables does not have any keys to other tables now, we can make triple join with ip_addresses table then with ports which have device_id, if this suitable i'll update or maybe you have any better idea how to get device_id with single join?

html/includes/api_functions.inc.php
zombah: Will fix in followup.

html/includes/api_functions.inc.php
zombah: Will fix in followup

html/includes/api_functions.inc.php
zombah: Will fix in followup

html/includes/api_functions.inc.php
zombah: Will fix in followup

html/includes/api_functions.inc.php
zombah: Just realized that to have hostname listing is better in list functions, i will add it and in next pr will change get_vlans to more suitable form for get function, to get single vlan by its id.

laf: Thanks, forgot we had list_vlans but yes, better to just allow passing the hostname into that function and append an extra sql query.

html/includes/api_functions.inc.php
laf: Leave it for now as is_read() then.

html/includes/api_functions.inc.php
laf: Then one function with the ability to pass hostname in would be good.

html/includes/api_functions.inc.php
```diff
@@ -526,6 +526,41 @@ function get_bgp()
...
+    foreach (dbFetchRows("SELECT `bgpPeers_cbgp`.* FROM `bgpPeers_cbgp` LEFT JOIN `devices` ON `bgpPeers_cbgp`.`device_id` = `devices`.`device_id` WHERE `bgpPeers_cbgp`.`device_id` IS NOT NULL $sql", $sql_params) as $bgp_counter) {
+        $host_id = get_vm_parent_id($device);
+        $device['ip'] = inet6_ntop($device['ip']);
+        if (is_numeric($host_id)) {
+            $device['parent_id'] = $host_id;
+        }
+        $bgp_counters[] = $bgp_counter;
+    }
+    $total_bgp_counters = count($bgp_counters);
+    if (!is_numeric($total_bgp_counters)) {
```
laf: This check seems pointless to me.

html/includes/api_functions.inc.php
```diff
@@ -822,11 +857,37 @@ function get_ip_addresses()
...
+        if ($ip_addresses_count == 0) {
+            api_error(404, "Port $port_id does not have IP Addresses");
+        }
+    } elseif (isset($router['id'])) {
+        check_is_read();
+        $network_id = $router['id'];
+        $ipv4   = dbFetchRows("SELECT * FROM `ipv4_addresses` WHERE `ipv4_network_id` = ?", array($network_id));
+        $ipv6   = dbFetchRows("SELECT * FROM `ipv6_addresses` WHERE `ipv6_network_id` = ?", array($network_id));
+        $ip_addresses_count = count(array_merge($ipv4, $ipv6));
+        if (!is_numeric($ip_addresses_count)) {
```
laf: Again, I think you can probably get rid of these.

html/includes/api_functions.inc.php
```diff
@@ -526,6 +526,41 @@ function get_bgp()
...
+    $sql_params = array();
+    $hostname   = $_GET['hostname'] ?: '';
+    $device_id  = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
+    if (is_numeric($device_id)) {
+        check_device_permission($device_id);
+        $sql        = " AND `devices`.`device_id` = ?";
+        $sql_params[] = $device_id;
+    }
+    if (!is_admin() && !is_read()) {
+        $sql .= " AND `bgpPeers_cbgp`.`device_id` IN (SELECT device_id FROM devices_perms WHERE user_id = ?)";
```
laf: Can you convert this sub query to a JOIN like the other code has for user lookup so things are consistent. Same goes for any other sub queries you do.

html/includes/api_functions.inc.php
```diff
@@ -1327,6 +1388,71 @@ function get_devices_by_group()
...
+    $vrfs       = array();
+    foreach (dbFetchRows("SELECT `vrfs`.* FROM `vrfs` LEFT JOIN `devices` ON `vrfs`.`device_id` = `devices`.`device_id` WHERE `vrfs`.`vrf_name` IS NOT NULL $sql", $sql_params) as $vrf) {
+        $host_id = get_vm_parent_id($device);
+        $device['ip'] = inet6_ntop($device['ip']);
+        if (is_numeric($host_id)) {
+            $device['parent_id'] = $host_id;
+        }
+        $vrfs[] = $vrf;
+    }
+    $total_vrfs = count($vrfs);
```
laf: Same here, probably can remove this check.

html/includes/api_functions.inc.php
```diff
@@ -1327,6 +1388,71 @@ function get_devices_by_group()
...
+    $app    = \Slim\Slim::getInstance();
+    $router = $app->router()->getCurrentRoute()->getParams();
+    $vrfId  = $router['id'];
+    if (!is_numeric($vrfId)) {
+        api_error(400, 'Invalid id has been provided');
+    }
+
+    $vrf       = dbFetchRows("SELECT * FROM `vrfs` WHERE `vrf_id` IS NOT NULL AND `vrf_id` = ?", array($vrfId));
+    $vrf_count = count($vrf);
+    if (!is_numeric($vrf_count)) {
```
laf: Same again, drop this.

html/includes/api_functions.inc.php
```diff
@@ -1343,6 +1469,88 @@ function list_ipsec()
...
+    foreach (dbFetchRows("SELECT `vlans`.* FROM `vlans` LEFT JOIN `devices` ON `vlans`.`device_id` = `devices`.`device_id` WHERE `vlans`.`vlan_vlan` IS NOT NULL $sql", $sql_params) as $vlan) {
+        $host_id = get_vm_parent_id($device);
+        $device['ip'] = inet6_ntop($device['ip']);
+        if (is_numeric($host_id)) {
+            $device['parent_id'] = $host_id;
+        }
+        $vlans[] = $vlan;
+    }
+    $vlans_count = count($vlans);
+    if (!is_numeric($vlans_count)) {
```
laf: Same again + the other count checks you have.

html/includes/api_functions.inc.php
```diff
@@ -239,6 +239,24 @@ function get_graph_generic_by_hostname()
...
+{
+    check_is_read();
+
+    $app           = \Slim\Slim::getInstance();
+    $router        = $app->router()->getCurrentRoute()->getParams();
+
+    $locations   = dbFetchRows("SELECT `locations`.* FROM `locations` WHERE `locations`.`location` IS NOT NULL");
+    $total_locations = count($locations);
+    if ($total_locations == 0) {
+        api_error(404, 'Locations does not exist');
```
laf: `Locations do not exist`

html/includes/api_functions.inc.php
```diff
@@ -822,11 +875,28 @@ function get_ip_addresses()
         check_device_permission($device_id);
         $ipv4   = dbFetchRows("SELECT `ipv4_addresses`.* FROM `ipv4_addresses` JOIN `ports` ON `ports`.`port_id`=`ipv4_addresses`.`port_id` WHERE `ports`.`device_id` = ? AND `deleted` = 0", array($device_id));
         $ipv6   = dbFetchRows("SELECT `ipv6_addresses`.* FROM `ipv6_addresses` JOIN `ports` ON `ports`.`port_id`=`ipv6_addresses`.`port_id` WHERE `ports`.`device_id` = ? AND `deleted` = 0", array($device_id));
+        $ip_addresses_count = count(array_merge($ipv4, $ipv6));
+        if ($ip_addresses_count == 0) {
+            api_error(404, "Device $device_id does not have IP Addresses");
```
laf: `Device $device_id does not have any IP addresses`

html/includes/api_functions.inc.php
```diff
@@ -822,11 +875,28 @@ function get_ip_addresses()
...
+        }
     } elseif (isset($router['portid'])) {
         $port_id = urldecode($router['portid']);
         check_port_permission($port_id, null);
         $ipv4   = dbFetchRows("SELECT * FROM `ipv4_addresses` WHERE `port_id` = ?", array($port_id));
         $ipv6   = dbFetchRows("SELECT * FROM `ipv6_addresses` WHERE `port_id` = ?", array($port_id));
+        $ip_addresses_count = count(array_merge($ipv4, $ipv6));
+        if ($ip_addresses_count == 0) {
+            api_error(404, "Port $port_id does not have IP Addresses");
```
laf: `Port $port_id does not have any IP addresses`

html/includes/api_functions.inc.php
```diff
@@ -822,11 +875,28 @@ function get_ip_addresses()
...
+        }
+    } elseif (isset($router['id'])) {
+        check_is_read();
+        $network_id = $router['id'];
+        $ipv4   = dbFetchRows("SELECT * FROM `ipv4_addresses` WHERE `ipv4_network_id` = ?", array($network_id));
+        $ipv6   = dbFetchRows("SELECT * FROM `ipv6_addresses` WHERE `ipv6_network_id` = ?", array($network_id));
+        $ip_addresses_count = count(array_merge($ipv4, $ipv6));
+        if ($ip_addresses_count == 0) {
+            api_error(404, "IP Network $network_id does not exist or empty");
```
laf: `IP Network $network_id does not exist or is empty`

html/includes/api_functions.inc.php
```diff
@@ -1327,6 +1397,68 @@ function get_devices_by_group()
...
+        $host_id = get_vm_parent_id($device);
+        $device['ip'] = inet6_ntop($device['ip']);
+        if (is_numeric($host_id)) {
+            $device['parent_id'] = $host_id;
+        }
+        $vrfs[] = $vrf;
+    }
+    $total_vrfs = count($vrfs);
+    if ($total_vrfs == 0) {
+        api_error(404, 'VRFs does not exist');
```
laf: `VRFs do not exist`

html/includes/api_functions.inc.php
```diff
@@ -1343,6 +1475,82 @@ function list_ipsec()
...
+        $host_id = get_vm_parent_id($device);
+        $device['ip'] = inet6_ntop($device['ip']);
+        if (is_numeric($host_id)) {
+            $device['parent_id'] = $host_id;
+        }
+        $vlans[] = $vlan;
+    }
+    $vlans_count = count($vlans);
+    if ($vlans_count == 0) {
+        api_error(404, 'VLANs does not exist');
```
laf: `VLANs do not exist`

html/includes/api_functions.inc.php
```diff
@@ -1343,6 +1475,82 @@ function list_ipsec()
...
+    $app            = \Slim\Slim::getInstance();
+    $router         = $app->router()->getCurrentRoute()->getParams();
+    $ipv4_addresses = array();
+    $ipv6_addresses = array();
+
+    $ipv4_addresses   = dbFetchRows("SELECT * FROM `ipv4_addresses`");
+    $ipv6_addresses   = dbFetchRows("SELECT * FROM `ipv6_addresses`");
+    $ip_addresses_count = count(array_merge($ipv4_addresses, $ipv6_addresses));
+    if ($ip_addresses_count == 0) {
+        api_error(404, 'IP Addresses does not exist');
```
laf: `IP addresses do not exist`

html/includes/api_functions.inc.php
```diff
@@ -1343,6 +1475,82 @@ function list_ipsec()
...
+    $app           = \Slim\Slim::getInstance();
+    $router        = $app->router()->getCurrentRoute()->getParams();
+    $ipv4_networks = array();
+    $ipv6_networks = array();
+
+    $ipv4_networks   = dbFetchRows("SELECT * FROM `ipv4_networks`");
+    $ipv6_networks   = dbFetchRows("SELECT * FROM `ipv6_networks`");
+    $ip_networks_count = count(array_merge($ipv4_networks, $ipv6_networks));
+    if ($ip_networks_count == 0) {
+        api_error(404, 'IP Networks does not exist');
```
laf: `IP networks do not exist`

zombah:
@laf Thank you for review! Where were my eyes... Submited fix

laf:
Thanks @zombah :D

PR #8020, opened 2018-01-04
Fix: Override -H for check_load service. Fixed backslash escaping.

Zmegolaz:
- check_load does not use the -H option, and need a service file to override that.
- I combined the load1, load5 and load15 values into one graph with multiple lines.
- The default graph and the ICMP graph had an extra \ at the end of every GPRINT line.
- The mysql service had an unescaped backslash.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


kkrumm1:
@laf Perhaps a youtube video going over services and how they work.

laf:
> @laf Perhaps a youtube video going over services and how they work.

I'll see when I get time.

PR #8026, opened 2018-01-04
docs: importance attaching services to a device

kkrumm1:
in librenms. Otherwise alerting on service it won't work. :)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Services.md
+++ b/doc/Extensions/Services.md
@@ -3,8 +3,8 @@ source: Extensions/Services.md
 
 Services within LibreNMS provides the ability to use Nagios plugins to perform additional monitoring outside of SNMP.
 
-These services are tied into an existing device so you need at least one device that supports SNMP to be able to add it
-to LibreNMS - localhost is a good one.
+**These services are tied into an existing device so you need at least one device that supports SNMP to be able to add it
+to LibreNMS - localhost is a good one. This is needed in order for alerting to work properly.** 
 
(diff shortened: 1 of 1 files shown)


doc/Extensions/Services.md
```diff
@@ -3,8 +3,8 @@ source: Extensions/Services.md
 
 Services within LibreNMS provides the ability to use Nagios plugins to perform additional monitoring outside of SNMP.
 
-These services are tied into an existing device so you need at least one device that supports SNMP to be able to add it
-to LibreNMS - localhost is a good one.
+**These services are tied into an existing device so you need at least one device that supports SNMP to be able to add it
```
murrant: It doesn't need to support snmp any more.

kkrumm1: Yes but the service check has to be attached to a device that is in librenms in order for the alerting to work. Also how sure are you on the SNMP? https://community.librenms.org/t/service-check-alert/2954/25

murrant: well crap, looks like we need to update the code to allow that!

murrant: @kkrumm1 I could not reproduce the need for snmp...

kkrumm1: Yeah, I saw the post... I was just going off the word of amaizenshtein.

PR #8029, opened 2018-01-04
Place hostname first on add device page.

murrant:
Not sure why but the existing order bothers me every time.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
:) It's always the small things that annoy us eh.

murrant:
I find that that annoyance is a strong indicator of something that isn't quite right.

The best UI is one you don't have to think about :)

PR #8033, opened 2018-01-05
device: Enable cisco-vrf discovery for junos os

zombah:
As cisco-vrf now works fine with Junos os it seems fine
to enable it by default

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Bad module name then if it's not cisco specific :(

I think we should update this to be just vrf, that will mean file name changes, doc updates, a check for the old config variable in includes/process_config.inc.php and set the new one to true if the old one was and then `discover_cisco-vrf` updating to `discover_vrf` in the `devices_attribs` table to that db selected modules work.

It's a big ask so if you don't fancy it, let me know and I'll do that.

zombah:
Agree. Sure i'll do. VRF is specific for Cisco and Arista, maybe to choose vendor agnostic name, like mib name, mpls-vpn or mpls-l3vpn?

laf:
You pick, basically it just needs to represent what it does now seeing as it's no longer cisco vrf specific.

Zmegolaz:
I'd say "VRF" is fine, it's not vendor specific and is used in Juniper too. ("set instance-type vrf")

zombah:
Yes i'll keep vrf, i found that it is used alot in webui,docs and etc places so i see no reason to create segmentation with other term variants.

murrant:
Looks good to me too.   (obviously needs the schema to be moved up)

zombah:
Rebased versus current master

laf:
Awesome work @zombah. Thanks for going through all of that :D

PR #8038, opened 2018-01-06
webui: device-groups refresh

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

before:
screen shot 2018-01-06 at 8 32 28 pm
screen shot 2018-01-06 at 8 32 52 pm

after:
screen shot 2018-01-06 at 8 31 55 pm
screen shot 2018-01-06 at 8 32 07 pm


laf:
Not sure about this one, we don't have a similar actions menu anywhere else I don't think so it not looks odd and unnatural.

crcro:
we (maybe) have something like this on ports page -> bulk remove deleted ports

laf:
Yeah that was also only added recently.

aldemira:
What about turning that action link into a button?

crcro:
i tried but looks a bit bigger even with btn-sm in my opinion ...

murrant:
I think it should be a button.  A link makes it seem like it goes somewhere. A button makes it seem like it does something.

If it is too big probably move it out of the header.

laf:
@crcro any update on this one?

laf:
@crcro Feel free to reply to this when you have time but for now we are closing the PR.

PR #8042, opened 2018-01-07
Move install to php53 branch if running an unsupported version of PHP

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Validations/Php.php
+++ b/LibreNMS/Validations/Php.php
@@ -53,7 +53,7 @@ class Php implements ValidationGroup
 
         // if update is not set to false and version is min or newer
         if (Config::get('update') && version_compare(PHP_VERSION, $min_version, '<')) {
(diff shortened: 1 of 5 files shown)


murrant:
Ok, I've tested this:

 * no updates
 * php 5.5 switch to php53
 * php 5.5 update in php53 branch
 * php 7.2 update switch to master
 * php 7.2 update in master branch

For the release version, I propose we create a new lightweight tag that way we can re-point the tag if we need to push an update.  Otherwise it will be stuck on the hardcoded sha until PHP is updated.

murrant:
We also need to decide a branch point for php53.  Right now it is at 1.35.

laf:
Can we not just move to a branch rather than a tag?

murrant:
I guess it would be the same thing effectively, no reason we need to use tags even if they are on the "release" train.  Thanks for clearing that up @laf.

murrant:
Both release and master will end up in the same php53 branch.  Seems pretty good.

This is pretty hard to test, but I have tested all the ones I could think of.

murrant:
https://youtu.be/jrA-1cG_wq4?t=24

PR #8043, opened 2018-01-07
feature: apps versioning

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

this would imply small changes to snmp extend scripts and polling scripts.
test was done using modified os-updates script.

screen shot 2018-01-07 at 11 51 58 pm


murrant:
Wouldn't it be better to do this with an alert?  It wouldn't require any custom code.

kkrumm1:
We have alert for this in the Alerts Collection - 
image

crcro:
@kkrumm1 .... it's not about the values returned by apps but about the scripts version itself ... ie: os-updates is now v1, you change it now and add new detection or fix something in script and make it v1.1, the user doesn't know about this change unless let's say he is constantly monitoring librenms-agent repo

murrant:
Oh, interesting idea.

crcro:
i think that this could be the inform part, but still struggling with travis and db change :|

laf:
Not sure why you've tagged me in that :)

I'm not a fan of the applications at all, they are flaky most of the time so I'm not going to get in the way of this PR as I'm in the position that I won't do any work on it anyway.

Having said that I'm not sure this is the right way to go :)

crcro:
Besides your view of apps part :) you have something against maintaining db tables ... if i recall well from past :)

laf:
> you have something against maintaining db tables ... if i recall well from past :)

You don't recall well (or I don't). I can't ever remember this discussion before. However:

  - Easily forgotten about and the table not kept up to date.
  - It gives me the impression that apps are changeable which will most likely lead to loss in data for people. Like the core code we have, it should be done right the first time.
  - Logging that apps are out of date every 5 minutes seems excessive.
  - What do you do about people who have installed a newer version of the app but is running old code (even down to them using the stable release of LibreNMS).

murrant:
I'm not a fan of storing application data in the database.  If this table gets corrupted or changed there is no easy way to fix and if a user changes it, we do not know about that.  Say the table gets truncated, it is just empty forever then.

crcro:
event entries a little more detailed, maybe something to docs should be added also to address a case of outdated app

screen shot 2018-01-09 at 12 34 58 pm

laf:
Personally, this is a no from me.

murrant:
I'm fine with a no too.

As I expressed in the past, we should really be using external solutions for this, such as collectd, but I don't have enough motivation to push that.

PR #8047, opened 2018-01-08
fix: ldap-authorization should be considered external

pmcnabb:
When sso auth type was added (1c6b7a9), some code for getting a username
was moved into HttpAuthAuthorizer. LdapAuthorizationAuthorizer uses the
same process to get the username and needs to be updated as well.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
This is actually the case for ADAuthorizationAuthorizer too.

Can you push getExternalUsername() up to AuthorizerBase and remove the individual implementations (except for SSO)?

pmcnabb:
I've tested this branch with an LdapAuthorizationAuthorizer configuration and it is working correctly.

PR #8052, opened 2018-01-09
Add snmp_getnext_multi() function

neg2led:
Added snmp_getnext_multi() function using syntax of existing snmp_get_multi() function.

This replicates the snmp_get_multi() function but calls snmpgetnext instead, returning an array like so;

```
array(1) {
  [4]=>
  array(2) {
    ["dot11manufacturerProductName"]=>
    string(15) "UAP-AC-Pro-Gen2"
    ["dot11manufacturerProductVersion"]=>
    string(35) "BZ.qca956x.v3.7.37.6065.170118.0908"
  }
}
```

This replicates the output of snmp_get_multi() - we won't know the index in advance, but consistency of outputs seems worth doing.

See #8005 for further details :)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/snmp.inc.php
+++ b/includes/snmp.inc.php
@@ -335,13 +335,10 @@ function snmp_getnext_multi($device, $oids, $options = '-OQUs', $mib = null, $mi
         $value             = trim($value, "\" \n\r");
         list($oid, $index) = explode('.', $oid, 2);
         if (!str_contains($value, 'at this OID')) {
-            if (is_null($index)) {
-                if (empty($oid)) {
-                    continue; // no index or oid
-                }
-                $array[$oid] = $value;
+            if (empty($oid)) {
+                continue; // no index or oid
             } else {
-                $array[$index][$oid] = $value;
+                $array[$oid] = $value;
             }
         }
     }
--- a/includes/snmp.inc.php
+++ b/includes/snmp.inc.php
@@ -342,7 +342,7 @@ function snmp_getnext_multi($device, $oids, $options = '-OQUs', $mib = null, $mi
             }
         }
     }
-    recordSnmpStatistic('snmpget', $time_start);
+    recordSnmpStatistic('snmpgetnext', $time_start);
     return $array;
 }//end snmp_getnext_multi()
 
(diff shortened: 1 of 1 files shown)


murrant:
Because this is getnext, perhaps omitting the index seems appropriate. Otherwise, how will you find the data in the result array?

For example we get firstOid and secondOid, but first oid is at index 32 and the second index is at index 43

Now we have an array like this:
```
array(
    32 => array('firstOid' => 'value'),
    43 => array('secondOid' => 'value),
)
```
You'd have to iterate through the array and search for it.

I think this makes more sense:
```
array(
    'firstOid' => 'value',
    'secondOid' => 'value',
)
```

neg2led:
I was looking to preserve the index number without having to account for it being a part of the key, since it's something that isn't known at the time of calling the function (and might be different for each returned result). We could return this;
 ```
array(
    'firstOid.index' => 'value',
    'secondOid.index' => 'value',
)
```
But now our calling routine won't know what it should be looking for as a key. Maybe a bit of both?

```
array(
    firstOid => array('value' => 'sometext', 'index' => 'somenumber'),
    secondOid => array('value' => 'sometext', 'index' => 'somenumber'),
)
```
Then it's as simple as $data['firstOid']['value'] but the index isn't lost along the way - If we don't mind losing the index entirely then I'm totally happy to strip it, though.

**Comment by murrant** [comment]
No, I'd say if you want to preserve the oid use the (non-existent)
`snmp_getnext_multi_oid() `

**Comment by neg2led** [comment]
Makes sense. Done and done :)

File: includes/snmp.inc.php [hunk 1]
```diff
@@ -307,6 +307,45 @@ function snmp_getnext($device, $oid, $options = null, $mib = null, $mibdir = nul
...
+        list($oid, $index) = explode('.', $oid, 2);
+        if (!str_contains($value, 'at this OID')) {
+            if (empty($oid)) {
+                continue; // no index or oid
+            } else {
+                $array[$oid] = $value;
+            }
+        }
+    }
+    recordSnmpStatistic('snmpget', $time_start);
```
murrant: The stat should be 'snmpgetnext'

PR #8055, opened 2018-01-09
Add support for ArbOS

TheMysteriousX:
This adds basic support for PeakFlow appliances. They don't really expose anything interesting by SNMP beyond what a generic device has.

I've referred to everything as 'arbos', but I don't know if the behaviour here covers APS/TMS devices too.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


TheMysteriousX:
I don't understand the test failure - I've not done anything that should affect 3com devices.

tests/snmpsim/arbos.snmprec
```diff
@@ -0,0 +1,630 @@
+.1.3.6.1.2.1.1.1.0 = STRING: Peakflow SP 8.3 (build HIU6)  System Board Model: S2600CO Serial Number: CG22541072COI
```
murrant: This is an snmpwalk, not an snmprec file.  The format is wrong.

That is what is causing tests to fail.

See https://docs.librenms.org/Developing/os/Test-Units/

murrant:
You can also remove the empty file includes/definitions/discovery/arbos.yaml (unless you intend to add some sensors there)

TheMysteriousX:
thanks for the pointer on the snmprec file - not sure how I ended up with a walk capture there.

I'll remove other yaml file - there aren't any sensors exposed that I can find.

One thing I'll try and add is load average, which is exposed through their own MIB. Raw CPU information isn't exposed.

TheMysteriousX:
I can think of three ways to do this. 

* I can tack it on to the end of ```includes/polling/ucd-mib.inc.php``` (logical, as it is the same information, just exposed at a different oid)
* I could add it into ```includes/polling/os/arbos.inc.php``` and still place it into the ```ucd_load``` rrd (may conflict with ```ucd.inc.php``` if Arbor start using standard oids)
* I can add an additional (duplicate) load average definition.

Any of these preferred (or something else?).

laf:
So this is a bit tricky as ucd-mib does a lot more than just load average but I don't think I'm personally a fan of tacking it on to the end of the os polling.

Maybe be worth adding a directory `includes/polling/ucd-mib/` and have the new OS file in that.

Then before where the load snmp_get is called do an is_file check for `includes/polling/ucd-mib/$os.inc.php`, if it exists load it, do you call for data in there and populate the variables needed for the rrd file. Then do `} else {` and the normal snmp call - make sense?

@murrant - Thoughts? Again this is better when it gets converted to OOP but seems an ok way to deal with this and any future devices that require similar.

murrant:
@laf In the processor rewrite branch, we always check for ucd mib if no other processors have been found.

TheMysteriousX:
Sounds like the best thing to do would be to wait until #8066 is merged before I add load averages in.

I've cleaned up this request, so I'm happy to submit for review, and I'll open another request with load at a later date.

laf:
> @laf In the processor rewrite branch, we always check for ucd mib if no other processors have been found.

Talking about load here rather than processors stuff so not related.

PR #8058, opened 2018-01-09
Add dependency info for device list in api calls

aldemira:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -315,8 +315,9 @@ function list_devices()
     }
     $devices = array();
     foreach (dbFetchRows("SELECT * FROM `devices` $join WHERE $sql ORDER by $order", $param) as $device) {
-        $dev_deps = dbFetchColumn("SELECT parent_device_id from device_relationships WHERE child_device_id = ?", array($device));
-        if ($dev_deps) {
+        $dev_deps = dbFetchRows("SELECT dd.device_id AS device_id, dd.hostname AS hostname FROM devices as d LEFT JOIN device_relationships AS dr ON dr.child_device_id = d.device_id LEFT JOIN devices AS dd ON dr.parent_device_id = dd.device_id WHERE d.device_id = ?", array($device));
+        
+        if (sizeof($dev_deps) > 1) {
             $device['depends_on'] = $dev_deps;
         }
         unset($dev_deps);
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -315,15 +315,11 @@ function list_devices()
     }
     $devices = array();
     foreach (dbFetchRows("SELECT * FROM `devices` $join WHERE $sql ORDER by $order", $param) as $device) {
-        $dev_deps = dbFetchRows("SELECT parent_device_id from device_relationships WHERE child_device_id = ?", array($device));
+        $dev_deps = dbFetchColumn("SELECT parent_device_id from device_relationships WHERE child_device_id = ?", array($device));
         if ($dev_deps) {
-            $tmp_arr = array();
-            foreach ($dev_deps as $dep) {
-                $tmp_arr[] = $dep['parent_device_id'];
-            }
-            $device['depends_on'] = $tmp_arr;
-            unset($tmp_arr);
+            $device['depends_on'] = $dev_deps;
         }
+        unset($dev_deps);
         $host_id = get_vm_parent_id($device);
         $device['ip'] = inet6_ntop($device['ip']);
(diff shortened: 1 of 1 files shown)


html/includes/api_functions.inc.php
```diff
@@ -315,6 +315,15 @@ function list_devices()
     }
     $devices = array();
     foreach (dbFetchRows("SELECT * FROM `devices` $join WHERE $sql ORDER by $order", $param) as $device) {
+        $dev_deps = dbFetchRows("SELECT parent_device_id from device_relationships WHERE child_device_id = ?", array($device));
```
murrant: dbFetchColumn puts a single column in a nice array ;)

html/includes/api_functions.inc.php
```diff
@@ -315,6 +315,11 @@ function list_devices()
     }
     $devices = array();
     foreach (dbFetchRows("SELECT * FROM `devices` $join WHERE $sql ORDER by $order", $param) as $device) {
+        $dev_deps = dbFetchColumn("SELECT parent_device_id from device_relationships WHERE child_device_id = ?", array($device));
```
laf: I already expect that people will ask for hostnames as well as ID's. Here's a totally unformatted but working example:

`select d.device_id, d.hostname, GROUP_CONCAT(dd.device_id), GROUP_CONCAT(dd.hostname) as parent_hostname from devices as d left join device_relationships as dr on dr.child_device_id=d.device_id left join devices as dd on dr.parent_device_id=dd.device_id GROUP BY d.hostname`

murrant:
Any thoughts on this being under it's own route?  Or some way to mark with parents or children?

laf:
> Any thoughts on this being under it's own route? Or some way to mark with parents or children?

How come?

aldemira:
I know using sizeof() is a bit silly, but empty() if(!) doesn't work. Let me know if you have a better solution.

laf:
@aldemira What about my comment about the query?

`select d.device_id, d.hostname, GROUP_CONCAT(dd.device_id), GROUP_CONCAT(dd.hostname) as parent_hostname from devices as d left join device_relationships as dr on dr.child_device_id=d.device_id left join devices as dd on dr.parent_device_id=dd.device_id GROUP BY d.hostname`

aldemira:
@laf Yeah, I got and simplified it so that if is a JSON array like 
[{ device_id: X, hostname: aaa}, {device_id: Y, hostname: bbb}] 
Isn't this what you wanted?

aldemira:
@laf I'm sorry, now I get it :)

laf:
Ok wasn't sure on the formatting of this in the end after seeing it returned as part of the json.

So: https://p.libren.ms/view/raw/a591f48e

Output below. I have removed the p.* from the select as I don't think we should be returning all the data for mac, ipv4 or ipv6 if someone is searching on it and just list the device info.

I've also added this to the get_device call so it works on single devices:

```
{
    "status": "ok",
    "devices": [
        {
            "device_id": "366",
            "hostname": "laf2",
            "sysName": "laf2",
            "ip": null,
            "community": "librenms",
            "authlevel": null,
            "authname": null,
            "authpass": null,
            "authalgo": null,
            "cryptopass": null,
            "cryptoalgo": null,
            "snmpver": "v2c",
            "port": "161",
            "transport": "udp",
            "timeout": null,
            "retries": null,
            "snmp_disable": "0",
            "bgpLocalAs": null,
            "sysObjectID": null,
            "sysDescr": null,
            "sysContact": null,
            "version": null,
            "hardware": "",
            "features": null,
            "location": null,
            "os": "generic",
            "status": "0",
            "status_reason": "icmp",
            "ignore": "0",
            "disabled": "0",
            "uptime": null,
            "agent_uptime": "0",
            "last_polled": null,
            "last_poll_attempted": null,
            "last_polled_timetaken": null,
            "last_discovered_timetaken": null,
            "last_discovered": null,
            "last_ping": null,
            "last_ping_timetaken": null,
            "purpose": null,
            "type": "server",
            "serial": null,
            "icon": null,
            "poller_group": "0",
            "override_sysLocation": "0",
            "notes": null,
            "port_association_mode": "1",
            "attribs": [],
            "dependancy": [
                {
                    "hostname": "laf",
                    "parent_device_id": "168"
                },
                {
                    "hostname": "localhost",
                    "parent_device_id": "398"
                }
            ]
        }
    ],
    "count": 1
}
```

laf:
@murrant how do you feel about the above?

murrant:
Looks fine to me.

PR #8060, opened 2018-01-09
 fix: Fixed the default fping config + changed variable name

laf:
Changed millisec to interval to match up with fping params name.
Changed default interval value to be same as timeout as code forces that anyway.
Added docs to try and explain the implications of setting a high timeout + performance docs to speed up polling.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


doc/Support/Performance.md
```diff
@@ -60,6 +60,31 @@ You can also set this globally with the config option `$config['snmp']['max_oid'
...
+You can change some of the default fping options used globally or per device. The defaults are:
+
+```php
+$config['fping_options']['retries'] = 3;
+$config['fping_options']['timeout'] = 500;
+$config['fping_options']['count']   = 3;
+$config['fping_options']['interval'] = 500;
+```
+
+If you're devices are slow to respond then you will need to increase the timeout value and potentially the interval value.
```
Zmegolaz: "you're" -> "your"

PR #8066, opened 2018-01-09
Major Processors rewrite

murrant:
# Generic changes
YAML definition support!
HOST-RESOURCES-MIB is now only tried if no other processors are found. (before was always ran)
UCD-SNMP-MIB is tried if if HOST-RESOURCES-MIB finds no processors (before was never tried, except manually selected os)
More generic and reusable discovery/poller module infrastructure.

# Processor types that are fixed
bdcom
binos
riverbed
saf-integra
ftos/dnos C Series
powerconnect Some models (changed index too)
enterasys (changed index too)
zynos (maybe)

# New support added
dnos Z Series

# Processors that will lose historical data
You can restore the old data by renaming the file to the new file name after the change.
## Changed index
fiberhome

## Changed to standard hr mib
axos
calix
dsm

## Changed to standard ucd mib
hpblmos

# Missing test data
vrp

* Most have guessed data based on previous processor discovery files.  This is usually indicated by usage % of 42.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

  fixes: #8098


laf:
I'll run this overnight to test.

One question, is this worth implementing (and in wireless as well) a cleanup function so we can run either when someone disables a module in the webui (or maybe just when poller next runs) which will drop the data from the db for this module and the device?

laf:
I have just merged in #7850 which added more cpu data so this will need to be taken in to account here. Sorry about that :(

murrant:
Sure we can do that, but lets keep it separate, this PR is already way to large.

laf:
I'll get some debug info soon but:

```
  | 2018-01-12 09:08:15 |   | Processor added: type hr index 196608 descr Intel Xeon E3-12xx v2 (Ivy Bridge)
-- | -- | -- | --
  | 2018-01-12 09:08:15 |   | Processor added: type hr index 196609 descr Intel Xeon E3-12xx v2 (Ivy Bridge)
  | 2018-01-12 00:33:27 |   | Processor Removed: hr 196608 Intel Xeon E3-12xx v2 (Ivy Bridge)
  | 2018-01-12 00:33:27 |   | Processor Removed: hr 196609 Intel Xeon E3-12xx v2 (Ivy Bridge)
```

Latest entry is me switching back to master.

image

murrant:
Notes:

Changing RRD index:

Missing test data:
vrp

laf:
Helpful at all? https://p.libren.ms/view/1071c01c

It's just a standard linux vm so nothing special

murrant:
@laf Seems fine until this: 
`SQL[DELETE T FROM `processors` T LEFT JOIN `devices` ON `devices`.`device_id` = T.`device_id` WHERE `devices`.`device_id` IS NULL]`

laf:
I'll run this overnight.

One big blocker for me is the loss in data, we shouldn't be doing that to users. If it's being done to tidy things up then we should allow those OSs that would lose data to stick with slightly less than ideal code to keep the historical data until such time we can find a way to rename files for all users.

It's definitely an amazing step forward so don't let the above annoy you, it's just we set out not to break installs where we can and this seems like it's not a small break for quite a few users.

> HOST-RESOURCES-MIB is now only tried if no other processors are found. (before was always ran)

Do you think we might hit instances where the old way was required or is that just most likely duplicate processors that were discovered previously.

murrant:
Renaming sounds fine, but I don't want it in the poller like it has been done previously.  Perhaps another process in the daily.sh.  I have minimized the amount of rrd name changes.
Reasons:
avaya-ers, sgos: zero-padded the index and are in a table, it seemed unreasonable to add a lot of code/yaml complexity for these two
fiberhome: processors are in a table, was previously hardcoded with two processors and with a mis-matched index
The rest are due to migrating to standard code instead of custom code.

I saw very little evidence that any of our os code implemented both custom HR and a custom MIB.  I don't expect anyone will lose anything here.

laf:
> Renaming sounds fine, but I don't want it in the poller like it has been done previously. Perhaps another process in the daily.sh. I have minimized the amount of rrd name changes.

Daily would be fine running alongside something like the schema updates. The biggest issue is rrdtool that now supports rrdcached without a shared nfs meaning we have no direct access to files.

If you're adamant that the best way forward with those OS' is to break the data we store then you'll need to do a notification to users warning them of this so they can manually rename if needs be.

laf:
Is this now ready to go? Are we doing a notification to warn people some data is expected to be lost?

murrant:
I am fine with a notification.  Based on my testing very few should be affected negatively.  I'll work up a notification, how much lead time?  A week?

laf:
~5 days?

Also, could you add docs for adding processors via yaml?

murrant:
Merge Feb 5.

PR #8068, opened 2018-01-11
Quick script to rename mibs to their proper names

murrant:
So I can stop re-writing it ;)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/scripts/rename-mibs.php
+++ b/scripts/rename-mibs.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * rename-mibs.php
+ *
+ * Rename mib files to the proper file name to match the mib name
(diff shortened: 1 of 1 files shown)


laf:
Is this not just a little easier to exec `grep 'DEFINITIONS ::= BEGIN' ./mibs/IF-MIB|cut -d ' ' -f1`?

murrant:
@laf the problem is some mib files have the definitions spread out over multiple lines and even comments in the middle...

murrant:
I've coded this tool about 3 times, so I figured I might as well save it so I don't keep writing it.

PR #8070, opened 2018-01-11
Drop tests for unsupported php versions

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,10 @@ services:
 matrix:
   fast_finish: true
   include:
+  - php: 7.2
   - php: 7.1
+    env: SKIP_STYLE_CHECK=1
   - php: 7.0
-    env: SKIP_STYLE_CHECK=1
-  - php: 5.3
-    sudo: required
-    dist: precise
-    env: SKIP_STYLE_CHECK=1
-  - php: 5.4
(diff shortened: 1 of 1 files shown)


laf:
Absolutely fine with this and you can just merge, but should we add `SKIP_UNIT_CHECK` as well along with `SKIP_STYLE_CHECK`.

murrant:
I don't think we should skip unit tests, might as well disable that PHP version all together if we skip unit tests.

That is why I disable PHP 7.0 when I sent that earlier PR.  I didn't think the middle versions were required.

laf:
> I don't think we should skip unit tests, might as well disable that PHP version all together if we skip unit tests.
> 
> That is why I disable PHP 7.0 when I sent that earlier PR. I didn't think the middle versions were required.

Surely between versions only the lint check is needed as we want to test the code works on those versions. phpunit should be the same no matter what the version then.

murrant:
lint checks the code for syntax not functionality.  We honestly only need to run lint on the oldest version because we are stuck with that syntax and cannot use newer language features.

At one point had some code that was fine syntax wise, but failed to run on certain versions of php 5.3.  phpunit would catch this, lint would not.

PR #8075, opened 2018-01-11
Prep for composer changes

murrant:
Non-disruptive changes for composer update

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


composer.json
```diff
@@ -13,7 +13,8 @@
         "irc": "irc://irc.freenode.org/#librenms"
     },
     "config": {
-        "platform": {"php": "5.3.9"}
+        "platform": {"php": "5.3.9"},
```
laf: FYI on this: https://github.com/symfony/symfony-standard/issues/858

I'm not sure we should remove this line here or if not having the lock file is fine. However we should actually ship the composer.lock file: https://ilikekillnerds.com/2014/09/should-you-commit-composer-lock-into-your-git-repository/

murrant: Yes, we will definitely ship composer.lock.  But it will be in the switch over PR.

The reason we hard-coded platform was to keep our dependencies at older versions.  Now that 5.6 is the minimum, we can remove that as well.  but not yet.

murrant: Actually, I can put composer.lock here and change that.

murrant:
Hmm, the problem with un-ignoring composer.lock is if someone runs composer update instead of composer install, it will likely break their updates.  I've added a bit of code to prevent accidental (or mis-guided) runs.

laf:
Running on my dev install to test. Will report back later on.

murrant:
Those all look fine to me.  As long as we use the composer-wrapper script it will re-download composer.

laf:
Fancy adding the composer wrapper call to daily then?

https://p.libren.ms/view/raw/abff667b

murrant:
It already calls composer wrapper @laf.

```
COMPOSER="php ${LIBRENMS_DIR}/scripts/composer_wrapper.php"
```

laf:
@murrant That just defines it.

I updated the patch: https://p.libren.ms/view/raw/f9218bb1

At present, `./daily.sh` doesn't install composer if it doesn't exist it only tests for the autoload file and installs the dependancies.

murrant:
The wrapper downloads composer anytime it is called, waits for it to finish, then executes the command.

laf:
> The wrapper downloads composer anytime it is called, waits for it to finish, then executes the command.

It doesn't in this case though:

```
-bash-4.2$ git checkout small-changes
Switched to branch 'small-changes'
Your branch is up to date with 'murrant/small-changes'.
-bash-4.2$ git pull
Already up to date.
-bash-4.2$ ls composer.phar
composer.phar
-bash-4.2$ rm composer.phar -f
-bash-4.2$ ./daily.sh
Checking PHP version                               OK
Updating to latest codebase                        OK
Updating SQL-Schema                                OK
Updating submodules                                OK
Cleaning up DB                                     OK
Fetching notifications                             OK
Caching PeeringDB data                             OK
-bash-4.2$ ls composer.phar
ls: cannot access composer.phar: No such file or directory
```

murrant:
@laf that's because composer wrapper wasn't called.

```
rm composer.phar -f
rm vendor/autoload.php
./daily.sh
```

murrant:
```
  - Updating phpmailer/phpmailer (v5.2.25 => v5.2.26): Downloading (connecting...)
Could not fetch https://api.github.com/repos/PHPMailer/PHPMailer/zipball/70362997bda4376378be7d92d81e2200550923f7, please create a GitHub OAuth token to go over the API rate limit
Head to https://github.com/settings/tokens/new?scopes=repo&description=Composer+on+rorix+2018-01-17+0831
to retrieve a token. It will be stored in "/opt/librenms/.composer/auth.json" for future use by Composer.
Token (hidden):
```

laf:
> @laf that's because composer wrapper wasn't called.

I'm aware of that but that's my issue and point here. I'm testing if composer doesn't exist, not that the vendor folder doesn't. This scenario is where someone installs everything fine but then deletes composer later on. At that stage we won't re-install it.

However that has just raised the question, the idea is to remove the vendor folder and the user will run composer install. However when we want them to update to a newer lib. Where in the code does that happen as I can't spot it, we don't ever re-run composer install - or is that part of the other PR?

murrant:
But, if you don't need composer, why does it need to exist?


That is absolutely not part of this PR.  I want to make that as a very small separate change.
Yes, in that change we will add: `composer-wrapper.ph --install` (which will update to the versions in composer.lock)

laf:
> But, if you don't need composer, why does it need to exist?

You've got it in validate so why not here?

> That is absolutely not part of this PR. I want to make that as a very small separate change.
> Yes, in that change we will add: composer-wrapper.ph --install (which will update to the versions in composer.lock)

Fine. Merge whenever you want.

PR #8086, opened 2018-01-15
Add Hardware, Version, and Serial strings to EqualLogic OS

peelman:
Uses snmpwalks because a given SNMP node can report for multiple members, which show up as unique indices.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/os/equallogic.inc.php
+++ b/includes/polling/os/equallogic.inc.php
@@ -1,0 +1,28 @@
+<?php
+/**
+ * equallogic.inc.php
+ *
+ * -Description-
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2018 Nick Peelman
+ * @author     Nick Peelman <nick@peelman.us>
+ */
+
+$hardware = implode(', ', explode(PHP_EOL, snmp_walk($device, 'eqlMemberProductFamily', '-Ovqa', 'EQLMEMBER-MIB')));
+$version = implode(', ', explode(PHP_EOL, snmp_walk($device, 'eqlMemberSerialNumber', '-Ovqa', 'EQLMEMBER-MIB')));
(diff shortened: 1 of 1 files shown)


peelman:
While you're looking at this:

Right now the Equallogic OS's definition includes only the Dell MIB directory; ignoring the EqualLogic MIB dir with all the relevant files in it.  Not sure if those were supposed to get moved, `equallogic` was left out by mistake.


```
diff --git a/includes/definitions/equallogic.yaml b/includes/definitions/equallogic.yaml
index 9777af9..862b98e 100644
--- a/includes/definitions/equallogic.yaml
+++ b/includes/definitions/equallogic.yaml
@@ -7,6 +7,7 @@ over:
     - { graph: device_storage, text: 'Storage Usage' }
 mib_dir:
     - dell
+    - equallogic
 discovery:
     - sysObjectId:
         - .1.3.6.1.4.1.12740.17.1
```

murrant:
@peelman that's incorrect.  Every OS automatically includes the directory of the same name as the OS.

peelman:
DOH!  of course you're right.  My apologies.

murrant:
Would you be willing to add test data for this?

laf:
I'm not sure that this is the right change.

Can you share the output of what this will look like?

Things like serial, hardware and version are designed to have a single entry so this might need to be done with get next to get the first value and the rest of the info adding to inventory.

peelman:
@laf i can and will in the future. but an Equallogic group can contain multiple arrays, so it’s entirely possible to receive multiple serials, models, etc from a polling run. insofar as i can tell, there isn’t a way to do this in a stricter way.

laf:
That really does sound like this is better handled in inventory rather than os polling then.

murrant:
Is this basically a chassis with "blades" in it?  I'm fine with that as each blade would have it's own model, serial, etc.

I did the same for E7 @laf...

peelman:
@murrant not quite; it can be one or more independent SANs acting together as a group; they allow things like seamless storage migrations between them and replication, among many other features.

We have three Eql SANs, but they are configured as individual groups.  The way I understand it based on the docs though, once you configure a "group" and add multiple SANs to them, they all will return the statuses of each other with SNMP; there isn't a way to query a single one any more (except for parsing out the individual id, obviously).  What you are querying when you query SNMP, even with just one SAN, is the Group's statuses, even if the group only contains one unit.

For example, you could have 2 Eql PS6100 and a PS6210 all belonging to the same group.  So when you query one, you are going to return data for all three, regardless of which one you query.  I hate it when they do things like this, but I can see advantages to the scaling of it, if the monitoring system is set up for it.

So I can see @laf's point that it should be an Inventory thing.  But I actually cloned what you had done for the Eltek Valere's, I had forgotten about the E7 configuration.  I'd like to get Inventory going for them to track controllers and disk independently, but that's going to have to wait for another slow day and those are pretty unpredictable for me.  And honestly we are trying to buy new SANs this year, so I may be writing modules for a StorageCenter or EMC setup by this Fall.

murrant:
Yes, I would use the inventory in that case.  Thanks.

laf:
> I did the same for calix @laf...

If I'd have spotted it then I'd of said the same thing. It's basically fudging data into columns not designed for that, especially from a webui perspective.

laf:
@peelman Closing this down. Feel free to re-open a new pull request or update this one making use of inventory

PR #8100, opened 2018-01-17
Move config loading into the Config class.

murrant:
Self contain dbFacile too.
Update rrdtool default setting to be similar to others.
Move silly cisco-entities.php into entity-physical.inc.php

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Testing on my dev install, will merge tomorrow if it works ok.

laf:
Tested full discovery, had polling running for > 6 hours. Web interface seems to work as expected so I'm happy with this.

PR #8101, opened 2018-01-17
changed check_mk service from xinetd to systemd

lwillek:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Proxmox.md
+++ b/doc/Extensions/Proxmox.md
@@ -13,16 +13,18 @@ In config.php, enable Proxmox:
 $config['enable_proxmox'] = 1;
 ```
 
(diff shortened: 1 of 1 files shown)


lwillek:
Please add Label "Documentation"

Proxmox changed from inetd using systemd now, so I adapted the documentation

laf:
Thanks for this.

Would you be able to put the systemd docs in as well as the xinetd so both exist and just indicate which version of proxmox uses which.

lwillek:
puh, systemd is out for a while now, but I understand the intention of your question. Let me rewrite the doc, ie. by adding a "xinetd section" for older proxmox systems, or for people using xinetd.

laf:
Thank you, it's to cover older installs as well so that would be awesome if you can.

laf:
image

PR #8102, opened 2018-01-17
Improved FortiManager support

tukezor:
Improved basic OS polling to include hardware, version, and serial number information
Added new mempool for accurate memory usage
Added Log Rate graph for FortiAnalyzer feature enabled FortiManagers
Updated FortiManager/FortiAnalyzer mib file

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/discovery/mempools/fortios.inc.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+if ($device['os'] == 'fortios') {
+    echo 'FORTIOS-MEMORY-POOL: ';
```
laf: You need to do a snmp_get here to check the device has mempools available before registering discover_mempool.

includes/polling/mempools/fortios.inc.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+// FortiOS Mempool
+echo 'FortiOS Mempool';
+
+$mempool['total'] = (snmp_get($device, 'FORTINET-FORTIMANAGER-FORTIANALYZER-MIB::fmSysMemCapacity.0', '-OvQ') * 1024);
```
laf: Please convert these two snmp_get calls to snmp_get_multi_oid(). You need to specify the MIB in the correct arg of that new function rather than tagging it onto the OID.

includes/polling/mempools/fortios.inc.php
```diff
@@ -0,0 +1,11 @@
...
+$mempool['used']  = (snmp_get($device, 'FORTINET-FORTIMANAGER-FORTIANALYZER-MIB::fmSysMemUsed.0', '-OvQ') * 1024);
+$mempool['perc'] = (($mempool['used'] / $mempool['total']) * 100);
+$mempool['free']  = ($mempool['total'] - $mempool['used']);
+
+echo '(U: '.$mempool['used'].' T: '.$mempool['total'].' F: '.$mempool['free'].') ';
```
laf: Either remove this line or change it to use `d_echo()` so it only appears in debug.

includes/polling/os/fortios.inc.php
```diff
@@ -1,3 +1,25 @@
 <?php
 
-$serial = snmp_get($device, 'fnSysSerial', '-OQv', 'FORTINET-FORTIANALYZER-MIB');
+use LibreNMS\RRD\RrdDefinition;
+
+$serial = snmp_get($device, 'FORTINET-CORE-MIB::fnSysSerial.0', '-Ovq');
```
laf: Please convert both of these to snmp_get_multi_oid() and also drop the MIB name from the oid and put it as an arg to the function.

includes/polling/os/fortios.inc.php
```diff
@@ -1,3 +1,25 @@
...
+$version = snmp_get($device, 'FORTINET-FORTIMANAGER-FORTIANALYZER-MIB::fmSysVersion.0', '-Ovq');
+$hardware = rewrite_fortinet_hardware($device['sysObjectID']);
+if ($hardware == $device['sysObjectID']) {
+    unset($hardware);
+}
+
+//Log rate only for FortiAnalyzer features enabled FortiManagers
+$entModes = snmpwalk_cache_oid($device, 'fmDeviceEntMode', null, 'FORTINET-FORTIMANAGER-FORTIANALYZER-MIB');
```
laf: How big is this walk? the idea in polling is to minimise the number of snmp calls. Can you not just test using `$log_rate = snmp_get($device, '.1.3.6.1.4.1.12356.103.2.1.9.0', '-Ovq');` seeing as that's all you care about?

sql-schema/230.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO `graph_types`(`graph_type`, `graph_subtype`, `graph_section`, `graph_descr`, `graph_order`) VALUES ('device',  'fortios_lograte',  'analyzer',  'Log Rate',  0);
```
laf: You will need to move this to `231.sql`

includes/polling/os/fortios.inc.php
tukezor: FortiManager and FortiAnalyzer are separate products, but you can enable most of the Analyzer features inside Manager. So the idea is to poll the log rate only if those Analyzer features have been enabled. There is no specific oid for that. 

That code checks if there are any Administrative Domains in fmg-faz mode. But you are right, in very large deployments that walk can get big. I think it should be safe to just get the first ADOM's mode and check if there is 'faz' in it.

tukezor:
That old Analyzer MIB was back from 2009. FortiManager and FortiAnalyzer do not have separate MIB's anymore, so there is just that combined MIB for both products. 

Thanks for the comments. I'll make the changes.

includes/polling/os/fortios.inc.php
laf: Or just do the log_rate snmp query. Running one snmp query to determine if we should run another is the same amount of time as just running the second one only and in fact will be slower if the first query tells us we should then run the second.

In short, you should only do what you're doing if we need the data from both queries which it doesn't look like we do here.

laf:
> That old Analyzer MIB was back from 2009. FortiManager and FortiAnalyzer do not have separate MIB's anymore, so there is just that combined MIB for both products.

Ok thanks, it can stay deleted then.

includes/polling/os/fortios.inc.php
tukezor: Log rate is only relevant if those Analyzer features are enabled. The log rate oid is present are those features enabled or not. So if we don't check that in any way, FortiManagers without analyzer features will get the log rate graph also. And its going to be constant 0. 

I understand that we want to keep snmp queries to the minimum. If you want it to be like you suggested, I will do it.

laf: Can the value of the log rate oid be 0 if it's enabled still? If not then you can check if it's > 0. If it can then you will need to find the best way to do a simple snmp get for the initial test, walking a tree which could be large shouldn't be done in polling.

laf:
Ok, that looks all good now.

The only thing left is unit test. https://docs.librenms.org/#Developing/os/Test-Units/ If you can add that then we can get this merged.

tukezor:
Added the test data. I Hope that's what you were asking for.

tests/snmpsim/fortios.snmprec
```diff
@@ -1 +1,41 @@
+1.3.6.1.2.1.1.1.0|4|
```
laf: This is perfect with the exception of sysDescr - is it genuinely blank?

tukezor: It's blank unless you manually set it.

laf: Ok, thanks :)

laf:
image

PR #8106, opened 2018-01-17
feature: migration of apps to collectd

crcro:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

just an ideea, with webui lineup and osupdates script test.

the test design was:
- host with collectd that reports data to server via network plugin (client)
- librenms host with collectd (server) and listening to collectd network clients and output to rrd in  /opt/librenms/collectd/

screen shot 2018-01-17 at 9 56 54 pm
screen shot 2018-01-17 at 9 57 04 pm

### DIFF
--- a/html/includes/collectd/functions.php
+++ b/html/includes/collectd/functions.php
@@ -66,8 +66,8 @@ function collectd_compare_host($a, $b)
 {
     $ea = explode('.', $a);
     $eb = explode('.', $b);
-    $i  = (count($ea) - 1);
-    $j  = (count($eb) - 1);
+    $i = (count($ea) - 1);
+    $j = (count($eb) - 1);
     while ($i >= 0 && $j >= 0) {
         if (($r = strcmp($ea[$i--], $eb[$j--])) != 0) {
             return $r;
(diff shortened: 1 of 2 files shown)


laf:
> librenms host with collectd (server) and listening to collectd network clients and output to rrd in /opt/librenms/collectd/

Can we not get it to store data in /opt/librenms/rrd/$hostname/ like normal?

murrant:
@laf CollectD puts the RRDs there, not LibreNMS...

@crcro I like the general direction, but it needs to be better developed.  I'm not particularly fond of the pretty_x() functions, they would be very long and a little nonsensical after adding all our apps.

laf:
> @laf CollectD puts the RRDs there, not LibreNMS...

I know, the question still stands though.

murrant:
The default according to the docs is 
```
$config['collectd_dir']  = '/var/lib/collectd/rrd';
```
https://docs.librenms.org/Support/Configuration/#collectd

crcro:
it could be done:
- changed the collectd_dir to /opt/librenms/rrd/ 
- collectd creates dirs under each host

the issue after this is that there are wrights issues and collectd cannot write data to rrd files.
still searching for a way to manage this ..

murrant:
Yeah, would require a re-write of the existing apps.  My experience is it will stay in the temporary state.  I actually don't care about apps that much, so I'm fine with that ;)

laf:
No further conversations have been had on this so I'm closing it down.

laf:
Proposing we close this down. Thanks for the work you put in @crcro but at present it's uncertain this is the way to go. @librenms/reviewers?

theherodied:
@laf, I use collectd for some things but I'm not positive it's the way to go either. I've looked into other projects like Prometheus to replace collectd.

laf:
Closing for now. Feel free to continue discussion.

PR #8111, opened 2018-01-18
device: Add Allied Telesis CPU info

mattie47:
Added support to poll both standalone devices and VCStacked device's CPU info.

-----

This works, however I don't like that on the device summary it shows the last polled CPU (e.g. CPU 8 in a stack of 8 switches), as usually the stack master will be stack member 1:

image

Can you see a way of modifying what I have (which I mostly copied from acos.inc.php) so Processor 1 will be what is shows on the device summary page?

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/processors/awplus.inc.php
+++ b/includes/discovery/processors/awplus.inc.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ * awplus.inc.php
+ *
+ * LibreNMS processor discovery module for Alliedware Plus
+ *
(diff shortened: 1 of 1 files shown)


laf:
> Can you see a way of modifying what I have (which I mostly copied from acos.inc.php) so Processor 1 will be what is shows on the device summary page?

That happens for all devices, it's not specific to this. It basically shows the last cpus description whilst aggregating all the data.

PR #8113, opened 2018-01-19
Add add_service_for_host endpoint to API

remyj38:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This pull request add the add_service_for_device API endpoint.

### DIFF
--- a/html/api_v0.php
+++ b/html/api_v0.php
@@ -186,6 +186,7 @@ $app->group(
                     '/services',
                     function () use ($app) {
                         $app->get('/:hostname', 'authToken', 'list_services')->name('get_service_for_host');
+                        $app->post('/:hostname', 'authToken', 'add_service_for_host')->name('add_service_for_host');
                     }
                 );
                 $app->get('/services', 'authToken', 'list_services')->name('list_services');
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -1516,13 +1516,13 @@ function validate_column_list($columns, $tableName)
 
 function add_service_for_host()
 {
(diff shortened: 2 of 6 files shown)


html/api_v0.php
```diff
@@ -186,6 +186,7 @@ function () use ($app) {
                     '/services',
                     function () use ($app) {
                         $app->get('/:hostname', 'authToken', 'list_services')->name('get_service_for_host');
+                        $app->post('/:hostname', 'authToken', 'add_service_for_host')->name('add_service_for_host');
```
laf: I'd change `add_service_for_host` to just `add_service`.

html/includes/api_functions.inc.php
```diff
@@ -1513,3 +1513,52 @@ function validate_column_list($columns, $tableName)
         $app->stop();
     }
 }
+
+function add_service_for_host()
+{
+    check_is_admin();
```
laf: We should also allow normal users to add these if they are allowed to access the device.

You can use `check_device_permission($device_id);` once you have the device id below.

html/api_v0.php
remyj38: It's already defined in https://github.com/librenms/librenms/blob/master/includes/services.inc.php#L37 and used in the function

html/includes/api_functions.inc.php
remyj38: Changed in 7344482

html/api_v0.php
laf: Good shout :)

laf:
Thanks very much for contributing this @remyj38 

image

PR #8123, opened 2018-01-22
device: Fix awplus $hardware for certain products

mattie47:
awplus.inc.php uses snmpgetnext for $hardware.

This works well for 99% of products however chassis
product SBx8100 does not show the chassis as the first product.

This adds an exception to follow for this product

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/os/awplus.inc.php
+++ b/includes/polling/os/awplus.inc.php
@@ -1,9 +1,15 @@
 <?php
 
 //$hardware and $serial use snmp_getnext as the OID for these is not always fixed.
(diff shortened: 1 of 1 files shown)


mattie47:
So this makes the following work for the 6 slot chassis and 12 slot chassis:

image



image

mattie47:
Why not use  sysObjectID.0 for all platforms?

Because BoardID has a slightly nicer formatted name across all platforms than what is returned by sysObjectID.

The SBx8100 is also the only awplus chassis product at present.

PR #8125, opened 2018-01-22
device: add suspended state to PAN OS

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

----------------------------

- Add "Suspended" state to Palo Alto Networks firewalls

### DIFF
--- a/misc/alert_rules.json
+++ b/misc/alert_rules.json
@@ -277,6 +277,10 @@
   "name": "Cisco NX-OS device has a bad fan"
   },
   {
+    "rule": "%devices.os = \"panos\" & %sensors.type = \"panSysHAState\" && %sensors.sensor_current = \"1\" && %sensors.sensor_prev = \"2\"",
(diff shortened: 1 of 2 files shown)


misc/alert_rules.json
```diff
@@ -276,10 +276,6 @@
   "rule": "%devices.os = \"nxos\" && %sensors.sensor_type = \"cefcFanTrayOperStatus\" && %sensors.sensor_current = \"[3-4]\"",
   "name": "Cisco NX-OS device has a bad fan"
   },
-  {
```
laf: Any reason this was removed, it should be valid still as it checks that the previous state was passive but is not active.

FTBZ: The rule was tested in deep yesterday and no way to match it and I didn't find a way to make it work. Don't understand why but since this is an important one I prefer to remove it to prevent users from using it if it's not working.

kkrumm1:
@FTBZ #8138

PR #8137, opened 2018-01-23
webui: added sysNames to pulldowns and the main page

aldemira:
Also made dev dependency list look more like devices page

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
So this needs a little update. The devices page does some checks on the hostname to decide what it should show. We should wrap the existing code from `html/includes/table/devices.inc.php` into a function and use it in both places:

```php
        if (format_hostname($device) !== $device['sysName']) {
            $hostname .= '<br />' . $device['sysName'];
        } elseif ($device['hostname'] !== $device['ip']) {
            $hostname .= '<br />' . $device['hostname'];
        }
```

aldemira:
@laf Can you give me a couple weeks to sort it out? I'm a bit too busy nowadays.

laf:
@aldemira Absolutely.

html/includes/functions.inc.php
```diff
@@ -1696,3 +1696,23 @@ function get_zfs_pools($device_id)
...
 }
+
+/**
+ * Returns the sysname of a device with a html line break prepended.
+ * if the device has an empty sysname it will return device's hostname instead
+ * And finally if the device has no hostname it will return an empty string
+ * @param device array
+ * @return string
+ */
+function add_sysname_or_hostname($device)
```
laf: Maybe `return_sysname_or_hostname`?

murrant: How about `get_device_name()`

aldemira:
Thanks guys, I wasn't keen on that function name as well :) 👍

html/includes/functions.inc.php
laf: I'm wondering if we pass a $html = false flag to it and it can then be re-used in other places.

aldemira: I removed the html bits altogether.

PR #8109, opened 2018-01-17
device: Allied Telesis inventory support

mattie47:
Ignore any commits below. I have not done anything here yet. 

As per screenshot from iReasoning mib browser, Allied Telesis awplus products do not [unfortunately] use the entity-mib, however most information is still available in the private enterprise mibs.

rscBoardTable (oid: .1.3.6.1.4.1.207.8.4.4.3.21.1.1) provides this in a table view.

image

My question is, how can lilbrenms import SNMP table information, since it's slightly different to doing an snmpget on individual OIDs?

Any assistance with what might help here would be much appreciated. 

Thanks,

Matt
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

PR #8139, opened 2018-01-23
device: Improve Allied Telesis Chassis product support

mattie47:
Fix - Wrong Serial number was displayed
Enhancement - Show the type of Controller card in chassis

Controller card can be either CFC400, or CFC960.

Tests:
- Checked Chassis x8106, and x8112
- Checked with CFC400 and CFC960
-- Checked behaviour with/without CFC in bay 1.5

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


mattie47:
@murrant 

As per highlighted: 

image

Is this the best way to specify a controller card in a chassis ? 

As you can see, I used $feature, but I'm not sure if something better than this exists for specifying a chassis' controller card. 

Thanks,

Matt

mattie47:
It does however look good when you see the list view:

image

includes/polling/os/awplus.inc.php
```diff
@@ -9,7 +9,16 @@
 $serial = snmp_getnext($device, "rscBoardSerialNumber", "-OQv", "AT-RESOURCE-MIB");
 
 // SBx8100 platform has line cards show up first in "rscBoardName" above.
+//Instead use sysObjectID.0
 if (strpos($hardware, 'SBx81') !== false) {
     $hardware = snmp_get($device, "sysObjectID.0", "-OQvs", "SNMPv2-MIB:AT-PRODUCT-MIB");
-    $hardware = str_replace('at', '', $hardware);
+    $hardware = str_replace('at', 'AT-', $hardware);
```
laf: Is this not risking matching at anywhere in the hardware string? Would `'/^at/'` not be better?

includes/polling/os/awplus.inc.php
```diff
@@ -9,7 +9,16 @@
...
+// Features and Serial is set to Controller card  1.5
+    $features = snmp_get($device, "rscBoardName.5.6", "-OQv", "AT-RESOURCE-MIB");
```
laf: No way to check sysObjectID or something we already know to determine the rcsBoardNames?

If not, you'll need to convert the two snmp_gets into snmp_get_multi_oid.

includes/polling/os/awplus.inc.php
mattie47: Yeah good point.

includes/polling/os/awplus.inc.php
mattie47: The problem is the device names suck in AT-PRODUCT-MIB. 

Unless I'm doing something stupidly wrong, this is what returns when polling several different products:

```librenms@librenmsdev:~$  snmpget -v2c -c librenms -OvQs -M /opt/librenms/mibs/:/opt/librenms/mibs/awplus 10.37.105.1 SNMPv2-MIB::sysObjectID.0
MIB search path: /opt/librenms/mibs/:/opt/librenms/mibs/awplus
atsbx908g2 ----> Should be AT-SBx908-GEN2
librenms@librenmsdev:~$  snmpget -v2c -c librenms -OvQs -M /opt/librenms/mibs/:/opt/librenms/mibs/awplus 10.37.105.4 SNMPv2-MIB::sysObjectID.0
MIB search path: /opt/librenms/mibs/:/opt/librenms/mibs/awplus
atx93028GTX ----> Should be AT-x930-28GTX
librenms@librenmsdev:~$  snmpget -v2c -c librenms -OvQs -M /opt/librenms/mibs/:/opt/librenms/mibs/awplus 10.37.105.5 SNMPv2-MIB::sysObjectID.0
MIB search path: /opt/librenms/mibs/:/opt/librenms/mibs/awplus
atXS916MXT ----> Should be AT-XS916-MXT
librenms@librenmsdev:~$  snmpget -v2c -c librenms -OvQs -M /opt/librenms/mibs/:/opt/librenms/mibs/awplus 10.37.105.6 SNMPv2-MIB::sysObjectID.0
MIB search path: /opt/librenms/mibs/:/opt/librenms/mibs/awplus
Cannot find module (FCMGMT-MIB): At line 28 in /opt/librenms/mibs//SW-MIB
Did not find 'connUnitPortStatEntry' in module #-1 (/opt/librenms/mibs//SW-MIB)
atx510L28GT ----> Should be AT-x510-28GT
librenms@librenmsdev:~$  snmpget -v2c -c librenms -OvQs -M /opt/librenms/mibs/:/opt/librenms/mibs/awplus 10.37.105.7 SNMPv2-MIB::sysObjectID.0
MIB search path: /opt/librenms/mibs/:/opt/librenms/mibs/awplus
atGS948MPX ----> Should be AT-GS948-MPX
```

rscBoardName simply provides the more correct device name. It just so happens the SBx8112 and SBx8106 don't have any "dashes" then product type after the name so this worked OK for using AT-PRODUCT-MIB

mattie47: >If not, you'll need to convert the two snmp_gets into snmp_get_multi_oid.

 And I should know this by now ;-)

includes/polling/os/awplus.inc.php
mattie47: I've spent a while on this and can't get something that will match.

'/^at/' doesn't seem to work for atSBx8106 or atSBx8112.

Unless you can give another suggestion, I really don't know what to try :-(

These are the only two options $hardware can be, and given that we only go down this path if SBx81 is true I guess this shouldn't be too concerning.

murrant:
Can we combine these to a single snmp_getnext_multi() call?
You could probably call `snmp_translate($device['sysObjectID'], 'AT-PRODUCT-MIB')` instead of re-fetching syObjectID.
Why is $hostname being set in this file?

murrant:
#8156

mattie47:
@murrant  - Thanks.

I've just tried using
    $hardware = snmp_translate($device['sysObjectID'], 'AT-PRODUCT-MIB');

```
SNMP[/usr/bin/snmptranslate -M /opt/librenms/mibs:/opt/librenms/mibs/ -m AT-PRODUCT-MIB -Os .1.3.6.1.4.1.207.1.14.114 2>/dev/null]
iso.*.*.114
```

As per above, it's not appending the OS location to -M. E.g. should be:

```
SNMP[/usr/bin/snmptranslate -M /opt/librenms/mibs:/opt/librenms/mibs/awplus/ -m AT-PRODUCT-MIB -Os .1.3.6.1.4.1.207.1.14.114 2>/dev/null]
iso.*.*.114
```

I assume it should be? 

Thanks,

Matt

mattie47:
Hi @murrant,

As per aa7590e I've updated my file however I still can't get the snmptranslate functionality working.


Please see https://p.libren.ms/view/4f7b6425 for what's happening.

Thanks,

Matt

murrant:
@mattie47 You will need #8156 for a properly working snmp_translate() function.

Could you give that a try?

murrant:
@mattie47 snmp_translate isn't working because you need to provide a device or a mibdir. Try:
```
$hardware = snmp_translate($device['sysObjectID'], 'AT-PRODUCT-MIB', null, null, $device);
```

mattie47:
@murrant  Thanks again. Tested both and that fixed things :-)

mattie47:
Is there anything stopping this from being merged? 

Thanks,

Matt

murrant:
Let me check the testing was ran against the new code.  Then we can merge it.

murrant:
@mattie47 actually, I just looked again and I have a couple of changes to suggest.

Since you are now fetching multiple points of data from rscBoardName and rscBoardSerialNumber, a walk would be better and faster.

Also, can you add some test data for this?

mattie47:
Hi @murrant,

OK no worries.

Is something like this acceptable? Sorry if it's crap. I'm still very much learning PHP...

```php
$data_array = snmpwalk_cache_multi_oid($device, 'rscBoardName', $data_array, 'AT-RESOURCE-MIB', '-OUsb');
$data_array = snmpwalk_cache_multi_oid($device, 'rscBoardSerialNumber', $data_array, 'AT-RESOURCE-MIB', '-OUsb');

d_echo($data_array);

if (strpos($hardware, 'SBx81') !== false) {
    $hardware = snmp_translate($device['sysObjectID'], 'AT-PRODUCT-MIB', null, null, $device);
    $hardware = str_replace('at', 'AT-', $hardware);
// Features and Serial is set to Controller card 1.5
    $features = $data_array['5.6']['rscBoardName'];
    $serial = $data_array['5.6']['rscBoardSerialNumber'];
```

It does work, but just not sure if it's the best way...

Thanks,

Matt

laf:
@mattie47 No point in doing the walks if you don't use them, so add them after the `if (strpos($hardware, 'SBx81') !== false) {`

mattie47:
@murrant  @laf  OK this is finally done.  Test data also updated.  I also made it clear that the serial number shown if device is a chassis is the serial number of the controller card. 

Thanks,

Matt

includes/polling/os/awplus.inc.php
```diff
@@ -3,13 +3,30 @@
...
+    $data_array = snmpwalk_cache_multi_oid($device, 'rscBoardName', $data_array, 'AT-RESOURCE-MIB', '-OUsb');
+    $data_array = snmpwalk_cache_multi_oid($device, 'rscBoardSerialNumber', $data_array, 'AT-RESOURCE-MIB', '-OUsb');
+
+    $hardware = snmp_translate($device['sysObjectID'], 'AT-PRODUCT-MIB', null, null, $device);
+    $hardware = str_replace('at', 'AT-', $hardware);
+
+// Features and Serial is set to Controller card 1.5
+    $features = $data_array['5.6']['rscBoardName'];
+    $serial = $data_array['5.6']['rscBoardSerialNumber'];
+    $serial = $serial. ' - CFC Bay 1.5';
```
laf: Last one for me on this, why append something to the end of the serial? The serial field should solely contain the actual serial number.

mattie47: @laf 

No worries. I tried to answer this in my last comment, but to provide some more context.

Essentially, we could have up to 24 serial numbers for 24 line cards, with 4 cards being controller cards in a 2x 12 bay Chassis stack.

Then there's also the serial number for the chassis shell, and serials for the PSUs etc. 

So as a user, I thought it was important to clarify *what* serial number we are actually reporting in LibreNMS. 

I mean, as a user which serial should I also care about? In almost all cases - The Controller Card. This is because all licensing is tied to it. 
 
So as a result, the serial number we most likely care about though is going to be either CFC Bay 1.5, or if not present then CFC Bay 1.6:

```
! Starting: Show system cmd 
! --------------------------------------------------------------------------------! 
! Stack member 1
! Board      ID   Bay     Board Name                        Rev   Serial number
! --------------------------------------------------------------------------------
! Chassis    374  Chassis AT-SBx8106                        B-0   R265E5008       
! Blade      353  Bay1    AT-SBx81XS6                       X8-0  A045624113500003
! Blade      317  Bay3    AT-SBx81GP24                      A-0   A04200a112800003
! Blade      352  Bay4    AT-SBx81GS24a                     C-1   A1XVD200E       
! Controller 376  Bay5    AT-SBx81CFC960                    B-0   A270E400A       
! Controller 376  Bay6    AT-SBx81CFC960                    B-0   A270E400U       
! PSU        319  PSUD    AT-SBxPWRSYS1/AC                  A-0   A045804135000064
! Fan module 375  FAN1    AT-SBxFAN06                       H-0   A04561A141900013
! --------------------------------------------------------------------------------
! --------------------------------------------------------------------------------
! Environment Status : Normal
! Bootloader version : 3.0.5
! Stack member 2
! Board      ID   Bay     Board Name                        Rev   Serial number
! --------------------------------------------------------------------------------
! Chassis    374  Chassis AT-SBx8106                        B-0   R265E3Z0L       
! Blade      381  Bay2    AT-SBx81XS16                      A-0   A04775H131100011
! Blade      380  Bay3    AT-SBx81GT40                      A-1   A26LD700Q       
! Controller 376  Bay5    AT-SBx81CFC960                    B-0   A270E400G       
! Controller 376  Bay6    AT-SBx81CFC960                    C-0   A270F602U       
! PSU        319  PSUD    AT-SBxPWRSYS1/AC                  A-0   A045804134300036
! Fan module 375  FAN1    AT-SBxFAN06                       H-0   A04561A141900010
! --------------------------------------------------------------------------------
! --------------------------------------------------------------------------------
! Environment Status : Normal
! Bootloader version : 3.0.2
! Current boot config: flash:/multi.cfg (file exists)
! System Name
```


Given the controller card can be in either bay 5 and/or 6, I felt it was worthwhile noting which controller card's serial number we're actually displaying in LibreNMS. It also makes it clear it's not the chassis shell's serial. 

It's also worth noting I've only made this code run for this 1 product. 

Thanks,

Matt

mattie47: Also, if there's major concerns I'm adding this to the field, perhaps it's something I could look at removing when I finally get to https://github.com/librenms/librenms/pull/8109.

Since there's no entity mib support at the moment, which serial number we're showing for said chassis product just isn't clear.  However, if the inventory support is in LibreNMS later, then the user of course can look at that tab to clarify further what serial number is in fact being displayed.

Since this isn't an option at the moment, I feel it's worth leaving the change as I currently have it. 

Thoughts?

Thanks,

Matt

laf: Ok based on that I think you should remove the text. Their should be no reason to explain what serial number we display just that we pick the one that makes the most sense.

For example on the Cisco ASAs we use the Chassis serial number and I'm sure we do the same elsewhere so it makes more sense to use the chassis serial. The rest of them belong in inventory.

mattie47: >Ok based on that I think you should remove the text. Their should be no reason to explain what serial number we display just that we pick the one that makes the most sense.

OK no worries. Will change in next commit.  

> I'm sure we do the same elsewhere so it makes more sense to use the chassis serial.

Allied Telesis literally ties nothing licensing wise to the Chassis's serial. It's all tied to the controller card[s] in the chassis. Is that the same for Cisco ASAs? I'm assuming ASAs bind it to the Chassis...

I can make the serial the chassis' if you want...I guess I'm reluctant because I can't see any use in knowing the Chassis' serial number other than perhaps Inventory management of physical devices in a network. 

Thanks,

Matt

laf: Use the one that makes the most sense :)

mattie47: Done :-) Thanks.

laf:
Thanks as always @mattie47 :)

PR #8149, opened 2018-01-25
Allow administrators to edit devices sysName

remyj38:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

-----
- New feature to allow administrators to edit the device sysName from edit page.
- Fixing indentation of html/pages/device/edit/device.inc.php

### DIFF
--- a/html/pages/addhost.inc.php
+++ b/html/pages/addhost.inc.php
@@ -49,6 +49,7 @@ if ($hostname !== false) {
                 'snmp_disable' => 1,
                 'os'           => $_POST['os'] ? mres($_POST['os_id']) : "ping",
                 'hardware'     => mres($_POST['hardware']),
+                'sysName'      => mres($_POST['sysName'])
             );
         } elseif ($_POST['snmpver'] === 'v2c' || $_POST['snmpver'] === 'v1') {
             if ($_POST['community']) {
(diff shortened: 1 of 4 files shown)


laf:
This will just revert back to what's found in polling / discovery.

Why allow this to be changed out of interest?

remyj38:
In my usage of LibreNMS, I have a lot of hosts which don't have SNMP (so don't have sysName).
For this kind of hosts, it's easier to find it if a sysName is set.

laf:
> In my usage of LibreNMS, I have a lot of hosts which don't have SNMP (so don't have sysName).
> For this kind of hosts, it's easier to find it if a sysName is set.

Ok but this change allows it for anyone. For snmp based devices the value will just be overwritten. You need to look at `html/pages/device/edit/snmp.inc.php`, that file allows certain input boxes to be shown when snmp is disabled only which is what you want.

You also shouldn't have the edit pencil button next to the input, just allow it be changed if it's non-snmp based.

Last thing, admin users should be allowed to change this like you have, but also normal users with permission to edit the device. You can use `device_permitted()` to determine this.

remyj38:
c4700d3d1c303d53e7cd4db44ca68dff48f13a90 follows your recommendations

murrant:
Wouldn't mind allowing this for all devices, to allow users to set custom display names on anything.

This is a good start, thanks.

PR #8161, opened 2018-01-29
F5 gtm support

dnapper:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Adds discovery and polling of F5 GTM module Wide IP and Pools discovery.


laf:
Thanks for this. Before we review the code can you share some screenshots of what the various new pages look like please.

Could you also show the impact on poll time for your devices and indicate roughly how big they are? The f5 module already can have a big impact on users so we need to understand the risk to existing installs here.

dnapper:
Sure. 

In our environment we don't run GTM services on the same devices as we run LTM so for me it will have no impact on poll times on existing devices. 

The code checks to see if the GTM module is running prior to discovering any new entities so I don't expect this will impact any devices that have the module disabled. 

This adds two new tabs to the load balancer page if GTM objects are discovered. 
gtm-newtabs

The pages look virtually the same as the current VS and Pool pages but the graphs are based on DNS requests handled not data throughput. 
gtm-wideipdetail
gtm-wideipdetail

I'll leave these pre devices with this running for a few days so you can see what the data looks like (and that it's actually polling :) )

One thing i've been struggling with is how to handle representing a service that has multiple pools associated.  In the end I decided that if I can sort out a way I can always add that later. 

This is with a pre-prod devices with ~30 objects on it. 

While on the main branch the discovery time and polling time is ~7s. 
Discovered in 7.116 seconds
SNMP [45/5.64s]: Get[10/0.77s] Getnext[1/0.07s] Walk[34/4.80s]
MySQL [158/0.06s]: Cell[30/0.00s] Row[25/0.00s] Rows[70/0.01s] Column[1/0.00s] Update[26/0.03s] Insert[5/0.01s] Delete[1/0.00s]
RRD [0/0.00s]: Update[0/0.00s] Create [0/0.00s] Other[0/0.00s]


Polled in 7.028 seconds
UPDATED!
#### Start Alerts ####
#### End Alerts ####
SNMP [57/5.61s]: Get[19/1.41s] Getnext[0/0.00s] Walk[38/4.20s]
MySQL [63/0.07s]: Cell[3/0.00s] Row[2/0.00s] Rows[20/0.00s] Column[3/0.00s] Update[33/0.06s] Insert[2/0.00s] Delete[0/0.00s]
RRD [75/0.00s]: Update[75/0.00s] Create [0/0.00s] Other[0/0.00s]

With the new additions discovery takes ~10s
Discovered in 10.139 seconds

SNMP [49/8.12s]: Get[11/0.87s] Getnext[1/0.07s] Walk[37/7.17s]
MySQL [704/0.60s]: Cell[30/0.01s] Row[26/0.00s] Rows[73/0.01s] Column[1/0.00s] Update[86/0.09s] Insert[487/0.48s] Delete[1/0.00s]
RRD [0/0.00s]: Update[0/0.00s] Create [0/0.00s] Other[0/0.00s]

Polling takes ~8s

Polled in 8.377 seconds
UPDATED!
#### Start Alerts ####
#### End Alerts ####
SNMP [64/6.76s]: Get[19/1.40s] Getnext[0/0.00s] Walk[45/5.36s]
MySQL [65/0.07s]: Cell[3/0.00s] Row[2/0.00s] Rows[22/0.01s] Column[3/0.00s] Update[33/0.06s] Insert[2/0.00s] Delete[0/0.00s]
RRD [195/0.01s]: Update[135/0.00s] Create [60/0.00s] Other[0/0.00s]

dnapper:
Update:
I found an issue where once I added an LTM to my test server the LTM discovery process was deleting the GTM objects. In order to fix it I split the GTM polling into it's own load balancer module and it appears to be working.

I'll leave it for a while to see if the problem comes back.

As a side note here is a comparison of LTM poling before and after the patch.
The polling time is quite a bit faster on my dev server but that's probably due to load.

DEV
polling-dev

PRE
polling-pre

html/pages/device/loadbalancer/gtm_pool.inc.php
```diff
@@ -0,0 +1,39 @@
...
+foreach ($components as $k => $v) {
+    foreach ($types as $type) {
+        if ($v['type'] == $type) {
+            $keep[$k] = $v;
+        }
+    }
+}
+$components = $keep;
+
+global $config;
```
laf: You shouldn't need this.

html/pages/device/loadbalancer/gtm_pool.inc.php
```diff
@@ -0,0 +1,39 @@
...
+
+if (is_file('pages/device/loadbalancer/'.mres($vars['subtype']).'.inc.php')) {
```
laf: You can drop the mres from these two lines, `$vars` has already had that applied.

html/pages/device/loadbalancer/gtm_wide.inc.php
```diff
@@ -0,0 +1,39 @@
...
+        if ($v['type'] == $type) {
+            $keep[$k] = $v;
+        }
+    }
+}
+$components = $keep;
+
+global $config;
+
+if (is_file('pages/device/loadbalancer/'.mres($vars['subtype']).'.inc.php')) {
```
laf: You don't need to use mres on these two lines.

includes/discovery/loadbalancers/f5-gtm.inc.php
```diff
@@ -0,0 +1,187 @@
...
+    $gtmWideStatusEntry = snmpwalk_array_num($device, '1.3.6.1.4.1.3375.2.3.12.3.2.1', 0);
+    $gtmPoolEntry = snmpwalk_array_num($device, '1.3.6.1.4.1.3375.2.3.6.2.3.1.1', 0);
+}
+
+/*
+ * False == no object found - this is not an error, OID doesn't exist.
+ * null  == timeout or something else that caused an error, OID may exist but we couldn't get it.
+ */
+
+if (!is_null($gtmWideIPEntry) || !is_null($gtmWideStatusEntry) || !is_null($gtmPoolEntry)) {
```
laf: Don't these need to be && if you want them all to match?

includes/discovery/loadbalancers/f5-gtm.inc.php
```diff
@@ -0,0 +1,187 @@
...
+    }
+}
+$components = $keep;
+
+// Begin our master array, all other values will be processed into this array.
+$tblBigIP = array();
+
+//Check if GTM modeul is running and run object discovery.
+if ((snmp_get($device, 'sysModuleAllocationProvisionLevel.3.103.116.109', '-Ovqs', 'F5-BIGIP-SYSTEM-MIB')) !=none) {
+    $gtmWideIPEntry = snmpwalk_array_num($device, '1.3.6.1.4.1.3375.2.3.12.1.2.1', 0);
```
laf: If we need all of these walks then it would be worth checking if each contains the data before trying the next and only carry on if we have them all (moving the check from below to here).

includes/discovery/loadbalancers/f5-ltm.inc.php
```diff
@@ -51,6 +51,13 @@
 $ltmPoolMemberEntry = snmpwalk_array_num($device, '1.3.6.1.4.1.3375.2.2.5.3.2.1', 0);
 $ltmPoolMbrStatusEntry = snmpwalk_array_num($device, '1.3.6.1.4.1.3375.2.2.5.6.2.1', 0);
 
+//Check if GTM modeul is running and run object discovery.
```
laf: Is this supposed to be here?

html/pages/device/loadbalancer/gtm_pool.inc.php
dnapper: Do you want me to remove it from the ltm module pages as well?

includes/discovery/loadbalancers/f5-gtm.inc.php
dnapper: You don't need to have a pool attached to a wide IP for it to function so there could be situations where people are using the device to service static DNS records as well as dynamic ones based on pool conditions.

html/pages/device/loadbalancer/gtm_pool.inc.php
laf: No, leave that. You can remove them in another PR but let's not mix things up here.

includes/discovery/loadbalancers/f5-ltm.inc.php
dnapper: no... whoops.

includes/discovery/loadbalancers/f5-gtm.inc.php
dnapper: Sure, there is no point moving onto look for status and pools if there are no wide IPs. 
Do you want me to do the same in the LTM discovery? We could skip trying to get pools and status codes if there are no virtual servers objects.

laf: Any saving is good, but again I'd keep it to another PR for LTM.

dnapper: Can do.

murrant:
Looks good, thanks @dnapper :)

PR #8162, opened 2018-01-29
Poll sensors as enum values instead of labels.

murrant:
Most of the time they should be polled as values anyway, but sometimes the mib might get loaded...  If we set -Oe, then we always get the numeric value of the poller.
All sensors should get a numeric value anyway.

This allows people to get away with mis-matched sensor names...

We still have to handle state sensors that return string values though.

Thoughts?

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Looks fine to me, can't think of any issues that might bite us on this one.

PR #8178, opened 2018-01-31
feature: Copy all email alerts to default contact

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---------------------------------------------------

- Add a way to copy all email alerts to default contact without ignoring other methods.

screen shot 2018-01-31 at 15 09 41

_Thanks @laf for the help._


includes/alerts.inc.php
```diff
@@ -317,6 +317,11 @@ function GetContacts($results)
         }
     }
 
+    # Copy all email alerts to default contact if configured.
+    if ($config['alert']['default_copy'] === true) {
+        $tmp_contacts[$config['alert']['default_mail']] = '';
```
laf: Before assigning it to $tmp_contacts, you need to check it doesn't exist already :)

sql-schema/233.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO `config` VALUES (798,'alert.default_copy','true','true','Copy all email alerts to default contact','alerting',0,'general',0,'0','0');
```
laf: This is breaking travis, you shouldn't insert the entry with a config_id set as they are auto incrementing.

See sql-schema/221.sql on how to do it.

includes/alerts.inc.php
```diff
@@ -317,6 +317,11 @@ function GetContacts($results)
         }
     }
 
+    # Copy all email alerts to default contact if configured.
+    if (!isset($tmp_contacts[$config['alert']['default_mail']]) && ($config['alert']['default_copy'])) {
+	        $tmp_contacts[$config['alert']['default_mail']] = '';
```
laf: This has a tab in from what I can see, needs replacing for 4 spaces :)

FTBZ: Oops, but I can't commit anymore: "error: failed to push some refs"

laf: You might need to git pull first.

PR #8182, opened 2018-02-01
docfix: Changed err-msg to message

iTeV:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
Changed the key `err-msg` to `message` in the API docs since the API doesnt return `err-msg` anymore, instead it returns `message` 
[source](https://github.com/librenms/librenms/blob/master/html/includes/api_functions.inc.php#L83)


laf:
Thanks very much for these fixes @iTeV :)

iTeV:
anytime!

PR #8193, opened 2018-02-03
Add Linux Mint OS icon svg

jquagga:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I added SNMP monitoring to a Linux Mint kiosk and wanted to swap out the tux icon for the Linux Mint icon.  Adding this file appears to have done that.  

This is the [Linux Mint svg from Wikipedia Commons](https://commons.wikimedia.org/wiki/File:Linux_Mint_logo_without_wordmark.svg) run through the online svg optimizer as suggested in the documentation.


laf:
You'll need to strip out the width / height from this. Should be able to do this with: https://jakearchibald.github.io/svgomg/

jquagga:
I can't get it to update in a way which doesn't include the height and width attribute and which looks right.  It's probably beyond my svg ability.  Hand editing it blew it away and the optimizer seems to keep the height / width.

PR #8197, opened 2018-02-05
Tweak multi_bits_separated graph rendering

peelman:
* remove extra backslashes that were getting rendered
* Aline aggregate in/out text with interface in/out text
* Adjust description length for graphs with widths > 500

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/graphs/generic_multi_bits_separated.inc.php
+++ b/html/includes/graphs/generic_multi_bits_separated.inc.php
@@ -132,7 +132,7 @@ if (!$noagg) {
     $rrd_options .= ' VDEF:totalin=aggrinbytes,TOTAL';
     $rrd_options .= ' VDEF:totalout=aggroutbytes,TOTAL';
     $rrd_options .= " COMMENT:' \\n'";
-    $rrd_options .= " COMMENT:'" . substr(str_pad('Aggregate', ($descr_len + 5)), 0, ($descr_len + 5)) . 'In' . "'";
+    $rrd_options .= " COMMENT:'" . substr(str_pad('Aggregate', ($descr_len + 4)), 0, ($descr_len + 4)) . ' In' . "'";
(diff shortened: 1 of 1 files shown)


laf:
Before:

image


After:

image

imho this adds far too much whitespace after the interface description.

peelman:
graph php

That's fine until you cross devices.  Then it tries to put the hostname in that field with the interface on teh second line.  That can make it problematic to read.  It should be dynamic or wide enough to be halfway accommodating.  18 Characters isn't wide enough.  40 Might be too much. I'd be willing to compromise there.

murrant:
Hmm, perhaps we pre-measure the description length, then we don't have extra padding when we don't need it.

peelman:
i was going to take a swing at looping through for the longest description (or 40chars) and set it; but i kept getting too distracted yesterday. i can try to find some time later this week maybe, unless @murrant beats me to it.

laf:
Ok what about this. @murrant will probably rewrite it but it seems to be ok :)

```
php > $init_modules = ['web'];
php > require 'includes/init.php';
php > var_dump(set_length(['Agg Indsadsdfdf something more made up but too long', 'Agg Out'], 40));
array(2) {
  [0]=>
  string(40) "Agg Indsadsdfdf something more made up b"
  [1]=>
  string(40) "Agg Out                                 "
}

php > var_dump(set_length(['Agg In', 'Agg Out'], 40));
array(2) {
  [0]=>
  string(7) "Agg In "
  [1]=>
  string(7) "Agg Out"
}
```
Code:
https://p.libren.ms/view/raw/9d5abd26

laf:
We can set a few more spaces in as a default so the limit will be $len + 2 for example.

laf:
bump

laf:
Closing for now, feel free to respond to my previous comments.

PR #8200, opened 2018-02-05
modified styles.css to make the device group menu a bit larger

chipgwyn:
The "Device Groups" menu seems unreasonably small.  Changed the css from max-height 200px to 800px.  This should allow for a much larger menu without scrolling off a reasonably sized screen.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


chipgwyn:
i've been using this change for a couple of months, haven't experienced any odd side-effects.

laf:
Can't help but think 800 is a bit too tall. My browser window is currently is about 900 tall and that's full screen.

You will also need to edit html/index.php and change the random number on this line:

`<link href="<?php echo($config['stylesheet']);  ?>?ver=291727421" rel="stylesheet" type="text/css" />`

chipgwyn:
Yeah I could see 800 maybe being a bit much.  I wish there was a way to make it go to maybe a max of 800, otherwise scroll when there's more entries than what the screen will take.  Alas I'm useless when it comes to CSS.  I tried several different combinations but couldn't work it out.

murrant:
Just do 600 I think.

laf:
@chipgwyn Any update on this one?

laf:
I've forced a css update with this so we can merge when it's completed it's checks.

PR #8207, opened 2018-02-06
Added new worldmap widget setting:

remyj38:
Signed-off-by: Rémy Jacquin <remy@remyj.fr>

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

----
Added new worldmap widget setting:
The user can chose how down devices under maintenance are shown: as a down device (default and as currently), as a warning device (orange icon) or as a up device. Up devices and warning devices are not shown if the map is configured to show only down devices


kkrumm1:
@remyj38 can you add a little tidbit about this to the world map doc, please. 
https://github.com/librenms/librenms//edit/master/doc/Extensions/World-Map.md

remyj38:
@kkrumm1 Done 😄

html/includes/common/worldmap.inc.php
```diff
@@ -93,6 +95,30 @@
             </select>
         </div>
     </div>
+    <div class="form-group">
+        <div class="col-sm-4">
+            <label for="maintenance" class="control-label">Down devices over maintenance should be shown as: </label>
```
laf: Maybe `Down devices in maintenance windows shown as:`

murrant:
Actually, why not show devices under maintenance as a totally different color?  We wouldn't need an option then.

Perhaps light blue.  Would also need a tooltip to note the status.

kkrumm1:
I agree with murrant because in the Availability map Device is shown as Yellow as warning for restarted in last 24 hours. This could be confusing with the world map having a color of orange. Users may think that is because of the warning (restarted).

html/includes/common/worldmap.inc.php
remyj38: @laf: I removed option in 288edca

remyj38:
@murrant @kkrumm1 : I removed the option and changed the color to light blue.

laf:
Thanks very much for this @remyj38.

PR #8211, opened 2018-02-06
Basic support for OpenBSD PF states

dagmartin:
This adds graphs for PF firewall state table on OpenBSD (might apply to FreeBSD as well, have not tested)

DO NOT DELETE THIS TEXT

#### Please note

- [ X ] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers


murrant:
Perhaps you should drop openbsd_ from the graph names?

dagmartin:
I wasn't really sure about the naming convention, most other related graphs I looked at seemed to have the vendor name in the filename (panos, fortios, etc.) and just "pf" might seem a bit generic? Since I've only tested this against OpenBSD I thought it made sense to name the files like that, but I'm happy to change if that is desired.

murrant:
@dagmartin even if the data collection is different, won't the data and graphs be the same?

dagmartin:
That's true. I've removed the openbsd prefix from the files.

PR #8220, opened 2018-02-07
Update librenms.logrotate

dragans:
- Change path /opt/librenms/logs/*log so that logrotate catches http logs (access_log and error_log).
- Added SU parameter to rotate files with librenms user, as it was throwing insecure permissions error without it.
- Added CREATE parameter to create new log files as librenms user.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/misc/librenms.logrotate
+++ b/misc/librenms.logrotate
@@ -7,5 +7,5 @@
     delaycompress
     missingok
     notifempty
-    create 644 librenms librenms
+    create 664 librenms librenms
 }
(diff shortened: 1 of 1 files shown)


misc/librenms.logrotate
```diff
@@ -1,9 +1,11 @@
...
-/opt/librenms/logs/*.log {
+/opt/librenms/logs/*log {
+    su librenms librenms
     weekly
     rotate 6
     compress
     delaycompress
     missingok
     notifempty
+    create 644 librenms librenms
```
murrant: You should allow the librenms group to write.  So: 664

dragans: Agreed :)

murrant:
Thanks :)

PR #8222, opened 2018-02-07
Filter disabled users, allow nested group membership for AD auth

network-guy:
Active Directory auth. Filter disabled users and allow nested groups.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Authentication/ADAuthorizationAuthorizer.php
+++ b/LibreNMS/Authentication/ADAuthorizationAuthorizer.php
@@ -180,7 +180,7 @@ class ADAuthorizationAuthorizer extends MysqlAuthorizer
         $ldap_groups = $this->getGroupList();
 
         foreach ($ldap_groups as $ldap_group) {
-            $search_filter = "(memberOf=$ldap_group)";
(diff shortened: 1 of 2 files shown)


murrant:
Should we be using: get_auth_ad_user_filter() and get_auth_ad_group_filter() functions there?

network-guy:
I don't believe so. get_auth_ad_group_filter() is already employed to get the ldap groups used in this filter. It doesn't pull in user membership information. get_auth_ad_user_filter() queries for a specific user, whereas the filter here checks for membership within a group. 

There is a config option for an auth_ad_user_filter, which defaults to "(objectclass=user)". We could potentially just update that to filter disabled users, but i'm not sure that gains us anything.

murrant:
I know we filter disabled users elsewhere, I was just wondering if we could consolidate them.

network-guy:
Short of setting a global LDAP filter string, I'm not sure where we could reuse the code. If you can identify something I would be happy to make the changes.

network-guy:
@murrant @laf Any other input on this, or can we look at getting it merged?

laf:
@network-guy I don't use AD so I can't test (or comment really). Hopefully @murrant can spare some time to do a final review.

PR #8233, opened 2018-02-08
Adding invalid hostnames is no longer fatal.

murrant:
Will still fail if they are unpingable.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/addhost.inc.php
+++ b/html/pages/addhost.inc.php
@@ -11,23 +11,19 @@ if ($_SESSION['userlevel'] < 10) {
     exit;
 }
 
-if (isset($_POST['hostname'])) {
-    $hostname = clean($_POST['hostname']);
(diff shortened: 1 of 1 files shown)


laf:
This disables snmp by default so users would add ping only devices.

Not really sure what the difference is, from testing adding some fake devices I don't see what's changed.

murrant:
When I tested snmp was not disabled by default.  I'll retest that.

Before, if the hostname was invalid, it would not try to ping it.

Now, it still shows an error, but it tries to ping/add it anyway.

This was for the user with underscores in his hostnames.

laf:
> This was for the user with underscores in his hostnames.

Is it not easier just to update the validation?

murrant:
The validation is correct in my opinion.

PR #8238, opened 2018-02-09
Fix panel device header

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


InsaneSplash:
BEFORE:
image

AFTER:
image

html/index.php
```diff
@@ -129,7 +129,7 @@
   <link href="css/leaflet.awesome-markers.css" rel="stylesheet" type="text/css" />
   <link href="css/select2.min.css" rel="stylesheet" type="text/css" />
   <link href="<?php echo($config['stylesheet']);  ?>?ver=291727421" rel="stylesheet" type="text/css" />
-  <link href="css/<?php echo $config['site_style']; ?>.css?ver=632417642" rel="stylesheet" type="text/css" />
+  <link href="css/<?php echo $config['site_style']; ?>.css?ver=632417692" rel="stylesheet" type="text/css" />
```
laf: This isn't needed when no css files are changed or did you miss committing that change?

InsaneSplash: :) Did then reverted them back... so forgot to revert this back too :)

laf:
I've tested this with mono and dark themes and don't get the change here. Can you explain in what scenario this is supposed to work?

InsaneSplash:
Very strange... Its basically the sub-menu bar within the device information pages that seems to have been missed with regards to the theme. The highlight is correct, but the background has not been set.... Ill dbl check my work :)

laf:
Hmm, might have me being blind but seems ok now.

PR #8242, opened 2018-02-11
Fix issues with new config init

murrant:
non $config variables defined in defaults.inc.php are no longer global, move them into $config.
include process_config.php changes since merge and delete file

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/process_config.inc.php
```diff
@@ -1,61 +0,0 @@
-<?php
```
laf: How come all of this file has been removed?

murrant: Its replacement is inside Config.php and is `process_config.inc.php`never called right now.

PR #8245, opened 2018-02-12
Add Billing Graphs & Data to API

richardlawley:
This PR makes billing graphs available through the API, along with the data backing them for use generating graphs in an external system.  Updated documentation demonstrates the new API functions available.

As part of this, the billing_graph and bandwidth_graph pages were moved into the standard graph structure.  The remaining billing_graph.php and bandwidth_graph.php pages have been converted to redirect to the new location, but this should really only affect people who have been accessing them directly, or that a page opened prior to updating librenms might try loading the old pages.  If preferred, we can simply remove these pages.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/API/Bills.md
+++ b/doc/API/Bills.md
@@ -134,7 +134,9 @@ Output:
 
 ### `get_bill_graph`
 
-Retrieve a graph associated with a bill
+Retrieve a graph image associated with a bill.
+
+NB: The graphs returned from this will always be png as they do not come from rrdtool, even if you have SVG set.
 
 Route: `/api/v0/bills/:id/graphs/:graph_type
 
@@ -149,13 +151,13 @@ curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/gra
 
 Output:
 
-(image)
+Graph Image
 
 ### `get_bill_graphdata`
 
 Retrieve the data used to draw a graph so it can be rendered in an external system
 
-Route: `/api/v0/bills/:id/graphdata/:graph_type
+Route: `/api/v0/bills/:id/graphdata/:graph_type`
 
 Example:
 
@@ -241,6 +243,8 @@ Output:
 ### `get_bill_history_graph`
 
 Retrieve a graph of a previous period of a bill
+
+NB: The graphs returned from this will always be png as they do not come from rrdtool, even if you have SVG set.
 
 Route: `/api/v0/bills/:id/history/:bill_hist_id/graphs/:graph_type`
 
--- a/doc/API/Bills.md
+++ b/doc/API/Bills.md
@@ -159,12 +159,19 @@ Retrieve the data used to draw a graph so it can be rendered in an external syst
 
 Route: `/api/v0/bills/:id/graphdata/:graph_type`
 
+Input:
+
+The `reducefactor` parameter is used to reduce the number of data points.  Billing data has 5 minute granularity, so
+requesting a graph for a long time period will result in many data points.  If not supplied, it will be automatically
+calculated.  A reducefactor of 1 means return all items, 2 means half of the items etc.
+
 Example:
 
 ```curl
 curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/graphdata/bits
 curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/graphdata/bits?from=1517443200
 curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/graphdata/bits?from=1517443200&to=1517788800
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/graphdata/bits?from=1517443200&to=1517788800&reducefactor=5
 ```
 
 Output:
@@ -190,7 +197,16 @@ Output:
         "ticks": [
             "1517750401",
             "1517756101"
-        ]
+        ],
+        "rate_95th": "251880417",
+        "rate_average": "146575554",
+        "bill_type": "cdr",
+        "max_in": 9888289942,
+        "max_out": 75848756,
+        "ave_in": 18029660.242105871,
+        "ave_out": 196447.38060137472,
+        "last_in": 3790227.9500000002,
+        "last_out": 122731.63333333333
     }
 }
 
--- a/doc/API/Bills.md
+++ b/doc/API/Bills.md
@@ -132,6 +132,66 @@ Output:
 }
 ```
 
+### `get_bill_graph`
+
+Retrieve a graph associated with a bill
+
+Route: `/api/v0/bills/:id/graphs/:graph_type
+
+Example:
+
+```curl
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/graphs/bits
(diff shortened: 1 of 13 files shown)


richardlawley:
The 5 scrut issues are false positives - either it has detected the wrong type signature for the 3rd party JpGraph library, or it's detecting a duplicate local function in different pages (only one will ever get included).

laf:
Let us know when you are done pushing updates so we can code review :)

laf:
I'm getting this for Transfer graphs within billing (confirmed it works ok in current master):

image

richardlawley:
Hmm, that's tricky to diagnose.  Can you let me know the URL of one of the images?  I've got this PR applied to our live setup and everything is working fine for me

doc/API/Bills.md
```diff
@@ -132,6 +132,66 @@ Output:
...
+
+Route: `/api/v0/bills/:id/graphs/:graph_type
+
+Example:
+
+```curl
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/graphs/bits
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/graphs/bits?from=1517443200
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/graphs/bits?from=1517443200&to=1517788800
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/graphs/monthly
```
laf: Probably worth documenting this returns a png no matter whether you say you want svg graphs or not.

doc/API/Bills.md
```diff
@@ -132,6 +132,66 @@ Output:
...
+### `get_bill_graphdata`
+
+Retrieve the data used to draw a graph so it can be rendered in an external system
+
+Route: `/api/v0/bills/:id/graphdata/:graph_type
+
+Example:
+
+```curl
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/graphdata/bits
```
laf: This isn't valid:

```
{
    "status": "error",
    "message": "from parameter must be supplied"
}
```

doc/API/Bills.md
```diff
@@ -132,6 +132,66 @@ Output:
...
+
+Output:
+
+(image)
+
+### `get_bill_graphdata`
+
+Retrieve the data used to draw a graph so it can be rendered in an external system
+
+Route: `/api/v0/bills/:id/graphdata/:graph_type
```
laf: Should this be called `/api/v0/bills/:id/data/:type`

Also missing the backtick on the line.

doc/API/Bills.md
```diff
@@ -176,4 +236,42 @@ Output:
...
+
+Output:
+
+(image)
+
+### `get_bill_history_graphdata`
+
+Retrieve the data for a graph of a previous period of a bill, to be rendered in an external system
+
+Route: `/api/v0/bills/:id/history/:bill_hist_id/graphdata/:graph_type`
```
laf: Maybe: `/api/v0/bills/:id/history/:bill_hist_id/data/:type`

doc/API/Bills.md
```diff
@@ -176,4 +236,42 @@ Output:
...
+
+Input:
+
+Example:
+
+```curl
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/history/1/graphdata/bits
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/history/1/graphdata/hour
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/bills/1/history/1/graphdata/day
```
laf: The output for /day:

```
            "17\nJan",
            "18\nJan",
            "19\nJan",
            "20\nJan",
```

Should we add `\n` here as that should be potentially dealt with on the client end. Instead do `17 Jan` or `17-Jan`

laf:
> Hmm, that's tricky to diagnose. Can you let me know the URL of one of the images? I've got this PR applied to our live setup and everything is working fine for me

Sure. It's just /bill/bill_id=4/view=transfer/

laf:
Graph image: 

`https://web01.lathwood.uk/graph.php?type=bill_historictransfer&id=4&from=1516060800&to=1518739199&imgtype=day&width=1190&height=250frequency`

Working one:

`https://web01.lathwood.uk/bandwidth-graph.php?bill_id=4&bill_code=&from=1516060800&to=1518739199&type=day&imgbill=1&x=1190&y=250frequency`

I have no idea where frequency is coming from in that!

laf:
Sorry hit the wrong button

laf:
Found it:

https://p.libren.ms/view/d6c5f89b

richardlawley:
Not quite sure what it is you're showing with that link - lines 13/22 in your diff don't seem to correspond with the same file from the PR.  Is it possible there's another change you've got merged in causing a conflict?  Comparing to https://github.com/librenms/librenms/pull/8245/files#diff-b96e6cb436083334eca02bb816d9c017 I don't have any lines which just contain `$bi .= '>";` - they are `$bi .= '$type'>";`.

I presume the images work fine if you remove frequency from the URL?

laf:
> I don't have any lines which just contain $bi .= '>"; - they are $bi .= '$type'>";.
> 
> I presume the images work fine if you remove frequency from the URL?

Correct. And it's the `$type` that's causing the issue.

doc/API/Bills.md
richardlawley: I think it's better as graphdata, because the data is pre-processed for use on a graph (number of data points is reduced).  The graph code was actually refactored so that this API call, the graph API calls and the graphs themselves all call the same functions to get data.

laf: Ok no worries.

doc/API/Bills.md
richardlawley: Just trying to work out the best way to handle this - it's applicable to all of the transfer graphs.  Wondering whether to just return the unix timestamps of the start/end of each item and let all formatting be done on the client end.  Only potential problem is if it introduces a timezone dependency

doc/API/Bills.md
laf: @richardlawley This now returns a http 500.

richardlawley: Unable to reproduce this again, on both my dev and live installs.  Do you get a log message, or perhaps when adding &debug to the URL?

laf: Your latest commit fixes that now :)

richardlawley: I've just found that the bills/1/history/1/graphdata/bits was giving a 500 error - is it possible you meant this route rather than bills/1/graphdata/bits ?

If so, this is now fixed.

laf: Possibly yes but it's definitely fixed as I checked out your previous commit and it broke again.

laf:
Big thanks as always for adding this information in @richardlawley

Harpreetsinghbedi:
Hello,

Thank you for add create/edit bill API in LibreNMS.

In LibreNMS create bill API "https://docs.librenms.org/#API/Bills/#create_edit_bill", there is an field name like "bill_custid". Please let me now, where from i will get this bill_custid?

Best Regards
Harpreet Singh

richardlawley:
The Create/Edit bill API was not added in this PR (or by me), and this is not the correct location to ask questions like this.  The best place to ask questions is the discord board: https://t.libren.ms/discord 

However, pretty sure that field is just free text and corresponds with the "Customer Reference" field in the web form.

PR #8246, opened 2018-02-13
Adding Gitlab to Transports

dword4:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


doc/Alerting/Transports.md
```diff
@@ -449,3 +449,13 @@ $config['alert']['transports']['jira']['password'] = 'myjirapass';
 $config['alert']['transports']['jira']['prjkey'][]  = 'JIRAPROJECTKEY';
 $config['alert']['transports']['jira']['issuetype'][]  = 'Myissuetype';
 ```
+
+## Gitlab
+
+LibreNMS will create issues for warning and critical level alerts however only title and description are set.  Uses Peronal access tokens to authenticate with Gitlab and will store the token in cleartext.
```
laf: `Personal` (SP)

dword4: Yeah unless LibreNMS stores the Issue ID somewhere you end up having to do a search through issues to match the alert to the existing issue.  I'm sure it can be done but since I based this off the Jira transport I don't believe it is built to resolve the issues yet, however that definitely merits investigation so that run-away recycling alerts don't just pile up.

dword4: I almost have auto-resolution worked out, however it seems like the code block below doesn't seem to actually work when I tested it

```
        // Don't create tickets for resolutions
        if ($obj['severity'] == 'recovery' && $obj['msg'] != 'This is a test alert') {
            return true;
        }```

In my testing setup I get tickets created on initial alert and when the alert clears.  Once I get this nailed down issues should be auto-closed when the alert clears however this brings me to another question.  Should this be a configurable behavior or just the default behavior for the Gitlab transport?

laf: Isn't that just to stop a new ticket being created for recovery?

dword4: The code block I pasted is supposed to do that, however in testing it doesn't appear to work.

laf: If you just do a `var_dump($obj);exit;` and then run alerts.php manually to see what data it has that might help work out why.

dword4: Okay I figured it out after putting some time in, $obj['severity'] never changes to recovery when say a device comes back online but $obj['state'] changes from 1 to 0 so that code would never work.  Should I bother with trying to flesh out a somewhat-shaky auto-resolve system to clear out issues once it clears or do you think this is sufficient for now?

laf: I think it's ok for now.

What I was thinking was having another column in the alerts table to store a serialized array of data you can store which would include anything back from the initial call and then when a recover comes through, you can access that data and do something else. I.e save the original issue number created - which I assume is returned in the original api call?

PR #8252, opened 2018-02-14
Add a filter on the LDAP request on getUserlist() function

pep-un:
Add the group filtering improve the speed of some pages (Dashboard and
Edit User mainly). If we don't apply this filter all the users are getted
from LDAP and used in to the loop.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


pep-un:
I just realise by fix totally brake the access management.

PR #8254, opened 2018-02-14
feature: Add more Dell iDrac state sensors

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Any chance of test data :)?

Rosiak:
Will add when I'm back from vacation :)

murrant:
@Rosiak you need to add the json file `tests/data/drac.json` too :) If you haven't generated that yet, run `./scripts/save-test-data.php -o drac -m sensors`

laf:
I've pushed the json file so this can be merged once it's passed tests.

laf:
Or not. I love our tests but sometimes, I hate our tests :(

murrant:
I think we need better feedback when things fail, but I have no idea how to do that as sometimes it is something esoteric.

murrant:
I get this output everytime I add it.  The reason you couldn't find the test bug is because it is a code bug:
image

murrant:
Oh and looking at it more closely, the bug is in the old state code.........

laf:
Thanks for taking a look at this @murrant

murrant:
No problem.  The fix was `-e` ;-)

PR #8255, opened 2018-02-14
Fix alerting email issues

murrant:
Variable collision
Only affects users that have not set the email delivery method in config.php.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
> Only affects users that have not set the email delivery method in config.php or the webui.

Our schema update biting people again. That's part of schema 051 and sets it to mail by default so everyone should have a default but I'm guessing it's not the case :(

laf:
This actually breaks things for me. The webui no longer allows the drop down for how to deliver mail. It's blank.

That config should be in the db, or at least the old one was.

PR #8263, opened 2018-02-18
Fix bgp peers discovery cleanup

murrant:
Take context into account, update index for that

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/bgp-peers.inc.php
+++ b/includes/discovery/bgp-peers.inc.php
@@ -8,7 +8,7 @@ if (Config::get('enable_bgp')) {
     if (key_exists('vrf_lite_cisco', $device) && (count($device['vrf_lite_cisco'])!=0)) {
         $vrfs_lite_cisco = $device['vrf_lite_cisco'];
     } else {
-        $vrfs_lite_cisco = array(array('context_name'=>null));
+        $vrfs_lite_cisco = array(array('context_name'=>''));
     }
 
     $bgpLocalAs = snmp_getnext($device, 'bgpLocalAs', '-Oqvn', 'BGP4-MIB');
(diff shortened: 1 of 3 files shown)


includes/discovery/bgp-peers.inc.php
```diff
@@ -125,12 +125,16 @@
...
+                $af_query = "SELECT bgpPeerIdentifier, afi, safi FROM bgpPeers_cbgp WHERE `device_id`=? AND bgpPeerIdentifier=? AND context_name=?";
+                foreach (dbFetchRows($af_query, [$device['device_id'], $peer['ip'], $device['context_name']]) as $entry) {
                     $afi  = $entry['afi'];
                     $safi = $entry['safi'];
                     if (!$af_list[$entry['bgpPeerIdentifier']][$afi][$safi]) {
-                        dbDelete('bgpPeers_cbgp', '`device_id` = ? AND `bgpPeerIdentifier` = ? AND afi=? AND safi=?', array($device['device_id'], $peer['ip'], $afi, $safi));
+                        dbDelete(
+                            'bgpPeers_cbgp',
+                            '`device_id`=? AND `bgpPeerIdentifier`=? AND context_name=? AND afi=? AND safi=?',
+                            [$device['device_id'], $peer['ip'], $device['context_name'], $afi, $safi]
```
laf: Is that the first time we've had an array shorthand in since the split :D

laf:
I can only test this on Arista. Works fine for those devices.

includes/discovery/bgp-peers.inc.php
murrant: No, there is some in the processor code I think.  Not ready for a mass replace, but I figured I might as well use it in new/changed code.  New is probably safer in case we need to backport.

murrant:
I tested it with several models using the test data.

PR #8274, opened 2018-02-21
save-test-data.php: Print a message when we encounter a bad module name

murrant:
Should we exit when we encounter this?

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
> Should we exit when we encounter this?

I think so, not sure it matters either way but at least exiting means the user should notice.

Issue #7842, opened 2017-12-01
Aerohive switches are cataloged as wireless devices.


daniagar:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

laf:
Thanks for the bug report or new device request, unfortunately you have not provided the information asked for in point 3 when you created this issue.

Please provide pastebin links to the information we need and we will re-open this issue.

daniagar:
![image](https://user-images.githubusercontent.com/34174425/33520848-51a68e1a-d7cb-11e7-8d52-f96810d2f6d6.png)

Hello, 
Sorry for the missing info. I am new to github.
Below are the information you requested. If anything else is needed please specify.

https://p.libren.ms/view/04a78aa4
https://p.libren.ms/view/e48fd799
https://p.libren.ms/view/9fc7c6e7

daniagar:
Also : 
====================================
Component | Version
--------- | -------
LibreNMS  | 1.33-194-g995b706
DB Schema | 220
PHP       | 5.4.16
MySQL     | 5.5.56-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
====================================

daniagar:
[ah_smi_mib.txt](https://github.com/librenms/librenms/files/1545850/ah_smi_mib.txt)
[ah_trap_mib.txt](https://github.com/librenms/librenms/files/1545851/ah_trap_mib.txt)
[...]

PR #8277, opened 2018-02-21
Added better detection for HiveOS devices

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This will need a notification but first let's agree that this is what we should do.

aerohive is the vendor rather than an OS. HiveOS is the OS but is shared over switches and aps so I've split the detection out.

Fixes: #7842


laf:
Notification sent for the 23rd of March. Do not merge until then.

PR #8280, opened 2018-02-21
fix: Dell N1548 CPU graphs

asteen-nexcess:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/OS/Powerconnect.php
+++ b/LibreNMS/OS/Powerconnect.php
@@ -59,11 +59,7 @@ class Powerconnect extends OS implements ProcessorDiscovery, ProcessorPolling
                     0
                 )
             );
-        } elseif (starts_with($device['sysObjectID'], '.1.3.6.1.4.1.674.10895.3024')) {
-            d_echo("Dell Powerconnect 8024F");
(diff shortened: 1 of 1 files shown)


LibreNMS/OS/Powerconnect.php
```diff
@@ -62,6 +62,9 @@ public function discoverProcessors()
         } elseif (starts_with($device['sysObjectID'], '.1.3.6.1.4.1.674.10895.3024')) {
             d_echo("Dell Powerconnect 8024F");
             return $this->discoverVxworksProcessors('.1.3.6.1.4.1.674.10895.5000.2.6132.1.1.1.1.4.9.0');
+        } elseif (starts_with($device['sysObjectID'], '.1.3.6.1.4.1.674.10895.3065')) {
```
laf: `starts_with` accepts an array as the second arg so seeing as you are just returning the same as the one above then remove the new elseif and change the one above to:

```php
} elseif (starts_with($device['sysObjectID'], ['.1.3.6.1.4.1.674.10895.3024', '.1.3.6.1.4.1.674.10895.3065'])) {
```

Give that a try and update the PR.

asteen-nexcess: @laf thanks! Updated.

PR #8282, opened 2018-02-21
Set Content-type header in the API transport if the text to be sent is valid JSON

nkaminski:
This change makes the API transport attempt to decode the request body text that is to be posted to the API endpoint as JSON prior to the data being posted. If the body text is identified as valid JSON, the Content-type HTTP header is set to 'application/json', which is required by many API endpoints that expect JSON formatted inputs.

#### Please note

#### Testers


laf:
@nkaminski Thanks for contributing :)

PR #8287, opened 2018-02-22
feature: Update jquery_lazyload

Rosiak:
- Removed the included library, instead only include the js file.
- Update to latest version

We should probably move to https://github.com/verlok/lazyload at some point, I tried to fiddle a bit with it, but that didn't end well(Also this version doesn't rely on jQuery).


laf:
Works for me. Thanks @Rosiak

PR #8292, opened 2018-02-24
Error for insufficient php version during install

murrant:
Remove platform from composer.json so users get an error if their php version isn't new enough.
Add extensions to composer.json too

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Shouldn't this have the .lock file updated as well?

murrant:
I don't think so as I'm not updating any library versions.

When updating the .lock file, we'll need to re-add `"platform": {"php": "5.6.4"}` or run it on that version of php.

PR #8293, opened 2018-02-24
feature: Added new alert rule builder UI

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 8293`

**NOTE to testers** This will break your alert mapping if you go back to master before this PR is merged.

This is ready for widespread testing!

It provides an update to the ui for creating alert rules. It's had limited testing so far so be warned.

Some things to note:

  - [x] ~~A console error will show when editing a new rule. This is expected.~~
  - [x] ~~Rules created the old way should continue to work. If you edit an old rule, it will be updated.~~
  - [x] ~~You can still create a new rule using the old system.~~
  - [x] ~~Interval no longer seems to be honoured so I need to look in to that.~~

To do:

- [x] ~~Automatically import old rules when they are edited.~~ Manually allow the importing of old rules.
- [x] Allow users to paste in an SQL query that will be parsed.
- [x] Fix sql gen when adding default rules
- [x] Collection fully converted to query_builder format
- [x] Allow add from collection to use query_builder format
- [x] (<,>,<=,>=) missing for integer type
- [x] List showing SQL instead of rule/query_builder
- [x] Convert to new format ~~button in old edit dialog or simply~~ - auto convert on edit
- [x] The webui layout / button will change. This is purely for testing so far.
- [x] Query Builder select too wide (layout issue not select box)
- [x] Remove old modal for adding / editing rules.
- [x] List all resolvable tables + fields
- [ ] Make IN + NOT IN work.

Please expect something to break so don't attempt to test this unless you are prepared.

image


laf:
Just an FYI that the interval value no longer seems to be honoured so I need to look into that.

murrant:
I definitely love this.

A couple of thoughts suggestions:
 * Should we store db_schema.yaml as json?  json_decode() is a lot faster.
 * Put macros before the normal fields instead of after, or sort them all alphabetically.
 * Pass id to mysql_type_to_php_type(), that way we can add some specific customizations/helpers.
 * For example, I changed all macros to Yes/No radio options in v2 instead of an text box

I assume you have some refinement for the webui planned after you get it functioning fully.  (Non-modal, doesn't dismiss when saving)

laf:
> Should we store db_schema.yaml as json? json_decode() is a lot faster.

No issue from my side, we don't need human readable file for that so json will be fine. I'll class it as out of scope for this PR for now.

> Put macros before the normal fields instead of after, or sort them all alphabetically.

Done.

> Pass id to mysql_type_to_php_type(), that way we can add some specific customizations/helpers.

What ID?

> For example, I changed all macros to Yes/No radio options in v2 instead of an text box

You can't use Yes/No for all macros, some require int such as bandwidth > x.

> I assume you have some refinement for the webui planned after you get it functioning fully. (Non-modal, doesn't dismiss when saving)

Once it's tested in enough detail yes. Although I think I'll keep the non-modal. I don't think it will look good in a confined space like that.

murrant:
Id is table.field

laf:
> Id is table.field

Added.

queryBuilder wasn't playing nice within a modal so it's a separate page for now.

Would love proper testing now and we can work out how to make the UI work later.

murrant:
Fixed a couple issues with the rule colleciton.
TODO:
 [ ] Fix sql gen when macros are used
 [ ] Collection fully converted to query_builder format
 [ ] Allow add from collection to use query_builder format
 [ ] List showing SQL instead of rule/query_builder
 [ ] Query Builder select too wide (layout issue not select box)

laf:
Thanks for taking a look :)

>  Collection fully converted to query_builder format

Any reason why?

laf:
I've tackled a few more on that list. Will get to the auto converting bit soon.

murrant:
I'd like to convert the saved rules so people can edit them with the new editor after adding.

Perhaps we should add an export button to make it easier to share people's rules.

laf:
> I'd like to convert the saved rules so people can edit them with the new editor after adding.

Unless I'm missing something when you import from the collection you can edit using the new ui as normal - do you mean some other way?

Export sounds good. They can just copy the query shown in the alert list though.

laf:
So I've update the edit rule to deal with the old format. It doesn't auto save but parses the old format ready for use. If the user chooses not to save then we just re-import next time they edit. If they do save, we store the query_builder value and it's then used going forward.

I've moved the double and float types to be strings, the reason for this is that REGEXP can be used on those but doesn't pass validation.

murrant:
Modal is working mostly.  Looks much better with larger size.
Select box works, but width is a little wonky, that needs some tweaking.
There are probably lots of unused things that need to be removed.
I don't understand what a "device limited rule" is so that may be broken.

laf:
> Modal is working mostly. Looks much better with larger size.

Ace, thanks for that.

> Select box works, but width is a little wonky, that needs some tweaking.

Any reason you're trying to narrow it down?

> I don't understand what a "device limited rule" is so that may be broken.

Not sure where you've got device limited rule from but I'm guessing you mean device specific rule? That's where you create a rule in the edit device section specifically so it's automatically bound to the device.

We potentially could update this to drop the device_id = -1 think and just say users need to bind rules to groups. It does mean it's an extra step for them + we can't auto convert those existing rules that are bound to a specific device without creating a unique group for each rule.

laf:
Oh and the table reload itself. We aren't using bootgrid here. A quick attempt to by just doing:

`$("#grid-basic").bootgrid();` resulted in the table disappearing. I didn't look much past that but it's easy enough to just update the relevant cells possibly as we do that on delete, just remove the td associated with that rule - or obviously get bootgrid working.

murrant:
@laf I don't want the select narrow :) It automatically sets width to the first option, which is '-----', I've got a little compromise for that merged now, seems ok.

Ok, I'll check the device create/edit rules later.

Reloading bootgrid won't work because we embed the data in the html, not ajax.  That is why I just reload the page for now.

laf:
> Reloading bootgrid won't work because we embed the data in the html, not ajax. That is why I just reload the page for now.

It would work with full bootgrid as the data is loaded in via ajax, I just wanted to check if it worked loading bootgrid against an existing table which it didn't and I didn't want to spend much time working that one out.

murrant:
@laf during my modal refactoring looks like I've broken editing old rules.  Any idea where the issue is?

laf:
> I'm wondering should we store the raw querybuilder json?
> Yeah, I know this requires some re-writing I would be willing.

I was doing that at some stage. It's pretty quick to implement that but out of interest why?

It makes parsing much better as it doesn't use the sql plugin but we'd only be able to display the full query in the alert rules list.

laf:
> @laf during my modal refactoring looks like I've broken editing old rules. Any idea where the issue is?

Looks like no code for the old way is running anymore. It just pulls in the query_builder value which is empty for old rules.

laf:
```
diff --git a/html/includes/forms/parse-alert-rule.inc.php b/html/includes/forms/parse-alert-rule.inc.php
index 7913085ea..7e277c809 100644
--- a/html/includes/forms/parse-alert-rule.inc.php
+++ b/html/includes/forms/parse-alert-rule.inc.php
@@ -27,6 +27,15 @@ if (is_numeric($alert_id) && $alert_id > 0) {
     $rule = $tmp_rules[$template_id];
 }
 if (is_array($rule)) {
+    if (empty($rule['query_builder'])) {
+        $sql_query = $rule['rule'];
+        $sql_query = str_replace('&&', 'AND', $sql_query);
+        $sql_query = str_replace('||', 'OR', $sql_query);
+        $sql_query = str_replace('%', '', $sql_query);
+        $sql_query = str_replace('"', "'", $sql_query);
+        $sql_query = str_replace('~', "REGEXP", $sql_query);
+        $rule['query_builder'] = $sql_query;
+    }
     $output             = array(
         'severity'      => $rule['severity'],
         'extra'         => $rule['extra'],
```

imcdona:
What are these tables for? I'd like to add comments to the schema updates in this push request.

``alert_group_map` 
`alert_rules`
`alert_map`
`alert_device_map``

laf:
`alert_group_map` = the mapping of groups to an alert rule.

`alert_rules` = the actual alert rules.

`alert_map` = is now `alert_device_map`

`alert_device_map` = the mapping of devices to an alert rule.

murrant:
Note to self remove macros from schema.

Rosiak:
Still testing this, so far no issues.

thomseddon:
I think this broke our alert mapping as we've just received a few hundred alerts for devices to which the alert isn't mapped too.

Checking in the code, I don't think it's observing groups or mappings correctly,

Previous (prior to this change)
```php
function GetRules($device)
{
    $groups = GetGroupsFromDevice($device);
    $params = array($device,$device);
    $where = "";
    foreach ($groups as $group) {
        $where .= " || alert_map.target = ?";
        $params[] = 'g'.$group;
    }
    return dbFetchRows('SELECT alert_rules.* FROM alert_rules LEFT JOIN alert_map ON alert_rules.id=alert_map.rule WHERE alert_rules.disabled = 0 && ( (alert_rules.device_id = -1 || alert_rules.device_id = ? ) || alert_map.target = ? '.$where.' )', $params);
}
```

Now we have:
```php
function GetRules($device)
{
    $groups = GetGroupsFromDevice($device);
    $params = array($device, $device);
    $where = "";
    foreach ($groups as $group) {
        $where .= " || group_id = ?";
        $params[] = $group;
    }
    $query = "SELECT DISTINCT a.* FROM alert_rules a LEFT JOIN alert_device_map d ON a.id=d.rule_id LEFT JOIN alert_group_map g ON a.id=d.rule_id WHERE a.disabled = 0 AND ((d.device_id=? OR d.device_id IS NULL) $where)";
    return dbFetchRows($query, $params);
}
```

Notice how the previous query checked for `alert_rules.device_id = -1` whereas the current query checks `d.device_id IS NULL`. The issue with the new query is that in the case there are no rows in alert_device_map that match the join (`a.id=d.rule_id `) then the value of `d.device_id` will be NULL (i.e. it will match in the absence of a rule, where it shouldn't).

As an example, see the rules matched below where there is nothing in the alert_device_map table:
```
MariaDB [librenms]> SELECT * FROM alert_device_map;
Empty set (0.01 sec)

MariaDB [librenms]> SELECT DISTINCT a.id, a.name, d.*, g.*  FROM alert_rules a LEFT JOIN alert_device_map d ON a.id=d.rule_id LEFT JOIN alert_group_map g ON a.id=d.rule_id WHERE a.disabled = 0 AND ((d.device_id='402' OR d.device_id IS NULL) );
+----+---------------------------------+------+---------+-----------+------+---------+----------+
| id | name                            | id   | rule_id | device_id | id   | rule_id | group_id |
+----+---------------------------------+------+---------+-----------+------+---------+----------+
|  3 | bgp session down                | NULL |    NULL |      NULL | NULL |    NULL |     NULL |
|  5 | Port status up/down             | NULL |    NULL |      NULL | NULL |    NULL |     NULL |
|  6 | Port utilisation over threshold | NULL |    NULL |      NULL | NULL |    NULL |     NULL |
+----+---------------------------------+------+---------+-----------+------+---------+----------+
3 rows in set (0.00 sec)

```



Also, `device_id`  was correctly updated to `d.device_id` but I think `group_id` should be updated to `g.group_id`.

thomseddon:
Also, I can submit a PR, but as @laf has been working on this so recently I thought he may be best place to understand all implications and check any other gotchas.

I think the schema will need to be changed, happy to build a PR if it would help.

laf:
@thomseddon We fixed that earlier, just re-run daily.sh

laf:
In fact you already have.

laf:
it probably should be:

`$query = "SELECT DISTINCT a.* FROM alert_rules a LEFT JOIN alert_device_map d ON a.id=d.rule_id LEFT JOIN alert_group_map g ON a.id=g.rule_id WHERE a.disabled = 0 AND ((d.device_id=? OR d.device_id IS NULL) $where)";`

Like you said, can you test and submit a PR @thomseddon

murrant:
@thomseddon Fixed, update your install
#8389

thomseddon:
Yeah that fixed it :+1: thanks

murrant:
And double fixed #8391

BradHooper:
Could we please add the operators IN and regex for number based values?

For example previously I could match on multiple BGP ASN in a list now I cannot.

murrant:
You can, but it means you have to add a lot of rules. I meant here btw #8353

PR #8294, opened 2018-02-25
ewc: more WirelessSensors and processor/mempools support

trs80:
Errors, NoiseFloor, RSSI, SNR and Utilization WirelessSensors
processor/mempools support using existing enterasys code

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/OS/Ewc.php
```diff
@@ -127,4 +138,167 @@ public function discoverWirelessClients()
...
+    /**
+     * Discover wireless frequency.  This is in MHz. Type is frequency.
+     * Returns an array of LibreNMS\Device\Sensor objects that have been discovered
+     *
+     * @return array Sensors
+     */
+    public function discoverWirelessFrequency()
+    {
+        $oids = snmpwalk_cache_oid($this->getDevice(), 'apRadioStatusChannel', array(), 'HIPATH-WIRELESS-HWC-MIB');
+        $ap_interfaces = $this->getCacheByIndex('ifName', 'IEEE8021-BRIDGE-MIB');
```
murrant: IEEE8021-BRIDGE-MIB does not contain ifName.
Probably try to avoid fetching again unless you don't have the data

murrant:
Looks great.  Just one item about ifName.

Something like this:
```php
$ifName = array_column('ifIndex', 'ifName', dbFetchColumn("SELECT ifIndex,ifName FROM ports WHERE device_id=?", $this->getDeviceId());
```

trs80:
Hmm, "$this->getDeviceId() is of type integer, but the function expects a array."  I'm still getting data though.

trs80:
Some of the sensor OIDs (RSS and SNR) can return -100 when there's no clients associated, is it possible to handle this in the WirelessSensor code somewhere so that no data is recorded?

BTW `LibreNMS/OS/Ceraos.php` has lots of `$ifNames = $this->getCacheByIndex('ifName', 'IF-MIB');`

murrant:
Yeah, that is fine.  But I would like to find a solution in the future.

murrant:
Oh and you need to wrap device id in an array.  Types that off the top of my head.

murrant:
That won't work either.  We don't like inter-module dependencies.

It will have to check for data then fall back to snmp.  Just put it back to what you had before (using the proper IF-MIB) and we can update all similar calls with some common code later.

murrant:
@trs80 Add some test data and this is good to merge.

https://docs.librenms.org/#Developing/os/Test-Units/#example-workflow

PR #8301, opened 2018-02-27
device: Add toner discovery and polling for Kyocera

corny:


laf:
Simples :)

PR #8304, opened 2018-02-27
Removed unwanted panel graph header

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


html/pages/graphs.inc.php
```diff
@@ -71,7 +71,6 @@
 
     print_optionbar_end();
 
-    print_optionbar_start();
```
laf: This leaves a rogue `print_optionbar_end()`. Are you sure it needs removing?

InsaneSplash: Yes, the one above this line is for the normal panel header... it seems there was a start() added, but never had an end.... It also changed the mini-graphs in the view to have a black background with the mono theme....

InsaneSplash: let me revisit this.... as I found one way down in the code..... very weird.... and I dont think its needed to be honest....

laf: What I mean is if you search for print_optionbar_ in the file, you now have 5 entries where as surely it should be 4 or 6? 3 starts and 3 ends or 2 of each?

InsaneSplash: Should be good now...

InsaneSplash: BEFORE:
image

AFTER:
image

PR #6920, opened 2017-06-30
refactor: Remove composer deps from the repository

murrant:
Allows users to manage installed libraries without conflicting with our git repo.  And also relieves us of distributing other's code.

We can specify target versions in composer.json and exact versions will be supplied by composer.lock.  This makes `composer install` install the exact versions we have tested with.

#### Testers

If you would like to test this pull request then please run:
```
git remote add https://github.com/murrant/librenms.git
git checkout -t murrant/composer-install
git reset --hard 394c4393a4fcf6c1ee42a6f6a07a839fe7cadc5d
./daily.sh
```

####  WARNING LibreNMS versions with known issues:
These are all fixed in 1.35

- Break updates: 0def643e0 to (fix #7864)
- Break LibreNMS until ./daily.sh is run again when distributed polling is enabled: 45b6a4c55 to 2e73b7529
- Break LibreNMS until ./daily.sh is run again: 3f4a35fb9 and older

PR #8305, opened 2018-02-27
Update composer_wrapper to use http links

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Ref: https://github.com/librenms/librenms/pull/6920#issuecomment-368911038


murrant:
What about users that only set http proxy?

murrant:
Also, I started something similar here.  But I wasn't sure it was needed.  What do you think?
https://github.com/librenms/librenms/compare/master...murrant:https-composer

laf:
I'd go with yours then? Can you add the composer version check fix and submit your branch?

murrant:
Sure

PR #8315, opened 2018-03-01
Fixing HP Proliant state sensors

mikenowak:
The support to HP Proliant disk discovery was originally introduced at #6124.

However that code contains a few errors that prevent it from doing what it was intended to do.

1) The MIB that contains cpqDaPhyDrvStatus and cpqDaPhyDrvSmartStatus is **CPQIDA-MIB**, not 
CPQSINFO-MIB.

2) The OID in the code that iterates the array in order to find the disk number should drop the last digit (.1) as that indicates the disk in position (1) and we end up missing the Disk # as follows:
```
Sensor Added: state cpqDaPhyDrvSmartStatus 10.0 Disk #
Sensor Added: state cpqDaPhyDrvStatus 00.0 Disk #
```

Having made the changes described herein, the following is seen in the recent events

```
State sensor Disk # has changed from (0) to (2)
State sensor Disk # has changed from (0) to (2)
Sensor Updated: state cpqDaPhyDrvSmartStatus 10.0 Disk #1
Sensor Updated: state cpqDaPhyDrvStatus 00.0 Disk #1
```

The only other issue that I can name is that the states are not expanded properly in the UI, i.e. 2 should equal to OK.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/sensors/state/hp.inc.php
+++ b/includes/discovery/sensors/state/hp.inc.php
@@ -24,9 +24,10 @@
  */
 
 $tables = array(
-    // One could add more entrys from deviceGroup, but this will do as a start
-    array('cpqDaPhyDrvStatus','.1.3.6.1.4.1.232.3.2.5.1.1.6.','cpqDaPhyDrvStatus','DriveStatus','CPQSINFO-MIB'),
-    array('cpqDaPhyDrvSmartStatus','.1.3.6.1.4.1.232.3.2.5.1.1.57.','cpqDaPhyDrvSmartStatus','SmartStatus','CPQSINFO-MIB'),
+    array('cpqDaPhyDrvTable','.1.3.6.1.4.1.232.3.2.5.1.1.6.','cpqDaPhyDrvStatus','cpqDaPhyDrvBay','CPQIDA-MIB'),
+    array('cpqDaAccelTable','.1.3.6.1.4.1.232.3.2.2.2.1.2.','cpqDaAccelCondition','cpqDaAccelCntlrIndex','CPQIDA-MIB'),
+    array('cpqHeFltTolPowerSupplyTable','.1.3.6.1.4.1.232.6.2.9.3.1.4.','cpqHeFltTolPowerSupplyCondition','cpqHeFltTolPowerSupplyBay','CPQHLTH-MIB','cpqHeFltTolPowerSupplyPresent'),
+    array('cpqHeFltTolFanTable','.1.3.6.1.4.1.232.6.2.6.7.1.9.','cpqHeFltTolFanCondition','cpqHeFltTolFanIndex ','CPQHLTH-MIB','cpqHeFltTolFanPresent'),
 );
 
 $x=0;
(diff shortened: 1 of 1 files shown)


laf:
Thanks for the fix :)

I think if you change:

`create_sensor_to_state_index($device, $state_name, $index);`

to:

`create_sensor_to_state_index($device, $state_name, $x . $index);`

It should fix the display issue, give it a try and update this PR.

mikenowak:
@laf,

thanks for that - I've added the change suggested by you and now the states are represented by nice bootstrap style buttons :)

I've also took this opportunity to extend this code and implement the following:

1) Added support for querying:
* raid ctrl accelerator status
* power supply status
* system fan status

2) Set the labels for the above to `CTRL # Status`, `PSU # Status`,  `FAN # Status`, and I've renamed `Disk #` to `HDD #`. Shall there be a naming convention in place that I am not aware of please let me know.

3) Added checks for PSU and FAN to ensure that they are truly installed `$temp[$index][$tablevalue[5]] == 'present'`, so that we wont alert if the component is not present due to such and such system configuration rather than failure.

4) Set `state_draw_graph` to 0 in all instances as I understand this would prevent the graphs from being created for these states, as these would be meaningless in my opinion. Please correct me if I am missing something here.

5) Last but not least, removed the SMART check as I am not getting anything out of that OID on my test system and based on my experience the SMART errors would result in `cpqDaPhyDrvStatus` becoming `predictiveFailure`.

Please let me know if you need anything else.

f0o:
I think travis died on this one

laf:
> Set state_draw_graph to 0 in all instances as I understand this would prevent the graphs from being created for these states, as these would be meaningless in my opinion. Please correct me if I am missing something here.

Well I'd say graphs are still useful. They show the historic state of the sensors over time which can be useful

> Last but not least, removed the SMART check as I am not getting anything out of that OID on my test system and based on my experience the SMART errors would result in cpqDaPhyDrvStatus becoming predictiveFailure.

I think it might be better to leave them in.

With this amount of changes, can you submit test data as well so we can ensure this code works going forward. https://docs.librenms.org/#Developing/os/Test-Units/#example-workflow

Rosiak:
Just want to add in that state_draw_graph to 0 doesn't make any difference right now. Code to enforce this was never implemented, but hopefully will be with time.

mikenowak:
> I think travis died on this one

Wrong number of spaces, should be good now

> Well I'd say graphs are still useful. They show the historic state of the sensors over time which can be useful

>Just want to add in that state_draw_graph to 0 doesn't make any difference right now. Code to enforce this was never implemented, but hopefully will be with time.

OK, thanks for confirming - this is now reverted to original value of 1.

> I think it might be better to leave them in.

OK, I'll try to do something smart here, since both of these come from the same snmp table.

>With this amount of changes, can you submit test data as well so we can ensure this code works going forward. https://docs.librenms.org/#Developing/os/Test-Units/#example-workflow

OK, will allocate time to this next week.

mikenowak:
Slight correction regarding the SMART status.

I just checked ILO 3 and 4 on Proliants G7, Gen8 and Gen9 and I dont see this exposed anywhere.

So my current thinking is that it isn't really needed unless somebody can prove otherwise please.

laf:
Here's some test data someone else has previously submitted which has the OID for the smart status:

```
.1.3.6.1.4.1.232.3.2.5.1.1.57.1.0 = INTEGER: 2
.1.3.6.1.4.1.232.3.2.5.1.1.57.1.1 = INTEGER: 2
```

murrant:
Unless you update to the state code to the new format, you can't change existing state translations.

https://docs.librenms.org/Developing/Sensor-State-Support/

murrant:
This currently cannot be tested. We don't discover sensors until the second discovery, because we poll 'hardware' instead of discover it.

murrant:
@mikenowak If they weren't in this, then you will need to open a new pull request.

murrant:
That's odd it works on my system.  Can you please stop by discord or community sometime to troubleshoot.

PR #8337, opened 2018-03-08
Borderless fullscreen geographical map, with just the nodes.

dagbdagb:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


cron410:
+1 ill be testing this out as well

cron410:
found some whitespaces as follows:

librenms@librenms:~$ ./scripts/github-apply 8337
<stdin>:13: trailing whitespace.
   margin-bottom: 0;
<stdin>:53: trailing whitespace.
 *
<stdin>:58: trailing whitespace.
 *
Checking patch html/css/fullscreenmap.css...
Checking patch html/pages/front/fullscreenmap.php...
Applied patch html/css/fullscreenmap.css cleanly.
Applied patch html/pages/front/fullscreenmap.php cleanly.
warning: 3 lines add whitespace errors.

laf:
This PR has got messy with commits :)

Because you are changing the body css it's causing weird behaviour now for me:

image

See the scroll bar randomly on the right?

cron410:
A completely different css file was added for the fullscreenmap php file. I
deleted the fullscreenmap.css file on my local install and the scroll bar
stays.

dagbdagb:
@laf 
sorry about the mess. :-)
First time I am doing anything more advanced than git pull with git. But the mess appears to result in a very nice feature, don't you think?

Do you want us to rediff the final result against a pristine checkout?

laf:
Ok so I don't seem to see the scroll bar anymore after a git pull and a force refresh.

I do see a big gray bar at the top, is that to be expected?

image

cron410:
No that is not expected. I don't know much about CSS or PHP so this PR
might need more time in the oven, and definitely more testing. What browser
and system are you running?

dagbdagb:
@laf 
That Should Not Happen (tm). Looks like some CSS is not kicking in for you. I'll set up a proper test env.

dagbdagb:
@laf
I did a fresh install and merged my PR. Map looks great, no gray bar. Is there anything particular about your environment, Neil?

laf:
Nothing special about my install. Tried in 3 browsers, Chrome, FF and Safari and it exhibits the same in all.

dagbdagb:
@laf 
Do you make use of $config['webui']['custom_css'] ?

murrant:
That gray bar is the top of the map... It happens when you have locations in a wide area and your browser window is tall.

murrant:
I have two suggestions.  

1. Change the default grouping setting in this map so it shows more individual pins.
2. Detect fullscreen mode (F11) and remove chrome and restore it when going back to window mode. Then rename this to just geographical map or something.

dagbdagb:
@murrant 
Thank you. @laf : The gray bar you see is simply because leaflet does not have any tiles for that area. You need to zoom a bit. Or set a default zoom in config.php. 
$config['leaflet']['default_zoom'] = 5;

As for your suggestions:
1. Can I change the default grouping without modifying includes/common/worldmap.inc.php ?
2. I do not understand, sorry. F11 is the common 'go fullscreen' key for chrome and ff on linux. Unsure if it is the same for windows. OSX is ctrl-cmd-....F? No detection should be required?
I can rename the map whatever you like.

cron410:
@dagbdagb I think @murrant means detecting the browser is in fullscreen mode, not detecting the actual keypress.

@laf I assume your grey bar issue wasnt fixed? Maybe try this to redetect the size of the screen?
setTimeout(function () {
    map.invalidateSize();
}, 0);

from here: https://stackoverflow.com/questions/31454717/leaflet-only-loads-one-tile

murrant:
The gray bar isn't an issue (per-se) it is just the background behind the map.  And there is no map to display on that part of the screen.  Could maybe do something with the zoom or something to combat this, but the fix could be worse than the issue if we aren't careful.

Yes, I meant detecting that the browser has gone fullscreen.  Some info here may be useful: https://stackoverflow.com/questions/16755129/detect-fullscreen-mode

cron410:
I found a fix for the grey background in Leaflet here:
https://stackoverflow.com/questions/13851888/how-can-i-change-the-default-loading-tile-color-in-leafletjs

I'll push it to @dagbdagb's repo

dagbdagb:
@murrant 
Dynamic styling (hide navbar depending on fullscreen or not) is a big can of worms at the bottom of a very deep hole. Bottom line is that F11 (or whatever the equivalent is on OSX) does not trigger the fullscreen event. And I really want a solution which catches that scenario.

So I need to use the onresize event and compare innerHeight and screenHeight. Whenever the delta is below some threshold, the browser is most likely fullscreen. And vice versa.

Will look into it.

dagbdagb:
Got a solution which appears to work. Will try to implement it for the map tomorrow.

dagbdagb:
Toggling fullscreen now also toggles the menubar.
Someone with better CSS-foo than myself could probably tune it further.

laf:
I've rebased this now so it only shows the relevant changed files.

I'm happy with it now, although should we change the name to 'world map' rather than 'Geographical'?

dagbdagb:
I'd say geographical is better. It does not have to be a *world* map, unless you zoom it out to cover the world.  So, 'geographical' is more generic. I use leaflet config options to zoom in on the area interesting me. (It is the one with the *green* bikeshed.....   :-)

cron410:
I only look at a state level, but surely everyone uses it differently.

On Mar 13, 2018 7:17 AM, "Neil Lathwood" <notifications@github.com> wrote:

I've rebased this now so it only shows the relevant changed files.

I'm happy with it now, although should we change the name to 'world map'
rather than 'Geographical'?

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/librenms/librenms/pull/8337#issuecomment-372631099>,
or mute
the thread

laf:
Geographical doesn't really strike me as obvious to what it is though. I see what you say about the world map but we already call this in the widgets and how much of the world map is shown is down to the user.

It's not really a huge deal but geographical just doesn't strike me as right.

dagbdagb:
I do not have a strong opinion about what we call it. I can live with 'bob'. I fail to see how geographical is wrong, but english is not my first language.

cron410:
Looks a lot better now, but I get different results at different page zoom levels. Using Chrome on Win10, the page shows the menu bar at the top which is great but the LMS logo diappears. If I zoom in a bit, the logo comes back. Resetting the page back to Zero zoom with Ctrl-0 the logo disappears again as expected. Then if I zoom out, the menu bar disappears completely. I'm willing to find a solution to toggle the menubar CSS field by checkbox if that works in a cleaner way.

dagbdagb:
@cron410 
Unsure if you have issues with CSS or the 'fullscreen detection logic' . You may try to fiddle with the 'fullscreen detection logic' first.
Check out the html here https://pastebin.com/Thfm7pjw. Save it as some file on a webserver and see if the detection logic makes sense in your browser/os combo.

dagbdagb:
Heh. Yeah. I can also see that the zoom-level impacts the fullscreen detection logic. The problem is that there is no good, cross-platform, fullscreen detection logic. And the platform/browser specific methods do not handle the use of F11 for fullscreen toggling. Hence I have to resort to heuristics.

I may look into resetting/setting the zoom-level, but I am not entirely convinced that is a great idea either.

murrant:
@cron410 you are zooming the the page, not the map.  How are you zooming?

dagbdagb:
ctrl-plus/ctrl-minus will exhibit the 'anomaly'. 
Be in fullscreen mode with regular zoom.
Hit ctrl-plus until the menubar shows up. The logic/heuristics considers the page to not be fullscreen anymore.

cron410:
It's not a huge issue for me but i could see other complaining about it in
the future. Until we can get the heuristics right, it might be better to
just have a toggle. Should I dig one up or do you guys have another plan?

cron410:
Looks like the issue is only on my tiny laptop screen. Works great on all my coworker's screens and the TV computer. I'm not gonna beat a dead horse.

cron410:
Are there any other issues preventing a merge?

kkrumm1:
Not for me.

murrant:
@cron410 nice work :)

PR #8339, opened 2018-03-08
Use composer.phar first and self-update it.

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Odd one here and no doubt it's me but `./validate.php` now fails with composer not being found.

If I print_r() $found and $matches in `./LibreNMS/Validations/Dependencies.php` I get:

OLD:
```
Array
(
    [0] => Composer version 1.6.3
    [1] => version
    [2] => 1.6.3
)
1
```

NEW:
```
Array
(
)
0
```

The output from composer wrapper:

`Composer version 1.6.3 2018-01-31 16:28:17`

Which matches the regex fine.

Switching back to master works ok.

murrant:
It is the ansi color output... It has the color codes in when we grep for it...

PR #8340, opened 2018-03-08
device: Use 5secs cpu interval for Extereme XOS

zombah:
Most network devices use cpu current counter, but
Extreme XOS miss it, use closest one to current
which is 5secs not 5mins and this makes cpu graph 
more granular.

Signed-off-by: Misha Komarovskiy <zombah@gmail.com>

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Out of interest why not use the 5min one. Using the a 5 second one will most likely give inaccurate results as it will spike from our polling code at the time and show high cpu usage all the time.

zombah:
5min flattening cpu graph too much. Maybe then go for 30sec counter? 5sec was first i noticed between other os who also don't have "current cpu".

murrant:
I would like to store multiple cpu usage periods and graph them all.  That way admins have all data available.  I haven't begun to think about this or what it would entail...  not any room for that on my plate right now.

murrant:
@zombah The problem with making the window too short is you add more measurement bias.

When we poll a device, the cpu usage goes up.  If you cut out most of the time where the device was idle, usage whill look higher than it is.

laf:
@zombah Any update on this?

laf:
Closing until we receive an update.

zombah:
Sorry i totally forgot about this. I like @murrant idea about storing multiple cpu vars and ability to graph them all. I tested 5secs counters for couple weeks there, didn't notice any polling impact but not big granularity bonus also, probably this is case specific. So i revoke this PR for now, storing multiple cpu counter sound much better idea than playing with specific one for specific device type.

PR #8336, opened 2018-03-08
Add detection of Axis Network Speaker

priiduonu:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [ x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

PR #8344, opened 2018-03-09
Sane default temperature limits for Axis cameras

priiduonu:
Default behaviour is to set low limit to (current temperature - 10 deg) and high limit to (current temperature + 20 deg) upon discovery. Changing that to abs values +5 and +35 deg respectively to be in realistic boundaries.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


FTBZ:
There is no default value by default, it's calculated on the first poll, as I know?

priiduonu:
Yes, it seems so, -10 deg and +20 deg from the current value on first poll. And when the current value happens to be near the expected minimum or maximum already, the calculated limits can be misleading. I have a couple of cameras running outdoor right now, with temperature sensors showing +8..9 degrees. I don't like the calculated low limit -2 deg, instead I want to be warned when the temperature gets below +5 (probably means a failing heater). On the other hand, the inside temperature can easily get over the calculated +28 in the summer.

laf:
Thanks for that.

As you are adding a new sysDescr for detection can you also add test data please: https://docs.librenms.org/#Developing/os/Test-Units/#example-workflow

murrant:
I see you've added the minimum required test data in the Other PR. 

I think you should add that here + sensor test data and combine this with #8336.

laf:
I've merged #8336 already. If you can update this PR with the rest of the test data that would be ace.

priiduonu:
The current PR in unrelated to #8336 and is only supposed to set the new default for the temperature sensors of the Axis cameras. I will try to open a new PR.

PR #8345, opened 2018-03-09
libvirt: Fix libvirt VM info

tracernz:
Note: I am a C/C++/Python guy, not a PHP guy, so be gentle/check carefully. I have tested this a bit on my qemu/kvm setup.

Commit message:
```Now we get all VMs (not just running ones), and we properly
display the power state. We now track VMs by UUID(permanent),
rather than temporary ID, which can change as other VMs are
deleted etc. This will reduce event log spam, and obviously
would be required if graphs etc. were added at a later date.

Note: possible VM states in current libvirt master branch:
- "no state",
- "running",
- "idle",
- "paused",
- "in shutdown",
- "shut off",
- "crashed",
- "pmsuspended"
```
Thus I modified the html to only consider "running" as ON, and everything else as OFF. Previously the code was actually wrong, and it displayed everything as ON. The states have not changed in at least 4 years: https://github.com/libvirt/libvirt/blame/master/tools/virsh-domain-monitor.c#L142-L150

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/print-vm.inc.php
+++ b/html/includes/print-vm.inc.php
@@ -10,10 +10,10 @@ if (getidbyname($vm['vmwVmDisplayName'])) {
 
 echo '</td>';
 
-if ($vm['vmwVmState'] == 'powered off') {
-    echo '<td class="list"><span style="min-width:40px; display:inline-block;" class="label label-default">OFF</span></td>';
-} else {
+if ($vm['vmwVmState'] == 'running' || $vm['vmwVmState'] == 'powered on') {
     echo '<td class="list"><span style="min-width:40px; display:inline-block;" class="label label-success">ON</span></td>';
+} else {
+    echo '<td class="list"><span style="min-width:40px; display:inline-block;" class="label label-default">OFF</span></td>';
 }
 
 if ($vm['vmwVmGuestOS'] == 'E: tools not installed') {
--- a/misc/db_schema.yaml
+++ b/misc/db_schema.yaml
@@ -1599,7 +1599,7 @@ vminfo:
     - { Field: id, Type: int(11), 'Null': false, Extra: auto_increment }
     - { Field: device_id, Type: int(11), 'Null': false, Extra: '' }
     - { Field: vm_type, Type: varchar(16), 'Null': false, Extra: '', Default: vmware }
-    - { Field: vmwVmVMID, Type: int(11), 'Null': false, Extra: '' }
+    - { Field: vmwVmVMID, Type: char(36), 'Null': false, Extra: '' }
     - { Field: vmwVmDisplayName, Type: varchar(128), 'Null': false, Extra: '' }
     - { Field: vmwVmGuestOS, Type: varchar(128), 'Null': false, Extra: '' }
     - { Field: vmwVmMemSize, Type: int(11), 'Null': false, Extra: '' }
--- a/includes/discovery/libvirt-vminfo.inc.php
+++ b/includes/discovery/libvirt-vminfo.inc.php
@@ -33,103 +33,89 @@ if (Config::get('enable_libvirt') && $device['os'] == 'linux') {
         if ($ssh_ok || !str_contains($method, 'ssh')) {
             // Fetch virtual machine list
             unset($domlist);
-            exec(Config::get('virsh').' -rc '.$uri.' list', $domlist);
+            exec(Config::get('virsh').' -rc '.$uri.' list --uuid --all', $domlist);
 
             foreach ($domlist as $dom) {
-                list($dom_id,) = explode(' ', trim($dom), 2);
+                $dom_id = trim($dom);
+                $dom_info = array();
 
-                if (is_numeric($dom_id)) {
+                if (strlen($dom_id) == 36) {
                     // Fetch the Virtual Machine information.
                     unset($vm_info_array);
-                    exec(Config::get('virsh').' -rc '.$uri.' dumpxml '.$dom_id, $vm_info_array);
-
-                    // Example xml:
-                    // <domain type='kvm' id='3'>
-                    // <name>moo.example.com</name>
-                    // <uuid>48cf6378-6fd5-4610-0611-63dd4b31cfd6</uuid>
-                    // <memory>1048576</memory>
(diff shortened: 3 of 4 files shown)


tracernz:
Fixed travis failure and force-pushed. I hadn't updated misc/db_schema.yaml... Can that be tested by ./scripts/pre-commit.php?

tracernz:
Ahhh, it occurs to me that the html is shared with at least VMWare SNMP as well... I'll have to figure out what it's states are and create a translation table for libvirt (should be somewhere in libvirt source code since it supports VMWare).

tracernz:
Sadly it's not contrained in libvirt... will have to dig elsewhere 
https://github.com/libvirt/libvirt/blob/6e6faf6d627d5ea291d9ff2378f7d1eb59e1d14e/src/vmware/vmware_driver.c#L281-L288

tracernz:
[Best info I could find on vmware MIB](https://github.com/Opmantek/nmis/blob/02eb6d5f13a84bf98f62e72c406e05a152089b27/models-install/Model-ESXi.nmis#L422-L427) says it's only two states are "powered on" and "powered off": 
I think I'd like to just modify the html to:
```php
if ($vm['vmwVmState'] == 'running' || $vm['vmwVmState'] == 'powered on') {
```
and keep the richer state info provided by libvirt. At least it will still show in the event log. Any objections? Alternatives would be to partially revert b7ac4932a45e864f24a050e19d3fc66c74573968, or add some additional states/colours to the html.

tracernz:
Made the change mentioned in previous message. Should be good to go now.

Future work related to libvirt:
- Support [libvirt-snmp](https://github.com/libvirt/libvirt-snmp), work needed on libvirt-snmp itself:
    - Needs a patch to build with recent net-snmp (I will try to submit upstream)
    - Should patch placeholder org id (12345) with real libvirt org id (36957)
    - Maintaining an aur package for arch users https://aur.archlinux.org/packages/libvirt-snmp-git/

html/includes/print-vm.inc.php
```diff
@@ -10,10 +10,10 @@
 
 echo '</td>';
 
-if ($vm['vmwVmState'] == 'powered off') {
-    echo '<td class="list"><span style="min-width:40px; display:inline-block;" class="label label-default">OFF</span></td>';
-} else {
+if ($vm['vmwVmState'] == 'running' || $vm['vmwVmState'] == 'powered on') {
```
laf: How do we know that this won't affect vmware based virtual machines?

I don't have any of either to test to be sure.

misc/db_schema.yaml
```diff
@@ -1599,7 +1599,7 @@ vminfo:
     - { Field: id, Type: int(11), 'Null': false, Extra: auto_increment }
     - { Field: device_id, Type: int(11), 'Null': false, Extra: '' }
     - { Field: vm_type, Type: varchar(16), 'Null': false, Extra: '', Default: vmware }
-    - { Field: vmwVmVMID, Type: int(11), 'Null': false, Extra: '' }
+    - { Field: vmwVmVMID, Type: char(36), 'Null': false, Extra: '' }
```
laf: Why is this changing? Again vmware seems to expect a numeric ID which was the same for libvirt before.

tracernz: _only_ running VMs even have an integer ID with libvirt. Even worse, the ID is ephemeral. Here is what happens when I start and stop a VM on a real machine:
```
$ virsh -c 'qemu:///system' list --all
 Id    Name                           State
----------------------------------------------------
 --snip--
 16    test                           running
 -     alarm-aarch64                  shut off

$ virsh -c 'qemu:///system' shutdown test
Domain test is being shutdown

$ virsh -c 'qemu:///system' list --all 
 Id    Name                           State
----------------------------------------------------
--snip--
 -     alarm-aarch64                  shut off
 -     test                           shut off

$ virsh -c 'qemu:///system' start test
Domain test started

$ virsh -c 'qemu:///system' list --all
 Id    Name                           State
----------------------------------------------------
--snip--
 17    test                           running
 -     alarm-aarch64                  shut off
```
As you can see the ID changed, which would cause a lot of garbage events in librenms currently. You are correct that this will affect VMWare, my mistake. I wonder if I need to make a new column for the UUID?

html/includes/print-vm.inc.php
tracernz: As far as I can tell VMWare returns either 'powered on' or 'powered off'. I'll spin up esxi in a VM and do some testing once I've made some changes to this PR.

tracernz:
> Other than the web ui that needed to be fixed, what's the reason for the other changes?

As above, + one command rather than 3 for efficiency's sake.

> You should also submit test data for this so we can be sure future changes won't be affected. If we can find someone with vmware running then we can get some data from them and test against this.

Good idea. I'll spin up an esxi VM and get some data.

murrant:
You'll need to add this to `tests/module_tables.yaml` to get it to capture the vminfo table.
```yaml
libvirt-vminfo:
    vminfo:
        excluded_fields: [device_id]
        order_by: vmwVmVMID
```

tracernz:
@murrant Thanks for the tip!

I've set up a test setup with libvirt+qemu on my desktop machine + esxi running as a guest (with nested virtualization enabled). I'm rewriting this branch in smaller commits. Will be ready later this week.

laf:
@tracernz Any update on this?

tracernz:
@laf Sorry, this slid off my radar a bit. I do have a WIP branch at https://github.com/tracernz/librenms/commits/mrc-libvirt2 awaiting some mechanism to supply test data for shell commands (`virsh` tool is used to gather the libvirt vm info). Ideally I would like to build something that works much the same as the snmp test data, except for shell commands (just static stdout data saved to a file basically).

I also made some patches so libvirt-snmp can compile against current net-snmp package... so SNMP may be an option for libvirt if distro packagers were to pick it up https://github.com/libvirt/libvirt-snmp/commits/master... that seems unlikely though sadly.

laf:
Closing this PR down for now. Please feel free to push more commits and let us know when it's ready.

PR #8351, opened 2018-03-09
Adding alerts.id to DescribeAlert() json

pozar:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/alerts.inc.php
+++ b/includes/alerts.inc.php
@@ -459,6 +459,7 @@ function RunJail($code, $obj)
  */
 function DescribeAlert($alert)
 {
+    global $alerts_id;
     $obj         = array();
     $i           = 0;
     $device      = dbFetchRow('SELECT hostname, sysName, sysDescr, hardware, version, location, purpose, notes, uptime FROM devices WHERE device_id = ?', array($alert['device_id']));
@@ -479,6 +480,7 @@ function DescribeAlert($alert)
     $obj['description']  = $device['purpose'];
     $obj['notes']        = $device['notes'];
     $obj['device_id']    = $alert['device_id'];
+    $obj['alerts_id']    = $alerts_id;
     $extra               = $alert['details'];
     if (!isset($tpl['template'])) {
         $obj['template'] = $default_tpl;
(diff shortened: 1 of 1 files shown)


includes/alerts.inc.php
```diff
@@ -459,6 +459,7 @@ function RunJail($code, $obj)
  */
 function DescribeAlert($alert)
 {
+    global $alerts_id;
```
murrant: Don't use a global, they are too unpredictable.  Load the id from the database instead.

includes/alerts.inc.php
```diff
@@ -479,6 +480,7 @@ function DescribeAlert($alert)
     $obj['description']  = $device['purpose'];
     $obj['notes']        = $device['notes'];
     $obj['device_id']    = $alert['device_id'];
+    $obj['alerts_id']    = $alerts_id;
```
murrant: The correct name would be "alert_id" instead of "alerts_id"

includes/alerts.inc.php
pozar: K. At this point in the code, I can't get that as I don't see a way to index back to alerts.id.  I can get that in RunAlerts() so I need to stuff that somewhere to use in DescribeAlert().  Would there be a better way to do this?

includes/alerts.inc.php
pozar: No problem changing that object name.  There seems to be this namespace reuse for "alert" where it sometimes gets used for rows in the alerts table and sometimes gets used for rows the alert_log table.  I was using alerts to try to avoid that.  Will change.

murrant:
The other one should probably be alert_log_id.

includes/alerts.inc.php
murrant: Why can't you store the id in the database? (adding a new field)

Also, looking below if you have multiple alerts, they will all use the alert_id of the last alert...

murrant:
I was looking at this and ended up with #8368

$alert is passed all the way through, so you don't need the global.
But I would have had to add it in 3 places to be consistent, so I consolidated things.

laf:
Thanks for taking a look a look at this. I’ve merged the other PR in place as it cleans things up a little. Please update and see if this works for you.

pozar:
Thanks gang.  Will loop back with Opsgenie to see that this is working.

PR #8363, opened 2018-03-13
Auto IPv6/IPv4 detection when adding host

murrant:
Fallback to ipv4 if snmpd isn't listening on ipv6 (as our example configures)
Update snmpd.conf example to listen on IPv6
refactor dnslookup() supports auto (ipv6 then ipv4) and hosts file for ipv4 at least.
Improve layout in the device snmp settings page

Tested adding:
ipv4 only
dual ipv4 listen only
dual dual listen

image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/addhost.php
+++ b/addhost.php
@@ -73,7 +73,7 @@ if (!empty($argv[1])) {
     $snmpver   = strtolower($argv[3]);
 
     $port      = 161;
-    $transport = 'udp';
+    $transport = 'auto';
 
     $additional = array();
     if (isset($options['b'])) {
--- a/doc/Support/SNMP-Configuration-Examples.md
+++ b/doc/Support/SNMP-Configuration-Examples.md
@@ -169,7 +169,12 @@ vi /etc/snmp/snmpd.conf
 
 ```
 # Change RANDOMSTRINGGOESHERE to your preferred SNMP community string
-com2sec readonly  default         RANDOMSTRINGGOESHERE
+com2sec  readonly  default      RANDOMSTRINGGOESHERE
+com2sec6 readonly  default      RANDOMSTRINGGOESHERE
+
+# listen on IPv4 and IPv6
+agentaddress udp:161
+agentaddress udp6:161
 
 group MyROGroup v2c        readonly
 view all    included  .1                               80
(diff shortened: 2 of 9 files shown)


laf:
This changes the default behaviour. Doing nothing more than checking this branch out and trying to add a host via the webui I get:

`Already have host with this IP localhost (127.0.0.1)`

murrant:
@laf good catch, I missed adding the boolean in the if statement when re-arranging.

murrant:
Testing this I discovered something.
When adding a device with dual stack, you can add it via both IPv4 and IPv6 because we only have one "IP" field...

murrant:
Maybe that only works for localhost since there is not an entry in the ipv6_addresses table...

laf:
So here's the next one :)

Trying to add www.google.com as icmp only device and it tries to use fping6 as the binary :) Adding www.intel.com and it works fine as it tries fping.

murrant:
That sounds correct to me, did you expect a different result?

laf:
Yes as I don't have a working ipv6 setup so it failed.

murrant:
Bahhhh, I'm guessing because we are doing a direct DNS lookup, we can't detect if IPv6 is functioning.

I really wish PHPs gethostbyname() supported IPv6

MrMaus13:
Something like this?

```
function check_ipv6_reachable($hostname) {
  $record = dns_get_record($hostname, DNS_AAAA);
  if(IsSet($record[0]["ipv6"])) return exec("/usr/sbin/fping6 -a ".$record[0]["ipv6"]);
}

if(check_ipv6_reachable("www.google.com")) echo "IPV6 Works!";

```

murrant:
unfortunately, we can't guarantee everything is pingable.  Aka some host blocks icmp echos but responds to snmp.

MrMaus13:
Isn't it possible to use isSNMPable for that then?

```
function check_ipv6_reachable($hostname) {
  $record = dns_get_record($hostname, DNS_AAAA);
  if(IsSet($record[0]["ipv6"])) return isSNMPable($record[0]["ipv6"]);
}

if(check_ipv6_reachable("www.google.com")) echo "IPV6 SNMP Works!";
```

I'm just beginning to understand the code, apologies in advance if I'm saying something stupid. ;-)

laf:
@murrant Can we close this or do you want to keep it open to come back to it?

murrant:
Yeah, closing this.  Can't come up with a good solution.

PR #8364, opened 2018-03-13
Allow full search on devices page

MrMaus13:
Allows you to search all columns on the devices overview page instead of only the 'hostname' column.
Useful for searching locations, system names, ip addresses, hardware types, etc.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/table/ports.inc.php
+++ b/html/includes/table/ports.inc.php
@@ -39,7 +39,7 @@ if (is_admin() === false && is_read() === false) {
 $sql .= ' LEFT JOIN `devices` AS `D` ON `ports`.`device_id` = `D`.`device_id`';
 
 if (!empty($vars['hostname'])) {
- $where .= ' AND (D.hostname LIKE "%i'.$vars['hostname'].'%" OR D.sysName LIKE "%'.$vars['hostname'].'%")';
+    $where .= ' AND (D.hostname LIKE "%i'.$vars['hostname'].'%" OR D.sysName LIKE "%'.$vars['hostname'].'%")';
 }
 
 if (!empty($vars['location'])) {
--- a/html/includes/table/ports.inc.php
+++ b/html/includes/table/ports.inc.php
@@ -40,7 +40,7 @@ $sql .= ' LEFT JOIN `devices` AS `D` ON `ports`.`device_id` = `D`.`device_id`';
 
 if (!empty($vars['hostname'])) {
     $where .= ' AND (D.hostname LIKE ? OR D.sysName LIKE ?)';
-    $param += array_fill(count($param),2,'%' . $vars['hostname'] . '%');
+    $param += array_fill(count($param), 2, '%' . $vars['hostname'] . '%');
 }
 
 if (!empty($vars['location'])) {
--- a/html/includes/table/ports.inc.php
+++ b/html/includes/table/ports.inc.php
@@ -40,7 +40,7 @@ $sql .= ' LEFT JOIN `devices` AS `D` ON `ports`.`device_id` = `D`.`device_id`';
 
 if (!empty($vars['hostname'])) {
     $where .= ' AND (D.hostname LIKE ? OR D.sysName LIKE ?)';
-    for($i=0;$i<2;$i++) $param[] = '%' . $vars['hostname'] . '%';
+    $param += array_fill(count($param),2,'%' . $vars['hostname'] . '%');
 }
 
 if (!empty($vars['location'])) {
--- a/html/includes/table/ports.inc.php
+++ b/html/includes/table/ports.inc.php
@@ -38,29 +38,28 @@ if (is_admin() === false && is_read() === false) {
 
 $sql .= ' LEFT JOIN `devices` AS `D` ON `ports`.`device_id` = `D`.`device_id`';
 
-if (!empty($_POST['hostname'])) {
-    $where .= ' AND `D`.`hostname` LIKE ?';
-    $param[] = '%' . $_POST['hostname'] . '%';
+if (!empty($vars['hostname'])) {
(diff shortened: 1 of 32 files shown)


dagbdagb:
is there any kind of library which allows for an intelligen search a la:

os:linux and host:foobar and not location:London*

and so on....

enilfodne:
Is there any policy with regards to code safety? Because the commit (and the files it modifies) are almost perfect examples for SQL injections.

laf:
Should just need `_POST` replacing with `vars` to be safe.

MrMaus13:
Ok, I changed it to vars, which seems to work fine.
How do 'vars' work? As a don't see them available anywhere?
Anyway, should i change all $_POST references with vars in the "devices.inc.php" file? Since they are really used all over the place.

laf:
So `$vars` is what we set for data processed via `_POST` and `_GET` but we run some sanity checks over it. If you can fix all _POST use that would be ace.

MrMaus13:
Ok, done. There are a lot of files with this issue as well. I'll take a look at them when i have some more time to spare.

murrant:
I disagree with using $vars, I would prefer $_REQUEST if we need to allow both post and get.

I know that is what a lot of the code is now, but PHP already does this for us.

MrMaus13:
That would defy the point and still leave you with insecure and injectable code; $_REQUEST is just as insecure as $_POST and $_GET:

From [PHP Manual]( http://php.net/manual/en/reserved.variables.request.php):
> Note:
> The variables in $_REQUEST are provided to the script via the GET, POST, and COOKIE input mechanisms and therefore could be modified by the remote user and cannot be trusted. The presence and order of variables listed in this array is defined according to the PHP variables_order configuration directive.

 
The reason for using $vars is not because we have to allow both post and get, it's because the input is sanitized in html/includes/vars.inc.php.

Makes it harder for a guest to drop a table using sql injection. There is a lot more code not checking on sanity of the input but for my change it made sense to at least change the files in the html/includes/table directory to be more secure.

murrant:
Good point.  I'm used to sanitizing the input at a different level.  Proceed.

html/includes/table/ports.inc.php
```diff
@@ -38,29 +38,28 @@
 
 $sql .= ' LEFT JOIN `devices` AS `D` ON `ports`.`device_id` = `D`.`device_id`';
 
-if (!empty($_POST['hostname'])) {
-    $where .= ' AND `D`.`hostname` LIKE ?';
-    $param[] = '%' . $_POST['hostname'] . '%';
+if (!empty($vars['hostname'])) {
+    $where .= ' AND (D.hostname LIKE "%i'.$vars['hostname'].'%" OR D.sysName LIKE "%'.$vars['hostname'].'%")';
```
laf: You need to put this back to `$where` and `$param`

MrMaus13: Ok I'll fix that;
Do you want the same for devices then? 
Because it's (almost) the same query.

laf: Yeah I saw that, you can fix if you like but it's not necessary. Just don't regress the code :)

murrant:
I wonder how this will perform with many devices? (say 10,000)

MrMaus13:
This one takes about 0.12sec, if you have any other queries to try, let me know.

```
MariaDB [librenms]> select count(*) from devices;
+----------+
| count(*) |
+----------+
|    51174 |
+----------+
1 row in set (0.01 sec)

MariaDB [librenms]> SELECT os FROM devices WHERE hostname LIKE "%rtm-vnn%" OR sysName LIKE "%rtm-vnn%" OR hardware LIKE "%rtm-vnn%" OR os LIKE "%rtm-vnn%" OR location like "%rtm-vnn%";
+---------+
| os      |
+---------+
| iosxe   |
| ciscosb |
| ios     |
| ios     |
| ios     |
| ios     |
| ios     |
| ios     |
| ios     |
| ios     |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ciscosb |
| ios     |
| ios     |
| ciscosb |
+---------+
36 rows in set (0.16 sec)

MariaDB [librenms]> 
```

dagbdagb:
This commit removes the ability to only see graphs for a subset of devices. 

For instance, I like to search for 'esx' when looking at servers, and then hit 'Memory' to illustrate how the ESXes are doing in that department. Prior to this commit, I would get graphs for all hosts with 'esx' in the hostname. After the commit I get graphs for all devices of type server.

laf:
@MrMaus13 I think it's because https://github.com/librenms/librenms/blob/a74086d82757a8ae1b4242ba7f2c4b28eab0d96c/html/pages/devices.inc.php#L94 section needs updating to use the new searchquery param

MrMaus13:
Right, I see, never thought of that.
I'll fix that this week, sorry @dagbdagb  :-)

dagbdagb:
No worries. 
Move fast, break often, break early.

laf:
Thanks @MrMaus13

MrMaus13:
Should be fixed: https://github.com/librenms/librenms/pull/8509

laf:
@MrMaus13 Thanks for fixing :)

@dagbdagb update and all should be well.

dagbdagb:
Looks good! Thanks!

PR #8371, opened 2018-03-13
Sane default temperature limits for Axis cameras

priiduonu:
Default behaviour is to set low limit to (current temperature - 10 deg) and high limit to (current temperature + 20 deg) upon discovery. Changing that to abs values +5 and +35 deg respectively to be in realistic boundaries.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Looks good, thanks :)

PR #8388, opened 2018-03-15
Updating the mibs file from vmware

mikeSimonson:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
You've added a whole heap of new MIBs which are already in the mibs folder. Please remove those from the vmware folder and only have vmware specific mibs there.

tracernz:
Can you put the source these were retrieved from etc. in the commit message please? It would help a bit for anyone trying to look at how they came to be in future.

mikeSimonson:
How do you spot those that are vmware specific ?
Those only come from the vmware archive.

tracernz:
> How do you spot those that are vmware specific ?

VMWARE-*

> Those only come from the vmware archive.

Yeah, they included all the generic MIBs their products use too.

murrant:
You can check this by running: 
`composer install` (to install development dependencies
`./vendor/bin/phpunit tests/MibTest.php --group mibs --filter=vmware` to check for problems in the vmware mib directory.

PR #8392, opened 2018-03-15
Snmpsim fail when it fails

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/Util/Snmpsim.php
```diff
@@ -26,6 +26,7 @@
 namespace LibreNMS\Util;
 
 use LibreNMS\Config;
+use LibreNMS\Exceptions\SnmpsimException;
```
laf: May as well remove all of these.

scripts/collect-snmp-data.php
```diff
@@ -27,8 +28,12 @@
 );
 
 if (isset($options['snmpsim'])) {
-    $snmpsim = new Snmpsim();
-    $snmpsim->run();
+    try {
+        $snmpsim = new Snmpsim();
+        $snmpsim->run();
+    } catch (SnmpsimException $e) {
```
laf: This was removed wasn't it?

murrant: Thought I removed all of them, haven't had a chance to double check.  Thanks.

Issue #8031, opened 2018-01-05
Updating ServerTech PDU device support 


papatango71:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

librenms@nms01-eqch2
[/opt/librenms]$ ./validate.php
====================================
Component | Version
--------- | -------
LibreNMS  | 1.33-304-ge16422a
DB Schema | 227
PHP       | 7.0.22-0ubuntu0.16.04.1
MySQL     | 10.0.31-MariaDB-0ubuntu0.16.04.2
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3
====================================

[OK]    Database connection successful
[OK]    Database schema correct

DISCOVERY :: https://p.libren.ms/view/32314270
>> Error'ed out 
  Paste endpoint: https://p.libren.ms/api/create
  ifStackStatus: Unknown Object Identifier (Sub-id not found: (top) -> ifStackStatus)
  MIB search path: /opt/librenms/mibs:/opt/librenms/mibs/sentry
  Cannot find module (CPQHOST-MIB): At line 5 in (none)
  cpqHoFileSysEntry: Unknown Object Identifier (Sub-id not found: (top) -> cpqHoFileSysEntry)

POLLER :: https://p.libren.ms/view/9f0e3607

SNMPBULKWALK :: https://p.libren.ms/view/aca5e8a3

Senrty3.mib :: https://p.libren.ms/view/6652bb1a
Sentry3OIDTree.txt :: https://p.libren.ms/view/8bf2fe49

[...]

PR #8407, opened 2018-03-16
feature: Improve Sentry4 discovery

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Should address #8031.

I have not processed:
st4Outlets
st4TemperatureSensors
st4HumiditySensors
st4WaterSensors
st4ContactClosureSensors
st4AnalogToDigitalConvSensors
st4FanSensors
st4EventInformation

I have not added test data either, as the data I tested was from one of the issues, hence not mine.

Indexes will change for frequency and power, so data will be lost.


laf:
We can just set the new index to be the same as the old, better to do that and preserve historical data than just wipe it out.

laf:
This is now ready I'd like someone else to take a look as it changes the way user_functions are called. Before we processed them before we ran divisor/multiplier which would then cause bad data.

For example, Fahrenheit to celsius where the value also needed to be divided by first resulted in the wrong values.

So, actually run the function once we have the correct known value.

PR #8411, opened 2018-03-17
nimbleos Changed to bit shift high << 32 and add low. Changed over graphs.

theherodied:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/nimbleos.yaml
+++ b/includes/definitions/nimbleos.yaml
@@ -4,8 +4,7 @@ type: storage
 icon: nimble
 over:
     - { graph: device_bits, text: 'Device Traffic' }
-    - { graph: device_processor, text: 'Processor Usage' }
-    - { graph: device_mempool, text: 'Memory Usage' }
+    - { graph: device_storage, text: 'Storage Usage' }
 mib_dir:
     - nimble
 discovery:
--- a/includes/discovery/storage/nimbleos.inc.php
+++ b/includes/discovery/storage/nimbleos.inc.php
@@ -19,7 +19,7 @@
  *
  * @package    LibreNMS
  * @link       http://librenms.org
- * @copyright  2018 theherodied
+ * @copyright  2018 Ryan Finney
  * @author     https://github.com/theherodied/
  */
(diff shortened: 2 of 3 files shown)


murrant:
Can you explain this, why would you want to concatenate the two numbers?

theherodied:
When I read the MIB file and saw the High Order Bytes and Low Order Bytes I began researching what they meant. I've seen several different takes on it but this is the one I'm tending to believe. I opened a ticket with Nimble to clarify/confirm.

https://www.zenoss.com/product/zenpacks/nimble
Jane Curry described it like this:
A modeler plugin - NimbleVolumeMap - which gathers id, name, various sizes, online status and number of connections. The three sets of size data are in 2 MIB values - high 32 bits and low 32 bits. The modeler combines these values to present a single value in the NimbleVolume object. We believe that the raw SNMP data is in MB. The modeler divides this by 1024 to present values as GB.

The code Jane used looks like this and can be found at https://github.com/jcurry/ZenPacks.community.Nimble/blob/master/ZenPacks/community/Nimble/modeler/plugins/NimbleVolumeMap.py

# MIB sez vol sizes are bytes.  They are actually MB. We want GB
om.VolSize = (( data['_VolSizeHigh'] << 32 )+ data['_VolSizeLow']) / 1024
om.VolUsage = (( data['_VolUsageHigh'] << 32 )+ data['_VolUsageLow']) / 1024
om.VolReserve = (( data['_VolReserveHigh'] << 32 ) + data['_VolReserveLow']) / 1024

I'm unsure if the 32 bitwise shift is needed.

theherodied:
This might provide some insight. I might need to change the code still.

theherodied:
I actually need to shift the high counter << 32 and add the low. I'll work on that.

theherodied:
I think this is correct now.

murrant:
Also, I **think** it will work like the existing code on 32bit platforms when you `<< 32` the number it should be 0 so, you will only use lowbits.

theherodied:
Is this all set to merge now?

PR #8414, opened 2018-03-19
Print mysql errors in debug output.

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/dbFacile.php
```diff
@@ -117,7 +117,9 @@ function dbQuery($sql, $parameters = array())
         $mysql_error = mysqli_error($database_link);
         if (isset($config['mysql_log_level']) && ((in_array($config['mysql_log_level'], array('INFO', 'ERROR')) && !preg_match('/Duplicate entry/', $mysql_error)) || in_array($config['mysql_log_level'], array('DEBUG')))) {
             if (!empty($mysql_error)) {
-                logfile(date($config['dateformat']['compact']) . " MySQL Error: $mysql_error ($fullSql)");
+                $error_msg =  "MySQL Error: $mysql_error ($fullSql)";
+                c_echo("%R$error_msg%n\n", $debug);
```
laf: Maybe drop the `$debug`? This is the kind of thing we want to see regardless of debug being enabled or not.

murrant: I'm fine with always printing this.  We'll get more reports of broken queries.

Issue #8163, opened 2018-01-29
Health State/Information Divisor value not being applied to low/high limits


mattie47:
I've been adding sensor/health support under https://github.com/librenms/librenms/pull/8140

I noticed that with the following:

```yaml
mib: AT-ENVMONv2-MIB
modules:
    sensors:
        voltage:
            data:
                -
                    oid: atEnvMonv2VoltageTable
                    value: atEnvMonv2VoltageCurrent
                    num_oid: .1.3.6.1.4.1.207.8.4.4.3.12.2.1.5.
                    descr: atEnvMonv2VoltageDescription
                    index: 'atEnvMonv2VoltageEntry.{{ $index }}'
                    divisor: 1000
                    low_limit: atEnvMonv2VoltageLowerThreshold
                    high_limit: atEnvMonv2VoltageUpperThreshold
```

The divisor of 1000 is not being applied to low_limit, or high_limit. 

My assumption, is that this is a bug however the workaround for me has been to simply hash out these two lines for now...

DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ ] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ ] Please provide as much detail as possible.
- [ ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

PR #8427, opened 2018-03-20
fix: Apply divisor / multiplier to high/low limits in sensors

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #8163 

If a numeric value is provided for _limits then it's used as is. If not, we apply the divisor / multiplier to those limits. I've been through all the yaml files and the only one which will have an issue with this is the one I've changed to hard coded limits. The devices value requires a divisor but the _limits don't.

Uncommented the _limit values from the yaml that the issue references as it was commented out due to this.

This will fix some high/low limits for yaml files where it should have been applied to but isn't.


murrant:
What do you think about having a different key for limits that don't need a divisor?
For example `raw_high_limit`?

laf:
We could I guess. This one vendor I've changed is the only one I've seen that doesn't support it as you'd expect so I'd prefer to keep things simple until we do actually need it.

murrant:
@laf, you can update all sensors by running `./scripts/save-test-data.php -m sensors`

laf:
Finally!

laf:
Well spotted @murrant should be fixed now.

Issue #8329, opened 2018-03-06
Xerox AltaLink C8000 Series - Toner Colors Incorrect


timconner:
**Problem Description:**
It looks like on these newer Xerox printers that the toner levels are reported on the wrong colors. The Xerox interface rounds the numbers so it's hard to tell exactly which colors switched around, but on this printer it is pretty clear that Black on the Xerox is being reported as Cyan in LibreNMS. Supporting pictures have been attached.

**Reported Levels:**
![reported](https://user-images.githubusercontent.com/22841670/37013305-2dd4bba2-20c7-11e8-890c-13760fb7a923.PNG)

Actual Levels:
![actual](https://user-images.githubusercontent.com/22841670/37013306-3044c530-20c7-11e8-89b6-ebe928d3661f.PNG)

**LibreNMS Version:**
```
====================================
Component | Version
--------- | -------
LibreNMS  | 1.37-21-g6c3473a
DB Schema | 239
PHP       | 7.2.2
MySQL     | 5.5.56-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
====================================
```

**Debug Information:**
Discovery: https://gist.github.com/timconner/36c9ceee830232315e7c81b70dbf358f
Poller: https://gist.github.com/timconner/067be652b54255f008e5c7e49abc0c51
SNMP Walk: https://gist.github.com/timconner/3d370949c496968cd5766ecd162285b5

I wish I could be more help in initial troubleshooting, but I'm a little new to SNMP troubleshooting and haven't been able to locate where the cause of the problem is.

--

DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
[...]

PR #8428, opened 2018-03-20
fix: Fixed and improved toner discovery

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #8329 

It doesn't break the device for which this fix was originally added for and the fix really shouldn't have been done. Working round one vendors bad implementation.

Also dropped walking of unneeded OIDs as these were just indexes so walk data we actually use.


laf:
Any downside to leaving the PN in though @murrant ?

murrant:
@laf SQL fails because the input is too long.

PR #8429, opened 2018-03-20
Move nested defaults to Config::processConfig()

murrant:
Fixes log_file when users set log_dir for example.

Removed stuff:
mibdir
dp_autocreate
base_url: setting http/https before config.php is loaded...
moved collectd setting to it's own config file

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
With ssl enabled (nginx) and no base_url set this breaks the WebUI as it tries to load content over http.

Setting it to the full url or `/` works fine but I fear not everyone will have base_url set (legacy most likely).

PR #8430, opened 2018-03-20
alert: Added support for disabling recovery notifications

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Added support for disabling sending recovery notifications.

Also:

  - blacklisted two more tables from being used in the query builder. No reason people should be alerting on those.
  - Also clear out the 'count' value for an alert so that alerts will be raised if matched again.

I've tested this with two of the exact same rules, one however had recovery alerts disabled. Max 5 alerts, repeated every 1 second.

I received 5 alerts for each, renamed the device so the alerts would clear and received just one recovery email. Then renamed the device back and alerts triggered again through the same sequence.

### DIFF
--- a/LibreNMS/Alerting/QueryBuilderFilter.php
+++ b/LibreNMS/Alerting/QueryBuilderFilter.php
@@ -32,7 +32,9 @@ use Symfony\Component\Yaml\Yaml;
 class QueryBuilderFilter implements \JsonSerializable
 {
     private static $table_blacklist = [
-        'device_group_device'
+        'device_group_device',
+        'alerts',
+        'alert_log',
     ];
 
     private $filter = [];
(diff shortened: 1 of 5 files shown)


laf:
One question will be, call it disable Recovery or Recovery notification and change the no_recovery to just recovery and invert the checks?

murrant:
Think we should add tooltips to all the fields.  It can help a lot to have just a bit of extra context.

Issue Recovery [[ on ]      ] is my vote :)

You can check if recovery is not set to default to old behavior.

laf:
> Issue Recovery [[ on ] ] is my vote :) Or just "Recovery"
> 
> You can check if recovery is not set to default to old behavior.

You mean jus default to if recovery doesn't exist then default to true? I can switch it around if needs be.

laf:
Updated

PR #8433, opened 2018-03-21
alert: Added ability to make notes for acking alerts + record who did so

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

We don't have any user level checks on acks right now so I've not introduced that.

This allows you to set an ack msg for an alert but also update it at any stage.

It also records who ackd the alert to eventlog


laf:
I've moved the schema file to 248

murrant:
During testing.  
1. The UI didn't change to show the alert was acknowledged (probably just refreshing bootgrid would work)
2. It said successfully ack'd even though it failed because I had not updated my schema yet.
3. In the notes, it would be better if it had a timestamp and user like `Sat Jul 23 02:16:57 2005 - Ack (murrant) ISP is down`
4. Maybe change the notes icon to indicate if there is no note yet.
5. A modal might look nicer, but I'm not fussed.
6. Also, I tend to prefer button groups I think instead of individual columns.
7. Not related, but Ack is yellow?  Odd since you can set alerts to warning, which is yellow...

murrant:
Light blue is acknowledged where I come from :-)

laf:
image

image

How's all that now?

murrant:
I think two tweaks and this is good. If you agree with the behavior, I've attached a patch.
1. append to notes instead of overwritting when acking
2. delete notes when the alert clears

```diff
diff --git a/html/includes/forms/ack-alert.inc.php b/html/includes/forms/ack-alert.inc.php
index 4eb94c43c..2e7c60f98 100644
--- a/html/includes/forms/ack-alert.inc.php
+++ b/html/includes/forms/ack-alert.inc.php
@@ -48,7 +48,11 @@ if (!is_numeric($alert_id)) {
 
     $data = ['state' => $state, 'open' => $open];
     if (!empty($ack_msg)) {
-        $data['note'] = date(Config::get('dateformat.long')) . " - Ack ({$_SESSION['username']}) $ack_msg";
+        $note = dbFetchCell('SELECT note FROM alerts WHERE id=?', [$alert_id]);
+        if (!empty($note)) {
+            $note .= PHP_EOL;
+        }
+        $data['note'] = $note . date(Config::get('dateformat.long')) . " - Ack ({$_SESSION['username']}) $ack_msg";
     }
 
     if (dbUpdate($data, 'alerts', 'id=?', array($alert_id)) >= 0) {
diff --git a/includes/alerts.inc.php b/includes/alerts.inc.php
index 3427f32e4..2d34c7d5f 100644
--- a/includes/alerts.inc.php
+++ b/includes/alerts.inc.php
@@ -230,7 +230,7 @@ function RunRules($device_id)
                 c_echo('Status: %bNOCHG');
             } else {
                 if (dbInsert(array('state' => 0, 'device_id' => $device_id, 'rule_id' => $rule['id']), 'alert_log')) {
-                    if (!dbUpdate(array('state' => 0, 'open' => 1), 'alerts', 'device_id = ? && rule_id = ?', array($device_id,$rule['id']))) {
+                    if (!dbUpdate(array('state' => 0, 'open' => 1, 'note' => ''), 'alerts', 'device_id = ? && rule_id = ?', array($device_id,$rule['id']))) {
                         dbInsert(array('state' => 0, 'device_id' => $device_id, 'rule_id' => $rule['id'], 'open' => 1, 'alerted' => 0), 'alerts');
                     }
                     c_echo(PHP_EOL . 'Status: %gOK');
```

laf:
Doesn't bother me too much either way so I've added it in.

PR #8440, opened 2018-03-21
Documentation updates for DMI support and making two links in the FAQ work

trs80:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
I feel like those line should be commented out by default.

Also, you can remove the numeric oids from the extends...

laf:
> I feel like those line should be commented out by default.
Definitely. They don't all work by default but the error will be returned and used then show in the webui.

laf:
@trs80 Any response to the previous comments?

murrant:
@trs80 looks like you need to sign/resign the CLA.  Thanks.

PR #8441, opened 2018-03-21
fix: Restored config items in sql queries

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

These have to exist in the database for it to work (I'd like to resolve that longer term).

I've also added them all into a new schema file, this will show as errors from build-base but means anyone who's installed in the last week will be fixed.


murrant:
Ah, I see what happened.  The query was really long and included non-macro config settings.

PR #8443, opened 2018-03-21
fix: Restored config items in sql queries

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Shouldn't matter for those who have got the config as it just won't re-add.


murrant:
@laf why were these restored?  I removed them for multiple reasons.
Might be missing from database (so we don't pull them from db).
Allow us to update them without database query.

laf:
@murrant I just reverted all that was removed because it broke most of the webui config support.

laf:
Absolutely.

Issue #5619, opened 2017-01-26
poller service starting multiple polling process for same machine


boudreau:
I'm using distribution pollers and starting using the poller service instead of crontab.

The poller service seem to be doing pretty good, but I have often multiple polleing process started with the same device ID to poll:

librenms 12553 59796  0 08:32 ?        00:00:00 /bin/sh -c /usr/bin/env php /var/www/librenms/poller.php -h 677 >> /dev/null 2>&1
librenms 13051 59796  0 08:32 ?        00:00:00 /bin/sh -c  /usr/bin/env php /var/www/librenms/poller.php -h 677 >> /dev/null 2>&1
librenms 13263 59796  0 08:32 ?        00:00:00 /bin/sh -c  /usr/bin/env php /var/www/librenms/poller.php -h 677 >> /dev/null 2>&1

This happens on the main poller/DB/web site as well as the other pollers.

It should not start multiple poller for the same device since there should be a DB lock on the device. Could there be a multiple queries before the lock is active in the DB?
How can I solve this?

Thanks for the help

Component | Version
--------- | -------
LibreNMS  | 08edfc60e06a1f79d0d91d5dd89a95f61d3a8945
DB Schema | 157
PHP       | 7.0.14
MySQL     | 5.5.50-MariaDB
RRDTool   | 1.5.6
SNMP      | NET-SNMP 5.7.2

DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for feature requests or bugs, please do not post issues asking for help or how to do X, Y or Z. 
You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Unless your issue is for a WebUI fix or feature then please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

laf:
@clinta any input on this?

clinta:
Can you provide a log from the poller service?

pollix:
[...]

PR #8450, opened 2018-03-22
[RFC] fix: poller-service locking does not work

TheMysteriousX:
Needs another look once I'm a little more awake, but this could use some testing and feedback.

I have no idea what's up with the original locking, other than it doesn't work. I found #5619, which didn't include a solution either, so I've ripped out the database locking, and replaced it with thread level locking.

It's not a fantastic fix due to the overall structure of the poller service, but it works for me.

After applying this, CPU load has dropped on my less capable switches from between 60-80%, down to ~10%

screenshot 2018-03-22 00 41 56

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/poller-service.py
+++ b/poller-service.py
@@ -48,6 +48,10 @@ config_file = install_dir + '/config.php'
 
 log.info('INFO: Starting poller-service')
 
+locklist = {}
+locklistlock = threading.Lock()
+
+dupedetect = {}
 
 class DB:
     conn = None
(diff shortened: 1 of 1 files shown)


laf:
I'm assuming this makes it non-distributed aware if the locking is done local.

If so, I doubt we can merge this in because it would break distributed poller setups.

TheMysteriousX:
I'll admit I didn't consider distributed polling; I could hide the alternative locking behaviour behind a toggle.

laf:
That could be an option.

@murrant had a PR in that replaced this script but we never merged it (I can't find the PR right now). Overall it replaced cron for everything so it would be better to focus on that if you fancy helping. It had an issue with auto updating but aside from that, I think it worked extremely well.

murrant:
I sent it as a PR here #8455

Still has the file handle leak as I haven't touched it.

murrant:
@TheMysteriousX Original locking worked with older versions of MySQL. But the server behavior changed at some point.

adam-bishop:
@murrant 's is the much better option - I didn't realise a rewrite was already in progress.

Re: the file descriptor leak, it's something I've seen before - if it's what I think it is, the cause FD's being inherited across process executions, even when calling exec*().

There's a bug in the python tracker on it: https://bugs.python.org/issue12107

Solution seems to be use Python3.4+, or don't rely on file descriptors auto closing. I suspect though it'll be in the MySQL or redis drivers, which makes fixing it impossible without patching upstream).

TheMysteriousX:
(signed in on the right account this time)

@murrant 's is the much better option - I didn't realise a rewrite was already in progress.

Re: the file descriptor leak, it's something I've seen before - if it's what I think it is, the cause FD's being inherited across process executions, even when calling exec*().

There's a bug in the python tracker on it: https://bugs.python.org/issue12107

Solution seems to be use Python3.4+, or don't rely on file descriptors auto closing. I suspect though it'll be in the MySQL or redis drivers, which makes fixing it impossible without patching upstream).

murrant:
@TheMysteriousX yeah, it has to do with the redis driver

murrant:
Closing this, continued at #8455

PR #8455, opened 2018-03-22
New python service

murrant:
Currently has a file handle leak (and will eventually run out of handles) related to the self update process.

Either need to fix that or rip out self-update and leave that up to cron or something.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


TheMysteriousX:
I got this running under python3.6; no changes really needed, it merged pretty cleanly. I'll add notes to the relevant lines, and keep an eye on the FD state.

discovery.php
```diff
@@ -123,7 +123,8 @@
 }
 
 global $device;
-foreach (dbFetch("SELECT * FROM `devices` WHERE status = 1 AND disabled = 0 $where ORDER BY device_id DESC", $sqlparams) as $device) {
+$devices = dbFetch("SELECT * FROM `devices` WHERE status = 1 AND disabled = 0 $where ORDER BY device_id DESC", $sqlparams);
```
TheMysteriousX: Query has changed slightly, but thats all.

config_to_json.php
```diff
@@ -7,8 +7,14 @@
  *
  */
 
-$init_modules = array();
+$init_modules = array('nodb');
 require __DIR__ . '/includes/init.php';
+try {
```
TheMysteriousX: Removed this section entirely mergedb was unable to be imported. Does it do anything?

LibreNMS/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+        """
+        self.config = config
+        self._db = None
+
+        if auto_connect:
+            self.connect()
+
+    def connect(self):
+        try:
+            import MySQLdb
```
TheMysteriousX: Replaced with this to avoid having to install the MySQL developer kit and a C compiler:

```
        try:
            import pymysql
            MySQLdb = pymysql
        except ImportError:
            try:
                import MySQLdb

            except ImportError:
                print("ERROR: missing the mysql python module:")
                print("Either through your os software repository or 'pip install mysqlclient'")
                raise
```

LibreNMS/service.py
```diff
@@ -0,0 +1,516 @@
...
+        for device in devices:
+            self.discovery_manager.post_work(device[0], device[1])
+
+    def discover_device(self, device_id):
+        if self.lock_discovery(device_id):
+            try:
+                info("Discovering device {}".format(device_id))
+                self.call_script('discovery.php', ('-h', device_id))
+                info('Discovery complete {}'.format(device_id))
+            except subprocess.CalledProcessError as e:
```
TheMysteriousX: This doesn't appear to be firing - ran out of time to pin down the root cause, but it seems that discovery.php is returning 0 when failing to discover under some circumstance:

```
[root@sdc-mon-001 librenms]# ./discovery.php -d -h 118
LibreNMS Discovery
===================================
Version info:
Commit SHA: faf48311b2c919d9bdde49220536cd0cb95f7930
Commit Date: 1521744011
DB Schema: 247
PHP: 7.1.14
MySQL: 5.5.56-MariaDB
RRDTool: 1.4.8
SNMP: NET-SNMP 5.7.2
==================================DEBUG!
Updating os_def.cache... Done
SQL[SELECT * FROM `devices` WHERE disabled = 0 AND snmp_disable = 0 AND `device_id` = '118' AND poller_group IN(1,3,2,4) ORDER BY device_id DESC]
SQL[SELECT * FROM devices_attribs WHERE `device_id` = '118']
ndc-mon-001.inf.virt.ja.net 118 linux SQL[INSERT INTO `device_perf` (`xmt`,`rcv`,`loss`,`min`,`max`,`avg`,`device_id`,`timestamp`)  VALUES ('3','0','100','0','0','0','118',NOW())]
Unpingable./discovery.php 118 2018-03-22 20:42:31 - 0 devices discovered in 1.958 secs
SNMP [0/0.00s]: Get[0/0.00s] Getnext[0/0.00s] Walk[0/0.00s]
MySQL [7/0.01s]: Cell[2/0.00s] Row[0/-0.00s] Rows[4/0.00s] Column[0/0.00s] Update[0/0.00s] Insert[1/0.00s] Delete[0/0.00s]
RRD [0/0.00s]: Update[0/0.00s] Create [0/0.00s] Other[0/0.00s]
[root@sdc-mon-001 librenms]# echo $?
0
```

discovery.php
```diff
@@ -151,3 +152,8 @@
 }
 
 logfile($string);
+
+if (count($devices) == 0) {
```
TheMysteriousX: if ($devices_discovered == 0) {

murrant:
I'll see if I can update this so it merges later.

murrant:
@TheMysteriousX If you want you can do a pull request against my branch.

```
git remote add murrant https://github.com/murrant/librenms.git
git fetch murrant
git checkout -b new-python-service murrant/new-python-service
git push -u origin new-python-service
```

All set up now to make changes.  After changes, commit and push.

Then go on github and go to your branch and click new pull request.
For the base select murrant/librenms and new-python-service and click Create Pull request.
I'll review your changes quick and merge them in.

If changes have been made in my branch to pull them in:
```
git fetch murrant
git checkout new-python-service (if not already on it)
git rebase murrant/new-python-service
```
After a rebase you have to --force to push to your branch

TheMysteriousX:
Will do.

Preliminary results under python3.4+ are promising:
screenshot 2018-03-23 17 48 58

This is with maintenance scheduled every 2 minutes.

So I think the best option for the auto update is to detect any version of python lower than that to disable the maintenance queue, and document the limitation.

I'll leave it running over the weekend to make sure that the descriptor leak is really absent on 3.4+, and not just slow.

laf:
Is this ready for testing (i,e up to date with any pull requests into Tony's repo).

Any docs for it?

TheMysteriousX:
There's a from me pull waiting for @murrant to merge, then I think it's good to go for some testing. 

Under python<3.4 the automatic restart is disabled, so if there is a critical schema update it'll break like it does with the current poller service.

Docs are included, and have a note on how to get python3.4 working on RHEL/CentOS 7 quickly.

Not sure how easy it is on debian/arch, but it would be worth someone investigating if they have a system handy.

laf:
> Rethinking this, what is the actual impact of the service not restarting? The poller processes don't persist beyond the lifetime of each invocation, so the only impact is that if there's a change to the poller service itself, it doesn't get picked up? My gut feeling is that any such change would be pretty substantial, and would probably require a notification to be posted first, like was done with the PHP minimum version change.

The wrappers we have change very little but they have and do change so requiring people to restart the service manually could be a pain. Would be easier to just say this doesn't work with python < 3.4?

Once the PR is merged let me know. I can test this over a 4 node distributed setup.

TheMysteriousX:
If this were a new thing, then yeah, I'd say stick it with a minimum version. The old poller service has the same limitation though, and has that fun locking bug.

If someone deploys it onto python <3.4, they're no worse off, and may even be better off if they're hitting the bug.

laf:
Is this ready for testing now?

TheMysteriousX:
(Above comment was from me; used the wrong account again).

laf:
Was just about to test this on distributed polling but I have a question.

Does daily.sh still need to run via cron?

How do you disable things like service and billing from running as I only want those to run on one poller (well billing at least).

laf:
> Not sure I follow your second question - do you mean you have a specific poller you want those to run on?
> 
> Otherwise for billing, the poller service should select a single poller when it's time to run the billing.

Yeah at present I run billing on one of the poller nodes so it only runs once.

What I want to make sure is that it continues to do so. I.e I've 4 pollers, billing should only run once otherwise it will generate 4 x the amount of data it should. Same for alerts.php.

TheMysteriousX:
It should do that - the queues only one poller does the work assignment, and once per billing/alerting cycle it will generate a single billing and alerting job to be run on a single poller.

Jobs cannot be worked my more than one poller.

laf:
Ok seems to be working fine.

image

Looks like all pollers have dropped load by about 1.

CPU hasn't changed but I think this might be because it looks like the devices are spread out over the 5 minute window where as before they are polled as fast as possible.

The `/pollers/tab=pollers/` page no longer works which is the insight into how the pollers have done. It's not a major show stopper but it is a big loss into if the pollers have all completed recently so it would be good to get that back working.

I don't seem to be getting any logs into syslog though.

laf:
Actually I'm not sure this is honoring the poller groups.

I've got devices pinned to specific pollers which are being polled by the wrong ones. I'm going to turn this off for now.

TheMysteriousX:
> Let's see then :) Btw are you on discord / irc?

No, but I can be. I'll add it to my autojoins later (username:adamb).

Should be easy enough to fix the poller stats, I'll take a look at what the existing distributed poller does.

Logging is weird, what OS are you running under? If you check journaltcl (```journalctl -xe```) do you see anything there?

I'm surprised by the distributed polling groups not functioning - I only have a 2 system setup, but it did appear to be coordinating properly. Can you share your distributed poller config?

TheMysteriousX:
I can see a case where the queues don't get created properly - '2,3' gets created as a queue literally called '2,3' rather than parsed and split into two queues. That should be a quick fix. I'd expect that to result in devices not getting polled at all though.

I can't find a path where work would get mis-assigned though - I'll set up a couple of extra nodes to match your config and see if I can replicate.

laf:
> I can see a case where the queues don't get created properly - '2,3' gets created as a queue literally called '2,3' rather than parsed and split into two queues. That should be a quick fix. I'd expect that to result in devices not getting polled at all though.
> 
> I can't find a path where work would get mis-assigned though - I'll set up a couple of extra nodes to match your config and see if I can replicate.

I don't mind retesting to save the hassle of you spinning up more nodes.

The idea with our setup is we have about 4k devices pinned to poller-1 and poller-2. Then the rest on poller-3 and poller-4.

However with this, I was seeing poller-3 and poller-4 trying devices pinned to 1 and 2.

murrant:
Ran across my to-do and it says multiple `poller groups`.  Hah. The rest was testing items.

laf:
Awesome. I'll merge it in and re-test :)

laf:
```
Traceback (most recent call last):
  File "librenms-service.py", line 6, in <module>
    import LibreNMS
  File "/opt/librenms/LibreNMS/__init__.py", line 6, in <module>
    from .service import Service, ServiceConfig
  File "/opt/librenms/LibreNMS/service.py", line 12, in <module>
    from contextlib import AbstractContextManager
ImportError: cannot import name AbstractContextManager
```

@TheMysteriousX I get that regardless of python version now :(

TheMysteriousX:
Whoops, looks like that interface was only added in v3.6

I'll just remove it, the functionality won't be changed.

laf:
Testing now in anger

TheMysteriousX:
Fantastic. Did you manage to figure out where logs were being directed to?

laf:
I haven't checked the logs yet.

The poller history is a little better:

image

Get data now which is good but the poll time is crazy high :)

TheMysteriousX:
It will be - I think it's a more useful number though. With the old poller, it will just show the poll time of the slowest device(s), which is known already.

This gives you an indication of capacity - i.e. you have 16 poller threads, that's 4800 seconds of capacity, which means you're running your poller nodes at around 40-50% of your capacity right now.

TheMysteriousX:
murrant/librenms#12 is needed to make this work when applied to the master branch.

laf:
> It will be - I think it's a more useful number though. With the old poller, it will just show the poll time of the slowest device(s), which is known already.

It showed you the total time taken for that poller rather than individual device poll times. I'd say it's definitely important as you need to know that all your devices were polled within the 5 min window.

> This gives you an indication of capacity - i.e. you have 16 poller threads, that's 4800 seconds of capacity per poller, which means you're running your poller nodes at around 40-50% of your capacity right now.

That's sort of a good thing to have but to be useful we'd need to show how many poller threads that poller ran. It does need displaying better though as 4800 isn't true unless you can show each pollers thread performance. The overall pollers performance based on 300 seconds is better suited imho :)

TheMysteriousX:
screenshot 2018-04-20 03 32 50
Knocked something together quickly - not pushed it yet as the code is pretty bad and the UI needs some styling.

murrant:
@TheMysteriousX nice

laf:
@TheMysteriousX Flipping awesome. Nice work :)

So this is still working well for us so I'll definitely approve this once the webui you've just shown is added in.

murrant:
I was thinking for multiple groups allowing per-group thread assignment, this would take extra UI or config options.

murrant:
I also have some questions:
Why did you need to have the threads have a database connection?  I was trying to avoid that and just have the manager have a connection.
Why verify_queue()?  Do you suspect a device might be put into the wrong work queue?

TheMysteriousX:
Neither MySQLdb or pymysql are thread safe.

The queue verification was written because the queue name collision problem was really hard to debug because the other log messages give false infomration

If a user encounters a locking problem though misconfiguration, or a bug introduced at a later date, the logs won't contain any useful information to diagnose it without having the worker do some self testing.

murrant:
Looking at the changes the only reason you access MySQL in the worker threads is for that verification.  (Well aware it is not thread safe, trying avoid using it in worker threads)

Can you give an example of how to create the issue?

laf:
@murrant / @TheMysteriousX Where are we with this? From my testing it's working well so I'd be happy to merge.

murrant:
@laf I kind of want to change it a bit when it comes to the configuration and group queues.

I was waiting on the Laravel PR, now I am currently working on Laravel Auth, so might have some time after that.

laf:
> I was waiting on the Laravel PR, now I am currently working on Laravel Auth, so might have some time after that.

@murrant Is that because of an issue or because of perfection as if it's for perfection and this is new, we could merge......

TheMysteriousX:
I've been away so haven't had chance to look at it recently - outstanding changes as far as I'm concerned are to implement a thread pool to allow safe concurrent db access to address the concern with that, and to pull in the UI change.

Thread pools should be easy, UI... depends on how extensive the changes are now the Laravel PR has been merged.

(Is 'login with facebook' broken on community.librenms btw? I get 'App not set up: This app is still in development mode, and you don't have access to it. Switch to a registered test user or ask an app admin for permissions.' trying to access it).

laf:
> (Is 'login with facebook' broken on community.librenms btw? I get 'App not set up: This app is still in development mode, and you don't have access to it. Switch to a registered test user or ask an app admin for permissions.' trying to access it).

Give it a try again now.

TheMysteriousX:
Still the same error I'm afraid.

laf:
@TheMysteriousX try again now

TheMysteriousX:
That did the trick, thanks!

murrant:
@laf not perfection.  I want to have one way to set up the configuration.
I added NODE_ID in the Laravel PR, so each node should have a unique identifier (not hostname) that we can use to tie configuration items to it.

What I want to do is have the pollers page, show all pollers registered (and unique configuration per poller in the database).  This way there does not need to be specific configuration for each node done locally other than the mysql and redis connection.

@TheMysteriousX What did you think about https://p.libren.ms/view/e8d4eaf4 ?

TheMysteriousX:
I've implemented per group threads as our diff @murrant  - seems to be working mostly ok, but I'm seeing that the billing and alerting jobs are running more often than I'd expect.

Should be able to find some time tomorrow to figure out why.

adam-bishop:
In the interest of getting this merged, I'm going to back out @murrant 's group thread change. I've spent a lot of time trying to stabilise it, but it's now at the point where it runs fine for a day or so then fails with an SQL deadlock, despite so far as I can tell queries only being executed on the main thread.

TheMysteriousX:
In the interest of getting this merged, I'm going to back out @murrant 's group thread change. I've spent a lot of time trying to stabilise it, but it's now at the point where it runs fine for a day or so then fails with an SQL deadlock, despite so far as I can tell queries only being executed on the main thread.

TheMysteriousX:
I was able to figure out where the problem was, so per thread groups for polling and discovery are left in.

Each node does end up with a connection as they need to report their state.

laf:
@TheMysteriousX Awesome work - is that in the PR you submitted to murrants branch?

I'd love to see this merged as our work install is now 3 releases behind because I've got this in testing :)

laf:
Is this ready for a test again?

jodell22:
**Ubuntu 16.04 Distributed Polling Setup**
**Poller Config**
`sudo update-alternatives --install /usr/bin/python python /usr/bin/python3 1` This changes the default python version from 2.7 to version 3 (3.5.2)
`sudo apt install python3-pip`
`sudo su` I got errors when trying to run pip with sudo. Switching to root with sudo su worked fine though.
`pip3 install pymysql`
`pip3 install redis`

**Redis Server**
I followed this guide to install a dedicated Redis server:
[https://www.digitalocean.com/community/tutorials/how-to-install-and-configure-redis-on-ubuntu-16-04](https://www.digitalocean.com/community/tutorials/how-to-install-and-configure-redis-on-ubuntu-16-04)

The only things missing from that guide were how to set it up for remote access, as well as setting up authentication. Remote access and auth can be set up in the redis.conf file:
`sudo vim /etc/redis/redis.conf`
Drop down to the Security section and uncomment the requirepass line like this:
`requirepass securePassword`
I also commented out the bind line in the network section, so it listens on all interfaces:
`#bind 127.0.0.1`

**Results**
We currently are using 3 pollers to scan just over 1000 devices. The pollers are easily chewing through all the devices within 5 minutes. One thing that I did notice is my poller cluster health table is not filling in properly:
screen shot 2018-06-04 at 8 51 33 pm

TheMysteriousX:
@jodell22, the fix for that is to remove the '+1' from the poller page.

There's (hopefully) final pull request in that fixes the broken table rendering and the test fails.

TheMysteriousX:
One thing it would be good to thoroughly check is service polling - it's not a feature I use.

jodell22:
Removing the +1 from the /opt/librenms/html/pages/pollers/pollers.inc.php
`$stat_count = count($stat_row) + 1;`
line fixed the table.

I'm not polling any services either.

laf:
I've updated to this latest branch for testing now.

This is a great screen:

image

Couple of questions. Is it worth hiding the `Standard Distributed Pollers` as it makes it look like something is broken?

Poller-1 for me deals with alerts but this doesn't show in the health page.

Awesome job on this. Really looking for it to be merged in so I can put my install back to normal :)

adam-bishop:
Could hide them, the current behaviour should be that there is a note saying "you have no standard pollers", is that happening?

Hm, are you sure it's poller-1 that's doing your alerts? I didn't touch that part of the code but I think it should be the master that executes the alerts. I'll double check that's the case.

laf:
> Could hide them, the current behaviour should be that there is a note saying "you have no standard pollers", is that happening?
 
I don't see it no.

> Nope, ignore me, it's put into a dispatch queue like the others. I'm not sure why it's always running on poller-1 for you.

Ignore me, just read the docs. This python service doesn't handle alerting anyway does it. You still need to run `alerts.php`.

I've been running this now for 4 days with no noticeable issues. 

Once that last bit of WebUI is sorted I'd be happy to merge. @murrant?

jodell22:
Just as an update:

We've bumped our system up to 4 pollers and just under 2000 devices on 5 minute poll intervals. We've been running this way for about 5 days now and everything is working pretty smooth. 

I did notice that 2 of my pollers stopped checking in a few days ago, but seem to still be polling devices. Not sure if anyone else has seen that?

cousard:
@TheMysteriousX 
I've run into a situation where it appears that the librenms-service is spinning up "poller.php -h all" instances that sit for long periods of time. I could be wrong but I don't think that is an intended action. 

Here is a 'ps auxf' from my master vm:
https://pastebin.com/raw/EbfeH6a2

There is a few of these instances that appear to have been running for hours. This is not isolated to the master. 

@jodell22, 
I think I saw something like this happening early one(failed to turn daily.sh off on a poller of mine). Re-applied the PR and saw it report in fine after that.

adam-bishop:
@jodell22 can you check your system log/journal for any python exceptions? If it's not checking in that probably means the dispatcher loop has failed.

@cousard absolutely not intended. Are you running the latest revision? That was one of the bugs I fixed when I first picked this up from @murrant. Possible it's regressed or the fix was incomplete. I'll see if I can reproduce it.

jodell22:
@adam-bishop Here's what I found for errors:

> Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]: Discovery_0-8(ERROR):Discovery poller exception! invalid expire time in set
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]: Traceback (most recent call last):
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:   File "/opt/librenms/LibreNMS/queuemanager.py", line 58, in _service_worker
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:     work_func(device_id)
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:   File "/opt/librenms/LibreNMS/service.py", line 327, in discover_device
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:     if self.lock_discovery(device_id):
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:   File "/opt/librenms/LibreNMS/service.py", line 480, in lock_discovery
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:     return self._lm.lock(lock_name, self.gen_lock_owner(), timeout, retry)
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:   File "/opt/librenms/LibreNMS/__init__.py", line 208, in lock
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:     return self._redis.set(key, owner, ex=int(expiration), nx=non_existing)
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:   File "/usr/local/lib/python3.5/dist-packages/redis/client.py", line 1171, in set
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:     return self.execute_command('SET', *pieces)
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:   File "/usr/local/lib/python3.5/dist-packages/redis/client.py", line 668, in execute_command
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:     return self.parse_response(connection, command_name, **options)
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:   File "/usr/local/lib/python3.5/dist-packages/redis/client.py", line 680, in parse_response
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:     response = connection.read_response()
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:   File "/usr/local/lib/python3.5/dist-packages/redis/connection.py", line 629, in read_response
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]:     raise response
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]: redis.exceptions.ResponseError: invalid expire time in set
Jun 12 12:59:59 ClientNMS-Poller4 librenms-service.py[5092]: Discovery_0-3(ERROR):Discovery poller exception! invalid expire time in set

cousard:
@adam-bishop 
I was running 1.39-70-g57a23a7. Removed the PR. updated with daily.sh and re-applied the PR. 

Validate.php output:
Component | Version
--------- | -------
LibreNMS  | 1.40-9-g78e3df4
DB Schema | 252
PHP       | 7.1.17
MySQL     | 5.5.56-MariaDB
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.2

Still running into the same issue. In the pastebin below you'll see the first instance under the service is a poller with -h all.
https://pastebin.com/raw/8RLpMZDi

TheMysteriousX:
@cousard I've identified your issue - the billing timer is broken and is putting events into the wrong queue. I'm not actually sure it ever worked reliably. I'll see if I can put together a fix tomorrow.

@jodell22 that's interesting, and I'll see if I can figure out why an expiry isn't being set; looking at the locking code I can see a few places a 'None' expiry could leak through. Wouldn't cause what you're seeing though.
Could you check that the message "Counting up time spent polling" isn't being logged? That would confirm I'm on the right track.

jodell22:
@TheMysteriousX I left a tail open over night on that poller's syslog file and I don't see any logs referencing "Counting up time spent polling"

Am I looking at the correct log location, or would it be showing up in the /opt/librenms/logs/librenms log?

laf:
Actually am seeing the `-h all` as well so happy to test any patches.

Still see `-h $id` as well though so it's not doing it on all threads.

TheMysteriousX:
murrant#15 should contain a fix for @cousard, and some improved robustness and debugability for @jodell22, should they see the issue reoccur.

I've also changed the behaviour of the poller page to just hide the table if there is no applicable poller type.

jodell22:
So I did a github-remove -d and a github-apply 8455 and now the service won't start:

> Jun 14 09:40:06 ClientNMS-Poller2 librenms-service.py[17971]: module 'LibreNMS' has no attribute 'Service'
Jun 14 09:40:06 ClientNMS-Poller2 systemd[1]: librenms.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Jun 14 09:40:06 ClientNMS-Poller2 systemd[1]: librenms.service: Unit entered failed state.
Jun 14 09:40:06 ClientNMS-Poller2 systemd[1]: librenms.service: Failed with result 'exit-code'.
Jun 14 09:40:09 ClientNMS-Poller2 systemd[1]: librenms.service: Service hold-off time over, scheduling restart.
Jun 14 09:40:09 ClientNMS-Poller2 systemd[1]: Stopped LibreNMS SNMP Poller Service.

TheMysteriousX:
@jodell22 Looks like github-apply doesn't work quite right. I needed to reset the working directory completely before running github-apply a second time.

laf:
I'm just running this in screen at the moment, do I need to stop the python service and restart for it to take effect?

TheMysteriousX:
It'll restart eventually (default timer for that is 24h), but if you want to pick up the changes faster, you can terminate it yourself.

There's one more bugfix going in once I've confirmed that it squishes that redis lock expiry message - need to run it for a few hours first.

murrant:
@TheMysteriousX I can make it apply nicely again by rebasing it, but you will need to `rebase` or `reset --hard` (this will destroy any local changes) your branch too

laf:
The latest changes seem to cause issues. Woke up to really patchy polling, missing data and 2 of the 4 pollers reporting:

```
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
performance(INFO):Counting up time spent polling
```

Up on restarting the service they just get:

```
poller-1(INFO):Attaching signal handlers on thread poller-1
poller-1(INFO):Using pure python SQL client
poller-1(INFO):Using pure python SQL client
poller-1(INFO):Using pure python SQL client
poller-1(INFO):Entering main LibreNMS service loop on 13170/poller-1...
poller-1(INFO):Groups: [0]
poller-1(INFO):Poller QueueManager created: 24 workers, 300s frequency
Poller_0-1(INFO):Creating queue poller:0
poller-1(INFO):Groups: [0]
poller-1(INFO):Alerting QueueManager created: 1 workers, 60s frequency
Alerting(INFO):Creating queue alerting:0
poller-1(INFO):Groups: [0]
poller-1(INFO):Services QueueManager created: 1 workers, 300s frequency
Services(INFO):Creating queue services:0
poller-1(INFO):Groups: [0]
poller-1(INFO):Discovery QueueManager created: 2 workers, 21600s frequency
Discovery_0-1(INFO):Creating queue discovery:0
poller-1(INFO):Groups: [0]
poller-1(INFO):Billing QueueManager created: 2 workers, 300s frequency
Billing(INFO):Creating queue billing:0
poller-1(INFO):LibreNMS Service: poller-1-cb9460f6-7070-11e8-93f4-44a8420e8020 started!
poller-1(INFO):Poller group 0 (default). Using Python 3.4.3 and redis locks and queues
poller-1(INFO):Maintenance tasks will be run every 1 day, 0:00:00
```

And it didn't appear to move on from there.

I've had to stop the service now on all and revert to the cron.

TheMysteriousX:
> poller-1(INFO):Entering main LibreNMS service loop on 13170/poller-1...

Can you check that the code has applied cleanly? That's some serious WTF there - the main event loop is running on a worker thread.

laf:
I'm just running this branch directly:

```
commit 6315245e3ad84075374a6c3e591a4c692d8d5ffd (HEAD -> new-python-service, origin/new-python-service, murrant/new-python-service)
Author: Adam Bishop <TheMysteriousX@users.noreply.github.com>
Date:   Thu Jun 14 06:35:06 2018 +0100

    * Remove implicit queue creation (#15)

    This behaviour is masking bugs. The default values should not be needed
      if the timers and queue managers are initialised properly.
    * Add extra log messages to troubleshoot threading problems
    * Fix billing events
    * Don't special case alerting
    * Make the stats logging robust against failure
    * Rename schema
    * Hide the poller tables if they are empty

commit 7caa6a80790d07cfb8188ec0048d7a3a7b6874bf
Merge: bdbc975d4 352916c8b
Author: Neil Lathwood <gh+n@laf.io>
Date:   Thu Jun 7 00:04:45 2018 +0100

    Merge pull request #14 from TheMysteriousX/new-python-service

    Test fixes and UI tweak

commit 352916c8b97e6c4dd7bfedf899676903ccdff668
Author: Adam Bishop <adam@omega.org.uk>
Date:   Tue Jun 5 11:51:58 2018 +0100

    Schema update

commit 28bf5b4178d776040abc1c1c9621d5ce87c4dd5f
Author: Adam Bishop <adam@omega.org.uk>
Date:   Tue Jun 5 11:46:04 2018 +0100

    Fix off by 1 error
    Fix test failure
    Fix ugly hover effect with rowspan
```

TheMysteriousX:
If you revert back to 7caa6a80790d07cfb8188ec0048d7a3a7b6874bf does it start running?

laf:
I've not explicitly checked since but that commit was working fine until I updated last night.

murrant:
I made schema changes, but I did not make a new schema file.

dsgagi:
Hi. Based on the input in #8757 , query in **fetch_services_device_list** should also drop "WHERE `disabled`=0". Thing is, I can't commit to murrant's branch :)

If you agree, I would also modify documentation FAQ page a little. My suggestion is below.

> What is the Difference between Disable Device and Ignore a Device?
> - Disable stops polling. ----> Disable stops device **SNMP** polling. Service polling will stay active.
> - Ignore disables alerting. ----> Ignore disables device alert triggering and clears already triggered alerts.

murrant:
@dsgagi you can create a pull request to it though.

TheMysteriousX:
@laf I think I've reproduced what you saw. If I run some pollers on 0e4c0fb599ae2f1d1065e7febd9794dc375cd763 and some on ba06209f1a8264eb853e899607a7eb5ab007f2f6, the non-master pollers on group 0 stop.

TheMysteriousX:
@murrant what was your intent with the discovery lock timer?

At the moment, it's defined as:
```
timeout = self.config.down_retry if retry else LibreNMS.normalize_wait(self.config.discovery.frequency)
```

normalize_wait emits the number of seconds until the next discovery window - because the discovery window is also aligned using normalize_wait, this means that some discovery jobs are kicked off with a lock expiry of <1 second.

My thinking is that for discovery jobs, it's probably fine to have them with a static timeout - maybe a frequency/10?

laf:
> @laf I think I've reproduced what you saw. If I run some pollers on 0e4c0fb and some on ba06209, the non-master pollers on group 0 stop.

Awesome. Thanks for taking a look.

murrant:
@TheMysteriousX normalize_wait is supposed to align the timers based on calendar time to prevent drift.
Sounds like there is a bug in the calculation?  I didn't quite understand what you described.

Oh, also, beside the schema update, python-dotenv is required now.

TheMysteriousX:
So, with 6 hour discovery intervals, it starts setting off discovery events at approximately 18:00 UTC, ±1 second after rounding.

The events that get fired off at 17:59:59 have a lock expiry of 1 second. The events fired at 18:00:00 have an expiry of 21600 seconds.

This makes the first few discovery events get launched twice, because the lock expires instantly.

murrant:
@TheMysteriousX 
```python
def normalize_wait(seconds):
    return math.ceil(seconds - (time() % seconds))
```

That should prevent it from scheduling before the upcoming schedule...

TheMysteriousX:
Ok, that works. I've sent another pull that forces the lock time to be at least 1 - without it,if a 0 second timeout works its way through somehow the worker thread will crash.

murrant:
I think I like your fix better.  Perhaps the min should be more that 1s though, just in case.

TheMysteriousX:
The fixes are both needed really - changing normalise wait makes sure that an event doesn't get scheduled too early, forcing a minimum just makes sure that if e.g. it's running in a VM and gets stunned for too long, or there's a leap second, or some other weirdness, we don't have a thread crash.

Could increase the minimum; It's difficult to pick a value though as the different timers have such wildly different runtimes.

TheMysteriousX:
@jodell22 @cousard @laf is everything looking happy on your systems?

laf:
@TheMysteriousX Will need to re-test, I rolled back to cron due to the last bug. Leave it with me.

jodell22:
@TheMysteriousX Working on updating all our pollers to the latest code now. Been a bit occupied with other things recently, so I'll let you know how things look with the latest code.

Everything was running great still when I checked it before updating today. This was on the build you released about 8 days ago. ba06209 I believe.

laf:
I moved to the latest commit and unfortunately polling seems to fail.

I can't afford to lose data on this prod system so I might have to back out of any further testing right now.

laf:
Scratch that, looks like it was ok just rrd not showing the latest data.

Would be interested to see what @jodell22 makes of it.

jodell22:
So polling seems to be working good. I did notice that the pollers don't appear to be checking in anymore though. The polling page shows them all red. 
screen shot 2018-06-21 at 8 02 36 pm

I did double check, and everything is still getting pulled in, though.

laf:
> So polling seems to be working good. I did notice that the pollers don't appear to be checking in anymore though. The polling page shows them all red.
> 
> 
> I did double check, and everything is still getting pulled in, though.

Same issue that I saw as well.

TheMysteriousX:
Did y'all run validate.php and execute the sql query it suggests after updating past 6b5df236d82774b068d44c9098b05d2e2fa8d79f?

jodell22:
@TheMysteriousX  I hadn't done that either. I just tried it on my db server and ran into this error:

> MariaDB [librenms]> ALTER TABLE `poller_cluster` ADD `node_id` varchar(255) NOT NULL  PRIMARY KEY AFTER `id`;
ERROR 1068 (42000): Multiple primary key defined

laf:
Ok so I've re-started the python daemon and this is collecting data as expected.

However the Poller Cluster Health page only shows 1 of 4 pollers. This happens to be the first poller that starts. Confirmed by starting poller-1 first then stopping them all and starting poller-4. Both times only that poller started first shows stats.

Also needs validate update to remove the reference to old pollers not completing and using the new stats to determine this if possible:

```
[FAIL]  The poller (poller-1) has not completed within the last 5 minutes, check the cron job.
[FAIL]  The poller (poller-2) has not completed within the last 5 minutes, check the cron job.
[FAIL]  The poller (poller-3) has not completed within the last 5 minutes, check the cron job.
[FAIL]  The poller (poller-4) has not completed within the last 5 minutes, check the cron job.
```

I'd love to get this out the door for a 1.41 release but that means it's got to be ready in a few days....:)

laf:
I have just seen this logged:

```
performance(ERROR):Unable to log performance statistics - is the database still online?
Traceback (most recent call last):
  File "/opt/librenms/LibreNMS/service.py", line 718, in log_performance_stats
    getattr(self.config, worker_type).frequency)
  File "/opt/librenms/LibreNMS/__init__.py", line 82, in query
    cursor.execute(query, args)
  File "/usr/local/lib/python3.4/dist-packages/pymysql/cursors.py", line 165, in execute
    result = self._query(query)
  File "/usr/local/lib/python3.4/dist-packages/pymysql/cursors.py", line 321, in _query
    conn.query(q)
  File "/usr/local/lib/python3.4/dist-packages/pymysql/connections.py", line 860, in query
    self._affected_rows = self._read_query_result(unbuffered=unbuffered)
  File "/usr/local/lib/python3.4/dist-packages/pymysql/connections.py", line 1061, in _read_query_result
    result.read()
  File "/usr/local/lib/python3.4/dist-packages/pymysql/connections.py", line 1349, in read
    first_packet = self.connection._read_packet()
  File "/usr/local/lib/python3.4/dist-packages/pymysql/connections.py", line 1018, in _read_packet
    packet.check_error()
  File "/usr/local/lib/python3.4/dist-packages/pymysql/connections.py", line 384, in check_error
    err.raise_mysql_exception(self._data)
  File "/usr/local/lib/python3.4/dist-packages/pymysql/err.py", line 107, in raise_mysql_exception
    raise errorclass(errno, errval)
pymysql.err.IntegrityError: (1048, "Column 'parent_poller' cannot be null")
```

TheMysteriousX:
I'm not going to be able to look at this for a few days, but my best guess is that the "on duplicate key" part of the sql statement isn't working right now the primary key has been changed.

TheMysteriousX:
if @laf posts atable dump of `poller_cluster`, I think you'll see node_id being set to `None` or similar, which is making the poller instances fight over a single single row.

That would mean that the root cause is that `node_id` is not being set somehow - perhaps it should be set to `poller_name` if that happens.

If I'm right, `SELECT id INTO @...` should be changed to target `node_id` - then it won't crash if `node_id` is somehow unset.

laf:
```
MariaDB [librenms]> select * from poller_cluster\G
*************************** 1. row ***************************
            id: 1
       node_id: None
   poller_name: poller-1
poller_version: librenms-service
 poller_groups: 3,2
   last_report: 2018-06-24 17:20:00
        master: 0
1 row in set (0.00 sec)

```

You are indeed correct :)

murrant:
Yeah, I was figuring that.  I think my fix would be to alert the user and give them a simple command to generate NODE_ID.

TheMysteriousX:
murrant#17 should make the validator consider any clustered pollers.

laf:
So I've pulled the latest changes in. `.env` has NODE_ID set in all pollers. Should the Poller health page be functioning as I only see a single poller still?

murrant:
@laf it takes up to 5 minutes to update the stats/poller table.

laf:
It's been a lot longer than that :)

murrant:
Hmm, maybe the SQL is failing.

TheMysteriousX:
What's being put into the db - is node_id still none?

laf:
Yup still None.

TheMysteriousX:
Could you grep your log for "Could not import .env" please, make sure it's cleanly importing the node_id from disk.

laf:
I'm actually just running this via screen right now so don't have that info :(

murrant:
Hmm, failing to parse .env should probably be fatal.

TheMysteriousX:
The code looks to be fine. I could find 4 ways to break it:
* Corrupting `.env` (manually)
* Installing `dotenv` instead of `python-dotenv`, or not having `python-dotenv` installed at all
* Having `.env` inaccessible by the 'librenms' user account due to *nix permissions or selinux
* Having duplicate rows in the SQL tables (because the SQL queries are inconsistently switching between `poller_name` and `node_id`).

Any of those sound possible @laf?

TheMysteriousX:
murrant#18 makes the SQL queries consistently use node_id, and makes failure to load the node_id 
 fatal.

I wouldn't recommend upgrading all instances right away as if it is a configuration problem, the poller won't start - do one that has redundancy, then figure out what the root cause is from there.

TheMysteriousX:
I noticed that too - I had a change that fixed it, but @murrant got there first with #8848

laf:
Ace. I'll merge that in as well :)

laf:
Merged - big thanks to @murrant and @TheMysteriousX for all of their work on this one :)

murrant:
@TheMysteriousX are you on Discord or IRC?  There is a bug with the stats and I'm not sure what to do about it.

PR #8456, opened 2018-03-23
devices: Add Phybridge OS support

vivia11:
Hello,
Adding support for NVT Phybridge.
Tested with a Flex Switch.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
You need to minimize your svg files and the icon should be inverted I think.
https://jakearchibald.github.io/svgomg/

laf:
I've just pushed the json test data needed for this.

however, the OS name is just the company name when it should be named after the firmware of the device. What is that name? If it doesn't have one then it should closely resemble the product name.

vivia11:
@laf 
From what I can gather the software version is: "SYS 3.1.9 PKG 7.0.0". Not sure if SYS or PKG is the name...
The model name is Flex Switch - 24 Port.

Should I rename the files/os to sys | pkg | flex?

laf:
Looks like it might be better as phybridge-flex

vivia11:
Okay I renamed the OS to phybridge-flex. Let me know if I'm missing something or if I should rename anything else.

PR #8458, opened 2018-03-23
Add config switch for selective polling to force HC Counter usage

cchance27:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/tests/YamlTest.php
+++ b/tests/YamlTest.php
@@ -84,6 +84,7 @@ class YamlTest extends TestCase
         'poller_modules',
         'discovery_modules',
         'icons',
+        'polling',
         'ifname',
         'good_if',
         'bad_if',


murrant:
You should change the yaml for the device this is intended to fix.  No reason to merge theoretical code ;)

murrant:
You should also add documentation here https://docs.librenms.org/Developing/os/Settings/

laf:
I worry about things like this having done many work arounds in port polling previously.

What we tend to find is that it's a limited set of devices or even firmware versions that require this and in reality the vendor should fix it.

cchance27:
While I agree the vendor should fix it and I’m working on that with cambium in my instance of the problem, I still think their should be a way for users to force HC to be used as the alternative is broken graphs or being forced to switch the device to full polling which in some cases might mean way more polling on larger port devices.

This just adds a last resort override for those users that are stuck.

cchance27:
will merge documentation changes this weekend as requested. 

Is their an explanation of the device yaml's and how to handle setting config flags?

murrant:
That is the wrong place in the documentation.
It should be documented in these two locations:
https://docs.librenms.org/Support/Configuration/
https://docs.librenms.org/Developing/os/Settings/

First one is for users using in config.php, second one is for developers in yaml files.

cchance27:
i had considered that but as selected_ports isn't part of general documentation (because its marked as experimental) and force_hc only affects selected_ports thats why i added it to that section, adding it to those sections seems odd without those sections having mention of the selected_ports

murrant:
I'm thinking selective port polling is not so experimental any more.

At least add it to this page, it is intended to collect ALL configurable settings for OS.
https://docs.librenms.org/Developing/os/Settings/

murrant:
@laf I find this preferable to also walking ifHCInOctects for selected port polling.  Thoughts?

murrant:
I'm unsure about it too.

cchance27:
So is this a no go ? im still running it to get the cambium stuff to poll correctly :S

laf:
@cchance27 It's not something I would merge personally.

PR #8462, opened 2018-03-23
Change the default alert delay to 1m

murrant:
Add regex type for numbers
Tooltips for all alert rule dialog items
Remove spaces in time settings.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@laf The reason for the duplicate tooltips was I wanted the tooltip on both parts of the field, but we break the field up because of the html structure.

I could probably try to restructure the html.

Issue #7921, opened 2017-12-19
Add Support For Tegile IntelliFlash


jdub01010101:
Good morning,

Per https://community.librenms.org/t/add-support-for-tegile-sans/2860, I am posting to have Tegile Intelliflash SAN support added to LibreNMS.  

DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.

Component | Version
--------- | -------
LibreNMS  | 1.33-245-gd6d4a3a
DB Schema | 224
PHP       | 7.1.12-2+ubuntu16.04.1+deb.sury.org+2
MySQL     | 10.0.31-MariaDB-0ubuntu0.16.04.2
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3

- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).

./discovery.php -h HOSTNAME -d | ./pbin.sh -- https://p.libren.ms/view/88bf844f
./poller.php -h HOSTNAME -r -f -d | ./pbin.sh -- https://p.libren.ms/view/86478af5
snmpbulkwalk -OUneb -v2c -c COMMUNITY HOSTNAME .  | ./pbin.sh -- https://p.libren.ms/view/7a42ae49
SNMP MIB -- https://p.libren.ms/view/a984fe91

- [x] Please provide as much detail as possible.

Tegile SANs are Active-Active normally, but our deployment is Active-Passive.  As such there are two controllers.  The pair share an IP but it does not respond to SNMP.  From what I can tell the controllers each need to be monitored separately.

- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

theherodied:
[...]

PR #8464, opened 2018-03-23
Added Tegile Storage OS detection, storage pool, and project storage.

theherodied:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #7921


theherodied:
Don't approve this yet. It's a work in progress. I just want testers to try it.

theherodied:
This is ready to merge. Tegile OS detection and pool usage has been added. 
More will be added in a future PR.

laf:
Thanks for this @theherodied.

Is tegile the correct os name? It should be based on the firmware / OS it's actually running so should this be IntelliFlash instead?

theherodied:
@laf, I went back and forth on that from what I saw in the disco and poller. I can update it but how would you want it named?

They call it Tegile IntelliFlash in the snmp output. 

Just intelliflash all lowercase?

theherodied:
@laf, let me know how this is now.
I didn't change the icon name as it's really the Tegile Logo I believe. I also left the mib directory as tegile/TEGILE-MIB. I left the mib named as they have it in the actual mib file itself.

theherodied:
Changed array to intelliflash1. Intelliflash2 will be for project usage as it has to be calculated differently.

includes/definitions/intelliflash.yaml
```diff
@@ -0,0 +1,11 @@
...
+text: 'Tegile IntelliFlash'
+type: storage
+icon: tegile
+over:
+    - { graph: device_storage, text: 'Storage Usage' }
+mib_dir:
+    - tegile
+discovery:
+    - sysDescr:
+      - Tegile IntelliFlash
```
laf: Looks like you can use `sysObjectID` instead for detection. From our stats `.1.3.6.1.4.1.43906.1` is unique to this device. sysObjectID is more stable for detection.

includes/discovery/storage/intelliflash.inc.php
```diff
@@ -0,0 +1,66 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.storage snmpgnu.org/licenses/>.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2018 Ryan Finney
+ * @author     https://github.com/theherodied/
+ */
+use LibreNMS\Config;
+if($device['os'] = 'intelliflash') {
```
laf: `=` will always match here :) Should be `==`

laf:
Test data updated, hopefully that fixes travis.

murrant:
Storage units is not set (you need to set it to 1) and the SQL is failing, that is why tests fail.

theherodied:
Fixes: #7921

PR #8470, opened 2018-03-25
Debian stretch settings

jepke:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/RRDCached.md
+++ b/doc/Extensions/RRDCached.md
@@ -159,7 +159,11 @@ For local RRDCached server
 $config['rrdtool_version'] = '1.5.5';
 $config['rrdcached'] = "unix:/var/run/rrdcached.sock";
 ```
-For remote RRDCached server
+For remote RRDCached server 
+Make sure you have network option in /var/default/rrdcached
+```bash
+NETWORK_OPTIONS="-L" 
+```
 ```php
 $config['rrdtool_version'] = '1.5.5';
 $config['rrdcached'] = "0:42217";
--- a/doc/Extensions/RRDCached.md
+++ b/doc/Extensions/RRDCached.md
@@ -156,7 +156,6 @@ BASE_OPTIONS="-B -F -R"
 
 For local RRDCached server
 ```php
-$config['rrdtool_version'] = '1.5.5';
 $config['rrdcached'] = "unix:/var/run/rrdcached.sock";
 ```
 For remote RRDCached server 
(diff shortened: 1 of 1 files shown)


doc/Extensions/RRDCached.md
```diff
@@ -124,6 +124,46 @@ OPTS="$OPTS -w 1800 -z 1800 -f 3600 -t 4"
...
+3. Restart the rrdcached service
+```bash
+    systemctl restart rrdcached.service
+```
+
+4. Edit /opt/librenms/config.php to include:
+
+For local RRDCached server
+```php
+$config['rrdtool_version'] = '1.5.5';
```
laf: This should be set to the version installed, so 1.6.0 in this case (don't worry about the minor version)

doc/Extensions/RRDCached.md
```diff
@@ -124,6 +124,46 @@ OPTS="$OPTS -w 1800 -z 1800 -f 3600 -t 4"
...
+$config['rrdcached'] = "unix:/var/run/rrdcached.sock";
+```
+For remote RRDCached server
+```php
+$config['rrdtool_version'] = '1.5.5';
+$config['rrdcached'] = "0:42217";
```
laf: This won't work with the example config above so I'd remove that to avoid confusion.

doc/Extensions/RRDCached.md
jepke: well i banged my head on that last 2 days but when it is set to 1.6.0 not a single graph on the overview tab is drawn. so that might be a bug if it supose to be working

doc/Extensions/RRDCached.md
jepke: you are right, made some change to add network option in rrdcached to enable networking

doc/Extensions/RRDCached.md
laf: Possibly, we'd need more info for that but that doesn't mean we should put the wrong info in docs.

jepke: totaly agree but my thoughts on the version variable where if above 1.5.5 put that as variable. after i managed to get it working that way :) 

but the error i got when having 1.6.0 in the config was. after i did revert it to 1.5.5 in the config it worked.
RRDTool Output ERROR: can not parse an empty rpn expression

laf: But then the config is not what version is installed and if we add any further checks later on this could break things - which if it was polling then people lose data.

If this surfaces someone with this setup who can help debug the issue then we can do that rather than purposefully tell people to set the wrong config from day one.

jepke: did remove the lines ill try to find the root cause tonight. would be great to have someone troubleshoot this with me

PR #8475, opened 2018-03-26
Update PDU2-MIB

salt-lick:
Also - it appears that some sensors for these Raritan PDUs still want to pull the MIB out of `/opt/librenms/mib` instead of the raritan folder.

This was [fixed](https://github.com/librenms/librenms/pull/8476).

Also this MIB adds humidity sensors.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Care to actually add support for the humidity sensors too?

https://docs.librenms.org/Developing/os/Health-Information/

laf:
@salt-lick Any response to the last comment?

salt-lick:
@laf Sorry for the delay. I was waiting for the vendor to reply to me. Apparently, these devices do not have a built-in humidity sensor. You can add an external sensor but unless you do there is no data present in the sensor table. I do not have these sensors to check what data they poll or what information is there. I honestly don't know what the next step would be. Close the pull? This is the latest MIB for the raritans and it should be backward compatible.

laf:
@salt-lick No we can merge for now. Can you (re)-sign the CLA - we've updated it.

PR #8478, opened 2018-03-27
newdevice: Add Last Mile CTM support

pheinrichs:
This PR adds support for Lastmile Gear CTM. 
Please review the custom state polling I've added. 

Thanks!

~~The blocker is in place as I'm waiting to see if there is a newer MIB from the manufacturer.~~
Update: This is the newest MIB available. Custom MIB will easiest solution. For now

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/OS/Lastmile.php
```diff
@@ -0,0 +1,44 @@
...
+use LibreNMS\Device\WirelessSensor;
+use LibreNMS\OS;
+
+class Lastmile extends OS implements WirelessClientsDiscovery
+{
+    /**
+     * This is a bit of a stretch. Clients are referencing `satellites`
+     * as this is a gps timing device.
+     */
+    public function discoverWirelessClients()
```
murrant: This seems like the complete wrong place for this data.  I'd rather not include it than put it here.

murrant:
@pheinrichs what is this device.  It appears to be a rebranded Cambium Canopy?

pheinrichs:
@murrant It is a GPS timing device.It is a custom firmware indeed. None of the cambium mibs / oid's would work.

murrant:
I mean the mib looks like a copy if the canopy mib.

Ok, gps satellites does not belong in the wireless clients stat.  You can either create a custom graph or omit it

murrant:
Also, update the os description please.

pheinrichs:
@murrant Consider it done, 
Would you be opposed to me adding a new wireless sensor for GPS satellites?

murrant:
@pheinrichs I think the gps satellites would be more of a generic thing.  Almost most any type of device could have gps in it.

includes/polling/os/lastmile.inc.php
```diff
@@ -0,0 +1,17 @@
...
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+
+$hardware = $device['sysDescr'];
+$version = snmp_get($device, '1.3.6.1.4.1.25868.1.1.0', '-Ovqs', '');
```
laf: You can drop that last `''`

includes/definitions/lastmile.yaml
```diff
@@ -0,0 +1,11 @@
+os: lastmile
```
laf: This looks like it's the vendors name. The OS name needs to be what the firmware is called or if that's not possible the device / type of device.

includes/definitions/discovery/lastmile.yaml
```diff
@@ -0,0 +1,149 @@
...
+            data:
+                - 
+                    oid: tempC
+                    num_oid: .1.3.6.1.4.1.25868.1.9.
+                    index: tempC.0
+                    descr: 'Temperature'
+        power:
+            options:
+                divisor: 10
+            data:
```
laf: I'm assuming this isn't in a table? :)

pheinrichs: Boy do I wish it was...

PR #8481, opened 2018-03-27
feature: Updating F5 Discovery Polling

dnapper:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This changes the method of polling for F5 objects.
The previous polling pulled 77 trees for each VS and 55 trees for each Pool. 
I update the polling to only pull the required 6 VS trees and 16 Pool trees. 

Attached is the improvement in discovery times for two of our lab devices. 
Before (~44s and ~31s respective)
[before.txt](https://github.com/librenms/librenms/files/1853922/before.txt)

After (~14s and ~11s respective)
[after.txt](https://github.com/librenms/librenms/files/1853923/after.txt)

I didn't change the method that is used to insert the objects into the database just the method that the data is gathered. 

I also incorporated the fix for the v4 address field from a previous PR into the pool member IP address as it seems that the issue made it's way to there as well. 

If there is a better way to do the array polling please let me know.  I found an snmpget function that takes in an arbitrary array of OID's but I didn't see one for snmpwalk.


murrant:
net-snmp snmpwalk only allows one oid.

Define an array that has all the names and oids in it then loop over it in a foreach loop.  It will be a lot easier to read.

includes/discovery/loadbalancers/f5-ltm.inc.php
```diff
@@ -44,38 +44,56 @@
...
+    name => '1.3.6.1.4.1.3375.2.2.10.1.2.1.1',
+    ip => '1.3.6.1.4.1.3375.2.2.10.1.2.1.3',
+    port => '1.3.6.1.4.1.3375.2.2.10.1.2.1.6',
+    defaultpool => '1.3.6.1.4.1.3375.2.2.10.1.2.1.19',
+    state => '1.3.6.1.4.1.3375.2.2.10.13.2.1.2',
+    errorcode => '1.3.6.1.4.1.3375.2.2.10.13.2.1.5',
+);
+
+foreach ($ltmVirtualServOID as $key => $value) {
+    $ltmVirtualServEntry[$key] = snmpwalk_array_num($device, $value, 0);
```
murrant: I would define `$ltmVirtualServEntry = [];` before the foreach loop (makes sure you don't have variable collisions)

Then you would also need to change the !is_null() calls to !empty() calls.

laf:
I know you've refactored the code to loop through an array to do snmpwalks but looking at the code it seems like if the name part doesn't contain data then the rest of the data wouldn't be used, seems to make sense to skip the loop if name contains no data.

dnapper:
I'll take a look at the device behaviour when there are no objects. I think i have a dev device around that I can factory reset.

dnapper:
Looks like if there are no objects it does return nothing.
I wasn't sure since it's a default license on the device and the module is always active. 

SNMP[/usr/bin/snmpbulkwalk -v2c -c COMMUNITY -Osqn -M /opt/librenms/mibs:/opt/librenms/mibs/f5 udp:HOSTNAME:161 1.3.6.1.4.1.3375.2.2.10.1.2.1.1]
.*.4.1.33*.* No Such Instance currently exists at this OID

SNMP[/usr/bin/snmpbulkwalk -v2c -c COMMUNITY -Osqn -M /opt/librenms/mibs:/opt/librenms/mibs/f5 udp:HOSTNAME:161 1.3.6.1.4.1.3375.2.2.5.1.2.1.1]
.*.4.1.33*.* No Such Instance currently exists at this OID

PR #8485, opened 2018-03-28
Override default device grouping radius

jepke:
Added an option to override the default group_radius 80. Usefull in fullscreenmap

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/common/worldmap.inc.php
+++ b/html/includes/common/worldmap.inc.php
@@ -124,7 +124,7 @@ if ($config['map']['engine'] == 'leaflet') {
         if (!empty($widget_settings['group_radius'])) {
             $group_radius = $widget_settings['group_radius'];
         } else {
-            $group_radius = Config::get('leaflet.group_radius', 80;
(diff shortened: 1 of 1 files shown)


html/includes/common/worldmap.inc.php
```diff
@@ -121,6 +121,8 @@
         }
         if (!empty($widget_settings['group_radius'])) {
             $group_radius = $widget_settings['group_radius'];
+        } elseif (!empty($config['custom_worldmap'])) {
+            $group_radius = $config['custom_worldmap']['group_radius'];
         } else {
             $group_radius = 80;
```
murrant: You can use this instead  (Config is our new config access class):
```
use LibreNMS\Config;
if (!empty($widget_settings['group_radius'])) {
    $group_radius = $widget_settings['group_radius'];
} else {
    $group_radius = Config::get('leaflet.group_radius', 80);
}
```

You also need to add it to the docs.  https://docs.librenms.org/Extensions/World-Map/

Looks like some other leaflet settings are missing there if you don't mind adding them.

PR #8496, opened 2018-04-01
Avoid XML Errors due to additional empty line before output

elberfeld:
The additional blank line before the xml declaration causes xml errors on some systems. 
This can be solved by removing the xml declaration.

Tested with Firefox 60.0b6 and Chrome 65 on Windows 10

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/graph-realtime.php
+++ b/html/graph-realtime.php
@@ -73,7 +73,7 @@ $height=125;            //SVG internal height : do not modify
 $width=300;             //SVG internal width : do not modify
 
 /********* Graph DATA **************/
-print('<?xml version="1.0" encoding="iso-8859-1"?>' . "\n");?>
+?>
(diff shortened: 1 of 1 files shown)


laf:
So version = 1.0 is the default without this anyway but encoding is `encoding="utf-8"` as a default. I'm not sure why `iso-8859-1` is used here though but it does mean the default for us changes.

I don't see a newline though before the definition line so I wonder if the fix needs to be somewhere else. Can you show the output you see from your browsers developer console?

elberfeld:
2018-04-14 23_38_18-http___warpsrvint_8088_graph-realtime php_type bits id 90 interval 2

As shown in the source view there is an additional newline.
No idea where this newline comes from, i can't see a couse within the code.

laf:
Can you check if you have a blank line in all other pages?

murrant:
Maybe a file that has a `?>`?

laf:
We don't seem to have any files that would be included that do.

murrant:
I couldn't seem to reproduce this and I don't think this would be the proper fix anyway.

If you can find the root cause, please let us know.

PR #8499, opened 2018-04-02
webui: Use toastr for operational messages on alert templates page

aldemira:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/modal/alert_template.inc.php
+++ b/html/includes/modal/alert_template.inc.php
@@ -225,11 +225,8 @@ function alertTemplateAjaxOps(template, name, template_id, title, title_rec)
         dataType: "html",
         success: function(msg){
             if(msg.indexOf("ERROR:") <= -1) {
-                $("#message").html('<div class="alert alert-info">'+msg+'</div>');
+                toastr.success(msg);
                 $("#alert-template").modal('hide');
-                setTimeout(function() {
-                    location.reload(1);
-                }, 1000);
             } else {
                 $("#error").html('<div class="alert alert-danger">'+msg+'</div>');
             }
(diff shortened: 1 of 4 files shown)


html/includes/modal/alert_template.inc.php
```diff
@@ -225,11 +225,8 @@ function alertTemplateAjaxOps(template, name, template_id, title, title_rec)
         dataType: "html",
         success: function(msg){
             if(msg.indexOf("ERROR:") <= -1) {
-                $("#message").html('<div class="alert alert-info">'+msg+'</div>');
+                toastr.success(msg);
                 $("#alert-template").modal('hide');
-                setTimeout(function() {
```
laf: It's not a huge thing but this stops the page reloading which means changing the template name won't show in the webui.

You could just update the row with the new name to avoid the reload, it's the only element that would change.

Otherwise, we'll be doing an overhaul on the templating system so you could leave this PR until that's done and we can take care of it all then.

aldemira: If this is going to create extra work for you I'm happy to close the PR. I'll wait for the overhaul, and I can fix it afterwards as well, doesn't matter.

laf: It's no extra work but for this PR to be merged you either need to re-add the setTimeout call or dynamically update the templates table with the updated template name after editing (the former is easier and fine)

html/includes/forms/alert-templates.inc.php
```diff
@@ -44,39 +43,49 @@
...
 } else {
-    $error = "You haven't given your template a name, it feels sad :( - $name";
+    $error = true;
+    $message = "You haven't given your template a name, it feels sad :( - $name";
 }
 
-if (!empty($ok)) {
-    die("$ok");
+if ($error) {
+    $status = array('status' => 1, 'message' => $message);
```
laf: Can you set status to ok or error as we do elsewhere.

PR #8507, opened 2018-04-04
Feature: Alert Transport Mapping

vivia11:
Hello all,

First off, I would like to say that I don't think this is even close to deployment. There's lots of work to be done before it can even be considered close. I felt like it would a waste of time for me to fully flesh out a feature if it didn't meet the needs of the community.

The feature allows for a user to add "Alert Contacts" which, right now, are emails associated with a contact name. A user can then associate a rule to specific alert contacts which would override any existing external transports. And only those alert contacts would be sent an alert.

Obviously, this needs a lot of work to incorporate the other external transports. There is currently one table (transport_email) which deals with the mail transport. Nothing exists for the other transports. (I work only with email, so that's the reason for the focus).  **Should we have a table for each transport - for users who wish to have multiple jira/ciscospark/slack etc transports?**

I haven't even bothered with adding in a update/delete feature to alert contacts. Right now the add button sits on the same page as the alert rules. **Should we have a separate page for alert contacts? Would it go under Alerts or Settings?**

Certain portions of the code are written with email in mind and not the other transports. These will probably have to change to accommodate the other transports.
Any suggestions/discussions about potential database schemas or the code itself is welcome.

**Note: Since I have not added a delete feature, do not test this code without knowing how to edit databases/tables.**

Screenshots anyone?
image

image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
This is a good start I've got a few things for you:
1. One table shared between all alert transports.
2. I've planned to allow multiple definitions per transport which would require a table to hold configured transports.  The reason I mention it is alert_contacts would need to be mapped to the correct configuration, not just a generic transport type.  Something to keep in mind, but probably something for a separate PR ;)
3. Visually, I think having an icon for the transport and then the contact description shown should be nice.

PR #8512, opened 2018-04-05
refactor: Use snmp.version config option to allow users to set versions available

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

`$config['snmp']['version']` was unused before. This allows people to disable snmp versions they don't need.


murrant:
Leave it in defaults, just commented.

trcyberoptic:
initial refector works for me, thanks!

laf:
> @laf Instead of setting in defaults, use the default option of Condig::get(). Put v2c first by default, as we try them in order, and trying other versions slows it down a lot. Plus we don't want to add devices by v1 if we can help it.

Maybe I should just drop the webui support in this PR then I can default this back to just v2c but it at least allows people to customise it.

Why set the defaults in Config::get() as you'd have to do it each time it's called (which would only be once now but probably not in the future?)

laf:
I've removed the addhost.php changes so this now maintains backwards compatibility.

PR #8522, opened 2018-04-08
update docs to reflect SMART monitoring having the ability to use either device name or disk serial number

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Update the docs to reflect the fix in https://github.com/librenms/librenms-agent/pull/168.

### DIFF
--- a/doc/Extensions/Applications.md
+++ b/doc/Extensions/Applications.md
@@ -843,7 +843,7 @@ and modify your snmpd.conf file accordingly:
 extend smart /usr/bin/sudo /etc/snmp/smart
 ``` 
 
-If you set useSN to 1, it is worth noting that you will loose history for that device each time you change it. You will also need to run camcontrol or the like on said server to figure out what device is actually corresponds with that serial number.
+If you set useSN to 1, it is worth noting that you will loose history for that device each time you change it. You will also need to run camcontrol or the like on said server to figure out what device actually corresponds with that serial number.
 
 ### Squid
 
--- a/doc/Extensions/Applications.md
+++ b/doc/Extensions/Applications.md
@@ -812,6 +812,7 @@ Anything starting with a # is comment. The format for variables is $variable=$va
 #This is a comment
 cache=/var/cache/smart
 smartctl=/usr/bin/env smartctl
+useSN=0
 ada0
 ada1
 ```
@@ -820,6 +821,7 @@ The variables are as below.
 ```
 cache = The path to the cache file to use. Default: /var/cache/smart
 smartctl = The path to use for smartctl. Default: /usr/bin/env smartctl
+useSN = If set to 1, it will use the disks SN for reporting instead of the device name.
 ```
 
 If you want to guess at the configuration, call it with -g and it will print out what it thinks
@@ -840,6 +842,8 @@ and modify your snmpd.conf file accordingly:
 ```
 extend smart /usr/bin/sudo /etc/snmp/smart
 ``` 
+
+If you set useSN to 1, it is worth noting that you will loose history for that device each time you change it. You will also need to run camcontrol or the like on said server to figure out what device is actually corresponds with that serial number.
 
 ### Squid
 
--- a/doc/Extensions/Applications.md
+++ b/doc/Extensions/Applications.md
@@ -812,7 +812,7 @@ Anything starting with a # is comment. The format for variables is $variable=$va
 #This is a comment
 cache=/var/cache/smart
 smartctl=/usr/bin/env smartctl
-useSN=0
+useSN=1
 ada0
 ada1
 ```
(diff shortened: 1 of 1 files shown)


laf:
It doesn't allow both though does it, I thought the script was updated to use the serial number solely.

VVelox:
@laf This is what https://github.com/librenms/librenms-agent/pull/168 is for. To allow either to be used.

The idea for serial number only was a major mistake.

Using serial number means loosing information each time a disk is replaced.

It also makes it very problematic to see if slot or line is going bad as it prevents historic analysis.

Also once in place in systems disks very rarely change place in systems. The idea of RAID cards and the like changing a lot in systems is not something that happens in production. If it happens, it is usually replacing a bad card, which results in nothing changing as far as device numbering or like.

enilfodne:
>The idea for serial number only was a major mistake.

I fail to see this as a mistake, but to each their own.

> Using serial number means loosing information each time a disk is replaced.

The data is still there, nothing is lost.

> It also makes it very problematic to see if slot or line is going bad as it prevents historic analysis.

I think this falls out of the scope of this application. This app is for tracking `S.M.A.R.T` data, not analysis or data aggregation.

> Also once in place in systems disks very rarely change place in systems. The idea of RAID cards and the like changing a lot in systems is not something that happens in production. If it happens, it is usually replacing a bad card, which results in nothing changing as far as device numbering or like.

A drive SN will never change (unless you modify the firmware), thus you can guarantee that the data you're looking at is only relevant to that drive. You can actually change the drive identifiers quite easily in VMs or when replacing disks in a ZFS pool.

VVelox:
> I think this falls out of the scope of this application. This app is for tracking S.M.A.R.T data, not analysis or data aggregation.

Wow. So you are telling me what the purpose of said app I wrote was.

Interesting and way to be a complete jerk.

No. Nothing about what I am saying falls outside of what I wrote this for. This is exactly what I wrote it for, tracking and displaying useful info on SMART data.

> A drive SN will never change (unless you modify the firmware), thus you can guarantee that the data you're looking at is only relevant to that drive. You can actually change the drive identifiers quite easily in VMs or when replacing disks in a ZFS pool.

Don't give a rats ass about VMs here as they are not relevant.

On top of this the if SN are applicable to you then you can use it. I don't see why you are getting so pissy about making it optional.

From a server administrative standpoint device ID is much more relevant as you need to track this over time. I don't care about any one specific drive. These get replaced.

On top of you dodged the issue of the supposed reason for this was changing RAID cards which is a rarity for when it comes to servers and doing so it is likely going to be replaced with the exact same time, resulting in stuff staying the same.

This is utterly akin to dumping graphs when the MAC of a network device changes.

> This is factually wrong. History is never lost, nothing is deleted.

No. But it is lost from a effective standpoint a you have to fuck around with renaming stuff and the like.

enilfodne:
> Wow. So you are telling me what the purpose of said app I wrote was.
>
> Interesting and way to be a complete jerk.
> 
> No. Nothing about what I am saying falls outside of what I wrote this for. This is exactly what I wrote it for, tracking and displaying useful info on SMART data.

Thank you for work and apologies if you interpreted this as snide comment. My point was about the project in general - LibreNMS is used to gather and record data. As such i don't see how troubleshooting application fits into this category.

> Don't give a rats ass about VMs here as they are not relevant.
>
> On top of this the if SN are applicable to you then you can use it. I don't see why you are getting so pissy about making it optional.
>
> From a server administrative standpoint device ID is much more relevant as you need to track this over time. I don't care about any one specific drive. These get replaced.
>
> On top of you dodged the issue of the supposed reason for this was changing RAID cards which is a rarity for when it comes to servers and doing so it is likely going to be replaced with the exact same time, resulting in stuff staying the same.
> 
> This is utterly akin to dumping graphs when the MAC of a network device changes.

I think a fair amount of LibreNMS users do care about VMs, because that's what they monitor, myself included.

I'm not pissy nor do i take this personally. I'm looking at this from a user standpoint where the change was made a few months ago and now its reverted.

I haven't dodged your question. I don't have RAID cards, i don't use RAID. As such i can't give a competent opinion about it.

VVelox:
> Thank you for work and apologies if you interpreted this as snide comment. My point was about the project in general - LibreNMS is used to gather and record data. As such i don't see how troubleshooting application fits into this category.

Right and the purpose of this is to gather data and record said data to make any sort of trouble shooting easier.

> I think a fair amount of LibreNMS users do care about VMs, because that's what they monitor, myself included.

But this app is not relevant to that. And should you wish to monitor it like you were before, this change will make zero difference if you are doing it from VM for some reason.

> I'm not pissy nor do i take this personally. I'm looking at this from a user standpoint where the change was made a few months ago and now its reverted.

It it changed the original code in a very dramatic fashion. It will make no difference to those with it already in place at all. This is hardly a good excuse.

VVelox:
Also sorry if I am coming off a bit grumpy and bitter, it is just frustrating that such a massive change was made and it was made non-togglable. Running on no sleep(yeah, bad reason for being grummpy, I know:()

Think of it this way, it is akin replacing all the device names in the interface with MAC addresses and expecting it to be useless... then if it changes dumping it if a bad card is replaced.

VVelox:
W00t! All good now and should make every one happy. :)

Now uses the SN and expanded the docs to reflect that.

PR #8532, opened 2018-04-09
refactor: Allow _perc and _current columns to be string for alert rules

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: https://community.librenms.org/t/trouble-with-storage-in-new-alert-rules-format/3724

### DIFF
--- a/LibreNMS/Alerting/QueryBuilderFilter.php
+++ b/LibreNMS/Alerting/QueryBuilderFilter.php
@@ -101,7 +101,7 @@ class QueryBuilderFilter implements \JsonSerializable
                     continue;
                 }
 
-                $type = $this->getColumnType($column_type, $column);
+                $type = $this->getColumnType($column_type);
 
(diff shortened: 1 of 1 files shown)


murrant:
Why are you making this change?  I'd prefer you make the change in the other part of the code, like where enum is handled.

laf:
It’s not for enum. It’s because currently these columns are restricted to integers but you need to also compare them to other columns.

murrant:
I don't understand "It’s not for enum".   I meant we need a standard way/place to define specific filter items.  That we way we can customize them and make it extra nice.  Right now there is only one custom thing, that is the enum one.  I expect we will add more...

murrant:
Yes, I feel better with it in that place.  But I do feel like we need a better way to handle "special cases".

I'd say put that here for now and we can maybe come up with something later if/when we add more.

PR #8534, opened 2018-04-10
Update alert rules to match auth failure on Cisco

sorano:
Cisco generates entires like these on authentication failure:

Login failed [user: ] [Source: ] [localport: 22] [Reason: Login Authentication Failed] at 08:09:04 cet Tue Apr 10 2018

The old regex would not match that since it looks for "failure" changing this to the more inclusive "fail" will catch Cisco failures while keeping current functionality.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/misc/alert_rules.json
+++ b/misc/alert_rules.json
@@ -113,7 +113,7 @@
     "name": "Too many wireless clients"
   },
   {
-    "rule": "%syslog.timestamp >= %macros.past_5m && %syslog.msg ~ \"@authentication failure@\"",
+    "rule": "%syslog.timestamp >= %macros.past_5m && %syslog.msg ~ \"@authentication fail@\"",
     "name": "Syslog, Authentication failure on Device"
   },
   {


sorano:
Might be worth adding that requires the following IOS config:

login on-failure log

sorano:
Sidenote: just noticed that this also catches 802.1x failures on Unifi AP's

kkrumm1:
You should make a new alert rule for Cisco devices. Since the original works with HPE devices

laf:
You can exclude OS from this rule but that feels hacky. What's the syslog output you see for this?

sorano:
HOSTAPD | ath1: STA 00:11:22:33:44:55 IEEE 802.1X: authentication failed - EAP type: 0 (unknown)

sorano:
Hello @laf 

Currently it looks like this:

^([^hostapd|^802\.1x]).*authentication fail.*

But this filters everything starting with any character inside the []group. So I wanted to filter it better with multiple [] but havent got around to it yet, hopefully I can look at it during this week.

murrant:
Sorano, I think you need to put each character in a [^] since you don't have negative look behind.

laf:
@sorano Let us know when you've got this updated.

PR #8535, opened 2018-04-10
api: Add option to output graphs as base64

pheinrichs:
Let me know what you would like updated. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -212,6 +212,7 @@ function get_graph_generic_by_hostname()
     $sensor_id    = $router['sensor_id'] ?: null;
     $vars         = array();
     $vars['type'] = $router['type'] ?: 'device_uptime';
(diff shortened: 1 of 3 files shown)


html/includes/api_functions.inc.php
```diff
@@ -244,6 +245,10 @@ function get_graph_generic_by_hostname()
     rrdtool_initialize(false);
     include 'includes/graphs/graph.inc.php';
     rrdtool_close();
+
+    if ($vars['output'] === 'base64') {
+        api_success(['image' => $base64_output, 'content-type' => get_image_type()], 'image');
```
laf: Do you need to add this to the other functions which return graphs?

pheinrichs: Yep, does the format and names seem ok? If so, i'll append it to the port methods as well

laf:
Yeah seems fine to me :)

PR #8538, opened 2018-04-10
Fix device group rule suggest, Auth before init

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/ajax_rulesuggest.php
+++ b/html/ajax_rulesuggest.php
@@ -25,13 +25,12 @@
 
 use LibreNMS\Authentication\Auth;
 
-session_start();
+$init_modules = array('web');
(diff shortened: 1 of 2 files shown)


laf:
I get `Unauthorized.` when trying to access the ajax_rulesuggest.php directly

html/ajax_rulesuggest.php
```diff
@@ -25,14 +25,13 @@
 
 use LibreNMS\Authentication\Auth;
 
-session_start();
```
laf: Was this a mistake? Doesn't work without it.

Soljia:
Works good for me! +1

PR #8541, opened 2018-04-10
devices: ibm-dpi os discovery

vivia11:
Hello all,

Tested with model IBM switched C19/C13 .

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/ibm-dpi.inc.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
+$version = end(explode(' ', trim(snmp_get($device, "sysDescr.0", "-OQv", "SNMPv2-MIB"))));
```
laf: We already have sysDescr value in `$device['sysDescr']`

vivia11: Mine are rebranded Lenovo PDUs actually. Maybe they're all the same now?

No MIB files that I can find. I've emailed IBM/Lenovo and we'll see if there are even any.

vivia11:
Mine are rebranded Lenovo PDUs actually. Maybe they're all the same now?

No MIB files that I can find. I've emailed IBM/Lenovo and hopefully they have MIBs for this (which I think they do).

laf:
This might be better to call it just dpi.

vivia11:
@laf 
After some digging, this is a model that is still owned by IBM, they apparently sold the other models of this (to Lenovo) type but withdrew this one and didn't sell it.

There are also no MIBs for this, confirmed by IBM.

Issue #7910, opened 2017-12-13
"From name" not used in mail transport?


olds463:
Opening issue instead of pull request as i'd like some of the regular contributors to confirm. 

It seems in the WebUI for "Email Transport" the "From name" does not get used. 

In the send_mall() function in includes/functions.php 

`        foreach (parse_email($config['email_from']) as $from => $from_name) {
            $mail->setFrom($from, $from_name);
        }
`

You are not using $config['email_name'] instead are looking to call parse_email which is expecting the "From email address" in WebUI ($config['email_from'] to be in the format of: 

`LibreNMS <user@domain.com>`

Is this field obsolete in the WebUI?

murrant:
Looks like an oversight.  It should probably try both. Feel free to send a fix ;)

PR #8550, opened 2018-04-11
fix: Use email_user variable as from name for emails if available

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I wouldn't call this a bug but fixed anyway. Fixes: #7910

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -933,7 +933,7 @@ function send_mail($emails, $subject, $message, $html = false)
 
         foreach (parse_email($config['email_from']) as $from => $from_name) {
             if (empty($from_name)) {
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -932,6 +932,9 @@ function send_mail($emails, $subject, $message, $html = false)
         $mail->Hostname = php_uname('n');
 
         foreach (parse_email($config['email_from']) as $from => $from_name) {
+            if (empty($from_name)) {
+                $from_name = Config::get('email_user', null);
```
murrant: null is the default for "default", so no need to specify it here.

PR #8554, opened 2018-04-12
Add 'dns name' and 'description' (ifAlias) in fdb search

wiad:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/table/fdb-search.inc.php
+++ b/html/includes/table/fdb-search.inc.php
@@ -5,7 +5,7 @@ use LibreNMS\Authentication\Auth;
 $param = array();
 
 $select = "SELECT `F`.`port_id` AS `port_id`, `F`.`device_id`, `ifInErrors`, `ifOutErrors`, `ifOperStatus`,";
-$select .= " `ifAdminStatus`, `ifAlias` AS `interface`, `ifDescr`, `mac_address`, `V`.`vlan_vlan` AS `vlan`,";
+$select .= " `ifAdminStatus`, `ifAlias`, `ifDescr`, `mac_address`, `V`.`vlan_vlan` AS `vlan`,";
 $select .= " `hostname`, `hostname` AS `device` , group_concat(`M`.`ipv4_address` SEPARATOR ', ') AS `ipv4_address`";
 
 $sql  = " FROM `ports_fdb` AS `F`";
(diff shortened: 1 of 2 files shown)


wiad:
As described in 
https://community.librenms.org/t/add-columns-for-dns-name-and-ifalias-in-fdb-search/3737/8

html/includes/table/fdb-search.inc.php
```diff
@@ -106,8 +116,10 @@
         }
         if ($entry['port_id'] == $endpoint_portid) {
             $endpoint_img = "<i class='fa fa-star fa-lg' style='color:green' aria-hidden='true' title='This indicates the most likely endpoint switchport'></i>";
+            $dnsname = gethostbyaddr(reset(explode(',', $entry['ipv4_address']))) ?: 'N/A';
```
laf: Do you need to put this in a function which caches the response otherwise you might just be sending the same query if the IP is listed multiple times.

laf:
I can't test this as my dev env doesn't have any IP info which would be impacted from this. It doesn't slow it down though so that's a +.

My work install is on 1.38 and this patch doesn't apply so can't test there.

The code looks sane although you will probably need to cache the data for the hostname look up as per my comment.

html/includes/table/fdb-search.inc.php
wiad: Not sure I understand correctly, but if there are multiple adresses in `$entry['ipv4_address']` the function `reset` will make sure only one is used.

wiad: Not sure if I understand correctly, but if there are multiple ip addresses in `$entry['ipv4_address']` the `reset` function will make sure we only use one of them (the first).

laf: The results in the query can have the same ip in multiple rows. I realise with reset you are getting just one but the next row might have the same IPs in and the lookup would be done again

wiad: hm, the gethostbyaddr lookup is only performed if `$entry['port_id'] == $endpoint_portid`, and `$endpoint_portid` can only be one. So unless there are multiple rows with the same  `$entry['port_id']` in the query result I don't see how that would happen.

laf: True, ignore me then :)

laf: I'd definitely be interested to see someone use this with a good amount of results returned to understand the implications.

wiad: well, I really appreciate your feedback! You have far better understanding of the code base so it's really valuable. And I would also like more testing of this in other environments...

html/includes/table/fdb-search.inc.php
```diff
@@ -35,16 +35,26 @@
...
     } elseif (isset($vars['searchby']) && $vars['searchby'] == 'ip') {
+        $ip = $vars['searchPhrase'];
+        $ip_search = '%'.mres(trim($ip)).'%';
         $sql .= " LEFT JOIN `ipv4_mac` AS `M` USING (`mac_address`)";
         $where  .= ' AND `M`.`ipv4_address` LIKE ?';
         $param[] = $ip_search;
+    } elseif (isset($vars['searchby']) && $vars['searchby'] == 'dnsname') {
+        $ip = gethostbyname($vars['searchPhrase']);
+        $ip_search = '%'.mres(trim($ip)).'%';
+        $sql .= " LEFT JOIN `ipv4_mac` AS `M` USING (`mac_address`)";
```
murrant: Search by says 'dnsname', but you are searching the ipv4_mac table.

wiad: So what would you call it? The dns name is resolved with gethostbyname to get an ip address which we can search for in lilbrenms db. So even if we end up by searching for ip address the search phrase the user inputs is a dns record name.

murrant: Yeah, I see now I missed that earlier, probably too sleepy.

I do wish you would try to avoid DNS lookups as much as possible though, they are slow. You could use our \LibreNMS\Util\IP class to test if it is an IP or not.

wiad: Yeah, good point about dns lookups. I try to avoid it as much as possible by only resolving one result (the top result). Avoiding it more would mean scrapping the feature altogether, which I of course could do but I really like it...  :)

Where would you suggest validating the ip would fit?

murrant: Many people add devices by DNS name.  It makes managing and searching for them easier.  If a histname is already a DNS name, why do a DNS lookup?

laf:
@librenms/reviewers Happy for this to be merged?

kkrumm1:
Looks good to me, i think this could be helpful.

PR #8558, opened 2018-04-13
ict-pdu os: Fixed polling for state sensor.

vivia11:
Removed if/else statement. Code was reading an index that was out of bounds (for sensor_index greater than 0) resulting in sensor_value of 0 which did not map to any state.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Any context around why this needs changing?

vivia11:
image

The if/else structure was causing the odd behaviour seen above. The first fuse port was correctly assigned a sensor_value, but all other fuse ports (due to the else structure) was given a sensor_value of 0 which does not map to anything.

The code was attempting to append '.0' to the oid and read the snmp value - but no such index exists. Not sure why the if/else structure was originally necessary, as the polling works fine without the if/else as seen below.

image

laf:
Luckily the chap who added this support used our pastebin service and posted his walk in the issue. I used it to test and you are correct this does fix it - however, removing the file also seems to. Can you check that and if so just remove the file please.

vivia11:
Removing the file also works for me.

murrant:
Any chance you could add test data for this device?

https://docs.librenms.org/#Developing/os/Test-Units/#example-workflow

laf:
I've added test data.

PR #8571, opened 2018-04-17
add json_app_get and convert fail2ban over to JSON

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This requires https://github.com/librenms/librenms-agent/pull/172 .


includes/polling/functions.inc.php
```diff
@@ -690,3 +690,75 @@ function convert_to_celsius($value)
...
+ * -4 : Empty return from snmp_get().
+ * -5 : Non-numeric version number.
+ *
+ * Other error codes are possible, but those are extend related and should be handled
+ * by the poller calling this function.
+ *
+ * If version, error, and errorString are not set, it is assumed it is a old JSON backend and
+ * sets them manually with a version of 0 and assumes a error of 0(no error).
+ */
+function json_app_get($device, $poid)
```
murrant: You can accept a string instead of an OID for this function.  Just use `string_to_oid()`

includes/polling/functions.inc.php
```diff
@@ -690,3 +690,75 @@ function convert_to_celsius($value)
...
+{
+    $returned_json = snmp_get($device, 'nsExtendOutputFull.'.$poid, '-O qv', 'NET-SNMP-EXTEND-MIB');
+
+    // make sure we actually get something back
+    if (empty($returned_json)) {
+        $parsed_json=json_decode('{"error":"-4","errorString":"Empty return from snmp_get()"}', true);
```
murrant: You could just use a native array here: `['error'=>'-4', 'errorString'=>'Empty return from snmp_get()']`

You could also throw an exception, new Exceptions can be created in `LibreNMS\Exceptions`.

includes/polling/functions.inc.php
```diff
@@ -690,3 +690,75 @@ function convert_to_celsius($value)
...
+        return $parsed_json;
+    }
+
+    $parsed_json=json_decode(stripslashes($returned_json), true);
+
+    if (json_last_error() === JSON_ERROR_NONE) {
+        if (empty($parsed_json)) {
+            // If we get here it means there are no keys in the array, meaining '{}' was was returned
+            $parsed_json{'error'}='-3';
```
murrant: Please use this style to access arrays in PHP `$parsed_json['error']`

includes/polling/functions.inc.php
```diff
@@ -690,3 +690,75 @@ function convert_to_celsius($value)
...
+            $parsed_json{'errorString'}='Blank JSON returned.';
+        } else {
+            // If the following is true, it is a legacy JSON app extend, meaning these are not set
```
murrant: You should probably return the raw output so we can process legacy apps without another snmpget.

includes/polling/functions.inc.php
VVelox: It is native. Just less common. Sure. I'll convert it over to from curly to square. :)

I'll look into the exceptions bit.

includes/polling/functions.inc.php
VVelox: This does not cover non-JSON stuff. That will require converting the backends and updating the pollers.

The legacy stuff AFAIK at this point is only ZFS as those variables are not present in the returned JSON.

While also for that, it prevents possibly breakages from happening if the backend errors and does not generates valid JSON, but not with some of the expected and commonly used keys.

Actually while I am thinking of this, I should add in a third option for this all for minimum return version, causing it to error when that happens.

includes/polling/functions.inc.php
murrant: No I mean you wrote a json array then convert it to a php array.  Just use a php array.

Well aware of using curly braces for array access, like I said our convention is square brackets.

VVelox: Doh! :)

Yeah. I see it now. Good point. =^.^=

murrant:
@VVelox Almost ready to merge, but this needs to support falling back to the existing data formats.  Otherwise it will break everyone's applications.

VVelox:
@murrant Hoi! Your thoughts on my latest update to the json_app_get function? Making the exception stuff optional. This will allow using exceptions on new stuff and shimming in code to support un-updated backends on converted ones.

murrant:
Maybe we should not be checking the version in this function and leave that up to the parser...

VVelox:
I see no reason with using an exception for this now. We can return the info and allow it to be processed as needed, if it is still able to proceed.

Also whoops. Speaking of which, just realized that I needed to fix the poller. The only time it should attempt to parse it the old way is if parsing it as JSON fails. Now only does so if -5 is returned.

murrant:
I have a suggestion for the Exceptions.  They should all extend one base exception so you can catch all json app exceptions at once if you choose to.

PR #8572, opened 2018-04-17
refactor: Updated the oxidized script

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Username is not always available so make it optional.

No need for librenms version so scrap that off.

Cut down function calls and avoid setting temp $user variable.

### DIFF
--- a/scripts/syslog-notify-oxidized.php
+++ b/scripts/syslog-notify-oxidized.php
@@ -4,20 +4,17 @@
 $init_modules = [];
 require __DIR__ . '/../includes/init.php';
 
+use LibreNMS\Config;
+
 function oxidized_node_update($hostname, $msg, $username = 'not_provided')
 {
-    global $config;
-
     // Work around https://github.com/rack/rack/issues/337
     $msg = str_replace("%", "", $msg);
     $postdata = ["user" => $username, "msg" => $msg];
-
-    $curl = curl_init();
-    curl_setopt($curl, CURLOPT_CUSTOMREQUEST, "PUT");
-    curl_setopt($curl, CURLOPT_POSTFIELDS, json_encode($postdata));
-    curl_setopt($curl, CURLOPT_USERAGENT, 'librenms');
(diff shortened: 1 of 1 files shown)


scripts/syslog-notify-oxidized.php
```diff
@@ -1,47 +1,37 @@
...
     $msg = str_replace("%", "", $msg);
-    $postdata = array("user" => $username, "msg" => $msg);
-
-    $version = `git rev-parse --short HEAD`;
-    if ($version != "") {
-        $version = "/".$version;
-    }
+    $postdata = ["user" => $username, "msg" => $msg];
 
     $curl = curl_init();
```
murrant: @laf I noticed we pull in http://requests.ryanmccue.info/ but it is only used for peeringdb.
Thoughts on using this all over instead of curl?

murrant:
Looks good, just a tangential question...

scripts/syslog-notify-oxidized.php
laf: Good shout, updated + used `Config::get()`

scripts/syslog-notify-oxidized.php
```diff
@@ -1,47 +1,34 @@
...
     $msg = str_replace("%", "", $msg);
-    $postdata = array("user" => $username, "msg" => $msg);
-
-    $version = `git rev-parse --short HEAD`;
-    if ($version != "") {
-        $version = "/".$version;
+    $postdata = ["user" => $username, "msg" => $msg];
+    $oxidized_url = Config::get('oxidized.url');
+    if (!empty($oxidized_url)) {
+        Requests::put("$oxidized_url/node/next/$hostname", [], json_encode($postdata), ['proxy' => get_proxy()]);
```
murrant: You sure the json_encode() is correct here?

murrant: Nevermind.

PR #8573, opened 2018-04-17
convert zfs over to use json_app_get

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This requires https://github.com/librenms/librenms/pull/8571 .


includes/polling/applications/zfs.inc.php
```diff
@@ -2,16 +2,17 @@
...
 $app_id = $app['app_id'];
-$options = '-O qv';
-$mib = 'NET-SNMP-EXTEND-MIB';
-$oid = 'nsExtendOutputFull.3.122.102.115';
-$json = snmp_get($device, $oid, $options, $mib);
 
-$zfs=json_decode(stripslashes($json), true);
+echo $name;
+
+$zfs=json_app_get($device, '3.122.102.115');
```
murrant: I think it would better to just use a string as the second argument.

Let me just drop this here:
```
function str2index($string)
{
    $ascii = array_map('ord', str_split($string));
    array_unshift($ascii, strlen($string));
    return implode('.', $ascii);
}
```

includes/polling/applications/zfs.inc.php
```diff
@@ -2,16 +2,17 @@
...
+if ($zfs{'error'} != '0') {
```
murrant: Our preferred style for array access in PHP is `$zfs['error']`

includes/polling/applications/zfs.inc.php
VVelox: I am not sure what you are trying to accomplish with this.

VVelox: Ahh! Okay! I see what you mean now. By string you mean "zfs" instead of '3.122.102.115'.

VVelox:
This requires https://github.com/librenms/librenms-agent/pull/179 and has been updated to work with the old version of the script as well.

VVelox:
Well I guess does not require as it will work with the old version, but that is the PR for updating the extend though.

VVelox:
@murrant W00t! Added and all good now. :)

murrant:
@VVelox The test data contains an extra comma (causing the tests to be skipped, fixed in #9015) 
Please fix the json, and the code because the tests fail.

VVelox:
@murrant Doh! Whoops. Fixed. Base64 encoded it instead of hex.

Thanks. :)

tests/data/linux_zfs-legacy.json
```diff
@@ -0,0 +1,409 @@
...
+                    "value": "76",
+                    "value_prev": null,
+                    "app_type": "zfs"
+                },
+                {
+                    "metric": "pools_arc_cap",
+                    "value": "45",
+                    "value_prev": null,
+                    "app_type": "zfs"
+                },
```
murrant: Comma here needs to be removed.

tests/data/linux_zfs-v1.json
```diff
@@ -0,0 +1,391 @@
...
+                    "value": "76",
+                    "value_prev": null,
+                    "app_type": "zfs"
+                },
+                {
+                    "metric": "pools_arc_cap",
+                    "value": "45",
+                    "value_prev": null,
+                    "app_type": "zfs"
+                },
```
murrant: Comma here needs to be removed.

laf:
also got a merge conflict that needs fixing.

murrant:
I can fix that. It is because I fixed the use of unquoted strings and changed to short array syntax.

Issue #8567, opened 2018-04-16
ePMP GPS up in radio web interface, down in Libre


mikerev:
**DO NOT DELETE THIS INFORMATION.**

> **Please read this information carefully.**

**GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z.** 

**You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).**

**If you have a feature request/changes please post this on our [community site](https://community.librenms.org/c/feature-requests).**

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [ ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ ] Please provide **ALL** info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ ] Please provide as much detail as possible.
- [ ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

https://p.libren.ms/view/6ee0ec65
https://p.libren.ms/view/0d226761
https://p.libren.ms/view/34574850

![kish](https://user-images.githubusercontent.com/10117960/38838828-a6b6b394-419d-11e8-9778-1c7c50a1b079.png)

mikerev:
@pheinrichs ?

laf:
This issue has been closed as it isn't classed as a bug or a new device request.

For help please use our Discord server or the community site: http://docs.librenms.org/Support/FAQ/#faq3

mikerev:
Why was this closed?

PR #8575, opened 2018-04-17
fix: ePMP gps state

pheinrichs:
fixes: #8567

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/discovery/epmp.yaml
+++ b/includes/definitions/discovery/epmp.yaml
@@ -13,7 +13,7 @@ modules:
                 -
                     oid: cambiumEffectiveSyncSource
                     num_oid: '.1.3.6.1.4.1.17713.21.1.1.7.'
-                    index: 0
+                    index: '{{ $index }}'
                     descr: GPS Status
                     states:
                         - { value: 1, generic: 0, graph: 1, descr: GPS Sync Up }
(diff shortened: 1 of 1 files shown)


pheinrichs:
Any idea why this would have stopped working? The above does resolve, just curious is all.

laf:
The `{{ $index }}` at the end isn't needed (never has and I'm not sure if it would have done anything or caused issues).

Specifying the index of 0 shouldn't be needed and `{{ $index }}` should work fine there - are you sure you needed to change that?

murrant:
@laf `{{ $index }}` isn't needed here, I allowed it in the new processors stuff.  I'll try to make it all consistent sometime.

pheinrichs:
Gotcha, so it was probably running 21.1.1.7.{{$index}}.{{$index}}'. So it should be okay like this?

laf:
Yes that should be fine although I think you should do `index: {{ $index }}` instead of `index: 0`

laf:
> @laf {{ $index }} isn't needed here, I allowed it in the new processors stuff. I'll try to make it all consistent sometime.

@murrant Would be good to standardise them although I don't think processors supports not having the `{{ $index }}` right now - didn't work for me earlier at least.

laf:
@pheinrichs Does that change to index still work for you so we can merge?

pheinrichs:
@laf Rediscovered and polled were a success. Good to go.

PR #4005, opened 2016-08-08
CISCO-NTP-MIB

adaniels21487:
A new Discovery/Poller module to collect NTP statistics from devices which support the CISCO-NTP-MIB

- Discovered peers are stored using components
- Statistics are displayed using 'applications'
- A critical alarm is raised when a stratum of 16 is reported.

PR #8580, opened 2018-04-18
device: Add NTP support for Allied Telesis AWPlus products

mattie47:
awplus: Add NTP support  …
- Fixed polling issues for awplus. Now offset and delay graphs work.

- html/includes/graphs/device/ntp_*.inc.php
--  Removed extra "\" from MAX count which showed up in Graph legend

- html/includes/graphs/device/ntp_delay.inc.php
-- Present legend in either seconds, or milliseconds (depending on vendor)

- html/includes/graphs/device/ntp_dispersion.inc.php
-- Present legend in either seconds, or milliseconds (depending on vendor)

- html/includes/graphs/device/ntp_offset.inc.php
-- Removed rrd minimum (-l 0). This is because offset can be below 0.
--- It's better for the graph to auto scale than set a fixed minimum to display here.

- html/includes/graphs/device/ntp_stratum.inc.php
-- Changed graph to go from 0 to max of 16. NTP Stratum can only be 0 to 16.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


mattie47:
So this works and is ready for someone to have a look over it. 

As you can see, I modeled this off the Cisco NTP support done under #4005 

I also made some minor modifications to the front end graphs. 

Here's a couple screenshots:

image
 
image



Thanks,

Matt

html/includes/graphs/device/ntp_delay.inc.php
```diff
@@ -19,9 +19,19 @@
 // We only care about our device id.
 $components = $components[$device['device_id']];
 
+//Specify whether value is ms or seconds.
+if ($device['os_group'] == 'cisco') {
+    $time=' (s)';
+    $virtlabel=' Seconds';
+} else {
+    $time = ' (ms)';
+    $virtlabel = ' Milliseconds';
```
laf: This needs fixing at poll time. Data should be stored the same no matter the device type so you'll need to convert the awplus from ms to s before storing it. The graph changes can then be reverted.

mattie47: OK. 

>Data should be stored the same no matter the device type so you'll need to convert the awplus from ms to s before storing it.

I would argue the cisco poller config here is in the wrong though.  

I've only got an old 1841 here but even the cisco CLI reports these stats in msec:

```
r1#sh ver | incl .bin
System image file is "flash:c1841-advipservicesk9-mz.151-4.M5.bin"
r1#sh ntp associations 

  address         ref clock       st   when   poll reach  delay  offset   disp
*~10.37.105.100   10.37.105.250    4     41     64   377  1.069 -25.587  5.027
 * sys.peer, # selected, + candidate, - outlyer, x falseticker, ~ configured
r1#sh nt
r1#sh ntp ?
  associations  NTP associations
  status        NTP status

r1#sh ntp st
Clock is synchronized, stratum 5, reference is 10.37.105.100 
nominal freq is 250.0000 Hz, actual freq is 250.0122 Hz, precision is 2**24
reference time is DE823257.2C3C884F (21:08:39.172 UTC Wed Apr 18 2018)
clock offset is -25.5874 msec, root delay is 56.56 msec
root dispersion is 163.06 msec, peer dispersion is 5.02 msec
loopfilter state is 'CTRL' (Normal Controlled Loop), drift is -0.000049008 s/s
system poll interval is 64, last update was 315 sec ago.
```

Funnily enough, the cisco NTP config doesn't work for me and reports most results as zero:

```
SNMP[/usr/bin/snmpbulkwalk -v2c -c COMMUNITY -Osqn -M /opt/librenms/mibs:/opt/librenms/mibs/cisco udp:HOSTNAME:161 .1.3.6.1.4.1.9.9.168.1.2.1.1]
.*.*.*.1.2.64158 1
.*.*.*.1.3.64158 *
.*.*.*.1.4.64158 123
.*.*.*.1.5.64158 *
.*.*.*.1.6.64158 123
.*.*.*.1.7.64158 0
.*.*.*.1.8.64158 4
.*.*.*.1.9.64158 4
.*.*.*.1.10.64158 6
.*.*.*.1.11.64158 6
.*.*.*.1.12.64158 -22
.*.*.*.1.13.64158 "00 00 00 0C "
.*.*.*.1.14.64158 "00 00 00 01 "
.*.*.*.1.15.64158 "
%i�"
.*.*.*.1.16.64158 "00 00 00 0F 00 00 00 08 "
.*.*.*.1.17.64158 "62 82 23 28 15 3F E6 D2 "
.*.*.*.1.18.64158 "DE 82 35 DE 1B C6 4A 5B "
.*.*.*.1.19.64158 "DE 82 35 DE 1B 76 C4 D0 "
.*.*.*.1.20.64158 853488881
.*.*.*.1.21.64158 255
.*.*.*.1.22.64158 0
.*.*.*.1.23.64158 "ew	D"
.*.*.*.1.24.64158 "62 82 24 B0 "
.*.*.*.1.25.64158 "ew

"
.*.*.*.1.26.64158 8
.*.*.*.1.27.64158 1
.*.*.*.1.28.64158 853488881
.*.*.*.1.29.64158 2

RRD[update /opt/librenms/rrd/10.37.105.110/ntp-10.37.105.100.rrd N:4:0:0:0]


Component: 92
    Index:      64158
    Peer:       10.37.105.100:123
    Stratum:    1.3.6.1.4.1.9.9.168.1.2.1.1.9.64158  = 4
    Offset:     1.3.6.1.4.1.9.9.168.1.2.1.1.23.64158 = 0
    Delay:      1.3.6.1.4.1.9.9.168.1.2.1.1.24.64158 = 0
    Dispersion: 1.3.6.1.4.1.9.9.168.1.2.1.1.25.64158 = 0
```


Some of the other graph changes are necessary because otherwise they simply do not display correctly.

mattie47: Ah. I'm an idiot. Perhaps the Cisco output above is actually working but because everything is less than a second, it's simply reporting 0 seconds for Offset/Delay/Dispersion. 

I've had a look at the Cisco MIB and can see for whatever reason they output these things in seconds, rather than milliseconds. I imagine this is probably historic given things like slower networks 2 decades ago...

So - I can convert my Allied Telesis stats to seconds, but as above it just seems so illogical given all my stats for Offset/Delay/Dispersion are less than 100msec

mattie47: @laf  OK have made requested changes. Thanks.

mattie47:
Now that I'm doing both in seconds, I'm just double checking the formatting, and making sure it all makes sense.

mattie47:
OK so - I realise I've modified the formatting of the graphs however I believe this is for the better.

Now all graphs nicely line up with the legend being in line for each graph:

e.g. 

image

mattie47:
Also not sure if I should *also* be unsetting $atNtpAssociationEntry in  https://github.com/mattie47/librenms/blob/a31b48e82ee4facf9f316622cfc304a83be8275a/includes/polling/ntp.inc.php#L29

laf:
Thanks for updating all of that.

Do we need to show on the graphs the value is in seconds still, it's been removed and people might not know what the data is representing.

mattie47:
Hey @laf,

>Do we need to show on the graphs the value is in seconds still, it's been removed and people might not know what the data is representing.

I'm confused. I've added seconds to the side of the graphs, and also in the legend on the following line:

https://github.com/mattie47/librenms/blob/a31b48e82ee4facf9f316622cfc304a83be8275a/html/includes/graphs/device/ntp_dispersion.inc.php#L24-L25

BTW - I'd been thinking about this PR over the w/e and I think the original change I made where I added two decimal places might be a good idea. It'll still obviously be in seconds, however if other vendors support the OID in milliseconds, then we get values like 0.20 seconds for 200 milliseconds.  That way the graphs won't simply be a whole bunch of 0 seconds results. 

Then of course vendors who only support seconds will appear as 1.00 seconds or 0.00 seconds etc. 

I might have a look at it on Monday if you're happy with that idea and update this.

Thanks,

Matt

laf:
> I'm confused. I've added seconds to the side of the graphs, and also in the legend on the following line:

Sorry didn't see that, could only see it being removed.



> BTW - I'd been thinking about this PR over the w/e and I think the original change I made where I added two decimal places might be a good idea. It'll still obviously be in seconds, however if other vendors support the OID in milliseconds, then we get values like 0.20 seconds for 200 milliseconds. That way the graphs won't simply be a whole bunch of 0 seconds results.
> 
> Then of course vendors who only support seconds will appear as 1.00 seconds or 0.00 seconds etc.
> 
> I might have a look at it on Monday if you're happy with that idea and update this.

Are you not adding the data in as 0.20? that should show as such in the graph.

mattie47:
>Are you not adding the data in as 0.20? that should show as such in the graph.

Nah, the "0" in the following lines leave the value as a whole number:

```
        $rrd_additions .= " GPRINT:DS" . $count . ":LAST:%7.0lf ";
        $rrd_additions .= " GPRINT:DS" . $count .    ":MIN:%7.0lf ";
        $rrd_additions .= " GPRINT:DS" . $count . ":MAX:%7.0lf\\l ";
```

This is how it was originally written. My previous commit made it:

```
        $rrd_additions .= " GPRINT:DS" . $count . ":LAST:%7.2lf ";
        $rrd_additions .= " GPRINT:DS" . $count .    ":MIN:%7.2lf ";
        $rrd_additions .= " GPRINT:DS" . $count . ":MAX:%7.2lf\\l ";
```

which added 2 decimal places.  I also need to remove "floor" from my next commit when changing milliseconds into seconds...

Aside from those minor things, anything else stick out I need to fix up?

Thanks,

Matt

laf:
> Aside from those minor things, anything else stick out I need to fix up?

I'd say the main thing for me would be could you use the MIB names / named OIDs in the snmpwalks? Other than that, no looks good to go.

mattie47:
Hi @laf,

I've tried to convert the OID numbers to the names instead, but I'm not having much luck.

While I can do:

$atNtpAssociationEntry = snmpwalk_array_num($device, 'atNtpAssociationEntry', 2);

What's returned is numerical:

```
Objects Found:
Array
(
    [1.3.6.1.4.1.207.8.4.4.4.502.10.1] => Array
```
	
I don't see a function in snmp.inc.php which returns named OIDs while allowing multiple dual indexes. 

Therefore I'm not sure how to reference below as named OIDs rather than numerical:

```
    foreach ($atNtpAssociationEntry['1.3.6.1.4.1.207.8.4.4.4.502.10.1'][2] as $index => $value) {
        $result = array();
        $result['UID'] = (string)$index;    // This is cast as a string so it can be compared with the database value.
        $result['peer'] = $atNtpAssociationEntry['1.3.6.1.4.1.207.8.4.4.4.502.10.1'][2][$index];
```

So I tried:

```
    foreach ($atNtpAssociationEntry['atNtpAssociationPeerAddr'] as $index => $value) {
        $result = array();
        $result['UID'] = (string)$index;    // This is cast as a string so it can be compared with the database value.
        $result['peer'] = $atNtpAssociationEntry['atNtpAssociationPeerAddr'][$index];
```

but as above, either I'm doing something stupid, or this simply isn't possible. 


Thoughts?


Thanks,

Matt

murrant:
@mattie47 snmpwalk_group() may be able to do what you want (try adjusting the depth to 0 or 2)

mattie47:
Thanks @murrant  

OK I can't see anything else I need to do to this. I think it's ready :-)

laf:
Test json data added.

mattie47:
@laf  Did you intentionally remove a whole lot of the test data? 

If not, I've run SNMPsim and generated the new json and can commit it. 

Thanks,

Matt

mattie47:
Hm okay so I ran into the same problem with generating the test data as @laf did. 

What I didn't notice before I committed was the following when running "./scripts/save-test-data.php -o awplus":

```
#### Load disco module ntp ####
+++
MySQL Error: Field 'app_status' doesn't have a default value (INSERT INTO `applications` (`device_id`,`app_type`)  VALUES ('203','ntp'))

>> Runtime for discovery module 'ntp': 0.0830 seconds with 52672 bytes
>> SNMP: [0/0.00s] MySQL: [46/0.05s] RRD: [0/0.00s]
#### Unload disco module ntp ####
```


Any idea what needs doing to fix this? 

Matt

murrant:
When creating the app in the database, you need to supply an empty string (or valid data) for the app_status field.

includes/discovery/ntp/awplus.inc.php
```diff
@@ -0,0 +1,130 @@
...
+
+    // Write the Components back to the DB.
+    $component->setComponentPrefs($device['device_id'], $components);
+    echo "\n";
+} // End if not error
+
+$module = strtolower($module);
+if (count($components) > 0) {
+    if (dbFetchCell('SELECT COUNT(*) FROM `applications` WHERE `device_id` = ? AND `app_type` = ?', array($device['device_id'], $module)) == '0') {
+        dbInsert(array('device_id' => $device['device_id'], 'app_type' => $module), 'applications');
```
murrant: When creating the app in the database, you need to supply an empty string (or valid data) for the app_status field.

mattie47: Thanks @murrant. I see now. 

So this works, and I assume is the correct way to do it:

```
+        dbInsert(array('device_id' => $device['device_id'], 'app_type' => $module, 'app_status' => '', 'app_instance' => ''), 'applications');
```

I guess I should fix up the same line in the cisco.inc.php config as well ? 

https://github.com/mattie47/librenms/blob/3728439fd88ffafb395f60c4202da11d151612d1/includes/discovery/ntp/cisco.inc.php#L126

murrant:
@mattie47 odd that that is A failing and B not failing for other PRs.

murrant:
@mattie47 you have caused a state_name collision :) and for the first time we are testing both that have the collision.  Can you rename the awplus sensor states?

mattie47:
I believe the state_name collision was caused by https://github.com/librenms/librenms/pull/8561 

However I've renamed the awplus ones instead here: https://github.com/librenms/librenms/pull/8671

PR #8591, opened 2018-04-20
Fix OIDs for temperature sensor for cimc os

vivia11:
Hello all,

Let me know if the snmp data isn't necessary.

Added a period that was missing on some OIDs for temperature sensing.

Removed $temp_board/cpu/mem[$oid][$index] so it is not passed to discover_sensor. It passes the values NULL anyways. The OID is appended with the index value, and then this new key is read from the $temp array. However, no such key value exists, resulting in NULL values. Instead of editing the string for each discover_sensor, I thought it would be simpler to remove it and have the values polled.
 
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
So the removed variable is current, we really need that otherwise high/low doesn't get calculated. Is it not available in the walk?

laf:
@vivia11 You might need to regenerate the actual snmprec file `collect-snmp-data.php`

includes/discovery/sensors/current/cimc.inc.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+$pwr_board = snmpwalk_array_num($device, '.1.3.6.1.4.1.9.9.719.1.9.14');
+
+/*
+* False == OID not found - this is not an error.
+* null  == timeout or something else that caused an error.
+*/
+if (is_null($pwr_board)) {
```
laf: We don't need this or the else in. Just check the `$pwr_board` is not empty and then run your code.

includes/discovery/sensors/current/cimc.inc.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+$pwr_board = snmpwalk_array_num($device, '.1.3.6.1.4.1.9.9.719.1.9.14');
```
laf: As you don't loop through this, can you not just use a single snmpget?

includes/discovery/sensors/current/cimc.inc.php
laf: Same for the other files.

includes/discovery/sensors/current/cimc.inc.php
```diff
@@ -0,0 +1,17 @@
+<?php
```
laf: Looking at this, it should be possible to do this in yaml, any reason you haven't?

laf:
Awesome work @vivia11 :)

PR #8599, opened 2018-04-23
Provide simple/complex Oxidized configs

hrabbit:
This patch provides the ability to override most aspects of the configuration output provided to Oxidized. This should be backwards compatible with any current maps defined previously but allows for extra features such as group overrides in Oxidized as well as IP changes (for hosts that are treated differently inside Oxidized), as well as ProxyHost additions. Of course, this goes further and allows for any flag that can be defined within Oxidized to be mapped in the config.

Examples

```php
```
```php
```
```php
```

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -1213,42 +1213,28 @@ function list_oxidized()
         $params = array($hostname);
     }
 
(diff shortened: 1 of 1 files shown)


hrabbit:
Style guide rejections.. I'll clean up and resubmit.

PR #8600, opened 2018-04-23
Allow more complete overrides on Oxidized output

hrabbit:
This patch provides the ability to override most aspects of the configuration output provided to Oxidized. This should be backwards compatible with any current maps defined previously but allows for extra features such as group overrides in Oxidized as well as IP changes (for hosts that are treated differently inside Oxidized), as well as ProxyHost additions. Of course, this goes further and allows for any flag that can be defined within Oxidized to be mapped in the config.

Examples

```php
```
```php
```
```php
```

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -1213,42 +1213,28 @@ function list_oxidized()
         $params = array($hostname);
     }
 
-    foreach (dbFetchRows("SELECT hostname,sysname,os,location FROM `devices` LEFT JOIN devices_attribs AS `DA` ON devices.device_id = DA.device_id AND `DA`.attrib_type='override_Oxidized_disable' WHERE `disabled`='0' AND `ignore` = 0 AND (DA.attrib_value = 'false' OR DA.attrib_value IS NULL) AND (`type` NOT IN ($device_types) AND `os` NOT IN ($device_os)) $sql", $params) as $device) {
-        if ($config['oxidized']['group_support'] == "true") {
-            foreach ($config['oxidized']['group']['hostname'] as $host_group) {
-                if (preg_match($host_group['regex'].'i', $device['hostname'])) {
-                    $device['group'] = $host_group['group'];
-                    break;
-                }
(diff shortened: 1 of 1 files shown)


laf:
So looking at this code it's not backwards compatible? If not, then you'll need to resolve that before we carry on as we can't break existing installs.

laf:
@hrabbit No worries. You can get this to work. Possibly in `processConfig()` you can set the new values based on the old ones.

You can't use the `deprecatedVariable()` like others do in there though as it sets it in the DB which isn't supported for this complex type of config

hrabbit:
> Possibly in processConfig() you can set the new values based on the old ones.

Thanks for that, I'll take a look at it. It's occured to me that the reasoning behind the 'maps' key was to allow for absolutely any flags to be set within it (including future Oxidized flags) and that would be included without clobbering into the LNMS local settings. I think processConfig() might allow me to move the existing 'groups' key to maps.groups which should prevent breakages.

hrabbit:
Hi @laf,

Sorry, but just noticed that deprecatedVariable() is using set() set override the config variables, however, if I'm reading this correctly, set won't save anything into the database unless the persist flag is defined as true (defaults to false)

https://github.com/librenms/librenms/blob/master/LibreNMS/Config.php#L225

I realise you don't need the below of course, but for clarity, figured it wouldn't harm anybody.

```php
 public static function set($key, $value, $persist = false, $default = '', $descr = '', $group = '', $sub_group = '')
    {
        global $config;

        if ($persist) {
            $res = dbUpdate(array('config_value' => $value), 'config', '`config_name`=?', array($key));
            if (!$res && !dbFetchCell('SELECT 1 FROM `config` WHERE `config_name`=?', array($key))) {
                $insert = array(
                    'config_name' => $key,
                    'config_value' => $value,
                    'config_default' => $default,
                    'config_descr' => $descr,
                    'config_group' => $group,
                    'config_sub_group' => $sub_group,
                );
                dbInsert($insert, 'config');
            }
        }
```

And... deprecatedVariable()

```php
private static function deprecatedVariable($old, $new)
    {
        if (self::has($old)) {
            d_echo("Copied deprecated config $old to $new\n");
            self::set($new, self::get($old));
        }
    }
```

I'm hoping I'm just missing something because my initial plan was to just add a local mapping method to the Config class that would just not call the persist option but upon seeing the above, I'm not sure it would be needed, and in fact might be just introducing duplicated code.

murrant:
Would be good to add some documentation for this too.
doc/Extensions/Oxidized.md

Issue #8602, opened 2018-04-23
RouterOS yaml sets Temp limits incorrectly


tslytsly:
New temperature discovery yaml for routeros has limits set incorrectly.
I think the intention was to set the limits to +70 and -40.
But because the sensor has a divisor of 10, these end up in LNMS as +7 and -4.

Easy to fix, I will submit a PR shortly.

**DO NOT DELETE THIS INFORMATION.**

> **Please read this information carefully.**

**GitHub issues is for new device requests only. Please do not post bugs or ask questions** 

**For help and support you can use our [Discord server](https://t.libren.ms/discord) or our [community site](https://community.librenms.org).**

**If you have a feature request/changes please post this on our [community site](https://community.librenms.org/c/feature-requests).**

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide **ALL** info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations
```
librenms@librenms:~$ ./validate.php
====================================
Component | Version
--------- | -------
LibreNMS  | 1.30-1008-g21e26a4
DB Schema | 247
PHP       | 7.0.28-0ubuntu0.16.04.1
MySQL     | 10.0.34-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3
====================================

[OK]    Composer Version: 1.6.4
[OK]    Dependencies up-to-date.
[OK]    Database connection successful
[OK]    Database schema correct
```

PR #8603, opened 2018-04-23
Changed limit values in routeros.yaml

tslytsly:
Limits do not account for the divisor of 10.
So need to be 10 x the actual value.
See https://github.com/librenms/librenms/issues/8602

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


tslytsly:
> Just need to fix other sensors.
what do you mean?

laf:
@tslytsly I've updated it now. However you need to re-run the `save-test-data.php` script rather than edit the json file directly.

murrant:
Yeah, I meant update test data, sorry.

Issue #8566, opened 2018-04-16
Eventlog full of "Sensor updated" messages


thalesog:
**DO NOT DELETE THIS INFORMATION.**

> **Please read this information carefully.**

**GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z.** 

**You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).**

**If you have a feature request/changes please post this on our [community site](https://community.librenms.org/c/feature-requests).**

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
====================================
Component | Version
--------- | -------
LibreNMS  | 1.38-57-g146fc985a
DB Schema | 247
PHP       | 7.0.27-0+deb9u1
MySQL     | 10.1.26-MariaDB-0+deb9u1
RRDTool   | 1.6.0
SNMP      | NET-SNMP 5.7.3
====================================
- [x] Please provide **ALL** info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

My LibreNMS new install is flooding "Sensor Updated" messages on EventLog
[Eventlog Print](https://i.imgur.com/pMfdTdl.png)

Debug from one of devices:
Discovery - [Link](https://p.libren.ms/view/c331f5de)
Poller - [Link](https://p.libren.ms/view/4ce3e187)
Snmpbulkwalk - [Link](http://repo.thalesog.com/output.txt)

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

murrant:
You can use another pastebin service.

laf:
Please create a separate issue.

PR #8604, opened 2018-04-23
 fix: Added indexes to two routeros temp sensors

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #8566 

Commit 1 is how I'd like to fix this but commit 2 retains historical data for those who had data from before the last refactor of these two lookups. Anything in the last 3 months will only have one sensor which round robin'd through both OIDs.


murrant:
@laf would it be better to set the type instead of doing weird stuff with index like the other entries?

laf:
> @laf would it be better to set the type instead of doing weird stuff with index like the other entries?
> 
> I kind of wish type defaulted to the oid value.

I actually prefer this way as it's specific to the sensor itself rather than a general name or the os name (which is a pain when renaming stuff). The issue for this stems from historic code where it used just the last index from the oid which as we know, isn't unique across the sensor types.

Might be worth dropping the index as a requirement and defaulting it to the value+index like we try and push people to anyway. People can override it if they like.

Overall, if we could get the renaming of rrds right then this would be a mute point as we could fix all things :)

PR #8613, opened 2018-04-25
added new TPLINK logo with brand name

czilian:
added "TP Link" underneath previous SVG graphic. (Logo itself is very little self-explanatory without)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Icons aren't meant to be self explanatory.  They are meant for quick identification.

So unless this is similar to another icon I don't think this is a good change.  I'm not a fan of text in the icons.

However, I do think you should add a logo image with the tplink text to the right. https://docs.librenms.org/Developing/os/Initial-Detection/#icon-and-logo

czilian:
Yes it is. Use this one then. Thx

murrant:
Sorry, thought I had those options checked for svgo

PR #8619, opened 2018-04-26
Expand the description of Application support

hanserasmus:
I was not aware that the applications can also be auto-discovered. I discovered it by chance, and when I mentioned it in discord channel, it seemed blatantly obvious to everyone else. Thought editing the docs might help future users.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


doc/Extensions/Applications.md
```diff
@@ -3,6 +3,8 @@ source: Extensions/Applications.md
 
 You can use Application support to graph performance statistics from many applications.
 
+Note that the monitoring of applications could be added before or after the hosts have been added to LibreNMS. If your device is already in LibreNMS, you will need to go and edit the pertaining device, and enable the application(s) you want to monitor. If the device is not in LibreNMS yet, don't sweat, because adding the agent or snmp extend scripts before adding the device will make the application auto-discoverable by LibreNMS.
```
laf: They should also be auto discovered even if setting up an app after the device is added in.

hanserasmus: You mean without manually enabling it per device, if the device is already in LNMS? Then maybe I should look at the page as a whole? Rethink the way all of it fits together? Some apps contain the part about "remember to edit your device in LNMS after configuring XYZ" and some apps don't. So maybe I should spend time over the weekend on this, if you agree?

murrant: Overview

If using an snmp extend then the application discovery module will pick it up. The application discovery module is enabled by default for most *nix operating systems. (otherwise you will have to manually enable that discovery module)

The unix-agent does not have a discovery module, only a poller module.  That poller module is always disabled by default.  It needs to be manually enabled if using the agent.  Some applications will be automatically enabled by the unix-agent poller module.

hanserasmus: Maybe we should add this overview to the page. I am not aware that it is written up somewhere else? I am willing to rewrite a lot of the stuff on the page, but don't want to create a lot of unnecessary work for the reviewers/mergers. If you want me to go ahead, I will do so over the weekend.

murrant: Please do go ahead.

hanserasmus: Have changed a lot in the file. Did not want to split it up into multiple PR's, as a lot of it was just copy and paste.

murrant:
Images can be added in `doc/img`

kkrumm1:
Looks good to me. @hanserasmus Good work brotha! :)

PR #8622, opened 2018-04-26
Poller taking too long -> device took too long

murrant:
We can't really alert on the poller too long since or alerts are device centric.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


misc/alert_rules.json
```diff
@@ -81,7 +81,7 @@
     "name": "IPSec tunnels down"
   },
   {
-    "rule": "%pollers.time_taken >= \"250\"",
+    "rule": "devices.last_polled_timetaken >= 290",
```
laf: Actually shouldn't this be %devices.last_polled_timetaken?

murrant: We strip out the % before processing anyway. Did you try it?

laf: No but all of the others have the % in. I'm sure I stripped them out when I did the large alerting PR but seem to recall you added them back in maybe?

murrant: Might have gotten reverted at some point.

laf: Also, this changes the alert from the pollers total time to a single device which is fine as the original one obviously won't work but it also needs the name of the rule updating.

murrant:
Removed all % from rules and renamed the poller taking too long rule.

laf:
Thanks for updating these @murrant

PR #8626, opened 2018-04-27
Enhance Cumulus Support

theherodied:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


theherodied:
Not ready to merge yet. I will add test data when I'm done adding more features.

murrant:
You are changing the index here, that might not work on other devices.  Perhaps you want to use `snmp_getnext_multi()`?

theherodied:
@murrant, does that look better?

murrant:
I think so

Rosiak:
@librenms/reviewers 
Should we really include all these mibs, when none of them are used?

theherodied:
I was wondering that myself. I was going to bring it up before any merge.

I think the cumulus-* mibs should be included. Not sure about the rest.

murrant:
Yeah, you can leave out the mibs for now, add them when you are using them ;)  We already ship 160MB of mibs :)

theherodied:
@librenms/reviewers, I think this is good and the blocker can be removed if the Travis check passes. Let me know if it needs anything else.

theherodied:
@laf think we can remove the blocker. I'm done with this for now.

PR #8627, opened 2018-04-27
Add support for Metamako MOS and Exablaze Fusion devices. IPv4-only fixes

gs-kamnas:
This PR adds support for discovering and polling Metamako devices running Metamako's MOS operating system, as well as Exablaze Fusion devices. Additionally, this changeset resolves an issue where validation fails on systems where IPv6 is disabled by unloading/omitting the relevant kernel modules (as opposed to disabling via sysctl).

#### Please note

#### Testers


includes/definitions/linux.yaml
```diff
@@ -34,3 +34,4 @@ register_mibs:
 discovery:
     - sysObjectID: .1.3.6.1.4.1.8072.3.2.10
     - sysDescr_regex: '/^Linux/'
+    - sysObjectID_except: .1.3.6.1.4.1.43296.
```
laf: You shouldn't need this here.

As your device is using `.1.3.6.1.4.1.43296.` for sysObjectID then it wouldn't match anyway.

includes/definitions/discovery/exablaze-fusion.yaml
```diff
@@ -0,0 +1,58 @@
+mib: EXALINK-FUSION-MIB
+modules:
+    sensors:
+        pre-cache:
```
laf: You can drop the `pre-cache` stuff here. It's only needed if you need access to other snmp data.

includes/polling/os/exablaze-fusion.inc.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+#EXALINK-FUSION-MIB::fusionInfoSerial = STRING: EXAFSN-B-00402
```
laf: You can drop the comments :)

includes/polling/os/exablaze-fusion.inc.php
```diff
@@ -0,0 +1,11 @@
...
+#EXALINK-FUSION-MIB::fusionInfoVersion = STRING: 1.5.0
+#EXALINK-FUSION-MIB::fusionInfoBoard = STRING: EBFSN-03
+#EXALINK-FUSION-MIB::fusionInfoSoftware = STRING: 2016-09-05 18:16:58 +1000 (c4573c4)
+
+$hardware = snmp_get($device, "fusionInfoBoard", "-Ovq", "EXALINK-FUSION-MIB");
```
laf: Please convert these four `snmp_get()` calls to a single `snmp_get_multi_oid()` call.

includes/definitions/linux.yaml
gs-kamnas: I was having issues here because Metamako devices have a sysDescr beginning with "Linux" though.

gs-kamnas:
@laf Will refactor and split into 2 PRs as requested.

PR #8630, opened 2018-04-27
Create Installation-Ubuntu-1804-Nginx.md

bfarmerjr:
A few tweaks are required to the Ubuntu 16.04 instructions to install on the newly released 18.04 version mostly relating to php being bumped to 7.2 from 7.0 and no mcrypt module for this php version in the distro repository.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Thanks @bfarmerjr :)

Could you copy the Apache one as well and update that? You also need to update mkdocs.yaml to add these two files otherwise they won't get built.

@librenms/reviewers Is it worth setting 16.04 as deprecated, remove from the menu and set 18.04 as current?

bfarmerjr:
I'll spin up another 18.04 VM and test/adjust apache instructions and make that one as well.

As far as deprecating 16.04, I personally would vote no as it is in support until April 2021.

Should I do another pull request for those two files or can they be added to this one? I don't have much experience with github, yet. I just decided since I took the time to figure this out, I would contribute instructions for others.

@murrant yes mcrypt (installing the module and enabling it) is in the existing [ubuntu 16.04 instructions](https://docs.librenms.org/#Installation/Installation-Ubuntu-1604-Nginx/). if it isn't used then that makes it even simpler to install and I can remove from this page

laf:
> As far as deprecating 16.04, I personally would vote no as it is in support until April 2021.

That's fine for Ubuntu but we don't need to support it :) The docs can still be there but we just choose to push 18.04.

Rosiak:
I agree with @laf, same approach we have used for CentOS(6 vs. 7).

paulgear:
At Ubuntu, we don't offer upgrades to LTS users until the first point release (18.04.1 in this case) is out, and I think it would be worth doing something similar here.  See https://wiki.ubuntu.com/Kernel/Support#A16.04.x_Ubuntu_Kernel_Support for example timelines.

murrant:
Seems reasonable to me to wait until 18.04.1 and then put 18.04 as our main install target and stop updating the 16.04 install docs.

laf:
Then for now it shouldn't be prominent within the nav for installation and should be added to https://github.com/librenms/librenms/blob/master/doc/Installation/index.md as a form of beta link.

murrant:
Also, there have been several updates to the install docs since this PR that should be carried over.

laf:
18.04.01 is due on July 26th

laf:
@bfarmerjr 18.04.01 is now out. Could you address the comments in this thread and update the install docs please.

doc/Installation/Installation-Ubuntu-1804-Apache.md
```diff
@@ -0,0 +1,154 @@
...
+
+Ensure date.timezone is set in php.ini to your preferred time zone.  See http://php.net/manual/en/timezones.php for a list of supported timezones.  Valid examples are: "America/New_York", "Australia/Brisbane", "Etc/UTC".
+
+    nano /etc/php/7.2/apache2/php.ini
+    nano /etc/php/7.2/cli/php.ini
+
+    a2enmod php7.2
+    a2dismod mpm_event
+    a2enmod mpm_prefork
+    phpenmod mcrypt
```
murrant: mcrypt isn't needed and doesn't exist.

doc/Installation/Installation-Ubuntu-1804-Apache.md
```diff
@@ -0,0 +1,154 @@
+source: Installation/Installation-Ubuntu-1804-Apache.md
+> NOTE: These instructions assume you are the **root** user.  If you are not, prepend `sudo` to the shell commands (the ones that aren't at `mysql>` prompts) or temporarily become a user with root privileges with `sudo -s` or `sudo -i`.
+
+**Please note the minimum supported PHP version is 5.6.4**
+
+## Install Required Packages ##
+
+    apt update
+    apt install apache2 composer fping git graphviz imagemagick libapache2-mod-php7.2 mariadb-client mariadb-server mtr-tiny nmap php7.2-cli php7.2-curl php7.2-gd php7.2-json php7.2-mysql php7.2-snmp php7.2-xml php7.2-zip python-memcache python-mysqldb rrdtool snmp snmpd whois
```
murrant: missing mbstring, please update this so it is based of the latest install docs for 16.04.

PR #8633, opened 2018-04-28
Allow more complete overrides on Oxidized output

hrabbit:
This patch provides the ability to override most aspects of the configuration output provided to Oxidized. This is now backwards compatible with any current groups defined previously but allows for extra features such as group overrides in Oxidized as well as IP changes (for hosts that are treated differently inside Oxidized), as well as ProxyHost additions. Of course, this goes further and allows for any flag that can be defined within Oxidized to be mapped in the config.

Examples

```php
```
```php
```
```php
```

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Config.php
+++ b/LibreNMS/Config.php
@@ -439,7 +439,7 @@ class Config
         self::deprecatedVariable('rrdgraph_real_95th', 'rrdgraph_real_percentile');
         self::deprecatedVariable('fping_options.millisec', 'fping_options.interval');
         self::deprecatedVariable('discovery_modules.cisco-vrf', 'discovery_modules.vrf');
-        self::deprecatedVariable('oxidixed.groups', 'oxidized.maps.groups');
+        self::deprecatedVariable('oxidixed.group', 'oxidized.maps.group');
 
         // make sure we have full path to binaries in case PATH isn't set
         foreach (array('fping', 'fping6', 'snmpgetnext', 'rrdtool') as $bin) {
--- a/LibreNMS/Config.php
+++ b/LibreNMS/Config.php
@@ -439,7 +439,7 @@ class Config
         self::deprecatedVariable('rrdgraph_real_95th', 'rrdgraph_real_percentile');
         self::deprecatedVariable('fping_options.millisec', 'fping_options.interval');
         self::deprecatedVariable('discovery_modules.cisco-vrf', 'discovery_modules.vrf');
-
+        self::deprecatedVariable('oxidixed.groups', 'oxidized.maps.groups');
 
         // make sure we have full path to binaries in case PATH isn't set
         foreach (array('fping', 'fping6', 'snmpgetnext', 'rrdtool') as $bin) {
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -1237,7 +1237,11 @@ function list_oxidized()
         $params = array($hostname);
     }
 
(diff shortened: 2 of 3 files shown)


laf:
You know you don't need to close and open pull requests to make changes, just edit the files and push the commits.

This isn't working 100% still, we have more than just oxidized.groups (it's oxidized.group btw) as available options. Please see (https://docs.librenms.org/#Extensions/Oxidized/#miscellaneous) for the others.

Could you also update the docs as well to reflect these changes.

hrabbit:
Thanks @laf,

I actually did the changes through the web interface and searched for a way to add more files but it didn't seem very clear. I'll pull my fork and do those changes locally.

As for the group(s) (I'll fix the spelling) side of it, I've purposely not touched anything relating to the LNMS side of the config and only added the parts that specifically get used in the output to Oxidized. I did use the docs to double check there weren't any others and group was the only one that was going to be needed to be mapped.

I'll put these changes in place and update this request.

thanks again

hrabbit:
@murrant, roger that! :+1:

LibreNMS/Config.php
```diff
@@ -383,7 +383,7 @@ private static function processConfig()
         self::deprecatedVariable('rrdgraph_real_95th', 'rrdgraph_real_percentile');
         self::deprecatedVariable('fping_options.millisec', 'fping_options.interval');
         self::deprecatedVariable('discovery_modules.cisco-vrf', 'discovery_modules.vrf');
-
+        self::deprecatedVariable('oxidixed.groups', 'oxidized.maps.groups');
```
laf: Should `oxidized.groups` not be `oxidized.group`?

hrabbit:
@laf, I've had to hold off due to other commitments but hoping in the next 3 or 4 days if that works. I'd love to have this included as it removes my painful hacks from existing. :)

laf:
@hrabbit That's fine, just let us know when it's ready.

hrabbit:
@laf + @murrant 
Hoping that passes now, I had to make some tweaks as I realised I completely bypassed the 'match' lookup that was previously defined.
Documentation should also be more clear now as to requirements.

html/includes/api_functions.inc.php
```diff
@@ -1237,42 +1237,30 @@ function list_oxidized()
...
-                }
-            }
-            if (empty($device['group'])) {
-                foreach ($config['oxidized']['group']['sysname'] as $host_group) {
-                    if (preg_match($host_group['regex'].'i', $device['sysname'])) {
-                        $device['group'] = $host_group['group'];
-                        break;
-                    }
-                }
+    foreach (dbFetchRows("SELECT hostname,sysname,os,location,INET6_NTOA(ip) AS ip FROM `devices` LEFT JOIN devices_attribs AS `DA` ON devices.device_id = DA.device_id AND `DA`.attrib_type='override_Oxidized_disable' WHERE `disabled`='0' AND `ignore` = 0 AND (DA.attrib_value = 'false' OR DA.attrib_value IS NULL) AND (`type` NOT IN ($device_types) AND `os` NOT IN ($device_os)) $sql", $params) as $device) {
```
laf: You can't do `INET6_NTOA(ip)` here as not all mysql versions support it.

Instead on the next line do:

`$device['ip'] = inet6_ntop($device['ip']);`

laf: Don't forget to keep the `ip` in the query though

PR #8640, opened 2018-04-30
Created a page to show all known VMs

aldemira:
I've created single page which shows all VMs known to LibreNMS. Also I thought not everyone would like to enable this option so it requires $config['show_allvm_list'] = true; in config.php to be enabled as a menu item (under devices)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/forms/get-vmlist.inc.php
+++ b/html/includes/forms/get-vmlist.inc.php
@@ -3,9 +3,8 @@ $vm_query = "SELECT a.vmwVmDisplayName AS vmname, a.vmwVmState AS powerstat, a.d
 
 if (isset($_POST['searchPhrase']) && !empty($_POST['searchPhrase'])) {
     #This is a bit ugly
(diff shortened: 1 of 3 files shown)


murrant:
I'd suggest counting the entries in the vminfo table instead of adding a config option.

aldemira:
@laf TBH I copied the SQL query from device dependencies ajax call. I'm going to fix that in a separate PR as well. 
screen shot 2018-05-10 at 13 03 29
screen shot 2018-05-10 at 13 28 24

laf:
@aldemira Actually would be good if you could add a license to the new file + update the one you did add using this template; https://docs.librenms.org/#Developing/Licensing/

laf:
@aldemira I've refactored the code a little bit. Can you test this patch and push it to this PR please:

https://p.libren.ms/view/3a29c5c3

PR #8642, opened 2018-04-30
Change Dell iDrac from Server to Appliance

theherodied:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/drac.yaml
+++ b/includes/definitions/drac.yaml
@@ -1,7 +1,7 @@
 os: drac
 text: 'Dell DRAC'
 icon: dell
-type: server
+type: appliance
 mib_dir:
     - dell
 discovery:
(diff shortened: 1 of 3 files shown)


theherodied:
@laf wasn't sure if we needed test data for this?

Thinking this should have an advance warning before being merged in case people had the appliance category in a rule.

Also, I wasn't sure if iLO should be server or if iDrac should be appliance. I went with server for both.

Rosiak:
I'm fine with either.
@librenms/reviewers ?

kkrumm1:
Sounds good to me.

murrant:
IMHO, ILO and iDRAC are appliances but supply server data.  I'm fine with either, but I agree it should be consistent.

theherodied:
Agreed. Will change iDrac instead and update PR.

PR #8643, opened 2018-05-01
Added support for graphing pf related stats for pfsense devices

utelisysadmin:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

In response to feature request:
https://community.librenms.org/t/add-support-for-begemot-pf-mib-for-pfsense-devices/3628


includes/definitions/pfsense.yaml
```diff
@@ -9,3 +9,5 @@ over:
 discovery:
     - sysDescr:
         - pfSense
+mib_dir:
```
murrant: This is redundant, we always append the directory with the same name as the OS.

murrant:
Other than the one small change, LGTM.

utelisysadmin:
Done.

mibs/pfsense/RSTP-MIB.txt
```diff
@@ -0,0 +1,325 @@
+--
```
laf: This mib already exists: `mibs/RSTP-MIB` so you can delete this one.

mibs/pfsense/BRIDGE-MIB.txt
```diff
@@ -0,0 +1,1483 @@
+--
```
laf: This mib already exists in `mibs/BRIDGE-MIB` so you can delete this one.

includes/polling/os/pfsense.inc.php
```diff
@@ -0,0 +1,147 @@
...
+$removals = $oids[0]['pfStateTableCount'];
+$matches = $oids[0]['pfCounterMatch'];
+$badoffset = $oids[0]['pfCounterBadOffset'];
+$fragmented = $oids[0]['pfCounterFragment'];
+$short = $oids[0]['pfCounterShort'];
+$normalized = $oids[0]['pfCounterNormalize'];
+$memdropped = $oids[0]['pfCounterMemDrop'];
+
+
+if (is_numeric($states)) {
```
laf: Not really sure if this will be a problem or not but as you check if each of these values are numeric then it could mean that in the event they aren't then from one poll to another you could end up with data out of order or the wrong number of DS' in the RRD if you have transient issues with the `snmp_multi_get()` call.

laf:
@utelisysadmin Any update on the comments we've left?

utelisysadmin:
Hi.
Apologies for the delay, been busy with other things. I'll remove the MIBs which are duplicate.

I do understand the comment about the number check, but I don't understand what do You want me to do about it. After all, all I did was copy the existing codebase for OpenBSD. All the OID used here are of type NUMERIC anyway.

includes/polling/os/pfsense.inc.php
murrant: @laf I don't think that is an issue here since the data is fetched by oid name.

laf: Actually I've just realised that these are all creating separate files anyway so this isn't an issue.

laf: If you can fix the mib issues from my comments I think we're good to go.

utelisysadmin:
Renamed all MIBs as requested.

utelisysadmin:
Yep, I forgot two. Renamed now.

PR #8646, opened 2018-05-01
Adding support for embedded graphs in email alerts

chasgames:
Created functionality to support embedded graphs in HTML emails.

To help test
```./scripts/github-apply 8646```

Add this line somewhere in your config.php
```$config['allow_unauth_graphs'] = true;```

Add this line in your HTML template and change the alt value to your graph.
```https://SERVERIP/GRAPHLINK```

Request references:
https://community.librenms.org/t/embed-graph-in-alert-email/3418/5
https://community.librenms.org/t/feature-request-embed-images-into-email
https://community.librenms.org/t/graphs-on-email/3749/4
https://community.librenms.org/t/alert-template-with-graph/3523

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -965,8 +965,7 @@ function send_mail($emails, $subject, $message, $html = false)
                 $image = file_get_contents($item, false, stream_context_create($arrContextOptions));
                 if ($config['webui']['graph_type'] == 'svg') {
                     $mail->addStringEmbeddedImage($image, $countcheck, 'graph.svg', 'base64', 'image/svg+xml');
-                }
-                else {
+                } else {
                     $mail->addStringEmbeddedImage($image, $countcheck, 'graph.png', 'base64', 'image/png');
                 }
                 $countcheck++;
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -950,28 +950,23 @@ function send_mail($emails, $subject, $message, $html = false)
         if ($html) {
             $mail->isHTML(true);
             //Addding functionality to support cid embedded graphs
-				$arrContextOptions=array(
-				    "ssl"=>array(
-				        "verify_peer"=>false,
-				        "verify_peer_name"=>false,
-				        "timeout" => 1200,
-				    ),
+            $arrContextOptions=array(
+                "ssl"=>array(
+                "verify_peer"=>false,
+                "verify_peer_name"=>false,
+                "timeout" => 1200,
+                ),
 				);
-
-				preg_match_all('/ alt=\"(.*?)\"/', $mail->Body, $match);
-
-				$countcheck = 0;
-				$newcidhtml = $mail->Body;
-				foreach ($match[1] as $item) {
-
-				$newcidhtml=preg_replace("/\bembedimage\b/i", $countcheck, $newcidhtml, 1);
-
-				            $image = file_get_contents($item, false, stream_context_create($arrContextOptions));
-				            $mail->addStringEmbeddedImage($image, $countcheck, 'graph.png', 'base64', 'image/png');
-				$countcheck++;
-				                }
-
-				$mail->Body = $newcidhtml;
+            preg_match_all('/ alt=\"(.*?)\"/', $mail->Body, $match);
+            $countcheck = 0;
+            $newcidhtml = $mail->Body;
+            foreach ($match[1] as $item) {
+                $newcidhtml=preg_replace("/\bembedimage\b/i", $countcheck, $newcidhtml, 1);
+                $image = file_get_contents($item, false, stream_context_create($arrContextOptions));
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jozefrebjak** [comment]
Hi, thank you for implementing this to librenms. So i applied 8646 to my librenms. Then i added to config.php `$config['allow_unauth_graphs'] = true;` and then I added
 
`http://server/graph.php?to=1525259400&id=1555&type=sensor_temperature&from=1525173000&height=494.5&width=1670.4`

But in my mail I got only this, 
graph

I am used google gmail destination address. What I am doing wrong ? 

Thank you.

**Comment by pheinrichs** [comment]
Thanks for this PR. This looks really good so far. Just a thought though, (laf or murrant might have other ideas). But instead of allowing uauth graphs.  Maybe use similar logic that the api uses for exposing a graph. 
The 'includes/graphs/graph.inc.php'  can take a $vars['output'] value of 'base64', this would capture the graph and output the $base64_output for you to use?

**Comment by chasgames** [comment]
@jozefrebjak Thanks for testing already! did you copy the link from the graph from the device overview page? I think you have to click into them (the 6 hour view page etc) and then copy the image link.

 If that doesn't work, can you logout, and paste that URL into your browser and check you can see the graph?

@pheinrichs Thanks for your suggestion :), it's currently based off the way current html emails are working so at the moment you need to set unauth graphs to true anyway. This request was just to make it so the images are embedded so not reliant on the connection to LibreNMS instance.

**Comment by jozefrebjak** [comment]
@chasgames  So i did logout and pasted url from copy image address and there was a problem, I cant get data, I only saw that empty graph.  So I looked into a config.php and there was a problem I had in config $config[‘allow_unauth_graphs‘] = true; and right is $config['allow_unauth_graphs'] = true;. Now it is working. Thank you for your help. I am going to test it into mails.

**Comment by murrant** [comment]
Two thoughts:
1. would be hard to embed a copy of the graph (with a link to the actual graph)?  That would avoid the unauth requirement.
2. We are about to merge Laravel hopefully, so after some work, we could use something link https://github.com/spatie/laravel-url-signer

**Comment by laf** [comment]
I think this is definitely a good thing but my plan was to approach it slightly different as I don't think calling a url to get the graph is the way forward. We should look at this once the laravel PR is merged in as I'll be updating how templates are done at that stage.

**Comment by chasgames** [comment]
Thanks for your feedback guys,

@murrant 
1 & 2.
Not sure how to do this without unauth because it's required at the moment for normal HTML graph emails. Perhaps by default, we could allow unauth for the instance itself for example:
```
$config['allow_unauth_graphs'] = true;
$config['allow_unauth_graphs_cidr'] = array('127.0.0.1/32');
```
This PR is for targetted for users with a private LibreNMS instance that want to see graphs on Office365,Gmail etc.. so the link wouldn't be accessible even if we include it.

@laf 
I can only think If the input isn't an image URL, then for usability we would want something similar to this:
https://community.librenms.org/t/send-to-dashboard-link/4218

Except Send to Alert.. !

Currently a URL is needed for HTML emails with src, so presumably, would you also want to change it for current HTML emails also?


Additionally, I want to overrule if SVG is set and embed a PNG for complete email support. It looks like some more code elsewhere is required for that and I may need some help.

laf:
It's definitely time to re-visit this now Laravel is merged in.

I've just submitted the PR which updates the templating to use Laravel Blade now as well.

Embedding the actual image in the email is the better way to do that and is supported by PHPMailer. We just need to work out someway (maybe in the templates) that someone can say, include this image. Maybe something like:

`@php attach_image("type=port_bits&id=$value['port_id'}&from=-24h&to=now&width=600&height=20") @endphp`

A function then can return ``.

Then we can pull out what cid's we have, use the `ref=""` to then generate the graph internally like we do for the API and attach it to the email (or other transports).

laf:
The other option is to switch across to Laravel notifications which I know @murrant has talked about before. However it expects a file so we'd still have to generate that file before hand anyway.

Looking at the laravel docs for notifications it does look like it would be a good move but I can't help but feel that we still would have to do a lot of the grunt work as they expect to be used programatically rather than from a template. Open to ideas though

laf:
> Not sure if I understand @laf this is what i've done already? Embedding the actual image using Phpmailer ($mail->addStringEmbeddedImage) , and using Alt as the reference URL.

I know :) However we're moving to use Laravel for templating (whether that's in my PR or an alternative) so we it needs changing from how you've done it be done in blade templates. We can then at that stage uses it for attaching images in other transports. My suggestion also avoids calling the webui with unauthed graphs.

laf:
It really does need to be re-done to support all transports rather than being email specific.

laf:
Closing this down for now, will need to be done so that it supports multiple transports + not utilise the public url for generating images.

PR #8650, opened 2018-05-02
Selective Port polling improvements

murrant:
Don't delete skipped ports when using selected polling
Improve output of port polling module.
Update ifOperStatus and ifAdminStatus even when skipping.

Warning! causes a semantic change where ifAdminStatus_prev and ifOperStatus_prev are updated every poll, so they will only be mismatched for one poll.  This could cause alerts to behave differently.

Add an optimization if less than 10% of the ports are disabled (or less than 5), walk the base oids.
Time improvement on test device Full: 12s, Selective: 8s, Optimized 4s. (no disabled ports on device, many down)

Another test device.

image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -880,10 +880,6 @@ foreach ($ports as $port) {
 
         // Update Database
         if (count($port['update'])) {
-            if ($port_id == 2666) {
-                var_dump($port['update'], $port['skipped']);
-            }
-
             $updated = dbUpdate($port['update'], 'ports', '`port_id` = ?', array($port_id));
             // do we want to do something else with this?
             if (!empty($port['update_extended'])) {
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -713,7 +713,6 @@ foreach ($ports as $port) {
         if (!empty($port['skipped'])) {
             echo " $port_id skipped.\n";
         } else {
-            var_dump($port['skipped']);
             // End parse ifAlias
             // Update IF-MIB metrics
             $_stat_oids = array_merge($stat_oids_db, $stat_oids_db_extended);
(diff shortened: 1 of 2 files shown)


9er:
I'm currently testing this PR. Ports that are down don't get deleted and stay in the ports list, yay \o/

The poller log suggests that the ports that are down, are still polled for packet/octet counters (e.g. my Interface 111 is down, and the log reports `ifHCOutUcastPkts.111 = 5934353`), even if if's down for a few runs. Which means that there is some SNMP load on the device for ports that are down and admin-down.

From my point of view, that's pretty much okay, because the port could have been up in between polls and could have had some traffic on them. And I don't wanna miss any traffic. Especially when I use it for billing.

But looking at the code, I'm not sure if this is intended. I would be okay with ports being skipped, if they are admin-down.

murrant:
@9er Thanks for testing.  Can you post the debug output of your poller run?

laf:
> Warning! causes a semantic change where ifAdminStatus_prev and ifOperStatus_prev are updated every poll, so they will only be mismatched for one poll. This could cause alerts to behave differently.

Why change this?


> Add an optimization if less than 10% of the ports are disabled (or less than 5), walk the base oids.
> Time improvement on test device Full: 12s, Selective: 8s, Optimized 4s. (no disabled ports on device, many down)

I've not read the code to check but if you're saying this now would override enabling selected port polling then I'm not a fan of this. If the entire feature of selected ports polling could detect when best to be used then yes, this would be good but not for when a user specifies to enable it but it's ignored.

includes/polling/ports.inc.php
```diff
@@ -174,45 +174,85 @@
...
+            $ports[$port['ifIndex']]['skipped'] = true;
+            return $port['deleted'] || $port['disabled'];
+        });
+
+        // if less than 5 ports or less than 10% of the total ports are skipped, walk the base oids instead of get
+        $polled_port_count = count($polled_ports);
+        $total_port_count = count($ports);
+        $walk_base = $total_port_count - $polled_port_count < 5 || $polled_port_count / $total_port_count > 0.9 ;
+
+        foreach ($table_base_oids as $oid) {
```
laf: Is this not supposed to check `$walk_base`?

murrant: Yeah, I think it is.

murrant:
Seeing some gaps in graphs on some devices.  Don't merge until I can investigate.

murrant:
I think I found the bug. It was an inverted filter causing polling to delete all ports every other poll.

PR #8657, opened 2018-05-03
Update Test-Units.md

mattie47:
Missing this step meant I never understood why I was always getting:

```
		(venv) librenms@librenmsdev:~$ php scripts/pre-commit.php -u --db --snmpsim            
		Running unit tests... PHP Fatal error:  Uncaught Error: Class 'PHPUnit_TextUI_Command' not found in /opt/librenms/vendor/phpunit/phpunit/phpunit:52
		Stack trace:
		#0 {main}
		  thrown in /opt/librenms/vendor/phpunit/phpunit/phpunit on line 52
		failed
```

I happened to discover this missing step from looking at the Travis CI logs for what needs doing.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Developing/os/Test-Units.md
+++ b/doc/Developing/os/Test-Units.md
@@ -33,6 +33,8 @@ one variant, please do not specify one.
 
 ## Running tests
 
+**Note:** To run tests, ensure you have run "composer install" from your LibreNMS root directory. This will read composer.json and install any dependencies required. 
+
 After you have saved your test data, you should run `./scripts/pre-commit.php -p -u` verify they pass.
 
 To run the full suite of tests enable database and snmpsim reliant tests: `./scripts/pre-commit.php --db --snmpsim -p -u`


murrant:
@mattie47 Maybe we should just run `./scripts/composer_wrapper.php` in the pre-commit script?

mattie47:
@murrant  - Probably not a bad idea to. I did always wonder why I could run some commands and not others, and I guess because of composer_wrapper.php. 

Personally - I still think it wouldn't be a bad idea to have a note about running composer install *somewhere* in the docs though. 

Despite the config files sitting in the root folder, I still never noticed or realized their importance. 

Thanks,

Matt

laf:
+1 to both of these (running composer_wrapper + updating docs) although shouldn't it be `./composer.phar`?

mattie47:
@laf 

"composer install" within the directory works, as does "./composer.phar install". 

I don't know which is "more" correct. Up to you guys. I guess if you do "composer install" it'll probably complain if composer isn't installed making it more obvious to the user to apt install composer if they don't already have it.

murrant:
composer_wrapper will run composer.phar, but it also downloads it if it is missing and runs self-update.

murrant:
It might be worth not using composer_wrapper.php for the developer stuff.  Gives a little look behind the curtain.

mattie47:
@laf  - Right. That makes sense. 

I'd originally followed https://docs.librenms.org/#Installation/Installation-Ubuntu-1604-Nginx/ when I set up the machine which sets the user's path to /opt/librenms, possibly why that "just worked" for me.

mattie47:
So.... I forgot about this PR and I'd deleted my Librenms repo after stuffing a number of things up in Git. I couldn't see a way of adding a commit to this again (probably is a way, but spend 20 min looking), so am closing this.

Have opened PR8707 instead...Sorry if this was a stupid way.

Issue #8353, opened 2018-03-09
Alerting overhaul - looking for feedback on what needs fixing


laf:
Hi Folks,

We are currently going through doing an overhaul on various aspects of the alerting system. This includes:

- [x] The rule builder (#8293)
- [x] Update the device / group mapping for alert rules (#8293)
- [x] Update the templating system to use something more solid (#8803)
- [x] Provide the ability to map specific transports to alert rules + provide custom configuration to those transports per alert rule (i.e multiple email addresses for mail transport). (#8660)
- [x] Allow transports to be configured multiple times / multiple 'destinations' (#8660)
- [ ] Attach / embed graphs
- [x] Ability to not send recovery emails (#8430)
- [x] Record the user who acknowledged an alert + message (#8433)
- [x] Add support for grouping rules based on MySQL AVG, MIN, MAX, etc (#9094)
- [ ] Track faults individually (for example each port or bgp session)

If you like the above but have nothing to add, simply use the 👍 against this post. Do not reply with comments.

If you would like to have additional functionality / changes done during this time please reply with the answers to these questions:

1. What do you require to be added or changed?
2. Will this benefit other users?
3. Would you class this as a nice to have or a requirement for an alerting system.
4. Are you willing to be involved in helping test? If so, please do not commit to this unless you have the time to spare. We will be looking for feedback within days of writing code and cannot wait weeks for responses.

Anything comments not appropriate to this or that don't provide the correct info will be deleted.

FTBZ:
Provide the ability to map specific transports to alert rules

murrant:
Allow transports to be configured multiple times
Allow transports to provide a list of targets (like multiple channels for Slack)

murrant:
**Huge project alert** Use Laravel Notifications for delivering alerts https://laravel.com/docs/5.6/notifications

Rosiak:
Ability to manually clear active alerts from the GUI.

Sneddo:
Permissions for alert acknowledgement. 

Currently accounts with global read user level can acknowledge alerts. It would be nice to have this as an optional configuration to allow only users with administrator access to the related device able to acknowledge alerts.

Sneddo:
Acknowledgment message.

Similar to recording the name of the user that acknowledged an alert, being able to add a short message to the alert would be very helpful. e.g. If a device is down due to a power outage this could be noted with an expected ETA, or record an incident ticket number etc.

laf:
[...]

PR #8660, opened 2018-05-03
Feature: Alert Transports Mapping

vivia11:
Hello all,
I guess this is the second version of a more fully fleshed out alert contact mapping feature. The old one was GH-8507

Transports to convert:

  - [x] API
  - [x] Cisco Spark
  - [x] Elasticsearch
  - [x] GitLab
  - [x] Philips Hue
  - [x] Jira
  - [x] Mail
  - [ ] ~~PagerDuty~~ - Requires a callback so leaving for now
  - [x] Nagios
  - [x] IRC
  - [x] Discord
  - [x] Rocket.chat
  - [x] Hipchat
  - [x] Pushover
  - [x] Boxcar
  - [x] Telegram
  - [x] Pushbullet
  - [x] VictorOps
  - [x] OpsGenie
  - [x] Clickatell
  - [x] PlaySMS
  - [x] Canopsis
  - [x] osTicket
  - [x] Microsoft Teams
  - [x] SMSEagle
  - [x] Syslog
  - [x] Slack

The intention is for this feature to have three different levels to it:
1. Alert rule to an alert contact mapping (where the code is at now)
2. Alert rule to an alert group (made up of alert contacts) mapping
3. Alert contact mapping to different transport configurations.

There will be three transport configuration types.
1. Default (the configuration that is held in the configs table)
2. None (no transport configuration - will explain later)
3. Other (a configuration that will be defined in a different able)

Take Mail transport for example. It can either be of a "default" or "other" configuration. The hope is that in the future, users can send mail from different mail servers if they wish.
However, for ciscospark which requires a room ID and an api-token, I've decided that it has no transport configuration. Most likely, every alert contact will contain a different room-id and an api-token - which is why it has the transport config of "none".
For other transports : I am not familiar with them, so hopefully the community can add support for these. I can definitely help!

To add support for each transport will require several things:
- addition to the UI
- addition to forms/alert-contacts.inc.php
- modifications to its object class

Screenshots
image
image
image

I'm not sure if this is the best way to do things, so please let me know if there's a better way to structure the code! Any comments on code/db schema,/UI etc is welcome and encouraged! 

The UI is heavily based on alert rules (front end is not my strong suit). And parts of the code are based on the code that was written for alert rules.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


jozefrebjak:
Its looking really great. I cant wait when I can use this in my install.  For now I am getting `sql-schema/248.sql: already exists in working directory`

murrant:
@vivia11 this needs to be updated there are merge conflicts with the main branch

murrant:
It doesn't have to be like this but awhile ago, I sketched out some ui for this. The contact select box you have is pretty close.

image

vivia11:
@murrant 
I changed the sql-schema file name, so that conflict should be resolved

Instead of adding transport configs per user, what about adding the functionality of contact groups? I think it covers the same use case as the one you describe.

You could have a contact group for each user. Or you could combine several users together into one contact group - it wouldn't be mapped on a per user case, but it allows for more flexibility that way.
With contact groups you could:
- have a generic contact group containing a distribution list email, a slack channel (and other generic transport types, I'm not familiar with them)
- map all types of transports for one user (ie all transports for user A) into one group
- map all types of transports for a group of users (all transports for users related power devices) into one group
- mix and match: have some generic transports and specific user transports
- map transports for team members who don't necessarily use librenms and so don't have an account on it

So contact groups would allow for:
1. Generic transport types
2. Transport configs for a specific user (you would just create a contact group for each user if you wish)
3. Both

Let me know what you think.

czilian:
Maybe you could consider a similar approach like for authentication making reference to objects residing either in local DB, Radius, AD, LDAP. Then it would be consistent from a security as well as management perspective.

murrant:
I'm not a very heavy alert users, it just goes to one slack channel for me.  Others are going to need to way in.  I don't want to make things too complicated.

FYI, in the future there will be a db entry for all users, no matter the auth method.

theherodied:
I like the idea of groups. Setting up groups now will make life simpler in the future. If you have a one off case for a specific user then you can create a group for them. 

It's the same theory everyone talks about with AD. Use groups, don't do one offs for accounts because of the pain it will cause later.

vivia11:
I don't want this PR to be too complicated either.
I'm hoping to keep this PR to alert mappings with contacts/contact groups.

There is definitely room in the future to extend the functionality of this feature. We can map a contact to a user in the future and we are already planning to be able to map contacts to different transport configurations (but in a different PR). We could even map contacts/contact groups to AD groups if there is a want/need for it.

I'm just hoping that we can get the basics of this feature out and once we have the foundation, we can take a look at the community feedback and plan our steps moving forward.

@murrant 
I like your UI sketch. I'm not the best at front end, but I'll try implementing your design.

laf:
Thanks for putting this together @vivia11 :)

So I'd planned to tackle this off the back of: https://github.com/librenms/librenms/issues/8353

My original idea to tackle this was to allow each transport to be configured in the rule but this idea allows them to be re-usable which I think is better. However I am wondering if it's worth moving all of the transports out of the config table into the table you've created here. Basically it just moves the configuration of transports to a different location.

I don't think the word contacts is right though, indicates it's email only so this should be changed and I think calling it transports wouldn't be far wrong if we moved them all to this table.

I can't tell what the two disabled buttons would be for although I'll guess the group is to just group together a few transports and then apply the group to the rule?

vivia11:
Aha, I guess I didn't explain my idea that well.

So right now for email transport configurations, we have the following config values:
- email address to deliver to
- email address to deliver from
- smtp host
- smtp port 
etc.

With the idea of contacts/transport separation, I'm saying:

Contact config:
- email address to deliver to
-------
Transport config:
- email address to deliver from
- smtp host
- smtp port
- everything else

It's basically an arbitrary definition we give to each transport type.
That means every "contact" object, is associated with:
 - a contact configuration
- a transport configuration 
- contact groups (more than one possible mapping)
This PR focuses only on the contact configuration/alert rule mapping and will use the default transport config (one in configs table). Although, with murrant's earlier comment, I might add contact-groups to this one as well.

Alert rules can only be mapped to contacts or contact-groups. The different transport configurations are applied at the contact level. Of course, the issue is, with the other transports which I have little to no knowledge of, I'm not sure where to draw the arbitrary line. 

So the two disabled buttons:
- one is for creating contact groups
- one is for creating different transport configurations: for my mail example, it would be creating different mail server configs - users would then have to use the add contact button to associate a contact with a transport config.

However, for some transport types, it doesn't make sense to draw an arbitrary line between contact and transport config. Eg ciscospark, so I added the option of just having a contact config, but no associated transport config. 

Hopefully I made sense explaining my difference between "contact" and "transport". And yeah, we can change the name contact - but I'm not sure what to call it either. 

Also, moving the configs to the other table sounds fine, but should that be done in a different PR? After we've tested this one out?

laf:
I think you've explained it fine - however I'm not sure I agree that this is the way forward then. Aside from mail, the 'contact' config would be the 'transport' config you are referring to (mail delivery options could be left in the config page). Pretty much all current transport configs we allow would need to be specified per contact anyway to allow those alerts to be routed differently within that transport.

As for moving the config for transports to another table being another PR then again I'd personally say no as this means we will have two lots of config in two tables we then need to merge together.

Users should just configure transports in one area and only once if needed (they can configure multiple if they want), then they can map the transport or group to an alert. I'm happy to help work on that as this was part of the alerting updates we were looking to do anyway.

vivia11:
I see what you mean. If Mail Transport is the only transport that is "different" then we can treat that one slightly different from the others.

If we don't need to distinguish between contact config and transport config (because as you said, they would all be per contact), this actually really simplifies the code and DB schema. So now we just need to map contacts to alert rules and contact-groups.  And since that becomes the case, then we should migrate the existing configs to the current table.

But tbh, I'm not sure how to do that through code?

On a second note, since I'll be editing the code - did you have a name you preferred over contact? I might as well change the files/vars while I'm at it.

murrant:
I was reading through the alerts wish list issue and I think some sort of alert template (what you called alert group) is the best way to solve things like alert escalation (like first alert these users, then alert these users).

vivia11:
I've added the ability to group different contacts together to form contact groups and these of course, can be mapped to an alert rule.

As previously discussed, I've changed some code (updated db schema) so that contacts can no longer be mapped to different transport configs - since this is probably unnecessary. This simplifies things :)

With the two transport classes I've implemented (mail and ciscospark), I've made so if the specified transport config doesn't work, the code reverts back to the default and attempts that instead. I'm not sure if this is a standard we can/should apply to the other transport classes. 

image

image

@murrant 
I'm not sure how to implement your UI without writing a lot of extra code since it would have to be dynamically generating modals for each configured contact. Since right now the code points to the same modal used to create/update a contact config. To be completely fair, I'm terrible with front end...

I'm wondering if there can be assistance in supporting the other transport classes/related UI work? Or we can create the support as requested?

murrant:
Don't worry about the front end too much for now.

However, I think for the configuration, you could add new method to the Transport interface to fetch the config template.
It would be similar to the current dynamically generated config, but would all be contained in the class. Defaults would be defined here as well.

An abbreviated example implementation:
```php
public function configTemplate()
{
    return [
        [
            'name' => 'email_user',
            'descr' => 'From name',
            'type' => 'text',
            'default' => Config::get('project_name', 'LibreNMS'),
            'required' => true,
        ],
        [
            'name' => 'email_from',
            'descr' => 'From email address',
            'type' => 'text',
            'pattern' => '[a-zA-Z0-9_\-\.\+]+@[a-zA-Z0-9_\-\.]+\.[a-zA-Z]{2,18}',
            'default' => Config::get('project_id', 'librenms') . '@' . php_uname('n'),
        ],
    ];
}
```

laf:
I'll need a bit more time to jump on this PR and offer my insight.

alydharshi:
@laf any further thoughts on this, we'd like to conclude this work and then move Vivia to different coding work. If there are any further items we need to work on for this we'd like to know soon. Thanks.

laf:
> @laf any further thoughts on this, we'd like to conclude this work and then move Vivia to different coding work. If there are any further items we need to work on for this we'd like to know soon. Thanks.

I really like this approach.

It needs some tidying up with regards to the naming of certain things + the table / column names to be more inline with how the rest are. Needs the other transports adding in to this as well + converting the current transport config into this new format.

Happy to help with all of this when I have time, grab me on discord if you are on there.

murrant:
I think the fundamentals are good now for the most part.  Could use a few passes of testing and refinement now.

Notes:

- UI naming does need some work.  I can't think of anything though... terrible at that.  Maybe Alert Targets or Notification Targets.  Notification Templates instead of Contact Groups?  Contacts doesn't seem 100% right with mixing contacts and channels.
- Should restrict the add contacts transport list to just configured transports.
- There should probably be some built ins like sysContact instead of that being a all or nothing global setting.
- configTemplate() function might need a rename since we probably need a generic template for the transport too, but we can deal with that later too.

vivia11:
@murrant and @laf 

1. I think Alert Targets (replace alert contacts) and Target Groups (contact groups) to me sound like reasonable names. If you guys agree with this, I can start renaming the variables/files.

2.
> The contact config should probably be stored as a json array in the database since it is dynamic.
So if I'm understanding this correctly, the entire contact config would be stored as a json string in a column in the db (alert_contacts) ? If I'm understanding correctly, that would be great and I could implement this. This would get rid of another table (alert_configs).

3. 
> Should restrict the add contacts transport list to just configured transports.
I'm not sure what you mean by this murrant?
I changed it recently so contacts can be configured as any transport type (there is no longer a distinction between contact config and transport config). And when you map a rule to a contact, only configured contacts will show up as an option.

laf:
> I think Alert Targets (replace alert contacts) and Target Groups (contact groups) to me sound like reasonable names. If you guys agree with this, I can start renaming the variables/files.

In essence the Alert contacts page contains: Transports (was Alert contact) and Transport Groups (was Contact group). Transports are being defined here along with grouping them together.

> So if I'm understanding this correctly, the entire contact config would be stored as a json string in a column in the db (alert_contacts) ? If I'm understanding correctly, that would be great and I could implement this. This would get rid of another table (alert_configs).

I'd be fine with that. We don't need to search or anything on that data so storing as json is fine (well, maybe serialized data).


> Should restrict the add contacts transport list to just configured transports.
> 
> I'm not sure what you mean by this murrant?
> I changed it recently so contacts can be configured as any transport type (there is no longer a distinction between contact config and transport config). And when you map a rule to a contact, only configured contacts will show up as an option.

Yeah maybe a mix up here. IMHO we should move all the transport configs into this new page. It makes the most sense and keeps it all in one.

laf:
After last nights conversation I've unsubscribed from this thread as it's obvious we won't reach an agreement on how it should work.

Tag me when it's ready for testing and I'll give it a test.

murrant:
I think that sounds fine with me @vivia11

vivia11:
@laf 
Testing would be great if there is time. :)

So I've allowed defaults to be configured on the new alert transports page. I haven't done anything with the old page, but I've made it so any new default configurations will override the old ones. Once we can get support for all the other transports, we can fully get rid of the old page and all transport configs will be held in the new table. I don't think we should get rid of the old page/code when we still don't have support for all the transports. This way transport support can be added gradually and hopefully everything will work simultaneously.

Instead of having the default transport configs in code (as you mentioned @murrant), I thought it would make more sense for them to be stored in the new table - that way users can configure two default emails, or four default slack channels if they wish etc. 

So regarding the prioritization of default vs non default transports. If an alert rule is mapped to at least one transport or transport group (note: you can map a rule to a default transport config), then the alert goes out to the mapped transports. If there is no mapping, then the alert goes out to all transports that are configured as a "default".

laf:
@vivia11 

Some points from me:

  - The icon for the menu should be changed. Maybe use car or bus.
  - Setting a transport to default. How is it decided which is the actual default if you have more than one or is it a case of they will all be used?
  - Would be good to move the new functions in `includes/alerts.inc.php` to OOPS - i.e to `LibreNMS\Alert\` somewhere.
  - Anywhere you access `$_POST` or `$_GET` change to `$vars`.
  - In `html/includes/forms/alert-transports.inc.php` where you grab config values, no way to use the array in the transport to assign that rather than having to basically do the same for each transport in the form file?
  - Need to move:

```
        $sql = "SELECT `transport_config` FROM `alert_transports` WHERE `transport_id`=?";
        $details = json_decode(dbFetchCell($sql, [$opts['alert']['transport_id']]), true);
```
That's in each transport to a central place.

Seems to work well in my limited tests.

I would say though, this definitely needs the other transports adding to this otherwise we will end up with two sections to configure transports.

I'd be happy to convert some of the other transports over if you test my templating PR :) #8803

vivia11:
@laf 
Thanks for the testing!
I'll start working on your suggestions ASAP and I'll try giving your PR test. :)

> Setting a transport to default. How is it decided which is the actual default if you have more than one or is it a case of they will all be used?

All defaults will be used if there is no rule mapping - kinda like how it currently is (alerts are sent out to all default configured transports). Except now, people are able to have more than one default email/slack channel etc.

html/includes/modal/edit_alert_transport.inc.php
```diff
@@ -0,0 +1,246 @@
...
+                        </div>
+                        <div class="form-group" title="The transport is default.">
+                            <label for="default" class="col-sm-3 col-md-2 control-label">Default Alert: </label>
+                            <div class="col-sm-2">
+                                <input type="checkbox" name="is_default" id="is_default">
+                            </div>
+                        </div>
+                    </form>
+<?php
+$transports = Config::get('alert.transports');
```
laf: So this only gets transports we've added to the DB. Not all of them will have db entries so we need to find another way to loop through the actual transports.

LibreNMS/Alert/Transport/Ciscospark.php
```diff
@@ -12,18 +12,20 @@
...
 class Ciscospark implements Transport
 {
     public function deliverAlert($obj, $opts)
     {
-        $token  = $opts['token'];
-        $roomId = $opts['roomid'];
-        $text   = strip_tags($obj['msg']);
-        $data   = array(
-            'roomId' => $roomId,
+        $details = AlertUtil::getTransportConfig($opts['alert']['transport_id']);
```
laf: We should try and move this into the Transport class, maybe in __construct() otherwise we're just duplicating the code all the time.

html/includes/modal/edit_alert_transport.inc.php
```diff
@@ -0,0 +1,246 @@
...
+    echo '<form method="post" role="form" id="'.strtolower($transport).'-form" class="form-horizontal transport">';
+    echo '<input type="hidden" name="transport-type" id="transport-type" value="'.strtolower($transport).'">';
+   
+    $tmp = call_user_func($class.'::configTemplate');
+
+    foreach ($tmp as $item) {
+        echo '<div class="form-group" title="'.$item['descr'].'">';
+        echo '<label for="'.$item['name'].'" class="col-sm-3 col-md-2 control-label">'.$item['title'].': </label>';
+        echo '<div class="col-sm-9 col-md-10">';
+        echo '<input type="'.$item['type'].'" id="'.$item['name'].'" name="'.$item['name'].'" class="form-control" ';
```
laf: We need to support checkboxes and selects here as well.

LibreNMS/Alert/Transport/Ciscospark.php
```diff
@@ -38,10 +40,54 @@ public function deliverAlert($obj, $opts)
...
+                'type' => 'text',
+                'required' => true,
+                'pattern' => '[a-zA-Z0-9\-]'
+            ]
+        ];
+    }
+
+    public static function configBuilder()
+    {
+        global $vars;
```
laf: We need to avoid using `globals` as well.

laf:
I've left some more comments inline. We've talked about this before but we also need to either convert existing transports into the new format or the transports themselves need to fall back to the old way.

laf:
https://p.libren.ms/view/raw/0c4ef5e1

That's my conversion of API. It supports both old and new methods currently.

vivia11:
>  we also need to either convert existing transports into the new format or the transports themselves need to fall back to the old way

Currently, for default transports, the code sends it out to any converted configured transports but also to old configured transport types after checking if the transport type is one that has not been converted. So it is not necessary for all transport types to be converted immediately, since the old default transports should still be used. Unless you meant something else? I could update Mail and Ciscospark to work simillarly to your API class if you wish to have two separate methods.

html/includes/modal/edit_alert_transport.inc.php
vivia11: Yes for the returned transport template, for type we could have options: text, checkbox, select etc. And the html code would have an if/else or case statement to deal with how it shows up

laf:
Maybe something like this:

```php
    public static function configTemplate()
    {
        return [
            'config' => [
                [
                    'title' => 'API Method',
                    'name' => 'api-method',
                    'descr' => 'API Method',
                    'type' => 'text',
                ],
                [
                    'title' => 'API URL',
                    'name' => 'api-url',
                    'descr' => 'API URL',
                    'type' => 'text',
                ],
            ],
            'validation' => [
                'api-method' => 'in:GET,POST',
                'api-url'    => 'required|url',
            ]
        ];
    }
```

Then:

```
        $result = call_user_func_array($class.'::configTemplate', []);
        $valid = array_map(function ($validation) use ($vars) {
            return Validator::make($vars, $validation);
        }, $result['validation']);;
```

We do need to get Laravel in for `ajax_form.php` to get this working but we should do that.

vivia11:
@laf 

I've added in the validation as recommended. 

> We do need to get Laravel in for `ajax_form.php` to get this working but we should do that.

Should that be done in a different PR, since that will affect all ajax form calls and not just ones that my PR uses? I've added in the validation for the transport classes and they can be used at a later date. I don't want to keep increasing the scope for this PR.

laf:
@vivia11 If you can mark on the list I put in your original post what transports you're working on so we can avoid duplication that would be ace :)

laf:
I've rebased this from master now as it was missing the Discord transport + needed laravel to make it work. You'll need to pull from your remote repo before carrying on (or at least should).

tmacgbay:
TEST: I put this PR in yesterday and the initial tests work without issue. I have three tiers of e-mail distribution groups each based on the priority of the alert.   (Alert-P1@<company>.com, Alert-P2@<company>.com, Alert-P3@<company>.com) "Alert Transport setup", association on an Alert Rule, and notification of e-mail on firing of the rule worked as expected.   We do not use any other transports currently.

I would note that calling it an Alert Transport makes it confusing. In truth, this is an action based on an alert that uses a particular Transport (e-mail in my case).  For future extensibility where a variety of actions could be executed on an alert event some of which may not use a transport, maybe we can call it an "Alert Action"?  i.e. in the future we may want to cascade another alert rule to check if a series of events are happening... or start/stop a service... 

On a side note, this PR is the last piece for us to have in place so that we can drop Solarwinds.  :-)

Maybe note the right place to ask but I am sure it is a quick answer - what is the command in Librenms to bring down the most recent commit?

jozefrebjak:
hello, really thank you for this PR. I am trying to test this but when i want create alert transport i get this error. What do you think where is the problem ? I added this PR by github-apply 8660. Thank you

error

tmacgbay:
@jozefrebjak you may have an older version, remove the PR and add it back in.  This problem happened to me and the remove/reinstall fixed it (forgot to report that above, I just assumed I originally had a bugged version).

tmacgbay:
TEST-INCIDENT:  I had only set up one rule to contain the new alert transport (Alert Action), it was the first rule in my list (Alerting on reboot).  all other alerts I have received are now sending to that alert transport value rather than to the default contact (which is all that I have on at the moment)

image
image
image

laf:
Ok so this is what I've tested:

- [x] Single transport set as default (mail)
- [x] Two transports set as default (mail)
- [x] Three transports set as default (mail)
- [x] Three transports set as default (mail, slack, pushover)
- [x] Single transport not set as default but mapped to one rule (mail)
- [x] Two transports not set as default but mapped to one rule (mail, slack)
- [x] Single transport not set as default but mapped to two rules (mail)
- [x] Two transports not set as default but mapped to one rule each (mail, slack)
- [x] Transport group with one transport set as default (mail)
- [x] Transport group with two transports both set as default (mail)
- [x] Transport group with one transport not set as default but mapped to one rule
- [x] Transport group with two transports not set as default but mapped to one rule
- [x] Transport group with two transports not set as default but mapped to two rules
- [x] Two Transport groups with one transport each not set as default but mapped to two rules, one each 
- [x] No transports mapped or set as default. Should fall back to mail

Not seen any issues with anything.

@tmacgbay Might be worth giving the latest version a go as I couldn't replicate your issue.

tmacgbay:
Not sure this will get through - I am on vacation to next week... looking forward to testing!!



-Tad

PR #8662, opened 2018-05-04
Added AP/Client Count detection for FortiWLC

centralscrutiniser:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

- Added MERU-* mibs under mibs/fortinet
- Added FortiWlc.php AP/Client count detection sensors under Librenms/OS
- Added initial polling configuration under includes/definitions/os/


mibs/fortinet/IPV6-TC
```diff
@@ -0,0 +1,67 @@
+IPV6-TC DEFINITIONS ::= BEGIN
```
murrant: IPV6-TC is a standard mib, it shouldn't be included here (it is already in the base directory)

config.php.default
```diff
@@ -1,51 +0,0 @@
-<?php
```
murrant: This file must not be deleted.

murrant:
Changes look good, just need to clean up the mibs and accidental file deletion.

Not required, but would be great if you added some test data too. https://docs.librenms.org/Developing/os/Test-Units/

includes/polling/os/fortiwlc.inc.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+use LibreNMS\RRD\RrdDefinition;
+
+$version = trim(snmp_get($device, "MERU-GLOBAL-STATISTICS-MIB::mwSystemGeneralVersion.0", "-OQv"), '"');
```
murrant: You don't need to trim() these others have done that as a legacy thing which hasn't been cleaned up.

Also, it would be better if you got all the data in one operation with snmp_get_multi(), there should be some other os polling files that use that you can look at for an example.

includes/polling/os/fortiwlc.inc.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+use LibreNMS\RRD\RrdDefinition;
```
murrant: You don't use that here, you can delete this line.

includes/polling/os/fortiwlc.inc.php
```diff
@@ -0,0 +1,7 @@
...
+
+$version = trim(snmp_get($device, "MERU-GLOBAL-STATISTICS-MIB::mwSystemGeneralVersion.0", "-OQv"), '"');
+$hardware = trim(snmp_get($device, "MERU-GLOBAL-STATISTICS-MIB::mwSystemGeneralModel.0", "-OQv"), '"');
+$hostname = trim(snmp_get($device, "MERU-GLOBAL-STATISTICS-MIB::mwSystemGeneralName.0", "-OQv"), '"');
```
murrant: $hostname does not do anything https://docs.librenms.org/Developing/os/Initial-Detection/#poller

murrant:
@centralscrutiniser You'll need to add test data for this device.  https://docs.librenms.org/Developing/os/Test-Units/

includes/polling/os/fortiwlc.inc.php
```diff
@@ -0,0 +1,4 @@
+<?php
+
+$version = snmp_get($device, "MERU-GLOBAL-STATISTICS-MIB::mwSystemGeneralVersion.0", "-OQv");
```
laf: Can you convert these two `snmp_get()` calls into one `snmp_get_multi_oid()`. You also don't and shouldn't prepend the MIB name to the OID, pass it as an arg to the function.

murrant:
Looking pretty good, if you can just add the snmprec with `./scripts/collect-snmp-data.php -h HOSTNAME`, I'll add the json data.  Feel free to add the json data too if you are up to getting snmpsim running.

centralscrutiniser:
Hi @murrant Thanks. I've updated the snmprec. If you're able to help with the test data, I'd be really grateful. Having difficulties with the snmpsim portion via ./scripts/collect-snmp-data.php. Probably need to discuss on Discord.

PR #8665, opened 2018-05-06
Feature: Plugins in the Port page

PipoCanaja:
Added new plugin menu_option in the "port" page, which contain hook calling public function port_container($device, $port) in plugins.
The display of this menu_option is conditionnal, and enabled with the following option in config.php: 
$config['enable_ports_plugins'] = 1;

pluginmenuoptionwithtestpluginenabled

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Plugins.php
+++ b/LibreNMS/Plugins.php
@@ -82,6 +82,16 @@ class Plugins
         return $plugin;
     }//end load()
 
+    public static function countHooks($hook)
+    {
+        // count all plugins implementing a specific hook
+        self::start();
+        if (!empty(self::$plugins[$hook])) {
+            return count(self::$plugins[$hook]);
+        } else {
+            return false;
+        }
+    }
 
     public static function call($hook, $params = false)
     {
--- a/doc/Extensions/Plugin-System.md
+++ b/doc/Extensions/Plugin-System.md
@@ -4,8 +4,9 @@ source: Extensions/Plugin-System.md
 > This will most likely be deprecated in favour of adding the possible extensions to the core code base.
 
 This documentation will hopefully give you a basis for how to write a plugin for LibreNMS.
+A test plugin is included in LibreNMS distribution.
 
-A test plugin is available on GitHub: https://github.com/laf/Test
+### Generic structure ###
 
 Plugins need to be installed into html/plugins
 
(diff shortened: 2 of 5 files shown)


murrant:
@PipoCanaja Thanks for your PR.  
Can you please link to an implementation? We are unlikely to merge something with nothing that uses it.

Also, because LibreNMS is an open source project, we are open to merging things directly into LibreNMS instead of requiring additional set up.

PipoCanaja:
Hello,

Basically, I have a plugin that add a few "SNMP Write" and "CLI Write" commandes to LibreNMS, which allow our NOC to shut/no shut ports, clear PortSecurity sticky Mac Addresses, View dot1X status, change Description and VLAN. 
This plugin is for the moment absolutely not portable (checks a specific description syntax to find if the port should be available for NOC, uses Huawei MIB to access our switch model etc etc ...).

As adding RW capability was clearly out of LibreNMS scope, I thought of this simple extension of the PluginStructure so everybody can actually extend LibreNMS in this (or other) directions, without breaking the AutoUpdate capability. 

What kind of implementation (in the test Plugin) would you like to see ? For the moment, as shown in the capture, I did exactly the same as the Test Plugin already did with the "Device" page : putting a simple text so that any future user would see where it's code would access to the GUI. 

I can easily provide a simple "Shut/NoShut" plugin, with the standard MIB, for instance. 

Bye

PipoCanaja:
Hello @murrant 

You'll find a test plugin that allows to Shut/No shut ports using the IF:MIB. 

https://github.com/PipoCanaja/InterfaceAdminPlugin

If you would like to include it directly into LibreNMS, I can add it to the currant Pull Request. 

PipoCanaja

laf:
I really wanted to get rid of the plugin system as it wasn't actively used but a few people keep popping up who do.

If we were to merge this then I would say that it would be good if you could add the plugin to https://github.com/librenms-plugins However you should also just use the existing device snmp creds rather than having another `$config` option that people have to set.

You should also remove the `$config` check, if you enable the plugin you shouldn't need to also setup a new config option.

PipoCanaja:
Hello Iaf

I don't know about the others, but I think that the plugin system is interesting to cover things that cannot/should not be included into librenms for any reason. It also allow some personnalisation without breaking the auto-update mechanism ... 

Concerning the plugin, I'll push it as asked. For the community, the point is that the current SNMP Community configured in LibreNMS is RO in my deployment. That's why I used a different one (in a quick and dirty way, I must say). The "nice" way would probably be to have a new property, per device, with RW community. 
But for the moment, I'll change to use the community stored in LibreNMS. 

For the $config check, I need to find a way to "ask" the plugins about which functions are provided, and if no plugin provide a "port_container" function, then I don't need to display the "plugin" menu_option at all. For the moment, the $config option does that manually. 

I'll review the code following those recommandation and come back to you.
Bye
PipoCanaja

PipoCanaja:
Hi @laf ,

I added a method in Plugins.php to count the number of Plugins implementing a specific hook. This allow to display the "Plugins" menu_option automatically as soon as at least a plugin implements it. 

I also corrected the Plugin itself to use LibreNMS community instead of a specific $config new param. 

And the push to https://github.com/librenms-plugins is on it's way. 

Bye

PipoCanaja:
Hi again,

Cannot make it  creating a PR to https://github.com/librenms-plugins . In fact, I cannot even clone it. Do I miss something or the rights do not allow me to clone it ? 

If you like, you can directly pick https://github.com/PipoCanaja/InterfaceAdminPlugin and add it to the https://github.com/librenms-plugins .

Thanx
PipoCanaja

laf:
> If you like, you can directly pick PipoCanaja/InterfaceAdminPlugin and add it to the @librenms-plugins .

I've cloned it, if this goes ahead then we'll give you access to the repo in librenms-plugins.

kkrumm1:
Looks good to me

PipoCanaja:
Hello
2 checks are still in "expected" status. Is it something normal, should I do something to correct ? 
Thanx
PipoCanaja

PipoCanaja:
Hello @laf 

2 checks are still in "expected" status. Is it something normal, could I do something to correct ? I would be very interested to get the PR merged so that I can get it into the "prod" machine instead of the test one ;) 

Thanx
PipoCanaja

laf:
I've just removed the `?>` from one of the files to kick the checks off again.

PipoCanaja:
Let me know if you need more documentation. The existing device_overview_container was missing and I added the port_container as well.

PipoCanaja:
Hello @kkrumm1 @laf @murrant ,
I am wondering if this PR would have a chance to get merged this week. Sorry to push but I would love to clean up the prod librenms with this before being out for a couple of days of holidays. 
Thanx and bye
PipoCanaja

murrant:
I'm kind of meh on the concept, but I'm not going to block it.

laf:
I'm going to merge it in and see where this goes. I did want the plugin system dropped as it's not really used but if it is useful and can be expanded then I'm ok with running it for a bit longer before we decide.

PipoCanaja:
Thanx ! 
Bye
PipoCanaja

PR #8674, opened 2018-05-08
Device: Add Open Access Netspire OS support

vivia11:
Adding OS support for OA Netspire devices.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


vivia11:
@laf 

Thanks for the suggestions, I've made some changes to the code. I'm not sure if I should be unsetting my extra variables as I use them or not...

includes/discovery/mempools/netspire.inc.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+if ($device['os'] === 'netspire') {
+    echo 'Netspire Memory Pool\n';
+    $usage = str_replace('"', "", snmp_get($device, 'netSpireDeviceStorageUsed.0', "-OQv", 'OACOMMON-MIB', 'openaccess'));
```
laf: You can drop the `openaccess` from the end of the `snmp_get()` call, this is already pulled in from the yaml file.

includes/polling/mempools/netspire.inc.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+$usage = str_replace('"', "", snmp_get($device, "netSpireDeviceStorageUsed.0", "-OQv", 'OACOMMON-MIB', 'openaccess'));
```
laf: Same here.

includes/polling/os/netspire.inc.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+$oids = array('netspireDeviceModelName.0', 'netSpireDeviceDeviceSerialNo.0');
+
+$data = snmp_get_multi_oid($device, $oids, '-OUQs', 'OACOMMON-MIB', 'openaccess');
+
+$hardware = $data[$oids[0]];
```
laf: This is easily broken I think. Take a look at how it's done in `includes/polling/os/ipolis.inc.php`

includes/definitions/netspire.yaml
```diff
@@ -0,0 +1,9 @@
+os: netspire
```
laf: So netspire appears to be a range of products. Does netspire cover them all (i.e they run the same firmware)?

vivia11: We have two different devices (audio controller and paging station) which run the same firmware. I am not sure if this extends to their entire range. However, their oacommon-mib covers their entire netspire range.

laf:
Awesome work as always :)

PR #8704, opened 2018-05-12
Added feature to filter delayed alerts on widget

DR3EVR8u8c:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

my work place is used to zenoss’ way to display warning and critical in different colors and events only turn critical after retries and delayed. when I introduce librenms, they feel strongly against the alert widget and alerts panel due to librenms has different concept of warning and different ways of alert.

Since my colleagues are used to watch display screen instead of email or other transporter, it is big concern for them if false/non-confirmed alerts keep showing on the dashboard. Although I have gone through the librenms performance tips and followed the procedures, we are still getting false alerts for Device up/down. Therefore, I decided to customize the display and only show delayed alerts, which is confirmed and sure.

I have made the changes on alert widget and it is working as expected so far. I thought someone else may want this feature as well.


DR3EVR8u8c:
have fixed the build job error, please pull again

murrant:
Looks ok to me.  My only concern is the setting name, "Delayed alerts" doesn't seem very clear to me.

DR3EVR8u8c:
Hi,
Thanks for your comment. What would you suggest if it is not delayed?

Since the "alerted" has already been used in librenms, I couldn't think of
any better name. Unless we change the current name of "alerted", I still
think"delayed" may be the best name of the state.

It shows what is happened and indicates that it has passed the delay period
of the alert rules.

Cheers,
Roger

DR3EVR8u8c:
Hello,
I have been thinking the name today and I am wondering if "fired alerts" feel more senses than that. just want to know your opinions before I go ahead to push up to the github.
Cheers,
Roger

laf:
`Triggered alerts`?

DR3EVR8u8c:
`Fired alerts` ?
Triggered alerts seems not clear as well.

laf:
> Fired alerts ?
> Triggered alerts seems not clear as well.

Is fine with me.

PR #8705, opened 2018-05-12
Feature: PoE graphs

PipoCanaja:
Hello,

I did some tests and enhancements in the POE polling code (Cisco, Huawei), and also added graphing code in the port page. 

Then code has been tested onto Cisco 2960X, and Huawei 5720 access switches.

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -4,175 +4,190 @@
 use LibreNMS\RRD\RrdDefinition;
 
 $data_oids = array(
-    'ifName',
-    'ifDescr',
-    'ifAlias',
-    'ifAdminStatus',
-    'ifOperStatus',
-    'ifMtu',
-    'ifSpeed',
-    'ifHighSpeed',
-    'ifType',
-    'ifPhysAddress',
-    'ifPromiscuousMode',
-    'ifConnectorPresent',
-    'ifDuplex',
-    'ifTrunk',
-    'ifVlan',
-);
+        'ifName',
+        'ifDescr',
+        'ifAlias',
+        'ifAdminStatus',
+        'ifOperStatus',
+        'ifMtu',
+        'ifSpeed',
+        'ifHighSpeed',
+        'ifType',
+        'ifPhysAddress',
+        'ifPromiscuousMode',
+        'ifConnectorPresent',
+        'ifDuplex',
+        'ifTrunk',
+        'ifVlan',
+        );
 
 $stat_oids = array(
-    'ifInErrors',
-    'ifOutErrors',
-    'ifInUcastPkts',
-    'ifOutUcastPkts',
-    'ifInNUcastPkts',
-    'ifOutNUcastPkts',
-    'ifHCInMulticastPkts',
-    'ifHCInBroadcastPkts',
-    'ifHCOutMulticastPkts',
-    'ifHCOutBroadcastPkts',
-    'ifInOctets',
-    'ifOutOctets',
-    'ifHCInOctets',
-    'ifHCOutOctets',
-    'ifInDiscards',
-    'ifOutDiscards',
-    'ifInUnknownProtos',
-    'ifInBroadcastPkts',
-    'ifOutBroadcastPkts',
-    'ifInMulticastPkts',
-    'ifOutMulticastPkts',
-);
+        'ifInErrors',
+        'ifOutErrors',
+        'ifInUcastPkts',
+        'ifOutUcastPkts',
+        'ifInNUcastPkts',
+        'ifOutNUcastPkts',
+        'ifHCInMulticastPkts',
+        'ifHCInBroadcastPkts',
+        'ifHCOutMulticastPkts',
+        'ifHCOutBroadcastPkts',
+        'ifInOctets',
+        'ifOutOctets',
+        'ifHCInOctets',
+        'ifHCOutOctets',
+        'ifInDiscards',
+        'ifOutDiscards',
+        'ifInUnknownProtos',
+        'ifInBroadcastPkts',
+        'ifOutBroadcastPkts',
+        'ifInMulticastPkts',
+        'ifOutMulticastPkts',
+        );
 
 $stat_oids_db = array(
-    'ifInOctets',
-    'ifOutOctets',
-    'ifInErrors',
-    'ifOutErrors',
-    'ifInUcastPkts',
-    'ifOutUcastPkts',
-);
+        'ifInOctets',
+        'ifOutOctets',
+        'ifInErrors',
+        'ifOutErrors',
+        'ifInUcastPkts',
+        'ifOutUcastPkts',
+        );
 
 $stat_oids_db_extended = array(
-    'ifInNUcastPkts',
-    'ifOutNUcastPkts',
-    'ifInDiscards',
-    'ifOutDiscards',
-    'ifInUnknownProtos',
-    'ifInBroadcastPkts',
-    'ifOutBroadcastPkts',
-    'ifInMulticastPkts',
-    'ifOutMulticastPkts',
-);
+        'ifInNUcastPkts',
+        'ifOutNUcastPkts',
+        'ifInDiscards',
+        'ifOutDiscards',
+        'ifInUnknownProtos',
+        'ifInBroadcastPkts',
+        'ifOutBroadcastPkts',
+        'ifInMulticastPkts',
+        'ifOutMulticastPkts',
+        );
 
 $cisco_oids = array(
-    'locIfHardType',
-    'locIfInRunts',
-    'locIfInGiants',
-    'locIfInCRC',
-    'locIfInFrame',
-    'locIfInOverrun',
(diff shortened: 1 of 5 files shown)


theherodied:
@pipocanaja what should we look for when testing? I'm not familiar with what was graphed previously with POE ports.

PipoCanaja:
Hello

On a Poe Switch, either Cisco or Huawei, you should be able to have a new graph in the "port" page (URL like http://myserver.mydomain/device/device=1/tab=port/port=50/ ), after the Interface Traffic, Packets and Errors. 

To check if the polling works, you can go in the RRD directory of the device you would like to ckeck (usually /opt/librenms/rrd/<mydevice>), and do a "ls *poe*" to see how it goes. 

Bye

PipoCanaja:
Hello @laf 

Done. Except 2 or 3 \n, the changes are now only linked to PoE code.

Bye

includes/polling/ports.inc.php
```diff
@@ -174,6 +174,20 @@
     'dot3StatsDuplexStatus',
 ];
 
+$vrp_poe_oids = array(
```
laf: Where is this used?

includes/polling/ports.inc.php
```diff
@@ -274,8 +288,41 @@
 }//end if
 
 if ($config['enable_ports_poe']) {
-    $port_stats = snmpwalk_cache_oid($device, 'pethPsePortEntry', $port_stats, 'POWER-ETHERNET-MIB');
```
laf: With your new code you've now stopped this working for any none ios, procurve or vrp devices, how come?

includes/polling/ports.inc.php
```diff
@@ -274,8 +288,41 @@
...
+    // Code by OS device
+    if ($device['os'] == 'procurve') {
+        // Not tested
+        $port_stats_poe = snmpwalk_cache_oid($device, 'pethPsePortEntry', array(), 'POWER-ETHERNET-MIB');
+    }
+
+    if ($device['os'] == 'ios') {
+        $port_stats_poe = snmpwalk_cache_oid($device, 'pethPsePortEntry', array(), 'POWER-ETHERNET-MIB');
+        $port_stats_poe = snmpwalk_cache_oid($device, 'cpeExtPsePortEntry', $port_stats_poe, 'CISCO-POWER-ETHERNET-EXT-MIB');
+        $port_ent_to_if = snmpwalk_cache_oid($device, 'portIfIndex', array(), 'CISCO-STACK-MIB');
```
laf: This + the extra walks below is going to slow polling down and could do so considerably. What's the reason for performing these?

includes/polling/ports/port-poe.inc.php
```diff
@@ -2,60 +2,76 @@
...
-    data_update($device, 'poe', $tags, $fields);
-
-    echo 'PoE ';
-}//end if
+if (($device['os'] == 'vrp')) {
+    //Tested against Huawei 5720 access switches
+    if (isset($this_port['hwPoePortEnable'])) {
+        $rrd_name = getPortRrdName($port_id, 'poe');
+        $rrd_def = RrdDefinition::make()
+            ->addDataset('PortPwrAllocated', 'GAUGE', 0)
```
laf: You're duplicating this block + `data_update()` 3 times, would be better to move these outside of the if blocks.

includes/polling/ports/port-poe.inc.php
```diff
@@ -2,60 +2,76 @@
...
+                'PortPwrAvailable'   => $this_port['hwPoePortMaximumPower'],
+                'PortConsumption'    => $this_port['hwPoePortConsumingPower'],
+                'PortMaxPwrDrawn'    => $this_port['hwPoePortPeakPower'],
+                   );
+
+        $tags = compact('ifName', 'rrd_name', 'rrd_def');
+        data_update($device, 'poe', $tags, $fields);
+        echo 'PoE(vrp) ';
+    }
+} elseif (($device['os'] == 'ios')||($device['os'] == 'ciscosb')) {
```
laf: You can do `if (str_contains($device['os'], ['ios', 'ciscosb'])) {` instead.

includes/polling/ports/port-poe.inc.php
```diff
@@ -2,60 +2,76 @@
...
+    // Code for Cisco IOS, tested on 2960X
+    if (isset($this_port['cpeExtPsePortPwrAllocated'])) {
```
laf: Where is `$this_port['cpeExtPsePortPwrAllocated']` coming from as it's not in the code anywhere before this?

includes/polling/ports/port-poe.inc.php
```diff
@@ -2,60 +2,76 @@
...
-
-    $tags = compact('ifName', 'rrd_name', 'rrd_def');
-    data_update($device, 'poe', $tags, $fields);
-
-    echo 'PoE ';
-}//end if
+if (($device['os'] == 'vrp')) {
+    //Tested against Huawei 5720 access switches
+    if (isset($this_port['hwPoePortEnable'])) {
+        $rrd_name = getPortRrdName($port_id, 'poe');
```
laf: This is also being duplicated

includes/polling/ports.inc.php
```diff
@@ -274,8 +288,41 @@
...
+        foreach ($port_ent_to_if as $p_index => $p_stats) {
+            if ($p_stats[portIfIndex] > 0) {
```
laf: Can you update to use `'portifIndex'` + any other indexes in arrays which don't contain `''`

mibs/huawei/HUAWEI-POE-MIB.mib
```diff
@@ -0,0 +1,720 @@
+
```
laf: Can you rename this to drop the `.mib` please.

includes/polling/ports/port-poe.inc.php
PipoCanaja: $this_port is defined in line 491 of includes/polling/ports.inc.php as &port_stats[$ifIndex].

includes/polling/ports/port-poe.inc.php
PipoCanaja: removed ciscosb for now as I cannot test it. ciscosb will fall in default value until somebody has a test bench for it.

includes/polling/ports.inc.php
PipoCanaja: On the way.

mibs/huawei/HUAWEI-POE-MIB.mib
PipoCanaja: OK

includes/polling/ports.inc.php
PipoCanaja: 'pethPsePortEntry' is now removed as it is not necessary for Cisco polling. cpeExtPsePortEntry and portIfIndex are mandatory : 
cpeExtPsePortEntry contains the actual value, but indexed by entIndex, and portIfIndex contains the mapping to change entIndex into ifIndex.
Then we have a foreach loop to do the actual mapping and get the usable values.

includes/polling/ports/port-poe.inc.php
PipoCanaja: done

includes/polling/ports.inc.php
PipoCanaja: The existing code was not working (POWER-ETHERNET-MIB is polled but never used in the code, and CISCO-POWER-ETHERNET-EXT-MIB is CISCO specific, but in case it was used by somebody, I restored the original code in the final 'else' statement now.

includes/polling/ports.inc.php
```diff
@@ -274,8 +274,43 @@
...
+            }
+        }
+    } elseif  ($device['os'] == 'vrp') {
+        echo 'HwPoePortEntry' ;
+        $port_stats = snmpwalk_cache_oid($device, 'HwPoePortEntry', $port_stats, 'HUAWEI-POE-MIB');
+    } else {
+        //Any other device, generic polling
+        $port_stats = snmpwalk_cache_oid($device, 'pethPsePortEntry', $port_stats, 'POWER-ETHERNET-MIB');
+        $port_stats = snmpwalk_cache_oid($device, 'cpeExtPsePortEntry', $port_stats, 'CISCO-POWER-ETHERNET-EXT-MIB');
+    }
```
PipoCanaja: Legacy polling in the 'else' statement.

PipoCanaja:
Looking at the test data, will come back with this later.

PipoCanaja:
Test data for huawei device with PoE antenna plugged in. You already have a bunch of Cisco devices but not 2960X so I'll try to capture the data and clean it to submit it as well.

PipoCanaja:
And done with Cisco 2960X

csalcedo1024:
Yup, not working for me either on a cisco 3560..

PR #8707, opened 2018-05-13
docs: Update Test-Units.md

mattie47:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Developing/os/Test-Units.md
+++ b/doc/Developing/os/Test-Units.md
@@ -33,7 +33,7 @@ one variant, please do not specify one.
 
 ## Running tests
 
-**Note:** To run tests, ensure you have executed "./scripts/composer_wrapper.php install" from your LibreNMS root directory. This will read composer.json and install any dependencies required. 
+**Note:** To run tests, ensure you have executed `./scripts/composer_wrapper.php install` from your LibreNMS root directory. This will read composer.json and install any dependencies required. 
 
 After you have saved your test data, you should run `./scripts/pre-commit.php -p -u` verify they pass.
 
--- a/doc/Developing/os/Test-Units.md
+++ b/doc/Developing/os/Test-Units.md
@@ -32,9 +32,8 @@ specify a different variant of the os, this will be tested completely separate f
 one variant, please do not specify one.
 
 ## Running tests
-**Note:** To run tests, ensure you have:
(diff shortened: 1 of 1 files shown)


murrant:
Seems to be a conflict there.  one line says to install the system wide composer, the other executes ./composer.phar which is downloaded by the `./scripts/composer_wrapper.php` script.

laf:
I'd prefer to get rid of suggesting to install composer by yum/apt.

mattie47:
>Seems to be a conflict there. one line says to install the system wide composer, the other executes ./composer.phar which is downloaded by the ./scripts/composer_wrapper.php script.

OK so I've gone off and done some googling...

In short:
- composer is a package manager for php, similar to npm for node. It differs from things such as apt, or YUM in that packages can be installed on a per project basis. 
- `apt install composer` would globally install composer, which isn't necessarily needed...because: 
-- we can install `composer.phar` in the project folder instead. This is the binary file for per project managed dependencies. 

**In terms of LibreNMS**
- ./composer.json contains the list of requirements/dependencies for LibreNMS. 
- The reason we don't necessarily need composer installed globally, is because we have ./scripts/composer-wrapper.php
- Composer-wrapper will attempt to download composer.phar into the libreNMS install directory 
- If Composer-wrapper can't download composer.phar, it will attempt to use a globally installed composer (if it is installed).
- If all else fails, it'll tell the user to install composer. 

What I don't understand is how composer-wrapper gets called in the first place? I thought perhaps through daily.sh, but this just seems to reference an already defined ${COMPOSER}... 

I see from the LibreNMS install instructions one of the required packages **is** composer, which then uses `composer --create-project` to initially set the libreNMS project up. But beyond that, I'm not sure...

laf:
Forget what composer is or isn't outside of LibreNMS. The composer_wrapper script handles installing composer into /opt/librenms/ as composer.phar so we avoid older versions installed by the distro in the bin path, they've caused us problems before.

Once composer is available as /opt/librenms/composer.phar then composer_wrapper handles invoking it to install the dependancies we need.

mattie47:
@laf,

OK - What do you actually want the docs here to say? 

`./scripts/composer_wrapper.php install` ?

laf:
Correct.

PR #8719, opened 2018-05-15
Feature: Adding VRP support for FDB table using HUAWEI-L2MAM-MIB

PipoCanaja:
Hello

Here is the code to load the FDB table using the Huawei MIB. Because the switch does not answer to Q-BRIDGE at all, and BRIDGE MIB does not provide the vlan information.

The test data (if necessary) will come after the other PR (POE polling) is merged, to have  a test file including all.

Bye

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/discovery/fdb-table/vrp.inc.php
```diff
@@ -0,0 +1,47 @@
+<?php
+/**
+ * vrp.inc.php
+ *
+ * Discover FDB data with HUAWEI-L2MAM-MIB
```
laf: Could you update your copyright to use this template please:

https://docs.librenms.org/#Developing/Licensing/#copyright

includes/discovery/fdb-table/vrp.inc.php
```diff
@@ -0,0 +1,47 @@
...
+                continue;
+            }
+            $vlan_id = isset($vlans_dict[$vlan]) ? $vlans_dict[$vlan] : 0;
+            $insert[$vlan_id][$mac_address]['port_id'] = $port_id;
+            d_echo("vlan $vlan mac $mac_address port ($ifIndex) $port_id\n");
+        }
+    }
+}
+
+echo PHP_EOL;
```
laf: Any reason for this? Probably better to add it to `echo 'HUAWEI-L2MAM-MIB:';` instead.

includes/discovery/fdb-table/vrp.inc.php
PipoCanaja: Done

includes/discovery/fdb-table/vrp.inc.php
PipoCanaja: Correct done. Validation ongoing.

murrant:
Your test data files seem to be incomplete.
When I was attempting to generate the test data, the fdb data was not mapping to a port.

murrant:
We don't support tests for the fdb-table module yet, added in #8723

PipoCanaja:
@murrant OK.

How should I proceed then ? I wait until #8723 is merged ?

PipoCanaja

PipoCanaja:
Hi @murrant and @laf 

Just a check to understand what's going on: the basic pre-commit test works but the advanced one (with test data) does not. When trying it on my machine, I get errors that do not seem to be related directly to this patch.So, the changes that are going on with the tests are impacting this merge ? 
If yes, could you let me know if there is something else I can do ? 
Thanx
PipoCanaja

murrant:
@PipoCanaja It is a problem with the sql query to capture the data.  I need to update it but haven't had any type yet.

PR #8721, opened 2018-05-15
FR 721 - Juniper SRX (Branch) SPU CPU, Memory and Flow Session Graphs

centralscrutiniser:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

- Added SPU CPU, Memory and Flow Session Graphs


includes/polling/os/junos.inc.php
```diff
@@ -1,5 +1,52 @@
 <?php
 
+use LibreNMS\RRD\RrdDefinition;
+
+$oid_list = 'jnxJsSPUMonitoringCurrentFlowSession.0 jnxJsSPUMonitoringCPUUsage.0 jnxJsSPUMonitoringMemoryUsage.0';
```
laf: Can we not get this data into the actual mempool and processors module?

centralscrutiniser: @laf  not sure I understand your request, can you explain?

laf: So we already have discovery and poller modules that deal with memory and processors so adding them here is counter to that + you can't alert on this data. So, if it's possible to get data from snmp that can be used in those modules it should be done there.

laf:
You've removed some config for bluecoat devices here + have removed the cpu and memory stuff.

If it's not possible to get it in the poller modules as I asked earlier then this PR was fine.

centralscrutiniser:
Hi @laf 

The Blue Coat code shouldn't have been in this PR. My apologies. 

The CPU (SPU) is actually already graphed by the existing JunOS processor code. Memory isn't, but as it's a pretty static graph, I've decided to remove it. 

The flow sessions are the interesting part, which is still included. 

Hope that makes sense.

laf:
@centralscrutiniser So you need to re-add the things you've removed and also put your schema changes in to a new file, you can't edit existing ones. Use `252.sql`

PR #8730, opened 2018-05-16
API: System endpoint, more health graphs

pheinrichs:
This PR adds memory, processor, disk usage to the /health endpoint.

At this time, no id specific info will be displayed about those graphs, only stacked graphs can be displayed for these.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


html/includes/api_functions.inc.php
```diff
@@ -825,6 +825,30 @@ function list_available_health_graphs()
                 'name' => 'device_'.$graph['sensor_class'],
             );
         }
+
+        $processor = dbFetchRows('SELECT * FROM `processors` WHERE device_id = ?', array($device_id));
```
murrant: Shouldn't these be count queries? `SELECT COUNT(*) ...`

murrant: That would be fine by me, all this needs to be converted to Eloquent eventually.  Eloquent is safe to use in API and Web UI.  It is disable in poller/discovery for now until we can send a notification for uses to increase their mysql max connections.

PR #8733, opened 2018-05-17
Updated installation instructions

hanserasmus:
Added a few steps I think might be useful to first time Libre users.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Installation/Installation-CentOS-7-Apache.md
+++ b/doc/Installation/Installation-CentOS-7-Apache.md
@@ -108,7 +108,7 @@ Install the policy tool for SELinux:
     setsebool -P httpd_can_sendmail=1
 
 ##### Allow fping
-Create the file http_fping.tt with the following contents:
+Create the file http_fping.tt with the following contents. You can create this file anywhere, as it is a throw-away file. The last step in this install procedure will install the module in the proper location.
 ```
 module http_fping 1.0;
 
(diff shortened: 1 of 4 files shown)


doc/Installation/Installation-CentOS-7-Apache.md
```diff
@@ -108,7 +108,7 @@ Install the policy tool for SELinux:
     setsebool -P httpd_can_sendmail=1
 
 ##### Allow fping
-Create the file http_fping.tt with the following contents:
+Create the file http_fping.tt with the following contents. You can create this file anywhere, as it is a throw-away file. The last step in this install procedure will install the module in the proper location.
```
murrant: This needs to be copied to the nginx install docs too.

doc/Installation/Installation-CentOS-7-Apache.md
```diff
@@ -138,6 +138,12 @@ Then run these commands
 
 #### Configure snmpd
 
+Create a backup of your old snmpd.conf that shipped with your CentOS install.
```
murrant: Not sure this is needed, if users want the default config back, they can just reinstall snmp.

doc/Installation/Installation-CentOS-7-Apache.md
```diff
@@ -171,6 +177,11 @@ Now head to the web installer and follow the on-screen instructions.
 
     http://librenms.example.com/install.php
 
+The web installer will prompt you to create a `config.php` file in your librenms install location. Please remember to set the permissions on config.php after you copied the on-screen contents to the file. Run:
```
murrant: This is definitely not needed. It is ok for the config.php file to have different permissions, as long as the web server can read it, and it definitely can since it wrote it.

Sure, validate.php will complain about it, but that is minor.

murrant: Yeah, web server has trouble making that file because of permissions and we can't fix the permissions for that folder because it would mess permissions.
https://github.com/ConsoleTVs/Charts

Issue #8732, opened 2018-05-16
New Device: Cambium CMM3


inkweb:
**Please do not post bugs or ask questions**

**For help and support you can use our [Discord server](https://t.libren.ms/discord) or our [community site](https://community.librenms.org).**

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Please provide **ALL** info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

    ====================================
    Component | Version
    --------- | -------
    LibreNMS  | 1.39-65-g0448c9e
    DB Schema | 250
    PHP       | 7.0.30-0ubuntu0.16.04.1
    MySQL     | 10.1.24-MariaDB-1~precise
    RRDTool   | 1.5.5
    SNMP      | NET-SNMP 5.7.3
    ====================================

Discovery: https://p.libren.ms/view/f19e3b76
Poller: https://p.libren.ms/view/bc96850d
SNMPWalk: https://p.libren.ms/view/89c191f3
MIB: http://support.cambiumnetworks.com/framed/onlinetools/content.CMM3-MIB.html

This device is essentially a switch with with 8 ports and POE injector. The current LibreNMS is detecting and graphing the main 10/100 port, but not the separate switch ports. It IS graphing the switch ports power states.

cchance27:
Ummm the cmm3 is not a switch it’s a pass through power and sync device so theirs no traffic to monitor nor are those ports exposed via snmp as they aren’t switch ports

inkweb:
I've got per-port bandwidth usage working fine in Cacti (which i'm trying to replace with LibreNMS). So the data must be there someplace.

cchance27:
From the etherwan or from the cmm itself? Could have sworn the internal ports are passthrough for cmm3/4/5

inkweb:
Ya, each of our 25 CMM3's has 8 individual bandwidth graphs.

laf:
Fixed in #8737

Issue #8731, opened 2018-05-16
New Device: Cambium CMM4


inkweb:
**Please do not post bugs or ask questions**

**For help and support you can use our [Discord server](https://t.libren.ms/discord) or our [community site](https://community.librenms.org).**

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Please provide **ALL** info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

    ====================================
    Component | Version
    --------- | -------
    LibreNMS  | 1.39-65-g0448c9e
    DB Schema | 250
    PHP       | 7.0.30-0ubuntu0.16.04.1
    MySQL     | 10.1.24-MariaDB-1~precise
    RRDTool   | 1.5.5
    SNMP      | NET-SNMP 5.7.3
    ====================================

Discovery: https://p.libren.ms/view/48a86f72
Poller: https://p.libren.ms/view/b18394ca
SNMPWalk: https://p.libren.ms/view/7ea3383c
MIB: http://support.cambiumnetworks.com/framed/onlinetools/content.CMM4-MIB.html

This device is essentially a switch with with 8 ports and POE injector. The current LibreNMS is detecting and graphing the main 10/100 port, but not the separate switch ports, nor their power state.

pheinrichs:
@inkweb Could you let me know what the output of this OID is?
.1.3.6.1.4.1.161.19.3.6.3.2.0

inkweb:
.1.3.6.1.4.1.161.19.3.6.3.2.0 = STRING: "CMM4 0a-00-3e-e8-12-20"

cchance27:
Cmm4 is a power and gps injection device those ports are passthrough ports not switch ports and not exposed packet info as their is no packet info or throughout because the device doesn’t touch data. All you can get from this device is the gps and power states

inkweb:
unfortunately i do not have a very good template for cacti for the CMM4, so its graphing very little. So you may be correct on this one. however it seems odd that the CMM3 would have per-port data, but the newer CMM4 would not.

cchance27:
Neither has one

The cmm4 ships with a cmm4 and an etherwan switch sometimes

[...]

PR #8737, opened 2018-05-17
Improved support for cambium cmm and added cmm4

pheinrichs:
- Added port discovery for cmm3 (not available for cmm4, not that i could find)
- Added discovery for cmm4
- Added Power port cmm4
- Removed Processor bug by disabling processor on device.

I am unable to test CMM4 and may need someone to do so or review for possible errors.

Fixes Issues #8732, #8731

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


tests/data/cmm4.json
```diff
@@ -0,0 +1,40 @@
+{
+    "os": {
+        "discovery": {
+            "devices": [
+                {
+                    "sysName": "",
+                    "sysObjectID": ".0.0.0.0.0.0.0.0.0.0",
```
laf: Is this genuinely the sysObjectID returned from SNMP?

murrant:
I'm curious about the port pollers, looking at the json files a lot of data is missing.
cmm4 seems to not supply if-mib data like ifOperStatus and ifSpeed, are these available at the normal IF-MIB OIDs or just this device provides bad data?
cmm seem as though the port poller is totally disabled because it is just an empty file.

pheinrichs:
@murrant I really can't confirm much on the cmm4. I don't have a device to confirm / test this on. 

After some more testing, the port implementation on cmm3 doesn't seem to be working quite right yet, Should merging an array to the port_stats var work? It seems as though the poller.php is attempting to insert into the ports_statistics table with a port_id of an empty string. Any idea what might be needed? 

If i purge the port_stats array entirely and just replace it with the newly created array that seems to work fine, just merging seems to cause issues.

laf:
Don't think you can use array_merge() on multi dimensional arrays can you?

murrant:
No, you would need to use something like array_replace_recursive()

pheinrichs:
I think this PR is finally okay? After a long time coming ;)

PR #8741, opened 2018-05-19
Fix Mimosa PtP frequency and power

murrant:
Add test data
There were also some sensor name collisions and an incorrect divisor.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Data is not working currently or is incorrect previously as far as I can tell.

PR #8748, opened 2018-05-21
Add Discord transport

theherodied:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Needs the webui + docs doing as well :p Thanks for adding this though

theherodied:
From sdef2

Discord server returns something like "empty response" http code Instead of http status 200
Don't remember which code exactly, but you need to modify the script accordingly else librenms thinks the alert failed
Line 55

theherodied:
@laf not sure what to do about the sql. I'm not entirely sure what's needed but tried it anyway. Even with the new sql it's giving me "Missing config name or value"

LibreNMS/Alert/Transport/Discord.php
```diff
@@ -0,0 +1,64 @@
+<?php
+/**
+ * Discord.php
+ *
+ * LibreNMS storage discovery module for Nimble Storage
```
f0o: wat?

theherodied: Copy and Paste error.

laf:
More code needed in `html/includes/forms/config-item.inc.php`.

Also, the changes to `html/pages/settings/alerting.inc.php` have muddled up. You've got slack/discord and rocket code mixed up.

laf:
@theherodied I've fixed the issue with the html page in the settings.

Tested and works well.

Just needs docs doing now.

theherodied:
@laf, I can't seem to get it working on mine. Not sure what's up with it.
I added the webhook url and hit test but nothing shows up. The "Test transport" button turns red then back to blue.

theherodied:
@laf , got it working. Just the "Test transport" button doesn't work.

Only thing I've gotten working so far is the basic webhook url. Not sure if I can get the channel, username, etc... working or if it's needed. Webhook is per channel so that might not matter.

laf:
> Only thing I've gotten working so far is the basic webhook url. Not sure if I can get the channel, username, etc... working or if it's needed. Webhook is per channel so that might not matter.

Do you mean message a specific user? I've not looked at the Discord api docs but channel works ok so I'm happy with that. If you can sort the docs we can get this merged in

theherodied:
https://github.com/librenms/librenms/pull/8795

theherodied:
@laf I'm good with this for now as well. It has been working well the past few days I've tested it.

Added docs and alphabetized the list.

PR #8702, opened 2018-05-12
Use Laravel authentication

murrant:
Use Laravel for authentication, but call into legacy auth code for now.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply 8702` and run ./build-base.php to update your db schema.

TODO:

- [x] Implement token auth for API
- [x] Remember Me (updated and tested)
- [x] Two Factor Auth
- [x] Unauth status screen
- [x] Remove legacy includes
- [x] API check auth
- [x] Remove unneeded legacy code
- [ ] Test each auth type (simple tests done for mysql, ldap, and active directory)
- [x] General Testing

Auth types tested:
 - [x] Mysql
 - [x] Active Directory
 - [x] LDAP
 - [x] Radius
 - [x] HTTP
 - [ ] HTTP-LDAP
 - [ ] HTTP-AD
 - [x] SSO

What to test:

 - login/logout
 - user management (if applicable)
 - api token creation
 - api access / full usage
 - other auth related things
 - 2FA

PR #8750, opened 2018-05-22
WIP - API v1 concept

pheinrichs:
I wanted to give this a go. I started this with the thought process of potentially using http://apidocjs.com for our new documentation on our API. 

Checklist of things to implement
- [x] - Route model binding for device ID or Hostname
- Get endpoints up to par
  - [x] Devices ( Waiting for #8852 )
  - [x] Ports (todo: use `previous` handle)
  - [x] Bills
  - [x] Address (todo: `arp routes`)
  - [ ] Graphs
  - [x] Groups
  - [x] Alerts
  - [x] Routing `add ip routes`
  - [x] Inventory
  - [x] Services
  - [x] Logs

- [ ]  - Auth middleware ( Waiting for #8702 )
- [ ] - Permissions ( Waiting for Auth)
- [ ] - Model Factories ( pull in faker package) https://github.com/librenms/librenmsv2/blob/develop/database/factories/ModelFactory.php
- [ ] - Api Tests


murrant:
Conceptual question. Are you thinking this a re-implementation of v0 or a restructuring?

laf:
> But overall I would love to improve the URL structure.

This is definitely needed. The current on became a bit of an unwieldy mess in the end.

app/Http/Controllers/Api/ApiController.php
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * Return an api response with pagination
+     *
+     * @param \Illuminate\Database\Eloquent\Collection $data
+     *
+     * @return \Illuminate\Http\JsonResponse
+     */
+    protected function paginateResponse($data)
+    {
+        $per_page = (request()->per_page ? request()->per_page : 15);
```
laf: Definitely think 50 should be the default maybe?

app/Http/Controllers/Api/Device/DeviceController.php
```diff
@@ -0,0 +1,319 @@
...
+        $snmpver = 'v2c';
+        $hostname     = $request->hostname;
+        $port         = (isset($request->port) ? $request->port : Config::get('snmp.port'));
+        $transport    = (isset($request->transport) ? $request->transport : 'udp');
+        $poller_group = (isset($request->poller_group) ? $request->poller_group : 0);
+        $force_add    = ($request->force_add ? true : false);
+        $snmp_disable = ($request->snmp_disable);
+        if ($snmp_disable) {
+            $additional = array(
+                'os'           => $request->os ? mres($request->os) : 'ping',
```
laf: If this is coming through laravel now then we don't need `mres()` do we?

app/Http/Controllers/Api/Device/DeviceGraphController.php
```diff
@@ -0,0 +1,67 @@
...
+namespace App\Http\Controllers\Api\Device;
+
+use App\Models\Device;
+use Illuminate\Http\Request;
+use App\Http\Controllers\Api\ApiController;
+
+class DeviceGraphController extends ApiController
+{
+    /**
+     * @api {get} /devices/:id/graphs Get Generic graphs
```
laf: This indicates it gets graphs. Maybe `/devices/:id/graphs/available`?

app/Http/Controllers/Api/ApiController.php
pheinrichs: Sounds good, i'll update accordingly

app/Http/Controllers/Api/Device/DeviceHealthController.php
```diff
@@ -0,0 +1,81 @@
...
+namespace App\Http\Controllers\Api\Device;
+
+use App\Models\Device;
+use Illuminate\Http\Request;
+use App\Http\Controllers\Api\ApiController;
+
+class DeviceHealthController extends ApiController
+{
+    /**
+     * @api {get} /devices/:id/health Get Health Sensors
```
laf: Same here, `/devices/:id/health/available`?

app/Http/Controllers/Api/Device/DeviceController.php
pheinrichs: Yep, you're right.

app/Http/Controllers/Api/Device/DeviceHealthController.php
```diff
@@ -0,0 +1,81 @@
...
+     *     }
+     *
+     */
+    public function index(Device $device)
+    {
+        return $this->objectResponse($device->sensors()->deviceSensors());
+    }
+
+    /**
+     * @api {get} /devices/:id/health/:class Get all sensors for a Health Class
```
laf: Ignore my last comment as this would clash, however I think we do need an alternative route for getting graphs available to getting graphs.

pheinrichs: I was thinking all graph routes would be routed through a base route of /api/v1/graphs/....
Thoughts?

laf:
Started commenting across all routes but it's probably easier to do it here.

We need to split the routes that produce data and produce graphs. I'm not sure the best approach but:

`/device/:id/data/health`
`/device/:id/graph/health`

Or something similar?

pheinrichs:
@laf I was thinking all graph routes would be routed through a base route of /api/v1/graphs/....
The reason being, this would leave the room for other graphs such as bills and further if needed while still maintaining uniformity.
```
/api/v1/graphs/devices/1/health
/api/v1/graphs/devices/1/wireless
/api/v1/graphs/bills/1
```
Thoughts?

app/Http/Controllers/Api/Device/DeviceGraphController.php
pheinrichs: This could probably be renamed to generic or something?

laf:
> Thoughts?

Zero, that looks good though :)

app/Http/Controllers/Api/Device/DeviceController.php
```diff
@@ -0,0 +1,336 @@
...
+            'v3.authlevel' => 'required_if:version,v3',
+            'v3.authname' => 'required_if:version,v3',
+            'v3.authpass' => 'required_if:version,v3',
+            'v3.authalgo' => 'required_if:version,v3',
+            'v3.cryptopass' => 'required_if:version,v3',
+            'v3.cryptoalgo' => 'required_if:version,v3',
+        ]);
+        $snmpver = 'v2c';
+        $hostname     = $request->hostname;
+        $port         = (isset($request->port) ? $request->port : Config::get('snmp.port'));
```
murrant: @pheinrichs FYI, you can simplify to this `$request->get('port', Config::get('snmp.port', 161))` (and add the default as 161) Same goes for anything else, were you need to supply a default, like snmpver

murrant:
Looking good so far @pheinrichs just a tip about pulling variables from $request when you need a default.

And I'm going to leave this here:  https://github.com/librenms/librenmsv2/blob/develop/database/factories/ModelFactory.php

app/Http/Controllers/Api/Device/DeviceController.php
pheinrichs: Oh perfect, thanks!

pheinrichs:
@laf @murrant How would we feel about versioning with the api docs? It's fairly straight forward to do, i'm just not sure how the versioning would work in tandem with future librenms updates.

Ex: a user updates the an api endpoint with new parameters. Does the API version increment along with librenms' current version, or would the user just update the latest patch number to the next available and be isolated?

laf:
> Ex: a user updates the an api endpoint with new parameters. Does the API version increment along with librenms' current version, or would the user just update the latest patch number to the next available and be isolated?

I think we only need to update the version under two scenarios:

Minor version increment, i.e: 1.1 if we release a breaking change. We could also do this on new endpoints being added so that the docs people have match what they expect to see.

Major version increment, i.e 2.0 if we refactor substantially. I.e this PR goes from 0.0 to 1.0.

laf:
So what I just said then :p

app/Http/Controllers/Api/RuleController.php
```diff
@@ -0,0 +1,229 @@
...
+            'interval' => $request->get('interval', 300)
+        ];
+
+        $rule = Rule::create([
+            'name'      => $request->name,
+            'disabled'  => $request->get('disabled', false),
+            'severity'  => $request->severity,
+            'rule'      => $request->rule,
+            'extra'     => json_encode($extra),
+            'query'     => '' // TODO: Actually generate query?
```
pheinrichs: @laf Where / wow is this generated?

laf: We auto generate it from the 'builder' column. See `html/includes/forms/alert-rules.inc.php`

app/Helpers/snmp.php
```diff
@@ -0,0 +1,72 @@
...
+ * @param array $device the we will be connecting to
+ * @param string $oids the oids to fetch, separated by spaces
+ * @param string $options extra snmp command options, usually this is output options
+ * @param string $mib an additional mib to add to this command
+ * @param string $mibdir a mib directory to search for mibs, usually prepended with +
+ * @return string the fully assembled command, ready to run
+ */
+function gen_snmp_cmd($cmd, $device, $oids, $options = null, $mib = null, $mibdir = null)
+{
+    global $debug;
```
pheinrichs: @murrant I'm trying to think of a good way to migrate these global variables. Do you have any thoughts?

laf:
@pheinrichs Let us know when you need testers on this one?

pheinrichs:
Still a ways off, unless we want to release this in beta. The things we need are the eloquent to be added for easier clean devices add / updated.

Secondly the graph endpoints need to be added and unsure if we want to add them via legacy or to re write them with the Rrdxport option

laf:
> Still a ways off, unless we want to release this in beta. The things we need are the eloquent to be added for easier clean devices add / updated.

Ok just checking in that was all.

> Secondly the graph endpoints need to be added and unsure if we want to add them via legacy or to re write them with the Rrdxport option

I think it would still need to support generating graphs directly + also rrdxport option.

cchance27:
Was wondering is this PR going to make it possible to pull graph/graph lists for the Interface Parsing groups? Like the transit/core/etc ... To my knowledge the current API is missing that ability, to be able to grab a list of all say Transit ports, and then be able to grab those transit graphs from the api.

laf:
> Was wondering is this PR going to make it possible to pull graph/graph lists for the Interface Parsing groups? Like the transit/core/etc ... To my knowledge the current API is missing that ability, to be able to grab a list of all say Transit ports, and then be able to grab those transit graphs from the api.

https://docs.librenms.org/API/PortGroups/#get_graph_by_portgroup

murrant:
Closing for now, will need to be redone.

PR #8751, opened 2018-05-22
Bug: Typo in new port/plugin.inc.php and Copyright cleaning

PipoCanaja:
Hello

My bad. Should have seen that previously but the commits were done on a clean git tree instead of the test machine itself (cause it was dirty, still learning git) and the file did not come as good as it should have. 

Here is a better one. 

Bye
PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/device/port/plugins.inc.php
+++ b/html/pages/device/port/plugins.inc.php
@@ -1,16 +1,24 @@
 <?php
 /*
- * LibreNMS
- *
  * This program is free software: you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by the
  * Free Software Foundation, either version 3 of the License, or (at your
  * option) any later version.  Please see LICENSE.txt at the top level of
  * the source code distribution for details.
+
+ * @package    LibreNMS
+ * @subpackage webui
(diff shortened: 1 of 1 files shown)


html/pages/device/port/plugins.inc.php
```diff
@@ -1,16 +1,24 @@
...
+
 $pagetitle[] = 'Plugins';
+$no_refresh = true;
 ?>
+
 <h3>Plugins</h3>
 <hr>
 <?php
-    LibreNMS\Plugins::call('port_container', $device, $port);
+\LibreNMS\Plugins::call('port_container', array($device, $port));
```
murrant: Can you use new style array.

html/pages/device/port/plugins.inc.php
```diff
@@ -1,16 +1,24 @@
...
+?>
```
murrant: And don't put `?>` at the end of files.  It will cause issues.

PipoCanaja:
Hi @murrant 

Should be better now, checks ongoing. 

PipoCanaja

PR #8757, opened 2018-05-23
Poll service check only if the associated device is available

dsgagi:
Poll service check only if the associated device is available.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/check-services.php
+++ b/check-services.php
@@ -61,14 +61,20 @@ if ($options['h']) {
     }
 }
 
-$sql = 'SELECT * FROM `devices` AS D, `services` AS S WHERE S.device_id = D.device_id ' . $where . ' ORDER by D.device_id DESC';
+$sql = 'SELECT * FROM `devices` AS D INNER JOIN `services` AS S ON S.device_id = D.device_id ' . $where .
+' LEFT JOIN `devices_attribs` as A ON S.device_id = A.device_id AND A.attrib_type = "override_icmp_disable"
+ ORDER by D.device_id DESC;';
 
 foreach (dbFetchRows($sql) as $service) {
-    // Run the polling function only if the associated device is up
-    if ($service['status'] === "1" || ($service['status'] === '0' && $service['status_reason'] === 'snmp')) {
+    // Run the polling function if the associated device is up, "Disable ICMP Test" option is enabled, 
+    // or service hostname/ip is different from the associated device
+    if ($service['status'] === "1" || ($service['status'] === '0' && $service['status_reason'] === 'snmp') ||
+    $service['attrib_value'] === 'true' || ($service['service_ip'] !== $service['hostname'] &&
+    $service['service_ip'] !== $service['ip'] )) {
         poll_service($service);
     } else {
-        d_echo("\nNagios Service - ".$service['service_id']."\nSkipping service check because device ".$service['device_id']." is down.\n");
+        d_echo("\nNagios Service - ".$service['service_id']."\nSkipping service check because device "
+        .$service['hostname']." is down due to icmp.\n");
     }
 } //end foreach
 rrdtool_close();
--- a/check-services.php
+++ b/check-services.php
@@ -80,11 +80,11 @@ foreach (dbFetchRows($sql) as $service) {
         // Mark service check as enabled if it was disabled previosly because device was down
         if ($service['service_disabled'] === "1") {
             dbUpdate(
-                    array('service_disabled' => '0'),
-                    'services',
-                    '`service_id` = ?',
-                    array($service['service_id'])
-                    );
+                array('service_disabled' => '0'),
+                'services',
+                '`service_id` = ?',
+                array($service['service_id'])
+            );
         }
         poll_service($service);
         $polled_services++;
@@ -95,18 +95,19 @@ foreach (dbFetchRows($sql) as $service) {
         // but only if it's not already marked as disabled
         if ($service['service_disabled'] === "0") {
             dbUpdate(
-                    array('service_disabled' => '1'),
-                    'services',
-                    '`service_id` = ?',
-                    array($service['service_id'])
-                    );
-            log_event("Nagios Service - {$service['service_desc']} ({$service['service_id']}) - 
-                      Skipping service check because device {$service['hostname']} is down due to icmp",
-                      $device,
-                      'service',
-                      4,
-                      $service['service_id']
-                     );
+                array('service_disabled' => '1'),
+                'services',
+                '`service_id` = ?',
+                array($service['service_id'])
+            );
+            log_event(
+                "Nagios Service - {$service['service_desc']} ({$service['service_id']}) - 
+                Skipping service check because device {$service['hostname']} is down due to icmp",
+                $device,
+                'service',
+                4,
+                $service['service_id']
+            );
         }
     }
 } //end service foreach
--- a/check-services.php
+++ b/check-services.php
@@ -77,7 +77,7 @@ foreach (dbFetchRows($sql) as $service) {
     if ($service['status'] === "1" || ($service['status'] === '0' && $service['status_reason'] === 'snmp') ||
         $service['attrib_value'] === 'true' || ($service['service_ip'] !== $service['hostname'] &&
         $service['service_ip'] !== inet6_ntop($service['ip']) )) {
-        // Mark service check as enabled if it was disabled previosly because device was down
+        // Mark service check as enabled if it was disabled previously because device was down
         if ($service['service_disabled'] === "1") {
             dbUpdate(
                 array('service_disabled' => '0'),
--- a/check-services.php
+++ b/check-services.php
@@ -77,15 +77,19 @@ foreach (dbFetch($query) as $device) {
  ORDER by D.device_id DESC;';
 
     foreach (dbFetchRows($sql) as $service) {
-    // Run the polling function if the associated device is up, "Disable ICMP Test" option is not enabled, 
+    // Run the polling function if the associated device is up, "Disable ICMP Test" option is not enabled,
     // or service hostname/ip is different from associated device
         if ($service['status'] === "1" || ($service['status'] === '0' && $service['status_reason'] === 'snmp') ||
         $service['attrib_value'] === 'true' || ($service['service_ip'] !== $service['hostname'] &&
         $service['service_ip'] !== $service['ip'] )) {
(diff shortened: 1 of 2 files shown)


check-services.php
```diff
@@ -64,7 +64,11 @@
 $sql = 'SELECT * FROM `devices` AS D, `services` AS S WHERE S.device_id = D.device_id ' . $where . ' ORDER by D.device_id DESC';
 
 foreach (dbFetchRows($sql) as $service) {
-    // Run the polling function
-    poll_service($service);
+    // Run the polling function only if the associated device is up
+    if ($service['status'] === "1") {
+        poll_service($service);
+    } else {
+        d_echo("\nNagios Service - ".$service['service_id']."\nSkipping service check because device ".$service['device_id']." is down.\n");
```
laf: I actually think we should log to eventlog as well:

`log_event("Nagios Service - {$service['service_desc']} ({$service['service_id']}) - Skipping service check because device {$service['hostname']} is down", $device, 'service', 4, $service['service_id'])`

check-services.php
```diff
@@ -64,7 +64,11 @@
 $sql = 'SELECT * FROM `devices` AS D, `services` AS S WHERE S.device_id = D.device_id ' . $where . ' ORDER by D.device_id DESC';
 
 foreach (dbFetchRows($sql) as $service) {
-    // Run the polling function
-    poll_service($service);
+    // Run the polling function only if the associated device is up
+    if ($service['status'] === "1") {
```
laf: I actually think we should maybe do:

` if ($service['status'] === "1" || ($service['status'] === '0' && $service['status_reason'] === 'snmp')) {`

So that if the device is marked as down for snmp then it may be up for icmp still and service checks may still work. @librenms/reviewers Thoughts?

check-services.php
dsgagi: I think it would be too much spam in the event log in this case.

In my case, one device has 5 service checks and they are polled every 5 minutes. So eventlog would be populated with 5 new logs every 5 minutes while the device is down.

PipoCanaja:
Hello

Are you really sure you this behaviour is suitable ? In my case, we use LibreNMS to poll our external services (web, SMTP, webapps etc etc) and none of the public IP addresses are allowed by security to answer PING)
So the devices are down, but we expect the services to be polled. 
This change would definitly break this. 

How could we change this behaviour to be optional ? 

Thanx 
PipoCanaja

dsgagi:
@PipoCanaja 

Do you add service checks on the monitored device page or some other device?

In our case, we add service checks on the monitored device page (for the same ip address/hostname) so if the device is unavailable it wouldn't make sense to try polling the services (which would slow down service polling because of service checks timeouts).

With this change, only services added to the unavailable monitored device will be skipped.

PipoCanaja:
It is exactly our case. The monitored device does not answer to Ping (security does not allow it) but only to HTTP (if web) or the applicative service port. So the only polling we can do is on the port itself. So we need a way to disable this change if the community decides to make this change the default behaviour.

dsgagi:
New config switch sounds like a good idea to me.

If reviewers agree, I would even leave if off by default for compatibility.

murrant:
There is already a disable icmp test setting per device and we could check the nagios checks to see if they match the device they are attached to as well.

check-services.php
dsgagi: @laf 
Just checking if you still think that we should log to the eventlog too or maybe you just missed my reply? :)

dsgagi:
Should we also skip checking services associated with disabled devices?

Just figured out that there is no condition for that.

murrant:
No, I think if the device is disabled skip.  But, if icmp check is disabled, poll no matter what... I think that seems right.

laf:
service-wrapper would need to mimic the same I'd say.

check-services.php
```diff
@@ -61,10 +66,68 @@
...
+ INNER JOIN `services` AS S ON S.device_id = D.device_id AND D.device_id = '.$device['device_id'].'
+ LEFT JOIN `devices_attribs` as A ON D.device_id = A.device_id AND A.attrib_type = "override_icmp_disable"
+ ORDER by D.device_id DESC;';
+
+    foreach (dbFetchRows($sql) as $service) {
+    // Run the polling function if the associated device is up, "Disable ICMP Test" option is not enabled,
+    // or service hostname/ip is different from associated device
+        if ($service['status'] === "1" || ($service['status'] === '0' && $service['status_reason'] === 'snmp') ||
+        $service['attrib_value'] === 'true' || ($service['service_ip'] !== $service['hostname'] &&
+        $service['service_ip'] !== $service['ip'] )) {
```
laf: `$service['ip']` is the device IP here so you will need to run that through `inet6_ntop()` first.

check-services.php
```diff
@@ -61,10 +66,68 @@
...
+                    'service',
+                    4,
+                    $service['service_id']
+                );
+            }
+        }
+    } //end service foreach
+
+    echo "#### Start Alerts ####\n";
+    RunRules($device['device_id']);
```
laf: I personally don't think running alerting in here is a good thing.

check-services.php
```diff
@@ -61,10 +66,68 @@
     }
 }
 
-$sql = 'SELECT * FROM `devices` AS D, `services` AS S WHERE S.device_id = D.device_id ' . $where . ' ORDER by D.device_id DESC';
+if (!isset($query)) {
+    $query = "SELECT * FROM `devices` WHERE `disabled` = 0 $where ORDER BY `device_id` ASC";
+}
+
+foreach (dbFetch($query) as $device) {
```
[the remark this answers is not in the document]
dsgagi: I added device loop because of the RunRules function below, so it would execute the function after polling all services from one device before continuing on to the next device and it's services.

check-services.php
dsgagi: I actually added this as a test, because we had alerting problem I mentioned in comment before.

> I made more changes to the check-services.php logic and I think alerting runs more stable with them. Our instance is running for several days with the changes, and no errors like the ones mentioned below happened after implementing them.
> Could not issue critical alert for rule 'Memory CRITICAL' to transport 'mail' Error: You must provide at least one recipient email address. (Memory CRITICAL is a rule fired based on service check result)
> Before, we had the above error firing several times a day.

I was reviewing alerting and saw RunRules is called in poller.php, so I added it in check-services.php too for the test. Also, kinda made sense to me as poller.php and check-services.php were not working synchronized while polling.

I just reviewed the eventlog of our instance. Before I added RunRules to the code, the error above occurred tens of times on every day. After the change, error still occurred on some days in last month, but 1-4 times tops. It helped, but didn't solve the problem :).

Can't quite figure out what happens that triggers the error. I suspect that RunRules triggers at the same time as alerts.php, but clears/deletes the alert after alerts.php started sending mail transport and because of that it doesn't have any contact to send too. By adding RunRules to check-service.php I guess that situation happened less often.

Off course, it's up to the librenms team to decide. I'm primarily a network/system engineer, so not much programming experience

laf: Well if it's not even a fix then I think it needs to be removed so that the loop around two SQL queries can be dropped.

check-services.php
dsgagi: Done.

check-services.php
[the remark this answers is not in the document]
dsgagi: Reverted back to one loop.

dsgagi:
Any suggestions to make codeclimate happy maybe? :)

Not sure if the logical expression could be simpler than how it is now.

check-services.php
```diff
@@ -13,7 +13,7 @@
  * the source code distribution for details.
  */
 
-$init_modules = array();
+$init_modules = array('alerts');
```
laf: Need to remove 'alerts' now that this is no longer running alerting.

dsgagi: Forgot about that. Done :)

check-services.php
```diff
@@ -61,10 +66,59 @@
...
 foreach (dbFetchRows($sql) as $service) {
-    // Run the polling function
-    poll_service($service);
-} //end foreach
+    // Run the polling function if the associated device is up, "Disable ICMP Test" option is not enabled,
+    // or service hostname/ip is different from associated device
+    if ($service['status'] === "1" || ($service['status'] === '0' && $service['status_reason'] === 'snmp') ||
+        $service['attrib_value'] === 'true' || ($service['service_ip'] !== $service['hostname'] &&
+        $service['service_ip'] !== inet6_ntop($service['ip']) )) {
+        // Mark service check as enabled if it was disabled previosly because device was down
```
laf: Just a spelling mistake on previosly.

dsgagi: Agree about documentation. Will write something up during the weekend or on Monday, busy Friday.

I mark service as disabled in the `service_disabled` column in order to use it as a logical condition for writing to eventlog. It would write to eventlog only on first loop while the device is down.

check-services.php
```diff
@@ -61,10 +66,59 @@
...
+        // but only if it's not already marked as disabled
+        if ($service['service_disabled'] === "0") {
+            dbUpdate(
+                array('service_disabled' => '1'),
+                'services',
+                '`service_id` = ?',
+                array($service['service_id'])
+            );
+            log_event(
+                "Nagios Service - {$service['service_desc']} ({$service['service_id']}) -
```
murrant: AFAIK, they shouldn't be called "Nagios Service"
https://www.monitoring-plugins.org/

doc/Extensions/Services.md
```diff
@@ -89,3 +89,10 @@ then you can run the following command to help troubleshoot services.
 ```
 ./check-services.php -d
 ```
+## Service checks polling logic
+
+By default, service checks will be skipped if the associated device is not pingable, and an appropriate entry will be populated in the event log. An exception to this rule are service checks with `IP address` parameter different from the associated device's IP address.
```
murrant: You can drop the "By default, "  Maybe try to make it a little less wordy.

dsgagi: What do you think about below?

```Service check is skipped when the associated device is not pingable, and an appropriate entry is populated in the event log. 
Service check is polled if it's `IP address` parameter is not equal to associated device's IP address, even when the associated device is not pingable.```

Seems simpler like this, I think :)

dsgagi:
Sorry @laf. I saw murrant's suggestion first and made the change right away.

Noticed the approval only after I submitted the change :(.

PR #8763, opened 2018-05-25
Fix: Update pi-hole app for new API

slashdoom:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


slashdoom:
Agent changes:
https://github.com/librenms/librenms-agent/pull/178
Info on community forum:
https://community.librenms.org/t/pi-hole-stats-not-pushed-to-influxdb/4155/8

laf:
This will break for those people using the old api.

Our app stuff doesn't do versioning right now until it's moved to passing data via json. Someone else is working on that so I'm not sure that we can merge this right now.

slashdoom:
Unfortunately, the old agent doesn't have a key/value format.  It just spit out numbers and the poller used the order of those number to decide what they were.  So I could probably make these able to work for either API version but I probably can't make it work with both the old and new agent.  Not sure if that would be good enough?

laf:
I do think that it needs to maintain backwards compatibility otherwise we just break things for everyone. Maybe @VVelox could chip in here as he's refactoring the apps to json.

slashdoom:
Yeah, that would be great.  I've looked at some of the app to json PR's and it looks cool but doesn't seem like it is currently written to work with the app agent.  Just snmp extends because the snmpget is written into the json parser unless I'm missing something.

laf:
For us to merge this you'd have to send a notification out to people to say wha they will need to do make this work for them. I'd suggest create a thread on the community site then posting a notification: https://github.com/librenms/librenms.github.io/pull/102/files

murrant:
@slashdoom why not allow snmp extend?

slashdoom:
Hey @murrant, this was written to do both agent and snmp extend.  My point above was just that the new apps to json functions don't appear to be written to work with the agent.  So if I wanted to use whatever version control features are built into the apps to json functions I would either have to figure out how to rework them work with the agent data or give up agent support altogether.

murrant:
Extremely psuedo code, but yes, converting them.

if (snmp_extend_data) {
  standard_data_format = convert_extend(snmp_extend_data)
} else {
  standard_data_format = convert_agen(agent_data)
}

Usually the simplest is just to have a key->value for each item.

murrant:
You could definitely adapt the json_app_get to check for agent data.

VVelox:
I would be happy to begin looking at helping with this. Sorry, been a bit here and there lately.

VVelox:
@murrant @slashdoom I can't get github to see slashdooms repo for submitting a patch to in the drop down. It will show me other repos but not theirs.

Any ways if this gets commited, I can easily throw together some JSON stuff for this.

Just been looking this over a bit more tonight and it will be easy.

murrant:
@VVelox you can just create a separate PR with your changes for now.

kkrumm1:
Been a year now closing for now. Feel free to reopen

PR #8764, opened 2018-05-25
webui: display full name in alert[log] table WIP

PipoCanaja:
Hello

In this PR, I display the full name in the device list. In one of our LibreNMS instances, running on Internet to check our public services (websites, etc) from outside, we have all alerts with device "www" and it of course make no sense without the domain "www.xxxx.com" or "www.yyyy.com" ... 

Probably not as good for internal network device monitoring (most of the LibreNMS users are in this situation).

Would it make sense to define a config option to allow long names and keep short names by default ? 

WIP: This patch is not meant to be merged yet, but to discuss what would be best. 

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


Rosiak:
A config option certainly makes sense.

laf:
`shorthost_target_length` already exists for you to set the length you want to see.

The length is personal preference and removing it shouldn't be the way to tackle this. Thanks for the PR but I feel this is already possible without your change.

PipoCanaja:
Hi,

Thanx, missed this in the documentation. Discussion was the right direction ! I implemented it and got the display exactly as needed.

Bye
PipoCanaja

PR #8769, opened 2018-05-26
Change max load from 5000 to 50000

isarandi:
Perhaps a load value of 50000% seems excessive, however it can happen. (Example: a heavily loaded compute node with 48 CPU cores that produces loads of ~15000%, printed as 150 by tools like htop).

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
why set a max at all?

isarandi:
Honestly, I have no idea why there is even an option to set any collection limits for any monitored quantity. Values outside the range are simply thrown away, which perhaps makes sense if we suspect that the unreasonably high or low value is somehow due to misconfiguration and doesn't represent reality.

But as this example demonstrates, sometimes extreme values do occur.

In short, if it's possible in this place of the API to not set a max value, that would probably be a good idea.

isarandi:
Should I close this and submit a new pull request or can I somehow make an edit here (to remove the max value)?

murrant:
You can go to the files changed tab and click the edit button.

Or update your branch and commit it if using git on your computer.

murrant:
Won't affect existing files.

PR #8773, opened 2018-05-26
EfficientIP SOLIDserver Detections

centralscrutiniser:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

- Added initial detection for EfficientIP SOLIDserver


includes/definitions/efficientip.yaml
```diff
@@ -0,0 +1,13 @@
+os: efficientip
```
laf: efficientip is the company isn't it? It should be named after the OS that the devices run. If the OS doesn't have a name then use solidserver.

laf:
I've pushed the other test file up, once it passes this is good.

murrant:

https://community.librenms.org/t/efficientip-how-to-get-dns-stat/11523/1

PR #8783, opened 2018-05-28
 Add possibility to chose -H options

Cormoran96:
Add possibility to choose if you want librenms add for you -H options for check your services.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Add possibility to choose if you want librenms add for you -H options for check your services.

### DIFF
--- a/includes/services.inc.php
+++ b/includes/services.inc.php
@@ -129,13 +129,13 @@ function poll_service($service)
 
     // If we do not have a cmd from the check script, build one.
     if ($check_cmd == "") {
-	if ($config['disable-H'] == 1) {
-		$check_cmd = $config['nagios_plugins'] . "/check_" . $service['service_type'];
-		$check_cmd .= " " . $service['service_param'];
-	} else {
-		$check_cmd = $config['nagios_plugins'] . "/check_" . $service['service_type'] . " -H " . ($service['service_ip'] ? $service['service_ip'] : $service['hostname']);
-		$check_cmd .= " " . $service['service_param'];
-	}
+        if ($config['disable-H'] == 1) {
+            $check_cmd = $config['nagios_plugins'] . "/check_" . $service['service_type'];
+            $check_cmd .= " " . $service['service_param'];
+        } else {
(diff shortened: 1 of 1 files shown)


laf:
Can you explain why this is needed?

Cormoran96:
This is necessary because some check does not support the -H option. For exemple check for proxmox (https://exchange.nagios.org/directory/Plugins/Operating-Systems/%2A-Virtual-Environments/OpenVZ/pve-2Dmonitor/details)
and many others that I had teste. I prefer to be free to add it by hand rather than being forced.

laf:
For the some checks that don't, all that's required is to create a custom service check like the others do.

Removing the -H will mean people have to go back to existing params and update them to include it again for specific ones or add the alternative passing of a hostname/ip the check requires.

Cormoran96:
Yes but my commit does not propose remove it but by default he stays, but if the user do not want it. Is it possible. with param $config['disable-H'] = 1;

Cormoran96:
Ok i don't know this way. How to configure it ? But, I think it make sense to add this options to leave the choice to the user.

laf:
Honestly, no, the real solution is a webui option per service check or a custom service check file. The former is preffered.

Cormoran96:
Yes I agree with you. But I have no idea🤔, how implement that.

Somone does have an idea with Model ?

Le jeu. 31 mai 2018 à 17:56, Neil Lathwood <notifications@github.com> a
écrit :

laf:
@Cormoran96 

You would need to:

  - Update `html/includes/modal/new_service.inc.php` to allow for `-H` to be disabled (checkbox using something like `<input type="checkbox" name="token-status" data-token_id="__ID__" data-off-text="No" data-on-text="Yes" data-on-color="danger" '.$api_disabled.' data-size="mini">` That's taken from somewhere else so update the names.
  - Update the services table by adding a new schema file in `sql-schema`, give it the next incremental file number and put your update query in to add the new column.
  - Update the `includes/services.inc.php` `poll_service()` to use this new option.

laf:
Closing this PR as it isn't the correct way to go on this.

PR #8803, opened 2018-06-05
Refactor alert templates to use Laravel Blade templating engine

laf:
I'll try and explain some of the changes in this.

1. I've added alerts-cli to the init modules so that only the alerts.php script will re-attempt to boot laravel.

2. The changes to debug was to stop laravel having issues with unset variables as it bails out on them. This may not be needed after I did the above but I've left it in.

3. I've basically removed the old template editing screen and put an indicator up that users need to convert the templates for which for now, I've left a simple convertor in place which has worked across all templates I've tried but it looked too complicated to auto convert in sql queries.

4. The codebase still supports the old templates and I've moved the old docs for users to still access.

5. Hopefully the class structure I chose works. The idea is we could add other templating engines but I actually hope we don't!

6. Added a fix in for Philips Hue transports as any config back in the array classed it as active so tried to send via that transport.


html/pages/templates.inc.php
```diff
@@ -13,3 +13,76 @@
...
+        ' @endif',
+        '@foreach ($faults as $key => $value)',
+        '@foreach ($contacts as $key => $value)',
+        '@endforeach ',
+        '@php echo \1; @endphp ',
+        '$value[\'string\']',
+        '{{ $\1[\'\2\'] }}',
+        '{{ $\1 }}',
+    ];
+    return preg_replace($find, $replace, $old);
```
murrant: You might need this to escape dollar sign followed by number.
```php
addcslashes($old, '$')
```

LibreNMS/Alert/Template/Template.php
```diff
@@ -0,0 +1,55 @@
...
+     * @param null $obj
+     * @return mixed
+     */
+    public function getTemplate($obj = null)
+    {
+        if ($this->template) {
+            // Return the cached template information.
+            return $this->template;
+        }
+        $this->$template = AlertTemplate::whereHas('map', function ($query) use ($obj) {
```
murrant: `$this->$template` -> `$this-template`

LibreNMS/Alert/Template/Template.php
```diff
@@ -0,0 +1,55 @@
...
+            $query->where('alert_rule_id', '=', $obj['rule_id']);
+        })->first();
+        if (!$this->$template) {
+            $this->$template = AlertTemplate::where('name', '=', 'Default Alert Template')->first();
```
murrant: `$this->$template` -> `$this-template`

html/pages/templates.inc.php
laf: You mean from the template like `$1`? If so where would this be from as we don't have any tables or variables starting with numbers.

laf:
Should be all good to go now.

Olen:
Getting this when trying to test different alerts:

```
./scripts/test-alert.php -v -h sw1.olen.net -r 1

PHP Fatal error:  Uncaught Error: Call to a member function connection() on null in /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:1058
Stack trace:
#0 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(1024): Illuminate\Database\Eloquent\Model::resolveConnection(NULL)
#1 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(859): Illuminate\Database\Eloquent\Model->getConnection()
#2 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(818): Illuminate\Database\Eloquent\Model->newBaseQueryBuilder()
#3 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(802): Illuminate\Database\Eloquent\Model->newQueryWithoutScopes()
#4 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(1374): Illuminate\Database\Eloquent\Model->newQuery()
#5 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(1386): Illuminate\Database\Eloquent\Mod in /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php on line 1058
```

Rule 1 is a simple "processors.processor_usage > 90"

But I get the same error with all hosts and rules I have tested so far

Olen:
Seems like the variable $transport is not set
I have added "Transport: {{ $transport }}" to my default template, and the alerts come with only "Transport: "
Also, @if ($transport == "mail") does not seem to work as expected

laf:
> Getting this when trying to test different alerts:
> 
> ./scripts/test-alert.php -v -h sw1.olen.net -r 1
> 
> PHP Fatal error:  Uncaught Error: Call to a member function connection() on null in /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:1058
> Stack trace:
> #0 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(1024): Illuminate\Database\Eloquent\Model::resolveConnection(NULL)
> #1 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(859): Illuminate\Database\Eloquent\Model->getConnection()
> #2 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(818): Illuminate\Database\Eloquent\Model->newBaseQueryBuilder()
> #3 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(802): Illuminate\Database\Eloquent\Model->newQueryWithoutScopes()
> #4 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(1374): Illuminate\Database\Eloquent\Model->newQuery()
> #5 /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(1386): Illuminate\Database\Eloquent\Mod in /opt/librenms/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php on line 1058
> Rule 1 is a simple "processors.processor_usage > 90"
> 
> But I get the same error with all hosts and rules I have tested so far

I wonder if this is what @murrant pointed out in the code review. If so, `./scripts/github-remove -d` then `./scripts/github-apply 8803` and re-try

laf:
> Seems like the variable $transport is not set
> I have added "Transport: {{ $transport }}" to my default template, and the alerts come with only "Transport: "
> Also, @if ($transport == "mail") does not seem to work as expected

I've just pushed an update to fix this.

You'll need to remove the patch and re-add + re-run `./composer.phar install`

laf:
I'm ok with how this is now.

laf:
Ok updated to have `$alert->FIELD` now in the templates.

Legacy templates should continue to run as normal.

Olen:
Since you are doing a lot of work on this anyway - one thing I have been missing in the alerts is the ability to show exactly what triggered the alert in the first place.
It could simply be a json_decode() of "select builder from alert_rules where id=?" added as $alert->ruleset or something before it is sent to the template engine.

Olen:
And this could be fixed already (I have not re applied the PR lately).  If not, this is just a reminder so it won't be lost.
I believe the Recovery Title is not being used anymore.

laf:
> And this could be fixed already (I have not re applied the PR lately). If not, this is just a reminder so it won't be lost.
> I believe the Recovery Title is not being used anymore.

I haven't gone to specifically fix it but it works for me.

Using the old or new format?

Olen:
> > I believe the Recovery Title is not being used anymore.

> I haven't gone to specifically fix it but it works for me.
> Using the old or new format?

The new format. But it did not work with an old template I still had around either.
I'll try to debug it a bit more.

Olen:
> Added $alert->builder. Not json_decoded() as then you'd need to iterate array returned. So you can just do {!! $alert->builder !!}

That is fine with me.  I'll probably not just display it raw (with {!! !!}, but I can easily json_decode it myself to do some nice "Alert triggered by macros.port_out_usage_perc > 75" or whatever I want.

Olen:
After updating now the Recovery Title is used correctly, however, for some reason $alert->transport is not set correctly when you have multiple transports using the same alert-template.

It is set correctly in the event-log:
Issued recovery for rule 'Service Critical' to transport 'irc'
But {{ $alert->transport }} will display "mail" on both irc and in the email.

laf:
Should be all done now.

No idea why travis is failing on this.

laf:
Resolved the conflict on `253.sql`

Any further feedback from anyway? @librenms/reviewers

pheinrichs:
@laf Tested, LGTM

laf:
Thanks for testing @pheinrichs 

This has now been rebased and conflicts resolved, should be good to go as far as I'm concerned.

laf:
Thanks for all the testing folks. Let's hope this works ok :)

PR #8810, opened 2018-06-08
Allow the hostname to resolve to the sysName, ie Dynamic DNS

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

The purpose of this update is to allow one to add and monitor a device using a dynamic dns name, or an entry that you do not want to display as the default name. This option allows you to change the displayed name to match the sysName regardless of the DNS name.

### DIFF
--- a/doc/Support/Configuration.md
+++ b/doc/Support/Configuration.md
@@ -284,6 +284,10 @@ Enable or disable the sysDescr output for a device.
 $config['force_ip_to_sysname'] = false;
 ```
 When using IP addresses as a hostname you can instead represent the devices on the WebUI by its SNMP sysName resulting in an easier to read overview of your network. This would apply on networks where you don't have DNS records for most of your devices.
+```php
+$config['force_hostname_to_sysname'] = false;
+```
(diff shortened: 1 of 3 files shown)


murrant:
Would a `always_use_sysname` setting work?  Or does that not cover your use case?

InsaneSplash:
I decided to split the options to implement in 2, so that the original option of IP to sysName could still be used on its own in case the user didn't want the sysName applied to their DynDNS host names etc. It just provides some flexibility. I guess one could also add the global option which could apply both rules?

InsaneSplash:
ETA? :)

murrant:
@InsaneSplash. You want dns to be changed to sysname, but not IPs?

murrant:
@InsaneSplash. You want dns to be changed to sysname, but not IPs?

laf:
From the code and config names I'd say yes that's the correct scenario as we already have ip -> sysname so basically with both these, sysName will just always be used but allows flexibility for all scenarios

PR #8811, opened 2018-06-08
Ability to enable debug output with wrappers

murrant:
Output is redirected to a per-device file in the log directory.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/discovery-wrapper.py
+++ b/discovery-wrapper.py
@@ -193,7 +193,7 @@ discovered_devices = 0
 parser = argparse.ArgumentParser(description='Spawn multiple discovery.php processes in parallel.',
                                  formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 parser.add_argument('-d', '--debug', action='store_true', default=False,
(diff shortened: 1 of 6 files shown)


discovery-wrapper.py
```diff
@@ -188,13 +190,16 @@ def memc_touch(key, time):
...
-    amount_of_workers = int(sys.argv[1])
-    if amount_of_workers == 0:
-        print "ERROR: 0 threads is not a valid value"
-        sys.exit(2)
-except:
-    amount_of_workers = 1
+parser = argparse.ArgumentParser(description='Spawn multiple discovery.php processes in parallel.',
+                                 formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+parser.add_argument('-d', '--debug', action='store_true', default=False,
+                    help="Enable debug output.")
```
laf: We should definitely warn here that this could lead to a large increase in disk usage if left enabled.

PR #8821, opened 2018-06-14
Allow ping checks to be ran separately from polling

murrant:
Allows ping checks at intervals not tied to the poller.  Pointless if you are not alerting on device status.
I updated the rrdstep.php script to treat ping-perf files separately and made it so it only converts if needed.

Docs here: https://docs.librenms.org/Extensions/Fast-Ping-Check/

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
I'm assuming this is to be invoked from cron?

I'm sure it works (haven't tested it) but is php really what we want to use here along with invoking it via cron? Seems like this would be better as part of the new poller service (or written in Go to support < 1 minute checks).

murrant:
Yes, I would add support to the python service too.
I don't know why php isn't ok.   fping does parallel pinging, one fping command pings all hosts.
Using PHP allows us to take advantage of Eloquent so we get nice things like this:
```php
$result = new DevicePerf($captured);
 $device->perf()->save($result);
```

cchance27:
Would be interesting to benchmark this against say a go pinger as i think the latency difference from a well coded go ping vs a php spawn of fping would be decent bit at least when scaled to say a few hundred or thousand devices, could be wrong though

murrant:
Why would a Go pinger be faster than a pinger written in C?

cchance27:
I’m referring to the wrapper+fping vs go-only instead of spawning a fping process

murrant:
@cchance27 feel free to code something up and test it.

laf:
If you can rebase this I'll check it out on a 4.5k+ device install

laf:
Just need docs.

I can only test this on my dev install:

```bash
-bash-4.2$ ./ping.php
Pinged 9 devices in 2.78s
-bash-4.2$ php artisan ping


  Not enough arguments (missing: "groups?* | Optional List of distributed poller groups to poll").
```

I know the artisan command isn't to be used but the code suggests it would return a message indicating that rather than showing missing options.

laf:
Probably just https://github.com/librenms/librenms/tree/master/doc/Extensions as a new file?

PR #8822, opened 2018-06-15
HPE ILO power fix

TomEvin:
cpqHeTemperatureThreshold is '0', which cause alerts. Fixed to '90'. Tested on Proliant Gen8-9.

### DIFF
--- a/includes/definitions/discovery/hpe-ilo.yaml
+++ b/includes/definitions/discovery/hpe-ilo.yaml
@@ -19,7 +19,7 @@ modules:
                     snmp_flags: '-OQUs'
                     index: 'cpqHeTemperatureCelsius.{{ $index }}'
                     descr: cpqHeTemperatureLocale
-                    high_limit: cpqHeTemperatureThreshold
+                    high_limit: '90' #cpqHeTemperatureThreshold
                 -
                     oid: cpqDaPhyDrvTable
                     value: cpqDaPhyDrvCurrentTemperature


laf:
Is this not just a bug with the firmware you are running? According to the mib `-99` is the unreported value so 0 is genuine or a bug.

laf:
I'm assuming the OID in https://github.com/librenms/librenms/pull/6996/files is what we are already using here. That was your PR - was that going to be wrong as well out of interest?

TomEvin:
Since that PR we switched to yaml, and yes, we are using the same OID-s.
In the meantime, I have checked there are updated HPE mib files too. Give me a couple of days, I'll check them too with the latest firmwares, but based on a quick look this cpqHeTemperatureThreshold OID not changed, but maybe there are new OIDs.

laf:
> Since that PR we switched to yaml

Indeed but the yaml was taken from that PR code so I expect the OID translates to the one you're proposing to remove.

Have static high / low values isn't ideal as they should be provided by the device. If that doesn't work then static high/low can be ok but we should be sure that it is broke before merging.

TomEvin:
I udpeted every software and firmware to the current versions, and I have checked the the SNMP with the latest MIBs, powersupply "cpqHeTemperatureThreshold" is still "0".
image
I do not have really any other idea then the fix threshold limit for this. :(

laf:
Ok let's go with that then. Can you add additional test data for this device please, we currently don't have any test data for it which means bad changes can break stuff: https://docs.librenms.org/#Developing/os/Test-Units/#example-workflow

TomEvin:
"Checked 2681 files in 48.7 seconds
No syntax error found
Tests ok, submit away :)"

Is that enough, or should I copy/paste the snmpres file here?

TomEvin:
Here is the hpe-ilo.snmprec: https://pastebin.com/mJQmfPXD
.json not generated because of "sh: 1: snmpsimd.py: not found"

laf:
I'll put the test data in a separate PR as this was submitted from OPs master branch we can't push to it.

Issue #8788, opened 2018-05-29
Add Occam\Calix OS 8.x


MHammett:
**Please do not post bugs or ask questions**

**For help and support you can use our [Discord server](https://t.libren.ms/discord) or our [community site](https://community.librenms.org).**

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Please provide **ALL** info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
https://p.libren.ms/view/6d242bd1
https://p.libren.ms/view/e549466b
https://p.libren.ms/view/ba32046e

- [x] Please provide as much detail as possible.
You support Occam OS (now Calix) older versions, but version 8 is not being recognized as Occam\Calix. It's being detected as generic.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

peelman:
We're trying to get get rid of our B6's, but they are going to have along tail; so some monitoring would be handy.

MHammett:
@peelman I added some older versions earlier. I just upgraded to 8.0.

https://github.com/librenms/librenms/issues/8359
https://github.com/librenms/librenms/issues/8361
https://github.com/librenms/librenms/issues/8362

murrant:
What would be the correct/best name for the OS?

MHammett:
Calix bought Occam. It was called OccamOS, but obviously post-acquisition, the Occam part has been stripped. However, it hasn't been replaced with CalixOS or anything like that. They appear just seem to call it B6 system, which would be the entire product family, not just the OS. @peelman any thoughts on that?

Here is how a few versions self-identify:

`DeKalb_01_02>show version
OccamOS System V5.5R7 for BLC 6440 01  [shelf:1f0, slot:2]
Kernel build id V5_5R7, made Tue Mar 10 18:31:48 PDT 2009
Booter build id V5_5R7, made Tue Mar 10 18:26:21 PDT 2009

Processor:  440GP Rev. C  at 396 MHz, revision  4.129 (pvr 4012 0481)
VCO: 792 MHz, PLB: 132 MHz, OPB: 66 MHz, EPB: 66 MHz, Sysclk: 33 MHz

Board assembly number is 900032-01, revision  number is D6, type is 1
Serial number is 11467390, MAC address is 00:02:86:00:2a:94
Manufactured by OCCM on 11/19/2004

[...]

PR #8827, opened 2018-06-15
Updated Calix detection

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Fixes: #8788


murrant:
Hmm, this is Occam hardware, I wonder if it has different mibs etc.

We really need to fix the calix os.  It should be called exa os.

PR #8828, opened 2018-06-15
Fix vars parsing on normal get urls

murrant:
(like a form with GET as the method)

The other fix is to modify all GET forms to use a url with a trailing slash.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/vars.inc.php
+++ b/html/includes/vars.inc.php
@@ -11,17 +11,21 @@ foreach ($_GET as $key => $get_var) {
     }
 }
 
+// handle normal encoded urls too
+$working_url = str_replace(['?', '&'], '/', $_SERVER['REQUEST_URI']);
+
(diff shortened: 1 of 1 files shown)


laf:
Is this worth not committing rather than having to try and test it's not broken anything?

murrant:
No, this is not really related to Laravel.

The problem is the form does not have a hardcoded action url.  So, if you visit the page without a trailing slash, the form does not work because the vars code can't parse it.

Seems like a simple fix to prevent the issue there and other places.

The other option is to rewrite the $vars parsing to just use Request.  But then the problem is that Laravel isn't booted for ajax calls right now...

murrant:
Doesn't matter to me.  We can close this and reopen if users report issues.

PR #8833, opened 2018-06-19
Disable autocomplete on SNMPv3 settings

KlaasT:
We have noticed that some Firefox password managers try to autocomplete AuthUser, AuthPass and CryptoPass with user credentials. This can lead to a security issue as the credentials of users are stored in plaintext inside the LibreNMS database.

The autocomplete="off" should disable this behaviour.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Just waiting on tests to complete now.

PR #8842, opened 2018-06-25
corrected display of minigraph when using sysName as hostname

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Corrected widget to display the minigraph when a sysName is used instead of a hostname.


html/includes/common/top-devices.inc.php
```diff
@@ -330,7 +330,7 @@
             $common_output[] = shorten_text($result['storage_descr'], 50).'</td><td class="text-left">';
             $common_output[] = overlib_link($link, print_percentage_bar(150, 20, $percent, null, 'ffffff', $background['left'], $percent.'%', 'ffffff', $background['right']), $overlib_content);
         } else {
-            $common_output[] = generate_device_link($result, generate_minigraph_image($result, $config['time']['day'], $config['time']['now'], $graph_type, 'no', 150, 21), $graph_params, 0, 0, 0);
+            $common_output[] = generate_device_link($result['hostname'], generate_minigraph_image($result, $config['time']['day'], $config['time']['now'], $graph_type, 'no', 150, 21), $graph_params, 0, 0, 0);
```
laf: `generate_device_link()` expects a full `$device` result to be passed. I think you'll need to use `format_hostname()` within `generate_device_link()` which will return the sysName if config.php is set up correctly.

murrant:
I've added tests.

I think when both these settings are enabled, we want basically everything to be sysName. (except when "hostname" is nto an IP or hostname)

PR #8857, opened 2018-06-30
Add a 'session' sensor for PanOS, SRX5800 Flow Accounting

TheMysteriousX:
I couldn't find any docs on adding a new sensor type so I've kind of winged it - I may have missed a few places that it needs to be added.

As an example, I've made our SRX5800's record the flow state on its SPU's - it looks like this:
screenshot 2018-06-30 22 25 24

I had a look at adding Linux TCP Established/Max, but it seems there's no maximum configured so it's of questionable use - can throw it in though if wanted, looks like:

screenshot 2018-07-01 01 03 35


laf:
So for me on this one is that these are in the wrong place. Sensors are for health information which these don't strike me as being applicable for.

Someone has attempted a PR to support a generic 'sensor' based system but I don't think it's ready yet. Ideally what we should have is one central place where you can register any type of 'sensors' against but say what section they belong in.

That needs time from someone though to get that in and make it work.

TheMysteriousX:
Mempools and CPU are implemented using the same code, so that's why I implemented it as health - being as the flow table is of a finite size and can run out.

laf:
> Mempools and CPU are implemented using the same code, so that's why I implemented it as health - being as the flow table is of a finite size and can run out.

Actually mempools and cpu have there own tables but that's for a good reason (well at atleast mempools) as it stores more than just a current value + they run there own disco/poller code.

Personally I'd like to see a more generic table that registers various health metrics and records the group it belongs to rather than trying to fit other things into the sensors table.

laf:
You don't need to close this, my responses are my personal opinion and not of the overall project. Someone else might disagree with me and want this merging.

murrant:
I'd be fine with this as a stop-gap.  At least we would have test data.  I do agree with what @laf said about needing a better place for all of it.

murrant:

murrant:
@TheMysteriousX I've updated this to use the count sensor, but you'll need to add test data for the sensors.

PR #8865, opened 2018-07-04
PSR-4 Plugin support 

daryl-peterson:
Add PSR-4 sub folder loading
Add PHP docblock to functions
Add composer psr4 install path

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/bootstrap/autoload.php
+++ b/bootstrap/autoload.php
@@ -3,16 +3,16 @@
 define('LARAVEL_START', microtime(true));
 
 /*
-|--------------------------------------------------------------------------
-| Register The Composer Auto Loader
-|--------------------------------------------------------------------------
-|
-| Composer provides a convenient, automatically generated class loader
-| for our application. We just need to utilize it! We'll require it
-| into the script here so we do not have to manually load any of
-| our application's PHP classes. It just feels great to relax.
-|
-*/
+  |--------------------------------------------------------------------------
+  | Register The Composer Auto Loader
+  |--------------------------------------------------------------------------
+  |
+  | Composer provides a convenient, automatically generated class loader
+  | for our application. We just need to utilize it! We'll require it
+  | into the script here so we do not have to manually load any of
+  | our application's PHP classes. It just feels great to relax.
+  |
+ */
 
 $plugin_dir = dirname(__DIR__) . "/html/plugins";
 
--- a/bootstrap/autoload.php
+++ b/bootstrap/autoload.php
@@ -14,7 +14,11 @@ define('LARAVEL_START', microtime(true));
 |
 */
 
-@include __DIR__ . '/../vendor/autoload.php';
+$plugin_dir = dirname(__DIR__) . "/html/plugins";
+
+/* @var $loader \Composer\Autoload\ClassLoader */
+$loader = include __DIR__ . '/../vendor/autoload.php';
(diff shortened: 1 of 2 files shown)


bootstrap/autoload.php
```diff
@@ -1,23 +1,26 @@
...
+  |--------------------------------------------------------------------------
+  |
+  | Composer provides a convenient, automatically generated class loader
+  | for our application. We just need to utilize it! We'll require it
+  | into the script here so we do not have to manually load any of
+  | our application's PHP classes. It just feels great to relax.
+  |
+ */
+
+$plugin_dir = dirname(__DIR__)."/html/plugins";
```
murrant: Don't edit this file: See https://getcomposer.org/doc/04-schema.md#psr-4

murrant: @daryl-peterson please read the link I posted.  It does the same thing but is much nicer.  Thanks.

daryl-peterson: Correct looking at the composer.json
"autoload": {
        "psr-4": {
            "App\\": "app",
            "LibreNMS\\": "LibreNMS",
            "LibreNMS\\Tests\\": "tests"
	}
}
Search path is listed for print_r($loader); Only relevant part here
[prefixDirsPsr4:Composer\Autoload\ClassLoader:private]
[LibreNMS\Tests\] => Array
(
    [0] => /local-librenms.org/vendor/composer/../../tests
)

[LibreNMS\] => Array
(
    [0] => /local-librenms.org/vendor/composer/../../LibreNMS
)

[App\] => Array
(
   [0] => /local-librenms.org/vendor/composer/../../app
   [1] => /local-librenms.org/vendor/composer/../laravel/laravel/app
)

Close but means all plugins would have to be int the same namespace as
LibreNMS\Plugins simply because the Plugins file does an include to the plugin file.
[classMap:Composer\Autoload\ClassLoader:private] => Array
(
	[LibreNMS\Plugins] => /local-librenms.org/vendor/composer/../../LibreNMS/Plugins.php
)

Even with the Plugins.php class doing an include a file like

html/plugins/DevReport/DevReport.php

LibreNMS\Plugins\DevReport
class DevReport
{

}
It will not work. because it expects the namespace to be LibreNMS\Plugins
LibreNMS\Plugins\DevReport will not work

So in order to be able to auto load a plugin without doing a manual include of the file is to either
add the 
"LibreNMS\\Plugins\\": "plugins" to the composer.json or add the path using the addPsr4 method to the loader.

daryl-peterson: actually you would have the add 
"LibreNMS\Plugins\": "html/plugins"  to the psr-4 section of composer.json

murrant: Exactly, that is much preferable imho.

daryl-peterson: Great so I'll add "LibreNMS\Plugins"  to the composer.json and revert the autoload. What about the changes to Plugins.php are those good ?

murrant: Yeah those seem fine to me.  Not a fan of long nested ifs myself.

But those changes will require a little testing.  Testing time is short on the maintainers side of things, so you might have to wait awhile before it is merged.

daryl-peterson: No problem. Thanks for the input

PR #8875, opened 2018-07-08
Updated shebang on librenms-service.py

RyanMorash:
Changed shebang to point to python3 as per documentation the service does not work with python2

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
I think python3 is available on pretty much all distros (that have python3)

PR #8893, opened 2018-07-10
discovery.php: don't exit(5) without new devices discovered

costasd:
Don't exit with a non-zero value in case no new devices were discovered, as non-zero codes are considered failed jobs for most cron implementations and produce logs like the following (in Debian):

```
cron.info CRON[963890]: (CRON) error (grandchild #963892 failed with exit status 5)
``` 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/discovery.php
+++ b/discovery.php
@@ -152,3 +152,8 @@ if (!isset($options['q'])) {
 }
 
 logfile($string);
+
+if ($doing !== 'new' && $discovered_devices == 0) {
+    # No discoverable devices, either down or disabled
+    exit(5);
+}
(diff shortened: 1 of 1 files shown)


TheMysteriousX:
It's not safe to remove: https://github.com/librenms/librenms/blob/e9e30c5e98d28c486b7a81974513b375b42a8d79/LibreNMS/service.py#L361

Could probably be implemented in another way - @costasd, why/how are you invoking a discovery run with no devices?

TheMysteriousX:
Assuming this is a ```-h new```, I think the correct fix is:

```
if ($doing !== 'new' && $discovered_devices == 0)
```

costasd:
@TheMysteriousX True, it's coming from the cronjob with `-h new`.

costasd:
Thanks @laf and @TheMysteriousX!

PR #8898, opened 2018-07-14
Fixed snmptraps.

TheGreatDoc:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Its a fix for the basic ones that comes with LibreNMS


snmptrap.php
```diff
@@ -21,22 +23,32 @@
 $init_modules = array();
 require __DIR__ . '/includes/init.php';
 
-$entry = explode(',', $argv[1]);
+// Creates an array with trap info
+while ($f = fgets(STDIN)){
```
laf: Why the switch away from `$argv[1]`?

TheGreatDoc: Dont ask me why but $argv[] only contains the script itself so I used other approach to parse info

laf: Are you sure that wasn't just `$argv[0]` that contained the script?

TheGreatDoc: Well... I'm almost sure. I've tried to dump $argv with `var_export($argv, true)` to know what were in the array and only the script was dumped.

With my new implementation, dumping $entry with `var_export($entry, true)` after the while dumped everything recieved in the trap.

As you know, Im not really a developer so I may have missed something that makes $argv to only dump the script.

snmptrap.php
```diff
@@ -46,6 +44,7 @@
 }
 
 if (!$device['device_id']) {
+    echo "unknow device\n";
```
murrant: Typo here.

murrant:
Looks like you made some good changes here @TheGreatDoc .

I suggest you add some tests to make sure things work properly and continue to work.
To make it easier to test move most the code in the snmptrap.php file into a function or class. So that snmptrap.php only captures input then calls a function.  Then in the test you can call that function and test the result (database changes, etc)

If you wanted to go crazy, you could make the code dispatch events for traps and then all we need to do is write handlers for each event we want. https://laravel.com/docs/5.4/events

snmptrap.php
```diff
@@ -22,22 +24,31 @@
...
 
-$entry = explode(',', $argv[1]);
+// Creates an array with trap info
+while ($f = fgets(STDIN)) {
+    $entry[] = $f;
+}
 
-logfile($argv[1]);
+// Formatting array
+$hostname = trim($entry['0']);
```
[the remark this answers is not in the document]
TheGreatDoc: I used as it was in some other file. Can't remember by now. Modifying

includes/snmptrap/authenticationFailure.inc.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
+log_event('SNMP Trap: Authentication Failure: ' . $device['sysName'], $device, 'auth', 3, $device['hostname']);
```
laf: Logging by device sysName is specific to your env. You either need to log using hostname or run it through `format_hostname()` to get the type the user wants.

includes/snmptrap/bgpBackwardTransition.inc.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+$bgppeerip = strstr(strstr($entry['5'], " ", true), ".");
+$bgppeerip = substr($bgppeerip, 1);
+$bgpstatus = trim(strstr($entry['5'], " "));
+
+$string = "bgpBack " . $bgppeerip . " - " . $bgpstatus . "\n";
+
+$fp = fopen('/root/trap.txt', 'a');
```
laf: Can you remove this debug pls :)

includes/snmptrap/authenticationFailure.inc.php
TheGreatDoc: Yeah. Done!

includes/snmptrap/bgpBackwardTransition.inc.php
TheGreatDoc: how embarrassing....

TheGreatDoc:
@laf 
Fixed the logging to sysName. Now is using format_hostname($device).
Debug removed... Shame on me.
I've also changed 'astext' to get it from config if available.

 @murrant 
Removed `' '` from $entry[]
About making a function. I can do it with a little bit help. Will be in touch in discord. I'm crazy, but no as much to make events in laravel. Maybe with a really clear example I can try, but Im not really a developer.

murrant:
Surprise, you are a developer @TheGreatDoc (it isn't as grand of a thing as you think).

Don't worry about the events.  To put it in a function, just move all the code inside a function that accepts the input as a string.

Take a look at the syslog handling for an idea (and for how to test).

TheGreatDoc:
@murrant LOL. I really think calling myself developer is an offense to real devs.

About the function. Do you like this way? I'm not sure how to do test. Syslog is pretty "universal" but, for example, to test bgpEstablished/bgpBackward a BGP session with specific Peer IP and Device is needed so I dont know where to start.

laf:
Not sure where Tony would say to do this but:

```
if (!$device['device_id']) {
    exit;
}

$file = $config['install_dir'].'/includes/snmptrap/'.$entry['1'].'.inc.php';
if (is_file($file)) {
    include "$file";
} else {
    echo "unknown trap ($file)";
}

```

Move that to `includes/snmptrap.inc.php` into a function called `process_trap($device, $data)`.

Then back in `snmptrap.php` call `process_trap($device, $entry)`.

A class would be better but if that's not something you feel comfortable with then no worries.

includes/snmptrap.inc.php
```diff
@@ -0,0 +1,18 @@
...
+function process_trap($device, $entry) {
+    global $config;
+
+    $oid = trim(strstr($entry[3], " "));
+    $oid = str_replace("::", "", strstr($oid, "::"));
+
+    $file = $config['install_dir'].'/includes/snmptrap/'.$oid.'.inc.php';
+
+    if (is_file($file)) {
+        include "$file";
```
murrant: The quotes could be removed here, but it is fine either way.

snmptrap.php
```diff
@@ -29,4 +29,20 @@
     $entry[] = $f;
 }
 
-snmptrap($entry);
+$hostname = trim($entry[0]);
+$ip = str_replace(array("UDP:","[","]"), "", $entry[1]);
+$ip = trim(strstr($ip, ":", true));
+
+$device = @dbFetchRow('SELECT * FROM devices WHERE `hostname` = ?', [$hostname]);
```
murrant: You should check the ip field too.  This might work.
```php
$device = dbFetchRow('SELECT * FROM devices WHERE `hostname`=? OR `hostname`=? OR `ip`=?', [$hostname, $ip, inet_pton($ip)]);
```

Remove the @ from these queries

snmptrap.php
```diff
@@ -29,4 +29,20 @@
...
+
+if (!$device['device_id']) {
```
murrant: This check should be `if (empty($device)) {`

snmptrap.php
```diff
@@ -29,4 +29,20 @@
...
+    $device = @dbFetchRow('SELECT * FROM ipv4_addresses AS A, ports AS I WHERE A.ipv4_address = ? AND I.port_id = A.port_id', [$ip]);
+}
+
+if (!$device['device_id']) {
```
murrant: This check should be `if (empty($device)) {`

snmptrap.php
```diff
@@ -22,24 +24,25 @@
...
+}
 
-logfile($argv[1]);
+$hostname = trim($entry[0]);
+$ip = str_replace(array("UDP:","[","]"), "", $entry[1]);
+$ip = trim(strstr($ip, ":", true));
 
-// print_r($entry);
-$device = @dbFetchRow('SELECT * FROM devices WHERE `hostname` = ?', array($entry['0']));
+$device = @dbFetchRow('SELECT * FROM devices WHERE `hostname` = ?', [$hostname]);
```
murrant: You should check the ip field too.  This might work.
```php
$device = dbFetchRow('SELECT * FROM devices WHERE `hostname`=? OR `hostname`=? OR `ip`=?', [$hostname, $ip, inet_pton($ip)]);
```

Remove the @ from these queries

snmptrap.php
```diff
@@ -22,24 +24,25 @@
...
 
 if (!$device['device_id']) {
```
murrant: This check should be `if (empty($device)) {`

snmptrap.php
```diff
@@ -22,24 +24,25 @@
...
-    $device = @dbFetchRow('SELECT * FROM ipv4_addresses AS A, ports AS I WHERE A.ipv4_address = ? AND I.port_id = A.port_id', array($entry['0']));
+    $device = @dbFetchRow('SELECT * FROM ipv4_addresses AS A, ports AS I WHERE A.ipv4_address = ? AND I.port_id = A.port_id', [$ip]);
 }
 
 if (!$device['device_id']) {
```
murrant: This check should be `if (empty($device)) {`

TheGreatDoc:
@murrant I did the changes.

Just one question for my knowledge. For device detection, in the original snmptrap.php were two sql queries. One for hostname and one for IP. They were already there.

In your changes, you ask me to edit the first query to include IP as hostname or IP as IP. 

`$device = dbFetchRow('SELECT * FROM devices WHERE `hostname`=? OR `hostname`=? OR `ip`=?', [$hostname, $ip, inet_pton($ip)]);`

Is the second one, that is a fallback if hostname is not found needed?

`$device = @dbFetchRow('SELECT * FROM ipv4_addresses AS A, ports AS I WHERE A.ipv4_address = ? AND I.port_id = A.port_id', [$ip]);`

Whats the difference betwen them for IP lookup?

includes/init.php
```diff
@@ -64,6 +64,7 @@ function module_selected($module, $modules)
 require_once $install_dir . '/includes/datastore.inc.php';
 require_once $install_dir . '/includes/billing.php';
 require_once $install_dir . '/includes/syslog.php';
+require_once $install_dir . '/includes/snmptrap.inc.php';
```
laf: We shouldn't be including this for all scripts, it should be just included in snmptrap.php.

I'm going to bump this to 1.43 release.

murrant: Honestly, it is just one function.  I would just put the function in one of the existing includes.

TheGreatDoc: I can either move the function to functions.php or the include to snmptrap.php

TheGreatDoc: I think that the laf option is better. Why need all (gui/cli) load the trap function (even if its only 1) if only snmptrap.php is going to use it? Also, dont know if for doing test you need to load the function in the test so Im lost.

Will do whatever you guys decide

TheGreatDoc: I really wish I could but thats really out of my skill. If you point me with good examples of function to classes maybe, with time, I can get it working in a class.
For now, I'm going to move the include to snmptrap.php.

laf:
Needs a rebase to fix the conflicts.

@TheGreatDoc Can you test all of this?

@murrant You would have been better to just fix the other snmptrap PR :p

TheGreatDoc:
@laf Tested and working perfect!

How do I properly rebase? Tried git `pull --rebase origin master` but it have conflicts and I'm unsure on what to do.

murrant:
@TheGreatDoc that was the right way.

I've fixed it here for you.
Device.php = merge all the use statements
composer.lock = accept upstream content hash.

murrant:
@TheGreatDoc I think composer.lock had something wrong. We'll see if this fixes.

murrant:
@TheGreatDoc there was a test that unset($_SERVER);  which messes up Laravel booting (this is the first test that boots Laravel).  So running this tests alone worked fine, but failed when running them all.

I tracked it down by adding this code to TestCase.php (which all tests extend so it runs in every test class).
```php
    public function testScriptName()
    {
        $this->assertTrue(isset($_SERVER['SCRIPT_NAME']), 'Broken $_SERVER variable!!!');
    }
```

So the first test where it failed or the one before was the that broke things.

PR #8903, opened 2018-07-16
Fixed device dependancy webui issues

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

When debugging the dependancy issue someone raised I found I couldn't actually manage them via the webui. This is because rather than the expected ID, a name is sent. Instead of trying to find the break I decided that it would make more sense to support both numerical IDs and hostnames.

### DIFF
--- a/html/includes/forms/delete-host-dependency.inc.php
+++ b/html/includes/forms/delete-host-dependency.inc.php
@@ -15,35 +15,38 @@
 use LibreNMS\Authentication\Auth;
 
 if (!Auth::user()->hasGlobalAdmin()) {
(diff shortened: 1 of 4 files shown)


laf:
image

That's current master. Definitely doesn't work for me.

laf:
Ok so this is because you can select the text you've entered as an option:

image

I've typed `laf` here, selecting only the last two are valid but the UI allows you to select the first which is where it fails. I can't see an initial fix in using select2 so I'll stick with this for now.

murrant:
@laf I think the correct easy fix is to disable tags
By deleting this line on all of the select2's
```
tags: true
```

PR #8907, opened 2018-07-16
Don't depend on String_Blade_Compiler so hard

murrant:
Allows things to boot if it is missing but displays a toast.
Allows Toastr facade to work on legacy pages too.

Not 100% sure if we want this.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/legacy_index.php
+++ b/html/legacy_index.php
@@ -303,17 +303,17 @@ foreach (dbFetchRows('SELECT `notifications`.* FROM `notifications` WHERE NOT ex
 }
 
 if (is_array($msg_box)) {
-    echo("<script>
+    echo "<script>
         toastr.options.timeout = 10;
         toastr.options.extendedTimeOut = 20;
-    ");
+        </script>
+    ";
+
     foreach ($msg_box as $message) {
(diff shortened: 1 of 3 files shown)


html/legacy_index.php
```diff
@@ -303,17 +303,17 @@ function popUp(URL)
...
-    ");
+        </script>
+    ";
+
     foreach ($msg_box as $message) {
-        $message['type'] = mres($message['type']);
-        $message['message'] = mres($message['message']);
-        $message['title'] = mres($message['title']);
-        echo "toastr.".$message['type']."('".$message['message']."','".$message['title']."');\n";
+        Toastr::add($message['type'], $message['message'], $message['title']);
```
laf: Is this working for you? I can't seem to get it working.

`FORCE=1 ./composer.phar remove wpb/string-blade-compiler`

I would have expected that to have led to a toastr notification but I don't see anything.

murrant: @laf you have to revert composer.json after removing it to get a validation error.
I did get a toastr after removing the package.

PR #8908, opened 2018-07-16
BUG - test-template.php upgrade after #8803 PR, following test-alert.php code

PipoCanaja:
Hello

Following this post, ( https://community.librenms.org/t/bug-test-template-php-not-running-after-laravel-upgrade/4639/9 ), here is the PR.

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [X] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

PR #8910, opened 2018-07-16
BUG - test-template.php upgrade after #8803 +  #8908 PR

PipoCanaja:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Close, but it should actually be AlertData instead of a collection.

PipoCanaja:
Probably something like this then.

PipoCanaja:
Looks pretty good to me, as you would say 👍

laf:
Thanks again @PipoCanaja

PR #8911, opened 2018-07-17
Custom error page

murrant:
Generic error page, just a dump of the current page with additional information.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


composer.lock
```diff
@@ -3487,50 +3487,6 @@
             ],
             "time": "2018-01-29T19:49:41+00:00"
         },
-        {
```
laf: Was this meant to be committed?

murrant: Nope, a quick update and I didn't notice this file snuck in.

PR #8926, opened 2018-07-20
WebUI - 'Disabled' and 'Down' state for devices/links rendered on NetworkMap

PipoCanaja:
Hello

Here is a proposal for a dedicated color on disabled devices displayed on the Network Maps. Usecase is that during a migration, we are "disabling" the old devices and we delete them only a few days/weeks later. This change allows to see the productive devices as well as the disabled ones and clearly know which is which. 

Thanx

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Got an example of what this looks like?

PipoCanaja:
Something like this, right now. 

image

Looking at the code, it seems that the color of the links is mapped to the load of the links. We could imagine the the hosts color changes : 
 - host disabled -> host GRAY, links GRAY as well, could be dotted (dotted would mean "not running"
 - host down -> host RED, links dotted RED
 - host up -> host BLUE, links color depending of the speed (like it is now). 

Bye

laf:
@PipoCanaja Looks fine to me. Is this still classed as WIP?

laf:
> Do you suggest me to commit the "try and fails" or to keep it "internal" (and providing you screenshots until then) and when something appears good, then I commit ?

Sorry I'm not sure what you mean by this.

PipoCanaja:
The map as in the screenshot is now committed. In my environnement, it seems good. But it should definitly be tested now in more networks.  

Let me know how it goes.

PR #8931, opened 2018-07-23
Make alert transport form compatible with sql strict mode

vivia11:
A quick fix to have the form compatible with users running sql strict mode (like my own team).

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/forms/alert-transports.inc.php
+++ b/html/includes/forms/alert-transports.inc.php
@@ -62,6 +62,8 @@ if (empty($name)) {
     $status = 'error';
     $message = 'Missing transport information';
 } else {
+    
+    $is_default = $is_default ? 1 : 0;
     $details = array(
         'transport_name' => $name,
         'is_default' => $is_default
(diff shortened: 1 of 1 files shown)


html/includes/forms/alert-transports.inc.php
```diff
@@ -62,6 +62,7 @@
     $status = 'error';
     $message = 'Missing transport information';
 } else {
+    $is_default = $is_default ? 1 : 0;
```
murrant: I think it would be better if you did this up above (line 53 & 55).

vivia11: That's a good suggestion, I'll make that change.

html/includes/forms/alert-transports.inc.php
```diff
@@ -46,15 +46,9 @@
 
 $transport_id        = $vars['transport_id'];
 $name                = $vars['name'];
-$is_default          = isset($vars['is_default']) ? $vars['is_default'] : null;
+$is_default          = isset($vars['is_default']) ? 1 : 0;
```
murrant: This isn't the same logic.

Try:
```php
$is_default = (int)(isset($vars['is_default']) && $vars['is_default'] == 'on');
```

PR #8935, opened 2018-07-24
Use PDO for MySQL queries

murrant:
Gut dbFacile
mres() now does nothing... could have some vulnerable queries, but most are ok

TODO
- [x] IRC client broke
- [x] Install broke

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/DB/Eloquent.php
+++ b/LibreNMS/DB/Eloquent.php
@@ -125,7 +125,7 @@ class Eloquent
     public static function DB()
     {
         // check if Laravel is booted
-        if (class_exists('DB')) {
+        if (defined('LARAVEL_START') && class_exists('DB')) {
             return \DB::connection();
         }
 
--- a/LibreNMS/DB/Eloquent.php
+++ b/LibreNMS/DB/Eloquent.php
@@ -33,11 +33,12 @@ class Eloquent
 {
     /** @var Capsule static reference to capsule */
     private static $capsule;
+    private static $legacy_listener_installed = false;
 
     public static function boot()
     {
         // boot Eloquent outside of Laravel
-        if (!defined('LARAVEL_START') && class_exists(Capsule::class)) {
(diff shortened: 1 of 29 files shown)


laf:
Checked out for testing on my dev install, leaving overnight.

laf:
I dont' seem to see DB queries now in debug out with `-v` for poller.php or discovery.php.

I'm not sure I'm getting eventlog data now:

`SQL[INSERT IGNORE INTO `eventlog` (`host`,`device_id`,`reference`,`type`,`datetime`,`severity`,`message`,`username`)  VALUES (:host,:device_id,:reference,:type,NOW(),:severity,:message,:username) {"host":168,"device_id":168,"reference":"NULL","type":"alert","severity":1,"message":"Issued critical alert for rule 'Localhost' to transport 'mail'","username":""} 3.79ms]`

Doesn't show up.

Is using IGNORE a wise thing, means that we won't see the mysql errors in the log file anymore.

murrant:
Ah yes:
```
SQL Error! SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '11-8' for key 'unique_alert' (SQL: INSERT INTO `alerts` (`state`,`device_id`,`rule_id`,`open`,`alerted`)  VALUES (:state,:device_id,:rule_id,:open,:alerted)) in includes/alerts.inc.php:236
```

I guess I need to fix the query...

laf:
Re-tested and seems ok still. Found the issue with the installer and it's because I moved /opt/librenms/ out of the way in one shell whilst staying in it in another. In the one I was already in that dir for, running `ll config.php` showed a file existed when in reality - it didn't.

murrant:
FYI, been using this on my production without issues.

murrant:
FYI, the last commit was to fix travis.   Didn't seem to have an affect on normal use.

laf:
> FYI, the last commit was to fix travis. Didn't seem to have an effect on normal use.

Indeed hence I didn't retest :) Nice work with this.

PR #8939, opened 2018-07-24
Add docs for using a base template

pheinrichs:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Alerting/Templates.md
+++ b/doc/Alerting/Templates.md
@@ -70,7 +70,7 @@ If you'd like to reuse a common template for your alerts follow below
 
 create a file in
 ` resources/views/alerts/template.blade.php`
-Scaffold this file with the following
+As an example, add the following:
 
 ```
 <html>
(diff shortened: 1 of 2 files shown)


doc/Alerting/Templates.md
```diff
@@ -65,6 +65,47 @@ Placeholders can be used within the subjects for templates as well although $fau
 
 The Default Template is a 'one-size-fit-all'. We highly recommend defining your own templates for your rules to include more specific information.
 
+## Base Templates
+If you'd like to reuse a common template for your alerts follow below
+
+create a file in
+` resources/views/alerts/template.blade.php`
+Scaffold this file with the following
```
laf: Maybe `As an example, add the following:`

pheinrichs:
Should I update the recommended folder path and / or give include a base template they can use and just ignore that file? 
My concern is that the path i chose is a bit generic and LibreNMS could use it in the future?

laf:
I'd be happy to just have this template file in resources/view/alerts/ and exclude it from git changes or just change the dir to something like you've suggested.

laf:
@pheinrichs If you can update this to how you want it to be we can merge for the tagged 1.42 release.

laf:
I was under the impression we were still discussing what's best with regards to excluding the dir as it's a standard name we might use ourselves later?

PR #8940, opened 2018-07-24
fix: Match interface counter64 OIDs with unsigned bigint

siegsters:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

#####

As discussed on Discord, this is a fix in DB schema so that interface counters that are now often reported as counter64, can be properly recorded in the DB, as unsigned bigint will match the max value of counter64. This created an issue when counters were above the signed value, with current interface speed, for example, not being properly reported, since it could not record the values in the database.


sql-schema/256.sql
```diff
@@ -0,0 +1,24 @@
+ALTER TABLE ports
+  MODIFY `ifInUcastPkts` BIGINT UNSIGNED,
```
murrant: each command must be a single line, so either collapse to one line, or make multiple alter statements.

siegsters: Changed to multiple alter statements

misc/db_schema.yaml
```diff
@@ -1062,6 +1062,31 @@ pollers:
   Indexes:
     PRIMARY: { Name: PRIMARY, Columns: [poller_name], Unique: true, Type: BTREE }
     id: { Name: id, Columns: [id], Unique: false, Type: BTREE }
+poller_cluster:
```
murrant: I think your branch is a little out of date.  Remove this and users changes below.

PR #8943, opened 2018-07-25
Set sql-mode explicitly from app

rkojedzinszky:
This is to not depend on global mysql setting.

Signed-off-by: Richard Kojedzinszky <krichy@nmdps.net>

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@rkojedzinszky thanks for this, I actually started to do the exact same thing here: https://github.com/librenms/librenms/pull/8935

Didn't think to remove it from the install docs and validator though. You missed some items with regards to the phpunit tests and the librenms service (not poller).

Not 100% sure what we want to do to resolve the two.

poller-service.py
```diff
@@ -55,6 +55,7 @@ class DB:
     def __init__(self):
         self.in_use = threading.Lock()
         self.connect()
+        sefl.query("SET SQL_MODE = ''")
```
laf: Spelling mistake here.

rkojedzinszky:
I didn't know that a refactor is in progress to work with strict mode also. Then there should be some way to set sql-mode to the desired value, but I think that also should be done explicitly in code.

poller-service.py
rkojedzinszky: Fixed.

murrant:
Do we actually need to disable strict mode in the python scripts?

laf:
> Do we actually need to disable strict mode in the python scripts?

Not strictly no as the current queries will work but someone could update them in the future.

laf:
> I didn't know that a refactor is in progress to work with strict mode also. Then there should be some way to set sql-mode to the desired value, but I think that also should be done explicitly in code.

It's just an on going thing.

laf:
@rkojedzinszky Any response to this?

PR #8953, opened 2018-07-26
Initial detection for SIAE Microelettronica ALFOplus2 microwave radio device

metavrs:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/siae-alfoplus2.inc.php
```diff
@@ -0,0 +1,3 @@
+<?php
+$version = trim(snmp_get($device, ".1.3.6.1.2.1.47.1.1.1.1.8.1", "-Oqv"), '"');
```
murrant: You don't need trim anymore.

includes/polling/os/siae-alfoplus2.inc.php
```diff
@@ -0,0 +1,3 @@
+<?php
+$version = trim(snmp_get($device, ".1.3.6.1.2.1.47.1.1.1.1.8.1", "-Oqv"), '"');
+$hardware = trim(snmp_get($device, ".1.3.6.1.2.1.47.1.1.1.1.9.1", "-Oqv"), '"');
```
murrant: You don't need trim anymore.

mibs/siae/IANA-ADDRESS-FAMILY-NUMBERS.mib
```diff
@@ -0,0 +1,130 @@
+
+  IANA-ADDRESS-FAMILY-NUMBERS-MIB DEFINITIONS ::= BEGIN
```
murrant: This is an example of a standard mib.  You can find it here: mibs/IANA-ADDRESS-FAMILY-NUMBERS-MIB.  Delete it and any others there are copies of in the mibs folder.

Also, rename the rest of the mibs with the script `./scripts/rename-mibs.php mibs/siae/`

includes/definitions/siae-alfoplus2.yaml
```diff
@@ -0,0 +1,12 @@
+os: alfoplus2
```
murrant: Does this device use the same MIBs or different MIBs as other saie devices?  It may share an OS and you probably don't want this specific of an OS name if that is a case.

If MIBs are only for this device class, this name is ok.

metavrs: The MIBs came directly from SIAE in a zip called:  ALFOplus2_MIB_R1.5.0.  However, it's likely several of the MIB sub-files are shared with their other devices.  

They don't seem to have an OS name that I could find.  Should I just rename it to "siae"?  I already named the mibdir and icons "siae".

mibs/siae/IANA-ADDRESS-FAMILY-NUMBERS.mib
metavrs: Will do.

includes/definitions/siae-alfoplus2.yaml
metavrs: Looks like they do have an OS name for this device.  SM-OS.  I'll make the changes.

includes/polling/os/sm-os.inc.php
```diff
@@ -0,0 +1,3 @@
+<?php
+$version = snmp_get($device, '.1.3.6.1.4.1.3373.1103.7.2.0', '-Oqv', '', '');
```
laf: As you've got the MIB files, can you use them for these snmp lookups.

Please also convert to snmp_get_multi_oid() instead of two separate calls.

metavrs: @murrant said I do not need to declare the mibs on numeric OID's.

laf: But you have the mibs so you should use them. Otherwise, no point in uploading the mibs.

metavrs:
I added mibs to pollers as requested, also removed the Aricent mib files, which while they were included in the official mib zip for this device, they had no impact on LibreNMS.  I also added the wireless sensor class for this OS.

metavrs:
Hello,  Do you need anything on my end to finish this up?  Thank you!

laf:
> Hello, Do you need anything on my end to finish this up? Thank you!

No I was just fixing up tests. All good now.

Issue #8950, opened 2018-07-26
wrong sensor discovery for linux server hostname which contains "ntc"


tomarch:
Hi,

I've some linux server with the string "ntc" int the hostname (for exemple "server-montcherie").
The discovery find some sensors which don't exist.

The faulty code is this line : `if (preg_match("/(Linux).+(ntc)/", $device['sysDescr'])) {`
This line is present in this file:
includes/discovery/sensors/charge/linux.inc.php
includes/discovery/sensors/current/linux.inc.php
includes/discovery/sensors/temperature/linux.inc.php
includes/discovery/sensors/voltage/linux.inc.php

There's not test on the snmp_get result (it return "No Such Instance currently exists at this OID" in my case) before use the discover_sensor function.

laf:
This issue has been closed as it isn't classed as anew device request.

For help please use our Discord server or the community site: http://docs.librenms.org/Support/FAQ/#faq3

PR #8956, opened 2018-07-27
linux sensors - check if value is valid before use discovery_sensor

tomarch:
Fix for issue #8950 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


tomarch:
Ok i've made the change

murrant:
You should probably add a test case with your ntc hostname as well.  Otherwise this might get broken again when this code is refactored.

laf:
I've added a test file in which should cover this.

PR #8966, opened 2018-07-31
Fix Debian Stretch rrdcached Base_Path

MHammett:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
The journal path below is similar...

MHammett:
I don't know enough to know if that should be changed or not. My system appears to work with the stated journal path, but that doesn't mean it's right.

murrant:
default should probably be under /var/lib or something.  Was there any created at install?  `locate rrdcached`

MHammett:
It looks like there's one in /var/lib.

https://p.libren.ms/view/5bc4d694

PR #8968, opened 2018-07-31
Log an event for each Service Status change

PipoCanaja:
Hello

Here is a proposal for a pièce of code to log (using log_events) each service status change onto the device itself. 

Let me know if we should create the "getStatusText()" function to convert the int to "OK", "Critical" etc etc . I searched for such a function but did not find it. 

I also add to use a hack for the SQL cause it would prevent the device_id valueto be available in the resulting $service when device_attribs is empty for the device. If you have a better way to suggest, I'll be happy to use it. 

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/check-services.php
+++ b/check-services.php
@@ -55,7 +55,9 @@ if ($options['h']) {
     }
 }
 
-$sql = 'SELECT * FROM `devices` AS D'
+// need to specify D.device_id as D_device_id because device_id gets overriden in the resulting raw by A.device_id if no attribs are defined for this device
+// TODO: a cleaner way
+$sql = 'SELECT D.device_id as D_device_id,D.*,S.*,A.* FROM `devices` AS D'
        .' INNER JOIN `services` AS S ON S.device_id = D.device_id AND D.disabled = 0 '.$where
        .' LEFT JOIN `devices_attribs` as A ON D.device_id = A.device_id AND A.attrib_type = "override_icmp_disable"'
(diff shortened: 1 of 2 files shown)


check-services.php
```diff
@@ -55,7 +55,9 @@
     }
 }
 
-$sql = 'SELECT * FROM `devices` AS D'
+// need to specify D.device_id as D_device_id because device_id gets overriden in the resulting raw by A.device_id if no attribs are defined for this device
+// TODO: a cleaner way
+$sql = 'SELECT D.device_id as D_device_id,D.*,S.*,A.* FROM `devices` AS D'
```
laf: Can be changed to 

```
SELECT D.*,S.*, attrib_value FROM `devices` AS `D`
```

includes/services.inc.php
```diff
@@ -113,7 +113,7 @@ function discover_service($device, $service)
     echo "$service ";
 }
 
-function poll_service($service)
+function poll_service($service, $device)
```
laf: Should be able to remove this second param now.

PipoCanaja:
Changes are done. I added the WIP tag to test that into production in my environnement to see if it is all right with more services than test machine. Should take a few days. 

PipoCanaja

PipoCanaja:
Works well in my production system as well. OK for me.

Bye
PipoCanaja

PR #8969, opened 2018-07-31
Allow to use full DN as value for member attribute instead of member: username

TheGreatDoc:
Allow to use full DN as value for member attribute instead of member: username

I dont use LDAP so this should be tested with both methods.

For using fulldn as user `$config['ldap_auth_userdn'] = true;` must be set in config.php

This comes from https://community.librenms.org/t/feature-request-full-dn-as-group-member-attibute-in-ldap-auth/4805

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/defaults.inc.php
+++ b/includes/defaults.inc.php
@@ -631,7 +631,7 @@ $config['auth_ldap_groupmemberattr']            = 'memberUid';
 $config['auth_ldap_emailattr']                  = 'mail';
 $config['auth_ldap_cache_ttl'] = 300;
 // How long in seconds should ldap* module cache user information in $_SESSION
-$config['auth_ldap_userdn']                     = true;
+$config['auth_ldap_userdn']                     = false;
 // Uses a users full DN as the value of the member attribute in a group (instead of member: username, it’s member: uid=username,ou=groups,dc=domain,dc=com).
 
 // Active Directory Authentication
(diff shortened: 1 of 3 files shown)


dtm7:
Tested change works for me with $config['auth_ldap_userdn' = true with our setup.

It's difficult for me to test the negative because our LDAP config requires group members to be at least prefixed with their CN. (member: uid=username, instead of just member: username).

I manually added the "uid=" prefix to the code in LdapAuthorizer.php for testing purposes:
`
$ldap_comparison = ldap_compare(
  $connection,
  $ldap_group,
  Config::get('auth_ldap_groupmemberattr', 'memberUid'),
  "uid=".$this->getMembername($username)
);
`

With this in place it looks to me like setting $config['auth_ldap_userdn'] = false works (for me) also.

Thanks!

includes/defaults.inc.php
```diff
@@ -631,6 +631,8 @@
 $config['auth_ldap_emailattr']                  = 'mail';
 $config['auth_ldap_cache_ttl'] = 300;
 // How long in seconds should ldap* module cache user information in $_SESSION
+$config['auth_ldap_userdn']                     = true;
```
laf: This changes the default for all installs.

Should be set to `false` to maintain backwards compatibility.

TheGreatDoc: Done!

PR #8974, opened 2018-08-01
Add group based permissions

remyj38:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Add a feature to allow normal users to see devices based on a devices group (faster than add hundreds of devices to the user).


murrant:
@remyj38 Thanks for working on this. I've personally avoided adding this feature until we have device groups that are not dynamic (IE you specifically add device_ids to the group and they never change).  I was thinking of allowing conversion between the two types to ease management.

I have no idea if this is a good design, but I do think that using dynamic groups for security could be a security nightmare.

Don't be discouraged, let me know your thoughts please.

remyj38:
@murrant The PR is almost ready, but I'll need some help for fixing DB tests fails.

I use dynamics groups (based on device name) and I want to delegate access to those devices to my customers. I'm definitely not sure if it's the best design to do this too, but, IMHO, without more sophisticated permission system, it's the easiest way to implement this feature.

murrant:
easy now could mean really hard later.  That is way I am hesitant.

In my mind I'd like to have non dynamic device groups where you manually add devices and they can't change or be manipulated by outside factors.  To ease management, I was thinking something like using dynamic groups to set them up initially or maybe a one way conversion from dynamic to static.

Then only allow static device groups to be used for security settings.

I'm not sure if that is too complex or not.

PR #8989, opened 2018-08-06
Option to filter availability map widget by device groups

shaheen69:
Added an option for device groups to filter devices/services displayed on the availability map to based on the selected device group.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/common/availability-map.inc.php
+++ b/html/includes/common/availability-map.inc.php
@@ -34,6 +34,8 @@ if ($config['webui']['availability_map_compact'] == 1) {
 
 $show_disabled_ignored = $widget_settings['show_disabled_and_ignored'];
 
+$filter_device_group = $widget_settings['filter_device_group'];
+
 if (defined('SHOW_SETTINGS')) {
     $common_output[] = '
     <form class="form" onsubmit="widget_settings(this); return false;">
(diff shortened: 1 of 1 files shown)


laf:
So before hand device groups was used with the setting `$config['webui']['availability_map_use_device_groups']`. It seems that this is now redundant and could be removed. Yes it would expose groups to all users but the menu bar has this anyway so I don't think it's something to worry about until we refactor auth privileges. 

You also need to fix your code formatting: https://travis-ci.com/librenms/librenms/jobs/138406893#L675

shaheen69:
Hi,

I've managed to correct the code formatting issues. I'll need to investigate further about compliance with the device group permissions as I've currently just run a query to the database to get all the device groups populated in the settings dropdown menu.

Regards,

Shaheen

laf:
> @laf as a note, the Laravel menu blade respects user permissions for the device menu.

We're not using them at the moment are we? print-menubar.inc.php is still where it's all defined + this is about device groups, which we don't hide from users anywhere at present.

laf:
> I've managed to correct the code formatting issues. I'll need to investigate further about compliance with the device group permissions as I've currently just run a query to the database to get all the device groups populated in the settings dropdown menu.

Sorry I'm not talking about that aspect. That we can live with. If you look in the file you've edited for `$config['webui']['availability_map_use_device_groups']`, I expect you can remove the check for this config value as it looks to be redundant now.

shaheen69:
When removing the check for `$config['webui']['availability_map_use_device_groups']` I found that it breaks the functionality in the menu: Overview->Maps->Availability

I think we should leave it intact as it doesn't seem to affect the widget on the dashboard.

laf:
What's the issue you came across as that config isn't used in the menu system itself.

Got a patch for what you tried?

shaheen69:
> What's the issue you came across as that config isn't used in the menu system itself.

Sorry, I wasn't clear, what I meant is...
If you go the full screen Availability Map in the located at 'Overview->Maps->Availability', it's uses the same file as the **availability-map widget** that is used on the dashboard. If I remove the `$config['webui']['availability_map_use_device_groups']` checks in the file `html/includes/common/availability-map.inc.php`, the 'device groups' filter on the full screen Availability Map no longer functions. Hence I think I should leave those checks in place on the widget, even though it doesn't affect the widget when removed.

> Got a patch for what you tried?

Not sure how to do this yet (send a patch), this is my first experience in contributing to an open source  project as well as using Git. Would that be something you could guide me on?

I also see that the widget has recently been updated, I'll resolve the conflicts for my changes and push an update shortly.

laf:
Sure thing, when you've edited the code you can run `git diff| ./pbin.sh` then pastebin the link provided

murrant:
FYI, I'm rewriting the dashboard widgets backend and this is implemented there.

murrant:
Future PR to come...

laf:
> FYI, I'm rewriting the dashboard widgets backend and this is implemented there.

@murrant I re-did the front end in v2. Would like to move that over to v1 at some stage as the tiling library was a current supported one.

murrant:
@laf most of the work I'm doing is on the backend.  So I'm not duplicating much work.

PR #8991, opened 2018-08-06
Bugfix: Correct config template for API transport

gs-kamnas:
Due to bug in the implementation of the alert transport mapping introduced in #8660  , the API transport is unable to be configured via the new interface. Only the 'get' method is displayed in the UI and the defined validation is expecting uppercase "GET" or "POST" as the method, making it impossible to configure the API transport.

This change corrects the name to option value mapping to uppercase GET and adds back the POST method option. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Api.php
+++ b/LibreNMS/Alert/Transport/Api.php
@@ -87,8 +87,8 @@ class Api extends Transport
                     'descr' => 'API Method: GET or POST',
                     'type' => 'select',
                     'options' => [
-                        'get' => 'get',
-                        'get' => 'get'
+                        'GET' => 'GET',
+                        'POST' => 'POST'
                     ]
                 ],
                 [


gs-kamnas:
@laf Can you explain where else the requirement for a lowercase method HTTP method name is? It appears that the method name is unconditionally converted to lowercase before being compared with lowercase "get".

Reference:
https://github.com/librenms/librenms/blob/66e88af8d55fa4c4aed23285687bdad0691394ad/LibreNMS/Alert/Transport/Api.php#L53

murrant:
Oh, I must have changed that in my PR.

Why is strtolower() anyway?  I think the old code used all caps.

laf:
Indeed @gs-kamnas. I was in the shower when commenting last :)

PR #9011, opened 2018-08-13
Device: Add support for Huawei iBMC

PipoCanaja:
Hello

Here is the support for iBMC server management unit of the Huawei servers. MibFile, YAMLs, tests data are included. 

It is important to add that in current version of iBMC (Huawei iBMC Management Console 3.04 and below), there is no way to set sysName online (which means that if you have more than 1 RH server, they are all sysNamed "iBMC" ) ... Current workaround is to manually snmpSet another name (same as hostname which is settable but not copied in the sysName OID ... )

Case is opened to get a proper solution from Huawei. 

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hello,

Done ;) 

Bye

includes/polling/os/ibmc.inc.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+$oids = array (
+    'version' => 'firmwareVersion.11.65.99.116.105.118.101.32.105.66.77.67',   // firmwareVersion."Active iBMC"
```
PipoCanaja: I did not manage to use the string indexed form of the OID, so I had to write it numerically. If you have a working syntax for firmwareVersion."Active iBMC", I am interested.

laf: No worries. That oid is indexed by firmwareName which is a bit silly.

PipoCanaja: I searched for a way to snmpwalk it (in the CLI) and did not even succeed in that ... Anyway, it works.

laf:
I've rebased this to try and fix the tests.

PR #9017, opened 2018-08-15
Added hiding of disabled/ignored ports in graph and device ports view.

Evil2000:
When disabling or ignoring ports in the "Ports Settings" (http://<nms_host>/device/device=<device_id>/tab=edit/section=ports/) then the ports are still shown under the Port tab for a specific device (http://<nms_host>/device/device=<device_id>/section=ports/) and in the Overall Traffic Graph.
To keep the view and graph more clear, this change only shows the ports which are not disabled or ignored.

DO NOT DELETE THIS TEXT

#### Please note

- [y] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

### DIFF
--- a/html/includes/graphs/device/bits.inc.php
+++ b/html/includes/graphs/device/bits.inc.php
@@ -4,7 +4,7 @@
 $ds_in  = 'INOCTETS';
 $ds_out = 'OUTOCTETS';
 
-foreach (dbFetchRows('SELECT * FROM `ports` WHERE `device_id` = ? AND `disabled` = 0', array($device['device_id'])) as $port) {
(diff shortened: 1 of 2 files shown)


laf:
Any reason you've put these values in quotes? You shouldn't need to.

Evil2000:
@laf You're right. I removed the qotes, so it's more clear to read,

murrant:
Hmm, if I'm not mistaken, ignore is only intended for alerting I don't think it should be used to hide things by default.

Evil2000:
@murrant Could you please post a link to the documentation where this is explained?

murrant:
https://docs.librenms.org/Support/FAQ/#what-is-the-difference-between-disable-device-and-ignore-a-device

PR #9025, opened 2018-08-16
Add basic support for IBM iSeries (AS400)

VanillaNinjaD:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
You are going to need to add test data for the new OS as described here. https://docs.librenms.org/Developing/os/Test-Units/
If you upload the snmprec, I can generate the json dump for you and add it.

includes/definitions/ibm-as400.yaml
```diff
@@ -0,0 +1,11 @@
+os: ibm-as400
```
laf: According to https://en.wikipedia.org/wiki/IBM_i this should be called ibm-i

includes/polling/os/ibm-as400.inc.php
```diff
@@ -0,0 +1,3 @@
+<?php
+$output = preg_split('/\s+/', $device['sysDescr']);
```
laf: Can you use another variable name other than `$output` to avoid conflicts anywhere else. Maybe `$ibmi_output`.

laf:
Thanks @VanillaNinjaD.

Last thing is to rename the snmprec test file. We also need a json test file, I can generate this for you but you've created this PR from your master branch. It would probably be better to close this PR and create a new one after you've done: `git checkout -b ibmi-i` then do `git push origin ibm-i` - assuming that `origin` is your librenms fork.

VanillaNinjaD:
Will do, I'm still so confused by git...

PR #9026, opened 2018-08-16
Cleaned up Trango Apex Lynx OS code and added wireless sensors

metavrs:
I removed an incorrect mib and added the correct one.  Changed initial detection to use mib instead of numeric oid. Removed trim function.  Added wireless sensor class.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Looks great.  There are several instances where you mess up the plurality of $sensors that should be fixed and this is ready.  I'm uploading the test json for you.

murrant:
Hmm, looks like the MIB is kinda broken. Underscores (_) are not allowed in oid names.  So, this made the os version not work.  If the vendor doesn't have an updated and fixed MIB, you can either revert numeric oid or change underscores to dashes in the MIB.

metavrs:
Hey murrant, any idea why this is failing?  I couldn't tell from looking at the details

laf:
I've just re-run `./scripts/save-data.php -o apex-lynx` to regenerate the data.

metavrs:
Thank you, I'll watch out for that next time

PR #9050, opened 2018-08-21
Fixing some URLs and Images in /doc/

eimann:
DO NOT DELETE THIS TEXT
I encountered a few broken links in the documentation files, especially for linked images. This has been fixed in this PR.

#### Please note

#### Testers

### DIFF
--- a/doc/Alerting/Rules.md
+++ b/doc/Alerting/Rules.md
@@ -82,4 +82,4 @@ Alert when:
 ### Alert Rules Collection
 You can also select Alert Rule from the Alerts Collection. These Alert Rules are submitted by users in the community :)
 If would like to submit your alert rules to the collection, please submit them here [Alert Rules Collection](https://github.com/librenms/librenms/blob/master/misc/alert_rules.json)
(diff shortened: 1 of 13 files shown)


kkrumm1:
Thanks for the PR but those images are loading up. Adding the /doc/ would bring the link to the image?

eimann:
Here's without the fix: https://github.com/librenms/librenms/blob/master/doc/Support/Adding-a-Device.md
Here's with the fix: https://github.com/eimann/librenms/blob/doc-fix/doc/Support/Adding-a-Device.md

laf:
> Here's without the fix: /doc/Support/Adding-a-Device.md@master
> Here's with the fix: eimann/librenms:doc/Support/Adding-a-Device.md@doc-fix

@kkrumm1 is right, this would break it for the docs.librenms.org site which is typically where people will read the docs. The perm fix for both would be to put the images within the dir of each file and reference them relatively.

eimann:
Thanks @laf. I'll give it some more work then and will close it for now.

PR #9054, opened 2018-08-21
When validation fails print links to the community site and Discord.

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/Validations/Database.php
```diff
@@ -194,7 +194,7 @@ private function checkSchema(Validator $validator)
         if (empty($schema_update)) {
             $validator->ok('Database schema correct');
         } else {
-            $result = ValidationResult::fail("We have detected that your database schema may be wrong, please report the following to us on IRC or the community site (https://t.libren.ms/5gscd):")
+            $result = ValidationResult::fail("We have detected that your database schema may be wrong.")
```
laf: The only thing with dropping this link is we could end up with loads of threads on updating permissions. I'd much more prefer to direct everyone to that one link.

PR #9027, opened 2018-08-17
New transport modules (Kayako Classic, SMSFeedback)

odvolk:
Hello! I've made a first version of Alert Transporting module for Kayako Classic Servicedesk. May be it is possible to use with Kayako OnPremise version (not sure).

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

PR #9056, opened 2018-08-22
Fixed router.db generation script for rancid 3.1+

odvolk:
Hi all! I noticed, that when you are using rancid v3.1+, router.db structure should be

`host**;**login**;**pass instead host**:**login**:**pass`

Maybe should it be fixed up?

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


odvolk:
UP: Fixed some documentation for Kayako and SMSFeedback transport modules

scripts/gen_rancid.php
```diff
@@ -50,7 +50,7 @@
         if ($devices['disabled']) {
             $status = "down";
         }
-        echo $devices['hostname'] . ':' . $rancid_map[$devices['os']] . ':' . $status . PHP_EOL;
```
laf: This will just break compatibility.

You'd need to add a config option so people can specify which version people are running.

Also, as mentioned [in](https://github.com/librenms/librenms/pull/9027), you need to split out your pull requests. docs changes for transports and a rancid change in one won't get merged. You also shouldn't make changes in your master branch. We have docs on using git, have a read.

odvolk: Thx for and idea. I'll close this pull request and make changes there first

PR #9057, opened 2018-08-22
Add Basic Support for Allworx VoIP Systems

VanillaNinjaD:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/allworx.inc.php
```diff
@@ -0,0 +1,4 @@
+<?php
+$hardware = $device['sysDescr'];
+$version = snmp_get($device, '1.3.6.1.2.1.27.1.1.4.1', '-OQv', '', '');
```
laf: You should add a license header to the files you add.

Can you please convert these two snmp_gets into multi gets - check other files for examples.

includes/definitions/allworx.yaml
```diff
@@ -0,0 +1,10 @@
+os: allworx
```
laf: allworx is the vendor name, do they not have an OS name these products run or a group name for the product line?

html/images/os/allworx.svg
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
laf: You need to remove the width/height + compress the svg.

Use https://jakearchibald.github.io/svgomg/ to do this.

includes/definitions/allworx.yaml
VanillaNinjaD: The have both the X (legacy) series and the Connect (current) series of telephone switches but they don't have an OS name and the software is unified between the models. I am happy to change it but I don't know what we'd change it to.

Any suggestions?

VanillaNinjaD: Since Allworx makes a line of Layer 3 smart switches I forgot about, I'll rename this OS to allworx_voip

VanillaNinjaD:
@laf Here is my best effort!

laf:
@VanillaNinjaD I've made a few teaks. Once it's passed tests, we'll merge in.

PR #9061, opened 2018-08-23
Added support for Teleste Luminato

jozefrebjak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


TheGreatDoc:
@jozefrebjak Check the travis CI. Seems like you uploaded a bitmap as svg.

includes/polling/os/teleste.inc.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+$version = snmp_get($device, "1.3.6.1.2.1.47.1.1.1.1.10.10", "-OQv");
```
laf: Do you have a mib for these OIDs?

You will also need to convert them to snmp multi gets (see other files as examples).

mibs/HOST-RESOURCES-MIB.mib
```diff
@@ -0,0 +1,1540 @@
+HOST-RESOURCES-MIB DEFINITIONS ::= BEGIN
```
laf: You've added a lot of standard mibs that we already have, please remove these as they aren't needed.

For any specific to this OS or vendor please upload to `mibs/teleste/` dir

includes/definitions/teleste.yaml
```diff
@@ -0,0 +1,12 @@
+os: teleste
```
laf: teleste appears to be to the vendors name rather than the OS these devices run. Please use the OS name for the product or if wasn't available then something referencing the product itself or product range would be better.

includes/polling/os/luminato.inc.php
```diff
@@ -0,0 +1,4 @@
+<?php
+$version = preg_replace('/[\r\n\"]+/', ' ', snmp_get($device, "swVersion.0", "-OQv", "TELESTE-LUMINATO-MIB"));
```
laf: Can you convert all of these snmp_gets into snmp_multi_get_oid please. Saves on SNMP queries.

Also, don't prepend Teleste to the hardware, it's supposed to just describe the model / hardware.

laf:
I've sorted out the tests so that should be fine. However the two mibs you've added reference another, `TELESTE-ROOT-MIB`, do you have this to upload to this pr please?

PR #9064, opened 2018-08-23
Database schema cleanup

murrant:
Remove several inconsistencies (mostly with data types and lengths)
Make things a bit nicer to implement with Laravel database migrations.
poller* tables had f'd up indexes
removed an unused field with screwy definition

probably needs to be tested on an older version of mysql/mariadb to make sure.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


misc/db_schema.yaml
```diff
@@ -1615,20 +1608,20 @@ users:
     - { Field: email, Type: varchar(64), 'Null': false, Extra: '' }
     - { Field: descr, Type: char(30), 'Null': false, Extra: '' }
     - { Field: level, Type: tinyint(4), 'Null': false, Extra: '', Default: '0' }
-    - { Field: can_modify_passwd, Type: tinyint(4), 'Null': false, Extra: '', Default: '1' }
-    - { Field: created_at, Type: timestamp, 'Null': false, Extra: '', Default: '1970-01-02 00:00:01' }
+    - { Field: can_modify_passwd, Type: tinyint(1), 'Null': false, Extra: '', Default: '1' }
+    - { Field: created_at, Type: timestamp, 'Null': false, Extra: '', Default: CURRENT_TIMESTAMP }
```
laf: I'll need to check this but I'm sure you can't have two CURRENT_TIMESTAMPs in one table with strict mode on

murrant: Could be, I just checked and I thought I was running in strict, but I was not.

laf:
```
260 -> 261 ...SQL Error! SQLSTATE[HY000]: General error: 1293 Incorrect table definition; there can be only one TIMESTAMP column with CURRENT_TIMESTAMP in DEFAULT or ON UPDATE clause (SQL: ALTER TABLE `users` CHANGE `created_at` `created_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ;)
  /opt/librenms/includes/sql-schema/update.php:92
```

murrant: I think only one could have the `on update CURRENT_TIMESTAMP` right?  default CURRENT_TIMESTAMP is ok?

laf: No it says you can't for default in that output

laf:
Leave the rebase on this now but I plan to merge this in Monday.

laf:
Rebased to try and get this out this morning but it's failing tests and I don't have any time to check into it.

PR #9068, opened 2018-08-23
WIP Fix peeringdb query to delete all.

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2397,7 +2397,11 @@ function cache_peeringdb()
                         }
                     }
                     $pdb_ix_peers_ids = implode(',', $peer_keep);
-                    dbDelete('pdb_ix_peers', "`pdb_ix_peers_id` NOT IN ($pdb_ix_peers_ids)");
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -2397,7 +2397,11 @@ function cache_peeringdb()
                         }
                     }
                     $pdb_ix_peers_ids = implode(',', $peer_keep);
-                    dbDelete('pdb_ix_peers', "`pdb_ix_peers_id` NOT IN ($pdb_ix_peers_ids)");
+                    if (empty($pdb_ix_peers_ids)) {
+                        dbDelete('pdb_ix_peers');
```
laf: I think this should be restricted to `ix_id`

murrant: I noticed that, the query was broken before.  Not sure they were ever deleting anything.

I can either restrict it to ix_id and do multiple deletes or collect all valid ids and do one delete per table.  I started coding the second option, but ran out of time.

murrant:
Just fixed in #9077

Issue #8416, opened 2018-03-19
Support for PBI DCH 4000P 


jozefrebjak:
**DO NOT DELETE THIS INFORMATION.**

> **Please read this information carefully.**

**GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z.** 

**You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).**

**If you have a feature request/changes please post this on our [community site](https://community.librenms.org/c/feature-requests).**

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [ ] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [ ] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [ ] Please provide **ALL** info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [ ] Please provide as much detail as possible.
- [ ] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

Please add support for PBI DCH-4000P DVB-S/S2 a multi deskrembler, MPEG-2/SD decoder.

Here is a captured discovery:
https://p.libren.ms/view/b84c794c

Poller
https://p.libren.ms/view/c59dccf7

SNMP:
https://p.libren.ms/view/5421941a

**Input status**
.1.3.6.1.4.1.1070.3.1.1.104.14.3.0 is for strenght of signal (dBm)
.1.3.6.1.4.1.1070.3.1.1.104.1.1.8.0 is for C/N (dB)
.1.3.6.1.4.1.1070.3.1.1.104.1.1.9.0 is for Eb/N0 (dB)
.1.3.6.1.4.1.1070.3.1.1.104.1.1.2.0 is for packet lenght (bytes)
.1.3.6.1.4.1.1070.3.1.1.104.1.1.3.0 is for Total Bit Rate (Mbps)
.1.3.6.1.4.1.1070.3.1.1.104.1.1.4.0 is for Valid Bit Rate (Mbps)

Component | Version
LibreNMS | 1.37-57-g69bedd0
DB Schema | 245
PHP | 7.0.22-0ubuntu0.16.04.1
[...]

PR #9072, opened 2018-08-24
WIP Added support for PBI Digital Decoder 

jozefrebjak:
There is a test in my environment. We are using multiple PBI DCH-4000P decoders.

image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


jozefrebjak:
#8416

includes/definitions/pbi.yaml
```diff
@@ -0,0 +1,10 @@
+os: pbi
+text: 'Linux'
```
laf: text should describe the OS / Product line.

jozefrebjak: @laf There is no MIB for that device, I can't find any, only that OIDs, this device is running linux, there is a tab with version and that linux os version 1b is not listed in snmpwalk.

snimka obrazovky 2018-08-25 o 16 42 51

What do you think, when are you talking about correct name ? 

I am not programmer, so when is code bad, please help me with this. I am just trying to add some devices, which we have on our network and will be really awesome, when it will be possible to monitor them. 

Thank you.

jozefrebjak: You are right, I changed os to IRD

jozefrebjak:
@laf Really thank you for help for this device. I am going to test this changes.

laf:
Sorry but it's set to Hz. The graphs should auto scale though and change the values once they get to ranges it makes sense to do this on.

Thanks for adding more support and letting me test creating a MIB :)

jozefrebjak:
OK, thank you for the info. This MIB is really good thing !

jozefrebjak:
Please help, in this line is a system-name of the device, but is there with the serial number, is there an possibility to extract only system-name ?

there is example of three devices

.1.3.6.1.4.1.1070.3.1.1.1.0 = STRING: "H6-SDI-Zemplin              9070113                 "
.1.3.6.1.4.1.1070.3.1.1.1.0 = STRING: "H10-f&h                     8520059                 "
.1.3.6.1.4.1.1070.3.1.1.1.0 = STRING: "H8s2-primy                  9240171                 "

laf:
> Please help, in this line is a system-name of the device, but is there with the serial number, is there an possibility to extract only system-name ?

You would just have to do a preg_match or explode on the space if it's static at one space.

PR #9074, opened 2018-08-24
Isolate poller and discovery modules

murrant:
An exception in one module should not interrupt the process for other modules.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
What does this protect us against?

murrant:
@laf when using eloquent/db queries if we get unexpected input and it throws an exception. (Like in the OSPF poller port_id cannot be null recently)  This will allow the polling process to continue with the remaining modules instead of stopping.  We still need to fix those issues, but the impact will be less.

PR #9077, opened 2018-08-24
Fix IN db queries

murrant:
Most were fine as they hardcoded the in into the query.
Change them all to use PDO properly.
Did not fix IRCBot, they are are all hardcoded.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2344,6 +2344,8 @@ function cache_peeringdb()
             $rand = rand(3, 30);
             echo "No cached PeeringDB data found, sleeping for $rand seconds" . PHP_EOL;
             sleep($rand);
+            $peer_keep = [];
+            $ix_keep = [];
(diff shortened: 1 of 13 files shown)


includes/functions.php
```diff
@@ -2396,11 +2398,19 @@ function cache_peeringdb()
...
-                    $pdb_ix_peers_ids = implode(',', $peer_keep);
-                    dbDelete('pdb_ix_peers', "`pdb_ix_peers_id` NOT IN ($pdb_ix_peers_ids)");
                 }
-                $pdb_ix_ids = implode(',', $keep);
-                dbDelete('pdb_ix', "`pdb_ix_id` NOT IN ($pdb_ix_ids)");
+            }
+
+            // cleanup
+            if (empty($peer_keep)) {
+                dbDelete('pdb_ix_peers');
```
laf: I'm not sure you can move this out of the loop, pdb_ix_peers is grouped by ix_id so you need to delete the peers that no longer are in that IX.

murrant: @laf we collect all ix_peers from all groups, then delete the ones we didn't see.
ix1 = 1,4,5,6
ix2 = 7,8,9,10

old code:
delete not in 1,4,5,6
delete not in 7,8,9,10

new code:
delete not in 1,4,5,6,7,8,9,10

your suggestion:
delete not in 1,4,5,6 where ix = 1
delete not in 7,8,9,10 where ix =2

Is that right?

laf: Good shout :)

PR #9082, opened 2018-08-25
Improved port polling in certain scenarios

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Small improvements to ports polling time. Tested with and without selected ports polling on Cisco IOS and Arista EOS devices.

### DIFF
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -175,8 +175,6 @@ $dot3_oids = [
     'dot3StatsDuplexStatus',
 ];
 
-$pvid_oids = 'dot1qPvid';
-
 // Query known ports and mapping table in order of discovery to make sure
 // the latest discoverd/polled port is in the mapping tables.
 $ports_mapped = get_ports_mapped($device['device_id'], true);
@@ -256,8 +254,7 @@ if ($device['os'] === 'f5' && (version_compare($device['version'], '11.2.0', '>=
                     $port_stats = snmp_get_multi($device, $extra_oids, '-OQUst', 'EtherLike-MIB', null, $port_stats);
 
                     if ($device['os'] != 'asa') {
-                        $extra_pvid_oids = "$pvid_oid.$ifIndex";
-                        $port_stats = snmp_get_multi($device, $extra_pvid_oids, '-OQUst', 'Q-BRIDGE-MIB', null, $port_stats);
+                        $port_stats = snmp_get_multi($device, "dot1qPvid.$ifIndex", '-OQUst', 'Q-BRIDGE-MIB', null, $port_stats);
                     }
                 }
             }
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -175,9 +175,7 @@ $dot3_oids = [
     'dot3StatsDuplexStatus',
 ];
 
-$pvid_oids = [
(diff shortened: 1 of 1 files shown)


laf:
@murrant I'm not 100% happy about `$walk_base = $total_port_count - $polled_port_count < 5 || $polled_port_count / $total_port_count > 0.9 ;` in ports polling.

I assume you added this in as you saw devices having issues when selected ports polling had issues? I'd like us to consider reverting this change.

Here's the impact of removing the check for me:

image

murrant:
I just put some numbers together, I would like to see something like it in there, but tweak the calculation.
I put it in there be cause certain devices took a lot longer to check all ports individually when they are doing it for every (or almost every) port.

laf:
> I just put some numbers together, I would like to see something like it in there, but tweak the calculation.
> I put it in there be cause certain devices took a lot longer to check all ports individually when they are doing it for every (or almost every) port.
> 
> Maybe 90% of the ports is too low and it should be 95% or more.

Personally for now, I'd like to remove the code - in this instance it's 100% of the ports that still pass.

At present we have 3 scenarios with this:

Selected ports polling off
Selected ports polling on
Selected ports polling on but falls back.

Users can only control the first two which means we're forcing long poll times for people who elect for selected ports polling in certain circumstances. I realise this also means the poll time may increase where option 3 happens but is no longer available.

laf:
> @laf if 100% of the ports are polled how can it be slower to run snmpget x #ports vs 1 walk? That doesn't make any sense to me.

Since when does snmp make sense.

This is repeatable across all the Arista's we run.

As for changing the base oids. We'd need to do something about `is_port_valid()` which is called before the gets to determine if the port should be even polled - this uses ifAlias, ifName, ifDescr and ifType.

murrant:
@laf funny thing about the is_port_valid() call, it doesn't use $port_stats, it uses $port (which is from the db).  Those three are unused.

laf:
@murrant Do those values get updated to the db later on because if not, a port that once fails is valid would never pass even if it should.

includes/polling/ports.inc.php
```diff
@@ -175,6 +175,10 @@
     'dot3StatsDuplexStatus',
 ];
 
+$pvid_oids = [
```
murrant: I feel like this should be inlined it is just one oid and I don't expect more to be added.

includes/polling/ports.inc.php
```diff
@@ -212,6 +216,7 @@
         $walk_base = $total_port_count - $polled_port_count < 5 || $polled_port_count / $total_port_count > 0.9 ;
 
         if ($walk_base) {
+            echo "Not enough ports for selected port polling, walking base OIDs instead\n";
```
murrant: This could be improved.  We could split out the $walk_base boolean and accurately describe why the walk is being performed.

laf: Seems pointless as I really do want to remove this code.

murrant: Let's look at that in a different PR.  This PR looks good to me, just the inline of that array + remove the implode there.

laf: I've sorted the array out but left the text as is. Seems a bit ott to set variables for both those outcomes just to be able to say which one caused the selected port polling to fail.

PR #9084, opened 2018-08-26
add a tool for working with JSON apps

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This is a handy little tool I wrote for generating snmprec and test data files for JSON apps.  Dealing with the metric stuff can be a bit annoying, especially in some cases it appears to want it in order as well. The metrics are checked for based on metric name in binary case insensitive. This outputs JSON test data using the sorting method.

Also handy for some one if they want to grab a quick look at how the metrics look flattened for writing alerts as well.


scripts/json-app-tool.php
```diff
@@ -0,0 +1,179 @@
...
+-m is handy if you want to grab a list of metrics a JSON app is returning
+   and how they data is flattned as well for when writing alert rules. -k
+   prints it in a slightly neater manner and in a manner and in tested
+   order.
+";
+    exit();
+}
+
+// make sure we have a JSON file to work with
+if (!isset($options[j])) {
```
murrant: This needs to be a string, not a constant. `$options[j]` -> `$options['j']`.  Same as everywhere.  This will cause code to fail in PHP 7.3 fyi ;)

scripts/json-app-tool.php
```diff
@@ -0,0 +1,179 @@
...
+        "1.3.6.1.4.1.8072.1.3.2.2.1.21.6.100.105.115.116.114.111|2|1\n".
+        "1.3.6.1.4.1.8072.1.3.2.2.1.21.".$oid."|2|1\n".
+        "1.3.6.1.4.1.8072.1.3.2.3.1.2.".$oid."|4x|".bin2hex($raw_json)."\n";
+    exit(0);
+}
+
+// prints the json test data file if asked to
+if (isset($options[j])) {
+    $test_data=array(
+        "discovery" => array(
```
murrant: You are missing a key here.
Should be 
```php
[
    'applications' => [
        'discovery' => [...],
        'poller' => [...]
    ]
]
```

scripts/json-app-tool.php
VVelox: Ohh! Thanks dude! That is great to know. :) Fixed.

scripts/json-app-tool.php
VVelox: Doh! Yeah. Thanks. Had to look at this for a bit before I realized what I was missing.

PR #9090, opened 2018-08-27
WIP Reduce the amount of initial data fetched by per-port polling.

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
I did rather odd test for this and set ifName,ifDescr and ifAlias all to null for a device. Even after a full discovery and poll the ports never came back.

Switching back to master worked ok.

PR #9094, opened 2018-08-27
Added ability to set a custom SQL query for alert rules.

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This would provide the flexibility some people need without us trying to shoe horn lots of things in to the Alert builder interface.

Also fixed the api for adding/editing rules.


murrant:
Tested adding via the api this:
```
{"rule_id":1333,"device_id":"-1", "name": "apitest", "builder":"{\"condition\":\"AND\",\"rules\":[{\"id\":\"devices.hostname\",\"field\":\"devices.hostname\",\"type\":\"string\",\"input\":\"text\",\"operator\":\"equal\",\"value\":\"localhost\"}],\"valid\":true}","severity": "critical","count":15,"delay":"5 m","mute":false}
```

It tried to update non-existent rule and didn't fail.
It didn't give a specific error when I tried to send old format alert rule, or json was invalid.
QueryBuilder was not converted to sql (missing toSql call)

laf:
@murrant Those issues should now be sorted.

laf:
Rebased again, should be good to go.

trs80:
This appears to have removed the transport restriction box in html/includes/modal/new_alert_rule.inc.php

Also I don't know if the recovery alerts checkbox is correctly hoooked up.

PR #9098, opened 2018-08-28
Update devices.bgpLocalAs field from varchar to int

TheGreatDoc:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Change bgpLocalAs field type in devices table from VARCHAR to INT to make it sortable


sql-schema/262.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `devices` CHANGE `bgpLocalAs` `bgpLocalAs` INT NULL DEFAULT NULL;
```
murrant: Needs to be unsigned or it won't be big enough.

PR #9104, opened 2018-08-29
Two fixes for Prometheus integration

angryp:
This is a small fix for Prometheus integration that does two things:
1. Forces metric names to be lowercase as per naming guidelines 
2. Removes slashes from the names of labels, since Prometheus does not support this symbol. 

Tested with pushgateway 0.5.2 and prometheus 2.2.1 and a monitoring environment of... 5 devices. 😃 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/prometheus.inc.php
+++ b/includes/prometheus.inc.php
@@ -31,20 +31,19 @@ function prometheus_push($device, $measurement, $tags, $fields)
                         $k = preg_replace('/([_]{2})/', '_', $k);
                         $k = strtolower($k);
                         $vals = $vals . "$k $v\n";
-                    }
-                }
+                    } 
+                } 
 
                 foreach ($tags as $t => $v) {
                     if ($v !== null) {
-                        if(strpos($v,"/") === false) {
+                        if (strpos($v, "/") === false) {
                             ;
-                        }
-                        else {
+                        } else {
                             $v = str_replace("/", "", $v); // Prometheus does not accept "/" symbol in label names, so we need to strip them out
-                        }
+                        } 
                         $promtags = $promtags . "/$t/$v";
-                    }
-                }
+                    } 
+                } 
 
                 $promurl = $config['prometheus']['url'].'/metrics/job/'.$config['prometheus']['job'].'/instance/'.$device['hostname'].$promtags;
                 $promurl = str_replace(" ", "-", $promurl); // Prometheus doesn't handle tags with spaces in url
--- a/includes/prometheus.inc.php
+++ b/includes/prometheus.inc.php
@@ -26,19 +26,29 @@ function prometheus_push($device, $measurement, $tags, $fields)
 
                 foreach ($fields as $k => $v) {
                     if ($v !== null) {
+                        $k = preg_replace('/([a-z]{1,})/', '$1_', $k); // Generally, metric naming in Prometheus is done with underscores without capital letters
+                        $k = substr($k, 0, -1);
+                        $k = preg_replace('/([_]{2})/', '_', $k);
+                        $k = strtolower($k);
                         $vals = $vals . "$k $v\n";
                     }
                 }
-                
+
                 foreach ($tags as $t => $v) {
                     if ($v !== null) {
+                        if(strpos($v,"/") === false) {
+                            ;
+                        }
+                        else {
+                            $v = str_replace("/", "", $v); // Prometheus does not accept "/" symbol in label names, so we need to strip them out
+                        }
                         $promtags = $promtags . "/$t/$v";
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


laf:
Are you saying prometheus support is currently broken?

Any examples of what this changes?

@cchance27 You added the original support, any comment on this?

angryp:
Without the fix for "/" symbol, Pushgateway will not register the metric and write to log: 
```
time="2018-08-30T08:35:03+03:00" level=debug msg="Failed to parse URL: /instance/localhost/measurement/storage/mib/hrstorage/descr//run/user/0, odd number of components in label string \"/instance/localhost/measurement/storage/mib/hrstorage/descr//run/user/0\"" source="push.go:66"
time="2018-08-30T08:35:03+03:00" level=debug msg="Failed to parse text, text format parsing error in line 1: invalid metric name" source="push.go:108"
```
With the fix, this problem is solved. 

As per the naming fix, it just becomes at least slightly compatible with the guidelines: https://prometheus.io/docs/practices/naming/

I also added a comment on metric type registering (currently all of them are "untyped"), but unfortunately lack the proper knowledge to implement this. In theory, script can take MIB description to declare metric type (like, GAUGE from MIB is mapped to Gauge type, etc.). On the other hand, documentation states they are not even used yet: 
> The Prometheus server does not yet make use of the type information and flattens all data into untyped time series. This may change in the future.
https://prometheus.io/docs/concepts/metric_types/

murrant:
You could also get type info from the rrd definition.

laf:
@angryp Please take a look at the travis output, some fixes are needed before this could be merged.

murrant:
I think it just happen to build when master was broken.  Restarted the travis ci build.

laf:
The code formatting was off, `} else {` split out on different lines and possibly more:

https://travis-ci.com/librenms/librenms/jobs/142595826#L676

angryp:
@laf Changes done according to Travis output.

includes/prometheus.inc.php
```diff
@@ -26,19 +26,28 @@ function prometheus_push($device, $measurement, $tags, $fields)
...
+                        $k = preg_replace('/([_]{2})/', '_', $k);
+                        $k = strtolower($k);
                         $vals = $vals . "$k $v\n";
                     }
                 }
-                
+
                 foreach ($tags as $t => $v) {
                     if ($v !== null) {
+                        if (strpos($v, "/") === false) {
```
laf: Does this not make more sense to just run str_replace() on $v all the time?

includes/prometheus.inc.php
```diff
@@ -26,19 +26,28 @@ function prometheus_push($device, $measurement, $tags, $fields)
 
                 foreach ($fields as $k => $v) {
                     if ($v !== null) {
+                        $k = preg_replace('/([a-z]{1,})/', '$1_', $k); // Generally, metric naming in Prometheus is done with underscores without capital letters
```
murrant: You could just use this function...

angryp: Snake case can be good, but unfortunately an array of metrics come with all capital letters like INBROADCASTPKTS, INDISCARDS, INERRORS, INMULTICASTPKTS, etc., which are then translated to i_n_b_r_o_a_d_c_a_s_t_p_k_t_s, i_n_d_i_s_c_a_r_d_s, etc. This is obviously unreadable, but this is how data is received by "prometheus_push()". If I understood correctly, it comes from includes/datastore.inc.php via data_update function, and it is called therefore from an array of places... How should I proceed?
Here is a poller debug, just in case: https://gist.github.com/angryp/c601f9cf4869b470c2e10a4affc68616

includes/prometheus.inc.php
angryp: Sure, did that. A haven't thought about a filesystem that consists of "/" itself, that is, a root filesystem. So it is needed to substitute with something. For now I added "_" symbol as a substitution. There are no associated errors produced.

angryp:
One additional problem found: some metrics can start with a number, for example - load average from ```ucd-mib``` module. 
Added additional check to swap alphabetical symbols and numeric in places.

includes/prometheus.inc.php
murrant: Snake case alone won't do the job ;-)

murrant:
FYI #9179

angryp:
Thanks @murrant, I will return to this PR once I finish with more important stuff at work (should take 2-3 days).

laf:
@angryp Any update on this?

angryp:
> 
> 
> @angryp Any update on this?

Hello @laf! Sorry for the inconveniences caused, but no, still no chance of continuing this. Unfortunately, the priority for the task has been lowered at work, so I cannot predict when I will be able to continue. Feel free to close, I will make a new request once the time comes.

murrant:
Closing for now, feel free to re-open.

PR #9106, opened 2018-08-29
Added Permission: Config Access

chasgames:
Added permissions for user level config access.

Need help testing. Is the SQL-SCHEMA file correctly made?

Settings -> Edit User -> Edit Permissions

image

Request references:
https://community.librenms.org/t/global-read-access-to-configs-adding-flag-for-this/1590
https://community.librenms.org/t/access-config-to-level-read/2610/2

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/device.inc.php
+++ b/html/pages/device.inc.php
@@ -353,8 +353,9 @@ if (device_permitted($vars['device']) || $permitted_by_port) {
             <i class="fa fa-bar-chart fa-lg icon-theme"  aria-hidden="true"></i> Alert Stats
             </a>
             </li>';
+        $can_access_configs = dbFetchCell('SELECT access_configs from users WHERE user_id = ?;', Auth::user()->user_id);
 
-        if (Auth::user()->hasGlobalAdmin()) {
+        if (Auth::user()->hasGlobalAdmin() || $can_access_configs == 1 ) {
             if (!is_array($config['rancid_configs'])) {
                 $config['rancid_configs'] = array($config['rancid_configs']);
             }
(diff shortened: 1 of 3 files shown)


laf:
I've often thought about fixing this and whilst we need to move to another user/permissions system, my initial thinking would be to create a new user type of SuperAdmin with a higher level of access which would include this.

Simple config option to allow Normal admin users access. We could automatically move users on level 10 to the new level and then allow people to edit the users.

chasgames:
Thanks Laf, so for example, bump the admin up to superuser and change admin into a normal user level where you can have custom access? 

I was thinking eventually to have something similar to Per Device Access, except Per Config Access. only allow certain device configs to certain users. That's where I thought per-user permissions might be handy place for this. As these users need network-wide view, but shouldn't have full config access to important core boxes.

chasgames:
Thanks Laf

Shall i look at making a new user level number 6

E.G Global Read & Config Access

I think ideally you could edit the user levels in the GUI and have tickboxes to check config access, billing access etc.. that might be a little too difficult for me though.

murrant:
Yes, LibreNMS is in desperate need of RBAC.  Once #8702 is merged, we could utilize an off-the-shelf Laravel RBAC.  We could incrementally add permission items after that.

laf:
Oh you want to have non-admin users have the ability to set configs. My idea was for restricting certain admins to being able to set config so no don't create another level.

Honestly, I'd prefer not to merge this PR and wait until we introduce a RBAC system.

chasgames:
In my scenario I have a few users which need to access the configs (config tab) on select devices and at the moment I have to make them administrators which is not ideal.

That's fine no problem.

murrant:
When we do RBAC, roles will be configurable above the default roles we ship.  You'll be able to mix whatever permissions you want.

laf:
Closing this as we should use RBAC rather than trying to fudge something else in to the existing system.

PR #9107, opened 2018-08-29
Fixed slack options not showing in the webui

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This will cause a loss of the options for slack (only author_name) but this isn't classed as critical to notify users imho.


murrant:
@laf no worries, I got you ;)

PR #9115, opened 2018-08-31
Fixed Total Chassis Power sensor_index for SmartAX MA5603T/MA5683T

TheGreatDoc:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

In SmartAX MA5683T/MA5603T Starting index for Per Card power starts in '0' instead '1'. '0' was also the index for Total Chassis Power.


includes/discovery/sensors/power/smartax.inc.php
```diff
@@ -22,13 +22,13 @@
...
-$pow_frame_oid = '1.3.6.1.4.1.2011.2.6.7.1.1.1.1.11';
+
 $power_frame_oid = '.1.3.6.1.4.1.2011.2.6.7.1.1.1.1.11.0';
 
 $power = snmp_get($device, $power_frame_oid, '-Ovq');
-$index = '0';
+$sensor_index = 'TotalChassisPower';
 
-discover_sensor($valid['sensor'], 'power', $device, $power_frame_oid, $index, 'smartax', 'Chassis Total', '1', '1', null, null, null, null, $power);
+discover_sensor($valid['sensor'], 'power', $device, $power_frame_oid, $sensor_index, 'smartax', 'Chassis Total', null, null, null, null, null, null, $power);
```
murrant: Why are you changing the multiplier and divisor to 0?  And why did you change the index?

TheGreatDoc: I've changed Index to sensor_index only for future understability. `index` is more interpreted as oid index, but this is sensor_index that can be (or not) oid index, and removed multiplier/divisor for clean as they are just '1', which I think is the same than passing them as null.

laf: If we have no reason to change the index we should leave it alone, same with the divisor / multiplier.

Test data is helpful so I'd revert the php changes and just submit the test data.

murrant: Well, I see that both discover_sensor() calls have the same index, so yes, one needs to be changed.

But, how we handle this in other cases it to change the type, not the index (from 'smartax' -> 'smartax-total' or something)

TheGreatDoc: Done! Just curious: Why type instead of index? Whats the difference?

PR #9122, opened 2018-09-01
Eltek Support

sippe2:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/table/mempool-edit.inc.php
+++ b/html/includes/table/mempool-edit.inc.php
@@ -6,7 +6,7 @@ $sql = " FROM `mempools` AS `M` LEFT JOIN `devices` AS `D` ON `M`.`device_id` = 
 $param[] = $device_id;
 
 if (isset($searchPhrase) && !empty($searchPhrase)) {
-    $sql .= " AND (`D`.`hostname` LIKE '%$searchPhrase%' OR `M`.`mempool_descr` LIKE '%$searchPhrase%' OR `S.`mempool_perc` LIKE '%$searchPhrase%' OR `M`.`mempool_perc_warn` LIKE '%$searchPhrase%')";
+    $sql .= " AND (`D`.`hostname` LIKE '%$searchPhrase%' OR `M`.`mempool_descr` LIKE '%$searchPhrase%' OR `S`.`mempool_perc` LIKE '%$searchPhrase%' OR `M`.`mempool_perc_warn` LIKE '%$searchPhrase%')";
 }
 
 $count_sql = "SELECT COUNT(`mempool_id`) $sql";
--- a/html/includes/table/processor-edit.inc.php
+++ b/html/includes/table/processor-edit.inc.php
@@ -6,7 +6,7 @@ $sql = " FROM `processors` AS `P` LEFT JOIN `devices` AS `D` ON `P`.`device_id` 
 $param[] = $device_id;
 
 if (isset($searchPhrase) && !empty($searchPhrase)) {
-    $sql .= " AND (`D`.`hostname` LIKE '%$searchPhrase%' OR `P`.`processor_descr` LIKE '%$searchPhrase%' OR `S.`processor_usage` LIKE '%$searchPhrase%' OR `P`.`processor_perc_warn` LIKE '%$searchPhrase%')";
+    $sql .= " AND (`D`.`hostname` LIKE '%$searchPhrase%' OR `P`.`processor_descr` LIKE '%$searchPhrase%' OR `S`.`processor_usage` LIKE '%$searchPhrase%' OR `P`.`processor_perc_warn` LIKE '%$searchPhrase%')";
 }
 
 $count_sql = "SELECT COUNT(`processor_id`) $sql";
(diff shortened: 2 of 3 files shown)


includes/polling/os/eltek.inc.php
```diff
@@ -0,0 +1,31 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2018 Mikael Sipilainen
+ * @author     Mikael Sipilainen <mikael.sipilainen@gmail.com>
+ */
+
+$version   = snmp_get($device, '.1.3.6.1.2.1.1.1.0', '-OQva', 'SNMPv2-MIB');
```
murrant: You don't need to re-fetch this, just use:
```php
$version = $device['sysDescr'];
```

murrant:
Thanks for the contribution.  You could probably update them in this file.

includes/polling/os/eltek.inc.php
sippe2: Hopefully updated version looks better now.

laf:
> Just wondering is there some clever way to get sysname, sysLocation and sysContact information from different oid? Eltek stores that kind of data to different place...

This seems like it's better to contact the vendor and ask them to implement the standard MIBs properly.

Got a link to the vendor / devices this is for please?

You also need to submit tests for this: https://docs.librenms.org/#Developing/os/Test-Units/#example-workflow

sippe2:
I'll try to make those tests. Only problem is snmpbulkwalk jams totally and snmpwalk complain 'oid not increasing' without -Cc command. So I have some understanding problems to make correct like snmprec query for snmpsim. Oid 1.3.6.1.4.1.12148.9 works perfectly but standard oid 1.2.1.1.0 not increasing after 1.2.7.5.1.1.0.0.0.0.4002...

murrant:
@sippe2 If you can just run `./scripts/collect-snmp-data.php -h HOSTNAME` and add the snmprec to this PR we can generate the test data.

If you would like to get snmpsim working, remove any currently installed snmpsim.  Then run `pip3 install snmpsim`.

murrant:
I was about to add the json file, when I noticed a couple things:
This OS should be caled webpower.
The version string should be just the version.
You should not jam unrelated data into the features string.

sippe2:
Ok.
So you suggest that I rename eltek to webpower (lowercase I suppose).
Did I understood correctly that the version should be just number like "4.11"?

Sorry about the confusion. The first time ever publishing any kind of code and I can see there is a lot of things to know. :)

sippe2:
Actually looks there is already os called webpower but different oid.

murrant:
There is indeed, how about eltek-webpower?

sippe2:
Hopefully looks better now :)

includes/polling/os/eltek-webpower.inc.php
```diff
@@ -0,0 +1,42 @@
...
+// ELTEK-DISTRIBUTED-MIB (version9) needs files SNMPv2-SMI SNMPv2-TC
+
+$output = preg_split("/[\s,]+/", $device['sysDescr']);
+        $version = $output[1];
+
+$oid = snmp_get_multi($device, 'systemSiteInfoControllerType.0 rectifierStatusType.1 batteryName.0 systemSiteInfoSystemSeriaNum.0 systemSiteInfoCustomer.0 systemSiteInfoLocation.0', '-OQUs', 'ELTEK-DISTRIBUTED-MIB');
+        $hardware = $oid[0]['systemSiteInfoControllerType'];
+        $features = 'Rectifier type: '.$oid[1]['rectifierStatusType'];
+        $features .= ', Battery name: '.$oid[0]['batteryName'];
+        $features .= ', Site Customer: '.$oid[0]['systemSiteInfoCustomer'];
```
murrant: You still need to remove these extra things from features.

sippe2:
Travis CI broken? I think it jammed during the tests...

murrant:
No, test data wasn't added yet, because of pending changes.  I've added it now.

sippe2:
Thanks for the help. :)

murrant:
Thanks for the contribution :)

murrant:
@sippe2 Hey, if these support UPS-MIB instead of doing what you have done, you just need to add `rfc1628_compat: true` to the os yaml file.

murrant:
@sippe2 looks like you have accidentally added some netagent2 changes to this PR.

sippe2:
Sigh... Hard to learn how git got works. Actually I just did not understood to create a new branch... and I just pushed new files here. Svn was easy but it was 10+ years ago. I'll try fo fix my mess. My bad. Sorry.

murrant:
https://docs.librenms.org/Developing/Using-Git/

sippe2:
Tavis CI says...

FAILURES!
Tests: 492, Assertions: 1246, Failures: 1.
Cleaning database...
snmpsimd: output at /tmp/snmpsimd.log
The command "php scripts/pre-commit.php -q -u --db --snmpsim --fail-fast" exited with 1.
0.01s$ bash -n daily.sh

Just wondering is it better to start over and create a new pull request in the new branch with all files what needed?

murrant:
Up to you, I can't help you out because it is on your master branch :)

PR #9125, opened 2018-09-01
New OS: Support for Glass Way EYDFA WDM Optical Amplifier

TheGreatDoc:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

NOTE: Those devices come all with same `sysName` so `$config['allow_duplicate_sysName'] = true;` is needed to have more than 1 working.


includes/definitions/gw-eydfa.yaml
```diff
@@ -0,0 +1,79 @@
+os: gw-eydfa
```
laf: Probably can just be `eydfa`

includes/definitions/gw-eydfa.yaml
```diff
@@ -0,0 +1,79 @@
...
+text: 'Glass Way WDM EYDFA'
+type: network
+icon: glassway
+over:
+- { graph: device_temperature, text: 'Temperatures' }
+- { graph: device_dbm, text: 'Optics' }
+mib_dir:
+- gw-eydfa
```
laf: Change the mib dir to `glassway`

TheGreatDoc:
Im stucked here. Dont know why Travis complains about sensors, i've not changed them, only did formattings fixes.

includes/polling/os/gw-eydfa.inc.php
```diff
@@ -0,0 +1,27 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2018 TheGreatDoc
+ * @author     TheGreatDoc
+ */
+
+$hardware = snmp_get($device, ".1.3.6.1.4.1.17409.1.3.3.2.2.1.4.1", "-OQv");
```
laf: Do you not have the OID names to use here? Also needs converting to multi gets.

includes/discovery/sensors/gw-eydfa.inc.php
```diff
@@ -0,0 +1,163 @@
+<?php
```
laf: I've had a quick general look at the MIB for this and it looks like this could be moved to yaml. Have you tried that?

includes/definitions/gw-eydfa.yaml
TheGreatDoc: eydfa is the kind of hardware, too generic for being an OS I think.

includes/definitions/gw-eydfa.yaml
TheGreatDoc: I'll do! 👌

includes/polling/os/gw-eydfa.inc.php
TheGreatDoc: Yes. I dont know why I used numeric.

includes/discovery/sensors/gw-eydfa.inc.php
TheGreatDoc: There are two points.
The table contains different sensor types and I dont know how to difference them.
The other is limits being this kind:
```
analogAlarmLO.13.1.3.6.1.4.1.17409.1.11.4.1.2.2
```
And yaml parser interpretes .13.1.3.6.1.4.1.17409.1.11.4.1.2.2 as index

includes/definitions/gw-eydfa.yaml
laf: No name for the firmware they run or the product range?

laf:
I'll give this a shot to move to yml.

TheGreatDoc:
I can try again with yml, but I'll need some help in how to select proper values for te table that contains different sensor types (Current and Voltage) and in the parse of the limits indexes (yml parse part of the oid as index)

includes/discovery/sensors/gw-eydfa.inc.php
murrant: It seems like that is correct.  The MIB says the index is analogParameterOID.

13 is the length of the string, the rest is the analogParameterOID value.

TheGreatDoc: Yes, I know it is the correct Index. Now the problem is:

Those are the two state sensors oids, one for each power supply:
```
.1.3.6.1.4.1.17409.1.1.2.1.4.11.1.3.6.1.4.1.17409.1.11.100.0.2
.1.3.6.1.4.1.17409.1.1.2.1.4.11.1.3.6.1.4.1.17409.1.11.101.0.2
```
which translates to:
```
NSCRTV-HFCEMS-PROPERTY-MIB::discreteAlarmState.11.1.3.6.1.4.1.17409.1.11.100.0.2
NSCRTV-HFCEMS-PROPERTY-MIB::discreteAlarmState.11.1.3.6.1.4.1.17409.1.11.101.0.2
```
I did several tests (maybe all wrong) and I cant get the 2 sensors working.

In the other part of the yaml. I dont know how to walk the table which have several sensor types.
oaPumpTable have:
```
Current in mA (divisor 1000)
Current in A (divisor 100)
Temperature in C (divisor 10)
```
Also, the Hi/Lo limits also comes like the state sensors, with:
```analogAlarmLO/analogAlarmLOLO/analogAlarmHI/analogAlarmHIHI" . length . sensor_oid```

I think all this cant be achieved with yaml, but as I said I can be wrong.

Im open to ideas ;)

**Comment by TheGreatDoc** [comment]
@laf I moved to yaml. The only I cant get working are the limits.

Yaml: https://pastebin.com/EXbNyE5y
Discovery Debug: https://pastebin.com/1dz4kcer

**Comment by TheGreatDoc** [comment]
Still stucked, even with pre-cache.

I've tried using:
```
low_limit: '{{ $analogAlarmLOLO }}'
and
high_limit: '{{ $analogAlarmHIHI.11.1.3.6.1.4.1.17409.1.11.2.0 }}'
```
With same result. No limit is set in the discovery

**Comment by TheGreatDoc** [comment]
Sometimes I'm like a newbie.
Disco debug: https://p.libren.ms/view/899379bf
Yaml: https://p.libren.ms/view/e0a73ad5

File: includes/definitions/gw-eydfa.yaml (hunk 1 above)
Comment (laf) [reply includes/definitions/gw-eydfa.yaml]: Stick with gw-eydfa then

File: includes/discovery/sensors/gw-eydfa.inc.php (hunk 4 above)
Comment (laf) [reply includes/discovery/sensors/gw-eydfa.inc.php]: Looks like you are right :( I've took a look at the data and without some work in the yaml processing code this isn't natively supported.

File: includes/discovery/sensors/gw-eydfa.inc.php [hunk 5]
```diff
@@ -0,0 +1,174 @@
...
+ * @author     TheGreatDoc
+ */
+
+$oids = snmp_walk($device, 'oaPumpTable', '-Osq', 'NSCRTV-HFCEMS-OPTICALAMPLIFIER-MIB');
+d_echo($oids."\n");
+
+if ($oids) {
+    echo 'GW EYDFA PUMP ';
+}
+
```
laf: For all of these below where you would do multiple snmp gets for the high/low values. I would do a `if (is_numeric($value)) {` check before them so that if the value isn't returned for whatever reason then we save on a bunch of gets.

TheGreatDoc:
I was using the sensor description as sensor index, so the code is the same in all sensors but I dont really care on the name of sensor index.
I've also change the $value inserted in discovery to be divided by $divisor. Before, in discovery, values were inserted without it, making them wrong until poller.

laf:
Hopefully the svg was just the last thing that needed fixing. You can use https://jakearchibald.github.io/svgomg/ to compress the file + you need to swap out width/height for a viewbox value - see existing files.

I've sorted this now so hopefully tests pass.

PR #9129, opened 2018-09-01
WebUI: Oxidized Device List: Link to config and refreshDevice Btn

PipoCanaja:
Hello all,

Here is a patch adding 2 new icons in each device of the Oxidized list. 
- One does directly link to the configuration display page. 
- The other uses the Oxidized API to queue a new scan of the configuration on the device. 

capture d ecran 2018-09-01 a 21 09 17

Concerning the code itself, I am pretty sure there is something better to do with the function : 
- function oxidized_node_update($hostname, $msg, $username = '')

I copied it from syslog-notify-oxidized.php and this is code duplication. But so far syslog-notify-oxidized.php  is not loading html/includes/functions.inc.php so putting it there would not help. 

Let me know what you suggest. 

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/forms/refresh-oxidized-node.inc.php
+++ b/html/includes/forms/refresh-oxidized-node.inc.php
@@ -15,17 +15,6 @@ header('Content-type: application/json');
 use LibreNMS\Config;
 use LibreNMS\Authentication\Auth;
 
-function oxidized_node_update($hostname, $msg, $username = '')
-{
-    $postdata = ["user" => $username, "msg" => $msg];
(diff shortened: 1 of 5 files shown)


murrant:
Why not put it in `includes\functions.php`?

PipoCanaja:
Yup. 
And then load it from syslog-notify-oxidized.php. I have nothing against :) I just wanted to ask you guys first :)

PR #9130, opened 2018-09-01
Fix edgeswitch temperatures

murrant:
There was a typo in the definition

https://community.librenms.org/t/ubuiquiti-ubnt-switches-not-showing-temperatures/5243

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


trs80:
So, per Discord chat, the num_oid wasn't changed when the oid was changed and now has a superfluous 1. at the end so temperatures are discovered correctly but not polled.

PR #9131, opened 2018-09-01
Disable test devices to block normal poller

murrant:
Still could happen, but much less likely.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Does this disable the real life device from being polled?

murrant:
No this disables the temporarily added deviced (based of the collected snmprec).

I noticed when running the python service, it would react too quickly to the test device and mess up the test data.

PR #9159, opened 2018-09-05
Updated hardware detection for VRP

jozefrebjak:
Please test this, my device os version Huawei VRP 5.170 (S5720 V200R010C00SPC600)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/os/vrp.inc.php
+++ b/includes/polling/os/vrp.inc.php
@@ -1,6 +1,6 @@
 <?php
 
-$hardware = trim(snmp_get($device, '.1.3.6.1.4.1.2011.2.33.20.1.1.1.3.0', '-OQv'), '"');
+$hardware = trim(snmp_get($device, '.1.3.6.1.4.1.2011.5.25.31.1.1.1.1.43.67108873', '-OQv'), '"');
 
 preg_match("/Version .*\n/", $device['sysDescr'], $matches);
 $version = trim(str_replace('Version ', '', $matches[0]));


PipoCanaja:
Hello,

This is probably not a good idea, this OID looks very specific to a particular device. 

Bye

murrant:
Sometimes it is appropriate to check $device['sysObjectID'], then snmpget the right value.

PipoCanaja:
Hi @murrant, @jozefrebjak 
It seems that there is absolutely no unified way to solve this. I have now 3 different families (switches, CloudEngines, AccessRouters), and 7 different devices types, and none reply their HW name to the same OID. 
Entity mib is not even a solution, some families are including some kind of virtual chassis in the Entity tree making it difficult to find the real HW you are speaking to. 
Even sysObjectID is not a flat list. 
Only option seems to identify a few recurring OIDs and poll them until one gives a non empty reply. I'll try to come with a PR in this direction.
PipoCanaja

murrant:
@PipoCanaja yes, laf and I have discussed moving os details (version, hw, etc) to discovery, but neither of us have gotten around to it.  Again, I would suggest a sysobjectid map so you only have to call snmpget once.

PipoCanaja:
I tried another approach with sysDescr. Works in the devices I tried. But I kept the OID variant for all devices I don't have that may only support it. 

https://github.com/librenms/librenms/pull/9167

PipoCanaja

laf:
@PipoCanaja It actually is probably better to do all of those as one single multi get rather than loop through an array.

murrant:
Closing this in favor of #9167

PR #9162, opened 2018-09-06
fix: edgeswitch fans

crcro:
fix: fan state 2(operational) should be mapped to 0 (no error)

tested on: EdgeSwitch 24-Port 250W, firmware 1.8.0.5106045  - latest atm

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
I don't like OS checks, perhaps set a flag on the OS or sensor instead.

crcro:
added bad_fanspeed to yaml definition and made checks based on it, hope this is a better approach

laf:
I'm against this really. Such a bodge within the code base to support something all other vendors get right.

murrant:
actually, @laf is right.  Maybe a different sensor type? Fan utilization?

crcro:
@laf is right, reverted fan rpm changes, leaved only the operational mapping

murrant:
@crcro looks like the other state generic values need to be corrected too.

murrant:
@crcro The test files don't have any data for fan state sensors.  Care to capture it with `./scripts/save-test-data.php -h HOSTNAME`

crcro:
@murrant I've runned `./scripts/save-test-data.php -o edgeswitch` after `./scripts/save-test-data.php -m sensors -o edgeswitch` and the resulted json is identical with the already existing one ...

murrant:
@crcro sorry, wrong command.  I meant `./scripts/collect-snmp-data.php -h HOSTNAME`

PR #9163, opened 2018-09-06
WebGUI Oxidized - Add author+message + refresh button

PipoCanaja:
Hello

Here is a PR for Autor and commit message of Oxidized configs. Also added the add-to-refresh-queue button.

capture d ecran 2018-09-06 a 11 55 35

Bye

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/device/showconfig.inc.php
+++ b/html/pages/device/showconfig.inc.php
@@ -260,7 +260,7 @@ if (Auth::user()->hasGlobalAdmin()) {
                 foreach ($config_versions as $version) {
                     echo '<option value="'.$version['oid'].'|'.$version['date'].'|'.$config_total.'" ';
                     if ($current_config['oid'] == $version['oid']) {
-                        $author = $version['author']['name'] . '(' . $version['author']['email'] . ')';
+                        $author = $version['author']['name'];
                         $msg = $version['message'];
                         if (isset($previous_config)) {
                             echo 'selected>+';
--- a/html/pages/device/showconfig.inc.php
+++ b/html/pages/device/showconfig.inc.php
@@ -333,28 +333,3 @@ if (Auth::user()->hasGlobalAdmin()) {
 }//end if
 
 $pagetitle[] = 'Config';
-
-echo '
-<script type="text/javascript">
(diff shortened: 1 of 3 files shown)


html/pages/device/showconfig.inc.php
```diff
@@ -318,3 +333,28 @@
 }//end if
 
 $pagetitle[] = 'Config';
+
+echo '
+<script type="text/javascript">
+    function refresh_oxidized_node(device_hostname){
```
laf: This function already exists in `html/pages/oxidized.inc.php`. You will need to move it to `html/js/librenms.js` rather than duplicate it.

PipoCanaja:
Hi @laf 

Function now moved to js/librenms.js. Also fixed a button issue in showconfig.inc.php when page was resized. 

Bye
PipoCanaja

murrant:
Mine all say the same, just the git author.   Is there a way you can map the update user to git author in oxidized?  Perhaps some docs updates are needed.

laf:
> Mine all say the same, just the git author. Is there a way you can map the update user to git author in oxidized? Perhaps some docs updates are needed.

This is an oxidized thing rather than something we need to document. 

https://github.com/ytti/oxidized/issues/940 is how you do it.

PipoCanaja:
It seems that the email address is actually not changing. It is the one that is configured int the output/git/email field of the oxidized config file. 

Does it really make sense to display it ? I am not so sure ...

Is there any situation where this email would be changing from this default value, to your knowledge ?

murrant:
I don't think oxidized would set the author info differently by itself, only if triggered externally with the the info as @laf linked.  It looks like we could add the from(email) field in the oxidized_node_update() function.

laf:
> It looks like we could add the from(email) field in the oxidized_node_update() function.

I'm not aware of any syslog from devices that would contain an email address. They normally just contain the logged in user.

You could drop the email address though.

murrant:
That's true I was more thinking of getting the email from auth, but that won't be accurate, so probably best to drop it from the display.

PipoCanaja:
OK, you did it first ;)

PR #9165, opened 2018-09-07
Added BGP Peer Descriptions

TheGreatDoc:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Added BGP Peer Description for all of us that does multiple sessions with same transit/peer asn.

To edit Peer description: Device -> Edit -> Routing tab


html/pages/device/edit/routing.inc.php
```diff
@@ -0,0 +1,85 @@
...
+            var data = $(this).val();
+            var $this = $(this);
+            $.ajax({
+                type: 'POST',
+                url: 'ajax_form.php',
+                data: {type: "routing-update", device_id: device_id, data: data, routing_id: routing_id},
+                dataType: "json",
+                success: function (data) {
+                    if (data.status == 'ok') {
+                        $this.closest('.form-group').addClass('has-success');
```
laf: I know this is used elsewhere but I've personally been moving more towards using the toastr method to display the noticications.

If we carry on the way you've done it here then we need to make the response more prominent by potentially adding a tick or cross as per:https://getbootstrap.com/docs/3.3/css/#forms-control-validation

TheGreatDoc: Tick/cross added. 
I do like the toastr method. Is there any implementation done to take as example? I could try to do that in another PR.

TheGreatDoc: Done! Cant be more easy

PR #9167, opened 2018-09-07
VRP: improving extraction of HW and Version

PipoCanaja:
As discussed in https://github.com/librenms/librenms/pull/9159

This new PR tries to extract as much as possible from SysDescr. If it fails, it uses OIDs (starting by the one that is currently use in existing code).

-> On devices that sysDescr matching the regexp, we'll have a speed improvement (no more snmpget) to fill the value
-> On devices that do not match the regexp, if the device uses the 1st OID, no speed difference compared to today. 
-> On devices that do not match sysDescr, nor the 1st OID, then we may improve the information displayed but at the cost of more OID polled at each poller occurence.

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/vrp.inc.php
```diff
@@ -1,6 +1,37 @@
...
+if (!empty($matches[1])) {
+    $hardware = "Huawei " . trim($matches[1]);
+} else {
+    // The HW cannot be extracted from sysdescr, let's try OIDs
+    $oidList[] = '.1.3.6.1.4.1.2011.2.33.20.1.1.1.3.0';
+    $oidList[] = '.1.3.6.1.4.1.2011.5.25.188.1.4.0';
+    $oidList[] = '.1.3.6.1.4.1.2011.5.25.183.1.25.1.5.1';
+    $oidList[] = '.1.3.6.1.4.1.2011.5.25.31.6.5.0';
+    foreach ($oidList as $oid) {
+        $hardware_tmp = trim(snmp_get($device, $oid, '-OQv'), '"');
```
murrant: you don't need to trim quotes.

PipoCanaja: Cleaned, as well as d_echo statements.

jozefrebjak:
Worked for me.

includes/polling/os/vrp.inc.php
```diff
@@ -1,6 +1,33 @@
...
+
+if (!empty($matches[2])) {
+    $version .= " (" . trim($matches[2]) . ")";
+}
+
+if (!empty($matches[1])) {
+    $hardware = "Huawei " . trim($matches[1]);
+} else {
+    // The HW cannot be extracted from sysdescr, let's try OIDs
+    $oidList[] = '.1.3.6.1.4.1.2011.2.33.20.1.1.1.3.0';
```
murrant: Don't declare an array like this, do it like this:
```php
oidList = [
    '.1.3.6.1.4.1.2011.2.33.20.1.1.1.3.0',
    '.1.3.6.1.4.1.2011.5.25.188.1.4.0',
    '.1.3.6.1.4.1.2011.5.25.183.1.25.1.5.1',
    '.1.3.6.1.4.1.2011.5.25.31.6.5.0',
];

Comment (murrant) [reply includes/polling/os/vrp.inc.php]: They do indeed:
```
HUAWEI-SRM-MIB::hw8060FrameDesc
HUAWEI-DEVICE-EXT-MIB::hwProductName
HUAWEI-STACK-MIB::hwStackDeviceType
HUAWEI-ENTITY-EXTENT-MIB::hwEntitySystemModel
```

PipoCanaja: Hi,
I did change the order, and resolve as much OIDs as possible considering the MIB Files we have. It is not as nice as your list, but it is improved. 
I also now take SNMP value first and sysDescr later. The idea is that SNMP is most of the time more detailed than sysDescr. 
Bye

murrant:
You need to fix the whitespace formating issue mentioned in travis output.
Also, test data needs to be updated as you have changed some data.
`./scripts/save-test-data.php -o vrp` (if you don't have snmpsim, you can install it with `pip3 install snmpsim`)

PipoCanaja:
Seems that this time the checks went good.

PR #9169, opened 2018-09-07
device: Initial support for CXR-Networks Terminal Server

PipoCanaja:
Hello

Support for CXR TerminalServers, tested on TS324. 

Tests data is attached. 

PipoCanaja
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/cxr-networks.inc.php
```diff
@@ -0,0 +1,3 @@
+<?php
+$hardware = trim(snmp_get($device, '.1.3.6.1.4.1.1425.1040.1.1.0', '-OQv'), '"');
```
laf: You don't need trim with these + can you convert to a multi get please.

includes/polling/ports/cxr-networks.inc.php
```diff
@@ -0,0 +1,55 @@
...
+    $curIfIndex = $offset + $index;
+    $port_stats[$curIfIndex]['ifDescr']="SerialPort$index";
+    $port_stats[$curIfIndex]['ifType']='rs232'; //rs232
+    $port_stats[$curIfIndex]['ifName']="Serial$index";
+    $port_stats[$curIfIndex]['ifAlias']="Port $index, " . $serialport_stats['terminalType'] . ", " . $serialport_stats['mode'] . ", " . $serialport_stats['baudRate']." ".$serialport_stats['nbParStop'];
+    if ($serialport_stats['aliasIpAddress'] != "0.0.0.0") {
+        $port_stats[$curIfIndex]['ifAlias'] .= ", Alias IP: " . $serialport_stats['aliasIpAddress'] . ":".$serialport_stats['tcpPort'];
+    }
+    if ($serialport_stats['remoteIpAddress'] != "0.0.0.0") {
+        $port_stats[$curIfIndex]['ifAlias'] .= ", Remote IP: " . $serialport_stats['remoteIpAddress'] . ":".$serialport_stats['remoteTcpPort'];
```
laf: Do you mean to override ifAlias here?

includes/polling/ports/cxr-networks.inc.php
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    $port_stats[$curIfIndex]['ifInOctets']=$serialport_stats['bytesReceiveFromV24'];
+    $port_stats[$curIfIndex]['ifOutOctets']=$serialport_stats['bytesSendToV24'];
+    $port_stats[$curIfIndex]['ifSpeed']=$serialport_stats['baudRate'];
+    $port_stats[$curIfIndex]['ifSpeed']=preg_replace("/[^0-9.]/", '', $port_stats[1000+$index]['ifSpeed']);
```
laf: You can merge this line with the one above

includes/definitions/cxr-networks.yaml
```diff
@@ -0,0 +1,12 @@
+os: cxr-networks
```
laf: Seems this is the vendors name. If they don't have a name for the firmware / os they run then please use something details the product line rather than just the vendor name.

includes/polling/ports/cxr-networks.inc.php
laf: I think this will lead to long ifaliases and you’ll need to shorten it down

includes/definitions/cxr-networks.yaml
laf: cxr-ts probably better

laf: cxr-ts probably better

includes/polling/ports/cxr-networks.inc.php
PipoCanaja: The MIB definition says 255 chars. We should be far lower than that. Any lower limit than that in mind ?

includes/definitions/cxr-networks.yaml
laf: Shorter would be better :)

PipoCanaja: Will do ... tomorrow ;)

PipoCanaja: Done ;)

PR #9177, opened 2018-09-10
Fixing Sentry 3 and 4 Temperature Sensors

sjtarik:
Currently Sentry 3 and 4 sensors not able to read temperature values, because snmp get options are not set correctly returning non-numerical values while code expects otherwise. Also current code only reads first two sensor values where there can be multiple environment sensors, especially with external environment sensors starting at index 5. Also discovered low/high limit values are not converted to celcius if user functions passed to discover sensor function so manually converting these values.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/discovery/sensors/temperature/sentry3.inc.php
```diff
@@ -1,5 +1,4 @@
 <?php
-
 $oids = snmp_walk($device, 'tempHumidSensorTempValue', '-Osqn', 'Sentry3-MIB');
```
murrant: There are a lot of snmp_gets here, it is going to be slow.

Try this, the dd() is to see the data...
```php
$data = snmpwalk_group($device, 'tempHumidSensorTable', 'Sentry3-MIB');
dd($data);
```

includes/discovery/sensors/temperature/sentry4.inc.php
```diff
@@ -22,32 +22,36 @@
 $divisor    = '10';
```
murrant: For Sentry4, you'll need a couple walks:

```
st4TempSensorConfigTable or st4TempSensorEventConfigTable not sure
st4TempSensorValue
```

includes/discovery/sensors/temperature/sentry4.inc.php
```diff
@@ -22,32 +22,36 @@
...
 $multiplier = '1';
 if ($oids) {
+    echo 'ServerTech Sentry4 Temperature ';
     $sentry_temp_scale = snmp_get($device, 'st4TempSensorScale.0', '-Ovq', 'Sentry4-MIB');
-    if ($sentry_temp_scale == 'fahrenheit') {
-        $user_func = 'convert_to_celsius';
-    } else {
-        $user_func = null;
```
murrant: Why did you delete this?  It is required if the data outputs farenheit.

sjtarik: In my librenms instance, limit values not converted although I passed the user_func to discover_sensor. So I need to convert low/high limit values manually. Just to be consistent I convert current data value manually too. Also passing convert_to_celsius function to discover_sensor does not work properly but fahrenheit_to_celcius works. I think these two bugs should be addressed separately.

includes/discovery/sensors/temperature/sentry3.inc.php
sjtarik: will update the diff shortly.

sjtarik:
@murrant I think current Travis CI is broken but not relevant to my commit. I checked the issues reported by codeclimate but they are not really addressable. Would you review this CR with these exceptions?

murrant:
Codeclimate is still being tuned, it is not a required check.

includes/discovery/sensors/pre-cache/sentry4.inc.php
```diff
@@ -32,6 +32,15 @@
...
 $pre_cache['sentry4_humid'] = snmpwalk_cache_oid($device, 'st4TempSensorEventConfigTable', $pre_cache['sentry4_humid'], 'Sentry4-MIB');
 
+echo 'st4TempSensorConfigTable ';
+$pre_cache['sentry4_temp'] = snmpwalk_cache_oid($device, 'st4TempSensorConfigTable', array(), 'Sentry4-MIB');
+
+echo 'st4TempSensorMonitorTable ';
+$pre_cache['sentry4_temp'] = snmpwalk_cache_oid($device, 'st4TempSensorMonitorTable', $pre_cache['sentry4_temp'], 'Sentry4-MIB');
+
+echo 'st4TempSensorEventConfigTable ';
+$pre_cache['sentry4_temp'] = snmpwalk_cache_oid($device, 'st4TempSensorEventConfigTable', $pre_cache['sentry4_temp'], 'Sentry4-MIB');
```
murrant: This table is walked twice.

murrant:
Looks great overall, just a duplicate walk.

We need test data for this to make sure it works right and doesn't get broken in the future.  https://docs.librenms.org/#Developing/os/Test-Units/

sjtarik:
@murrant I realize this is a new device os, so I want to properly follow the guide you shared and add other items too for device discovery, poller profile etc not just test data in a separate pull request. If everything looks fine, can you merge this one? I will send the discovery pull request by next Wed.

laf:
> @murrant I want to properly follow the guide you shared and add other items too for device discovery, poller profile etc not just test data in a separate pull request. If everything looks fine, can you merge this one? I will send the discovery pull request by next Wed.

We really do need the test data before we can merge this.

sjtarik:
@laf @murrant I have a few questions. 
1. Do I just need to upload snmprec and json files?

2. There are sentry snmprec files for sentry3 and sentry4 as sentryX_smart.snmprec and sentryX_switched.snmprec although OS names are sentry3 and sentry4. Should I just create new snmprec entries with different names or should I name them after os names as sentry3.snmprec

3.  I am running into issues collecting the snmp data, some oid's are missing in snmprec and json test file although they are in sensors table and updated fine on webpage too during discovery and polling. Even in collect php script debug I can see OIDs for these are collected fine but not in output snmprec file. Also for sentry3 snmprec file not generated at all although command line says it was generated. What is the best way to troubleshoot?

sjtarik:
> 
> What command are you running to generate the snmprec file?

I am using collect-snmp-data script as follows:

```
./scripts/collect-snmp-data.php -h 475 -n -d

Updated snmprec data /opt/librenms/tests/snmpsim/sentry4.snmprec
Verify this file does not contain any private data before submitting!
```
And for sentry4, snmprec files are being created with missing OID values that are polled, discovered, stored in database successfully by modules.

For sentry3, snmprec file not generated although output says otherwise.

```
Updated snmprec data /opt/librenms/tests/snmpsim/sentry3.snmprec

Verify this file does not contain any private data before submitting!
```

sjtarik:
@laf @murrant I just added the snmprec and test json files for sentry3 and sentry4_switched8i variant and tests are passing. Could you please review?

I run into some issues as mentioned above comment but when I pass a file name to collect-data script it generated the snmprec files with right content. For save data script I had to do a change to honor -f --file parameter. In current shape, it does not honor it but after you merge it I will just send another pull request to fix it as below if you like?

```
try {
    $no_save = isset($options['n']) || isset($options['no-save']);
    foreach ($os_list as $full_os_name => $parts) {
        list($target_os, $target_variant) = $parts;
        echo "OS: $target_os\n";
        echo "Module: $modules_input\n";
        if ($target_variant) {
            echo "Variant: $target_variant\n";
        }
        echo PHP_EOL;

        update_os_cache(true); // Force update of OS Cache
        $tester = new ModuleTestHelper($modules, $target_os, $target_variant);
        if (isset($options['f'])) {
            $tester->setJsonSavePath($output_file);
        }
        $test_data = $tester->generateTestData($snmpsim, $no_save);

        if ($no_save) {
            print_r($test_data);
        }
    }
} catch (InvalidModuleException $e) {
    echo $e->getMessage() . PHP_EOL;
}
```

laf:
@sjtarik Thanks for the first pull request.

image

sjtarik:
thanks @laf, I will keep it coming :) I think next one would be on save data script and poller profiles for sentry devices (removing bgp/ospf checks from pdu polling).

PR #9193, opened 2018-09-12
Updated example for new alerting engine

pmusolino-rms:
Using the example of port utilization caused errors when trying to use it with the new alert system.  Putting variables in curly braces and removing php calls resolved it.  Updating docs to reflect this.

Also removed examples with hash tags in the body of the template as pasting this in will cause syntax errors by the alerts script.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Alerting/Templates.md
+++ b/doc/Alerting/Templates.md
@@ -116,7 +116,6 @@ Unique-ID: {{ $alert->uid }}
 Rule: @if ($alert->name) {{ $alert->name }} @else {{ $alert->rule }} @endif
 @if ($alert->faults) Faults:
 @foreach ($alert->faults as $key => $value)
-  #{{ $key }}: {{ $value['string'] }}
 @endforeach
 @endif
 Alert sent to:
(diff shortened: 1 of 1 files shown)


doc/Alerting/Templates.md
```diff
@@ -281,7 +280,6 @@ Alert-ID: {{ $alert->id }} <br>
 Rule: @if ($alert->name) {{ $alert->name }} @else {{ $alert->rule }} @endif <br>
 @if ($alert->faults) Faults:
 @foreach ($alert->faults as $key => $value)
-#{{ $key }}: {{ $value['string'] }}<br>
```
murrant: Why did you remove this one?

doc/Alerting/Templates.md
```diff
@@ -116,7 +116,6 @@ Unique-ID: {{ $alert->uid }}
 Rule: @if ($alert->name) {{ $alert->name }} @else {{ $alert->rule }} @endif
 @if ($alert->faults) Faults:
 @foreach ($alert->faults as $key => $value)
-  #{{ $key }}: {{ $value['string'] }}
```
murrant: Why did you remove this?

doc/Alerting/Templates.md
pmusolino-rms: The lines starting with hash tags cause the parser to throw a syntax error.  If they are supposed to be comments and you would like to keep the line then I can amend it to use the {{-- comment --}} which I think is the laravel comment formatting.

doc/Alerting/Templates.md
pmusolino-rms: The lines starting with hash tags cause the parser to throw a syntax error.  If they are supposed to be comments and you would like to keep the line then I can amend it to use the {{-- comment --}} which I think is the laravel comment formatting.

murrant: I think the hash is supposed to be a literal hash in the result.  No special meaning.
Really odd that it would cause a syntax error.

PR #9196, opened 2018-09-13
Allow trusted proxy via APP_TRUSTED_PROXIES

murrant:
Set to '*' by default to emulate legacy behavior.
Set up doc describing environment variables

I could be convinced to set the default to null instead and require all reverse proxy users to set APP_TRUSTED_PROXIES.

Pulls in new requirements, you will get an error if composer cannot install the dependency automatically via `daily.sh`. If testing run `./scripts/composer_wrapper.php install --no-dev`.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Util/Env.php
+++ b/LibreNMS/Util/Env.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * Env.php
+ *
+ * -Description-
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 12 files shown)


sorano:
Just wanted to note I tested this on my installation (LibreNMS as http backend behind HAProxy) and it works for me.

ivankisic:
Tested on the latest LibreNMS with HTTPD and HAProxy, works fine.

murrant:
Thanks for the feedback.

spinza:
This did not fix it for me.  The login form still produces http urls which it should not.  The logo on the login form for example is being fetched from http instead of https.  Also the form action is http not https.

spinza:
This does not fix the problem for me.  The login form URLs revert to http instead of https.

PR #9212, opened 2018-09-16
Disable used rules in template map for select them

TheGreatDoc:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you map 2 templates to the same rule, it will always use the one with lower id.
With this, you wont be able to map a second template to the same rule. It will disable the option and tell you what template is already mapped to that rule.

### DIFF
--- a/html/includes/forms/attach-alert-template.inc.php
+++ b/html/includes/forms/attach-alert-template.inc.php
@@ -14,31 +14,50 @@
 
 use LibreNMS\Authentication\LegacyAuth;
 
-header('Content-type: text/plain');
+header('Content-type: application/json');
 
 if (!LegacyAuth::user()->hasGlobalAdmin()) {
-    die('ERROR: You need to be admin');
+    $response = array(
+        'status'  => 'error',
+        'message' => 'Need to be admin',
+    );
+    echo _json_encode($response);
+    exit;
 }
 
+$status = 'error';
+$message = 'Ops. Something happened. Enable debug and check librenms.log';
+set_debug(true);
 if (!is_numeric($_POST['template_id'])) {
-    echo 'ERROR: No template selected';
-    exit;
+    $message = 'ERROR: No template selected';
 } else {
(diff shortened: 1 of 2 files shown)


TheGreatDoc:
Thanks @laf. You actually can remove the rule clicking in the upper empty option. 
But what I noticed is, as <option>'s are loaded in /templates/ if you de-select one rule to use in another template, you must reload.

laf:
> Thanks @laf. You actually can remove the rule clicking in the upper empty option.
> But what I noticed is, as <option>'s are loaded in /templates/ if you de-select one rule to use in another template, you must reload.

Probably worth changing the blank one to be `<option>--Clear Rules--</option>`.

To do this properly you'd need to return this data in the json response then it wouldn't need a page reload.

TheGreatDoc:
I think im done. Only needs some debug clean up.

Please, test it now and let me know if you like it.

laf:
Seems that this now allows you to allocate a second template to a rule, says it's ok but doesn't actually do it. Would be much better to warn people that it didn't work or stop people doing it like you did before.

TheGreatDoc:
Reverted.

laf:
I've done a small update when editing a template that the associated rules for that template are selectable again so you can continue to select the existing rule + others.

TheGreatDoc:
@laf you did the same I reverted in [b36d591faf251f00b06080985aed26d330864a85](https://github.com/librenms/librenms/pull/9212/commits/b36d591faf251f00b06080985aed26d330864a85)

And now, as you said before, if you open a mapping (removing the disabled attr from that options), then cancel and open another mapping will have the other mapping options enabled, making it possible to select that alert rules.

I was thinking in making another ajax call on "Cancel" button to disable selected again.

laf:
> To do this properly you'd need to return this data in the json response then it wouldn't need a page reload.

^^ You need to do it that way. At present you've got a combination of loading data into the modal which is loaded on the initial page load then a call to ajax for which ones are already selected.

It really needs all doing over ajax.

PR #9214, opened 2018-09-16
Speed up testing based on changed files

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This should work on the cli as well as from travis - I hope.

### DIFF
--- a/scripts/pre-commit.php
+++ b/scripts/pre-commit.php
@@ -124,7 +124,7 @@ if ($map['php'] === 0) {
 }
 
 // If we have no php files and no OS' found then also skip unit checks.
-if ($map['php'] === 0 && empty($map['os'])) {
+if ($map['php'] === 0 && empty($map['os']) && !$os) {
     putenv('SKIP_UNIT_CHECK=1');
 }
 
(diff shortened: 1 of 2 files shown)


laf:
`Travis CI - Pull Request Successful in 3m — Build Passed`

Should have done it quicker but I'll take that.

scripts/pre-commit.php
```diff
@@ -80,15 +117,45 @@
...
 foreach (array_keys($options) as $opt) {
     $ret = 0;
     if ($opt == 'l' || $opt == 'lint') {
         $ret = run_check('lint', $passthru, $command_only);
     } elseif ($opt == 's' || $opt == 'style') {
         $ret = run_check('style', $passthru, $command_only);
     } elseif ($opt == 'u' || $opt == 'unit') {
-        $ret = run_check('unit', $passthru, $command_only, compact('fail_fast', 'os', 'module'));
+        if (!empty($map['os']) && $map['php'] === 0) {
+            foreach ($map['os'] as $os) {
```
murrant: Instead of doing a loop, you can provide a regex in the os key:

```php
    if ($options['os']) {
        $phpunit_cmd .= " --group os --filter \"@{$options['os']}.*\"";
    }
```

laf: Well spotted. See if the update I've just done works.

PR #9216, opened 2018-09-17
AD Authorization fixes

murrant:
Remove mres() and $_SESSION usage.
Remove broken addUser function and use Mysql addUser.
Extract common AD auth code to ADUtils

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
FYI, I tested both ad and ad-auth.  I can test the "external" auths by hardcoding a username, except for sso. Also, I now have radius and ldap servers (and the work AD).

TheMysteriousX:
Works with SSO fine.

murrant:
@TheMysteriousX thanks for testing :) Definitely let us know if we break sso, as we can't fully test that.
Good to remove a lot of duplicate code with this one and there was a lot of instances where the originally copied code was out of sync. So hopefully squashed some bugs too.

PR #9218, opened 2018-09-17
validate.php: Updated file permissions check

murrant:
Always post generic full fix commands.  Use chmod instead of setfacl in case the filesystem doesn't support acls.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/Validations/User.php
```diff
@@ -63,6 +63,12 @@ public function validate(Validator $validator)
         // Let's test the user configured if we have it
         if (Config::has('user')) {
             $dir = Config::get('install_dir');
+
+            // generic fix
+            $fix = "chown -R $lnms_username:$lnms_groupname $dir\n" .
```
laf: Maybe prepend this with: `As the root user` or add `sudo` before each file.

PR #9222, opened 2018-09-18
Cisco: change notKnown status to unknown and not warning

FTBZ:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

------------------------------------

The cRFStatusPeerUnitState is used for the VSS Peer Status, problem is that a lot of devices return a warning because the VSS Peer State is in warning when the status is "notKnown".

But a notKnown status should be categorized as LibreNMS "Unknown" and gray but not as a "Warning" and yellow status.

### DIFF
--- a/includes/discovery/sensors/state/cisco.inc.php
+++ b/includes/discovery/sensors/state/cisco.inc.php
@@ -50,7 +50,7 @@ foreach ($tables as $tablevalue) {
         if ($state_index_id !== null) {
             if ($state_name == 'cRFStatusUnitState' || $state_name == 'cRFStatusPeerUnitState') {
                 $states = array(
-                    array($state_index_id,'notKnown',0,1,1) ,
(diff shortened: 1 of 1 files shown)


laf:
Looks like the updated pre-commit.php hasn't ran unit tests when it should. I'll take a look.

murrant:
Also, because this uses the old style state sensor declaration, it will not update the database.
You need to update to this yaml or the php style here: https://docs.librenms.org/Developing/Sensor-State-Support/

murrant:
@laf it ran 993 tests, which is all of them.

laf:
> @laf it ran 993 tests, which is all of them.

@murrant Ah it's moved to 7.2 now hasn't it. Forgot about that.

laf:
@FTBZ Any response to murrants comment on this?

PR #9225, opened 2018-09-18
VRP: Discovery improvement for optical power + Mib upgrade

PipoCanaja:
Better sensor usage (to overcome bad device answers on some OIDs). 
Some Cloud Engine devices are replying with dBm (divisor 100) and switches are replying with uW (positive value, and should use user_func for conversion).
OIDs :  hwEntityOpticalRxPower and hwEntityOpticalTxPower

This yaml should handle the case better, by using another OID in case hwEntityOpticalRxPower and hwEntityOpticalTxPower are not positive. This OID is in the same table, and the cloud engines I tested them with are now OK (instead of having NaN value, received from the uw_to_dbm() function used with a negative value ... 

And more up to date versions of the Huawei MIB files

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


jozefrebjak:
I tested this pull on my Huawei S5720-32X-EI-24S-AC and it is looks OK

PipoCanaja:
OK, test data is now good. The PR is ready for review.

PR #9227, opened 2018-09-19
NAC Polling (Network Access Control)

joseaugustocardoso:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

screen shot 2018-09-18 at 21 02 03

This code check the authentication status on Cisco switch ports.
When enable the NAC on config fle, sill apear on device page a new tab called NAC, and inside that, a grid with all authentication status.

Config to enable NAC:
$config['enable_ports_nac'] == 1

Grid columns:

Port
MAC Address
IP Address
Authorization
Domain (Data / Voice)
Mode
Username
Time Out
Time Left
Authentication
Method (802.1x/MAB)


includes/polling/ports.inc.php
```diff
@@ -363,6 +363,54 @@
...
+            $port_index_nac = substr($PortAuthSessionEntryNAC, 0, strpos($PortAuthSessionEntryNAC, "."));
+            $port_auth_id_nac = strstr($PortAuthSessionEntryNAC, "'" );
+            $port_auth_id_nac = substr($port_auth_id_nac, 0, -1 );
+            $port_auth_id_nac = substr($port_auth_id_nac, 1);
+            $dbreult = dbFetchRow('SELECT * FROM `ports_nac` WHERE `auth_id` = ?' , $port_auth_id_nac);
+            if ($dbreult == null){
+                echo "Not found\n";
+                dbInsert(array('auth_id' => $port_auth_id_nac), 'ports_nac');
+            }
+            $IPHextpDec = explode(' ',$PortAuthSessionEntryParameters['cafSessionClientAddress']);
```
murrant: Try IP::fromHextString() here

includes/polling/ports.inc.php
```diff
@@ -363,6 +363,54 @@
     }
 }
 
+//NAC Polling
+if ($config['enable_ports_nac']) {
```
murrant: I think this should be it's own module instead of a port of the ports poller.

murrant:
Also, if you create a new module, you might want to add it to module_tables.yaml.  Then add some test data for the module. https://docs.librenms.org/Developing/os/Test-Units/

laf:
Also travis will show you a number of code formatting updates that need to happen.

+ You're using `dbQuery()` for updates and things, please use the specific functions for those like `dbUpdate()`.

includes/polling/ports.inc.php
joseaugustocardoso: It was my first idea, but discussing on Discord, laf said something about use the actual port structure.

laf08/30/2018
So that might be something better to be part of the existing ports polling

Jose Augusto Brazil [<o>]08/30/2018
yes, could be.. but if I do that, I think will not possible disable this polling if somebody doesn't use NAC, and maybe will increase the polling time... or not?

laf08/30/2018
That depends on how much additional information you need to collect but that can be controlled by a config option

joseaugustocardoso:
> Also travis will show you a number of code formatting updates that need to happen.
> 
> * You're using `dbQuery()` for updates and things, please use the specific functions for those like `dbUpdate()`.

I tried, and on some cases I used. The problem is the function expect receive an array and my data in on a simple variable string. I tried somethings but w/o success  :(

murrant:
If it is a string you simply need to wrap it in an array.

$string -> [$string]

includes/polling/ports.inc.php
murrant: `use LibreNMS\Util\IP;`

murrant: You will also need to surround it with a try/catch block in case it gets invalid input.  Putting all the code that relies on it returning a correct IP inside the try/catch.

joseaugustocardoso:
> > Also travis will show you a number of code formatting updates that need to happen.
> > 
> > * You're using `dbQuery()` for updates and things, please use the specific functions for those like `dbUpdate()`.
> 
> I tried, and on some cases I used. The problem is the function expect receive an array and my data in on a simple variable string. I tried somethings but w/o success :(

Done. Thank you for advising.

joseaugustocardoso:
> [CLA assistant check](https://cla-assistant.io/librenms/librenms?pullRequest=9227) 
> Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/librenms/librenms?pullRequest=9227) before we can accept your contribution.
> 
> **Jose Augusto Cardoso** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/librenms/librenms?pullRequest=9227) it.

I already accepted the CLA

includes/polling/ports.inc.php
```diff
@@ -380,17 +380,17 @@
...
-            dbQuery("UPDATE `ports_nac` SET `PortAuthSessionUserName` = '".$PortAuthSessionEntryParameters['cafSessionAuthUserName']."' WHERE `ports_nac`.`auth_id` = '".$port_auth_id_nac."';");
-            dbQuery("UPDATE `ports_nac` SET `PortAuthSessionAuthzBy` = '".$PortAuthSessionEntryParameters['cafSessionAuthorizedBy']."' WHERE `ports_nac`.`auth_id` = '".$port_auth_id_nac."';");
-            dbQuery("UPDATE `ports_nac` SET `PortAuthSessionTimeOut` = '".$PortAuthSessionEntryParameters['cafSessionTimeout']."' WHERE `ports_nac`.`auth_id` = '".$port_auth_id_nac."';");
-            dbQuery("UPDATE `ports_nac` SET `PortAuthSessionTimeLeft` = '".$PortAuthSessionEntryParameters['cafSessionTimeLeft']."' WHERE `ports_nac`.`auth_id` = '".$port_auth_id_nac."';");
-            dbQuery("UPDATE `ports_nac` SET `device_id` = '".$device['device_id']."' WHERE `ports_nac`.`auth_id` = '".$port_auth_id_nac."';");
+            dbUpdate(array('port_index' => $port_index_nac), 'ports_nac', '`auth_id` = ?', array($port_auth_id_nac));
```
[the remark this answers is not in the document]
joseaugustocardoso: Done. Thank you for advising.

PipoCanaja:
Hi,
I can't test it as I don't have anymore Cisco devices doing dot1X but will be happy to extend it to Huawei devices as soon as this one is commited ! 
So if there is a choice, always take the multivendor way ;) 
Thanx

html/includes/print-nac.inc.php
```diff
@@ -0,0 +1,69 @@
+<?php
+echo '<table class="table table-hover table-condensed">
```
laf: You'll need to convert this table to use bootgrid.

We have lots using it already for examples.

html/pages/device.inc.php
```diff
@@ -435,6 +435,14 @@
                 </a>
                 </li>';
         }
+        
+        if ($config['enable_ports_nac'] == 1) {
```
laf: Convert this check to a dbFetch to see if data exists in the new tables for this device.

includes/polling/ports.inc.php
```diff
@@ -362,6 +363,55 @@
         $port_stats = snmpwalk_cache_oid($device, 'cpeExtPsePortEntry', $port_stats, 'CISCO-POWER-ETHERNET-EXT-MIB');
     }
 }
+            
+//NAC Polling
+if ($config['enable_ports_nac']) {
```
laf: I was the one who said put this in ports but I didn't realise how much data this would pick up. I'd definitely agree this should be it's own polling module with the ability to call OS files specifically. Ideally, it would be done using OOP rather than the old code format.

laf:
@joseaugustocardoso Any news on this?

joseaugustocardoso:
> @joseaugustocardoso Any news on this?

I'll check all changes requests soon.
And reply each one.

includes/polling/ports.inc.php
joseaugustocardoso: Ok, I told sometime to understand the code to do what I did, and to replace that to a module, I have to study more things on LNMS code. Im not sure if it will be quick, but I'll hard work on that.
Maybe we can release it as is, while I replace this to a module. What do you think?

html/includes/print-nac.inc.php
joseaugustocardoso: screen shot 2018-11-02 at 00 36 49

joseaugustocardoso: done. Thanks for advising.

html/pages/device.inc.php
joseaugustocardoso: I have created a new module called Cisco-NAC
screen shot 2018-11-06 at 23 48 51

includes/polling/ports.inc.php
joseaugustocardoso: Done!
screen shot 2018-11-06 at 23 48 51

joseaugustocardoso: Done.
screen shot 2018-11-06 at 23 48 51

includes/polling/cisco-nac.inc.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+use LibreNMS\Util\IP;
+
+if (!get_dev_attrib($device, 'poll_cisco-nac', 'true')) {
```
murrant: It would be better to add a discovery module to detect if nac should be polled or not.

includes/polling/cisco-nac.inc.php
```diff
@@ -0,0 +1,83 @@
...
+    $ports_mapped = get_ports_mapped($device['device_id'], true);
+    $ports = $ports_mapped['ports'];
+    $PortAuthSessionEntry = snmpwalk_cache_oid($device, 'cafSessionEntry', array(), 'CISCO-AUTH-FRAMEWORK-MIB');
+    $cafSessionMethodsInfoEntry = snmpwalk_cache_oid($device, 'cafSessionMethodsInfoEntry', array(), 'CISCO-AUTH-FRAMEWORK-MIB');
+    foreach ($PortAuthSessionEntry as $PortAuthSessionEntryNAC => $PortAuthSessionEntryParameters) {
+        $port_index_nac = substr($PortAuthSessionEntryNAC, 0, strpos($PortAuthSessionEntryNAC, "."));
+        $port_auth_id_nac = strstr($PortAuthSessionEntryNAC, "'");
+        $port_auth_id_nac = substr($port_auth_id_nac, 0, -1);
+        $port_auth_id_nac = substr($port_auth_id_nac, 1);
+        $CheckExistIndex = dbFetchRow('SELECT * FROM `ports_nac` WHERE `port_index` = ?', $port_index_nac);
```
murrant: Still missing schema file to define ports_nac table.

includes/polling/cisco-nac.inc.php
joseaugustocardoso: I didnt find yet a way to detect the NAC enable.

includes/polling/cisco-nac.inc.php
joseaugustocardoso: Done

includes/polling/cisco-nac.inc.php
```diff
@@ -13,11 +13,43 @@
         $port_auth_id_nac = strstr($PortAuthSessionEntryNAC, "'");
         $port_auth_id_nac = substr($port_auth_id_nac, 0, -1);
         $port_auth_id_nac = substr($port_auth_id_nac, 1);
-        $CheckExistIndex = dbFetchRow('SELECT * FROM `ports_nac` WHERE `port_index` = ?', $port_index_nac);
-        $CheckExistAuth = dbFetchRow('SELECT * FROM `ports_nac` WHERE `auth_id` = ?', $port_auth_id_nac);
+        $port_auth_domain = $PortAuthSessionEntryParameters['cafSessionDomain'];
         $IPHextoDec = IP::fromHexString($PortAuthSessionEntryParameters['cafSessionClientAddress']);
-        if ($CheckExistIndex['port_index'] != null) {
-            echo "Port Index Was Found. Updating...\n";
+        $CheckExistIndex = dbFetchRows('SELECT port_index FROM `ports_nac` WHERE `port_index` = ?', $port_index_nac);
```
murrant: Instead of dbFetchRows(), you can use dbFetchRow() to selected a single row (that way you can avoid the [0] below)

Even better, if you are selecting a single item, you can dbFetchCell() to select a single value from the db and it won't be in an array.

murrant:
Thanks for the updates so, to split your poller into discovery and polling (which will reduce the polling cost).

Discovery:
You need to execute snmpwalks and then create/update/delete entries in the database

Poller:
look up data from the db to determine what to poll
snmpget multi or snmpwalk data for those items and update the database.

This allows us to have the discovery enabled on all relevant devices and not affect polling performance for devices that don't have any data to poll. :)

includes/defaults.inc.php
```diff
@@ -778,6 +778,7 @@
 $config['poller_modules']['cisco-voice']                 = false;
 $config['poller_modules']['cisco-cbqos']                 = false;
 $config['poller_modules']['cisco-otv']                   = false;
+$config['poller_modules']['cisco-nac']                   = false;
```
laf: I think this should just be `nac` as I'm sure cisco won't be the only ones to implement network access control support.

joseaugustocardoso: Could be, but this module will work only with Cisco switches, because Im using Cisco's MIB.

includes/polling/cisco-nac.inc.php
joseaugustocardoso: Done. It make all sense. Thanks for advising.

joseaugustocardoso:
> Thanks for the updates so, to split your poller into discovery and polling (which will reduce the polling cost).
> 
> Discovery:
> You need to execute snmpwalks and then create/update/delete entries in the database
> 
> Poller:
> look up data from the db to determine what to poll
> snmpget multi or snmpwalk data for those items and update the database
> sometimes is is appropriate to add/remove items in the poller.
> 
> This allows us to have the discovery enabled on all relevant devices and not affect polling performance for devices that don't have any data to poll. :)

@murrant, thank you for explanation. But in this case, I believe that all ports must be check on each polling process (5 mins), because the switch port can be very dynamic, then the authentication polling must follow that. 
Btw, I agree that I need to find a way to identify if the switch has the NAC feature running to enable the module automatically. The problem is that I didnt find yet a way to verify that. The MIB check return data same if the NAC is not running. So, at this time, on my opinion, this module must be enable manually :(

includes/defaults.inc.php
PipoCanaja: Hi,
The idea would be to actually reuse most of the inside of if but add support for other switches. I am already thinking of my huawei devices on which I run NAC... 
So basically, I would follow laf and suggest to change name to nac or dot1x or whatever is constructor agnostic. 
PipoCanaja

murrant:
If there is no way to determine if NAC is enabled, then, yeah, probably a manual poller module.

Next step is to add the ability to test this module and then add some test, add the module to tests/module_tables.yaml and then collect snmp data (see https://docs.librenms.org/Developing/os/Test-Units/)

Also, your poller code could be improved with regards to readability.   Try reading through the code imagining you don't know what it does.  Comments, whitespace, and simplifying code can help.  Ask me on discord sometime if you need more specific pointers.

murrant:
@joseaugustocardoso if you can add test data `./scripts/collect-snmp-data.php -h HOSTNAME`, I'll help you out a bit with this.

murrant:
@joseaugustocardoso is this code working for you?  LGTM and works ok in my testing.

joseaugustocardoso:
> @joseaugustocardoso is this code working for you? LGTM and works ok in my testing.

I making a stress test, on real NAC environment. Back to you soon.

joseaugustocardoso:
> > @joseaugustocardoso is this code working for you? LGTM and works ok in my testing.
> 
> I making a stress test, on real NAC environment. Back to you soon.

@murrant, is not working as well anymore.
You removed all logic on includes/polling/nac.inc.php
I made some combination of if/else covering all situations...

murrant:
@joseaugustocardoso what did the if else cover?
The current code creates/updates/deletes records as necessary.

joseaugustocardoso:
> @joseaugustocardoso what did the if else cover?
> The current code creates/updates/deletes records as necessary.

The logic is:
If (Authentication ID) collected was found on DB, do nothing (ignore), it will match many times.

Else if (Auth ID and Port Index) collected was not found on DB, create on DB (normally this is will be the first polling process).

Else if (Port Index and Domain (Voice or Data)) collected is same on DB (auth ID is different), then update this DB line with a new authentication information. If the switch port has 2 devices connected (ip phone and PC) we will have 2 authentication ID, each one per domain (data and voice)

So, I have stressed this logical many times, creating some situations to make sure all possibilities will be covered.

murrant:
Ok, sounds like I just used the wrong columns to see if a record exists.  It should use port_id and domain (not auth id).

murrant:
@joseaugustocardoso Is it correct now?

joseaugustocardoso:
> @joseaugustocardoso Is it correct now?

Yes, thats perfect right now!

I just committed a new change to reorder AuthZ to stay side by side with AuthC column.

murrant:
@joseaugustocardoso Thanks for the contribution!

By the way, you should update your git author email address to match the email you use for github :)

laf:
Looks like you might need to re-save the test json data.

murrant:
I think the data capture needs an order by clause.

murrant:
@joseaugustocardoso Very close here.  But we need an additional piece of data to determine unique entries here.

Using the test data (which contains five auth entries on a single port) results in one entry (username5) with the current code.

What column should we use?

Maybe you could share some real snmp data with me privately.  The test data is too abstract for me to figure it out.

joseaugustocardoso:
> @joseaugustocardoso Very close here. But we need an additional piece of data to determine unique entries here.
> 
> Using the test data (which contains five auth entries on a single port) results in one entry (username5) with the current code.
> 
> What column should we use?
> 
> Maybe you could share some real snmp data with me privately. The test data is too abstract for me to figure it out.

Ok, I'm working on that.

Btw, how can I send you the file privately?

murrant:
Probably need to update the test data to better match reality/test this.

FTBZ:
Do not use this, it will not work:
> $config['enable_ports_nac'] == 1

Use this to enable the NAC module:
`$config['poller_modules']['nac'] = true;`

FTBZ:
@joseaugustocardoso It would be really great to have a menu dedicated only to NAC and not need to go on each of the switches.

PipoCanaja:
@joseaugustocardoso @FTBZ Could you give me your feedback on Cisco NAC after I added Huawei support ? 
You can also have a look at https://github.com/librenms/librenms/pull/9706 which is mainly changing the display (width of cells and tooltips)
Thx

PR #9228, opened 2018-09-19
Update syslog display and backend

murrant:
select boxes now dynamically load instead of loading all on pageload
select and table ajax backends completely redone with Laravel (just syslog for now)
duplicate url creation to Url utility class for now (uses Device model instead of array)
build short hostname functionality into Device->displayName() helper

tested with over 1 million syslog entries :1st_place_medal: 

https://community.librenms.org/t/pruning-syslog-items-from-librenms-dbs/2485

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
The select boxes are a little small and unusable imho:

image

Aside from that, works fine and looks so much better :) Wonder if you could enable multi-select on this, that could be pretty cool.

murrant:
yeah, sizing them with ajax is tricky. it doesn't have all the data.  I can set initial sizes.

PR #9247, opened 2018-09-21
Check sysDescr for JunOS version.

Zmegolaz:
Some JunOS routers (seems to be mainly larger ones like MX480/960+) doesn't have the OID checked for software version. This patch checks sysDescr, where it seems to be mentioned in most (all?) models.

This saves an SNMP query each poll too.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/os/junos.inc.php
+++ b/includes/polling/os/junos.inc.php
@@ -22,6 +22,10 @@ if (is_numeric($srx_sess_data[0]['jnxJsSPUMonitoringCurrentFlowSession'])) {
 
 $version = snmp_get($device, 'jnxVirtualChassisMemberSWVersion.0', '-Oqv', 'JUNIPER-VIRTUALCHASSIS-MIB');
 if (empty($version)) {
+    preg_match('/kernel JUNOS (\S+),/', $device['sysDescr'], $jun_ver);
+    $version = $jun_ver[1];
+}
+if (empty($version)) {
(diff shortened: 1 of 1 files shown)


includes/polling/os/junos.inc.php
```diff
@@ -21,6 +21,10 @@
 }
 
 $version = snmp_get($device, 'jnxVirtualChassisMemberSWVersion.0', '-Oqv', 'JUNIPER-VIRTUALCHASSIS-MIB');
+if (empty($version)) {
+    preg_match('/kernel JUNOS (\S+),/', $device['sysDescr'], $jun_ver);
+    $version = $jun_ver[1];
```
laf: Doesn't setting this here now mean the next check never matches so could be removed.

Zmegolaz: Probably, but I'm not 100% sure all sysDescr are formatted this way and will be matched by the regex. I thought leaving the old code behind as a fallback isn't much of an issue.

zombah:
@Zmegolaz Imho its connected not to specific MX series models but to combination of newer JunOS version for MX series. All my MX480/MX960 devices with 14.2R7 and 15.1 versions parsed fine by oid, devices with decent version of 16.1R6 are not.

laf:
@Zmegolaz Any response to the comment from zombah.

PR #9256, opened 2018-09-23
Add sitemap.xml to robots.txt

TheGreatDoc:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/robots.txt
+++ b/doc/robots.txt
@@ -1,2 +1,3 @@
 User-agent: *
 Disallow:
+Sitemap: https://docs.librenms.org/sitemap.xml


doc/robots.txt
```diff
@@ -1,2 +1,3 @@
 User-agent: *
 Disallow:
+Sitemap: https://docs.librenms.org/sitemap.xml
```
laf: How come you've added this in?

We disallow bots so they won't see this value anyway.

TheGreatDoc: I know almost 0 about SEO but:

`User-agent: *` means all robots are allowed
`Disallow:` Nothing is disallowed to index

So I've added the sitemap in a hope of Google indexing the docs site so when someone search about librenms and is covered in the docs it will return docs as result

PR #9257, opened 2018-09-23
Add feature to sort alert schedules by status

remyj38:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

It's should be very useful to sort alert status by status.

### DIFF
--- a/html/includes/table/alert-schedule.inc.php
+++ b/html/includes/table/alert-schedule.inc.php
@@ -31,7 +31,13 @@ if (empty($total)) {
     $total = 0;
 }
 
-if (!isset($sort) || empty($sort)) {
+if (isset($sort) && !empty($sort)) {
+    list($sort_column, $sort_order) = explode(' ', trim($sort));
+    if ($sort_column == 'status') {
+        $sort_by_status = true;
+        $sort = "`S`.`start`  $sort_order";
+    }
+} else {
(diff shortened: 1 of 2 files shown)


html/includes/table/alert-schedule.inc.php
```diff
@@ -108,6 +109,18 @@
     );
 }
 
+if (substr($original_sort, 1, 6) == 'status') {
```
laf: Got an example output of before and after of what this code block is doing?

html/includes/table/alert-schedule.inc.php
```diff
@@ -31,7 +32,7 @@
     $total = 0;
 }
 
-if (!isset($sort) || empty($sort)) {
+if (!isset($sort) || empty($sort) || substr($sort, 1, 6) == 'status') {
```
laf: The ports.inc.php file in this dir does the sort by:

`list($sort_column, $sort_order) = explode(' ', trim($sort));`

It would be better to stick to the same method here if you can.

laf:
@remyj38 Any response to the above?

html/includes/table/alert-schedule.inc.php
remyj38: Updated to use the same syntax that the ports.inc.php file

remyj38:
@laf Updated with your suggestion

PR #9258, opened 2018-09-23
Add alerts schedule notes into device notes

remyj38:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---
Having alerts schedules notes into devices notes improves device status traceability.

### DIFF
--- a/html/includes/forms/schedule-maintenance.inc.php
+++ b/html/includes/forms/schedule-maintenance.inc.php
@@ -129,6 +129,11 @@ if ($sub_type == 'new-maintenance') {
             foreach ($_POST['maps'] as $target) {
                 $target = target_to_id($target);
                 $item   = dbInsert(array('schedule_id' => $schedule_id, 'target' => $target), 'alert_schedule_items');
+                if ($notes && get_user_pref('add_schedule_note_to_device', false)) {
+                    $device_notes = dbFetchCell('SELECT `notes` FROM `devices` WHERE `device_id` = ?;', array($target));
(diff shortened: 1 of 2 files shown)


html/includes/forms/schedule-maintenance.inc.php
```diff
@@ -129,6 +129,9 @@
             foreach ($_POST['maps'] as $target) {
                 $target = target_to_id($target);
                 $item   = dbInsert(array('schedule_id' => $schedule_id, 'target' => $target), 'alert_schedule_items');
+                if ($notes) {
+                    dbQuery('UPDATE `devices` SET `notes` = CONCAT(IFNULL(`notes`, ""), "' . date("Y-m-d H:i") . ' Alerts delayed:' .  $notes . '") WHERE `device_id` = ' . $target . ';');
```
murrant: Don't use dbQuery like this, it is vulnerable to sql injection.  You should always use the parameters array.

Why not just fetch notes and then set it?

Will the notes ever be removed?

laf:
I can't help but feel this might not be what people want to happen by default.

html/includes/forms/schedule-maintenance.inc.php
remyj38: MySQL can concatenate strings, so why don't use it ?

I changed to use dbupdate instead of dbquery, with parameters. So, needed to fetch notes before update.

Nop, never removed

remyj38:
Personal setting added to enable the feature, disabled by default

PR #9259, opened 2018-09-24
Allow zoom to be decimal and switch zoomSnap to 0.1

cchance27:
Allow 1 decimal place for initial zoom and lower zoom snap interval to improve initial zoom in widgets.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


cchance27:
not going to lie here i wanted to add this as an adjustable value in the widget settings and default settings via config, but got super confused with the widget settings not working, and then after some testing on my side i couldn't find a reason not to just set the default zoomSnap to 0.1 and change the sprintf to allow 1 decimal place. 

This allows a bit nicer/more precise initial zooming (especially on widgets)

PR #9261, opened 2018-09-24
[WIP] Allow librenms to execute SNMP via SSH

TheMysteriousX:
I'm not exactly sure if you'd want this changeset, so opening a pull request early to get some feedback.

I needed to poll  a device on a network that I didn't have direct access to, and DNS magic/SNMP proxy seemed unappealing, so I spent half an hour seeing if I could convince librenms to use netsnmp on a remote system.

It worked, and it's surprisingly performant.

There's some OpenSSH magic to make the SSH connection persist for the entire duration of the polling, caveats are the target system needs net-snmp, and the librenms mibs in the same location as the host.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/addhost.php
+++ b/addhost.php
@@ -75,7 +75,7 @@ if (isset($options['j'])) {
 
     $jump = explode('@', $options['j']);
 
-    if (count($jump === 2)) {
+    if (count($jump) === 2) {
         $jump_user = $jump[0];
         $jump_hostname = $jump[1];
     } else {
--- a/addhost.php
+++ b/addhost.php
@@ -16,7 +16,7 @@ use LibreNMS\Exceptions\HostUnreachableException;
 $init_modules = array();
 require __DIR__ . '/includes/init.php';
 
-$options = getopt('Pbg:p:f::');
+$options = getopt('j:Pbg:p:f::');
 
 if (isset($options['g']) && $options['g'] >= 0) {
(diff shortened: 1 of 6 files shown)


murrant:
Just a question, not saying anything should be this way or that.

You are running the snmp commands on the remote host?  Why not just tunnel the snmp through the ssh connection?

TheMysteriousX:
Driven by my need for the feature to get this device monitored. Wrapping the snmp shelll commands needs the fewest code changes to prove make it work without spawning a connection with every command, or totally wrecking performance.

It's more difficult to use a forwarded port with the imperative style of the poller, and because net-snmp doesn't support SOCKS:
* devices.port needs to be dynamically assigned and maintained through the lifecycle of the polling session to stop two processes trying to use the same port (+ hard limit of 20k devices, ish, don't know if anyone would have a deployment that large)
* devices.hostname needs to be ignored and replaced
* the connection needs to be held open as the control process would see as idle
* the connection needs to be released somehow, even if the poller crashes

A little thought is also needed to make sure that it isn't difficult to extend in future with e.g. stunnel or wireguard

If you like the idea I'll extend it with the above and some GUI changes to display the jump host details.

murrant:
Closing this for now, feel free to re-open.

PR #9262, opened 2018-09-25
remove apc-ats os and merge sensors into apc

tomarch:
Thanks @FTBZ for the #9221 !
But i think we don't need a new os "apc-ats" because:
-it need to add exception in "apc" os definition  (ATS use same firmware and so the same Sysdesc)
-ATS have other sensor which is already present in "apc" discovery

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
@FTBZ Any comments?

FTBZ:
On my side the metrics are returning errors and are not working. The problem is that to fix these metrics, we risk breaking things for PDU users. That's why I created the new OS. 

After if the metrics do not work for me, it does not matter as long as I have the state sensors.

FTBZ:
Nice to know, seems better now with the new firmware. The MIB file was updated too, can you add the new version to the commit?

murrant:
Instead of deleting the test data, rename it to APC (and update it if necessary)

tomarch:
I've add the test data.
thanks.

murrant:
You need the snmprec too.  I would suggest the name apc_ats.snmprec and apc_ats.json for the files.

tomarch:
We want to remove apc-ats OS as we don't need it.
There's already a apc.snmprec, so i only remove apc-ats.snmprec

murrant:
yeah the  underscore separates the os and the variant, a dash means it is part of the os name.

apc-ats  (OS is apc-ats)
apc_ats (OS is apc, variant is ats)

We use variant to allow multiple test datas per os.

tomarch:
OK. thanks for this information.

laf:
Updated tests.

murrant:

PR #9301, opened 2018-10-05
device: MGE UPS support improvement (incl. traps)

PipoCanaja:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/discovery/sensors/load/rfc1628.inc.php
```diff
@@ -5,7 +5,7 @@
 $load_data = snmpwalk_group($device, 'upsOutputPercentLoad', 'UPS-MIB');
 
 foreach ($load_data as $index => $data) {
-    $load_oid = ".1.3.6.1.2.1.33.1.4.4.1.5.$index";
+    $load_oid = ".1.3.6.1.2.1.33.1.4.4.1.5.".$index.".0";
```
laf: We've not had any complaints about these and I feel this might break things (doesn't look like we have any test data, I do though and none of it shows with a .0)

PipoCanaja: Would it be something specific to the Galaxy 7000 ? I can (and will) provide test data for my units. But I did not find any other test data matching this RFC anyway so I cannot say if this change breaks anything or not ... 
I'll try to check the MIB itself to get convinced what is the "correct" way, and search for a proper way to handle this. In the meantime, I'll put a WIP on this.

PipoCanaja: Hello,
I did change the behaviour to detect if the data comes as an array (meaning this ".0" is necessary) and do the necessary changes for the Galaxy to work without impacting existing device. 
I also added the test data from our device. 
This should now be a lot safer for any other device.

includes/discovery/sensors/voltage/rfc1628.inc.php
```diff
@@ -25,15 +25,23 @@
     );
 }
 
+
 $output_volts = snmpwalk_group($device, 'upsOutputVoltage', 'UPS-MIB');
 foreach ($output_volts as $index => $data) {
-    $volt_oid = ".1.3.6.1.2.1.33.1.4.4.1.2.$index";
+    $volt_oid = ".1.3.6.1.2.1.33.1.4.4.1.2.".$index;
```
laf: This change seems pointless, may as well revert it.

includes/definitions/discovery/eaton-mgeups.yaml
```diff
@@ -0,0 +1,234 @@
+mib: MG-SNMP-UPS-MIB
+modules:
+    sensors:
+        power:
+            data:
+                -
+                    oid: upsmgOutput
```
laf: For all of these, you've used a high level folder / OID when you should be using the Table. See this pic:

image

So use upsmgOutputPhaseTable in this one example but the rest also need to be updated.

laf: It will always do a walk but if you will be re-using data within the folder OID then it will be fine to use it as we cache the original walk so long as you aren't walking a lot of data.

It's a trade off between walking too much data to cache and not caching enough and having to perform multiple walks. Neither is perfect just use the best option.

nonobzh29:
This commit introduced a bug regarding the input and output voltage values.
It drops by a factor of 10 ! (23.2V instead of 232V).
Please fix it asap, I didn't like to be waked up by the alert last night after the update ...

PipoCanaja:
Hi @nonobzh29 
It seems that all eaton devices are not behaving the same. 
Could you PasteBin your device discovery.php and poller.php details ?
Thanx

nonobzh29:
Hi @PipoCanaja 
I confirm it's fixed now! Thanks!

murrant:

PR #9311, opened 2018-10-09
Fix bug of alerting by ping.php

DR3EVR8u8c:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

## please refer the Help ticket for more information:
https://community.librenms.org/t/please-help-with-for-fast-ping-checks-problems/5586

**Bug description:**
the PingCheck.php saves the device status after it run rules. Therefore, the rules cannot be detected correctly as the devices table hasn’t been updated. While Poller.php update the devices status in db before it run the rules detection.


laf:
@murrant Does this look like the correct way to fix this?

murrant:
@laf no, this is not correct.  After the device model has been saved, it is no longer dirty.

It is correct to just move the save call inside the isDirty check right before RunRules().

murrant:
Actually, because of last_ping.  We can't do it that way.

PR #9315, opened 2018-10-10
Discovery YAML: do not implicitly append $index

murrant:
Require it explicitly. Makes it easier to understand and matches the style of other values

Questionable YAML changes (either broken before or now broken):
secureplatform: haStatCode
ptp600: receiveModulationMode

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Out of interest why do you prefer to have to specify `{{ $index }}`. It's just more to write in the file which means it's more error prone for things like spelling mistakes, missing brackets etc.

murrant:
Often people don't understand that the index is appended to the end.  Also, many other places require {{ $index }} to be specified, this is the only place that adds it.

**We gain a lot of clarity and consistency for just a little extra verbosity.**

laf:
> Also, many other places require {{ $index }} to be specified, this is the only place that adds it without that.

In yaml this came first. You added {{ $index }} in processors :p Needs a rebase anyway. I've personally never come across a PR that has added it by my mistake.

murrant:
I can think of at leaat four people asking in chat why discovery worked and polling didn't, with the cause being an incorrect oid because they didn't know index is appended.

I'd like to remove that friction.  I think this is the best way.

PR #9320, opened 2018-10-12
 Updated to use new theme for docs site

laf:
Test site: http://web01.lathwood.uk:8443/

If this is good for a merge then don't, it may not deploy the site correctly so I'd like to merge it in and test / fix on the fly if needs be.


Rosiak:
I'm cool with this, wish the load time will be the same as your test site ;-)

laf:
> I'm cool with this, wish the load time will be the same as your test site ;-)

It should / will with this new template.

murrant:
@laf I modified this to just use the base material theme.
Probably should restrict the theme version to be safe.

scripts/deploy-docs.sh
```diff
@@ -1,18 +1,16 @@
...
 FULL_REPO="https://${GH_TOKEN}$GH_REPO"
-THEME_REPO="https://github.com/librenms-docs/theme_v2.git"
 
 if [ "$EXECUTE_BUILD_DOCS" != "true" ]; then
     echo "Doc build skipped"
     exit 0
 fi
 
-pip install --user 'jinja2<2.9' mkdocs pymdown-extensions
+pip install --user mkdocs mkdocs-material pymdown-extensions
```
laf: mkdocs-material by default needs mkdocs 1.x, mkdocs that's currently being installed in travis is 0.17.5 so I'm not sure this is going to work

murrant:

PR #9322, opened 2018-10-15
Added option to disable port discovery and poll if ifAdminStatus down

fmarton:
Issue:
I have network with many nodes and many ports in a node. I want to discover the used ports only. In this case there are less port.rrd so the summary size is smaller for whole network.

Changes:
ifAdminStatus can be used as filter for port discovery and poll
ifName is printed out in debug mode when discovery and poll

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Support/Configuration.md
+++ b/doc/Support/Configuration.md
@@ -516,7 +516,7 @@ Examples:
 $config['bad_if'][] = "voip-null";
 $config['bad_iftype'][] = "voiceEncap";
 $config['bad_if_regexp'][] = '/^lo[0-9].*/';    // loopback
-$config['bad_ifAdminStatus'][] = '/down/';  
+$config['bad_ifAdminStatus'][] = 'down';  
 ```
 Numerous defaults exist for this array already (see includes/defaults.inc.php for the full list). You can expand this list
 by continuing the array.
@@ -531,7 +531,7 @@ by continuing the array.
 
 `bad_ifalias_regexp` is matched against the ifAlias value as a regular expression.
 
-`bad_ifAdminStatus` is matched against the bad_ifAdminStatus value as a regular expression. 
+`bad_ifAdminStatus` is matched against the bad_ifAdminStatus value. 
 
 ### Interfaces that shouldn't be ignored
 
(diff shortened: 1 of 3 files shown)


laf:
Have you tried selected port polling, this does the same thing

fmarton:
Similar but the goal is different. Selected Port Polling doesnot  poll the not used ports to optimize poller modul. My solution dont discover the admin down ports, not in the deleted ports list, not shown in the node summary.  It is working as IfDesc filter, can be used with or without per port polling feature.

I have many nodes with many interfaces, I  want to see not used port nowhere .

laf:
> My solution dont discover the admin down ports, not in the deleted ports list, not shown in the node summary. It is working as IfDesc filter, can be used with or without per port polling feature.

Well you might not see the data but without selected port polling it's still polling the data regardless.

I'm in two minds about this but wouldn't stop it from being merged by others.

However, you do need to document this new config option: https://docs.librenms.org/#Developing/os/Settings/#ignoring-interfaces

fmarton:
Documented in 'Interfaces to be ignored' section.

laf:
I do wonder if we could just make this lot of config options more dynamic like `
```
$config['bad_int']['ifAdminStatus'][] = '/^down$/';
$config['bad_int']['ifDescr'][] = '/Ethernet1/';
```

murrant:
My plan is to make the port configuration screen a lot more powerful/informative.  Telling what ports will be polled, hidden, not polled and allowing control of that (including surfacing selective port polling)

fmarton:
> I do wonder if we could just make this lot of config options more dynamic like `
> $config['bad_int']['ifAdminStatus'][] = '/^down$/';
> ```
Please remember ifAdminStatus is from IF-MIB and the possible values are up, down and testing. It is not worth to use regexp in this case.

fmarton:
> My plan is to make the port configuration screen a lot more powerful/informative. Telling what ports will be polled, hidden, not polled and allowing control of that (including surfacing selective port polling)

Current GUI is good enough. maybe the mass or automatically selection needxs improvment. One by one setting is not effective if there are many ports.

laf:
> Please remember ifAdminStatus is from IF-MIB and the possible values are up, down and testing. It is not worth to use regexp in this case.

It's not about one item in the MIB rather being able to use anything we currently poll for. But using the regex is worth it because you can restrict it to only the values you want to see.

Rosiak:
```
$config['bad_int']['ifDescr'][] = '/Ethernet1/';
```
Ability to filter sensors based on this would be great, we have had requests for this quite recently.

murrant:
Closing, feel free to reopen discussions.

PR #9336, opened 2018-10-18
option to turn off icmp check for polling while fast-ping check is implemented

DR3EVR8u8c:
While the Fast-Ping Check feature has been implemented, the default ping check during polling is not helpful. And the default ping sometimes cause small gap in the graph due to the fping failure.
Due to the reason above, I would like to add an option to turn off the icmp check in polling while do not stop the fast-ping check.
Thanks and regards,
Roger

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


DR3EVR8u8c:
don't know why this PR including some past commits from @laf and @murrant. I expected that the PR only includes my last 2 changes. Please let me know if you want me to close this PR and create new one.

murrant:
Seems fine in the diff.

By the way you understand that these two ping checks are not interchangeable.  You will have less data by only using ping.php.

DR3EVR8u8c:
@murrant , May I know what data will be missing if use ping.php only?

DR3EVR8u8c:
Sorry @murrant, I may make mistake. initially I assumed that the config['icmp_check']=false will stop all pings including fast-ping. but that seems not the case. if so, then the option may be not required. I will close the PR now.

murrant:
@DR3EVR8u8c The performance tab data is not populated by ping.php.  ping.php sends 1 ping (if successful), the poller sends a minimum of 3 pings (by default) so it can get average latency, etc.

PR #9338, opened 2018-10-18
Removed redundant grouping from SQL query

angryp:
This change is made for "FDB table" web interface page, which uses redundant information for "GROUP BY" statement. More information available in community forums: https://community.librenms.org/t/fdb-table-page-produces-high-load/5831

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
We don't have anywhere near that amount of data but with this pr it went from 2.25s to 0.11s.

it does break with ONLY_FULL_GROUP_BY enabled though.

> What if we pulled back the joins and ran separate queries to fetch port and device information? That way it only has to fetch it for the current page of entries.

Aren't we only pulling that for the current page now anyway as it's limited to 50 results. It's only the count query but that should be a quick query.

murrant:
Yeah, the old query causes a full scan.  I'd say this is ok for now, but when this is rewritten we will probably have to use a different strategy.

PR #9343, opened 2018-10-18
snmp_max_oid per Os support and snmpv1 multi_oid fix

alangregory:
_The difference between SNMPv1 and the later versions becomes apparent when one of the OIDs being requested is not valid. SNMPv2c (and SNMPv3) will display what information they can._

So if one oid is wrong or doesn't exist, v1 will trhow an error instead of showing the working ones.
It's safe to get one by one.
#### Please note

#### Testers

### DIFF
--- a/includes/snmp.inc.php
+++ b/includes/snmp.inc.php
@@ -222,12 +222,24 @@ function snmp_get_multi_oid($device, $oids, $options = '-OUQn', $mib = null, $mi
 {
     $time_start = microtime(true);
 
-    if (is_array($oids)) {
-        $oids = implode(' ', $oids);
+    if ($device['snmpver'] == 'v1') {
+        $oids = explode(" ", $oids);
+        $data = [];
+        foreach ($oids as $oid) {
+            $cmd = gen_snmpget_cmd($device, $oid, $options, $mib, $mibdir);
+            $query = trim(external_exec($cmd));
+            if (!empty($query)) {
+                $data[] = $query;
+            }
+        }
+        $data = implode("\n", $data);
+    } else {
+        if (is_array($oids)) {
+            $oids = implode(' ', $oids);
+        }
+        $cmd = gen_snmpget_cmd($device, $oids, $options, $mib, $mibdir);
+        $data = trim(external_exec($cmd));
     }
-
-    $cmd = gen_snmpget_cmd($device, $oids, $options, $mib, $mibdir);
-    $data = trim(external_exec($cmd));
 
     $array = array();
     $oid = '';
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2481,7 +2481,12 @@ function get_device_oid_limit($device)
 {
     global $config;
 
-    $max_oid = $device['snmp_max_oid'];
+    //Some Os only work with specific max_oid values
+    if ((isset($device['os'], $config['os'][$device['os']]['snmp_max_oid']) && $config['os'][$device['os']]['snmp_max_oid'])) {
+        $max_oid = $config['os'][$device['os']]['snmp_max_oid'];
+    } else {
+        $max_oid = $device['snmp_max_oid'];
+    }
 
     if (isset($max_oid) && $max_oid > 0) {
         return $max_oid;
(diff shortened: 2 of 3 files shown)


murrant:
Hmm, we will lose performance on v1 with this.  Are you sure that is the behavior of all v1, or just a few devices?  (although, my memory seems to indicate I have ran into this before)

alangregory:
This was taken from http://net-snmp.sourceforge.net/wiki/index.php/TUT:snmpget section Multiple Values
It's v1 beahavior.

laf:
So it's not that v1 doesn't support it, it's that if we ask for an OID not available it would fail.

Where are you seeing this issue?

Wonder if we could just re-run the get without the one that caused the failure. That would have a performance impact for the ones requesting the wrong OIDs but for everything else it will be the same. The fact this is the first I've heard about it indicates it's not a huge issue.

laf:
From our stats, v1 is about 10% of the total so I'd have expected to have heard of others having issues if this was a big thing.

murrant:
I remember switching our discovery to multiget, which triggered this issue.

I think maybe we could try multi then fallback to single if there is an error for v1?  That would reduce the impact.  Is it worth it though?

alangregory:
I'm testing with some devices who support v1 only, and another v2c Os behave the same manner, no output if one oid is wrong.
Another alternative would be to check Os snmp_max_oid.

murrant:
@alangregory I did some testing:
```
snmpget -v1 murrayap sysObjectID.0 sysDescr.0 sysDescr.1 sysContact.0
Error in packet
Reason: (noSuchName) There is no such variable name in this MIB.
Failed object: SNMPv2-MIB::sysDescr.1

SNMPv2-MIB::sysObjectID.0 = OID: NET-SNMP-MIB::netSnmpAgentOIDs.10
SNMPv2-MIB::sysDescr.0 = STRING: UAP-nanoHD 3.9.54.9373
SNMPv2-MIB::sysContact.0 = STRING: root@localhost
```

So, that device returns all the data it has after a bad oid.  It does look like the format differs from v1-v2c:
```
snmpget -v2c  watchguard sysObjectID.0 sysDescr.0 sysDescr.1 sysContact.0 
SNMPv2-MIB::sysObjectID.0 = OID: SNMPv2-SMI::enterprises.3097.1.5.49
SNMPv2-MIB::sysDescr.0 = STRING: M400
SNMPv2-MIB::sysDescr.1 = No Such Instance currently exists at this OID
SNMPv2-MIB::sysContact.0 = STRING: email
```

alangregory:
Understood, it maybe be half device related to, i've got no additional output on some devices.
Maybe add a check on get_device_oid_limit ?

murrant:
@alangregory that is probably a good solution.

includes/snmp.inc.php
```diff
@@ -221,14 +221,30 @@ function snmp_get_multi($device, $oids, $options = '-OQUs', $mib = null, $mibdir
...
-    if (is_array($oids)) {
+        $oid_limit = get_device_oid_limit($device);
+    if (!is_array($oids)) {
+        $oids = explode(" ", $oids);
+    }
+    $oid_number = count($oids);
+
+    if ($oid_limit < $oid_number) {
+        $data = [];
+        while (!empty($oids)) {
```
murrant: If you use array_chunk(), you can avoid the if/else and make this a lot simpler.

alangregory: Changed.
There's a problem with "\n" as divider.
Some result may have it inside response, like : 
```
"""
.1.3.6.1.2.1.1.3.0 = 3402322978\n
.1.3.6.1.2.1.1.6.0 = \n
.1.3.6.1.2.1.1.4.0 = \n
.1.3.6.1.2.1.1.5.0 = xxxxxxxxxxxx\n
.1.3.6.1.2.1.1.2.0 = .1.3.6.1.4.1.9.1.1639\n
.1.3.6.1.2.1.1.1.0 = Cisco IOS XR Software (Cisco ASR9K Series),  Version 6.1.3[Default]\r\n
Copyright (c) 2017 by Cisco Systems, Inc.
"""
```

PR #9313, opened 2018-10-09
WebUI: edituser page with new Auth system

PipoCanaja:
Hi team,

This PR was started because I lacked the "Edit Permission" capability for Active Directory users. I tried to find some kind of "minimal" changeset to achieve this.

And I like the little encouragement cause **Edit Permissions** works so far :) (tested it 5 minutes, time needed to create and edit permission on one new user) 

I'll continue digging in this direction, but I would like to get your opinion and advice (and probably help as well) on this. I suppose you would prefer the complete refactor of the page (Auth object, Device object, getting rid of all select/update DB, etc etc), but I don't feel comfortable at all starting this, so I started "step by step" for the moment. 

Let me know,

Thx,

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [X] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #8868, opened 2018-07-04
Laravel migrations

pheinrichs:
This branch is functional currently but the validation errors are through the roof when using the migrations.

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

PR #9348, opened 2018-10-19
New User Management

murrant:
image

Waiting on #8868 and #9422 some things really need tests here.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Does not collide with #9313, utilizes edituser still to manage device permissions.


murrant:
@TheGreatDoc nope, thanks :)

laf:
Doesn't save when unselecting Can Modify Password.

I'm not a fan of the new user / edit user form - it looked better as it was when it was a bit more condensed and concise, just seems to be trying to take up the entire page for the sake of it now.

The old edit user / add user pages still exist.

Does anything need to be done with `./adduser.php` or `install.php` add user?

Could you make it so that deleting a user doesn't refresh the page and just removes the row. We do this in the alert-rules if you want a code example to look at.

Aside from that, all other options adding / deleting saving / updating works for me.

murrant:
I didn't look at the old one, just put the form together, I'll take a look.  Could definitely remove without refresh.

laf:
> Perhaps you like modal style better?

No opinion either way. Edit could work with it but add user won't as you could be on any page.

laf:
Looks better now :)

So just the not being able to unselect Allow user to change password option. You can enable it and it saves but not disable it.

I realise the old edituser page needs to stay as the perms are done through it but I think it would be good to redirect the standard access or /edituser/ to this new page + change the link for `Change...` in the permissions page to point to the new users mgmt page.

Same question about it affecting `./adduser.php` or `install.php` stands just in case.

murrant:
@laf I also need to add 2fa unlock and remove.

murrant:
no affect to ./adduser.php and install.php, doesn't touch that code.

voipmeister:
Hi @murrant , do you still need testers for these changes?

murrant:
@voipmeister yes, but this is out of date at the moment.  I'll update it sometime.  I was waiting so I could added automated testing of the two factor auth stuff since it has several tricky scenarios.  That is almost ready.

voipmeister:
Understood, if you mention me here I'd be happy to do some testing 🙂

laf:
@murrant You need this testing now?

murrant:
I think so.  I was attempting to write some dusk tests, but I'm really bad at that.

html/pages/edituser.inc.php
```diff
@@ -11,49 +10,46 @@
...
+if (!LegacyAuth::user()->isAdmin()) {
     include 'includes/error-no-perm.inc.php';
 } else {
     if ($vars['user_id'] && !$vars['edit']) {
-        /** @var User $user */
-        $user = User::find($vars['user_id']);
-        $user_data = $user->toArray(); // for compatibility with current code
-
+        $user_data = LegacyAuth::get()->getUser($vars['user_id']);
         echo '<p><h2>'.$user_data['realname']."</h2><a href='edituser/'>Change...</a></p>";
```
laf: This needs to be updated to `/users`

murrant:
Fixed all those (and some other items, check the log if curious)

Still not sure what to do with tests, I'll probably just remove them or do something simple.

murrant:
We have basic login tests with and without 2fa, so that is better than nothing :)

murrant:
@voipmeister ready for testing

voipmeister:
> 
> 
> @voipmeister ready for testing

I missed this, will try to test this weekend!

chasgames:
Nice I like this view :+1: 

Just noting two minor issues on my instance, unfortunately i have no dev environment setup yet to test any fixes

1 - LibreNMS logo changes, i have dark mode set, so every other page pulls (https://LIBRENMS/images/librenms_logo_dark.svg), but on manage users it pulls a different logo (https://LIBRENMS/images/librenms_logo_light.svg)

2 - The custom menu doesn't work for anything under https://LIBRENMS/users , i'm not sure but i wonder if LegacyAuth::get()->canManageUsers() blocks it 
https://community.librenms.org/t/how-to-add-a-custom-menu-item/3761/4?u=chas

murrant:
1. Must be some disparity the way the code gets the logo...
2. Custom menu support isn't in the new menu system.  Could probably be added back...

chasgames:
Ahh thanks, 2. is rather indispensable for me, hopefully for others too as currently the only way to share custom aggregations e.g https://librenms/iftype/type=customportdescription/ and other links which don't fall under any other menu.

Thanks for fixing my other code too! I wasn't sure if LegacyAuth was still a thing :)

murrant:
@chasgames you mean this menu??
image

chasgames:
Ahh i remember now exploring custom_descr a few years ago, i forgot it automatically adds to the menu.

I think the reason was we have around 20 different aggregations, so putting a little symbol next to it (Font Awesome icons) made it a bit more user friendly for the rest of the users so i put it in the custom menu.

I do use it for portal links to other applications like an ip manager, and to my weather animation (https://community.librenms.org/t/weathermap-histogram/1052/4?u=chas) but i suppose this could live somewhere else.

But good point i could switch back to the custom_descr in config.php if the new menu system is not supporting it

murrant:
@chasgames https://github.com/murrant/librenms/commit/f47e5eac2baf1627977e0eae3f937ac08fd8b63e

Also, logo theme was already fixed in that branch before you reported :)

chasgames:
Sorry to keep updating this one , just noticed another difference;

Majority of my usernames have the dot between the firstname and lastname. Creating a new user today, but dots no longer are allowed. I'm struggling to see where this validator is in the code though.

image

murrant:
app/Http/Requests/StoreUserRequest.php

There are better spots than this.  Often I miss comments on closed PRs ;)

murrant:

https://community.librenms.org/t/dots-in-username/12887/2

PR #6388, opened 2017-04-10
newdevice: Eaton/MGE UPS reorganization

murrant:
MGE split Galaxy (and Upsilon) to apc-mgeups
All others to eaton-mgeups
Remove mge ups from eatonups os.
Remove powerware it is a duplicate of eatonups

Not ready for commit!  No changes to sensor discovery/polling, other than powerware removal.
I need feed back to know if I have categorized these UPS correctly.

mgepdu.snmprec and mgeups-galaxy.snmprec are updated from user stats.  The rest are guessed.

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you signed the [Contributors agreement](http://docs.librenms.org/General/Contributing/) - please do NOT submit a pull request unless you have (signing the agreement in the same pull request is fine). Your commit message for signing the agreement must appear as per the docs.
- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

PR #9356, opened 2018-10-20
fix: #9301: Eaton/APC Galaxy seems to be the exception with divisor 10 for voltage

PipoCanaja:
Hello

This PR is a follow up of #9301.

The APC Galaxy family was Eaton but is now APC. As a first step, I now corrected the regexp and exclusion regexp so that it ends up with apc-mgeups OS. 
Then the RFC 1628 code is adapted for apc-mgeups. 

@laf : Concerning #9301, should'nt I have seen with the tests that I was breaking things ? It never came up, neither on my server nor during the PullRequest process ?
pre-commit.php skips all tests now in fact. Is it normal or is it something misconfigured on my server ?

Bye

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hello again 

@laf, @murrant , it seems that #6388 was meant to do some cleaning on the eaton/apc devices. And reading at the yaml files, it seems that the Galaxy device (which is now APC) should be apc-mgeups instead of eaton-mgeups. But the yaml file for eaton-mgeups lacks a regexp exclusion for Galaxy...
Do I get it right ? 

PipoCanaja

PipoCanaja:
Yep ,right, the files are existing but they only include 2 lines : sysDescr and sysObjectID which means that it cannot catch anything about divisors ... 

I did add the rss notification as asked. I did not put any date in the text cause I don't know when this PR would be merged. 

PipoCanaja

PipoCanaja:
And test data for Galaxy 7000 is now in this PR.

murrant:
When should we merge this?  Release is already sneaking up on us.

PR #9359, opened 2018-10-22
New Location Geocoding

murrant:
- Fix location so it is a regular database relation (this allows multiple devices to be accurately linked to one location and saves api calls)
- Parse coordinates from the location more consistently
- Add settings to webui
- ~~Used [PHP Geocoder](http://geocoder-php.org/), which has lots of backends and is well tested. (also includes reverse and geoip)~~
- Google Maps, Bing, Mapquest, and OpenStreetMap supported initially.
- Default to OpenStreetMap, which doesn't require a key.  They will liberally hand out bans if you exceed 1 query per second though.
- All other Geocoding APIs require an API key. (Google requires a credit card on file, but seems to be the most accurate)
- Update all (I think) sql queries to handle the new structure
- Remove final vestiges of override_sysLocation as a device attribute
- Update existing device groups and rules in DB
- Tested all APIs with good/bad location, no/bad/good key, and no connection.
- Cannot fix advanced queries that use location

This blocks #8868

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


sql-schema/270.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE devices ADD location_id int NULL AFTER location_id;
```
laf: Invalid AFTER location_id :)

sql-schema/270.sql
```diff
@@ -0,0 +1,11 @@
...
+ALTER TABLE locations MODIFY lat double(10,6);
+ALTER TABLE locations MODIFY lng double(10,6);
+
+DELETE t1 FROM locations t1 INNER JOIN locations t2 WHERE t1.id < t2.id AND t1.location = t2.location;
+INSERT INTO locations (location, timestamp) SELECT devices.location, NOW() FROM devices WHERE devices.location IS NOT NULL AND NOT EXISTS (SELECT location FROM locations WHERE location = devices.location);
+UPDATE devices INNER JOIN locations ON devices.location = locations.location SET devices.location_id = locations.id;
+ALTER TABLE devices DROP location;
```
laf: I know for us, this is going to break groups and I think removes the ability to use location for groups now as the `locations` table can't reference back to a device id.

murrant: We could join the table for all queries.

I'll check alerts and groups and see what our options are.

Bleh, new this was going to be a tricky one.

murrant: Fixed

murrant:
Dang, PHP Geocoder modules don't support php 5.6...  Reverting :(

html/includes/print-menubar.php
```diff
@@ -187,11 +187,11 @@
                 echo('
                     <li role="presentation" class="divider"></li>
                     <li class="dropdown-submenu">
-                    <a href="#"><i class="fa fa-map-marker fa-fw fa-lg" aria-hidden="true"></i> Geo Locations</a>
+                    <a href="locations"><i class="fa fa-map-marker fa-fw fa-lg" aria-hidden="true"></i> Geo Locations</a>
```
laf: This stops you being able to click on a mobile to expand that menu. When you do it just takes you to the locations url.

includes/polling/core.inc.php
```diff
@@ -79,14 +79,18 @@
     }
 }
 
-if ($poll_device['sysLocation'] && $device['location'] != $poll_device['sysLocation'] && $device['override_sysLocation'] == 0) {
-    $update_array['location'] = $poll_device['sysLocation'];
-    $device['location']       = $poll_device['sysLocation'];
-    log_event('Location -> ' . $poll_device['sysLocation'], $device, 'system', 3);
-}
+if ($device['override_sysLocation'] == 0 && $poll_device['sysLocation']) {
```
laf: This now stops you overriding the devices location but still have it geo lookup what that value is.

sql-schema/271.sql
```diff
@@ -0,0 +1,18 @@
+ALTER TABLE devices ADD location_id int NULL AFTER location;
+ALTER TABLE locations MODIFY lat double(10,6);
+ALTER TABLE locations MODIFY lng double(10,6);
+
+DELETE t1 FROM locations t1 INNER JOIN locations t2 WHERE t1.id < t2.id AND t1.location = t2.location;
+INSERT INTO locations (location, timestamp) SELECT devices.location, NOW() FROM devices WHERE devices.location IS NOT NULL 
+AND NOT EXISTS (SELECT location FROM locations WHERE location = devices.location);
```
laf: This is on a new line and it shouldn't be.

sql-schema/271.sql
```diff
@@ -0,0 +1,18 @@
...
+UPDATE devices INNER JOIN locations ON devices.location = locations.location SET devices.location_id = locations.id;
+ALTER TABLE devices DROP location;
+
+ALTER TABLE locations MODIFY location varchar(255) NOT NULL;
+CREATE UNIQUE INDEX locations_location_uindex ON locations (location);
+
+UPDATE alert_rules SET builder=REPLACE(builder, 'devices.location', 'locations.location');
```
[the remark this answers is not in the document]
murrant: I don't think I can properly update raw queries, because I would need to insert the glue too.

murrant:
As far as I'm concerned this is ready to merge.

laf:
limited testing but I think their is a bug (aside from it seems to favour openstreetmap as the default, from the code it looks like google should be the default still).

If an address doesn't have it's geo cords found then it gets set to 0.00000, 0.000000 and the date/time updated for the 2 day cache. If you update the date/time to anything > 2 days I don't think it will then ever attempt a geo lookup again. It's only when lat/lng are null as well it will do this.

laf:
> As for OpenStreetMap, is the setting set in your DB?

Seems so yes.

> Yes, after the coordinates are non-null, we stop trying to resolve GPS coords, this is by design. You cannot edit the location string of a location, so after they have been resolved there is no need to resolve them again. If the sysLocation string of a device changes, a new location will be created. I hope to add the ability to manually adjust gps coordinates of locations sometime, but this PR seemed big enough.

But what about if the mapping engine doesn't have the address one day but does the next, at present it would mean the coords would never be found.

I'm going to merge in for now anyway.

PR #9364, opened 2018-10-22
Redesign alert template modal

crazy-max:
* "Attach template rules" modal merged to "Alert template" modal
* Use select2
* Autocompletion available
* Fix orphans occurrences in db while removing rule or template
* Use 'pre' font family for alert template textarea

## Preview

image

image

image


TheGreatDoc:
Nice! But there is a little issue (that was fixed by me :P) and now is back.

You can select the same rule in two different templates. That will make the rule to use the template with lowest ID.

crazy-max:
@TheGreatDoc Yes indeed, is it better now ?

image

html/includes/modal/alert_template.inc.php
```diff
@@ -37,6 +37,10 @@
                             <label for="template">Template: </label>
                             <textarea class="form-control" id="template" name="template" rows="15"></textarea>
                         </div>
+                        <div class="form-group">
+                            <label for="rules_list">Attach template to rules: </label>
+                            <select id="rules_list" name="rules_list" class="form-control" multiple data-live-search="true" data-max-options="10"></select>
```
laf: Do we need to limit to 10 here?

html/js/bootstrap-select.min.js
```diff
@@ -0,0 +1,9 @@
+/*!
```
laf: This looks like it's better maintained than select2 which we currently use - any reason you've picked it though as it does mean we now load 2 libraries for the same thing.

crazy-max: I didn't know there was another library like this. I'll look into replacing select2.

html/includes/modal/alert_template.inc.php
crazy-max: I have kept the [old behavior](https://github.com/librenms/librenms/blob/master/html/includes/modal/attach_alert_template.inc.php#L35) but we can definitely remove this limit.

html/js/bootstrap-select.min.js
laf: Don't spend too long. From the looks of this one, it might be better to convert to this.

crazy-max: I wanted to say I will check where select2 is used in the project and replace with it with bootstrap-select :)

murrant: Select2 is pretty mature.  I've been building a lot of ajax request handlers for it because previously the code just loads everything in page.  Sometimes causing nightmare page load times.

Will probably will switch to something virtual dom based (vue.js/react) at some point.

laf: Yeah we'd discussed this off github and this new one doesn't support ajax anyway so it'll be converted to select.

crazy-max: @murrant @laf Ok np I will switch to select2 and try to mimic the bootstrap-select UX

crazy-max:
@laf @murrant Ok I have mimic bootstrap-select using select2 :

image

What do you think ?

crazy-max:
@laf For this kind of feature yeah you right

crazy-max:
:tada:

murrant:

PR #9365, opened 2018-10-22
Add F5 APM active sessions graph

JohnSPeach:
Add an APM section to the F5 graphs, with the number of active sessions graphed.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/os/f5.inc.php
+++ b/includes/polling/os/f5.inc.php
@@ -11,7 +11,7 @@ if (is_numeric($sessions)) {
     $rrd_def = RrdDefinition::make()->addDataset('sessions', 'GAUGE', 0);
 
     $fields = array(
-        'sessions' => $sessions,        
+        'sessions' => $sessions,
     );
 
     $tags = compact('rrd_def');


sql-schema/269.sql
```diff
@@ -1 +1 @@
-INSERT INTO `graph_types`(`graph_type`, `graph_subtype`, `graph_section`, `graph_descr`, `graph_order`) VALUES ('device',  'apm_sessions',  'apm',  'Active Sessions',  0);
+INSERT INTO `graph_types`(`graph_type`, `graph_subtype`, `graph_section`, `graph_descr`, `graph_order`) VALUES ('device',  'bigip_apm_sessions',  'apm',  'Active Sessions',  0);
```
TheGreatDoc: We are already in 269 Schema, you cant modify it. You should create 270.sql.

Also, if you are modifiying a current existing graph type (changing apm_sessions to bigip_apm_sessions) you must do an UPDATE instead of INSERT

JohnSPeach:
Reverted 269.sql and added 270.sql

laf:
This isn't adding anything new it's just renaming the data currently stored, why is that?

JohnSPeach:
> This isn't adding anything new it's just renaming the data currently stored, why is that?

It needed the bigip prefix for the graphs to display automatically. Prior to this the APM tab had blank graphs and once I clicked on one I had to manually select bigip_apm_sessions.

includes/polling/os/f5.inc.php
```diff
@@ -14,7 +14,6 @@
         'sessions' => $sessions,
     );
 
-    $tags = compact('rrd_def');
```
murrant: You accidentally deleted this line, please restore it.

JohnSPeach: Replaced missing line

JohnSPeach:
TBH, I'm not entirely sure. I changed the name everywhere and I'm not in the office today. I can certainly see how it goes with just the graph file......

murrant:
Thanks, that would be great.  It would prevent users from losing historical data (which they haven't been able to view)...

JohnSPeach:
Fresh pull request created against the current master

PR #9366, opened 2018-10-22
Show device "features" in correct column

Jellyfrog:
Align the device overview with the device specific overview.

Before:
image

After:
image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


html/includes/table/devices.inc.php
```diff
@@ -207,9 +207,13 @@
         $location = substr($device['location'], 0, 32);
     }
 
+    if ($device['features']) {
```
laf: You can just wrap device features in brackets inline below rather than setting the variable with it.

Jellyfrog: It simply follows how it's done in other files. The argument is null often so then it would be an empty `()`

html/includes/table/devices.inc.php
```diff
@@ -207,9 +207,13 @@
...
+        $device['features'] = '('.$device['features'].')';
+    }
+
     if ($subformat == 'detail') {
-        $platform = $device['hardware'] . '<br>' . $device['features'];
-        $os = $device['os_text'] . '<br>' . $device['version'];
+        $platform = $device['hardware'];
+        $os = $device['os_text'] . '<br>' . $device['version'] . ' ' . $device['features'];
```
laf:
```suggestion
        $os = $device['os_text'] . '<br>' . $device['version'] . ' ' . empty($device['features']) ? '' : "({$device['features']})";
```

html/includes/table/devices.inc.php
laf: Ternary :)

html/includes/table/devices.inc.php
Jellyfrog: @laf fixed

murrant:
Why is this the correct column?  I think there is confusion in the code base what "features" actually is.  Some treat it as hardware features some as software features.

Jellyfrog:
> Why is this the correct column? I think there is confusion in the code base what "features" actually is. Some treat it as hardware features some as software features.

It aligns it with how it is in the device overview and the device "overlib"

Jellyfrog:
> Good point. My guess is it was done this way originally to shorten the Operating System column and use unused space in the platform column.

That was my guess too, think we should keep it uniformed tho, it's a bit confusing now.

murrant:

PR #9370, opened 2018-10-23
Show port description and dns name in FDB table

Jellyfrog:
- Added Port Description field to FDB Table
- Added DNS Name field to FDB Table
- Fixed sorting by port in FDB Table

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/table/fdb-search.inc.php
+++ b/html/includes/table/fdb-search.inc.php
@@ -60,10 +60,12 @@ if (isset($vars['searchPhrase']) && !empty($vars['searchPhrase'])) {
         $where  .= ' AND `F`.`mac_address` LIKE ?';
         $param[] = $mac_search;
     } else {
-        $where  .= ' AND (`V`.`vlan_vlan` = ? OR `F`.`mac_address` LIKE ? OR `P`.`ifAlias` LIKE ?)';
+        $sql .= " LEFT JOIN `ipv4_mac` AS `M` USING (`mac_address`)";
+        $where  .= ' AND (`V`.`vlan_vlan` = ? OR `F`.`mac_address` LIKE ? OR `P`.`ifAlias` LIKE ? OR `M`.`ipv4_address` LIKE ?)';
         $param[] = (int)$search;
         $param[] = $mac_search;
         $param[] = '%' . $search . '%';
+        $param[] = '%' . gethostbyname(trim($vars['searchPhrase'])) . '%';
     }
 }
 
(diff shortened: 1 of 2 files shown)


murrant:
How about adding the field with `data-visible="false"` so it can be displayed if wanted, but is hidden by default.

Also, it feels like this field should be next to the port (interface) field.  That would need to be fixed in the search table too.

Jellyfrog:
> How about adding the field with `data-visible="false"` so it can be displayed if wanted, but is hidden by default. It already displays port (which happens to have ifAlias in the popup).

I think it should be visible by default, its good information.

> Also, it feels like this field should be next to the port (interface) field. That would need to be fixed in the search table too.

Agree, but now it follows how other pages sort the fields, should I add a commit moving the description field for all pages? And maybe rename it to "Port Description" ?

jozefrebjak:
For me working perfectly. Nice feature

murrant:
The reason I suggested hiding it by default is it is duplicate information.

html/includes/table/fdb-search.inc.php
```diff
@@ -56,23 +57,24 @@
...
         $where  .= ' AND `F`.`mac_address` LIKE ?';
         $param[] = $mac_search;
     } else {
-        $where  .= ' AND (`V`.`vlan_vlan` = ? OR `F`.`mac_address` LIKE ?)';
+        $sql .= " LEFT JOIN `ipv4_mac` AS `M` USING (`mac_address`)";
+        $where  .= ' AND (`V`.`vlan_vlan` = ? OR `F`.`mac_address` LIKE ? OR `P`.`ifAlias` LIKE ? OR `M`.`ipv4_address` LIKE ?)';
         $param[] = (int)$search;
         $param[] = $mac_search;
+        $param[] = '%' . $search . '%';
+        $param[] = '%' . mres(gethostbyname(trim($vars['searchPhrase']))) . '%';
```
laf: You shouldn't need to use `mres()` here, `$vars` already has done that.

html/pages/device/ports/fdb.inc.php
```diff
@@ -7,7 +7,9 @@
             <th data-column-id="mac_address">MAC address</th>
             <th data-column-id="ipv4_address">IPv4 Address</th>
             <th data-column-id="interface">Port</th>
+            <th data-column-id="description">Description</th>
             <th data-column-id="vlan">Vlan</th>
+            <th data-column-id="dnsname">DNS Name</th>
```
laf: For me this is fairly useless so just has N/A throughout. I imagine it's useful for others but not all.

Jellyfrog: Hidden by default now.

PR #9375, opened 2018-10-24
Support for Alcatel OmniPCX

PipoCanaja:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/alcatel-omnipcx.yaml
```diff
@@ -0,0 +1,13 @@
+os: alcatel-omnipcx
+group: alcatel
+text: 'Alcatel OmniPCX'
+type: server
```
laf: I wonder if appliance would better fit here.

includes/definitions/discovery/alcatel-omnipcx.yaml
```diff
@@ -0,0 +1,35 @@
...
+                    descr: 'Pbx Role'
+                    index: 'pbxRole{{ $index }}'
+                    state_name: pbxRole
+                    states:
+                        - { descr: Main , graph: 0, value: 1 , generic: 0 }
+                        - { descr: Standby , graph: 0, value: 2 , generic: 0 }
+                        - { descr: Active PCS , graph: 0, value: 3 , generic: 0 }
+                        - { descr: Inactive PCS , graph: 0, value: 4 , generic: 0 }
+                        - { descr: Unknown , graph: 0, value: 0 , generic: 3 }
+                -
```
laf: You need states values fr all state sensors.

includes/definitions/alcatel-omnipcx.yaml
```diff
@@ -0,0 +1,13 @@
+os: alcatel-omnipcx
```
laf: This can be renamed to `omnipcx`

includes/definitions/discovery/alcatel-omnipcx.yaml
PipoCanaja: I was afraid of this. But then, how could I have a sensor that just displays a "count" ? not a percentage, not a temperature, just a "count" ?

includes/definitions/alcatel-omnipcx.yaml
PipoCanaja: OK, will be.

includes/definitions/alcatel-omnipcx.yaml
PipoCanaja: That makes sense yes. Will be changed.

includes/definitions/discovery/alcatel-omnipcx.yaml
PipoCanaja: I know, that's why I expected this ;) but the unit of the state sensor is # and it is exactly displayed like I would like to see ;) Except the title "State" of course ;)

PipoCanaja: So, to recap, what should I do ? create a new type of sensor, named count, that would more or less share its code with state, except the state translation, and the event when the value changes ?

murrant: @PipoCanaja yeah, best solution would be to create a new sensor type.

It would not share much with states.  Other than the display could be similar.

PipoCanaja: All right, will have a look at that. Should I add that to the same PR or create a new one ?

PipoCanaja: Did add it to this one, for now. Let's continue the discussion in the main conversation.

PipoCanaja:
New "versatile" type added, unitless, so we can display anything that does not fit the standard sensors.

doc/Developing/os/Health-Information.md
```diff
@@ -31,6 +31,7 @@ Currently we have support for the following health metrics along with the values
 | temperature                     | C                           |
 | voltage                         | V                           |
 | waterflow                       | l/m                         |
+| versatile                       |                             |
```
laf: What's this new unit supposed to show?

laf: Just read in your comment. Not sure about this one. I can definitely see the benefit but how will you deal with things like min max scales etc?

I'd prefer to see one of the other PRs that adds custom OIDs in supported rather than go down this route.

PipoCanaja: Unit would, if we push this concept further, be a parameter of this versatile type. And min/max would be defined as today in the YAML (but of course, auto-compute of the min/max value would not make sense here). 

In a PBX like here, creating a dedicated type for each entry would be too much (we can count PhonesSets, SIP subscriptions, etc etc). But we still need a way to show these. 

What do you mean by "one of the other PRs" ?

laf: Min / max I'm referring to scales rather than thresholds. I.e what the graphs have in the Y axis as the min/max.

8738 and 7925 both try and cover this in a more generic way.

PipoCanaja: Ok.
May be I'll take another route right now, which is to do a standard polling instead of Yaml, with a dedicated page. This would allow this PR to continue its life without creating this graph type that would later on need to be changed/suppressed. 
What do you think ?

murrant: If you want to get this merged, just omit that sensor for now and add it later.

PipoCanaja: In the process of doing that. WIP will be removed when tests are done.

includes/snmp.inc.php
```diff
@@ -468,7 +468,7 @@ function snmpwalk_cache_oid($device, $oid, $array, $mib = null, $mibdir = null,
         $oid               = trim($oid);
         $value             = trim($value, "\" \\\n\r");
         list($oid, $index) = explode('.', $oid, 2);
-        if (!strstr($value, 'at this OID') && isset($oid) && isset($index)) {
```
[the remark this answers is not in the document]
murrant: You probably need to leave the 'at this OID', not sure though.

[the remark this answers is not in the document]
PipoCanaja: Yes I did keep both "at this oid" and "isset(oid)" tests. But I did put the review I line to early in github ;) 
-         if (!strstr($value, 'at this OID') && isset($oid) && isset($index)) {
+         if (!strstr($value, 'at this OID') && isset($oid)) {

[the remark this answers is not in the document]
PipoCanaja: just added a check for oid not empty (that should never happen).

PipoCanaja:
OK for review.

includes/snmp.inc.php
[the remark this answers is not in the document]
murrant: well isset($oid) is redundant now :)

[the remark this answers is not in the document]
PipoCanaja: correct ;)

PR #9381, opened 2018-10-24
Graph of overall poller performance

murrant:
Add new tab on pollers page/menu.
Improve tab layout/display.
Move (copy) poll-log to a tab.
FIXED: Graphs have odd brokenness for me, probably need to build them some other way.

Not 100% sure the calculations are correct.

image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


TheGreatDoc:
I will say that at least total time is not correct or Im reading it wrong:

image

3k seconds avg ;)

murrant:
Remember that this is not real time, it is poller time.  Try dividing it by the number of poller threads.

TheGreatDoc:
I have 16 threads. 3070 / 16 = 192 seconds.

In my dev install, with 6 devices, poller usually finish in ~10 seconds

murrant:
@TheGreatDoc Try this:
```
 find rrd -name poller-perf.rrd| xargs -I{}  rrdtool lastupdate --daemon unix:/run/rrdcached/rrdcached.sock {}|grep :|cut -d' ' -f2|awk 'BEGIN { sum=0 } { sum+=$1 } END {print sum }'

1835.59
```

murrant:
Darn test code...

TheGreatDoc:
I have rrdcached for production. Dev install is without rrdcached. 
How do I get the same without rrdcached?

murrant:
@TheGreatDoc delete the server option.  Also, I fixed the issue.

TheGreatDoc:
Tested and looks awesome.

`Cur` and `Min` shows 0 like yours. Still working or it is going to be 0? If is going to be 0, can we remove them?

murrant:
https://stackoverflow.com/questions/43267942/last-value-becomes-0-sometimes-in-rrd-graph-when-using-if-and-un-to-set-unknown

resources/views/layouts/menu.blade.php
```diff
@@ -364,13 +364,14 @@
                         <li><a href="{{ url('authlog') }}"><i class="fa fa-shield fa-fw fa-lg" aria-hidden="true"></i> Auth History</a></li>
                         <li role="presentation" class="divider"></li>
                         <li class="dropdown-submenu">
-                            <a href="#"><i class="fa fa-th-large fa-fw fa-lg" aria-hidden="true"></i> Pollers</a>
```
laf: The one thing with this change is it makes the sub-menu unusable on a mobile. Clicking the Pollers drop down now just takes you to the pollers page so the sub items in the menu are now only accessible by going to the main pollers page and clicking through the tabs from there.

I'll merge anyway but personally, I think it would be good to restore that.

murrant:

PR #9391, opened 2018-10-26
Change unpolled devices toast to be based on rrd step

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/Checks.php
+++ b/app/Checks.php
@@ -82,9 +82,10 @@ class Checks
                 Toastr::error("<a href='notifications/'>$notification->body</a>", $notification->title);
             }
 
-            $rrdstep = Config::get('rrd.step', 300);
-            if (Device::isUp()->where('last_polled', '<=', Carbon::now()->subSeconds($rrdstep * 2))->exists()) {
-                Toastr::warning('<a href="poll-log/filter=unpolled/">It appears as though you have some devices that haven\'t completed polling within the last 15 minutes, you may want to check that out :)</a>', 'Devices unpolled');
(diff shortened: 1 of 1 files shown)


app/Checks.php
```diff
@@ -82,7 +82,8 @@ public static function postAuth()
                 Toastr::error("<a href='notifications/'>$notification->body</a>", $notification->title);
             }
 
-            if (Device::isUp()->where('last_polled', '<=', Carbon::now()->subMinutes(15))->exists()) {
+            $rrdstep = Config::get('rrd.step', 300);
+            if (Device::isUp()->where('last_polled', '<=', Carbon::now()->subSeconds($rrdstep * 2))->exists()) {
```
laf: Any reason to change this from 15 mins to in essence 10 mins. There was no real reason to pick 15m in the first place but interested why change it.

murrant: Yes.

murrant: But to answer your question, because I was changing from a flat number to a factor of rrd step, and I didn't think 3x made sense.  If someone has 1hr polling, it has to fail for 3hrs before the alert.

Maybe 3x does make sense.  Open to thoughts of course.

laf: I don't think many people are going to do 1 hour polling. 1 minute is more likely so this would lead to the message after 2 mins.

Either way, can you also update the actual message returned as it's hard coded to 15 minutes.

murrant: The reason it was added is someone is doing 30 minute polling.  1hr was just round numbers.

PR #9393, opened 2018-10-26
Update Installation-CentOS-7-Nginx.md

TheGreatDoc:
By default on CentOS 7, /var/lib/php/session is root:root.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


TheGreatDoc:
Better to just create a new session dir and set it up in php.ini? 

Or whats the correct fix?

murrant:
The user should have permissions after following the docs.

Long term, the plan is not to use the phone session directory.

murrant:
image

TheGreatDoc:
Sorry but I dont understand what you mean :(

Seem like "now", by default, /var/lib/php/session is owned by root:root.

After chgrp to apache works like a charm, but as I said by default and following the guide step by step, is root group.

laf:
As part of the packer installs I have: `chgrp apache /var/lib/php/session/`

It does look like it's needed. Been a while since I installed from scratch :(

PR #9399, opened 2018-10-29
Use which to solve executables path

crazy-max:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


crazy-max:
The same apply to librenms/librenms-agent#209

crazy-max:
@laf You right I just see that! And it is currently available on [librenms-agent repo](https://github.com/librenms/librenms-agent/blob/master/agent-local/nfsstats) so I think we should delete to avoid misunderstanding

PR #9401, opened 2018-10-30
bug-fix and new features routeros

takenalias:
Fixed incorrect OID for rate, renamed rate to TX-Rate as per update from mikrotik.

Added support for link distance

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/OS/Routeros.php
```diff
@@ -170,10 +172,10 @@ public function discoverWirelessRate()
             $sensors[] = new WirelessSensor(
                 'rate',
                 $this->getDeviceId(),
-                '.1.3.6.1.4.1.14988.1.1.1.8.1.13.' . $index,
```
laf: Is this definitely supposed to be moved?

takenalias: Yep. Last time I put the wrong OID into the 60g rate. That and the fact that the station end, the index seems to be 4 not 1, meant I had to get it to look at the station table as well. Initially the raw rate and the distance values   werealmost identical mesurements 23m and 2.3gigbit which is what made me think it was all fine.

image

And for the normal wireless. 
image
image

PR #9410, opened 2018-10-31
format server graphs with units for memory and disk usage

VirTechSystems:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


html/includes/common/server-stats.inc.php
```diff
@@ -87,34 +87,96 @@
 
         $i = 0;
         foreach ($mem as $m) {
+            $mem_used = $m['used'];
```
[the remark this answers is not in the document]
VirTechSystems: That's a cool function.  I'm still a little unfamiliar with the file structure.  Would I put it in html/includes/functions.inc.php and call it from there?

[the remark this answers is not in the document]
VirTechSystems: Or maybe modify the format_bi function in common.php in includes/ to return a hash of the bytes, units, and combined value so it can be used here?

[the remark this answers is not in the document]
laf: Does `format_bi()` not do what you need anyway?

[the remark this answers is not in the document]
VirTechSystems: Only partially.  The graph widget requires the current value and symbol to be passed to it in different values.  That's why I made the function return a hash of all the parts. 

The other option would be a substring of what format_bi returns. Trim off the last two characters or something like that.

Neither way seems great to me.  The hash is more flexible, but the substring is simple.  What are your thoughts on getting the converted bytes and symbol as separate values?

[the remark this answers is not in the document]
laf: I'd say the substring (or a preg_match) would be more preferred. You can put that in another function which calls format_bi.

Alternatively the function I linked to could be used, it's a lot simpler and that can return the hash.

includes/common.php
```diff
@@ -634,6 +634,25 @@ function format_bi($value, $round = '2', $sf = '3')
     return number_format(round($value, $round), $sf, '.', '').$ext;
 }
 
+function format_graph_bi($current, $total, $round = 2)
```
laf: So very close, you can't have this function accept 2 values to format. It should be a totally generic function that just accepts an int value and round value. You'll have to call the function twice but that's fine.

includes/common.php
```diff
@@ -634,6 +634,27 @@ function format_bi($value, $round = '2', $sf = '3')
     return number_format(round($value, $round), $sf, '.', '').$ext;
 }
 
+function format_graph_bi($value, $round = 2, $pow = 0)
```
laf: I'm confused at why you're making this function more complicated than the example on stack overflow?

VirTechSystems: The $pow parameter is to make sure both the current value and total value end up being the same unit, otherwise the graph represent the values incorrectly.  There is the possibility of ending up with something like GB for one value and TB for another depending on the current values and totals and how they divide.

VirTechSystems: Any thoughts on this?  Not sure what direction to go with any changes.

murrant: Why is that important if they are labeled properly?

VirTechSystems: It makes the graphs much more readable not having to convert everything mentally from MB to GB or TB to see what a servers resources look like.

Plus, it shows that 10% of your disk free might be 100GB and isn't a critical problem.

PR #9414, opened 2018-11-02
Improve documentation for service plugins

2E0PGS:
This pull request documents a list of which plugins come pre installed.

This also documents where you can find the documentation, source code and publisher of the included plugins.

Originally discussed with `laf#1551` aka @laf on Discord. Since I was having trouble finding our what each plugin really was. One example is the following:

> I have no clue what the differences are between or where they come from `ssl_cert`, `cert_expire` however i do know `http` which is actually `http_check` can check cert expiry too.

Note I did add the `[TOC]` text to the head of the document since the page is longer now this should help website visitors navigate the page easier.

Cheers,
2E0PGS

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


doc/Extensions/Services.md
```diff
@@ -1,12 +1,144 @@
...
+# Setting up services
 
 Services within LibreNMS provides the ability to use Nagios plugins to perform additional monitoring outside of SNMP.
 
 **These services are tied into an existing device so you need at least one device to be able to add it
 to LibreNMS - localhost is a good one. This is needed in order for alerting to work properly.**
 
+## Pre installed plugins
+
+List of pre installed plugins in the format: \<name\> - \<origin\>
```
TheGreatDoc: Quick comment: They dont come pre installed. 

Afaik those plugins comes with the nagios-plugins/nagios-plugins-all packages for Ubuntu and CentOS respectively.

doc/Extensions/Services.md
```diff
@@ -1,12 +1,144 @@
...
+
+Plugins will only load if they are prefixed with "check_" and they have that prefix stripped out when displaying in the "Add Serice" GUI "Type" dropdown list.
+
+* [ajp](https://github.com/bzed/pkg-nagios-plugins-contrib/tree/master/check_ajp) - pkg-nagios-plugins-contrib
```
laf: I did mention this on discord but I really don't think we should be maintaing a list like this.

2E0PGS: > The list of plugins can change, honestly I think just detailing about what they are where people can find them generally is the best option

You did write that. Maybe I interpreted that wrong however I still think a list like this is super useful. I personally have visited this myself several times as a reference since I created it.

If not I guess I could always move this to my own repo specifically for maintaining a list of plugins found in librenms. However I think others would benefit from finding it easily here.

At least my information regarding how plugins are loaded should be considered useful. That isnt documented anywhere to my knowledge so someone trying to script their own would be very confused until they found out it must be prefix with `check_`.

doc/Extensions/Services.md
2E0PGS: I cannot find such package under Ubuntu repos however their does appear to be one for centos.

>To avoid confusion, I'll swap pre installed with supported or anything else to dont let the user think they come already with the LibreNMS install.

Yes that is better wording.

However with the VM images they are pre packaged.

doc/Extensions/Services.md
laf: That's fine, I won't stand in the way of this being merged but I won't be merging it myself.

doc/Extensions/Services.md
murrant: In many distros the package is called monitoring-plugins

2E0PGS:
I have added a new commit which instead points to the two main sources of documentation and the packages for the two distro's.

Note not all plugin documentation is from `monitoring-plugins` so I also linked `pkg-nagios-plugins-contrib` which is where you will find the other half of plugin docs.

Hopefully this is better. I guess I got a bit carried away creating documentation I would prefer myself being more centralised however as you said others probably wouldn't maintain it.

PR #9425, opened 2018-11-07
Fixed sysDescr can't get value

jasoncheng7115:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

---

Previous use:
`preg_match('/Router Model: ([\w ]+), Version: ([\S\w\.]+)/', $poll_device['sysDescr'], $tmp_draytek);`

There is no sysDescr String, and the Hardware & Version cannot be obtained.

The polling results:
2018-11-07 10_18_52-router-006 - capture - librenms

---

After testing, I made the following modifications so that he could read and split correctly:
`preg_match('/Router Model: ([\w ]+), Version: ([\S\w\.]+),/', snmp_get($device, '.1.3.6.1.2.1.1.1.0', '-OQv'), $tmp_draytek);`

The fixed polling results:
2018-11-07 10_18_10-router-006 - capture - librenms

### DIFF
--- a/includes/polling/os/draytek.inc.php
+++ b/includes/polling/os/draytek.inc.php
@@ -3,7 +3,7 @@
  * draytek.inc.php
  * @author     Jason Cheng <sanyu3u@gmail.com>
  */
-preg_match('/Router Model: ([\w ]+), Version: ([\w\.]+)/', $poll_device['sysDescr'], $tmp_draytek);
+preg_match('/Router Model: ([\w ]+), Version: ([\S\w\.]+),/', snmp_get($device, '.1.3.6.1.2.1.1.1.0', '-OQv'), $tmp_draytek);
 $hardware = $tmp_draytek[1];
 $version  = $tmp_draytek[2];
 unset($tmp_draytek);


laf:
`$poll_device['sysDescr']` exists with the output for the lookup of the sysDescr value in snmp which is all your snmpget is doing.

It would be good to see a debug from your devices poller output.

jasoncheng7115:
> `$poll_device['sysDescr']` exists with the output for the lookup of the sysDescr value in snmp which is all your snmpget is doing.
> 
> It would be good to see a debug from your devices poller output.

Yes, snmpget can be taken out,

However, in the original draytek.inc.php the value of $poll_device['sysDescr'] is empty.
The hardware and OS cannot be analyzed.

laf:
If you can take the snmp_get out then what are you using in it's place?

includes/polling/os/draytek.inc.php
```diff
@@ -3,7 +3,7 @@
  * draytek.inc.php
  * @author     Jason Cheng <sanyu3u@gmail.com>
  */
-preg_match('/Router Model: ([\w ]+), Version: ([\w\.]+)/', $poll_device['sysDescr'], $tmp_draytek);
```
murrant: If you use `$device['sysDescr']` does it work?  It would be really helpful if you provided the sysDescr from the device.

murrant:
Fixed in #9466

PR #9430, opened 2018-11-08
Add new sysDescr string for AlliedWare Plus products.

luukp:
Release 5.4.8-2.1 of AlliedWare Plus will change the format of the
sysDescr string. This patch updates Allied Telesis yaml files to work with
this change.

Signed-off-by: Luuk Paulussen <luuk.paulussen@alliedtelesis.co.nz>

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


luukp:
The failure above is legitimate, so I will need to revisit this change.

luukp:
I fixed the issue mentioned in my previous comment.  I think it would also be useful to add a simulator file for the new release, but I think it may be better to do that as a separate patch (once the software releases.)

laf:
We need tests for this as well. Can you at least post an example sysDescr output?

mattie47:
@luukp I might not work at ATL now, but feel free to flick me an email if you need help with anything.

mattie47:
@laf 
cc @luukp  @lorranmira

In terms of tests, do you just want a new awplus snmp rec file which contains the new sysDescr? 

E.g. `awplus_1.snmprec` plus generated `awplus_1.json`?

Thanks,

Matt

luukp:
I am in the process of generating snmprec and json files for the new variant, which has not been straightforward due to various issues around snmpsim.

A number of other MIBs now have additional support as well in the new version (e.g. ipNetToPhysical, ipNetToMedia, entityMIB), and I am wondering if it makes sense to include output from these in the new MIB. Obviously some of the MAC/IP data is private, so I'm wondering what the best practice is for the simulated data?

laf:
If you want to just add snmprec files then we can can generate the json data.

To be fair, you could just give us the sysObjectID and sysDescr although full test data is better if it's different than what we already have.

luukp:
I've added the snmprec file to the commit.

PR #9444, opened 2018-11-11
api: Add list_links and get_link api calls

zombah:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/API/Switching.md
+++ b/doc/API/Switching.md
@@ -69,3 +69,81 @@ Output:
   ]
 }
 ```
+
+### `list_links`
+
+Get a list of all or per hostname discovered Links.
+
+Route: `/api/v0/resources/links`
+
+Input:
+
+  - hostname = Either the devices hostname or id
+
+Example:
+```curl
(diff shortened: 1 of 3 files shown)


doc/API/Switching.md
```diff
@@ -69,3 +69,81 @@ Output:
...
+Route: `/api/v0/resources/links`
+
+Input:
+
+  - hostname = Either the devices hostname or id
+
+Example:
+```curl
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/resources/links
+curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/resources/links?hostname=host.example.com
```
laf: I wonder if this route should be `/api/v0/devices/host.example.com/links`?

zombah: Yes resonable, i'll prepare it in follow-up commit.

laf: Can you do it in this one, no point in releasing something that will change, especially if people start using it.

zombah: Yes sure in this PR. Just added as separate call.

PR #9450, opened 2018-11-15
SAF CFM wireless device support

Schouwenburg:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


Schouwenburg:
Hello, it looks like the build failed purely on a style issue. I made multi-line function declaration and calls to work around the style problem of a maximum of 120 characters on a row. So either way I will run into a style issue. Could you please advise or approve?

includes/discovery/entity-physical/entity-physical.inc.php
```diff
@@ -25,6 +65,28 @@
...
+if ($device['os'] == 'saf-cfml4') {
+    $entity_array = array();
+    $device_array = array();
+    echo ' saf-cfml4Anatomy';
+    $oid = '.1.3.6.1.4.1.7571.100.1.1.2.22';
+    $row = 0;
+    $device_array = snmpwalk_cache_oid($device, $oid, $entity_array, 'SAF-MPMUX-MIB');
+    // Descr, VendorType, ContainedIn, Class, ParentRelPos,
+    //  Name, HardwareRev, FirmwareRev, SoftwareRev, SerialNum, MfgName, ModelName, Alias, Alias, AssetID, IsFRU
+    $entity_array[++$row] = return_entity_array('CFM L4', 'CFM L4', '0', 'chassis', '-1', 'Chassis', '', '', '', $device_array[0]['serialNumber'], 'SAF', 'CFM L4', '', '', 'true');
```
murrant: 1. You  don't need to use the $row variable `$entity_array[] = 'something';` just appends it at the next index.
2. foreach would be a lot cleaner instead of for (what if the device has  a different number of responses than you expect.
3. You could probably loop through this in a way that you wouldn't need the function and still have it clean looking.

Schouwenburg:
Hello Tony, thanks for looking into this and the comments. Without the function is indeed possible, but the amount of rows needed was a bit extreme I think I still have a copy of it so I could easily build it back but I’m not sure what would be better. Not needing the function will be cleaner as it’s the only function in that file. Anyway, I’ll use you’re suggestions and issue a new pull request tomorrow.

Thanks,

Janno

Verstuurd vanaf mijn iPhone

Schouwenburg:
I changed includes/discovery/entity-physical/entity-physical.inc.php.
- I removed the use of the extra function to make it more readable
- I still needed to use $row in this new setup because else code would be messy with a lot of array_push and temporary values. I think the current setup will make most sense and is good readable
- Foreach was implemented as it is more readable, but I did not make it more intelligent like handling an unexpected amount of radios or containers. This setup already takes care of different scenarios that could be encountered like the optional modules inserted into the containers. This OS is for a single unit type (SAF CFM L4) and will not run into any issues of other variations. Also the MIB structure has clearly not been built to support different flavours of other models of this device.

includes/definitions/discovery/saf-cfml4.yaml
```diff
@@ -0,0 +1,423 @@
...
+  sensors: 
+    state: 
+      data: 
+        - 
+          descr: 'Active link'
+          index: 'activeLink.{{ $index }}'
+          num_oid: '.1.3.6.1.4.1.7571.100.1.1.2.22.5.1.{{ $index }}'
+          oid: activeLink
+          state_name: activeLink
+          states:
```
laf: Would you mind formatting these states as per other examples:

```
                    states:
                        - { value: 1, descr: other, graph: 1, generic: 3 }
                        - { value: 2, descr: unknown, graph: 1, generic: 3 }
                        - { value: 3, descr: ok, graph: 1, generic: 0 }
                        - { value: 4, descr: non-critical, graph: 1, generic: 1 }
                        - { value: 5, descr: critical, graph: 1, generic: 2 }
                        - { value: 6, descr: non-recoverable, graph: 1, generic: 2 }
```

includes/definitions/saf-cfml4.yaml
```diff
@@ -0,0 +1,12 @@
+os: saf-cfml4
```
laf: Any reason this is a new OS rather than adding to the `saf` OS we have?

includes/definitions/discovery/saf-cfml4.yaml
Schouwenburg: I changed the format of this yaml files as requested. It indeed looks better this way.

Schouwenburg:
@laf no more user-pending

includes/definitions/saf-cfml4.yaml
laf: Not having the same MIB across product(s) is fine. Do they not have other models within this range?

The short is, we don't want to be too specific with the OS names as it leads to duplicates and more work to keep thins up to date.

Schouwenburg: Hello heroes, Yes I understanding the point. The OS I made supports a product line of 3 different models with 3 types of optional modules and 4 types of radios. That covers the whole product line of the L4. There will not come any other models as this product line has been surpassed by newer product lines. Best, Janno

includes/discovery/entity-physical/entity-physical.inc.php
```diff
@@ -25,6 +26,57 @@
     echo ' hwEntityBomEnDesc';
     $entity_array = snmpwalk_cache_oid($device, 'hwEntityBomEnDesc', $entity_array, 'ENTITY-MIB:HUAWEI-ENTITY-EXTENT-MIB');
 }
+if ($device['os'] == 'saf-cfml4') {
```
murrant: This is how I would format this section:

```php
if ($device['os'] == 'saf-cfml4') {
    $entity_array = [];
    echo ' saf-cfml4Anatomy';
    $oid = '.1.3.6.1.4.1.7571.100.1.1.2.22';
    $device_array = snmpwalk_cache_oid($device, $oid, [], 'SAF-MPMUX-MIB')[0];

    $entity_array[] = [
        'entPhysicalVendorType' => 'CFM L4',
        'entPhysicalContainedIn' => '0',
        'entPhysicalClass' => 'chassis',
        'entPhysicalParentRelPos' => '-1',
        'entPhysicalName' => 'Chassis',
        'entPhysicalSerialNum' => $device_array['serialNumber'],
        'entPhysicalMfgName' => 'SAF',
        'entPhysicalModelName' => 'CFM L4',
        'entPhysicalIsFRU' => 'true',
    ];

    foreach ([1 => 'rf1Version', 2 => 'rf2Version'] as $index => $item) {
        $entity_array[] = [
            'entPhysicalDescr' => $device_array[$item],
            'entPhysicalVendorType' => 'radio',
            'entPhysicalContainedIn' => 1,
            'entPhysicalClass' => 'module',
            'entPhysicalParentRelPos' => $index,
            'entPhysicalName' => "Radio $index",
            'entPhysicalIsFRU' => 'true',
        ];
    }

    foreach (range(1, 4) as $index) {
        $entity_array[] = [
            'entPhysicalDescr' => 'Module Container',
            'entPhysicalVendorType' => 'containerSlot',
            'entPhysicalContainedIn' => 1,
            'entPhysicalClass' => 'container',
            'entPhysicalParentRelPos' => $index + 2,
            'entPhysicalName' => "Slot $index",
            'entPhysicalIsFRU' => 'false',
        ];
    }

    foreach ([1 => 'm1Description', 2 => 'm2Description', 3 => 'm3Description', 4 => 'm4Description'] as $index => $item) {
        if (!str_contains('N/A', $device_array[$item])) {
            $entity_array[] = [
                'entPhysicalDescr' => $device_array[$item],
                'entPhysicalVendorType' => 'module',
                'entPhysicalContainedIn' => $index + 3,
                'entPhysicalClass' => 'module',
                'entPhysicalParentRelPos' => 1,
                'entPhysicalName' => "Module $index",
                'entPhysicalIsFRU' => 'true',
            ];
        }
    }
}
```

includes/definitions/saf-cfml4.yaml
murrant: I think laf's point is it should probably just be saf-cfm right?

SAF's products are CFIP, CFQ, CFM

includes/discovery/entity-physical/entity-physical.inc.php
Schouwenburg: Thanks for this suggestion, I did not know how to do "snmpwalk_cache_oid($device, $oid, [], 'SAF-MPMUX-MIB')**[0]**". I used your code and corrected it a bit and added an extra condition as not all cfm devices have modular cards.

LibreNMS/Device/Sensor.php
```diff
@@ -358,9 +358,12 @@ private static function fetchSnmpData($device, $sensors)
             $snmp_data = array_merge($snmp_data, $multi_data);
         }
 
-        // deal with string values that may be surrounded by quotes
+        // deal with string values that may be surrounded by quotes, scientific number format and remove non numerical characters
         array_walk($snmp_data, function (&$oid) {
-            $oid = trim($oid, '"') + 0;
+            $oid = preg_replace('/[^-0-9.e]/', '', $oid);
```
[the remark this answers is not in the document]
Schouwenburg: Hello Murrant. If this is the wrong place than please help me find the right place. I think this might be a great place to generically solve the issues as described below. Furthermore the two times preg_replace are really needed:
The first one removes all characters that should not be in a number or scientific number. So it filters thinks like additional things like V for voltage, A for amperage, C for celcius, and also things like "TX=7123.123Mhz' or 'dB; are cleaned. But the 'e' stays as it might be a scientific number. Then after trying to convert the scientific number to a numeric value it might still have been no scientific number so then we need to clean the value in the second preg_replace.

includes/definitions/saf-cfml4.yaml
Schouwenburg: Yes, I looked into this and you are right. So I will update all with a renamed os.

Schouwenburg: I know changed the OS name from saf-cfml4 to saf-cfm.

LibreNMS/OS/SafCfml4.php
```diff
@@ -0,0 +1,177 @@
...
+        return array(
+            // SAF-MPMUX-MIB::cfml4radioTxFrequency
+            new WirelessSensor(
+                'frequency',
+                $this->getDeviceId(),
+                '.1.3.6.1.4.1.7571.100.1.1.2.22.3.6.0',
+                'saf-cfml4-tx',
+                'cfml4radioR1TxFrequency',
+                'Radio 1 Tx Frequency',
+                null,
```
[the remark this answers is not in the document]
Schouwenburg: Resolved

LibreNMS/Device/Sensor.php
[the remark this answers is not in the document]
murrant: Hmm, this parsed `2.1E-5` incorrectly for example.

murrant:
@Schouwenburg you didn't like my changes?  They seemed to work in all cases I tested.

Schouwenburg:
Hello Tony, I tested it too and scientific numbers where not converted to normal decimal values. In the saf cfm os support as it is today I do not use any scientific numbers anymore so for now this is not important although I suspect this will become an issue for someone in the future. Your code worked great for the removal of extra letters or text around values and that is most important to me. Your code looks clean, that is also great. Best, Janno

murrant:
@Schouwenburg The goal is not to convert them to decimal numbers, it is to convert them to numeric.
They will be stored in the DB then displayed in the UI as decimal.

murrant:
This should be a more accurate regex:
```php
            preg_match('/-?\d+(.\d+)?(e-?\d+)?/', $oid, $matches);
            if (isset($matches[0])) {
                $oid = $matches[0] + 0;
            } else {
                $oid = trim('"', $oid); // allow string only values
            }
```

PR #9457, opened 2018-11-17
Added ability to record traceroutes for devices down due to ICMP

laf:
This can be used in alerting ($alert->debug['traceroute']) or viewed in the webui:

image

We could have the ability to turn on per device the ability to record this all the time.

At present it does it every time the device is down if the config option is on, we could do it when the status changes to down and no further.

Will add a small increase in poll time for the down device but that's probably not likely to be longer than a full poll for a device - but it could.

### DIFF
--- a/doc/Alerting/Templates.md
+++ b/doc/Alerting/Templates.md
@@ -50,7 +50,7 @@ Placeholders are special variables that if used within the template will be repl
 - ping min (if icmp enabled): `$alert->ping_min`
 - ping max (if icmp enabled): `$alert->ping_max`
 - ping avg (if icmp enabled): `$alert->ping_avg`
-- debug (array) This will contain:
+- debug (array) If `$config['debug']['run_trace] = true;` is set then this will contain:
   - traceroute (if enabled you will receive traceroute output): `$alert->debug['traceroute']`
   - output (if the traceroute fails this will contain why): `$alert->debug['output']`
 - Title for the Alert: `$alert->title`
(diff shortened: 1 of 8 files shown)


doc/Alerting/Templates.md
```diff
@@ -50,6 +50,9 @@ Placeholders are special variables that if used within the template will be repl
 - ping min (if icmp enabled): `$alert->ping_min`
 - ping max (if icmp enabled): `$alert->ping_max`
 - ping avg (if icmp enabled): `$alert->ping_avg`
+- debug (array) This will contain:
```
murrant: Mention the config setting here too I think.

doc/Support/Configuration.md
```diff
@@ -100,6 +100,15 @@ $config['icmp_check'] = false;
 
 If you would like to do this on a per device basis then you can do so under Device -> Edit -> Misc -> Disable ICMP Test? On
 
+#### traceroute
+LibreNMS uses traceroute / traceroute6 to record debug information when a device is down due to icmp AND you have `$config['debug']['run_trace] = true;`
```
swilsonskytap: Missing single quote after run_trace.

murrant: Can you send a pull request to fix?

swilsonskytap: Sure.  I'm new to this, but I should be able to figure it out and get that going.

murrant: Easy way go to the docs page and click the edit button ;)

swilsonskytap: Thanks.  I'll get this figured out eventually :)   Anyway, looks like someone else corrected it.

chasgames:
this is epic 👏🏻

MrBlueSkye:
Great feature :100:

murrant:

https://community.librenms.org/t/add-new-tool/7696/4

Issue #8230, opened 2018-02-08
Add support for AKCP SPX+ monitoring devices


olivierbeytrison:
- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Component | Version
--------- | -------
LibreNMS  | 1.36-36-g5474519
DB Schema | 234
PHP       | 7.0.22-0ubuntu0.16.04.1
MySQL     | 10.0.33-MariaDB-0ubuntu0.16.04.1
RRDTool   | 1.5.5
SNMP      | NET-SNMP 5.7.3

[Discovery](https://p.libren.ms/view/22db0ce3)
[Poller](https://p.libren.ms/view/bb882fc7)
[SNMP Walk (didn't finish because of errors)](https://p.libren.ms/view/fc4c0b6c)
[SNMP Walk of the enterprise OID](https://p.libren.ms/view/47a8aeaf)

The new monitoring device [SPX+ by AKCP](http://www.akcp.com/products/sensorprobe-plus/spx-plus/) doesn't use the previous MIB (like for sensor probes or security probes). It uses now a new MIB with a dynamic table listing the different sensors.

Would it be possible to add support for this sensor ?

boxcomrob:
The device we are using is the SP2+, it's in the same range as the SPX+ mentioned above but is a smaller less-comprehensive device.

I am new to SNMP so hopefully this is the correct information. I do have a MIB from AKCP covering this whole range (a copyright notice mentions it is not allowed to be disclosed though, but I will request permission to post it here if it helps) but I'm not really sure how to use it, it shows hundreds of validation errors and won't import into any monitoring software I have tried so far. 

With our SP2+ device, we are using a combined temperature/humidity sensor on port 1, a smoke detector (configured as a dry contact) on port 2, and a dry contact door switch on port 3.

====================================
Component | Version
--------- | -------
LibreNMS  | 1.37-60-g9001b19
DB Schema | 245
PHP       | 7.0.25-0ubuntu0.16.04.1
[...]

PR #9460, opened 2018-11-19
Updated detection for AKCP devices

laf:
Fixes: #8230


murrant:
:/  akcp seems to be a duplicate of ethernetprobe2  ( or at least a rebranded version)

Other than that, this looks fine.

PR #9463, opened 2018-11-19
Syslog name translation

TylerSweet:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This helps to solve an issue where syslog data is not being associated with the proper device because the following conditions are true:
1) The device is forwarding syslog through another host 
2) The device's resolved DNS name does not match the SysName

In our case, we poll devices on one IP address but Syslogs are forwarded from a second interface to another host, which collects them before forwarding them onto LibreNMS. 
Because the sending IP address doesn't match, as well as the hostname it sends from doesn't match any hostname that LibreNMS has, the logs are never associated with a device. This solves that problem, by allowing users to provide a translation table in config.php. For example:

`$config['syslog_xlate'] = array(
        'loopback0.core7k1.noc' => 'n7k1-core7k1',
        'loopback0.core7k2.noc' => 'n7k2-core7k2'
);`

### DIFF
--- a/includes/syslog.php
+++ b/includes/syslog.php
@@ -5,7 +5,6 @@
 // $device_id_ip = @dbFetchCell("SELECT device_id FROM ipv4_addresses AS A, ports AS I WHERE A.ipv4_address = '" . $entry['host']."' AND I.port_id = A.port_id");
 use LibreNMS\Config;
 
-
 function get_cache($host, $value)
 {
     global $dev_cache;
--- a/includes/syslog.php
+++ b/includes/syslog.php
@@ -3,6 +3,7 @@
 // FIXME : use db functions properly
 // $device_id_host = @dbFetchCell("SELECT device_id FROM devices WHERE `hostname` = '".mres($entry['host'])."' OR `sysName` = '".mres($entry['host'])."'");
 // $device_id_ip = @dbFetchCell("SELECT device_id FROM ipv4_addresses AS A, ports AS I WHERE A.ipv4_address = '" . $entry['host']."' AND I.port_id = A.port_id");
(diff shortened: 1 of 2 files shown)


includes/syslog.php
```diff
@@ -57,6 +57,9 @@ function process_syslog($entry, $update)
     }
 
     $entry['host'] = preg_replace("/^::ffff:/", "", $entry['host']);
+    if (array_key_exists($entry['host'], $config['syslog_xlate'])) {
```
murrant: Please use \LibreNMS\Config instead of accessing $config directly.

murrant:
You will also need to add docs.

includes/syslog.php
```diff
@@ -57,8 +59,8 @@ function process_syslog($entry, $update)
     }
 
     $entry['host'] = preg_replace("/^::ffff:/", "", $entry['host']);
-    if (array_key_exists($entry['host'], $config['syslog_xlate'])) {
-        $entry['host'] = $config['syslog_xlate'][$entry['host']];
+    if (array_key_exists($entry['host'], $syslog_xlate)) {
```
murrant: Close, try this:
```php
if (Config::has('syslog_xlate.' . $entry['host'])) {
    $entry['host'] = Config::get('syslog_xlate.' . $entry['host']);
}
```
or just

```php
if ($new_host = Config::get('syslog_xlate.' . $entry['host'])) {
    $entry['host'] = $new_host;
}
```

TylerSweet: Oh, those look much cleaner. I like the second one. I'll be submitting the changes soon.

PR #9471, opened 2018-11-20
New OS: Schleifenbauer, entity-physical improvements, various html page fixes

### LINKED ISSUE #9360: New device: Schleifenbauer PDU
**Please do not post bugs or ask questions**

**For help and support you can use our [Discord server](https://t.libren.ms/discord) or our [community site](https://community.librenms.org).**

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Please provide **ALL** info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

Adding this to LibreNMS identifies it as a server. The type should be Power.
sysName on multiple devices is always returning schleifenbauer.  
| Schleifenbauer | .1.3.6.1.4.1.31034.11.1 | 

SNMP Example from Vendor: http://documentation.schleifenbauer.eu/SNMP/snmp%20examples.txt 
Vendor MIB: http://documentation.schleifenbauer.eu/SNMP/PDU_mib_244.zip 

[generic-10.70.2.42.txt](https://github.com/librenms/librenms/files/2500775/generic-10.70.2.42.txt)
[poller-10.70.2.42.txt](https://github.com/librenms/librenms/files/2500777/poller-10.70.2.42.txt)
[discovery-10.70.2.42.txt](https://github.com/librenms/librenms/files/2500774/discovery-10.70.2.42.txt)

martijn-schmidt:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

#### Questions
- Is there is a cleaner way to populate all states >1 for the sdbMgmtStsDuplicateDevices and sdbMgmtStsNewDevices sensors in includes/definitions/discovery/schleifenbauer.yaml?
- In a large databus ring, you may see timeouts during the discovery process regardless of snmpbulkwalk|snmpwalk, this can mostly be mitigated by setting a global or device-specific timeout 10 and retries 3. However, I could not find if this is possible to set that for this specific OS so that there is a sane default, rather than relying on user configuration changes for a properly functioning discovery?

#### Changes

- Fix various little typos in comments, documentation, or html output.
- Add a few nicecase() calls in html pages where appropriate.
- Add power_consumed and power_factor sensors.

- Fix and add various things on the device entphysical tab:
-- Resolve flickering during loading (missing liClosed for nested items).
-- Clicking a sensor actually takes you to the matching graph now, instead of the health tab.
-- Add two new icons for "stack" and "powerSupply"
-- Make stored entPhysicalAlias and entPhysicalAssetID data visible in the inventory.

- Create discover_entity_physical() function:
-- Add discover_entity_physical() function into includes/discovery/functions.inc.php
-- This allows for an easy implementation of OS-specific entity-physical discovery.
-- Update includes/discovery/entity-physical/entity-physical.inc.php to use the new function.

- Add schleifenbauer OS support, closes #9360:
-- We detect the master unit in includes/polling/os/schleifenbauer.inc.php by looking at object sdbMgmtCtrlDevUnitAddress to find the unit at position 1 of the ring since the device itself would otherwise not return any data about firmware, hardware, serial, etc.
-- Opted to go for PHP based instead of YAML based discovery for most sensors due to several dual-index objects in the MIB and lack of RFC-based ENTITY-MIB support.
-- Opted to use the serial-number and phase-number of the device as the discover_sensor($index) value since the databus address may change frequently. This keeps the graphs linked to the correct PDU regardless of its unit address, assuming you run your rediscovery at the right time.
-- Populates entity-physical databases with data collected through the proprietary MIB, builds its own entityPhysicalIndex values based on sdbMgmtCtrlDevUnitAddress, methodology documented in the comments. I didn't want to rely on sdbDevIdIndex since that had a much larger range and could return 0 (see "other remarks" below), resulting in trouble with the entPhysicalIndex calculation.

Other remarks:
- It is currently not possible to set ring-wide MIB-2 management values SysName, SysLocation, SysContact for this OS. I will ask the vendor to support that properly together with @Knot3n.
- sdbDevCfEntry contains interesting data about external sensor modules, but I don't have access to one of those - so I'll be asking the vendor for demo hardware to implement that beyond populating the entity-physical database.
-- In a large databus ring, you may see timeouts during the discovery process regardless of snmpbulkwalk|snmpwalk, this can mostly be mitigated by setting a global or device-specific timeout 10 and retries 3.
- If you have a largeish databus ring, and the ring is opened, this significantly slows down the polling cycles to the point where they can last >5 minutes. Discovery, of course, also slows down to a crawl at that point so you will see that sensors affected by SNMP timeouts are (temporarily) deleted. It's strongly advised for end users to set up a high priority alert for the "Databus Ring State" sensor.
- Unit addresses may change due to user modifications in the databus. If you run a device rediscovery immediately after making your changes in Schleifenbauer's webinterface, the code picks up on the new unit addresses and keeps your original graph data intact.
- The MIB specifies that object sdbDevIdIndex must be between 1 and 65535, but it is possible to have a device with unit address 0 and that results in an out of range sdbDevIdIndex value. I think it may be "not a bug, but a feature" and will contact the vendor about that with @Knot3n.


murrant:
Answers: 
sdbMgmtStsDuplicateDevices isn't a state sensor.  State sensors have finite pre-defined states like notPresent, ok, warning, critical.
try tweaking the repeaters option for the devices and report sub-optimal performance to the manufacturer.  Or is the problem limited bandwidth on the "databus ring"?

I'm not really a fan of nicecase() and try to avoid it if possible :/

murrant:
Thanks for the PR!

Answers: 
sdbMgmtStsDuplicateDevices isn't a state sensor.  State sensors have finite pre-defined states like notPresent, ok, warning, critical.
try tweaking the repeaters option for the devices and report sub-optimal performance to the manufacturer.  Or is the problem limited bandwidth on the "databus ring"?

Is power_consumed the best name for that sensor?  What does it measure?
I'm not really a fan of nicecase() and try to avoid it if possible :/

martijn-schmidt:
Hey @murrant - thanks for the feedback.

What do you think I should do with the sdbMgmtStsDuplicateDevices sensor and its friends then? None of the other predefined sensor classes really seem to fit the bill. I figured it was closest to a state sensor, since a value of 0 means one thing (e.g. things are OK), and a value of >1 means another thing (e.g. things are Critical) but with a bit more information about how severe the Critical problem actually is. Maybe the following would help - do you have a preference for one of those two?

- Modify the code for the state sensor class to allow "ranges" or ">n" and "<n" style operators for the values in the state table, where the detected value is then used as a variable in the descr defined in the state table.
- Alternatively, create a generic "count" or "amount" sensor class for sensors such as sdbMgmtStsDuplicateDevices, sdbMgmtStsNewDevices, sdbMgmtStsDevices, sdbMgmtStsAddressableDevices, etc.

Yes, there is limited bandwidth on the databus ring.. the vendor's own web UI is also quite slow in larger rings so it's not just something innate to the snmp implementation. I believe my colleagues were previously told by the vendor that the only way to really speed it up is to carve your databus rings into smaller segments or even forego the databus ring architecture altogether, giving each device its own ethernet uplink. But in a colocation environment where you don't have a switch in every cabinet with a PDU installed, or where the customer might even manage their own PDU via ethernet in terms of switchable outlets while the PDU is still part of the databus ring for measurement purposes by the datacenter operator, that is not always viable.

The power_consumed sensor measures the accumulated lifetime power consumption of that PDU, expressed in kWh (Kilowatt Hours). If you have a power draw of 2 kW, and you keep it at that level for 24 hours, you will have consumed 48 kWh worth of power. "power_consumed" is an up and to the right style graph which relates itself to the "power" graph much like a "GB of data consumed" graph would relate to the "Mbps of traffic" measurements made for a network device. The advantage here is that it's more accurate than multiplying "power" and "time", and that it keeps count on the device itself regardless of missed polling cycles.

As for nicecase(), I added that way since I figured it was the closest to what the LibreNMS codebase already did to clean up certain sensor names before displaying them, like is the case for dbm -> dBm. But as long as we can get rid of the underscore in quality_factor or chromatic_dispersion before displaying them in html, I'd be happy.

martijn-schmidt:
Such a finite state OID is available as `sdbMgmtStsRingState` (1 = ok / 0 = critical) for `sdbMgmtStsRingBreachIndex`. I have implemented it using the former, with ok/critical style output, but the latter also indicates the unit ID of the PDU where the ring has been breached.

Unfortunately there is no finite state OID for `sdbMgmtStsDuplicateDevices` or `sdbMgmtStsNewDevices` at this time, so we could remove those two sensors from the PR code for the time being. Then I can work with @Knot3n to ask the vendor for a new OID with ok/warn/critical style indications so they can be added back to the LibreNMS codebase "cleanly" when a Schleifenbauer firmware update becomes available?

About `power_usage` instead of `power_consumed`: I'd considered that name actually, but felt it'd be too similar to the existing `power` sensor in terms of meaning.. potentially raising doubts about what it is. A closer match, if `power_consumed` is not desirable, could be the `energy` physical quantity?

The official SI unit for `energy` is Joule, rather than the non-SI unit kWh which is most widely used in the datacenter industry. But it's easy to convert kWh to Joule and vice versa, so if there's a strong preference to use SI units in LibreNMS I'm sure I could build a `kwh_to_joule` function and discover/poll the sensors with `user_func` to convert the data into SI units, much like the [Health Information documentation](https://docs.librenms.org/Developing/os/Health-Information/) refers to using `fahrenheit_to_celsius` as a `user_func`.

Quoting from https://www.quora.com/How-do-you-convert-joules-to-kWh for reference:

>A watt is energy consumed at the rate of one joule per second.
>
>There are 3600 seconds in an hour. There are 1000W in a kW (the k means times 1000).
>
>So:
>kWh x 3600s/h x 1000 = J
>A little algebra to reverse:
>1 kWh = J/3,600,000
>
>So divide the number of joules by 3.6x10^6 to get the number of kWh.

The downside of doing it that way is that the sensor measures with 1 kWh granularity - and each kWh will be converted to 3600000 Joule. With the vendor MIB allowing measurements for kWh between 0 and 16777215, we could potentially get graphs up to 60397974000000 Joule - and I have seen devices with lifetime power consumption measurements of 255186 kWh in real-world production.

martijn-schmidt:
No problem, I appreciate the back-and-forth responses since it'll get us a better result in the end :)

Firstly, the number sensor: do we want name it `count` with unit `#`, or `amount` with unit `#`?

As for measuring stored energy: I think you specifically mean electric charge, like in a UPS, or like in a smartphone which could for example have 3450 mAh battery capacity? I believe that would use its own SI unit Coulomb (electric charge) rather than Joule (energy), and it appears to be something different as per [Quora](https://www.quora.com/What-is-the-difference-between-mAh-and-kWh) - I'm falling back to that site for some information since I'm not an electrician myself:

> A milliampere-hour (mAh) is a unit of electric charge (Current×Time) whereas kilowatt-hour (kWh) is a unit of electric energy (Voltage×Current×Time). You can convert mAh into kWh by simply multiplying with a voltage.

So just like we also have separate graphs and OID measurements for `power`, even though it could be derived from `voltage * current`, I feel it'd also be worthwhile to have separate graphs and OID measurements for `electric energy` versus `electric charge`. 

For the SI versus non-SI unit conversation I also did some more digging, and I personally feel this [reddit post](https://www.reddit.com/r/explainlikeimfive/comments/4h8m57/eli5_why_are_batteries_charge_measured_in_mah/) sums it up nicely:

> mAh or Ah is a much more convenient unit to use, though. Coulomb would be correct by SI standards, but measuring device lifetime in seconds isn't very practical. The expected battery lifetime of the average electronic device is in measured in hours, so introducing lots of conversion factors would make it difficult to mentally size a battery.

Then to get back to the subject at hand: my personal preference would go to either `electric_energy` or `power_consumed` both of which would be expressed in the unit kWh, and `electric_charge` with unit mAh for energy stored in a battery, since these two units appears to be the standard in at least the datacenter business.

murrant:
Yes, I accidentally conflated Amp hour and Watt hour. (writing them out like that makes it pretty obvious)

power_consumed seems fine in that case.

For count `#` is my preference.  However, the units should probably be configured by the sensor, but that would take some code updates and probably a new database field.

martijn-schmidt:
@murrant I have added the count sensor and modified the Schleifenbauer discovery yaml to use that instead of the state sensor where applicable. I think this should cover it?

murrant:
@librenms/reviewers thoughts on the count sensor?

laf:
> @librenms/reviewers thoughts on the count sensor?

Seems fine to me. The only issue will be the stacked graphs for it could be useless if various count graphs are added at some stage as they can depict various things the scales may not work.

Aside from that, this PR needs the logos to be compressed (https://jakearchibald.github.io/svgomg/)

Mibs have been uploaded but appear not to be used, could you switch to using the mib please.

Any polling needs to be optimised. Please use snmp multi get rather than multiple single gets.

murrant:
Not sure if it is the correct way to go, but how about adding a nullable field on the sensors table called group?  We could use it for grouping sensors together in the ui.

laf:
> Not sure if it is the correct way to go, but how about adding a nullable field on the sensors table called group? We could use it for grouping sensors together in the ui.

Sounds good. I suggested the same thing in the custom oid PR.

murrant:
I was thinking about this and sensor_type is basically unused (it is just part of the sensor_index really).  Should I try to re-purpose that, remove that and add a group field, or just add a group field?

murrant:

martijn-schmidt:
Thanks for the help to get the last bits and pieces done, murrant!

murrant:
Thanks for the PR :)

PR #9476, opened 2018-11-25
Fix do not include template text in HTML page

lowinger42:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


lowinger42:
No, 
there is no reason to include the text from all templates to the web page, just to allow javascript to check if the text {/if} is inside one of the templates and show the upgrade warning?  

The template text is not used for anything else except this check. It was added in #9413.

In a slow computer the page looks very funny, with all the html templates in the last column visible/flashes, until the column is hidden.

laf:
Awesome, thanks for fixing.

Issue #9485, opened 2018-11-29
Unknown column 'location' after merge of PR #9359 at ajax_search.php


empi89:
After the latest daily.sh run with autoupdate enabled I was unable to search for device names with the global search in the upper right. I have investigated on it and found that the ajax_search.php returned:
```
SQL Error! SQLSTATE[42S22]: Column not found: 1054 Unknown column 'location' in 'where clause' (SQL: SELECT * FROM `devices` WHERE `hostname` LIKE '%core%' OR `location` LIKE '%core%' OR `sysName` LIKE '%core%' OR `purpose` LIKE '%core%' OR `notes` LIKE '%core%' ORDER BY hostname LIMIT 12) (SQL: SELECT * FROM `devices` WHERE `hostname` LIKE '%core%' OR `location` LIKE '%core%' OR `sysName` LIKE '%core%' OR `purpose` LIKE '%core%' OR `notes` LIKE '%core%' ORDER BY hostname LIMIT 12)
  /var/opt/librenms/html/ajax_search.php:49
[]
```

Which seems to be related to https://github.com/librenms/librenms/commit/3e35ee0e7d1ad20289e5adecf35a52e7d3e06177 of @laf and @murrant.

PR #9486, opened 2018-11-29
Fix for #9485

empi89:
This is the fix for https://github.com/librenms/librenms/issues/9485.


laf:
Thanks for fixing this. FYI you don't need to create an issue to do a PR. We don't use github issues for bugs / issues.

PR #9488, opened 2018-11-29
Added support for Vigintos Modulator & Ampiflier

jozefrebjak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@jozefrebjak Looks like you could have used yaml for this any reason not to?

mibs/viginitos/VEL-MIB
```diff
@@ -0,0 +1,41 @@
+VEL-MIB DEFINITIONS ::= BEGIN
```
murrant: The MIB directory should match the os name. (you have an extra i here)

jozefrebjak: it was mistake, I changed it

jozefrebjak: thank you :)

jozefrebjak:
There is how it is looks like in a web, its something like host controller for DVB modulator and amplifier

snimka obrazovky 2018-12-21 o 22 33 00
snimka obrazovky 2018-12-21 o 22 32 51

murrant:
I'm saying according to the MIB it doesn't guarantee they will always be at those OIDs (on different devices).  It is just a list of values.

It might be more reliable to use host2DevParUnit to determine what type of sensor it is.

murrant:
@jozefrebjak how are you coming along with this?

jozefrebjak:
@murrant  so, i deleted all php discovery and replaced them with yaml. I used your advice to do this.

PR #9493, opened 2018-11-29
Fixed ifType being removed when a port is down

PipoCanaja:
DO NOT DELETE THIS TEXT

#### Please note

Hello,

This PR is adding ifType to table_base_oids.  Without it, a port down looses its type, making it difficult to filter on this criteria. It seems that ifType is not a costly OID so it should not be to bad to the performance. 
I run this change on my setup for a while, but this is of course not significant considering the large amount of devices around. At least on a few Cisco switches and mostly Huawei devices, it did not make a change to add it. 

Waiting for feedbacks ;) 

Bye
PipoCanaja

#### Testers


murrant:
Perhaps it would be better to prevent ifType from being deleted when a port is down (I'm assuming selective port polling)?

PipoCanaja:
Assuming right for selective port polling. 
The idea would be to get ifType from discovery process, and to avoid deleting it during polling ? I'll have a look at this approach, which makes sense.
Bye for now

PipoCanaja:
Quick update, seems that the discovery process does not fill ifType, or at least does not save it. So it is not as easy as I first hoped ;) 
So I suppose we should first add ifType into the discovery process (which means it will be at least updated every occurence of discovery) and keep the value in between.

This could even be chosen as the default behaviour for this OID cause ifType is a very stable value I would say, stable enough to leave poller and be only in discovery ... What do you think ?

laf:
> I suppose we should first add ifType into the discovery process (which means it will be at least updated every occurence of discovery) and keep the value in between (I have a patch for this already).

This makes more sense than increasing polling times.

PipoCanaja:
Polling time unchanged in current state of the PR. Instead, the data already collected during the discovery process is saved or updated, and the polling keeps the values unchanged if selective polling is enabled.

murrant:
Tests can be fixed my running `./scripts/save-test-data.php -m ports`, this will cause a lot of modified files.  If you want I can run it and push it to your branch...

PipoCanaja:
Hi @murrant 
I would agree to your proposition to run the tests. I am not comfortable doing it. 
PipoCanaja

laf:
I've pushed the tests up.

PR #9495, opened 2018-11-29
Dynamic_discovery_get_value in can_skip_sensor to use all oids in skip_values

PipoCanaja:
Hello

While working on dynamic discovery of Huawei switch, I faced an issue where I needed to filter sensors with skip_values but the oid to use was not in the current table. Instead, I went with pre_cache, which works well for displaying variables with {{ xxxxx }}. 
But no success. 
While searching the reason, it appeared that the magic is done by dynamic_discovery_get_value for the variable replacement, but not in can_skip_sensor. 
This patches slightly changes the signature of can_skip_sensor function, and now uses dynamic_discovery_get_values fed with the complete $pre_cache. 

Works well in my lab and prod now. 

New YAML files using this feature will come in dedicated PRs for Huawei VRP switches. 

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Device/YamlDiscovery.php
+++ b/LibreNMS/Device/YamlDiscovery.php
@@ -90,7 +90,7 @@ class YamlDiscovery
                         }
                     }
 
-                    if (static::canSkipItem($current_data['value'], $current_data, $group_options, $snmp_data)) {
+                    if (static::canSkipItem($current_data['value'], $index, $current_data, $group_options, $snmp_data)) {
                         continue;
                     }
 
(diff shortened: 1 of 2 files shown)


murrant:
@PipoCanaja interesting, I forgot there were two implementations of this.

Perhaps it should be unified like so:

```php
function can_skip_sensor($value, $yaml_item_data, $group_options, $item_snmp_data = array())
{
    return YamlDiscovery::canSkipItem($value, $yaml_item_data, $group_options, $item_snmp_data);
}
``

PipoCanaja:
I didn't know yet that a 2nd implémentation was there ... I would go with removal as well. I'll try it and adapt the PR. 

PipoCanaja

PipoCanaja:
Hi again,
I had to remove the fact that YamlDiscovery::canSkipItem was private.

PR #9496, opened 2018-11-30
Fixed integers fields in alert rules to be string

TheGreatDoc:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
I'd really like to see the editor allow field selection by dropdown instead of this work-around.

I understand that isn't necessarily an easy task :)

TheGreatDoc:
I know. I want to do field type by operator dropdown.

But! This is going to take way more time than what I did here and I think we need to fix it quick.

murrant:
Yeah, can you add `//TODO implement field selection and change back to integer` comment on that line?

TheGreatDoc:
Done. But... If field selection by operator is implemented, is this function neede anymore?

murrant:
Yes, that would be done in javascript

PR #9499, opened 2018-11-30
Add Device Dragonwave Harmony Enhanced Mc

djamp42:
Add OS/Wireless Sensors Support for Dragonwave Harmony Enhanced MC Radios

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/OS/HarmonyEnhancedMc.php
```diff
@@ -0,0 +1,90 @@
...
+        $sensors = array();
+        foreach ($oids as $index => $entry) {
+            $sensors[] = new WirelessSensor(
+                'rssi',
+                $this->getDeviceId(),
+                '.1.3.6.1.4.1.7262.4.5.12.203.1.1.5.' . $index,
+                'harmony_enhanced_mc',
+                $index,
+                'RSL Radio ' .$index,
+                'null',
```
murrant: this should just be null, not 'null', same with the next one and the divisor (10 instead of '10')

murrant:
Thanks for the PR!

Your OS name seems to be a specific hardware device no an OS.  A good rule of thumb is looking at how the MIBs are laid out.  Does the MIB cover more than just one device?

djamp42:
I went through some of the other products, and only product that uses the same is "harmony enhanced". Other harmony radios use completely different MIBS. I'll update everything to just "harmony_enhanced" if that sounds good?

murrant:
That sounds good @djamp42 Thanks.

djamp42:
Okay i have updated the OS name as described above, tested against my production server, all SNMP is working and OS is being discovered correctly. Thanks

PR #9362, opened 2018-10-22
WIP - Added audiocodes discovery and basic graphs

PipoCanaja:
Hi,

Here is a basic Audiocodes support. It was mostly tested with MP118, with 8 FXS ports. Anything else is not tested. 

Please try this PR on test servers which have access to different audiocodes devices and contact me in this post so we can exchange snmpbulkwalks of your devices in order to improve. 

Thanx

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [X] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

fixes #8954

PR #9508, opened 2018-12-01
Audiocodes initial support

PipoCanaja:
Second attempt for this Audiocodes basic support. The goal is to have it merged, and more sensors will be added later, as I own only 1 model to do the tests on, and I did not succeed adding devices without properly testing them locally. 

Code here is a cleaned subset of https://github.com/librenms/librenms/pull/9362 

I don't expect to add more to this PR, so if the auto tests are all right, it will be ready for merge. 

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/discovery/audiocodes.yaml
```diff
@@ -0,0 +1,153 @@
...
+                    descr: 'FXS {{ $index }} ({{ $callerDisplayCallerDisplay }}, {{ $channelsStartingPhoneNum }})'
+                    index: 'hookStatus{{ $index }}'
+                    state_name: acAnalogFxsFxoHookState
+                    states:
+                        - { descr: offHook, graph: 0, value: 0, generic: 0 }
+                        - { descr: onHook, graph: 0, value: 1, generic: 0 }
+                    skip_values:
+                        -
+                            oid: acAnalogFxsFxoType
+                            op: '!='
```
murrant: If off-hook the sensor will be deleted during discovery.

includes/definitions/discovery/audiocodes.yaml
```diff
@@ -0,0 +1,153 @@
...
+                    descr: 'E&M {{ $index }} ({{ $callerDisplayCallerDisplay }}, {{ $channelsStartingPhoneNum }})'
+                    index: 'hookStatus{{ $index }}'
+                    state_name: acAnalogFxsFxoHookState
+                    states:
+                        - { descr: offHook, graph: 0, value: 0, generic: 0 }
+                        - { descr: onHook, graph: 0, value: 1, generic: 0 }
+                    skip_values:
+                        -
+                            oid: acAnalogFxsFxoType
+                            op: '!='
```
murrant: should this be `=`?  I assume 2 is not present/configured.

PipoCanaja: The idea is to create 3 times the same OID depending on the value of acAnalogFxsFxoType but change the description statically in each of the 3 instance. Unfortunately, there is no way to get "FXO", "FXS" of "E&M" from a variable ... So the '!=' is correct here.

includes/definitions/discovery/audiocodes.yaml
PipoCanaja: I filter on acAnalogFxsFxoType, not on acAnalogFxsFxoHookState. So the discovery is not impacted at all iwth HookState. 
This bloc is 3 times the same sensor declare (based on acAnalogFxsFxoType) but depending on the device, I create it with FXO, FXS or E&M description.

PR #9511, opened 2018-12-03
Correct sound for pushover.net

sorano:
pushover.net should now use sound= instead of critical.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Alerting/Transports.md
+++ b/doc/Alerting/Transports.md
@@ -319,7 +319,7 @@ Get your Access Token from your Pushbullet's settings page and set it in your tr
 ## Pushover
 If you want to change the [notification sounds](https://pushover.net/api#sounds) then add it in Pushover Options:
 
-`sound_critical=falling`
+`sound=falling`
 
 Enabling Pushover support is fairly easy, there are only two required parameters.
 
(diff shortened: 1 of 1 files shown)


doc/Alerting/Transports.md
```diff
@@ -319,7 +319,7 @@ Get your Access Token from your Pushbullet's settings page and set it in your tr
 ## Pushover
 If you want to change the [notification sounds](https://pushover.net/api#sounds) then add it in Pushover Options:
 
-`sound_critical=falling`
+`sound=falling`
```
murrant: According to the source code,
sound_critical, sound_warning, and sound_ok are set to the sound field in the pushover api...  I don't think just sound will go through.

sorano: Well, I was the one that commited sound_critical=falling to the documentation.
Yesterday I realised that the sound was not the falling tune, I changed to sound=falling and then it started to work again.
🤷‍♂️

murrant: See #9519

I think setting sound will always use the same sound, (instead of different ones for up/down).  You could add that to the docs.

murrant: @sorano I merged the fix so now the original docs work, but also, sound= works to set the sound for all status at once.  Do you want to add that or just leave it as is?

sorano: I will add it to docs in a couple of hours.

PR #9514, opened 2018-12-03
Alert schedule refactor

murrant:
Allow maintenance schedule check in Laravel code.
Fixes to the schedule editor
Change the target selection to select2 like the alert rule editor.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
Initial testing looks ok, need to wait for maintenance window to finish now and see if alerts are ok then test reoccurring + groups.

laf:
Not sure if this is an issue from how it was before but you can't set any date in the future. The default seems to stop at the current date time but if you set a future date then you're still restricted on the hours you can set it for.

murrant:
@laf I ran into that during testing and tried to fix it.  Perhaps I didn't fix all the cases...

murrant:
@laf ok, I think I have all the cases covered now for limiting the datetime picker...

laf:
Now still seems wrong.

If you pick reoccuring and set the start date to tomorrow, you won't be able to pick any hour, it will limit you.

Non-reoccuring ones also don't allow you to set a future date and time past the current time - date itself works just not the time.

murrant:
@librenms/reviewers anything holding this up?  I'd like to consider merging the migrations PR this month and having this merged would make it easier.

laf:
Re-occuring schedules still don't allow you to set a time later than the current time even if you select a later date.

murrant:
@laf I can't reproduce that.  In recurring start hour cannot be after end hour and end hour cannot be before start hour.  Recurring must hour/min must be contained in a single day.

laf:
> @laf I can't reproduce that. In recurring start hour cannot be after end hour and end hour cannot be before start hour. Recurring must hour/min must be contained in a single day.
> 
> Something like this might be more appropriate.
> [codepen.io/caseymhunt/pen/kertA](https://codepen.io/caseymhunt/pen/kertA)

Makes sense.

Probably not worth merging this late hour for 1.47.

PR #9516, opened 2018-12-03
Added Wireless discovery to Huawei Vrp

PipoCanaja:
A large set of Huawei VRP devices (switches) can become Wireless Lan Controllers, as well as the AC (Access Controller) family. 
This PR detects any answer in the WLAN mibs of Huawei and gather some data. 

Please let me know if I am going the right direction. I tried to use the LibreNMS/OS/xxxx with classes and co, and was partly successfull. WirelessSensors are discovered that way. 
But doing so for accesspoints polling would result in multiple time polling the same OIDs without caching capabilities (I use snmpwalk_group which does not cache) for every class, which would be very expensive in time.

So I went back to the ciscowlc route where polling is done in includes/polling/os/xxxx.inc.php. If a cleaner way is possible without rewriting all the access_points part of the code, I am interested ;) 

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
FYI, the access points code is due for a rewrite to bring it in line with the Wireless code.  But considering I don't have a controller like this, I did not have working knowledge and lost interest and never completed it.

PipoCanaja:
@murrant 
It makes sense. I'll go with current philosophy first to get something working, and will improve it as it goes. 
My questions for the moment are : 
- The snmpwalk_group() function is great whenever you get data indexed by mac-address or other funny types but it does not seems to be cached. Any other way to get the same kind of walk with caching enabled ? 
- So everytime you call it, it costs time in the discovery (not too bad) or polling (bad 🤒 ). In order to avoid this, I try to do it only once, which is not compatible with the multiple "interfaces" and methods of the Wireless code. A way to get around this could be to have some "pre_cache" method called for the device, where the snmpwalk_group is called, and that creates data freely accessible from all the interface methods later.

murrant:
@PipoCanaja you can cache data on the OS, it persists.  Check the getCacheByIndex() function to see.  Feel free to implement that on the specific OS class, and we can pull it up if it is useful to others.

Basically the goal is to get rid of pre_cache as it is unconditional, so even if the data isn't needed, it is fetched.  getCacheByIndex() the first one that calls it fetches the data and caches it, next one(s) get the cached.

PipoCanaja:
I'll try to see how it goes with the snmpwalk_group part, because the fact that OIDs are converted into MAC Addresses and back, makes it more difficult to cache. This job is done directly by snmpwalk (with the proper option).

PipoCanaja:
Hello @murrant,

The function "public function getCacheTable($oid, $mib = null)" seems the be very close to my needs. I only miss the capacity to give it the depth parameter to influence the indexing, parameter that I can give to snmpwalk_group. Exemple :

 $vapInfoTable = snmpwalk_group($this->getDevice(), 'hwWlanVapInfoTable', 'HUAWEI-WLAN-VAP-MIB', 3, array());

The "3" at the end allow snmpwalk_group to properly index a table with multiple indexes (depth). 

And upgrading "getCacheTable" to pass the option is also not so easy, cause we should also add a way in the cache hash to store the value of the option. If we don't, then the same OID being accessed twice in the code, but with different depth values in snmpwalk_group parameters, would certainely break.

murrant:
Indeed, but I think when you access a table the "group" parameter is determined by the structure of that data, not how you want to use it.  So it should be common that the group depth is the same for the same data.

What I'm trying to say is maybe we don't add that hashing complexity until it is needed.

PipoCanaja:
Theoretically, you are right ;)  
The only question I have is how to troubleshoot the issue that would appear in case we have somebody accessing the same OID with the group but different depths.
I'll give it a try, sometimes during the Xmas break. And the discussion will continue afterward.

LibreNMS/OS.php
```diff
@@ -130,21 +130,22 @@ public function getCacheByIndex($oid, $mib = null, $snmpflags = '-OQUs')
...
      */
-    public function getCacheTable($oid, $mib = null)
+    public function getCacheTable($oid, $mib = null, $depth = null)
     {
         if (str_contains($oid, '.')) {
             echo "Error: don't use this with numeric oids!\n";
             return null;
         }
 
         if (!isset($this->cache[$oid])) {
```
murrant: This isset doesn't match the lines below.
You shouldn't need $depth as a key right?  Snmp tables have a fixed number of items in the index of a table.

PipoCanaja: Unless you can be sure that nobody will ever call the function twice with different $depth ... 
The point here is that for complex tables, you may have a situation where you want to see them with different values at different places of the code (to loop over it in a different way). Having this cached without $depth as a key, means that a piece of code my ask for a depth and receive a cached value with another depth (and fail ...). 
I agree this is probably not so often, but it would be fairly difficult to troubleshoot. As a security and for coherence of the result of the cached value compared to the "non cached" equivalent, a nested hastable does not cost that much CPU (LibreNMS waits for IO 99% of the time (Network IO, Disk IO ...)).

Nice catch for the isset, missed it ! :(

includes/definitions/discovery/vrp.yaml
```diff
@@ -30,7 +30,15 @@ modules:
...
-                    skip_value_lt: -50
+                    skip_values: 
+                        - 
+                            oid: hwEntityOpticalMode
+                            op: '='
+                            value: '1'
+                        - 
+                            oid: hwEntityOpticalTemperature
+                            op: '<'
+                            value: '-50'
```
PipoCanaja: the AC6605 does add 'garbage' values to the table if the sensors are meaningless, so more filtering is necessary. Does not break existing families (as far as Travis, and my production says)

LibreNMS/OS.php
murrant: Depth is dictated by the how many values are in the index of the snmp table (you can see this in the mib).  This really can't change.

PipoCanaja: In a perfectly defined mib, why not, but in real life, you sometimes try multiple depths in order to make your life easier in the loop.
Caching the depths values acts as a check. The same reason we check parameters in a function call. Useless until somebody calls the function with the wrong parameters :)
$depth as key can't be wrong and does not cost much of memory/performance. Not caching it exposes a risk if the function is misused (on purpose or by mistake). 

If you prefer that I remove it, I will of course.

LibreNMS/OS.php
```diff
@@ -130,21 +130,22 @@ public function getCacheByIndex($oid, $mib = null, $snmpflags = '-OQUs')
      * DO NOT use numeric oids with this function! The snmp result must contain only one oid.
      *
      * @param string $oid textual oid
-     * @param string $mib mib for this oid
+     * @param string $mib mib for this oid (optional)
+     * @param string $depth depth for snmpwalk_group (optional)
      * @return array array indexed by the snmp index with the value as the data returned by snmp
      */
-    public function getCacheTable($oid, $mib = null)
+    public function getCacheTable($oid, $mib = null, $depth = null)
```
murrant: $depth should default to 1, not null otherwise default is 0 instead of 1 like parent function.

PipoCanaja: done

PipoCanaja:
OK. I'll probably update that when my devices (currently in the lab and testing phase) are moved to production, but this will take another couple of months so let's go.

PipoCanaja:
@murrant OK for me to merge.

PR #9522, opened 2018-12-05
get geolocation at first poll

pcw11211:
based on the logic of code, we will have to wait 2 days from adding new device for lat and lng to be updated

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

based on the code, system will populate lat and lng only after 2 day after adding new device

(if lat & lang is not null system will never update lat and lng)

there is no need to wait for 2 days


murrant:
Keep mind, it will only query on every poll if it has failed every time before.  Once geocoding succeeds for a location, it is not run again.

PR #9531, opened 2018-12-07
feature: Macros for xDP neighbours alert rules

PipoCanaja:
Simple macro to ease alert rules checking if an xDP entry exists and if this entry points to a known device. 

"port_has_xdp_neighbours": "(%macros.port && %links.local_port_id = %ports.port_id)",
We check here if we have a port and if an entry in links (all neighbours known in Librenms) matches the current port_id. 

"port_has_xdp_neighbours_device": "(%macros.port_has_neighbours && (%links.remote_port_id IS NOT NULL))",
We add to previous macro a check remote_port_id. If it is NOT NULL, then we know the remote_port_id is known to LibreNMS. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/misc/macros.json
+++ b/misc/macros.json
@@ -25,6 +25,8 @@
     "port_in_usage_perc": "((%ports.ifInOctets_rate*8) \/ %ports.ifSpeed)*100",
     "port_out_usage_perc": "((%ports.ifOutOctets_rate*8) \/ %ports.ifSpeed)*100",
     "port_now_down": "%ports.ifOperStatus != %ports.ifOperStatus_prev && %ports.ifOperStatus_prev = \"up\" && %ports.ifAdminStatus = \"up\" && %macros.port",
+    "port_has_xdp_neighbours": "(%macros.port && %links.local_port_id = %ports.port_id)",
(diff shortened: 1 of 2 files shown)


misc/macros.json
```diff
@@ -25,6 +25,8 @@
     "port_in_usage_perc": "((%ports.ifInOctets_rate*8) \/ %ports.ifSpeed)*100",
     "port_out_usage_perc": "((%ports.ifOutOctets_rate*8) \/ %ports.ifSpeed)*100",
     "port_now_down": "%ports.ifOperStatus != %ports.ifOperStatus_prev && %ports.ifOperStatus_prev = \"up\" && %ports.ifAdminStatus = \"up\" && %macros.port",
+    "port_has_xdp_neighbours": "(%macros.port && %links.local_port_id = %ports.port_id)",
+    "port_has_xdp_neighbours_device": "(%macros.port_has_neighbours && (%links.remote_port_id IS NOT NULL))",
```
laf: This doesn't seem to work and I think it's because you reference another macro which then references another macro.

PipoCanaja: It does not work because I should not commit before going to bed and just after renaming a macro :D With all testing done before renaming ... 

Anyway, With the correct name, it works beautifully.

PR #9534, opened 2018-12-10
webui: Improve alert details display for bgp and sensors

PipoCanaja:
This change is improving the default display of alerts for BGP Peers and Sensors. The best solution would probably be, IMO, to use the templating system that is already creating emails and all other transports to create this text. This would allow everybody to customize the alerts display in the GUI. 

But this looks like a too big change for me right now. So until then, this PR will give more details and links for BGP and Sensor based alerts. 

capture d ecran 2018-12-10 a 21 59 44

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/functions.inc.php
+++ b/html/includes/functions.inc.php
@@ -1193,7 +1193,7 @@ function alert_details($details)
 
         if ($tmp_alerts['sensor_id']) {
             $details = "Current Value: " . $tmp_alerts['sensor_current'] . " (" . $tmp_alerts['sensor_class'] . ")<br>  ";
-            $details_a[] = array();
+            $details_a = array();
 
             if ($tmp_alerts['sensor_limit_low']) {
                 $details_a[] = "low: " . $tmp_alerts['sensor_limit_low'];
--- a/html/includes/functions.inc.php
+++ b/html/includes/functions.inc.php
@@ -1193,7 +1193,7 @@ function alert_details($details)
 
         if ($tmp_alerts['sensor_id']) {
             $details = "Current Value: " . $tmp_alerts['sensor_current'] . " (" . $tmp_alerts['sensor_class'] . ")<br>  ";
-            $details_a = array();
+            $details_a = [];
 
             if ($tmp_alerts['sensor_limit_low']) {
                 $details_a[] = "low: " . $tmp_alerts['sensor_limit_low'];
--- a/html/includes/functions.inc.php
+++ b/html/includes/functions.inc.php
@@ -720,23 +720,6 @@ function generate_sensor_link($args, $text = null, $type = null)
         $args['graph_type'] = "sensor_" . $type;
     }
 
-    $details = "Current Value: " . $args['sensor_current'] . " (" . $args['sensor_class'] . ")<br>  ";
-
-    if ($args['sensor_limit_low']) {
-        $details_a[] = "low: " . $args['sensor_limit_low'];
-    }
-    if ($args['sensor_limit_low_warn']) {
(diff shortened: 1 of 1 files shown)


html/includes/functions.inc.php
```diff
@@ -704,7 +704,62 @@ function generate_port_link($port, $text = null, $type = null, $overlib = 1, $si
...
+    }
+
+    if (!isset($args['hostname'])) {
+        $args = array_merge($args, device_by_id_cache($args['device_id']));
+    }
+
+    $content = '<div class=list-large>' . $text . '</div>';
 
+    $content .= "<div style=\'width: 850px\'>";
+    $graph_array = array();
```
murrant: can you define this array like this?
```php
$graph_array = [
    'type' => $args['graph_type'],
    'legend' => 'yes',
    ...
];
```

html/includes/functions.inc.php
```diff
@@ -704,7 +704,62 @@ function generate_port_link($port, $text = null, $type = null, $overlib = 1, $si
     }
 }//end generate_port_link()
 
+function generate_sensor_link($args, $text = null, $type = null)
```
murrant: usually on these functions the array argument is passed directly to the graph link function.  Also, you should probably pass the sensor as a separate parameter.

html/includes/functions.inc.php
```diff
@@ -1133,6 +1188,37 @@ function alert_details($details)
...
+            if ($tmp_alerts['sensor_limit_low_warn']) {
+                $details_a[]= "low_warn: " . $tmp_alerts['sensor_limit_low_warn'];
+            }
+            if ($tmp_alerts['sensor_limit_warn']) {
+                $details_a[]= "high_warn: " . $tmp_alerts['sensor_limit_warn'];
+            }
+            if ($tmp_alerts['sensor_limit']) {
+                $details_a[]= "high: " . $tmp_alerts['sensor_limit'];
+            }
+            $details .= implode(', ', $details_a);
```
murrant: Scrutinizer: "The variable ``$details_a`` does not seem to be defined for all execution paths leading up to this point."

html/includes/functions.inc.php
```diff
@@ -704,7 +704,62 @@ function generate_port_link($port, $text = null, $type = null, $overlib = 1, $si
...
+        return overlib_link($url, $text, $content, null);
+    } else {
+        return fixifName($text);
+    }
+}//end generate_sensor_link()
+
+
+function generate_sensor_url($sensor, $vars = array())
+{
+    return generate_url(array('page' => 'graphs', 'id' => $sensor['sensor_id'], 'type' => $sensor['graph_type'], 'from' => $config['time']['day']), $vars);
```
murrant: $config was not pulled from global here.  Instead of doing that, use the \LibreNMS\Config class.  We are slowly removing $config from the code.

PipoCanaja: done

html/includes/functions.inc.php
PipoCanaja: I used the same structure of code as ports, accesspoints... Except that the var is named $args instead of $sensor (the content is anyway $tmp_alerts). 
I would prefer to avoid touching all of that code. I think that if we refactor all, the best would be to use the Alert class and have the templating system do that job completely. 
So what change would you exactly expect at this step ? only renaming $args to $sensor ?

html/includes/functions.inc.php
PipoCanaja: done

html/includes/functions.inc.php
PipoCanaja: done

PR #9536, opened 2018-12-11
Fixed openstreetmap support in device not using custom tiles

laf:
Fixes: https://community.librenms.org/t/config-leaflet-tile-url-for-local-tile-server-not-working-on-device-page/6486


murrant:
@laf looks good but you missed the init_map() call in resources/views/locations.blade.php

laf:
Done.

PR #9544, opened 2018-12-14
Added automatic generation of the list of available transports (alerting)

hexdump0x0200:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This patch allows you to avoid manually adding a transport to the list of available transports. In addition, in custom installations, where their own transports were added, the need to manually update the file "html/includes/modal/edit_alert_transport.inc.php" broke the automatic update (git detected a conflict and required manual solution). With this patch, you can avoid this, add local transports that remain outside the git repository and there is no need to create and maintain your fork in order to add your transports or add them to the upstream (they may have no value outside the local installation).

### DIFF
--- a/html/includes/modal/edit_alert_transport.inc.php
+++ b/html/includes/modal/edit_alert_transport.inc.php
@@ -43,11 +43,15 @@ if (LegacyAuth::user()->hasGlobalAdmin()) {
 
 // Create list of transport
 $transport_dir = Config::get('install_dir').'/LibreNMS/Alert/Transport';
+$transports_list = array();
 foreach (scandir($transport_dir) as $transport) {
     $transport = strstr($transport, '.', true);
     if (empty($transport)) {
         continue;
     }
+    $transports_list[] = $transport;
+}
+foreach ($transports_list as $transport) {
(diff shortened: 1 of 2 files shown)


html/includes/modal/edit_alert_transport.inc.php
```diff
@@ -39,37 +39,18 @@
...
-                                    <option value="smsfeedback-form">SMSFeedback</option>
-                                    <option value="syslog-form">Syslog</option>
-                                    <option value="telegram-form">Telegram</option>
-                                    <option value="twilio-form">Twilio</option>
-                                    <option value="victorops-form">Victorops</option>
-                                    <!--Insert more transport type options here has support is added. Value should be: [transport_name]-form -->
+<?php
+
+// Create list of transport
+$transport_dir = Config::get('install_dir').'/LibreNMS/Alert/Transport';
```
laf: We already do all of this later on. It would make more sense to store the results of `scandir()` and then re-use that later as well. Make sense?

hexdump0x0200: Done. Also removed duplicate code.

PR #9547, opened 2018-12-14
Add support for Eaton Matrix Telecom Inverters

peelman:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/eaton-matrix.inc.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+$hardware = trim(snmp_get($device, "matConName.0", "-OQv", "TELECOM-MIB"), '" ');
```
laf: Can you convert these 3 to a multi get please.

peelman:
@murrant the docs need to be updated for num_oid then:

Via [here](https://docs.librenms.org/Developing/os/Health-Information/)
> This is the numerical OID that contains value. This should always be without the appended index.

Will amend it for here.  I will also do teh capture for the snmprec.  Didn't realize there was a script for that.

includes/polling/os/eaton-matrix.inc.php
peelman: Wilco.  Won't be able to get to this until Monday, but will try to do them first thing so you guys can review at your leisure.

PR #9550, opened 2018-12-15
webui: translate state in health/state display

PipoCanaja:
Hi

Goal of this PR is to display the state translation in the state tabs of the GUI. 

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/device/health/sensors.inc.php
+++ b/html/pages/device/health/sensors.inc.php
@@ -22,7 +22,7 @@ foreach (dbFetchRows('SELECT * FROM `sensors` WHERE `sensor_class` = ? AND `devi
     if (($graph_type == 'sensor_state') && !empty($state_translation['0']['state_descr'])) {
         $sensor_current = get_state_label($sensor['state_generic_value'], $state_translation[0]['state_descr'] . " (".$sensor['sensor_current'].")");
     } else {
-        $current_label = get_state_label_color($sensor);
+        $current_label = get_sensor_label_color($sensor);
         $sensor_current = "<span class='label $current_label'>".format_si($sensor['sensor_current']).$unit."</span>";
     }
 
--- a/html/pages/device/health/sensors.inc.php
+++ b/html/pages/device/health/sensors.inc.php
@@ -19,7 +19,21 @@ foreach (dbFetchRows('SELECT * FROM `sensors` WHERE `sensor_class` = ? AND `devi
         $sensor_descr = $sensor['sensor_descr'];
     }
 
-    $sensor_current = format_si($sensor['sensor_current']).$unit;
+    $current_label = "label-success";
+    if (!is_null($sensor['sensor_limit_warn']) && $sensor['sensor_current'] > $sensor['sensor_limit_warn']) {
+        $current_label = "label-warning";
+    }
+    if (!is_null($sensor['sensor_limit_low_warn']) && $sensor['sensor_current'] < $sensor['sensor_limit_low_warn']) {
+        $current_label = "label-warning";
+    }
+    if (!is_null($sensor['sensor_limit']) && $sensor['sensor_current'] > $sensor['sensor_limit']) {
+        $current_label = "label-danger";
+    }
+    if (!is_null($sensor['sensor_limit_low']) && $sensor['sensor_current'] < $sensor['sensor_limit_low']) {
+        $current_label = "label-danger";
+    }
+
+    $sensor_current = "<span class='label $current_label'>".format_si($sensor['sensor_current']).$unit."</span>";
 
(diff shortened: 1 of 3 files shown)


PipoCanaja:
Only the health tab of a device is done right now. 

capture d ecran 2018-12-15 a 23 34 31

I also had to change a little bit the min/max display for all sensors, in order to check for NULLity cause the format_si() function does loose this information, which replaced any undefined min or max limit with a wrong '0'. 

capture d ecran 2018-12-15 a 23 07 42

Next to come, the main state list.

html/pages/device/health/sensors.inc.php
```diff
@@ -16,12 +20,28 @@
...
+
     echo "<div class='panel panel-default'>
-            <div class='panel-heading'>
-                <h3 class='panel-title'>$sensor_descr <div class='pull-right'>$sensor_current | $sensor_limit_low <> $sensor_limit</div></h3>
-            </div>";
+        <div class='panel-heading'>
+        <h3 class='panel-title'>$sensor_descr <div class='pull-right'>$sensor_current";
+
+    //Display low and high limit if they are not null (format_si() is changing null to '0')
+    if (!is_null($sensor['sensor_limit_low'])) {
```
laf: All looks except I don't like the low / high values being in red. Looks like they are triggering something rather than just informational.

Might even be worth putting them in label boxes like the current value so it's all in keeping. You could change the label colour based on it being high or low and even if it's breached the current value.

PipoCanaja: Yep, agree. Red is not nice, it was an attempt to make it "different" than the value itself. Will try to see what I can do with label boxes.

PipoCanaja: With a standard value, the label display is pretty nice. 
capture d ecran 2018-12-15 a 23 54 41

But if you have low/high limit associated with a state, which is not forbidden, and works well (I use it in production), the display is somehow unclear : 
capture d ecran 2018-12-15 a 23 55 32

What do you think ?

laf: I like the second way more but what about having orange and blue or something for high / low and then red if they are breached?

PipoCanaja: I would say that too many colors might become difficult to read as well. 

I can try to move "current value" into label (that would be green or red depending on the conformity with min/max values)

And min/max values themselves, if availiable, would remain gray (label-default). 

But that won't be for today ;)

murrant: Maybe instead of the the high low use arrow icons?

PipoCanaja: Here is the variant with text labels, including warning limits, and current_value get its color based on the 4 limits. The limits themselves are label-default.

For icons, we would need to make it clear if it is limit or warn_limit, and if we are the low or high limit. I don't see "visually" how to do that. If somebody has a sketch of it, feel free. 

capture d ecran 2018-12-16 a 12 14 54

murrant:
```
-----
 / \
/ | \
  |
```

Something like that? I couldn't find a font awesome icon that matches :(
Also tooltip text with a full description of what the field is helps too.

laf: I'd much prefer avoiding just using icons that rely on people hovering over to work out what they are. It's not really intuitive (we do this elsewhere in the webui and it confuses people).

PipoCanaja:
All right, let's remove WIP then.

PipoCanaja:
Sorry to dismiss the approval, but instead of a new PR, let's continue the work in this one. 
- The display into the /health/metric=state/view=details is now also showing the state in textual form. Makes it easier to use. 
- The other metrics now get the label with color as well
- the color checking is moved to a function. 

capture d ecran 2018-12-19 a 00 07 02
capture d ecran 2018-12-19 a 00 14 37

murrant:
Looks great.

Can we trim the trailing 0s from the sensor values?

Hint: "1.00000" + 0 == 1

PipoCanaja:
> Can we trim the trailing 0s from the sensor values? 
> Hint: "1.00000" + 0 == 1

Yes, looks like the approval got dismissed again ;)

murrant:
Travis failure is unrelated, merging

PipoCanaja:
It seems adding 0 did break the m of k multiplicator, so it breaks the display of any mW, mA or kV ... I open a new PR to try to solve this.

PipoCanaja:
I think the best way is to patch "format_si()" instead.

murrant:
Thaks

PipoCanaja:
@murrant : done in #9575

PR #9555, opened 2018-12-17
Adding Sonicwall SMA 400 support

marvink87:
- Added metric clients for use in SSL VPN client alerting
- Support for specifying multiple sensors for clients metric
- Added health sensor for SSL VPN clients
- RAM info supported
- CPU info supported
- Added test data
- Updated mib files

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


marvink87:
I had trouble squashing the merge from upstream. Github told me it was unable to merge automatically after that, so that's why it's 2 commits.

marvink87:
When specifying multiple sensors the individual sensor limit's aren't really useful in the graph. I noticed I left those in right after the merge request. Just added another commit to remove these dashed lines.

murrant:
This is the fourth PR to add a Count Sensor type.  We're still a little unsure about how to do it.

murrant:

murrant:
@marvink87 I updated your pull to use the new count sensor.  But the snmprec is missing some oids used by your code.  Can you re-capture the snmprec with this PR applied?  Let me know if you need an help.

marvink87:
@murrant I just updated the test data

murrant:
Thanks.

marvink87:
Now that you've defined sensor groups the metric doesn't have to be as generic as count, maybe 'clients' would be a better fit?

Everything seems to be working though. The only fault I noticed is a cosmetic one. In the health summary the graph is weird, the y axis shows numbers using exponentials which isn't needed for a number around 30.

Also I noticed you removed the ability to aggregate multiple sensors. Should I open a new issue for this feature or is there already a generic way to do this?

murrant:
The sensor description "Clients" should be enough.  You could add a "Clients" group, but it just adds another header in the ui, but you only have one "count" sensor so it doesn't seem helpful.

marvink87:
I noticed not all data was updated so I deleted the snmprec & json files and recreated them.

@murrant Travis tests will probably fail again for the ports. Should I manually add ifAlias like you did? Is this a known bug or am I doing something wrong?

murrant:
@marvink87 it is failing because when the tests are run it uses the code in master (which now fills in ifAlias in discovery).  If you rebased/merged master, and recreate the test data it will be correct.  Or you could just fill in ifAlias by hand ;)

marvink87:
@murrant Still the same error, I tried manually editing but couldn't figure out what to edit yet. The result was that pre-commit failed on my own machine as well.

I think it has something to do that the device doesn't actually return a ifAlias as is seen in the snmprec, maybe there's an assumption made in save-test-data.php?

murrant:
@marvink87 now tests are failing because you have an if ignore rule set :)

PR #9570, opened 2018-12-20
Storage Graph that shows used Bytes rather than a percentage utilization

sinky:
Usefull if volume can grow in total size. Which causes the percentage Graph to drop.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/graphs/storage/used.inc.php
+++ b/html/includes/graphs/storage/used.inc.php
@@ -1,0 +1,60 @@
+<?php
+
+$scale_min = '0';
+$scale_max = '100';
+
+require 'includes/graphs/common.inc.php';
+
+$rrd_options .= ' -b 1024';
+
+$iter = '1';
+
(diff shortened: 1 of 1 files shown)


laf:
Do you just see people calling for this graph manually for now?

I wonder if we document this in an FAQ for now if it was to be merged?

murrant:
He means there isn't an obvious way to find this graph.  You could only find it by going to the single graph view and selecting it from the dropdown.  Which means basically only you would be using it.  To merge it other people should be able to benefit.

sinky:
Mhh, the DropDown is there anyway, now with only one item. So why not to provide the posibility to change the graph?

Example to demonstrate:
In the middle there is an drop in usage about 10%.
graph_usage

But this is not a result of deleting data. It is because of the increase in the total disk space of the volume.
graph_used

murrant:
@sinky Good graph, but it needs to be discover-able so people can benefit. Feel free to resubmit.

PR #9578, opened 2018-12-21
Update Installation-Ubuntu-1804-Apache.md

Function-10:
Added commands to save /etc/mysql/mariadb.conf.d/50-server.cnf and exit vi editor

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Installation/Installation-CentOS-7-Apache.md
+++ b/doc/Installation/Installation-CentOS-7-Apache.md
@@ -46,6 +46,11 @@ Within the `[mysqld]` section please add:
 innodb_file_per_table=1
 lower_case_table_names=0
 ```
+
+Save the configuration file and exit the text editor
+
+Restart mysql
+
     systemctl enable mariadb
     systemctl restart mariadb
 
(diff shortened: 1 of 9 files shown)


murrant:
This should probably be a "tip" further up about how to exit VI.  And it should be in all the install docs, not just one.

laf:
We tell people to use vi a few times but imho it's not our place to teach people how to use an editor.

Function-10:
Updated document to only give instructions to exit the text editor.
Removed keyboard input.

laf:
Thanks, this does need updating in all of the install guides though to remain consistent.

Function-10:
> Thanks, this does need updating in all of the install guides though to remain consistent.

Sorry, I thought I had. Do I need to submit a new pull request?

laf:
> Sorry, I thought I had. Do I need to submit a new pull request?

No just make sure you edit the files within the same branch: https://github.com/Function-10/librenms/tree/patch-1

murrant:
I don't think this is adding much and would prefer to keep the install docs as succinct as possible.

PR #9592, opened 2018-12-24
VRP: implement NAC data polling (WIP)

PipoCanaja:
Hello

Here is a simple implementation with Huawei VRP devices pushing into the existing fields. I am wondering if we could extend in a flexible way the data model to adapt to different MIB philosophy of multiple vendors : 
https://community.librenms.org/t/huawei-nac-and-nac-model/6587

We should keep this in WIP for the moment.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hi
I will add the test data (probably next week so I can test a little bit before) but it seems that the result is not bad.
vlan and time_elapsed column were added, and by default, a huawei device will have these visible. On all others, I kept the exact same columns visible (but the vlan column could probably be default visible. )
Here is default display for huawei switch. 
capture d ecran 2019-01-08 a 15 38 02

murrant:
@PipoCanaja you have some upstream commits showing up here.  You will need to rebase to remove them.

PipoCanaja:
> @PipoCanaja you have some upstream commits showing up here. You will need to rebase to remove them.

Yes, git is just magic to me, sometimes ...

html/pages/device/nac.inc.php
```diff
@@ -49,10 +67,26 @@
             };
         },
         formatters: {
+            "time_interval": function (column, row) {
+                var value = row[column.id];
+                if (value > 59) {
+                    var date = new Date(null);
```
murrant: you could probably use moment.js here

PipoCanaja: OK. Will check that.

murrant:
The migration needs to be converted for this PR https://laravel.com/docs/5.4/migrations

PipoCanaja:
@murrant Here is the result with moment.duration(xxxxx).humanized(). The value is clearly less precise, what do you think ?

capture d ecran 2019-01-16 a 23 07 53

murrant:
@PipoCanaja I figured you would do something like this: 
```javascript
function formatDuration(seconds) {
    var duration = moment.duration(seconds, 's');
    var years = duration.years(),
        months = duration.months(),
        days = duration.days(),
        hrs = duration.hours(),
        mins = duration.minutes(),
        secs = duration.seconds();

    var output = '';
    if (years) {
        output += years + 'y ';
    }
    if (months) {
        output += months + 'm ';
    } else {
        if (days) {
            output += days + 'd ';
        } else {
            output += hrs + ':' + mins + ':' + secs;
        }
    }

    return output.trim();
}
```

PR #9594, opened 2018-12-27
WIP Huawei GPON OLT improvements

jozefrebjak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


TheGreatDoc:
Amazing @jozefrebjak 

BUT this is going to be a mess. We need to match old sensors index to be the same or old will be removed (with their historical data) and new ones added.

jozefrebjak:
@TheGreatDoc Yes it will be mess, but as for me I don't care about history. I don't know if will be possible to use old index, if yes it will be great but I don't know how.

jozefrebjak:
@TheGreatDoc we also need to change how is OLT discovered. Now it is used for discovery: sysDescr: - 'Huawei Integrated Access Software', what is used also for Huawei MDU in our case ma5620, what is not very good. We need to use sysObjectID I believe for better discovery, also there is different sensors for OLT and MDU. Will we need to make another os for example smartax-mdu. Some of the sensors from out are working on the MDU too, but some not.

murrant:
@jozefrebjak looks like some nice improvement.

When I am doing these changes.  I usually use the testing system to minimize the user impact.
Create new tests file before the changes.  Switch to the branch with the changes and run tests.  Make as many items match as I reasonably can then.  Once I have it in a good state update the test data to fix the ones I want to change (like improved descriptions, etc).

murrant:
@jozefrebjak Pretty close here.  Some devices that used to be detected as smartax no longer are.

jozefrebjak:
@laf i edited some indexes, so now will be used old rrd for processors and for total power of frame. I tried to use hwslotindex for temeprature of boards, but its does not work. So I leave a new indexes for that, @murrant  which devices do you think ? I am tryinig to make support for smartax-mdu for example, what is someting different device than this OLT for example, so for example I dont want to be that device detected with same os, where is different type of device and that device has different sensors etc. So I dont know, will we make another smartox-olt os for example and leave old smartox ?

PR #9606, opened 2019-01-02
Allow grouping of sensors

murrant:
Improves visual identification for sensors

 - [x] Docs

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
I like the idea of this. However is it the best thing to allow the groups to be absolutely anything? I'm not sure either way but it strikes me as personal at that point and one persons opinion of a group might not be anothers?

murrant:
@laf Yeah, I think so.  Right now groups only apply within a single OS.  We don't use them on all sensors.

I think of group like an meta description. If they need to be standardized more later, we can do that.  I don't have enough info about how this will be used now to figure that out.

murrant:
@librenms/reviewers This is ready.  I'll probably merge if there a no objections.

PR #9622, opened 2019-01-07
Nokia vrf bgp

vitalisator:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

##### Changes

This is my try to implement MPLS VRF instances. (not cisco vrf-lite)
this also adds support for BGP over VPRN (virtual private routed network) on Nokia routers
That works for me fine, but I'm not a programmer, so feel free to refactor or improve.
This PR currently not include the GUI implementation of VRF. (mybe one can do that or at lest give me a advice on that). So all VRF BGP Sessions are currently under main routing Tab.


includes/discovery/bgp-peers.inc.php
```diff
@@ -5,6 +5,71 @@
 use LibreNMS\Util\IP;
 
 if (Config::get('enable_bgp')) {
+    if ($device['os'] == 'timos') {
+        $bgpPeersCache =snmpwalk_cache_multi_oid($device, 'tBgpPeerNgTable', [], 'TIMETRA-BGP-MIB', 'nokia');
```
murrant: the mib directory 'nokia' is extraneous here, it is already defined in the os yaml.

vitalisator: in the timos.yaml is defined group: nokia, I think that what you mean

vitalisator:
@murrant 
bgpLocalAs is one per vrf instance. This is additionally. The one under device table is still present for the Base routing instance. 
test data - ok.

> You probably need to update tests/module_tables.yaml also to account for the new vrf reference (similar to how it works for bgpLocalAs).
I'm currently not really understand the purpose for tests/module_tables.yaml. You mean somewhat like this? 
 - { left: bgpPeers.vrf_id, right: vrfs.vrf_id, select: [bgpLocalAs] }

murrant:
You need to run `./scripts/build-schema.php` to update the schema verification file.

vitalisator:
should be ready for merge now.

vitalisator:
@murrant, thank you for advice.

murrant:
@vitalisator I've update modules_tables.yaml, I think it is right now.

I added the test data to my dev librenms and it looks a little odd.  There is no local IP and the poller says no bgp peers.

vitalisator:
@murrant I've updated test data, please test again.

vitalisator:
```
SELECT * FROM `vrfs` where device_id = 13;                                                                                                                                                                                                                                                                        
+--------+---------+-----------------+------------+------------------------------+-----------------------+-----------+                                                                                                                                                                                                                
| vrf_id | vrf_oid | vrf_name        | bgpLocalAs | mplsVpnVrfRouteDistinguisher | mplsVpnVrfDescription | device_id |                                                                                                                                                                                                                
+--------+---------+-----------------+------------+------------------------------+-----------------------+-----------+                                                                                                                                                                                                                
|      1 | 1       | Base            |          0 | NULL                         | Base                  |        13 |                                                                                                                                                                                                                
|      2 | 3       | vprn118208010   |      39835 | 39835:118208010              | vprn118208010         |        13 |                                                                                                                                                                                                                
|      3 | 4094    | vpls-management |          0 | NULL                         | vpls-management       |        13 |                                                                                                                                                                                                                
|      4 | 4095    | management      |          0 | NULL                         | management            |        13 |                                                                                                                                                                                                                
+--------+---------+-----------------+------------+------------------------------+-----------------------+-----------+                                                                                                                                                                                                                
4 rows in set (0.00 sec)
```

In this case the only one VRF instance has a Local AS configured, the others are default management instances. Or what exactly is the problem?

murrant:
Yes, I saw that.  But none of the get local AS.  Check the Travis CI output.

murrant:
The database migration will need to be converted: https://laravel.com/docs/migrations

database/migrations/2019_01_16_195644_add_vrf_id_and_bgpLocalAs.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('bgpPeers', function (Blueprint $table) {
+            $table->integer('vrf_id')->after('device_id');
```
murrant: Should vrf_id be nullable?

vitalisator:
@murrant I think this one can be merged now.

murrant:
Thanks Vitali :)

PR #9643, opened 2019-01-10
Fix if discover

evheros:
This PR is te revert 1 change in PR #9493. In that PR the discovery is updating IfName, IfAlias and IfDescr. 

As in the Poller there are more checks the discovery will overwrite or NULL some fields. This PR will also copy IfName, IfAlias and IfDescr to empty IfName, IfAlias and IfDescr the same way as the poller is doing.

Also i've removed the IfAlias update as this was overwriting it when it was set manually.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@PipoCanaja does this affect the change you made, or does it still work as intended?

PipoCanaja:
Basically, the idea is to get as much as possible from discovery cause if you have selective port polling, then the poller does not collect data if port is admin down. So if discovery does not do it either, then the port does not get updated at all, and we lack some info in LibreNMS. 

My main concern was on ifType , and that triggered the change on my side. (we have a plugin here that is interested in ifType value for operationnal tasks).

So I would probably go with the version were the tests in discovery are done the same way as polling. And discovery itself is not as "time constrained" as polling, so it should be all right in that matter.

evheros:
Added the ifname, ifalias and ifdescr copy if it is emtpy or null

ToDo: Removal of Transit: in description

PipoCanaja:
Hi,

Seems to be a good way to do it. I'll try it on lab and let you know. 

PipoCanaja

PipoCanaja:
Looking at the results of the tests, it appears that it changes the behaviour. -> More interfaces are seen. Right now, I don't really see why in fact ... 
@murrant : Any idea why travis sees more interfaces now than before ?

evheros:
I realy would like this patch to be included in 1.48 as the previous update introduced some problems on production systems.

@PipoCanaja and @murrant  Is this blocking or can this PR beeing merged?

includes/discovery/ports.inc.php
```diff
@@ -44,9 +44,25 @@
 foreach ($port_stats as $ifIndex => $port) {
     // Store ifIndex in port entry and prefetch ifName as we'll need it multiple times
     $port['ifIndex'] = $ifIndex;
-    // handle ifName and ifAlias the same way we do in the poller
-    $ifName = empty($port['ifName']) ? $port['ifDescr'] : $port['ifName'];
-    $ifAlias = empty($port['ifAlias']) ? $port['ifDescr'] : $port['ifAlias'];
+
+    // When devices do not provide ifDescr data, populate with ifName data if available
```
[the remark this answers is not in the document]
evheros: I've added the 'is_port_valid' function before doing the copy actions so no new ports are added. The travis test still fails but as far as i can see this is "correct" because we are now updating the fields that previously were empty and were filed by the poller. 
Discovery is now filling them the same way the poller does.

evheros:
I've added the 'is_port_valid' function before doing the copy actions so no new ports are added. The travis test still fails but as far as i can see this is "correct" because we are now updating the fields that previously were empty and were filed by the poller.
Discovery is now filling them the same way the poller does, so should not break anything. Only update info on ports previous not beeing updated.

includes/discovery/ports.inc.php
```diff
@@ -64,7 +82,7 @@
             $ports_db[$port_id]['deleted'] = '0';
             echo 'U';
         } else { // port is existing, let's update it with some data we have collected here
-            dbUpdate(array('ifType' => $ifType, 'ifName' => $ifName, 'ifAlias' => $ifAlias, 'ifDescr' => $ifDescr), 'ports', '`port_id` = ?', array($port_id));
```
murrant: Why did you remove ifAlias update?

evheros: If you set the description manualy in the webinterface the ifAlias is changed in the DB. If the discover is running after this it is updating it with the info from the device. And so the manual set description (ifAlias in DB) is reverted back and the manual changed setting is gone.

murrant: Interesting.  Is it worth checking the device attribute to add this info on first discover and not clobber it later?

PipoCanaja: Yep, that would make sense in that case. I was not aware of the manual setting as I do not use it but, the set of ifalias should end up in a dedicated function that does this check anytime an update is tried.

murrant:
```php
        if (!empty($device['attribs']['ifName:' . $port['ifName']])) {
            // ifAlias overridden by user, don't update it
            unset($snmp_data['ifAlias']);
        }
```

(this assumes you've extracted the arrays to a var called $snmp_data)

murrant:
I've taken a crack at this. Thoughts?

PR #9645, opened 2019-01-10
Deprecate PHP < 7.1.3

laf:
Doing this whilst I've got a cold, probably a better way to do it but it should work.

Only tested whilst running php 7.2.4

DO NOT MERGE UNTIL FEB 1st.

### DIFF
--- a/LibreNMS/Validations/Updates.php
+++ b/LibreNMS/Validations/Updates.php
@@ -72,7 +72,7 @@ class Updates extends BaseValidation
                         "You are on the PHP 5.3 support branch, this will prevent automatic updates.",
                         "Update to PHP 5.6.4 or newer (PHP 7.2 recommended) to continue to receive updates."
                     );
-                } elseif ($versions['local_branch'] == 'php70') {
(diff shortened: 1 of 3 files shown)


LibreNMS/Validations/Updates.php
```diff
@@ -70,7 +70,12 @@ public function validate(Validator $validator)
                 if ($versions['local_branch'] == 'php53') {
                     $validator->warn(
                         "You are on the PHP 5.3 support branch, this will prevent automatic updates.",
-                        "Update to PHP 5.6.4 or newer (PHP 7.1 recommended) to continue to receive updates."
+                        "Update to PHP 5.6.4 or newer (PHP 7.2 recommended) to continue to receive updates."
+                    );
+                } elseif ($versions['local_branch'] == 'php70') {
```
murrant: Tested fine, but shouldn't the branch be `php56`?

murrant:
Does it make sens to keep the php53 stuff?  I'm not sure it can be reached.

PR #9651, opened 2019-01-11
Fix victorops recovery alerts

amtypaldos:
VO will not accept an int as an entity ID, this causes issues when issuing recovery alerts as the entity_id is the only way to link incidents.  I have tested this with VO and incidents will alert AND recover.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Victorops.php
+++ b/LibreNMS/Alert/Transport/Victorops.php
@@ -41,7 +41,7 @@ class Victorops extends Transport
         $url = $opts['url'];
 
         $protocol = array(
-            'entity_id' => ($obj['id'] ? $obj['id'] : $obj['uid']),
+            'entity_id' => str($obj['id'] ? $obj['id'] : $obj['uid']),
             'state_start_time' => strtotime($obj['timestamp']),
             'entity_display_name' => $obj['title'],
             'state_message' => $obj['msg'],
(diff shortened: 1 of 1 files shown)


LibreNMS/Alert/Transport/Victorops.php
```diff
@@ -41,7 +41,7 @@ public function contactVictorops($obj, $opts)
         $url = $opts['url'];
 
         $protocol = array(
-            'entity_id' => ($obj['id'] ? $obj['id'] : $obj['uid']),
+            'entity_id' => str($obj['id'] ? $obj['id'] : $obj['uid']),
```
murrant: should be `(string)` to cast to string.

amtypaldos:
```suggestion
            'entity_id' => strval($obj['id'] ? $obj['id'] : $obj['uid']),
```

amtypaldos: Sorry was so excited that I fixed it I forgot the function that PHP used to convert a variable to a string.

murrant: sure

murrant: Make sure you test it with the correct code :)

amtypaldos: Just confirmed that is the correct code I tested with, sorry about the mix up.

PR #9656, opened 2019-01-11
WIP: Poller-group check in poll-billing.php

Swashy:
Poll billing currently has no concept of a distributed architecture, the php code behind it does not check for the poller-groups number that a device belongs to.

For a distributed architecture where polled devices can only be polled by their respective LibreNMS poller, this means a lot of wasted time as the SNMP poll must time-out for every single port.

This change adds in a simple conditional that checks a port's device's poller-group. If the poller-groups don't match, then it won't poll the port.

This is currently working on my own distributed setup. It's dropped the total execution time from 120+ seconds to 5-25 seconds.

Possible improvements:
- Can this check be done with only 1 database query or less?
- Is the config option grabbed properly and accounted for?
- Could port billing avoid SNMP polling and simply grab the data it needs through RRD data? I'm not familiar enough with RRD to say.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/poll-billing.php
+++ b/poll-billing.php
@@ -55,6 +55,23 @@ foreach ($query->get(['bill_id', 'bill_name']) as $bill) {
         $host    = $port_data['hostname'];
         $port    = $port_data['port'];
 
+        // This conditional will check if "distributed_poller_group" is set in config.php.
+        // If it is, check to see if the port we're trying to bill on is for a device in our poller group,
+        // if the poller_groups don't match, don't do billing and continue to next port
+        if (isset($config['distributed_poller_group'])) {
+            // Fetch the associated device ID based on the port_id from the `ports` table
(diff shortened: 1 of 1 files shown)


murrant:
It should accept poller group from the CLI rather than checking the distrubuted poller setting.

The easier way to do this would be to only select ports that are in that poller group.
```php
$port_list = dbFetchRows('SELECT * FROM `bill_ports` as P, `ports` as I, `devices` as D WHERE P.bill_id=? AND I.port_id = P.port_id AND D.device_id = I.device_id AND D.poller_group = ?', [$bill_id, $poller_group]);
```

murrant:
Cannot use RRD data, it is not accurate, this is why billing exists as it is in the first place :)

murrant:
Closing this because of no response.  Feel free to re-open.  It should be updated to have a more specific sql query as I mentioned.

Issue #7541, opened 2017-10-25
Pfsense Memory not polling correct info


kkrumm1:
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our irc channel ##librenms on freenode to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [X] Please include all of the information between the 
`====================================
Component | Version
--------- | -------
LibreNMS  | 1.32-77-gc50ddc3
DB Schema | 211
PHP       | 7.0.24
MySQL     | 5.5.56-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
====================================
[OK]    Database connection successful
[OK]    Database schema correct
` section of `./validate.php` which you can run from the cli.
- [X] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Discovery output: https://gist.github.com/kkrumm1/9fa35764b1590fbe62f5e242577eda13#file-pfsense-discovery
Poller output: https://gist.github.com/kkrumm1/f1978ad1b268ab5eac36903e1962b0ba#file-pfsense-poller
SNMP walk: https://gist.github.com/kkrumm1/8763c82040803c2ccad0e95d0dc4c4a0#file-pfsense-snmp-walk

screenshot from Pfsense 
![image](https://user-images.githubusercontent.com/27959419/31974244-1820fa8a-b8f0-11e7-9c52-55879da61f0c.png)

Screenshot from LibreNMS
![image](https://user-images.githubusercontent.com/27959419/31974282-7c75dcda-b8f0-11e7-8662-414314a4ed1f.png)

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

tomarch:
Pfsense GUI show only used memory (without shared and cache), but librenms show used + shared + cache.
You can see that in the over graph : 
[...]

PR #9659, opened 2019-01-13
FreeBSD mempools Fix

vitalisator:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

#### Fixes
On some (opnsense, pfsense, ...) FreeBSD based devices that using bsnmpd daemon, mempools show wrong value for hrStorageSize.
This fix use the real metrics from hrMemorySize.0 for affected devices.
This behaviour is only on bsnmpd based devices, net-snmp based are not affected.


includes/polling/mempools/hrstorage.inc.php
```diff
@@ -12,5 +14,9 @@
 
 $mempool['units'] = $entry['hrStorageAllocationUnits'];
 $mempool['used']  = ($entry['hrStorageUsed'] * $mempool['units']);
-$mempool['total'] = ($entry['hrStorageSize'] * $mempool['units']);
+if ($device['sysObjectID'] == '.1.3.6.1.4.1.12325.1.1.2.1.1') { // bsnmpd based devices, like FreeBSD, opnsense, pfsense ...
```
murrant: Maybe just check if hrMemorySize is good or hrStorageSize is bad instead of if this is *BSD?

vitalisator: I have no idea how to evaluate if hrMemorySize is good or hrStorageSize is bad, both values are present and can be polled.
But I know that on *BSD the hrStorageSize values is wrong.

murrant: So there are values there, they are just wrong?? That seems really odd.  What are the two values?

vitalisator: here can you read some explanation to this behavior:
https://lists.freebsd.org/pipermail/freebsd-stable/2014-March/078146.html
So that is a old one, but still the same bug on bsnmpd

vitalisator: in #7541 this issue was addressed too, but no one had the right solution :-)

murrant:
@vitalisator fun, ok.  Would you be willing to add some test data for this since it is an edge case?
`./scripts/collect-snmp-data.php -h HOSTNAME`

PR #9665, opened 2019-01-14
Move the HttpAuthAuthorizer guest user check from userExists() to authenticate()

jthiltges:
User creation starts with a call to `userExists()`. When using `auth_mechanism = http-auth` and `http_auth_guest`, `userExists()` returns true and creation fails. Move the guest user check to the `authenticate()` function to avoid this issue.

This change works in my environment, but we're not using guest users, so my testing is limited. I submitted this as WIP for that reason.

This should make HttpAuthAuthorizer `authenticate()` match with [ADAuthorizationAuthorizer](https://github.com/librenms/librenms/blob/1f1fcf56fbd644ccebe2e7e239374fd1b1687a77/LibreNMS/Authentication/ADAuthorizationAuthorizer.php#L60) and [LdapAuthorizationAuthorizer](https://github.com/librenms/librenms/blob/1f1fcf56fbd644ccebe2e7e239374fd1b1687a77/LibreNMS/Authentication/LdapAuthorizationAuthorizer.php#L87).

A forum user also seems to have run across the same issue: https://community.librenms.org/t/adduser-php-says-user-already-exists/1798

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Authentication/HttpAuthAuthorizer.php
+++ b/LibreNMS/Authentication/HttpAuthAuthorizer.php
@@ -18,20 +18,11 @@ class HttpAuthAuthorizer extends MysqlAuthorizer
             return true;
         }
 
-        if (Config::has('http_auth_guest') && parent::userExists(Config::get('http_auth_guest'))) {
+        if (Config::has('http_auth_guest') && $this->userExists(Config::get('http_auth_guest'))) {
             return true;
         }
 
         throw new AuthenticationException('No matching user found and http_auth_guest is not set');
-    }
-
-    public function userExists($username, $throw_exception = false)
-    {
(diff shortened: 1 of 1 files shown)


murrant:
This seems reasonable to me. (Still needs testing)

However, the userExists() function is no longer needs to be overridden.  Delete it from the class file and update all calls to just be $this->userExists();

jthiltges:
Good point on the userExists() function. I've removed it as suggested. Thank you!

murrant:
In testing when the user already exists it tries to add it to the database. Because it checks if the given username exists, but then tries to add the guest username.

image

PR #9669, opened 2019-01-15
Refactor FDB Tables to Laravel

murrant:
Hopefully much better performance with large tables.
Better dns resolution (limit the number of IPs tried per MAC)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


app/Http/Controllers/Table/FdbTablesController.php
```diff
@@ -0,0 +1,270 @@
...
+    protected function findIps($mac_address)
+    {
+        if (!isset($this->ipCache[$mac_address])) {
+            $ips = Ipv4Mac::where('ipv4_mac.mac_address', $mac_address)->pluck('ipv4_address');
+
+            $dns = 'N/A';
+            // don't try too many dns queries, this is the slowest part
+            foreach ($ips->take(2) as $ip) {
+                $hostname = gethostbyaddr($ip);
+                if (!IP::isValid($hostname)) {
```
Jellyfrog: Better to cache all responses, valid or not, won't change per run anyway.

app/Http/Controllers/Table/FdbTablesController.php
```diff
@@ -0,0 +1,270 @@
...
+     * @return \Illuminate\Support\Collection
+     */
+    protected function findIps($mac_address)
+    {
+        if (!isset($this->ipCache[$mac_address])) {
+            $ips = Ipv4Mac::where('ipv4_mac.mac_address', $mac_address)->pluck('ipv4_address');
+
+            $dns = 'N/A';
+            // don't try too many dns queries, this is the slowest part
+            foreach ($ips->take(2) as $ip) {
```
Jellyfrog: Doesn't match with comment;
>Better dns resolution (limit to 4 IPs tried per MAC)

app/Http/Controllers/Table/FdbTablesController.php
murrant: good point

murrant: @Jellyfrog I looked at the code and we cache the results for each mac every time.  So there shouldn't be any duplicated requests.

Jellyfrog: Yeah I misread, sorry.
Wouldn't it make more sense to cache per IP instead of mac? Multiple mac-addresses could have the same IP (different networks).

Jellyfrog: Also maybe making the DNS resolving a config option?

PR #9675, opened 2019-01-16
Syslog Hook for Junipers JunOS

nova-2nd:
Old message is not generated for SRX-110. This one catches CLI and UI config commits

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


scripts/syslog-notify-oxidized.php
```diff
@@ -20,7 +20,7 @@
     oxidized_node_update($hostname, $msg, $matches['user']);
 } elseif (preg_match('/HWCM\/4\/CFGCHANGE/', $msg, $matches)) { //Huawei VRP devices CFGCHANGE syslog
     oxidized_node_update($hostname, $msg);
-} elseif (preg_match('/UI_COMMIT: User \\\\\'(?P<user>.+?)\\\\\' .*/', $msg, $matches)) {
+} elseif (preg_match('/UI_COMMIT_PROGRESS: Commit operation in progress: commit complete/', $msg, $matches)) {
     oxidized_node_update($hostname, $msg, $matches['user']);
```
PipoCanaja: Hi

- The old message is still sent by other devices / versions ? if yes we should not remove it, but only add the COMMIT_PROGRESS one. 
- Is there a way to catch the user doing the change in the New message ? If no, then the oxidized_node_update should be updated accordingly

PR #9679, opened 2019-01-16
Cable Modem Graphs for TopVision OS

djamp42:
Graphs Total/Registered/Offline Cable modems for the TopVision CMTS.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hello @djamp42 

Could you add testdata to your PR ? 
https://github.com/librenms/librenms/blob/master/doc/Developing/os/Test-Units.md

Thanx

djamp42:
Yup, updated test data, thanks

PipoCanaja:
Thanx. But you should also run ./scripts/save-test-data.php etc etc. 

Thanx

djamp42:
> Thanx. But you should also run ./scripts/save-test-data.php.

So when i run the following command

`./scripts/save-test-data.php -o topvision -m os -d `

it reads all the snmprec files and then hangs at listening endpoint. First time i've ever used this so it might be something obvious.  It never updates the topvision.json file.

```
  SNMPv3 USM SecurityName: simulator
  SNMPv3 USM authentication key: auctoritas, authentication protocol: MD5
  SNMPv3 USM encryption (privacy) key: privatus, encryption protocol: DES
  Maximum number of variable bindings in SNMP response: 64
  --- Transport configuration
  Listening at UDP/IPv4 endpoint 127.1.6.1:1161, transport ID 1.3.6.1.6.1.1.0
```


  Listening at UDP/IPv4 endpoint 127.1.6.1:1161, transport ID 1.3.6.1.6.1.1.0

PipoCanaja:
It seems that your snmpsim config is not working properly. It could be the iptables / firewall rules on the machine. If you could troubleshoot in that direction, it would probably solve the issue. (look at the address and the port snmpsim uses, it is clearly non standard)

djamp42:
So this is a custom graph and data doesn't get entered into the database.  @murrant in discord said the following. Can you let me know what else needs to be done if any?

> no, test data doesn't test rrd info, only database, so it doesn't make sense.

PipoCanaja:
Right. Only the snmprec file gets data if the DB is not changed. So this looks good. 

@librenms/reviewers Could somebody else have a look ?

PR #9687, opened 2019-01-17
Device: Axis sensor state fix

kkrumm1:
Fix Axis sensor states for Generic alerts and status. 
0 = OK
1 = Warning
2 = Critical
3 = Unknown

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@kkrumm1 unfortunately, this is the old style state sensor, it will not update existing states in the DB.  (so this basically only fixes it for new installs)  Need to convert to YAML or the new PHP style as shown in the state sensor docs.

PR #9691, opened 2019-01-17
Huawei iBMC absent state grey

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hi,

That was a question to define if absent was normal or not, and I did not really found a proper answer so I stayed with green for absent. 
Basically, a change of state from OK to absent is important. But a disk that was never installed should be green (absent is expected and normal in that case)

I am perfectly ok to change it, and i must say I did not try to find similar sensors to compare the values.

murrant:
Up to you a lot of other MIBs call it 'notpresent'.

@PipoCanaja you choose to merge or close this.  No worries.

PipoCanaja:
The majority of the /not( )?present/i states are already type "3" so I would say ibmc should be treated the same. Let's merge this.

PR #9692, opened 2019-01-17
Add device group filter to widgets

Jellyfrog:
Adds a device group filter to most widgets

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Table/EventlogController.php
+++ b/app/Http/Controllers/Table/EventlogController.php
@@ -36,6 +36,7 @@ class EventlogController extends TableController
     {
         return [
             'device' => 'nullable|int',
+            'device_group' => 'nullable|int',
             'eventtype' => 'nullable|string',
         ];
     }
@@ -61,7 +62,11 @@ class EventlogController extends TableController
      */
     public function baseQuery($request)
     {
-        return Eventlog::hasAccess($request->user())->with('device');
+        return Eventlog::hasAccess($request->user())
+            ->with('device')
+            ->when($request->device_group, function ($query) use ($request) {
+                $query->inDeviceGroup($request->device_group);
+            });
     }
 
     public function formatItem($eventlog)
--- a/app/Models/Device.php
+++ b/app/Models/Device.php
@@ -429,6 +429,15 @@ class Device extends BaseModel
         return $this->hasDeviceAccess($query, $user);
     }
 
+    public function scopeInDeviceGroup($query, $deviceGroup)
+    {
+        return $query->whereIn('device_id', function ($query) use ($deviceGroup) {
+            $query->select('device_id')
+                ->from('device_group_device')
+                ->where('device_group_id', $deviceGroup);
+        });
+    }
+
     // ---- Define Relationships ----
 
     public function alerts()
--- a/app/Models/DeviceRelatedModel.php
+++ b/app/Models/DeviceRelatedModel.php
@@ -36,8 +36,11 @@ class DeviceRelatedModel extends BaseModel
 
     public function scopeInDeviceGroup($query, $deviceGroup)
     {
-        $groups = $deviceGroup instanceof DeviceGroup ? $deviceGroup->devices()->pluck('devices.device_id') : [];
-        return $query->whereIn($this->getTable() . '.device_id', $groups);
+        return $query->whereIn('device_id', function ($query) use ($deviceGroup) {
+            $query->select('device_id')
+                ->from('device_group_device')
+                ->where('device_group_id', $deviceGroup);
+        });
     }
 
     // ---- Define Relationships ----
--- a/app/Models/Port.php
+++ b/app/Models/Port.php
@@ -6,7 +6,7 @@ use DB;
 use Illuminate\Database\Eloquent\Builder;
 use LibreNMS\Util\Rewrite;
 
-class Port extends BaseModel
+class Port extends DeviceRelatedModel
 {
     public $timestamps = false;
     protected $primaryKey = 'port_id';
@@ -210,11 +210,6 @@ class Port extends BaseModel
     }
 
     // ---- Define Relationships ----
-
-    public function device()
-    {
-        return $this->belongsTo('App\Models\Device', 'device_id', 'device_id');
-    }
 
     public function events()
     {
--- a/resources/views/widgets/settings/availability-map.blade.php
+++ b/resources/views/widgets/settings/availability-map.blade.php
@@ -56,9 +56,9 @@
 
     <div class="form-group">
         <label for="device_group-{{ $id }}" class="control-label">@lang('Device group')</label>
-        <select class="form-control" name="device_group" id="device_group-{{ $id }}"  data-placeholder="@lang('All Devices')">
+        <select class="form-control" name="device_group" id="device_group-{{ $id }}" data-placeholder="@lang('All Devices')">
             @if($device_group)
-                <option value="{{ $device_group->id }}" selected> {{ $device_group->name }} </option>
+                <option value="{{ $device_group->id }}" selected>{{ $device_group->name }}</option>
             @endif
         </select>
     </div>
@@ -66,7 +66,7 @@
 
 @section('javascript')
     <script type="text/javascript">
-        init_select2('#device_group-{{ $id }}', 'device-group', {}, '{{ $device_group ? $device_group->id : '' }}');
+        init_select2('#device_group-{{ $id }}', 'device-group', {});
 
         function toggle_availability_type(el, id) {
             console.log(el.value);
--- a/app/Http/Controllers/Table/SyslogController.php
+++ b/app/Http/Controllers/Table/SyslogController.php
@@ -34,6 +34,7 @@ class SyslogController extends TableController
     {
         return [
             'device' => 'nullable|int',
+            'device_group' => 'nullable|int',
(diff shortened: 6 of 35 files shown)


app/Http/Controllers/Table/EventlogController.php
```diff
@@ -61,7 +62,15 @@ protected function filterFields($request)
      */
     public function baseQuery($request)
     {
-        return Eventlog::hasAccess($request->user())->with('device');
+        $query = Eventlog::hasAccess($request->user())->with('device');
+
+        if ($request->device_group) {
+            $query->whereHas('device.groups', function ($query) use ($request) {
```
murrant: Should probably make this a query scope on the Device model (like hasAccess)

Jellyfrog: True, will fix.

murrant: That would also make it so you don't have to use all those clones below.  Those aren't very nice to look at :)

app/Http/Controllers/Table/EventlogController.php
```diff
@@ -62,13 +62,11 @@ protected function filterFields($request)
...
     public function baseQuery($request)
     {
-        $query = Eventlog::hasAccess($request->user())->with('device');
-
-        if ($request->device_group) {
-            $query->whereHas('device.groups', function ($query) use ($request) {
-                $query->where('id', $request->device_group);
-            });
-        }
+        $query = Eventlog::hasAccess($request->user())->with(['device' => function ($query) use ($request) {
```
Jellyfrog: @murrant Like this you had in mind?

murrant:
@Jellyfrog I've merged #9716 
You can use that now, feel free to update the device group query if needed.

murrant:
Hey @Jellyfrog are you planning on finishing this?

Jellyfrog:
> Hey @Jellyfrog are you planning on finishing this?

Hi, yes, forgot about it.

app/Models/DeviceRelatedModel.php
```diff
@@ -36,8 +36,9 @@ public function scopeHasAccess($query, User $user)
 
     public function scopeInDeviceGroup($query, $deviceGroup)
     {
-        $groups = $deviceGroup instanceof DeviceGroup ? $deviceGroup->devices()->pluck('devices.device_id') : [];
-        return $query->whereIn($this->getTable() . '.device_id', $groups);
+        return $query->whereHas('deviceGroup', function ($query) use ($deviceGroup) {
```
Jellyfrog: @murrant what do you say about doing like this instead?
Will generate the SQL:
```sql
select * from `eventlog`
where exists (
  select * from `device_group_device`
  where `eventlog`.`device_id` = `device_group_device`.`device_id`
  and `device_group_device`.`device_group_id` = ?
)
``` 
That should be more efficient :)

Jellyfrog:
@murrant 
What do you think about doing like this instead: https://github.com/librenms/librenms/pull/9692/commits/d01d9773d9b358fd28c84019dc9c0f13981d0faa

Will generate the SQL:
```sql
select * from `eventlog`
where exists (
  select * from `device_group_device`
  where `eventlog`.`device_id` = `device_group_device`.`device_id`
  and `device_group_device`.`device_group_id` = ?
)
``` 
That should be more efficient :)

app/Models/Port.php
```diff
@@ -209,13 +209,25 @@ public function scopeHasAccess($query, User $user)
         return $this->hasPortAccess($query, $user);
     }
 
+    public function scopeInDeviceGroup($query, $deviceGroup)
```
murrant: Port is already a DeviceRelatedModel, so you shouldn't need this to be redefined here.

app/Models/DeviceRelatedModel.php
```diff
@@ -46,4 +47,9 @@ public function device()
     {
         return $this->belongsTo('App\Models\Device', 'device_id', 'device_id');
     }
+
+    public function deviceGroup()
+    {
+        return $this->HasMany('App\Models\DeviceGroupDevice', 'device_id', 'device_id');
```
murrant: You should copy the groups() definition from the device model.  No need to create a model for the pivot table.

Then you just need to make everything consistent.

Jellyfrog: I know it's not needed, but it optimizes the SQL.
Current:
```sql
SELECT *
FROM `eventlog`
WHERE EXISTS
    (SELECT *
     FROM `device_group_device`
     WHERE `eventlog`.`device_id` = `device_group_device`.`device_id`
       AND `device_group_device`.`device_group_id` = '2')
ORDER BY `datetime` DESC
LIMIT 50
OFFSET 0;
```

With the following change:
```php
return $query->whereHas('groups', function ($query) use ($deviceGroup) {
    $query->where('device_groups.id', $deviceGroup);
});
```
The SQL becomes:
```sql
SELECT *
FROM `eventlog`
WHERE EXISTS
    (SELECT *
     FROM `device_groups`
     INNER JOIN `device_group_device` ON `device_groups`.`id` = `device_group_device`.`device_group_id`
     WHERE `eventlog`.`event_id` = `device_group_device`.`device_id`
       AND `device_groups`.`id` = '2')
ORDER BY `datetime` DESC
LIMIT 50
OFFSET 0
```
So this is simply a optimized version.

app/Models/Device.php
```diff
@@ -15,7 +15,7 @@
 use LibreNMS\Util\Url;
 use LibreNMS\Util\Time;
 
-class Device extends BaseModel
+class Device extends DeviceRelatedModel
```
Jellyfrog: @murrant This is up to you, a bit quirky, but it saves having to duplicate functions from `DeviceRelatedModel.php`

murrant: I think I prefer the duplicated function...  It doesn't contain any code so the duplication is minor.

app/Models/DeviceRelatedModel.php
murrant: This returns a DeviceGroupDevice model not a DeviceGroup model.

resources/views/widgets/settings/availability-map.blade.php
```diff
@@ -56,17 +56,17 @@
 
     <div class="form-group">
         <label for="device_group-{{ $id }}" class="control-label">@lang('Device group')</label>
-        <select class="form-control" name="device_group" id="device_group-{{ $id }}"  data-placeholder="@lang('All Devices')">
+        <select class="form-control" name="device_group" id="device_group-{{ $id }}" data-placeholder="@lang('All Device groups')">
```
murrant: This should still be "All Devices".

resources/views/widgets/settings/availability-map.blade.php
```diff
@@ -56,17 +56,17 @@
...
+                <option value="{{ $device_group->id }}" selected>{{ $device_group->name }}</option>
             @endif
         </select>
     </div>
 @endsection
 
 @section('javascript')
     <script type="text/javascript">
-        init_select2('#device_group-{{ $id }}', 'device-group', {}, '{{ $device_group ? $device_group->id : '' }}');
+        init_select2('#device_group-{{ $id }}', 'device-group', {});
```
murrant: When you remove this, is the current device group pre-selected?

Jellyfrog: Yes, and now it shows the name from the `<option>' instead of the id

murrant: Oh yeah, you have to pass an object there with name and id (then you can avoid the select in html).  I'm not sure which one is better...

app/Models/Device.php
```diff
@@ -476,6 +483,11 @@ public function groups()
         return $this->belongsToMany('App\Models\DeviceGroup', 'device_group_device', 'device_id', 'device_group_id');
     }
 
+    public function groupDevices()
```
murrant: I don't want this relationship or the DeviceGroupDevice model.

app/Models/DeviceGroupDevice.php
```diff
@@ -0,0 +1,10 @@
+<?php
```
murrant: Remove this.

app/Models/DeviceRelatedModel.php
```diff
@@ -46,4 +47,9 @@ public function device()
     {
         return $this->belongsTo('App\Models\Device', 'device_id', 'device_id');
     }
+
+    public function groupDevices()
```
murrant: remove

murrant:
You should either just check the table directly in your scope:
```
return $query->whereIn('device_id', function ($query) use ($deviceGroup) {
    /** @var Builder $query */
    $query->select('device_id')->whereIn('device_id', function ($query) use ($deviceGroup) {
    /** @var Builder $query */
    $query->select('device_id')
        ->from('device_group_device')
        ->where('device_group_id', $deviceGroup);
})
        ->from('device_group_device')
        ->where('device_group_id', $deviceGroup);
})
```

or change the basequery so instead of `Device::query()` (query is often ommited)

`DeviceGroup::find($deviceGroup)->devices()`

or a fake device group if you want to avoid the lookup
```
$dg = new DeviceGroup();
$dg->id = $deviceGroup;
$query = $dg->devices()->hasAccess...
```

murrant:

PR #9698, opened 2019-01-19
Improve index check for compatibility with different versions of the DDOS 

acl:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/discovery/storage/datadomain.inc.php
```diff
@@ -22,6 +22,9 @@
         $total = $storage['fileSystemSpaceSize'] * $units;
         $used = $storage['fileSystemSpaceUsed'] * $units;
         if ($descr == "/data: post-comp") {
+            if (is_null($index)) {
+                $index=$tmp_index;
```
murrant: $tmp_index is undefined...

murrant: Why use fileSystemResourceIndex at all?

acl: I was torn on this as well. Older DataDomain versions do have the fileSystemResourceIndex on their MIB file where as modern DataDomains have completely removed it. 

I don't know if it served any purpose at all in the past. I agree with your change and I'll keep testing to see if it's used in different setups. Thanks.

murrant: someone probably just didn't realize fileSystemResourceIndex  is in the table index.

Issue #9709, opened 2019-01-21
Request for update cron.d/librenms  in install docs section


mhzgh:
Hi there,

wanted to mention a small thing, which would improve docs and possible problems in auto-update feature.

When working trough the documentation for ubuntu https://docs.librenms.org/Installation/Installation-Ubuntu-1804-Apache/ i got tricked (once again) by cron shell invocation, when it won't source the environment variables. Yes this is described in the cron man-pages, but I think this not very broadly known.

So due to daily.sh triggering the git updates, it will rely on proxy settings and thus fail without proper ENV sourcing in many infrastructures.

IMO a small hint on that won't hurt, especially as the problem does not necessarily become obvious when you try to troubleshoot and re-execute as the librenms user manually (ENV is normaly loaded properly when you su into the librenms user directly)
I am referring to the "Cron job" chapter of the Ubuntu docs  in that case.

Edit: Maybe it would also be an option to mention git proxy settings instead, when referring to the dependencies required for installation.

Keep up the good work, has been a great experience using librenms so far!

Best Regards

PR #9720, opened 2019-01-23
Install docs, note proxy config

mhzgh:
As discussed here:
https://github.com/librenms/librenms/issues/9709

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Installation/Installation-CentOS-6-Apache-Nginx.md
+++ b/doc/Installation/Installation-CentOS-6-Apache-Nginx.md
@@ -324,6 +324,8 @@ Create the cronjob
 
     cp librenms.nonroot.cron /etc/cron.d/librenms
 
+> NOTE: Keep in mind  that cron, by default, only uses a very limited set of environment variables. You may need to configure proxy variables for the cron invocation. Alternatively adding the proxy settings in config.php is possible too. The config.php file will be created in the upcoming steps. Review the following URL after you finished librenms install steps: https://docs.librenms.org/Support/Configuration/#proxy-support
+
 ### Copy logrotate config ###
 
 LibreNMS keeps logs in `/opt/librenms/logs`. Over time these can become large and be rotated out.  To rotate out the old logs you can use the provided logrotate config file:
--- a/doc/Installation/Installation-CentOS-7-Apache.md
+++ b/doc/Installation/Installation-CentOS-7-Apache.md
@@ -154,6 +154,8 @@ Edit the text which says `RANDOMSTRINGGOESHERE` and set your own community strin
 
     cp /opt/librenms/librenms.nonroot.cron /etc/cron.d/librenms
 
+> NOTE: Keep in mind  that cron, by default, only uses a very limited set of environment variables. You may need to configure proxy variables for the cron invocation. Alternatively adding the proxy settings in config.php is possible too. The config.php file will be created in the upcoming steps. Review the following URL after you finished librenms install steps: https://docs.librenms.org/Support/Configuration/#proxy-support
+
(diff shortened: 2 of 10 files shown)


murrant:
How do you ensure that?

mhzgh:
I thought about it a moment but decided not to give a specifc example. In the most cases this should be possible by preceding the actual command with a source invocation in the cron job. E.g. . "~/.profile; daily.sh;" But this assumes the users personals files were "skel-created" during the user creation. This could be blocked, when the home directory already existed before the users was created. Or maybe there are other mechanism which inject http_proxy variable. Also settings could reside in profile.d folder, but are not sourced and so on... Was not too sure how to give an most-case-true example here... IMO once there is a hint to keep that in mind, users could review the place their settings come from and act accordingly - what do you think?

murrant:
for http_proxy, you can set it in config.php and I think it is exported for git commands, not sure though.  If it is not, it could be done.

murrant:
Yeah, I can't think of a silver bullet to solve the env in cron.  That is why I asked ;)

mhzgh:
> 
> 
> for http_proxy, you can set it in config.php and I think it is exported for git commands, not sure though. If it is not, it could be done.

That does sound even better to me, right. Daily.sh hands over to php anyways, and the whole application is php - so it doesnt need to be the cron-only solution.

Maybe a hint to review/set these variables during the Web-UI setup makes more sense in that case? Maybe even a lookup-check for validate.php?

Oh and https_proxy should be hitting here, to be precise - at least was in my case.

murrant:
A note about setting the proxy in config.php to ensure it works during cron or setting the env for cron would be good. You can link to https://docs.librenms.org/Support/Configuration/#proxy-support

That gives a concrete way forward, while also giving a hint for people that want to solve it outside config.php.

mhzgh:
Uhm, didn't even notice the Configuration Docs/Environment pages before to be honest...
Sure, can do.

> A note about setting the proxy in config.php to ensure it works during cron or setting the env for cron would be good. You can link to https://docs.librenms.org/Support/Configuration/#proxy-support
Could you give a bit more detail on what you mean by that? Not sure "where" to do modifications now :)

From our discussion I understand that adding a hint nearby the "cron copy" command is not the most optimal location. It should be mentioned in a more general scope nearby config.php.
Since the Install docs hand over that config.php creation part to the WebUI, I am not sure if the Chapter "Web Installer" would be a good place to add further config.php modification hints.

mhzgh:
So you mean e.g. adding an additional step/page in between the "Create user" and "Generate config.php" of the web UI installation wizard, that prompts to configure a http(s) proxy and mentions/refers to the things we discussed?

murrant:
@mhzgh don't worry about that, just thinking out loud.

Add a line return before your note and duplicate it to all the install docs and we can merge this.

mhzgh:
Oh, ok. Wouldn't feel too confident to extend the web-ui installer without further knowledge :)
Reworked the NOTE a bit to include the ENV or "config.php" references, will update it to the other OS descriptions too

murrant:
You can do it in the same PR.  If you want to do it from the github web interface, what I usually do is go to the changes tab, click the edit button (just an easy way to get to the right branch).  Then click on the path breadcrumb at the top to go to a different file and click edit on that file.

mhzgh:
Great, thanks for the guidance :)

Changed all the install docs I could find.

murrant:
Heh, we usually only update the current install docs... So you were a bit of an over-achiever there.

Thanks :)

PR #9722, opened 2019-01-23
Updated JumpCloud authentication example

dsmfool:
Updated JumpCloud authentication example.  Relevant Community Support Thread: https://community.librenms.org/t/ldap-with-jumpcloud/6883

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Authentication.md
+++ b/doc/Extensions/Authentication.md
@@ -182,8 +182,6 @@ An example config setup for use with Jumpcloud LDAP as a service is:
 
 ```php
 $config['auth_mechanism'] = "ldap";
-unset($config['auth_ldap_group']); #Remove built-in group
-unset($config['auth_ldap_groups']); #Remove built-in groups
 $config['auth_ldap_version'] = 3;
 $config['auth_ldap_server'] = "ldap.jumpcloud.com"; #Set to ldaps://ldap.jumpcloud.com to enable LDAPS
 $config['auth_ldap_port'] = 389; #Set to 636 if using LDAPS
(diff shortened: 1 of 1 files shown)


doc/Extensions/Authentication.md
```diff
@@ -182,17 +182,20 @@ An example config setup for use with Jumpcloud LDAP as a service is:
 
 ```php
 $config['auth_mechanism'] = "ldap";
+unset($config['auth_ldap_group']); #Remove built-in group
```
murrant: Instead of unset(), I'd say overwrite them like this:

```php
$config['auth_ldap_groups'] = [
    '{group}' = ['level' => 10];
];
```

doc/Extensions/Authentication.md
```diff
@@ -182,17 +182,20 @@ An example config setup for use with Jumpcloud LDAP as a service is:
...
+unset($config['auth_ldap_groups']); #Remove built-in groups
```
murrant: Instead of unset(), I'd say overwrite them like this:

```php
$config['auth_ldap_groups'] = [
    '{group}' = ['level' => 10];
];
```

doc/Extensions/Authentication.md
murrant: This can be omitted it is unneeded.  LibreNMS removes the default group if it is unchanged.

dsmfool: With this line removed the setting still remains in the "Global Settings" page as:
`auth_ldap_group | cn=groupname,ou=groups,dc=example,dc=com`

I would rather not see the default or placeholder setting.  Though, it seems like this parameter isn't required at all for successful authentication.

doc/Extensions/Authentication.md
dsmfool: This code block works, but I only have one group for my LibreNMS admins.  My suggested change is for a basic example.  If you don't unset the variable it just appends your desired group to the defaults.  Perhaps include this codeblock in an advanced example?

doc/Extensions/Authentication.md
murrant: @dsmfool yeah, the code discards any entries that are set to that.

Feel free to keep the unset in your local config.  I just don't think it should be in the docs ;)

doc/Extensions/Authentication.md
murrant:
```php
$config['auth_ldap_groups'] = [
    '{admin_group}' => ['level' => 10],
];
```

boom, only one group.  I added the second to illustrate how multiple groups would be added.  (rather than duplicating the entire entry like the current docs)

murrant: Honestly, it is probably best if we remove the default groups.  I can't imagine why they are set.

doc/Extensions/Authentication.md
dsmfool: I agree that it shouldn't have to be unset in the docs, but it should perhaps be shown how to remove them or overwrite them until the default groups are removed.

I'm not very familiar with Git, and unaware how to modify my commit for this pull request to reflect these requested changes to my pull request.

murrant: @dsmfool go to the files changed tab and click the edit button.  That will allow you to edit this file.

You can cancel and browse around to other files in that branch to edit them too if you want.

dsmfool: Done.  Hopefully I did it correctly.

PR #9727, opened 2019-01-24
Ruckus Wireless updates (ZD/SZ/Unleashed/Hotzone)

djamp42:
Changes:
1) Device Support for Ruckus Zone Director / Smart Zone / Unleashed / Hotzone
2) Added Host Info/CPU/MEM/client/ap count to all devices
3) Added per SSID client count to Zone Director and Smartzone Devices
4) Pull AP Licenses count for Zone Director and Smartzone

Notes:

1) Zone Director 1200 and Unleashed AP's share the same sysObjectID. (.1.3.6.1.4.1.25053.3.1.5.15). 
I tried my best to get the logic right in the discovery yaml files.
The issue is apon intial discovery it discovers the ZD1200 as a unleashed AP, 
and at somepoint after (5mins?) it discovers correctly as a ZoneDirector, and never has the issue again. Not sure if this is even a issue.

2) Ruckus ships a differnt RUCKUS-ROOT-MIB for each of the above products, so i had to get the differnces and mash them into 1 file.
The RUCKUS-ROOT-MIB here has been edited to support all the polling between devices, you will not find it from ruckus.

3) The OS Ruckuswireless is now only for Ruckus ZoneDirector, at some point the OS name for this should be updated, but i think that would require
a notice to be sent out to the user base. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


djamp42:
So it looks like some of the checks are failing because of this issue

> Zone Director 1200 and Unleashed AP's share the same sysObjectID. (.1.3.6.1.4.1.25053.3.1.5.15).
> I tried my best to get the logic right in the discovery yaml files.
> The issue is apon intial discovery it discovers the ZD1200 as a unleashed AP,
> and at somepoint after (5mins?) it discovers correctly as a ZoneDirector, and never has the issue again. Not sure if this is even a issue.

includes/definitions/ruckuswireless.yaml
```diff
@@ -14,5 +14,12 @@ register_mibs:
     ruckusZDWLANAPTable: RUCKUS-ZD-WLAN-MIB
     ruckusZDWLANAPRadioStatsTable: RUCKUS-ZD-WLAN-MIB
 discovery:
-    - sysObjectID:
-        - .1.3.6.1.4.1.25053.3.1
+    - sysObjectID: .1.3.6.1.4.1.25053.3.1.5
+      snmpget:
+          oid: '.1.3.6.1.4.1.25053.1.2.1.1.1.1.9.0'
```
murrant: This snmpget should probably be symetric with the exclusion above.

includes/definitions/ruckuswireless-unleashed.yaml
```diff
@@ -0,0 +1,26 @@
...
+text: 'Ruckus Wireless Unleashed'
+type: wireless
+icon: ruckus
+mib_dir:
+    - ruckus
+over:
+    - { graph: device_bits, text: Traffic }
+    - { graph: device_wireless_ap-count, text: 'Connected APs' }
+    - { graph: device_wireless_clients, text: 'Number of Clients' }
+register_mibs:
```
murrant: You can remove the register_mibs on the new OS.  It is only on the existing one for legacy reasons.

murrant:
```
NMP['/usr/bin/snmpget' '-v2c' '-c' 'ruckuswireless' '-Oqv' '-M' '/home/travis/build/librenms/librenms/mibs' '-t' '3' 'udp:127.1.6.2:1162' 'RUCKUS-ZD-SYSTEM-MIB::ruckusZDSystemModel.0']
MIB search path: /home/travis/build/librenms/librenms/mibs
Cannot find module (RUCKUS-ZD-SYSTEM-MIB): At line 1 in (none)
RUCKUS-ZD-SYSTEM-MIB::ruckusZDSystemModel.0: Unknown Object Identifier
```
This is the reason detection fails :)  It can't find RUCKUS-ZD-SYSTEM-MIB.  The problem is it hasn't added the mibdir to the search path because it is generic.  This could probably be fixed in the discovery code...

includes/definitions/discovery/ruckuswireless.yaml
```diff
@@ -5,4 +5,4 @@ modules:
               -
                   oid: ruckusZDSystemCPUUtil
                   num_oid: '.1.3.6.1.4.1.25053.1.2.1.1.1.5.58.0'
-                  type: ruckuszd
+                  type: ruckuswireless
```
murrant: Changing this will cause loss of history data.

djamp42:
Okay  i think i have all the above issues fixed, but now the existing ZD1100 snmpsim is failing because of the following.

(This is a Eltek OID so i'm not sure how it's getting in here. 
`[SNMP] snmpget ruckuswireless_zd1100 .1.3.6.1.4.1.12148.10.2.6.0 (1.3.6.1.4.1.12148.10.2.6.0): no data
`

(This is country code that was just added, i suppose i need to generate a updated zd1000 snmpsim.
`[SNMP] snmpget ruckuswireless_zd1100 .1.3.6.1.4.1.25053.1.2.1.1.1.1.9.0 (1.3.6.1.4.1.25053.1.2.1.1.1.1.9.0): no data
`

djamp42:
I actually found a zd1106 to test with, so i updated both the zd1100 and zd1200 test data.

djamp42:
Well i've fixed all the issues i know about, not sure if anyone else has tested this besides @utelisysadmin. I tested on my production librenms and everything seems fine as well.

utelisysadmin:
I've tested my PR in production too. I'll try to apply your today and see how it goes.

djamp42:
The only thing I noticed was when I manually hit rediscover on a zone director it started an entirely new graph for client count, so old data was lost.. however if I didnt hit that, librenms would eventually find all the changes and apply them, now my client graph has a single line from before and at the time of upgrade changes to per ssid lines.

utelisysadmin:
I've applied the PR. I'll let it work for a while, but so far it looks good for ZoneDirector,SmartZone and Unleasehed devices.

djamp42:
Does it ever match? I'm pulling SZ SNMP Client data directly and notice there is a lag between the SSIDs and the total count, it seems like the smartzone is not updating them both at the same time. It does eventually match up but changes again after clients disconnect and reconnect. This would get confusing to the user with just 1 SSID. Even though we are graphing exactly what ruckus is providing. I thought about adding code to limit the output to just the SSID if only 1 SSID is detected, i'll play around with that.

djamp42:
Ahh also found the AP's is the TOTAL aps, (down AND up).. So i'll need to update that sensor description also. This is the only AP count OID, so without polling APs directly like the WLC does we are stuck with this.

djamp42:
@utelisysadmin try again, will not display/poll client system total if only 1 SSID. Also updated Connected APs to Total APs.

utelisysadmin:
I've re-applied PR. I do understand why it shows You "Total APs". For me the number matches the currently "UP" number of APs in the SZ/ZD/Unleashed UI.

djamp42:
You dont have any down aps, the smartzone is reporting both up and down aps.

djamp42:
Basically your ap count will only change if you delete or add aps to the smartzone. It's the total aps in the system up and down.

utelisysadmin:
I cannot confirm this. I have a location with 114 APs, from which 48 are UP. The rest are down(confirmed by ZD dashboard). LibreNMS shows me 48 APs.

djamp42:
> I cannot confirm this. I have a location with 114 APs, from which 48 are UP. The rest are down(confirmed by ZD dashboard). LibreNMS shows me 48 APs.

This is only for Smartzone, i never checked ZD. Regardless i found a way to graph both in the smartzone and working on a fix.

utelisysadmin:
Yes, You are right. I just checked on a location with ZD and it show 22, even tough 1 is down.

djamp42:
Okay can you re-apply should graph total and connected APs. for smartzone. However you need to apply #9777 also because i needed to pass some SNMP flags that are not supported currently.

utelisysadmin:
I've applied both. Let's see how it goes.

utelisysadmin:
Just reporting I've had no issues so far.

murrant:
@djamp42 btw, you can run the tests locally and select just a specific test with `./scripts/pre-commit.php --snmpsim --db -o ruckuswireless-sz` for example

djamp42:
> @djamp42 btw, you can run the tests locally and select just a specific test with `./scripts/pre-commit.php --snmpsim --db -o ruckuswireless-sz` for example

Yeah i found that out a week ago, but i have somehow messed up my local git repo, not detecting new changes to files anymore, probably didn't help i'm learning git at the same time... In anycase everything is passing on my system, the final edits should allow everything to pass.

murrant:
@djamp42 You are doing great :)   I restarted the last build it failed for some random reason.

PR #7156, opened 2017-08-14
newdevice: Updated Arista entity-physical support to use high/low values from device

laf:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

Added checks for is_nan and is_infinite as the data returned from the calculations returned either -INF or NAN.

Fixes: #7078

---
This change is [Reviewable](https://reviewable.io/reviews/librenms/librenms/7156)

PR #9728, opened 2019-01-24
Fix Arista interface bias current divisor

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Divisor issue caused by #7156

This is incorrect on newer Arista versions(I guess the bug has been fixed in the OS).
Tested on EOS 4.20.8M & 4.21.3F

I think we should remove this workaround and advise users to upgrade, if asked.
Otherwise we are required to maintain a list of bugfree/working versions.


Rosiak:
I assume we're good on this? :)

PR #9729, opened 2019-01-24
Add support for Transition NIDs

JoshWeepie:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hello @JoshWeepie 
Thank you for your PR. Could you please add to it the MIB files from Transition Networks ? 
Thanx

JoshWeepie:
>           Could you please rename the files exactly as the mib name ? (without the extension mostly) and also remove the standard mib files (in the subdir) as we should already have them all ?

Removed extensions and standard mibs.

PipoCanaja:
Hello @JoshWeepie 
It seems that you removed too much now ... All mibs are gone. Rename got wrong ? You need to use "git mv nameOfFileNotGood.mib NAME-MIB"
PipoCanaja

JoshWeepie:
Ah oops, I'll get this fixed

murrant:
Sorry to be a bit late here.  But why have you included MIBs?  You don't seem to use them.

JoshWeepie:
I included them because @PipoCanaja asked that they be included in the PR, should we keep them or exclude them?😄  You're correct we are not currently using them and the polling/traffic graphs for our NIDs works correctly.

murrant:
I think in this case because there are so many.  We can just leave them off for now.  If you send a PR that adds sensors or something else that uses them, go ahead and include them there.

Also, you still need to add test data.

murrant:
@JoshWeepie Do you plan to add test data? (snmprec at minimum) Otherwise we will need to close this.

JoshWeepie:
I do plan on getting it added, we've been super busy and I had an issue trying to get the script to work correctly.

JoshWeepie:
@laf snmprec added

laf:
Thanks @JoshWeepie. Is their a better name for this, transition is the company name which we'd avoid using. Do they call the OS something? Model range that this works for?

JoshWeepie:
I don't think they call the OS anything, I am successfully polling S3290's and S3280's with the transition.yml file in this commit. Here's the datasheet for one of these https://www.transition.com/products/media-converters/s3290-series/

murrant:
@laf you have vlan interfaces ignored.  We should probably clear user settings when running tests right?

Issue #9712, opened 2019-01-21
New OS Request - fiberstore(fs.com)


Cormoran96:
**./discovery.php -h HOSTNAME -d | ./pbin.sh**
https://p.libren.ms/view/1d971906

**./poller.php -h HOSTNAME -r -f -d | ./pbin.sh**
https://p.libren.ms/view/d1231b65

**snmpbulkwalk -OUneb -v2c -c COMMUNITY HOSTNAME .  | ./pbin.sh**
https://p.libren.ms/view/67872cd3

vendor site:
https://www.fs.com

PipoCanaja:
Hi @Cormoran96 
Do you have the MIB files for these devices ?
PipoCanaja

Cormoran96:
No I don't find it.

PipoCanaja:
This Issue was a duplicate of #9117 so let's close it and continue the discussion in the other issue and in the PR #9734

Issue #9117, opened 2018-08-31
Support for new device Fiberstore Switches


sp1rr3:
**Please do not post bugs or ask questions**

**For help and support you can use our [Discord server](https://t.libren.ms/discord) or our [community site](https://community.librenms.org).**

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Please provide **ALL** info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [X] Please provide as much detail as possible.
- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

Hi, 
Please add support for Fiberstore switches this one is a S5850-48T4Q right now its added as a generic device and missing sensors.

discovery.php output - https://p.libren.ms/view/a4134bad
poller.php output - https://p.libren.ms/view/da1a55b4
snmpwalk output - Gets a crash from database with command provided in FAQ

laf:
We need the snmpwalk I'm afraid.

murrant:
Did you try snmpwalk instead of snmpbulkwalk?

sp1rr3:
Here comes the snmpwalk it seems as its the pbin that crashes.
[snmpwalk.txt](https://github.com/librenms/librenms/files/2359902/snmpwalk.txt)

murrant:
This is for a `APC Web/SNMP Management Card`

sp1rr3:
Damm sorry, same filename diffrent folder... here is the correct one.
[snmpwalk.txt](https://github.com/librenms/librenms/files/2361953/snmpwalk.txt)

PipoCanaja:
hi @sp1rr3 
Do you have the mib files for these devices ? 
Thanx
PipoCanaja

tsumaru720:
@PipoCanaja  Depends on the specific model
Vendor has MIBs for S2800 and S3800 available - Can provide a direct download URL from their site if you like?

PipoCanaja:
That would be a good beginning yes. I searched on their website with no success so far.

tsumaru720:
https://img-en.fs.com/file/user_manual/fs-s3800-mibs.zip

tsumaru720:
I think the device I have, whilst is FS.com - wont be detected with your current proposed definition - possibly needs a different OID adding to the FS.yaml? 

[...]

PR #9734, opened 2019-01-25
FS.COM (Fiberstore) 'GBN' and 'SWITCH' devices support

PipoCanaja:
Following https://github.com/librenms/librenms/issues/9712 and https://github.com/librenms/librenms/issues/9117

Mibs are quite buggy but some data is polled (Serial, HW, SW Version, CPU and Memory) on some devices. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


Cormoran96:
Works fine for me. Thanks

PipoCanaja:
@Cormoran96 
Which type of device ? Which sensors are discovered ? 
Thx

Cormoran96:
My bad.

I thought the processor graph start graph but no. 

screenshot from 2019-01-28 09-55-50

PipoCanaja:
So the 3900 family is different than the 3800 ... Their SNMP implementation is really bad ... 

If you can collect snmpwalk as described here, I'll check if something can be done : 
https://docs.librenms.org/Support/FAQ/#faq20

Cormoran96:
> If you can collect snmpwalk as described here, I'll check if something can be done 

I already do that here  #9712

html/images/os/fs.svg
```diff
@@ -0,0 +1,140 @@
+<?xml version="1.0" encoding="utf-8"?>
```
murrant: This needs to be optimized, I usually just paste it into https://jakearchibald.github.io/svgomg/

murrant:
Looks good overall.  I'm a but unsure about the OS name.
According to the MIBs, it should probably be called GBN.
The enterprise ID in the mib is GCOM Technologies, these are either rebranded GCOM or that is just a random enterprise ID:

```
--      Note that the admin enterprise ID (13464) is used in the sample
--      MIBs.  The customer "must" replace the admin enterprise ID with
--      the registered enterprise identifier for the specific company or
--      product.  The customer must also replace the ADMIN information
--      located in the 'MODULE-IDENTITY' in each sample MIB.
```

PipoCanaja:
In fact, it seems that they use multiple different mibs, from different origins ... One of them is indeed GBN according to the mib filenames. 
But as only very few OIDs are available, I suggest to stick with one OS (and handle the different OIDs in it) instead of creating as many OSes as devices from this vendor...

LibreNMS/OS/Fs.php
```diff
@@ -55,6 +55,8 @@ public function discoverProcessors()
         }
         // Tests OID from SWITCH MIB.
         $processors_data = snmpwalk_cache_oid($this->getDevice(), 'ssCpuIdle', [], 'SWITCH', 'fs');
+
+        $processors = [];
```
PipoCanaja: Not good. If you do so, you potentially forget the processors that were discovered in the previous foreach. 
New commit on its way.

PipoCanaja:
@murrant whenever you have time to review this, I would suggest we merge it. If new models of this vendor are to be integrated, we'll extend this later.
From my point of view, it looks OK.

includes/definitions/fs.yaml
```diff
@@ -0,0 +1,18 @@
...
+over:
+over:
+    - { graph: device_bits, text: 'Device Traffic' }
+    - { graph: device_processor, text: 'Processor Usage' }
+    - { graph: device_mempool, text: 'Memory Usage' }
+mib_dir:
+    - fs
+discovery:
+    - sysObjectID: 
+        - .1.3.6.1.4.1.51134
```
murrant: That is a lot of sysObjectIDs.  I'm concerned this OS isn't the right scope.  OS should be defined per OS/MIB set, not per vendor.

PipoCanaja: OK. I don't really understand the mess they did with so many sysObjectIDs. For the moment we have 2 mib groups (GBN* and SWITCH) so we will start with 2 OSes grouped in "fs".

murrant: I'm pretty sure all of the Fiberstore equipment is just rebranded oem equipment.

PipoCanaja: That would make sense ... Anyway, I'll still do that split ASAP.

PipoCanaja: Split done. Right now, 2 groups.

includes/discovery/mempools/fs-switch.inc.php
```diff
@@ -0,0 +1,24 @@
...
+// Discovery for FS-switch devices.
+//
+
+if ($device['os'] == 'fs-switch') {
+    echo 'FS : ';
+    $mempools_array = snmpwalk_cache_oid($device, 'memTotalReal', [], 'SWITCH', 'fs');
+    $mempools_array = snmpwalk_cache_oid($device, 'memTotalFree', $mempools_array, 'SWITCH', 'fs');
+    $mempools_array = snmpwalk_cache_oid($device, 'memTotalUsed', $mempools_array, 'SWITCH', 'fs');
+    d_echo($mempools_array);
+    if (is_array($mempools_array)) {
```
murrant: is_array is always true, you can just remove the if

murrant:
Looking pretty close.  Just one thing I commented on.  Also curious as to why you picked `fs-other`.

includes/discovery/mempools/fs-switch.inc.php
PipoCanaja: done

PipoCanaja:
I ended up with fs-gbn and fs-switch according to the mib they follow. You should not find any fs-other anymore in the current status of the PR.

PR #9711, opened 2019-01-21
Mattermost Alert Transport

gpant:
Added support for Mattermost notifications (Mattermost Transport).

George Pantazis <gpant@eservices-greece.com>

fixes #5659

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #9673, opened 2019-01-15
Added Alerta Alert Transport

GitStoph:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [ ] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

Apologies if i didn't do this right. Running out of time at work and wanted to get it up quick. 
Reference: https://github.com/librenms/librenms/issues/9672         @PipoCanaja

PR #9743, opened 2019-01-26
Revert "Mattermost Alert Transport"

TheGreatDoc:
Reverts librenms/librenms#9711

Due to that PR removing .htaccess files and making unable to load webui


TheGreatDoc:
@gpant if you dont mind doing a new PR but without removing the .htaccess files ;)

Also, the PR removes html/includes/collectd/config.php and resources/views/alerts/templates/default.blade.php 

Im pretty sure that shouldnt be there

gpant:
@TheGreatDoc I will re submit the PR using git directly, I didn't notice vscode did that in the second commit I am really sorry.

zombah:
@TheGreatDoc and why you reverted #9673? It seems it is different PR

zombah:
Yes revert commit is empty, didn't noticed, sorry for the noise

PR #9748, opened 2019-01-28
Files for Proper Discover of Tycon Power Systems TPDINs (all models)

jbdupont:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

These are changes and additional files needed to properly discovery the Tycon TPDIN, Tycon TPDIN2 and Tycon TPDIN-SC48-20 environmental/power monitoring devices used largely in the WISP (wireless internet service providers) community. The SNMP tables in the devices are extremely basic (i.e. all current, voltage and temperature readings are in a single table). This seemed to warrant the use of direct PHP sensory discovery modules. These are put into type "Power" since that is their primary purpose when used (that is, to monitor current, voltage and temperature).
Further, the TPDIN2 and TPDIN-SC48-20 use the same Enterprises branch and sub-branch, but the OIDs in use on each device, while similar and overlap, mean different things. This seemed to warrant a separate MIB folder under Tycon for each device.
Finally, the TPDIN unit (this is their original unit) uses the same Enterprises branch as the SERVERSCHECK.MIB and serverscheck.yaml discovery file. In order for the Tycon unit to be properly discovered, I had to rename serverscheck.yaml to something else (or delete it). Not sure how to indicate this, but I didn't want to include the file rename in the pull because I don't know what else it's used for - clearly it is there for some reason. The user will have to decide I guess.

Example screen shots are provided.
tycon-tpdinsc48-over
tycon-tpdin2-over
tycon-tpdin-over


murrant:
Tests are failing because your discoveries overlap.
We try to stick to sysObjectID for discovery if possible

jbdupont:
> Tests are failing because your discoveries overlap.
> We try to stick to sysObjectID for discovery if possible

Right. The struggle I had is that the SNMP implementation in the TPDIN2 and TPDIN-SC48-20 use the same Enterprises branch and the same OIDs underneath, but the two devices use the same OIDs differently. So when I left discovery using sysObjectID (which I tried), the TPDIN-SC48-20 would always only discover as a TPDIN2 and the sensors were wrong.

jbdupont:
Well.... let me know if you want me to do anything else with this. As I explained, there is conflict with the ServersCheck OS and the original TPDIN OS (they both use the same MIB enterprises ID (I can only surmise this because the TPDIN device was an OEM'd box from them and Tycon didn't apply their own enterprises ID until the TPDIN2). You probably have more creative ways to deal with this, but I renamed the serverscheck.yaml so it was at the end (alphabetically) of the file list so that the one I created would be matched up first.

The second conflict is that the TPDIN2 and the TPDIN-MPPT-SC48 use the same OID table (and have a different MIB), but use them for different things so discovery can't be done using sysObjectID alone.

The tpdin.yaml discovery file that is currently included in the master build doesn't work. A) it's only for the TPDIN2, not the original TPDIN; B) because the OID table it uses isn't broken into separate branches, the YAML definitions process only created the sensors for the 4th input of each sensor class (voltage, current, temperature). Unfortunately, there are four voltage inputs, four current inputs and two temp inputs on this device. C) because it uses sysObjectID, the TPDIN2 and TPDIN-MPPT-SC48 are being discovered as the same device, when they are not.

I'm not sure how else I can contribute to this effort. I can provide SSH access to my LibreNMS server and it has SNMP access to all three of Tycon's TPDIN models if you want to examine yourself. Otherwise, if what I've done isn't acceptable (I'm not a dev so it may not be), then I can delete the Pull request so it's not hanging out there.

murrant:
Sorry, I'm struggling to understand what needs to be done here.

As far as LibreNMS is concerned if a device uses the same MIB, it is the same device os, even if they are from different vendors.

If we just set all three to the same OS it wouldn't work right?  What differences would there be?

jbdupont:
Thanks for the message back, I appreciate it. Sorry to be so wordy, trying to get context and explanations right.

Everything I submitted in the Pull request was a result of discovery not working right for any of the three devices in question in the current master built.

1) Each device (TPDIN, TPDIN2, TPDIN-MPPT-SC48) uses a different MIB file.
2) TPDIN2 and TPDIN-MPPT-SC48 share the same sysObjectID, but need to be identified as two different devices because they use the same OIDs for different things; hence the use of sysDescr instead. It seemed necessary for them to be ID'd as different for LibreNMS to use different MIBs, which had to be in different folders.
3) TPDIN has the same enterprises ID as the serverscheck (ID 17095), but has its own MIB. In current master build, TPDIN device is discovered as a "ServersCheck" device, which does nothing to discovery the actual sensors. So using sysObjectID or sysDescr fails because the serverscheck.yaml is matched first. The MIB referenced in serverscheck.yaml is not correct for the TPDIN. The only way around this I found was to delete or move to the end (alphabetically) serverscheck.yaml. I don't know how else to work around this problem.
4) YAML based discovery of the sensors for all three devices is failing because the OID table structure in each is not "sane". That's why the use of PHP files for each sensor class for each device.
5) In current master build, only the last voltage, current and temperature inputs are "discovered" because of the poor OID table structure of the TPDIN2. There are actually four voltage and four current inputs along with two temp inputs, but only the last one of each are automatically discovered (refer to item 4 above).

PR #9753, opened 2019-01-28
Sorted ports, loadbalancer vips, pools by ifName and label

Serazio:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/ajax_listports.php
+++ b/html/ajax_listports.php
@@ -23,7 +23,10 @@ if (!LegacyAuth::check()) {
 }
 
 if (is_numeric($_GET['device_id'])) {
-    foreach (dbFetch('SELECT * FROM ports WHERE device_id = ? ORDER BY ifName', array($_GET['device_id'])) as $interface) {
+    $results = dbFetch('SELECT * FROM ports WHERE device_id = ?', array($_GET['device_id']));
+    $results = collect($results)->sortBy('ifName', SORT_NATURAL);
+
+    foreach ($results as $interface) {
         $interface  = cleanPort($interface);
         $string = addslashes(html_entity_decode($interface['label'].' - '.$interface['ifAlias']));
         echo "obj.options[obj.options.length] = new Option('".$string."','".$interface['port_id']."');\n";
--- a/html/ajax_listports.php
+++ b/html/ajax_listports.php
@@ -23,7 +23,7 @@ if (!LegacyAuth::check()) {
 }
 
 if (is_numeric($_GET['device_id'])) {
-    foreach (dbFetch('SELECT * FROM ports WHERE device_id = ?', array($_GET['device_id'])) as $interface) {
+    foreach (dbFetch('SELECT * FROM ports WHERE device_id = ? ORDER BY ifName', array($_GET['device_id'])) as $interface) {
         $interface  = cleanPort($interface);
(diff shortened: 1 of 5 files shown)


murrant:
The only thing is ordering gets wierd because ge1-10 comes before ge1-2.

Serazio:
I'll check later if "ORDER BY LENGTH(ifName) , ifName" will work better here.

PipoCanaja:
On some devices, sorting by ifIndex will probably provide the expected result, and the computing cost would be the best (sorting integers is always better than strings) but we'll not find a solution that fits all devices ...

Serazio:
I don't think that the performance aspect should matter here. 
I checked the query time with an juniper device with over 2900 interfaces and the times are not that higher. 

If this sorting would happen while polling, i totally agree with you, as this will generate a lot of overhead. But in this case it's just the overview and the bills

`#time mysql -hxxx -ulibrenms -p123 librenms -e "select * from ports where device_id = 1050 ORDER BY LENGTH(ifName), ifName;" > time.test

real	0m0.081s
user	0m0.020s
sys	0m0.013s

# time mysql -hxxx -ulibrenms -p123 librenms -e "select * from ports where device_id = 1050 ORDER BY ifName;" > time.test

real	0m0.085s
user	0m0.035s
sys	0m0.004s

# time mysql -hxxx -ulibrenms -p123 librenms -e "select * from ports where device_id = 1050" > time.test

real	0m0.067s
user	0m0.029s
sys	0m0.007s

#time mysql -hxxx -ulibrenms -p123 librenms -e "select COUNT(*) as count from ports where device_id = 1050"
+-------+
| count |
+-------+
|  2921 |
+-------+

real	0m0.014s
user	0m0.009s
sys	0m0.000s`

murrant:
@Serazio we weren't debating performance.

We were saying MySQL queries do alphabetical sort, humans expect natural sort (https://en.wikipedia.org/wiki/Natural_sort_order)

PipoCanaja:
@Serazio You could also try to reorder the result of the SQL query in PHP using natsort(). That may help getting a correct natural display.

Serazio:
I think i found a suitable way, to do this sorting. 
Basicly the following works for the overview:

```    $results = dbFetch("SELECT * FROM `ports` WHERE device_id = ? AND `deleted` != '1' AND `disabled` = 0", array($device['device_id']));
    usort($results, function($a, $b) {
        return strnatcasecmp($a['ifName'], $b['ifName']);
    });

    foreach ($results as $data) {
```

I'll update the pull request in the next hours

laf:
> On some devices, sorting by ifIndex will probably provide the expected result, and the computing cost would be the best (sorting integers is always better than strings) but we'll not find a solution that fits all devices ...

I can't help but agreeing with this but for a slightly different reason. Ordering by ifIndex (for ports overview) means you get them grouped on the type typically, ordering by name now means they ports are alphabetic and all over the show.

This is one of the things that a per user setting would be ideal.

murrant:
Yeah, I think ifIndex would be best here for device oriented port lists.

laf:
@Serazio Any update on this? You will also need to rebase.

Jellyfrog:

PR #9754, opened 2019-01-28
fix snmp debug clipping off quote when obscuring data

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/common.php
+++ b/includes/common.php
@@ -102,7 +102,7 @@ function external_exec($command)
         $replacements = [
             '-c\' \'COMMUNITY\'',
             '-u\' \'USER\'',
-            '-u\' \'USER\'',
+            '-U\' \'USER\'',
             '-A\' \'PASSWORD\'',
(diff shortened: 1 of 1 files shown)


includes/common.php
```diff
@@ -90,23 +90,23 @@ function external_exec($command)
...
-            '-c\' \'COMMUNITY',
-            '-u\' \'USER',
-            '-u\' \'USER',
-            '-A\' \'PASSWORD',
-            '-X\' \'PASSWORD',
-            '-P\' \'PASSWORD',
-            '-H\' \'HOSTNAME',
+            '-c\' \'COMMUNITY\'',
+            '-u\' \'USER\'',
+            '-u\' \'USER\'',
```
TheGreatDoc: Should be this -U instead of -u?

TheGreatDoc: Only one of the two

murrant: indeed, nice catch

PR #9757, opened 2019-01-29
LibreNMS/Alert/Transport/Elasticsearch - Send timestamp with timezone

pobradovic08:
Add timezone to timestamp field, otherwise ES will think it's UTC

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Elasticsearch.php
+++ b/LibreNMS/Alert/Transport/Elasticsearch.php
@@ -83,7 +83,7 @@ class Elasticsearch extends Transport
         }
 
         $data = array(
-            '@timestamp' => strftime("%Y-%m-%dT%T"),
+            '@timestamp' => strftime("%Y-%m-%dT%T%z"),
             "host" => gethostname(),
             "location" => $obj['location'],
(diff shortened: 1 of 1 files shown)


murrant:
Can we just send a unix_timestamp?
Aka: `'@timestamp' => date(),`

pobradovic08:
We could do:
`'@timestamp' => time()*1000,`

Apparently it's expecting microseconds since epoch and both  `time()` and `microtime(true)` are iterpreted as Jan 18th 1970 by ES

pobradovic08:
Unfortunately if we send just a timestamp then elastics sets the type of the field to `number` by default (instead of `date`). This can be fixed by creating a template on ES side but I think it's easier for everyone to send a timestamp as a string that will automatically be recognized as a date by ES.

murrant:
@pobradovic08 Perhaps instead of sending a timezone (which can bring daylight savings headaches) Just send the UTC time.  `\Carbon\Carbon:now('UTC')->toISOString()`

murrant:
Updated to use ISO 8601 format

PR #9759, opened 2019-01-29
Update Mattermost transport with configurable author_name

zombah:
Regarding Mattermost api doc  [author_name is optional](https://docs.mattermost.com/developer/message-attachments.html#author-details) 
And no reason to force hostname to author_name because hostname
already exist in alert topic, so make author_name configurable.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Mattermost.php
+++ b/LibreNMS/Alert/Transport/Mattermost.php
@@ -38,6 +38,7 @@ class Mattermost extends Transport
             'username' => $this->config['mattermost-username'],
             'icon' => $this->config['mattermost-icon'],
             'channel' => $this->config['mattermost-channel'],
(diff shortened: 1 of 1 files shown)


TheGreatDoc:
Hi @zombah 

Thanks for your PR. As for testing this a Mattermost server is needed, do you mind posting 2 screnshoots, one before and one after the patch?

zombah:
Current behaviour, triple hostname:
mattermost-with-patch-hostname-author
With patch, configurable author_name is empty:
mattermost-with-patch-empty-author

gpant:
The reason I had the author name set to the hostname is that it made it a lot easier to get the hostname when using the mattermost API to access the web hook stream. Now for doing the same exact work there is need for regex, having the author name blank or on a static value does not have any benefit.

PR #9764, opened 2019-01-31
Check for session_handler

Gorian:
Add a check to only display an error for the session file path being writable if `session_handler` is set to `file`

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Validations/Php.php
+++ b/LibreNMS/Validations/Php.php
@@ -59,7 +59,7 @@ class Php extends BaseValidation
     private function checkSessionDirWritable(Validator $validator)
     {
         $path = session_save_path() === '' ? '/tmp' : session_save_path();
-        if (!is_writable($path)) {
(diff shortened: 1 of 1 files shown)


Gorian:
Should probably write something to do checks if `file_handler` equals `memcache` or `redis`...

LibreNMS/Validations/Php.php
```diff
@@ -59,7 +59,7 @@ private function checkVersion(Validator$validator)
     private function checkSessionDirWritable(Validator $validator)
     {
         $path = session_save_path() === '' ? '/tmp' : session_save_path();
-        if (!is_writable($path)) {
+        if (!is_writable($path) and ini_get(save_handler) == "file") {
```
murrant: We generally use && instead of `and` and put quotes around save_handler, it is a string, not a constant.

murrant:
Closing we don't officially support other handlers.

PR #9767, opened 2019-01-31
Dantel WebMon Device Support

network-guy:
Support for Dantel WebMon units. Currently only polls temperature and humidity for the onboard sensors.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/webmon.yaml
```diff
@@ -0,0 +1,10 @@
+os: webmon
+text: 'Webmon Edge'
+type: environment
+icon: dantel
```
PipoCanaja: Hi,
You should include here a "mib: webmon" statement, and move your mib file to a webmon folder. doing this would allow better sorting and avoid unnecessary load of the webmon mib for other devices. 
The logic of mib loading is : 
- Load all files in mibs folder (as they are supposed to be "standard" mib files)
- Load all OS specific files from mibs/<folder>/ and <folder> is defined with the "mibs:" statement of the YAML. 

Bye

network-guy:
Requested changes have been made.

includes/discovery/sensors/temperature/webmon.inc.php
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2019 Mike Williams
+ * @author     Mike Williams <mike@mgww.net>
+ */
+
+$env = snmpwalk_group($device, 'pSlot1Table', 'WEBMON-EDGE-MATRIX-MIB');
+
+if (is_array($env)) {
```
murrant: snmpwalk_group always returns an array.  You should check if it is empty instead.

includes/polling/os/webmon.inc.php
```diff
@@ -0,0 +1,17 @@
...
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2019 Mike Williams
+ * @author     Mike Williams <mike@mgww.net>
+ */
+
+$version = trim(snmp_get($device, '.1.3.6.1.4.1.994.3.4.7.1.82.0.0', '-OQv', 'WEBMON-EDGE-MATRIX-MIB'), '"');
```
murrant: You don't need the trim, snmp_get already does that.

includes/discovery/sensors/humidity/webmon.inc.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2019 Mike Williams
+ * @author     Mike Williams <mike@mgww.net>
+ */
+
+$env = snmpwalk_group($device, 'pSlot1Table', 'WEBMON-EDGE-MATRIX-MIB');
+
+if (is_array($env)) {
```
murrant: This sensor looks like it could be implemented in yaml, if so, please move it there.

If you keep in in PHP: snmpwalk_group always returns an array. You should check if it is empty instead.

includes/discovery/sensors/temperature/webmon.inc.php
network-guy: Fixed

includes/polling/os/webmon.inc.php
network-guy: Fixed

includes/discovery/sensors/humidity/webmon.inc.php
network-guy: Fixed. The SNMP implementation is weird and I couldn't get it to pull the limits properly using YAML.

PipoCanaja:
Hi @network-guy 

Another LibreNMS user runs WebMon devices, and seems to have a different card in Slot 1. Your implementation suppose that temperature and humidity, which is not always the case. 
I'll propose a new implementation using "count" sensor (a neutral one) in order to try to match all type of cards in all slots.

PR #9779, opened 2019-02-05
Fixes constraint 'poller_cluster_node_id_unique'.

amanualgoldstein:
While the `poller_cluster_stats` table will support accounting for poller clusters where the number of participating poller instances N > 2, the `poller_cluster` table will always fold registrations from pollers into a single row for its parent cluster.

The `poller_cluster_stats` table may be populated as expected only once `poller_cluster` has a mapping for the corresponding `(node_id, poller_name)` pair.

Therefore, the underlying constraint `poller_cluster_node_id_unique` is redefined as unique with respect to both `poller_cluster.node_id` as well as `poller_cluster.poller_name`.

Additionally, relevant code paths in `Service.log_performance_stats` are updated to select `parent_poller_id` accordingly.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/service.py
+++ b/LibreNMS/service.py
@@ -708,7 +708,7 @@ class Service:
                            .format(self.config.node_id, self.config.name, "librenms-service", ','.join(str(g) for g in self.config.group), 1 if self.is_master else 0))
 
             # Find our ID
-            self._db.query('SELECT id INTO @parent_poller_id FROM poller_cluster WHERE node_id="{0}"; '.format(self.config.node_id))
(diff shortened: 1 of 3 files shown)


amanualgoldstein:
Amended previous constraint name to indicate it is now defined over two columns.

murrant:
@amanualgoldstein I don't understand what you are fixing here.  node_id should always be unique and poller_name can change (creating stale entries)

murrant:
I can't figure out what you are trying to fix.  Either your setup is wrong (node_id) should be unique or the fix is wrong.

Please explain it to me.  Closing for the time being.  Happy to re-open.

PR #9801, opened 2019-02-10
Workaround of bad SNMP implementation in EDS device.

PipoCanaja:
The EDS devices are sometimes replying to snmpget with an extra ".0" after the index. This patch checks this situation and extracts the correct value if available. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/sensors/state/eds.inc.php
+++ b/includes/polling/sensors/state/eds.inc.php
@@ -1,0 +1,6 @@
+<?php
+//Workaround for bad behaviour of the SNMP engine in EDS device
+//a ".0" is added in snmpget compared to snmpwalk of the same table. 
+if(!array_key_exists($sensor['sensor_oid'], $snmp_data) && array_key_exists($sensor['sensor_oid'].".0", $snmp_data)) {
+    $sensor_value = trim(str_replace('"', '', $snmp_data[$sensor['sensor_oid'].".0"]));
+}
--- a/includes/polling/sensors/state/eds.inc.php
+++ b/includes/polling/sensors/state/eds.inc.php
@@ -1,5 +1,5 @@
 <?php
-//Workaround for bad behaviour of the SNMP engine in EDS device
+//Workaround for bad behaviour of the SNMP engine in EDS device.
 //a ".0" is added in snmpget compared to snmpwalk of the same table. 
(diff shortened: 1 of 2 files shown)


murrant:
@PipoCanaja So, you need to omit {{$index }} in the yaml?  We can update the tests to allow that.

PipoCanaja:
> @PipoCanaja So, you need to omit {{$index }} in the yaml? We can update the tests to allow that.

Nope, not needed. 
You need to snmp[get|walk] the exact <OID>.<index> and you get a reply with <OID>.<index>.0 = "value" or you snmpwalk the table and you get the list of <OID>.<index> as expected ... 
So the solution implemented here and tested with the device is the only one I see so far. 
And it works with the test data (which are not buggy thanks to snmpsim) and the real device.

murrant:
@PipoCanaja but all this does is change the oid stored in the database then fetch it.  Why not just store the correct oid (without the .0)?

PipoCanaja:
In fact not: 
You need to query the correct OID+index for each sensor (as every the discovery finds it). But the reply you receive when queried with snmpget appears with an extra ".0" which makes the matching request/value fail. So this patch will check if the matching failed, and if yes, will try to match with an extra ".0". 
This ensures that a better implementation in a later upgrade of the device (or the test data in snmpsim) works as well as the real buggy device :)

PipoCanaja:
Exact. It is really stupid but at least we have a solution ...

PipoCanaja:
@murrant 
Do you have a way to force the unqueue of travis ? I did force a 2nd run, that succeeded but the original one is still stuck doing nothing ...

murrant:
@PipoCanaja you can restart within the Travis-CI web interface.  Do you have access to that?

PipoCanaja:
I added a dot in a comment to force travis-ci to run again cleanly. It was saying "queued" but when opening the job in Travis, it said "done" ... 
We'll see how it goes  this time.

murrant:
Ah, failure between github and travis.

PipoCanaja:
> Ah, failure between github and travis.

Now, it's good.

PipoCanaja:
> Probably shouldn't have added support in the first place and waited for the vendor to fix it. My bad :)

It seems that we'll have to wait a long time for the fix ... Anyway, it is not too ugly in the end :)

PR #9804, opened 2019-02-11
FDB table with history capabilities

PipoCanaja:
Hi,

Let's work at the implementation for  FDB table history.

Cleaning is done in daily.sh

PipoCanaja

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 9804`

### DIFF
--- a/app/Http/Controllers/Table/FdbTablesController.php
+++ b/app/Http/Controllers/Table/FdbTablesController.php
@@ -163,7 +163,8 @@ class FdbTablesController extends TableController
             'first_seen' => 'unknown',
             'last_seen' => 'unknown'
         ];
-
+        
+        // diffForHumans and doDateTimeString are not safe
         if ($fdb_entry->updated_at) {
             $item['last_seen'] = $fdb_entry->updated_at->diffForHumans();
         }
--- a/app/Http/Controllers/Table/FdbTablesController.php
+++ b/app/Http/Controllers/Table/FdbTablesController.php
@@ -160,9 +160,16 @@ class FdbTablesController extends TableController
             'vlan' => $fdb_entry->vlan ? $fdb_entry->vlan->vlan_vlan : '',
             'description' => '',
             'dnsname' => $ip_info['dns'],
-            'first_seen' => $fdb_entry->created_at->toDateTimeString(),
-            'last_seen' => $fdb_entry->updated_at->diffForHumans(),
+            'first_seen' => 'unknown',
+            'last_seen' => 'unknown'
         ];
+
+        if ($fdb_entry->updated_at) {
+            $item['last_seen'] = $fdb_entry->updated_at->diffForHumans();
+        }
+        if ($fdb_entry->created_at) {
+            $item['first_seen'] = $fdb_entry->created_at->toDateTimeString();
+        }
 
         if ($fdb_entry->port) {
             $item['interface'] = Url::portLink($fdb_entry->port, $fdb_entry->port->getShortLabel());
--- a/app/Http/Controllers/Table/FdbTablesController.php
+++ b/app/Http/Controllers/Table/FdbTablesController.php
@@ -45,7 +45,7 @@ class FdbTablesController extends TableController
         return [
             'port_id' => 'nullable|integer',
             'device_id' => 'nullable|integer',
-            'serachby' => 'in:mac,vlan,dnsname,ip,description',
+            'serachby' => 'in:mac,vlan,dnsname,ip,description,first_seen,last_seen',
             'dns' => 'nullable|in:true,false',
         ];
     }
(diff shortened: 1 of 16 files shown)


murrant:
The daily script matches a lot of other tables and saves mysql load during polling.

(Off topic, we should have a validation check for records older than cleanup values in the validation script)

database/migrations/2019_02_10_220000_add_dates_to_fdb.php
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('ports_fdb', function (Blueprint $table) {
+            $table->timestamp('date_discovered')->nullable();
```
murrant: You should use the standard Laravel "timestamps" https://laravel.com/docs/5.7/migrations

PipoCanaja: Correct, I ended up with this one instead of $table->timestamps(); so that I could change the names. But the DB structure is actually the same.
 
It is pretty simple to change now so let's go.

PipoCanaja:
OK, I'll have a look at how daily.sh does his magic. Never opened it so far.

PipoCanaja:
Ahaha,
A funny issue with Travis now. How do we deal with timestamps in a test ? Of course, they don't match : 

```
          1 => Array (
              'mac_address' => '001b382a7bc6'
 -            'date_discovered' => '2019-02-11 21:44:39'
 -            'date_last_seen' => '2019-02-11 21:44:39'
 +            'date_discovered' => '2019-02-11 21:24:40'
 +            'date_last_seen' => '2019-02-11 21:24:40'
              'ifIndex' => 9
              'vlan_vlan' => 1170
          )
```

murrant:
@PipoCanaja usually I exclude fields from that from being captured.  Kinda means we don't test them, but those fields don't really work in integration tests.

database/migrations/2019_02_10_220000_add_dates_to_fdb.php
murrant: You get niceties like Eloquent updating the timestamps automatically.

PipoCanaja:
> @PipoCanaja usually I exclude fields from that from being captured. Kinda means we don't test them, but those fields don't really work in integration tests.

@murrant : How do you exclude them ? just not collecting FDB data at all ?

database/migrations/2019_02_10_220000_add_dates_to_fdb.php
PipoCanaja: In fact, we need to use Eloquent "create" and "update" for the auto-update of the timestamps. But at least it is now ready for these niceties :)

PipoCanaja:
> @murrant : How do you exclude them ? just not collecting FDB data at all ?
Found it.

app/Http/Controllers/Table/FdbTablesController.php
```diff
@@ -152,6 +160,8 @@ public function formatItem($fdb_entry)
             'vlan' => $fdb_entry->vlan ? $fdb_entry->vlan->vlan_vlan : '',
             'description' => '',
             'dnsname' => $ip_info['dns'],
+            'first_seen' => strval($fdb_entry->created_at),
```
murrant: created_at should be an instance of carbon, so you can use that to get the format you want.

murrant:
@PipoCanaja I tested this and it works great.  The only thing I was wondering is about a way to show only current fdb entries.

app/Http/Controllers/Table/FdbTablesController.php
PipoCanaja: Done. Last seen is now in diffForHumans()

PipoCanaja:
> @PipoCanaja I tested this and it works great. The only thing I was wondering is about a way to show only current fdb entries.

@murrant : Good question. My idea now is to only display and allow sorting. That should provide both the current view and the history view depending on how you filter the data. Most probably users will let us know if they lack a way of displaying things.

daily.php
```diff
@@ -90,6 +90,11 @@
     }
 }
 
+if ($options['f'] === 'ports_fdb') {
+    $ret = lock_and_purge('ports_fdb', 'updated_at < DATE_SUB(NOW(), INTERVAL ? DAY)');
```
murrant: You need to set this default value and add documentation

PipoCanaja: default value is already set here : 
includes/defaults.inc.php

I'll do the documentation if you are OK with current implementation.

PipoCanaja: Doc was updated

PR #9806, opened 2019-02-12
Fix the divisor error found in routeros.yaml.

difanz:
RouterOS device returns the data in the correct form, thus there's no
need to use divisor.

Tested against RouterOS 6.43 and 6.42 (current long-term and stable branch), and can confirm the behavior change.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@difanz I think you've enabled the MIBs in your snmp.conf file (which parses the output and formats it).  The raw SNMP data can only return integers (not decimals) so the divisors are needed.  If you remove the MIB declaration in your snmp.conf, it will fix the values for you.

PR #9637, opened 2019-01-10
Added Alertmanager transport

angryp:
This is a new Transport pull request that offers integration with Alertmanager from Prometheus stack. 
Here is a depiction of how alerts look like from Alertmanager interface. Please note I have real hostnames crossed out here for security reasons. 
screenshot_1

Initially the code is based on Slack Transport, since it was easy to read for me. If there are some issues - please let me know and I will try to improve. 

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [ ] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #9807, opened 2019-02-12
Fixed Alertmanager transport

angryp:
This is a fix for [9637](https://github.com/librenms/librenms/pull/9637) pull request, which offered Alertmanager transport integration. What we did achieve is correct processing of resolved alerts - in my playground it was only a coincidence they were processed normally. After setting this up in a pre-prod environment, we saw alert resolutions are only processed after timeout in Alertmanager itself. 

This fix changes JSON layout according to this article: https://prometheus.io/docs/alerting/clients/

The only thing I remain uncertain with - if `date` function use is correct considering movement to Laravel. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Alertmanager.php
+++ b/LibreNMS/Alert/Transport/Alertmanager.php
@@ -41,16 +41,16 @@ class Alertmanager extends Transport
     public static function contactAlertmanager($obj, $api)
     {
         if ($obj['state'] == 0) {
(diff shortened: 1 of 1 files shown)


murrant:
The date function is fine. we have [Carbon](https://carbon.nesbot.com/docs/), but for simple things like this date() is better.

PR #9824, opened 2019-02-14
Fix logging too much

murrant:
Web servers are not running as the librenms user, so they are not capable of using daily rotation by Laravel yet...
Also mark all the files in config as non-editable

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
May be we could also add how to change in .env the level from debug to warning for instance.

murrant:
@PipoCanaja I got things a little mixed up in the Laravel PR.  The file logging SHOULD have been error and the stderr should have been debug (stderr is set when debug is enabled on the cli)

Should be correct now.

PR #9835, opened 2019-02-16
add support for sagemcom

vitalisator:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/Snmptrap/Handlers/EquipStatusTrap.php
```diff
@@ -0,0 +1,61 @@
...
+     * @param Device $device
+     * @param Trap $trap
+     * @return void
+     */
+    public function handle(Device $device, Trap $trap)
+    {
+        $severity = 0;
+        $state = $trap->getOidData('EQUIPMENT-MIB::equipStatus.0');
+
+        if ($state == 'warning' or $state == 'major' or $state == '5' or $state == '3') {
```
TheGreatDoc: Why number and string? They come randomly?

Also, could you change operators to be `||` instead `or`?

LibreNMS/Snmptrap/Handlers/EquipStatusTrap.php
```diff
@@ -0,0 +1,61 @@
...
+        } elseif ($state == 'critical' or $state == '4') {
+            $severity = 5;
+        } elseif ($state == 'minor' or $state == '2') {
+            $severity = 3;
+        } elseif ($state == 'nonAlarmed' or $state == '1') {
+            $severity = 1;
+        } else {
+            $severity = 0;
+        }
+        log_event('SNMP Trap: Equipment Status  ' . $state, $device->toArray(), 'state', $severity, $device->hostname);
```
TheGreatDoc: If this state trap is the same as polled in the discovery yaml, do you think you can do it to directly change the state in database too?

Like the linkUp/linkDown traps.

LibreNMS/Snmptrap/Handlers/LogTrap.php
```diff
@@ -0,0 +1,69 @@
...
+        $index = $trap->getOidData($index);
+
+        $logObject = $trap->getOidData('LOG-MIB::logObject.'.$index);
+        $logName = $trap->getOidData('LOG-MIB::logName.'.$index);
+        $logEvent = $trap->getOidData('LOG-MIB::logEvent.'.$index);
+        $logPC = $trap->getOidData('LOG-MIB::logPC.'.$index);
+        $logAI = $trap->getOidData('LOG-MIB::logAI.'.$index);
+        $state = $trap->getOidData('LOG-MIB::logEquipStatusV2.'.$index);
+
+        if ($state == 'warning' or $state == 'major' or $state == '5' or $state == '3') {
```
TheGreatDoc: Like the other file, change operators to be `||`

TheGreatDoc:
Thanks for your PR.

Awesome work.

LibreNMS/Snmptrap/Handlers/EquipStatusTrap.php
vitalisator: States numbers and strings just for sure, because this should work for different models and releases of ionos.

LibreNMS/Snmptrap/Handlers/EquipStatusTrap.php
vitalisator: not sure if this make sense. Would the direct change of the database trigger an immediately Alarm? Or would this only happen on next poller run?
Also I do not want to spend so much time on it :-)

html/images/os/sagemcom.svg
```diff
@@ -0,0 +1,560 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
murrant: This is more of a logo than an icon.  It also needs to be minimized.
https://docs.librenms.org/Developing/os/Initial-Detection/#icon-and-logo

includes/definitions/ionos.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - .1.3.6.1.4.1.1038
+poller_modules:
+            entity-physical: false
+            cisco-vrf-lite: false
+            storage: false
+            hr-mib: false
+            ucd-mib: false
+            arp-table: false
+            bgp-peers: false
+            vlans: false
```
murrant: Please don't disable all these modules like this. Some of these don't even exist others won't actually cause any additional load.  You may add ones that have a tangible benefit.

LibreNMS/Snmptrap/Handlers/LogTrap.php
```diff
@@ -0,0 +1,68 @@
...
+        $index = $trap->findOid('LOG-MIB::logIndex');
+        $index = $trap->getOidData($index);
+
+        $logName = $trap->getOidData('LOG-MIB::logName.'.$index);
+        $logEvent = $trap->getOidData('LOG-MIB::logEvent.'.$index);
+        $logPC = $trap->getOidData('LOG-MIB::logPC.'.$index);
+        $logAI = $trap->getOidData('LOG-MIB::logAI.'.$index);
+        $state = $trap->getOidData('LOG-MIB::logEquipStatusV2.'.$index);
+
+        if ($state == 'warning' || $state == 'major' || $state == '5' || $state == '3') {
```
murrant: You could actually do this with a single array
```php
$severity_map = [
    'warning' => 4,
    'major' => 4,
    '5' => 4,
    '3' => 4,
    'critical' => 5,
    ...
];

$severity = $severity[$state] ?? 0;
```

LibreNMS/Snmptrap/Handlers/EquipStatusTrap.php
```diff
@@ -0,0 +1,61 @@
...
+        } elseif ($state == 'critical' || $state == '4') {
+            $severity = 5;
+        } elseif ($state == 'minor' || $state == '2') {
+            $severity = 3;
+        } elseif ($state == 'nonAlarmed' || $state == '1') {
+            $severity = 1;
+        } else {
+            $severity = 0;
+        }
+        log_event('SNMP Trap: Equipment Status  ' . $state, $device->toArray(), 'state', $severity, $device->hostname);
```
murrant: The last parameter should not be hostname, just omit it.  The device is already referenced.

murrant:
Looks pretty good overall!  just a few small code style changes and the icon.

html/images/os/sagemcom.svg
vitalisator: changed

vitalisator:
please review/merge.

vitalisator:
@murrant , would be fine to go forward here :-)

PR #9837, opened 2019-02-18
Added resources/sensors api call to list all sensors

zombah:
Simple api call to get full list of all sensors

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/api_functions.inc.php
+++ b/html/includes/api_functions.inc.php
@@ -2029,31 +2029,12 @@ function get_link()
 
 function list_sensors()
 {
+    check_is_read()
+
     $app        = \Slim\Slim::getInstance();
     $router     = $app->router()->getCurrentRoute()->getParams();
-    $sql        = '';
-    $sql_params = array();
-    $hostname   = $router['hostname'];
-    $device_id  = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
-    if (is_numeric($device_id)) {
-        check_device_permission($device_id);
-        $sql        = " AND `sensors`.`device_id`=?";
-        $sql_params = array($device_id);
-    }
-    if (!LegacyAuth::user()->hasGlobalRead()) {
-        $sql .= " AND `sensors`.`device_id` IN (SELECT device_id FROM devices_perms WHERE user_id = ?)";
-        $sql_params[] = LegacyAuth::id();
-    }
 
-    $sensors       = array();
-    foreach (dbFetchRows("SELECT `sensors`.* FROM `sensors` LEFT JOIN `devices` ON `sensors`.`device_id` = `devices`.`device_id` WHERE `sensors`.`sensor_id` IS NOT NULL $sql", $sql_params) as $sensor) {
-        $host_id = get_vm_parent_id($device);
-        $device['ip'] = inet6_ntop($device['ip']);
-        if (is_numeric($host_id)) {
-            $device['parent_id'] = $host_id;
-        }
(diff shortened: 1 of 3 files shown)


html/includes/api_functions.inc.php
```diff
@@ -2027,6 +2027,42 @@ function get_link()
...
+        $sql_params = array($device_id);
+    }
+    if (!LegacyAuth::user()->hasGlobalRead()) {
+        $sql .= " AND `sensors`.`device_id` IN (SELECT device_id FROM devices_perms WHERE user_id = ?)";
+        $sql_params[] = LegacyAuth::id();
+    }
+
+    $sensors       = array();
+    foreach (dbFetchRows("SELECT `sensors`.* FROM `sensors` LEFT JOIN `devices` ON `sensors`.`device_id` = `devices`.`device_id` WHERE `sensors`.`sensor_id` IS NOT NULL $sql", $sql_params) as $sensor) {
+        $host_id = get_vm_parent_id($device);
```
murrant: $device is not defined yet...

zombah: Yes, missed that. Now i see that i wrongly duplicated this construct, it looks devices table specific with it's ip and parent_id stuff. Maybe to just use check_is_read for such list functions?

html/includes/api_functions.inc.php
```diff
@@ -2029,31 +2029,12 @@ function get_link()
...
-    $sensors       = array();
-    foreach (dbFetchRows("SELECT `sensors`.* FROM `sensors` LEFT JOIN `devices` ON `sensors`.`device_id` = `devices`.`device_id` WHERE `sensors`.`sensor_id` IS NOT NULL $sql", $sql_params) as $sensor) {
-        $host_id = get_vm_parent_id($device);
-        $device['ip'] = inet6_ntop($device['ip']);
-        if (is_numeric($host_id)) {
-            $device['parent_id'] = $host_id;
-        }
-        $sensors[] = $sensor;
-    }
+    $sensors       = dbFetchRows("SELECT `sensors`.* FROM `sensors` WHERE `sensors`.`sensor_id` IS NOT NULL);
```
murrant: sensor_id cannot be null, fyi

murrant: You could also use Eloquent here:

```php
$sensors = \App\Models\Sensor::hasAccess(Auth::user())->get();
$total_sensors = $sensors->count();
```

zombah: @murrant MySQL schema says that NULL is default for sensor_id:
```
MariaDB [librenms]> describe sensors;
+---------------------------+------------------+------+-----+-------------------+-----------------------------+
| Field                     | Type             | Null | Key | Default           | Extra                       |
+---------------------------+------------------+------+-----+-------------------+-----------------------------+
| sensor_id                 | int(10) unsigned | NO   | PRI | NULL              | auto_increment              |
```
i think it's good to have such sanity check same way all other api calls have
or better not?

zombah: @murrant Count is not working for some reason with Eloquent

murrant: @zombah it is auto_increment and also not nullable.

Count works for me:
```
php artisan tinker
Psy Shell v0.9.9 (PHP 7.3.3 — cli) by Justin Hileman
>>> $sensors = \App\Models\Sensor::hasAccess(User::first())->get();
>>> $sensors->count();
=> 13
```

PR #9838, opened 2019-02-18
Added power and fan sensors to VRP

jozefrebjak:
Added

image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/discovery/vrp.yaml
```diff
@@ -119,4 +119,35 @@ modules:
                     index: '{{ $index }}'
                     skip_values: -1
                     divisor: 1000000
-
+        load:
+            data:
```
PipoCanaja: You probably want "fan" and not "load" here :)

PipoCanaja:
Hi @jozefrebjak 

Could you create a new variant instead of changing the existing one for tests ? Your device is not the same as the existing one in tests. 

Thx

includes/definitions/discovery/vrp.yaml
jozefrebjak: there is value in percentage instead of rpm, so we cant use fan category.

jozefrebjak:
> Hi @jozefrebjak
> 
> Could you create a new variant instead of changing the existing one for tests ? Your device is not the same as the existing one in tests.
> 
> Thx

OK I will make another test data

includes/definitions/discovery/vrp.yaml
jozefrebjak: image

PipoCanaja: If you have the full speed in an OID, you can do the math in yaml and convert to the proper unit (RPM). If you don't, then you should probably use count and a group to explain what it is.

PipoCanaja:
You should revert the old snmprec but recompute (save-test-data) the json in case necessary.

includes/definitions/discovery/vrp.yaml
jozefrebjak: I found some of the maximum speed of the fan, but there is more than one module of fan, with different speeds, and i cant find an oid with that information for now. For VRP this will be the very first load sensor, and look at the picture in description i believe that its clear about what is going on. I can add group, this is not a problem.

PipoCanaja: If it is not exactly what the load class is for, then we must use "count" which is the only generic class we have. Then groups and name of the sensor will explain the user what it is.

murrant: Can't use count here.  You could add a new sensor for this, otherwise, drop it.

jozefrebjak: @murrant  i dont know how to add new sensor, so i dropped load of the fan.

PR #9839, opened 2019-02-18
Syslog hostname translation broken

TylerSweet:
For some reason, this is now broken. Accessing the array this way fixes it. I suspect it has something to do with the periods in the host names

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Good catch, would cause an error if the index is not set.  But error check is disabled in legacy code.

PR #9843, opened 2019-02-18
Add support for Infinera-Coriant Groove

nickhilliard:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hello
Thanx for your PR. Here are a few questions:
- Why did you code most of the sensors instead of using YAML ? 
- Could you improve the test data, the snmprec is quite empty for the moment, and the json is not there at all: https://docs.librenms.org/Developing/os/Test-Units/

Bye
PipoCanaja

nickhilliard:
> Why did you code most of the sensors instead of using YAML ?

Couple of reasons:  the interface name is not presented in snmp - this needs to be reconstructed in code.  I tried using yaml, and can get the information off the box, but can't create meaningful labels for the UI.  Is there a way of using $pre_cache[] array data in yaml?  If so, this might solve that problem.

Also, these devices don't have any of the usual IF-MIB contents.  All these interface instances need to be reconstructed from several other entity tables in the coriant mib.

> Could you improve the test data, the snmprec is quite empty for the moment, and the json is not there at all: https://docs.librenms.org/Developing/os/Test-Units/

ok, will take a look at that.

PipoCanaja:
You can actually precache data in yaml file and use other fields to build descriptions. You can look at `includes/definitions/discovery/vrp.yaml` which precaches entPhysicalName to group each SFP rx and tx power, for instance. 
It is clearly not as powerful as coding, so if needed, you are perfectly correct to code instead of Yaml.

nickhilliard:
i looked at vrp.yaml for some ideas, but it wasn't as flexible as needed.  There was an additional complication because the network interface indexes were non numeric, which meant that this needed to be simulated in code.

If the yaml processing code is looked at in future, it would be great to see coded pre-cache data available in the yaml configs - if this happened, a good chunk of this code could be replaced by yaml.

otherwise, I've updated the test data and this is running clean.  I also did a rename + logo change because Coriant got bought out by Infinera.  i.e. no further substantial changes planned on this side for this pull request.

TheGreatDoc:
@nickhilliard There is already pre-cache for yaml if that is what you mean

PipoCanaja:
Looks pretty good then. @TheGreatDoc, what do you think ?

includes/polling/os/infinera-groove.inc.php
```diff
@@ -0,0 +1,35 @@
...
+ * @author     Nick Hilliard <nick@foobar.org>
+ */
+
+echo 'CORIANT-GROOVE-MIB: ';
+
+// CORIANT-GROOVE-MIB::neType.0
+$hardware = snmp_get($device, '.1.3.6.1.4.1.42229.1.2.2.1.3.0', '-OQv', '+CORIANT-GROOVE-MIB', 'infinera-groove');
+
+// CORIANT-GROOVE-MIB::softwareloadSwloadVersion.1
+$version  = snmp_get($device, '.1.3.6.1.4.1.42229.1.2.9.2.1.1.3.1', '-OQv', '+CORIANT-GROOVE-MIB', 'infinera-groove');
```
TheGreatDoc: Can you move this to a snmp_get_multi?

nickhilliard:
ok, moved to snmp_get_multi().

PR #9849, opened 2019-02-20
Fix graphs for ASA announcing all interfaces in type l2vlan (by default filtered)

PipoCanaja:
Cisco ASA announces all its interfaces in l2vlan type (at least in multi-context mode). Or by default these are ignored in graphs (which is good generally speaking). Which gives an empty errored device_bits overall graph for each ASA contexts. 
 
In order to avoid this, this PR add a condition to ignore the l2vlan filtering for ASA.

The other option would be to create a filter list "per os" but it seems a little bit overkill only for this ASA issue. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/graphs/device/bits.inc.php
+++ b/html/includes/graphs/device/bits.inc.php
@@ -8,6 +8,12 @@ foreach (dbFetchRows('SELECT * FROM `ports` WHERE `device_id` = ? AND `disabled`
     $ignore = 0;
     if (is_array($config['device_traffic_iftype'])) {
         foreach ($config['device_traffic_iftype'] as $iftype) {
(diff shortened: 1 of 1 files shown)


murrant:
Why filter it only in the graphs?  Are the interfaces useful elsewhere?  You could just filter the ports completely by updating the OS yaml.

PipoCanaja:
In fact, I want to "unfilter" them in the graphs. 
Polling works beautifully but because of the filtering, they are not included in the aggregation graph on device page, graph which ends up beeing "Error drawing graph" (because all RRDs are filtered, so nothing remains). 
I don't think we can do anything in the YAML (cause this part works perfectly).

PR #9860, opened 2019-02-22
Alertmanager fix 2

angryp:
This is once again a small fix for Alertmanager transport. We found ourselves in a situation where we need to pass two label/value pairs from LibreNMS to Alertmanager, so rewrote the Transport to offer multiple labels to be specified. This has also been documented.
Also due to [latest release](https://github.com/prometheus/alertmanager/releases/tag/v0.16.0) Alertmanager V1 HTTP API endpoint has been deprecated, so it had to be changed as well. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Alerting/Transports.md
+++ b/doc/Alerting/Transports.md
@@ -36,7 +36,7 @@ It has built-in functionality for deduplicating, grouping and routing alerts bas
 
 LibreNMS uses alert grouping by alert rule, which can produce an array of alerts of similar content for an array of hosts, whereas Alertmanager can group them by alert meta, ideally producing one single notice in case an issue occurs. 
 
-The one and only possible parameter to be passed is `source` - this is required to distinguish LibreNMS alerts from alerts coming from different sources. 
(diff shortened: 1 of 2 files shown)


doc/Alerting/Transports.md
```diff
@@ -46,6 +46,7 @@ The one and only possible parameter to be passed is `source` - this is required
 | ------ | ------- |
 | Alertmanager URL      | http://alertmanager.example.com |
 | Alertmanager Options: | source=librenms |
```
TheGreatDoc: As you removed `source` in Alertmanager.php, do you still need to set it up?

angryp: Depends on the setup a user has, really. Initially it appeared logical to me (!) that one would like to specify source as one of the labels. Then I stumbled upon a requirement to add two labels which would not be usable for majority of users (mine will be source and escalation number), so ended up offering user to specify whatever he wants. Code block at line 68 does it. 
So in end, if it is "source" user wants - he can use "source". If it is something else - it can be something else.

angryp:
Just FYI: we had launched this script in our production this week and it works successfully so far.

PR #9864, opened 2019-02-22
remove calls to unavailable {{ $variable }} from YAML discovery

PipoCanaja:
As discussed in Discord, this PR removed unused variables from YAML Files instead of leaving them untouched.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Need to review all the test files now ...

murrant:
Having second thoughts on this one.  Leaving the original code makes it more obvious that the data was not found.

Hopefully, only the developer sees unreplaced data.

Thoughts?

PipoCanaja:
> Having second thoughts on this one. Leaving the original code makes it more obvious that the data was not found.
> 
> Hopefully, only the developer sees unreplaced data.
> Thoughts?

My issue is that sometimes, due to the fact that YAML does not allow conditions, you might want to display a variable if it is available, and nothing if it is not. In that case, it makes sense to have the "remove if not available" condition. 

Exemple : VRP, the OS of Huawei, is theoretically the same on all devices of the vendor (switches, routers, wireless Lan controllers). But they are in fact different flavours, and the SNMP implementation is very unstable between flavours. This current patch would allow to use 2 or 3 variables in one description, considering that only one is available for each flavours of VRP. 

That beeing said, I agree that we should display a warning in debug whenever a variable is not replaced, so that any person running ./discovery.php -v -d would see it (in case it is a mistake). I'll add that.

murrant:
Sounds good to me.

PR #9871, opened 2019-02-24
newdevice: Added support for Alpha Comp@s

cppmonkey:
Added support for the Alpha Comp@s AC/DC monitoring system.

Unfortunately the only way to see if it has switched to battery, is to see the battery discharging (negative current draw).

Our unit doesn't report input voltage...

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hi @cppmonkey 
Seems that your test data is not matching your current code. Is it possible that you did not run ./scripts/save-test-data.php with your last version of code ? 
You can check the results in TravicCI and track the issue down. 
Bye
PipoCanaja

cppmonkey:
I think I've found the issue.

create_state_index($state_name, $states) doesn't update the database. The states had changed in the includes, but hadn't updated the database.
Is it possible to purge a particular state_name from the database?

PR #9877, opened 2019-02-25
OpenGear OS remove single quote around type OOB Management

esundberg:
This is just something i noticed while looking up all the device types.

1. I remove the Single quotes around Device type OOB Management
2. Is OOB Management a correct device type? Should this be something else?

### DIFF
--- a/includes/definitions/opengear.yaml
+++ b/includes/definitions/opengear.yaml
@@ -1,6 +1,6 @@
 os: opengear
 text: Opengear
-type: 'OOB Management'
+type: OOB Management
 icon: opengear
 mib_dir:
     - opengear


PipoCanaja:
Hi,
- The quote itself seems perfectly fine. Why would you remove it ? 
- It seems that OOB Management is indeed not an official type, so that should probably be corrected. Would Appliance make sense ? 
PipoCanaja

esundberg:
what does it take to add more device types? May be we should add  "Console Server" for the Open Gear and Perle Devices.

PipoCanaja:
@librenms/reviewers : Any advices on adding new types? Should we avoid having too much of them ?

murrant:
Yeah, I think we should.  I don't see much reason to make a change here.

esundberg:
@librenms/reviewers Since OpenGearOS type is currently set to 'OOB Management'. To conform with the standard device types should change this to Network, Appliance, or Server? 


librenms@librenms-devel:~/includes/definitions$ fgrep type * | cut -d ':' -f 3 | sort | uniq 
grep: discovery: Is a directory

 appliance
 camera
 collaboration
 environment
 environmental
 firewall
 loadbalancer
 management
 network
 OOB Management
 power
 printer
 proxy
 server
 storage
 timing
 wireless
 workstation

librenms@librenms-devel:~/includes/definitions$ fgrep OOB *
opengear.yaml:type: OOB Management

PipoCanaja:
For the CXR remote console, I used "network". 
"appliance" could also match. It seems that remote console are mostly used to do OOB management for network equipements.

murrant:
you could just change it to 'management'

murrant:
Closing this for now, feel free to resubmit

PR #9879, opened 2019-02-26
Add Junos VirtuallChassis ports polling

rmagomedov:
Junos EX/QFX switches can be stacked together by so called VC ports. It's good to be able to monitor utilization of all these VC ports. With IF-MIB it is only possible to see ports of VirtuallChassis Master switch which is not enough. There is JUNOS-VIRTUALlCHASSIS-MIB containing interface counters of all VC ports for system.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


rmagomedov:
Is it ok if I add required snmp data to existing test file or it's better to create a new one? If it's fine, I updated tests/snmpsim/junos_ex.snmprec file. I cannot rebuild junos_ex.json with 'scripts/save-test.data.php -o junos -v ex'. When I run it, it stucks on 'Listening at UDP/IPv4 endpoint 127.1.6.1:1161, transport ID 1.3.6.1.6.1.1.0' on nothing happens after.

PipoCanaja:
Hi @rmagomedov ,
You should disable your firewall on 127.0.0.0/8 (all the loopback interfaces) and it will succeed.

rmagomedov:
Test data is updated.

PipoCanaja:
Hi,
It seems that the JSON file is not built against the SNMPREC file that is attached to the Pull request. If you look at the Travis CI results, you can see there is an issue. 
Could you first try to do : ```
./scripts/pre-commit.php -o junos ```
That one should run tests for junos (all variants, including 'ex') and you can fine tune the test in order to see a success here. 
Then you can commit and push again the correct files to the PR. 
PipoCanaja

PipoCanaja:
for some reasons, the snmprec file in this PR does not create any ports, and your JSON file does have ports. So either the snmprec file is not the same in your environnement than here, or your php code is not the same.

rmagomedov:
I checked out my working directory to the add-junos-vc-ports branch, the branch of the pull request. And all steps I do, I do it from this directory. The junos_ex.snmprec is updated there and I generate junos_ex.json in with `./scripts/save-test-data.php -o junos -v ex` command and it goes well - the data file is updated and it contains all expected VC ports. Right after that, from the same directory, I run  unit tests `./scripts/pre-commit.php --db --snmpsim -p -u` and it fails. So generating the .json file works well(expected ports are in place), but at the same time the unit tests show empty port array and fail

rmagomedov:
Here is how part of debug of `./scripts/save-test-data.php -o junos -v ex -m ports --debug` looks like. 

[save-test-data-debug.txt](https://github.com/librenms/librenms/files/2923890/save-test-data-debug.txt)

It looks like the unit test script is not even trying to poll jnxVirtualChassisPortTable of JUNIPER-VIRTULCHASSIS-MIB.

PipoCanaja:
The save-test-data.php should always be run without the -m <module> option, cause that's how the tests are run in the automated tests ... 

If that does not help, as junos_ex.snmprec was existing before your change, here is a suggestion : 

- Could you try to restore the original  junos_ex.snmprec and  junos_ex.json 
- Collect a new snmprec (junos_exPortPoll.snmprec for instance)
- Run a save-test-data without specifying any module (so you ensure all modules are run). 
- Then try the pre-commit.php -o junos that should succeed.

If all is good that way, then the issue is the mix of old snmprec data and new, and you can keep the new test file and leave the existing one untouched.

rmagomedov:
Sorry for the delay. I tried all of the above, nothing helped. I collected new separate junos_qfx.snmprec with `./scripts/collect-snmp-data.php -h <IP> -o junos -v qfx` command. Like before, save-test-data generates .json with expected VC ports, but pre-commit fails.

I am not sure but it looks like 'pre-commit' even is not trying to run `includes/polling/ports/junos-vc-ports.inc.php` needed for polling of VC ports while 'save-test-data' is.

murrant:
Please add your test files here and we can check it out.

rmagomedov:
Like I said, yesterday I found out that if I include my VC port polling php code with 'require' instead of 'require_once' tests pass.

Today I also noticed that tests pass even if the code include with 'requite_once' when test unit do not use OS variants. For example:
-  Moved JUNIPER-VIRTUALCHASSIS-MIB data from junos_ex.snmprec to junos.snmprec
-  Renamed junos_qfx.snmprec I collected myself to qfx.snmprec

rmagomedov:
In order to complete this pull request I can add the MIB data to junos.snmprec instead of junos_ex.snmprec, I think it's ok but a little illogical because looks like this junos.snmprec was collected from Juniper SRX firewall which I believe cannot have VirtuallChassis ports, but probably it does not matter for the unit tests.

I can also add that before I ran into this problem with unit tests. The test passed fine on my other production server on the php56 brach - Debian 9, PHP7.0

rmagomedov:
I have moved the snmp data from junos_ex.snmprec to junos.snmprec and now the tests passed

murrant:
Yes, it should be require (because it is being used like a function).
Looks like you got the unit tests working, nice.  Just one style fix and this should be g2g.

rmagomedov:
The tests passed

includes/polling/ports/junos-vcp.inc.php
```diff
@@ -0,0 +1,70 @@
...
+
+foreach ($junos_vcp_stats as $index => $vcp_stats) {
+    // VirtuallChassis MIB uses string indexes so dummy integer indexes for
+    // VC ports need to be created.
+
+    // Check if index string has expected format and decompose it to form
+    // dummy integer index of it.
+
+    if (preg_match('#^(\d{1,2})\.vcp-255/(\d)/(\d{1,2})$#', $index, $matches)) {
+        if (strlen($matches[1]) == 1) {
```
murrant: Instead of these if statements, you could just us str_pad():
```php
str_pad($matches[1], 2, '0', STR_PAD_LEFT)
```
1 => 01
99 => 99
432 => 432

murrant:
Looks good just a small change to reduce complexity a bit

includes/polling/ports/junos-vcp.inc.php
rmagomedov: Yeah, this looks more compact and easier to read. Done

PR #9889, opened 2019-03-01
New Device: ATS - Automatic Transfer Switch

TheGreatDoc:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


TheGreatDoc:
I cant get the High/Low limit work for the temperature sensor.

It snmpget the value but set it as Array when discovering sensor

```
Caching data: sensors SNMP['/usr/bin/snmpwalk' '-v2c' '-c' 'COMMUNITY' '-OteQUs' '-Ih' '-m' 'ATS-MIB' '-M' '/opt/librenms/mibs:/opt/librenms/mibs/ats' 'udp:HOSTNAME:161' 'emdConfigTempHighSetPoint.0']
emdConfigTempHighSetPoint.0 = 40

SNMP['/usr/bin/snmpwalk' '-v2c' '-c' 'COMMUNITY' '-OteQUs' '-Ih' '-m' 'ATS-MIB' '-M' '/opt/librenms/mibs:/opt/librenms/mibs/ats' 'udp:HOSTNAME:161' 'emdConfigTempLowSetPoint.0']
emdConfigTempLowSetPoint.0 = 5
```
And then:
```
Final sensor value: 29
Cur 29, Low: Array, Low Warn: null, Warn: null, High: Array
```

murrant:
probably because the code expects tables.  If you dump out the data after pre-cache, you should see it as an array.


FYI, the logo seems to be a copy of the icon. (icon is used if the logo doesn't exist)

TheGreatDoc:
yes, its a "copy". 

About pre-cache, it worked for the other limits and walk/return are exactly the same.

murrant:
You are looking at the raw snmpwalk output, not the data after it is returned by the snmp function.  It is in an array that doesn't track with what you expect.

TheGreatDoc:
I do understand what you mean. What I say is: why it works for the other limits?

What pre-cache does is snmpwalk and all limits returns the same in the walk. Just 1 oid with 1 value.

murrant:
because it is grouping the data by the index.  If you remove the extra index, it strips a level of arrays.

TheGreatDoc:
@murrant totally forgot about this.

So, how should I proceed? Just remove the temp limits from yaml?

murrant:
Probably need to move that part to php

Issue #8015, opened 2018-01-03
Grandstream ATA Support


CoMMyz:
Would like to add support for Grandstream ATA devices. The new firmware released supports SNMP finally and they provided the MIB files.

Here are the required information from a live working HT812 device.
discovery https://p.libren.ms/view/f0229bea
poller https://p.libren.ms/view/63ffadd3
snmpbulkwalk https://p.libren.ms/view/684aa29d

MIB attached: 
[MIB_HT8xx.zip](https://github.com/librenms/librenms/files/1601451/MIB_HT8xx.zip)

Thank you
DO NOT DELETE THIS INFORMATION.

> Please read this information carefully.

GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z. 

You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).

If you have a feature request please post this on our [community site](https://community.librenms.org/c/feature-requests).

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [x] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.
- [x] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
====================================
Component | Version
--------- | -------
LibreNMS  | 1.33-297-g34ed39c
DB Schema | 227
PHP       | 5.4.16
MySQL     | 5.5.56-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
====================================

- [x] Please provide ALL info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
- [x] Please provide as much detail as possible.
The Grandstream HT812 is a device that provides 2 FXS ports over SIP typically used for providing telephony to analog telephone devices.

- [x] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

PR #9906, opened 2019-03-05
Grandstream basic support

PipoCanaja:
Grandstream very basic support (MIB pretty bad).

Had to exclude "Linux HT" from the linux.yaml file to match these devices. If anybody has a better idea ... 

Will close https://github.com/librenms/librenms/issues/8015

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
@murrant Just a note to double check the change on the linux.yaml file ... Hope it does not break anything ... but ... more eyes is always better

murrant:
@PipoCanaja I think linux is treated specially, it is check AFTER all other OS are checked.

PipoCanaja:
I am not so sure cause during my tests, without the change in the linux.yaml, the device was always detected as linux and not grandstream-ht.

PipoCanaja:
Let's WIP this, I'll test again if I can avoid the change in linux.yaml. Will be a little bit less available in the next few days, except if the weather is bad for skiing ...

PipoCanaja:
Works correctly now without touching linux.yaml, so I reverted that.

PR #9916, opened 2019-03-07
Fixed sysName can't display because newline character

jasoncheng7115:
Fixed sysName can't display because newline character.
The reason is that the sysName retrieved by SNMP has a newline character

See forum for details:
https://community.librenms.org/t/ipv4-ipv6-mac-address-arp-table-page-empty/7524/6

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/pages/ports.inc.php
+++ b/html/pages/ports.inc.php
@@ -94,9 +94,9 @@ if ((isset($vars['searchbar']) && $vars['searchbar'] != "hide") || !isset($vars[
     $output .= "<option value=''>All Devices</option>";
 
     if (LegacyAuth::user()->hasGlobalRead()) {
-        $results = dbFetchRows("SELECT `device_id`,`hostname`, `sysName` FROM `devices` ORDER BY `hostname`");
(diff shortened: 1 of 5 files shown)


murrant:
Thanks @jasoncheng7115 but this fix is in the wrong place.  It should be either at poller/discovery or in the web page (or both).

jasoncheng7115:
> Thanks @jasoncheng7115 but this fix is in the wrong place. It should be either at poller/discovery or in the web page (or both).

Oh.. Ok, thank you for your prompt.
Let me see how to modify it.

murrant:
Probably the web page for sure, the line returns won't bother other places this is used, but it will here because of the way it is injected into the javascript.

jasoncheng7115:
> Probably the web page for sure, the line returns won't bother other places this is used, but it will here because of the way it is injected into the javascript.

I already know how to modify it, from the poller entry,
Submit PR after modification.

jasoncheng7115:
This PR modification method is not good enough. In addition, I will submit a new PR.

PR #9923, opened 2019-03-08
Fix database validation for MySQL 8

llcoolkm:
The validate script reports all timestamp columns as non-compliant because MySQL introduced a new schema in version 8. The extra field now adds "DEFAULT_GENERATED for columns that have an expression default value.". See https://dev.mysql.com/doc/refman/8.0/en/columns-table.html

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Thanks for the fix @llcoolkm

PR #9929, opened 2019-03-11
Patch fix sql error on gengroupsql

dracmic:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/device-groups.inc.php
+++ b/includes/device-groups.inc.php
@@ -106,7 +106,7 @@ function GenGroupSQL($pattern, $search = '', $extra = 0)
 
     $tables = array_unique($tables);
     $pattern = rtrim($pattern, '&|');
-    
+
     if ($tables[0] != 'devices' && dbFetchCell('SELECT 1 FROM information_schema.COLUMNS WHERE TABLE_NAME = ? && COLUMN_NAME = ?', array($tables[0],'device_id')) != 1) {
         //Our first table has no valid glue, prepend the 'devices' table to it!
         array_unshift($tables, 'devices');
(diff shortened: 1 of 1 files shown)


PipoCanaja:
Hi @dracmic 
Could you detail the bug you are facing ? any way to reproduce it ?
PipoCanaja

includes/device-groups.inc.php
```diff
@@ -153,6 +152,11 @@ function GenGroupSQL($pattern, $search = '', $extra = 0)
     if (!empty($join)) {
         $join = '('.rtrim($join, '& ').') &&';
     }
+  //patch sql error adding extra devices table in query when it has 3 elements 
+    $pos=array_search('devices',$tables);
```
murrant: This line is unused.

It would possibly be better to just call `array_unique($tables)` to remove any duplicate tables.

includes/device-groups.inc.php
```diff
@@ -153,6 +152,11 @@ function GenGroupSQL($pattern, $search = '', $extra = 0)
...
+    if (count($tables)==3) {
+    array_pop($tables);
```
murrant: It would possibly be better to just call `array_unique($tables)` to remove any duplicate tables.

This current code removes the last table when there happens to be 3 tables.

dracmic: Hello,

I am open to suggestion, but being a beginner in writing code and I just like to debug. I just wanted to help. I can redo the commit with array_unique.

murrant: You can also update this PR.  Currently the code doesn't work as you intend (in some cases).

I think just `$tables = array_unique($tables);` should be all you need to add.

PR #9933, opened 2019-03-11
Corrected active_count for Alert icon color

PipoCanaja:
Hi,

I did receive questions from my colleagues on the Alert Icon color in the menubar. Red or green does not follow the implicit logic.

After a check, it seems that the SQL calculating it is wrong. As far as I see it, we want:
- Green if all alerts in DB are OK (status 0) or ACK (status 2)
- Red in any other cases. 

That means : ```AND `alerts`.`state` NOT IN (0,2)'``` instead of ```AND `alerts`.`state` NOT IN (0,1)' ```as it is now. 
Moreover, Devices with disabled or ignore state should not be taken into account.

Here is a patch for legacy and laravel icon coloring. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/Http/ViewComposers/MenuComposer.php
+++ b/app/Http/ViewComposers/MenuComposer.php
@@ -222,7 +222,6 @@ class MenuComposer
         $alert_status = AlertRule::select('severity')
             ->isActive()
             ->hasAccess($user)
-            ->groupBy('severity')
             ->leftJoin('devices', 'alerts.device_id', '=', 'devices.device_id')
             ->where('devices.disabled', '=', '0')
             ->where('devices.ignore', '=', '0')
--- a/app/Http/ViewComposers/MenuComposer.php
+++ b/app/Http/ViewComposers/MenuComposer.php
@@ -223,6 +223,10 @@ class MenuComposer
             ->isActive()
             ->hasAccess($user)
             ->groupBy('severity')
+            ->leftJoin('devices', 'alerts.device_id', '=', 'devices.device_id')
+            ->where('devices.disabled', '=', '0')
+            ->where('devices.ignore', '=', '0')
+            ->groupBy('severity')
             ->pluck('severity');
 
         if ($alert_status->contains('critical')) {
--- a/app/Models/AlertRule.php
+++ b/app/Models/AlertRule.php
@@ -52,7 +52,7 @@ class AlertRule extends BaseModel
     {
         return $query->enabled()
             ->join('alerts', 'alerts.rule_id', 'alert_rules.id')
-            ->whereIn('alerts.state', [1, 3, 4]);
+            ->whereNotIn('alerts.state', [0, 2]);
     }
 
     /**
--- a/app/Models/AlertRule.php
+++ b/app/Models/AlertRule.php
@@ -52,7 +52,7 @@ class AlertRule extends BaseModel
     {
         return $query->enabled()
             ->join('alerts', 'alerts.rule_id', 'alert_rules.id')
-            ->where('alerts.state', 1);
+            ->whereIn('alerts.state', [1, 3, 4]);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


PipoCanaja:
We should also probably ignore "devices.ignore = 1" in the count for this, I suppose ?

murrant:
@PipoCanaja yeah, can you check the functionality of the Laravel menu (menu.blade.php and MenuComposer.php)

We might be able to remove the old menu and just use the Laravel menu. That is technically possible, just not sure about the details.

PipoCanaja:
Not clear on how to test these. Should we review and merge this first (so we correct the behaviour of the legacy code) and I'll check on how to test the laravel version ?

murrant:
Go to the all locations page (/locations)

PipoCanaja:
correct, issue is not corrected on locations. So I'll dig that.

PipoCanaja:
Seems to be OK now, but it is really not my thing ! I had SQL lessons at university but that was loooooong before eloquent :D :D 
Basically, I wanted to add the join for devices ignore and enabled.
Open to any suggestion for cleaner solution.

PipoCanaja:
@murrant Alert Transports are not in the Laravel version of the menu. 
Legacy: 
Capture d’écran 2019-03-12 à 19 28 48

Laravel: 
Capture d’écran 2019-03-12 à 19 28 42

That's out of scope of this PR. I'll start a new one later on.

app/Http/ViewComposers/MenuComposer.php
```diff
@@ -223,6 +223,10 @@ public function compose(View $view)
             ->isActive()
             ->hasAccess($user)
             ->groupBy('severity')
+            ->leftJoin('devices', 'alerts.device_id', '=', 'devices.device_id')
+            ->where('devices.disabled', '=', '0')
+            ->where('devices.ignore', '=', '0')
```
PipoCanaja: Should we include this in "isActive" ? I mean, an alert can be active if and only if those 2 where conditions on the device are met.

app/Models/AlertRule.php
```diff
@@ -39,7 +39,7 @@ class AlertRule extends BaseModel
      */
     public function scopeEnabled($query)
     {
-        return $query->where('disabled', 0);
+        return $query->where('alert_rules.disabled', 0);
```
PipoCanaja: disabled is a column name that exist in other tables.

app/Http/ViewComposers/MenuComposer.php
murrant: Not in a simple way.  Joining tables can have implications on the query, so I don't like hiding it in a scope.

app/Http/ViewComposers/MenuComposer.php
```diff
@@ -223,6 +223,10 @@ public function compose(View $view)
...
+            ->groupBy('severity')
```
murrant: groupBy('severity') is already set on this query ;)

app/Http/ViewComposers/MenuComposer.php
murrant: But maybe :D

app/Models/AlertRule.php
```diff
@@ -52,7 +52,7 @@ public function scopeIsActive($query)
     {
         return $query->enabled()
             ->join('alerts', 'alerts.rule_id', 'alert_rules.id')
-            ->where('alerts.state', 1);
+            ->whereIn('alerts.state', [1, 3, 4]);
```
murrant: I noticed you use whereNotIn in the legacy index and whereIn here.  Not an issue just an inconsistency.

app/Http/ViewComposers/MenuComposer.php
PipoCanaja: I'll follow your 1st advice and keep it as it is for now.

app/Models/AlertRule.php
PipoCanaja: Corrected.

PipoCanaja:
Tests in my environnement are good and result are the same with Laravel and legacy code.

PR #9951, opened 2019-03-13
Display Up/Down time in Device List

PipoCanaja:
This is an attempt for Uptime Downtime display in the device list. Thought ? I tried to reuse most of existing code, only adding the minimum.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/Models/Device.php
+++ b/app/Models/Device.php
@@ -396,6 +396,11 @@ class Device extends BaseModel
         return $this->status;
     }
 
+    public function getLastPolled()
+    {
+        return strtotime($this->last_polled);
+    }
+
     public function getIconAttribute($icon)
     {
         $this->loadOs();
--- a/app/Models/Device.php
+++ b/app/Models/Device.php
@@ -13,6 +13,7 @@ use LibreNMS\Util\IP;
 use LibreNMS\Util\IPv4;
 use LibreNMS\Util\IPv6;
 use LibreNMS\Util\Url;
+use LibreNMS\Util\Time;
 
 class Device extends BaseModel
 {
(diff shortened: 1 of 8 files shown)


PipoCanaja:
Capture d’écran 2019-03-14 à 00 09 03

app/Models/Device.php
```diff
@@ -251,10 +251,14 @@ public function canAccess($user)
             ->where('device_id', $this->device_id)->exists();
     }
 
-    public function formatUptime($short = false)
+    public function formatUptime($short = false, $downtime = false)
```
murrant: Instead of a function that does two things, how about two functions that do one thing :)

I know I wrote it, but I don't really like the formatUptime. It is in the wrong place.  Probably move it to the Time util class as Time::formatInterval(CarbonInterval $interval) (or it could just accept seconds)

Then the above would be something like:
```php
'uptime' => Time::formatInterval($device->status ? $device->uptime : ($now - $device->last_polled)),
```

PipoCanaja: For compatibility issue, the function Time::formatInterval uses seconds only.

app/Models/Device.php
```diff
@@ -396,6 +396,11 @@ public function getStatus()
         return $this->status;
     }
 
+    public function getLastPolled()
```
murrant: last_polled should be a carbon object, so you don't need this function.

PipoCanaja: I cannot access $device->last_polled from DeviceController.php: 
OK: ```'uptime' => Time::formatInterval($device->getStatus() ? $device->uptime : time() - $device->getLastPolled(), 'short'),```
NOK:``` 'uptime' => Time::formatInterval($device->getStatus() ? $device->uptime : time() - $device->last_polled, 'short'),```

murrant: Try $device->lastPolled

PipoCanaja: Still not. This one does not fail but the $device->lastPolled is empty.

murrant:
@PipoCanaja ok, I remembered you have to add it to casts on the device model:
```php
    protected $casts = [
        'last_polled' => 'datetime',
        'status' => 'boolean',
    ];
```

Then you can get downtime like this:
```php
$device->last_polled->diffInSeconds()
```

murrant:
Also, you should update Device:formatUptime() to use Time::formatInterval() to remove code duplication.  (or remove it entirely)

Sorry this PR has been a bit of monkeying around, but the Eloquent models haven't settled in yet, so sometimes they need to be improved a bit.

PipoCanaja:
Much better now. Cleaning of the various formatUptime() I found using grep :) statrs to look more consistant.

app/Models/Device.php
```diff
@@ -387,6 +364,16 @@ public function statusName()
 
     // ---- Accessors/Mutators ----
 
+    public function getStatus()
```
murrant: Can you please remove getStatus() and getLastPolled(), just use the property accessors: ->status and ->last_polled

PR #9955, opened 2019-03-14
Update documentation to reflect matching behaviour

kiwibrew:
Added a note to help users with matching patterns for Oxidized OS matching patterns.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Oxidized.md
+++ b/doc/Extensions/Oxidized.md
@@ -108,6 +108,8 @@ To match on a device os of edgeos then please use the following:
 $config['oxidized']['maps']['group']['os'][] = array('match' => 'edgeos', 'group' => 'wireless');
 ```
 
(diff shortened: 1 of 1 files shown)


murrant:
A good addition, but it is a little more than just lowercase.  It requires the system name for an os such as hpe-msa

PipoCanaja:
> A good addition, but it is a little more than just lowercase. It requires the system name for an os such as hpe-msa

By the way, Is there an easy way to see the os_name other than digging into the YAML file or the DB ?

murrant:
I could not think of one other than the raw yaml, or db.

PipoCanaja:
> I could not think of one other than the raw yaml, or db.

If I think of a place in the GUI that could provide this information, I'll add it.

murrant:
How about adding it in tooltips or something?

PR #9965, opened 2019-03-15
Patch \r and \n in RocketChat Transport for channel,emoji and icon

dracmic:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@dracmic what causes these to have line returns?

Also, wouldn't trim() be better than str_replace?

dracmic:
@murrant from investigation is directly from form when adding the details from Javascript in to the database because from debugging they are in db with \r and \n.

murrant:
fixed by #10001

PR #9969, opened 2019-03-15
Add HTML transport example for ms teams

sorano:
and fix missing " in Processor Alert with Graph example.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Alerting/Templates.md
+++ b/doc/Alerting/Templates.md
@@ -304,12 +304,26 @@ Rule: @if ($alert->name) {{ $alert->name }} @else {{ $alert->rule }} @endif <br>
 {{ $key }}: {{ $value['string'] }}<br>
 @endforeach 
 @if ($alert->faults) <b>Faults:</b><br>
-@foreach ($alert->faults as $key => $value)<img src="https://server/graph.php?device={{ $value['device_id'] }}&type=device_processor&width=459&height=213&lazy_w=552&from=end-72h><br>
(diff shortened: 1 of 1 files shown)


doc/Alerting/Templates.md
```diff
@@ -304,14 +304,28 @@ Rule: @if ($alert->name) {{ $alert->name }} @else {{ $alert->rule }} @endif <br>
...
-@foreach ($alert->faults as $key => $value)<img src="https://server/graph.php?device={{ $value['device_id'] }}&type=device_processor&width=459&height=213&lazy_w=552&from=end-72h><br>
+@foreach ($alert->faults as $key => $value)<img src="https://server/graph.php?device={{ $value['device_id'] }}&type=device_processor&width=459&height=213&lazy_w=552&from=end-72h"><br>
 https://server/graphs/id={{ $value['device_id'] }}/type=device_processor/<br>
 @endforeach 
 Template: CPU alert <br>
 @endif
 @endif
 ```
 
+#### MS Teams formatted default template:
```
murrant: Why is this ms teams specific?

sorano: Well, I guess it's a somewhat modified truth and probably could be used as a generic template as well.

Reason I called it teams specific is that teams supports limited tags:
https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/bots/bots-text-formats

So if someone decides to update the template in the future maybe they would add tags that were not supported in teams.

murrant: Hmm, seems like we should have a template collection like we do for alert rules.

PR #9973, opened 2019-03-15
Optional automatic sensor limits

amigne:
When a device does not give sensor limit values, LibreNMS automatically sets limits, based on the current value measured during the first discovery.

In environments where equipments providing such limit information and others that don't are mixed, it is impossible to determine at first if the limit value is provided by the vendor or by the tool.

In cases where the LibreNMS administrator would trust the vendor values (for instance, to generate alerts) and prefer have no values set at all when the vendor does not return any (these values can be manually set, if needed), a new configuration parameter can be set. If this parameter is true (default value), the discoverer will determine an automatic value; if the parameter is false, LibreNMS does not determine any automatic value when the device does not provide any.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
The name of the config variable doesn't feel quite right.  How about `sensors.guess_limits`?

TheGreatDoc:
Do you mind also documenting this new option in https://docs.librenms.org/Support/Configuration/ ?

amigne:
> Do you mind also documenting this new option in https://docs.librenms.org/Support/Configuration/ ?

@TheGreatDoc I'll do it with the update of my PR.

PipoCanaja:
I guess you can remove the WIP tag in the title to get this merged.

murrant:
@amigne naming might be trivial, but it is the only thing we can't fix trivially ;)

PR #9999, opened 2019-03-19
Bug fix if locations table is first in $tables array

dracmic:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/device-groups.inc.php
+++ b/includes/device-groups.inc.php
@@ -140,6 +140,10 @@ function GenGroupSQL($pattern, $search = '', $extra = 0)
                     $tables[] = $tmp;
                 }
             }
+              // patch  if  first table is locations
+               if ($tables[0]=='locations') {
+                $tables=array_reverse($tables);
+               }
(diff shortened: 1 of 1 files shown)


includes/device-groups.inc.php
```diff
@@ -140,6 +140,10 @@ function GenGroupSQL($pattern, $search = '', $extra = 0)
                     $tables[] = $tmp;
                 }
             }
+              // patch  if  first table is locations
+               if ($tables[0]=='locations') {
+                $tables=array_reverse($tables);
```
murrant: This will only fix some problems.  $tables is user generated, so 'devices' might not be the last field.

dracmic: murrant, but I cannot say the opposite thing but the issue on SQL query even written on discord is "SELECT DISTINCT(locations.device_id)" and on where part of line 162 of file.

murrant:
Fixed in #10000

PR #10006, opened 2019-03-20
Add Device Arris Apex

djamp42:
Tested with Arris APEX-1000

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/arris-apex.yaml
```diff
@@ -0,0 +1,11 @@
+os: arris-apex
+text: 'Arris Apex'
+type: network
+icon: arris
+mib_dir:
+    - arris
+over:
+    - { graph: '', text: '' }
```
PipoCanaja: Isn't it any modem counter available that you could poll and display here ? In a CMTS this is usually something you would like to see displayed as the over graph.

PipoCanaja:
Hello @djamp42 
Thanx for your PR. Could you have a look at the Codeclimate issues and solve them ? only a few blanks to remove as far as I can see.
And as I wrote in the YAML file, I am surprised not to see any modem counter. That seems surprising for a CMTS. You can use the "Count" sensor type for this kind of data.

djamp42:
@PipoCanaja This is not a CMTS, it's a IP to QAM Modulator (EdgeQAM) used in cable tv headends.. I'll fix the code climate later tonight.

PipoCanaja:
Other common value in cable operator world is noise. Nothing else to add in that direction ? 
Concerning the tests, unless you add other sensors, should be already good.

djamp42:
No, this is sending out the qam signal so noise isn't measured here, nothing in the upstream direction either on this thing either. I looked to tried to find something for the hover graphs but nothing really made sense.

PR #10012, opened 2019-03-22
Add SysName to Oxidized view

smiles1969:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 10012`

### DIFF
--- a/html/includes/functions.inc.php
+++ b/html/includes/functions.inc.php
@@ -1349,9 +1349,15 @@ function get_oxidized_nodes_list()
         }
         echo "
         </td>
+        ";
+        if ($config['force_ip_to_sysname'] === false) {
+        echo "
         <td>
         " . $device['sysName'] . "
         </td>
+        ";
+        }
+        echo "
         <td>
         <i class='fa fa-square text-" . $fa_color . "'></i>
         </td>
--- a/html/includes/functions.inc.php
+++ b/html/includes/functions.inc.php
@@ -1350,7 +1350,7 @@ function get_oxidized_nodes_list()
         echo "
         </td>
         ";
-        if ($config['force_ip_to_sysname'] === false) {
+        if (Config::get('force_ip_to_sysname') === false) {
         echo "
         <td>
         " . $device['sysName'] . "
--- a/html/includes/functions.inc.php
+++ b/html/includes/functions.inc.php
@@ -1349,15 +1349,9 @@ function get_oxidized_nodes_list()
         }
         echo "
         </td>
-        ";
-        if (!Config::get('force_ip_to_sysname')) {
-            echo "
-            <td>
-            " . $device['sysName'] . "
-            </td>
-            ";
-        }
-        echo "
+        <td>
+        " . $device['sysName'] . "
+        </td>
         <td>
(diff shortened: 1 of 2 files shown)


PipoCanaja:
Hi @smiles1969 
Thanx for your PR. This patch can make sense in some situation, but not really if you have "force_ip_to_sysname" enabled in the LibreNMS instance. In that case you end up with 2times the sysname displayed, which is a little bit annoying : 

Capture d’écran 2019-03-23 à 00 55 42

May be you could replace this column by the IP address in case the "force_ip_to_sysname" config option is set. Or make the column hidden by default, so it is only enabled when needed.

smiles1969:
I have added the config checks as suggested

html/includes/functions.inc.php
```diff
@@ -1349,6 +1349,15 @@ function get_oxidized_nodes_list()
         }
         echo "
         </td>
+        ";
+        if ($config['force_ip_to_sysname'] === false) {
```
murrant: Please use `Config::get('force_ip_to_sysname')` instead of accessing the global variable.

smiles1969: Done

murrant:
@smiles1969 sorry, I missed this the first time but I just noticed you used strict comparison.  A loose comparison would be better as the value may be null.  Or simply just `!Config::get('force_ip_to_sysname')`

smiles1969:
I've changed to !Config::get instead of == loose comparison and tested on my local repository.

PipoCanaja:
@smiles1969 Could you have a look at Travis tests ?

PipoCanaja:
Looks good. @librenms/reviewers  if you have a test setup with oxidized and not force_ip_to_sysname, do not hesitate to give feedback.

murrant:
@smiles1969 thanks for all the changes, but I actually realized something when looking at this today.
We should always send the column data and have it exist, just hide it when appropriate

```
echo '<th data-column-id="sysname" data-visible="' . (!Config::get('force_ip_to_sysname')  ? 'true' : 'false') . '">SysName</th>';
```

html/includes/functions.inc.php
```diff
@@ -1349,6 +1349,15 @@ function get_oxidized_nodes_list()
         }
         echo "
         </td>
+        ";
+        if (!Config::get('force_ip_to_sysname')) {
```
murrant: There should be no if statement here, just output the column

html/pages/oxidized.inc.php
```diff
@@ -30,6 +30,8 @@
                             <thead>
                             <tr>
                                 <th data-column-id="hostname" data-order="desc">Hostname</th>
+                                <?php echo '<th data-column-id="sysname" data-visible="' . (!Config::get('force_ip_to_sysname')  ? 'true' : 'false') . '">SysName</th>';
```
murrant: you could move most of the html outside the php tags, just the true/false ternary needs to be inside <?php

murrant:
Looks good :)

Issue #8480, opened 2018-03-27
New device/OS:  UPS Generic with I-Stars Network SNMP card


fabio-jgs:
**DO NOT DELETE THIS INFORMATION.**

> **Please read this information carefully.**

**GitHub issues is for known/validated bugs, please do not post issues asking for help or how to do X, Y or Z.** 

**You can use our [Discord server](https://t.libren.ms/discord) to ask questions or our [community site](https://community.librenms.org).**

**If you have a feature request/changes please post this on our [community site](https://community.librenms.org/c/feature-requests).**

Please confirm each of the sections below by putting an `x` in the box like `[x]`.

- [X] Is your install up to date? [Updating your install](http://docs.librenms.org/General/Updating/)
      Please do not submit an issue if your install is not up to date within the last 24 hours or on a stable monthly release.

- [X] Please include all of the information between the `====================================` section of `./validate.php` which you can run from the cli.
Component | Version
--------- | -------
LibreNMS  | 1.37-85-ga1d80e394
DB Schema | 246
PHP       | 7.0.27-0+deb9u1
MySQL     | 10.1.26-MariaDB-0+deb9u1
RRDTool   | 1.6.0
SNMP      | NET-SNMP 5.7.3

- [X] Please provide **ALL** info asked for [here](http://docs.librenms.org/Support/FAQ/#faq20).
./discovery.php https://p.libren.ms/view/6f06dad1
./poller.php https://p.libren.ms/view/95786f49
snmpbulkwalk https://p.libren.ms/view/0792b7f6

- [X] Please provide as much detail as possible.

There is an implementation for monitoring of generic UPS with SNMP CARD from manufacturer EAST - Global Energy. More information at: http://www.eastups.com/en/products_show_30_39_110.html.
Using the standard MIB UPS-MIB it is possible to identify: Load, Frequency, Temperature, Battery charge, and status.
Tests were performed with 3 different devices with the same card

- [X] Please do NOT post more than 10 lines of debug information here, use a pastebin service or GitHub Gists.

Love librenms? Please consider supporting our collective:
👉  https://t.libren.ms/donations

spencerbutler:
Hey @fabio-jgs Thanks for your patience. If you can kindly test this PR #10032, it may resolve your issue. Feedback welcome.

PipoCanaja:
Hi @fabio-jgs 
This device is now discovered correctly. Feel free to reopen this issue if you find something missing. 
Bye

PR #10032, opened 2019-03-27
Added support for Generic UPS-MIB (os: upsmib)

spencerbutler:
This aims to support any generic UPS that conforms to the UPS-MIB standard, and doesn't fit anywhere else. It **still needs an ICON**, submissions encouraged.

Fixes:
#8480
(probably others, I'll note them if I find them.)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hello
This support is already built in LibreNMS, with the name rfc1628. This is a yaml option, and any UPS with this yaml option just basically need to be declared and have an icon. Everything else is automated :) So this PR is not necessary. 
If a value is missing, you would have to extend the existing rfc1628 support.

PR #10034, opened 2019-03-27
WIP fix email duplicates

murrant:
add mail_use_bcc setting to put all to addresses in bcc instead
Send one email to the server instead of multiple individual mails

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Just a thought, I'm not sure it makes sense.  I don't see much reason to bcc instead of to.

laf:
> Just a thought, I'm not sure it makes sense. I don't see much reason to bcc instead of to.

The only reason for us not to use it (at work that is) is so you have transparency of who else gets the alerts.

SourceDoctor:
mmh BCC for Mail Transports?
What about a middle way with an option switch between CC and BCC with CC as default?
But by default i would also prefer to see who else get's this mail

murrant:
Needs rebase

PR #10041, opened 2019-03-28
Added support for East iStars UPS (os: istars)

spencerbutler:
Fixes:
#8480 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hi @spencerbutler 
This one is the one that impelements rfc1628 badly, is it ? 
Let me know if you have some tests done by requester or if we merge it right away. 
Bye

spencerbutler:
> Hi @spencerbutler
> This one is the one that impelements rfc1628 badly, is it ?
> Let me know if you have some tests done by requester or if we merge it right away.
> Bye

No feedback from the requester, as of yet. Let's wait a bit to see if any comes in.

spencerbutler:
> @spencerbutler If the image files are the same, you can put only the 'os' and it will be always used, instead of duplicating the file (and the space used).

ah, good advice. thanks and fixed.

PipoCanaja:
LGTM

spencerbutler:
> Instead of redefining UPS-MIB sensors, use this in the os yaml:
> 
> ```yaml
> rfc1628_compat: true
> ```

fixed

includes/discovery/sensors/current/rfc1628.inc.php
```diff
@@ -34,6 +34,9 @@
     if (count($output_current) > 1) {
         $descr .= " Phase $index";
     }
+    if (is_array($data['upsOutputCurrent'])) {
```
murrant: good, they messed up their rfc1628 implementation...

It would be better to check `isset($data['upsOutputCurrent'][0])` I think

spencerbutler: There were a few other instances of this check in rfc1628 code, I updated them too.

includes/discovery/sensors/load/rfc1628.inc.php
```diff
@@ -7,7 +7,7 @@
 foreach ($load_data as $index => $data) {
     $load_oid = ".1.3.6.1.2.1.33.1.4.4.1.5.$index";
 
-    if (is_array($data['upsOutputPercentLoad'])) {
+    if (isset($data['upsOutputPercentLoad'][0])) {
```
laf: Any reason this was changed?

spencerbutler: Yes, @murrant suggested it was a better way to check, so I updated all references to the upsDataFoo from is_array to isset.

laf:
The tests are still failing: https://travis-ci.com/librenms/librenms/jobs/191452775

This looks like it's because it's affecting the existing data for sensors - are you sure this is backwards compatible?

spencerbutler:
> The tests are still failing: https://travis-ci.com/librenms/librenms/jobs/191452775
> 
> This looks like it's because it's affecting the existing data for sensors - are you sure this is backwards compatible?

I will investigate this further.

murrant:
Hmm, looks like your changes to rfc1628.inc.php broke some other devices.

spencerbutler:
> Hmm, looks like your changes to rfc1628.inc.php broke some other devices.

Yeah, I'll see if I can fix that. Otherwise, I'll just revert the changes I made outside the scope of this PR.

murrant:
@spencerbutler Still working on this?

spencerbutler:
> @spencerbutler Still working on this?

I've been out of play, but I do intend on fixing this up. I'll make an effort to finish it this week.

murrant:
@spencerbutler your changes to all the rfc1628 files breaks everything. :)

spencerbutler:
> @spencerbutler your changes to all the rfc1628 files breaks everything. :) Either revert them or change it so it doesn't break other devices.

Reverted.

includes/discovery/sensors/current/rfc1628.inc.php
murrant: Other rfc1628 types have this code block:
```php
    if (is_array($data['upsOutputVoltage'])) {
        $upsOutputVoltage_value = $data['upsOutputVoltage'][0];
        $volt_oid .= ".0";
    }
```

You should probably update the oid being discovered as well.

tests/data/istars.json
```diff
@@ -1181,14 +1181,14 @@
...
-                    "sensor_oid": ".1.3.6.1.2.1.33.1.3.3.1.2.1",
+                    "sensor_oid": ".1.3.6.1.2.1.33.1.3.3.1.2.1.0",
                     "sensor_index": "3.2.0.1",
                     "sensor_type": "rfc1628",
                     "sensor_descr": "Input",
                     "group": null,
                     "sensor_divisor": 10,
                     "sensor_multiplier": 1,
-                    "sensor_current": 0,
+                    "sensor_current": 60,
```
murrant: Hey look, your test data works :)

Issue #9880, opened 2019-02-26
New Device: Cisco FirePower 2140


07stuntar1:
Hi I just tried to add Cisco Firepower 2140 Security Appliance to LibreNMS and it only shows "octeon"
Generic Device.

Cisco Firepower Threat Defense, Version 6.3.0.1 (Build 85), ASA Version 9.10(1)8
System Nameocteon
IP Address:
Operating SystemGeneric Device
Object ID.1.3.6.1.4.1.9.1.2407

====================================
Component | Version
--------- | -------
LibreNMS  | 1.48.1-85-gee76057
DB Schema | 2019_02_10_220000_add_dates_to_fdb (132)
PHP       | 7.2.14
MySQL     | 5.5.60-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
====================================

discovery 
https://p.libren.ms/view/51f4ae7d

poller
https://p.libren.ms/view/db1b7d76

snmpbuilkwalk
https://p.libren.ms/view/d43587bb

Please let me know if you need anything else.
Thank You

spencerbutler:
These files have been verified, and I will submit a PR in the near future.

07stuntar1:
Thank You

spencerbutler:
@07stuntar1 Here is an initial [commit](#10046) to get some graphs drawn for you. I plan on adding some sensors and state information to it, if I am able. Give that a shot and let me know what you think.

07stuntar1:
@spencerbutler 
Did I pull down the right configs I ran
./scripts/github-apply 10046

Still no graphs.

spencerbutler:
> @spencerbutler
> Did I pull down the right configs I ran
> ./scripts/github-apply 10046
> 
> Still no graphs.

did you rediscover the device?

07stuntar1:
I just deleted the device and ran a the discover and poller php script for the ip address
![image](https://user-images.githubusercontent.com/30091711/55205132-b74fe200-5175-11e9-9f5d-531dcd3658d9.png)

spencerbutler:
This is what I get with your snmp data.
![ciscofpr-2140td](https://user-images.githubusercontent.com/5126075/55205562-6524b600-51a1-11e9-9382-5b2d7f1648b7.png)

07stuntar1:
[...]

PR #10046, opened 2019-03-29
Updated OS (ftd) for Cisco FirePOWER devices

spencerbutler:
Fixes:
#9880 

There is a similar OS named ftd, but most of the modules are disabled. The CISCO-PRODUCTS-MIB lists several Cisco FirePOWER devices in the same OID range -- I don't know if they all need their own OS, or if they can be grouped into the same OS. Please advise.

I would like to get some sensors added to this OS, hence the "WiP" tag.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
ftd should in fact be ciscoftd to be consistant, but renaming an OS is a bit of a problem for the actual users. So I would avoid the rename. 
Now, most probably all Firepower devices should be included in ftd OS, unless you find an incompatibility that would justify not to. 
I would suggest to start this approach first, and only spend time on the new OS if that approach fails.

spencerbutler:
The only incompatibility I found was that the sensors, mempool, and processor modules are disabled. Do you have any idea why it is configured in that manor?

PipoCanaja:
Nope, not at all. I would still suggest to start from that OS, with the snmpwalk data of the issue, try re-enabling the discovery (which any user can disable from config.php if he wants) and see what you discover. There is no definitions/discovery/ftd.yaml file for the moment, so everything that is coming today is completely standard, not at all specific to ftd anyway.

spencerbutler:
I did start with that OS and moved to this one, named from the CISCO-PRODUCTS-MIB definitions (ie ciscoFpr2140td  OBJECT IDENTIFIER ::= { ciscoProducts 2407 }) because the discovery modules are disabled it ftd. There is only one Ftd device listed in the CISCO-PRODUCTS-MIB, and it's OID is not included in the ftd OS, so I'm at a lost for why: a) it's name doesn't match the documentation b) it's modules are disabled.

PipoCanaja:
The OS name in LibreNMS is absolutely arbitrary. Only depends on what name was chosen by the contributor. 
Discovery modules disabled or not is not important. As the FirePower devices are running linux, may be the contributor wanted to avoid the default linux values, it will be difficult to see. But adding new sensors to an existing OS does not harm existing devices that would not have this sensor. It is necessary to split only if devices behave differently enough.

spencerbutler:
> I think you just need to update the ftd os.

That is the plan. I had an unexpected event carry me away from coding for the past 2 weeks. I'll have a new commit for this "soon".

spencerbutler:
Added to OS (ftd).

PR #10051, opened 2019-04-01
WIP Add offset variable in YAML discovery

PipoCanaja:
This PR allows the use of an offset in both discover_sensor and in the YAML file definition (including in the SQL sensor table), that will allow this calculation to occur without using PHP like we have to now: 
$current = ($value / $divisor * multiplier) + $offset;

Necessary for devices that report temperature as such : 
myTemp : Integer from 0 to 1000, in tenth of Celcius, with origin at -30C
- INT: 0 -> -30C
- INT: 10 -> -29C
- INT: 300 -> 0C

To solve this, with the $offset, one would do : 
```
 value: myTemp
 divisor: 10
 const: 30
```

- All tests need to be rerun (cause the table sensors is upgraded) or const should be ignored until tests are re-run.
- This patch being merged will impact all PR waiting in the pipe. 
- I have a few devices (PacketLight) that use extensively this and PR will come when this one is evaluated and merged. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Measuring temperature like that seems very confusing.  Why would you offset it by 30 degrees?

PipoCanaja:
Basically, they don't want to send negative values. They do the same with Gain (dB) which is also translated to be zero originated.

PipoCanaja:
@murrant 
In term of code, this change does the magic needed. 
- Now we need to decide if this new variable is good to go (conceptually).
- And if yes if the naming is good (could be "adder" also, for instant). 

If all this goes to a validated concept, then I could either: 
- Rerun all tests that have "sensors" items (meaning they will lack the "adder" line in them).
- Filter this new item from the validation for a while (seems a bad idea but it is still a choice :) 

Let me know.

murrant:
I think the code looks good to be, but how about the name "offset"?  Also, it might be important to note in the docs whether it is applied before or after multiplier/divisor.

PipoCanaja:
'offset' seems good indeed. I'll change that and emphasize the priorities in the computation of the value.

PipoCanaja:
Doing a few tests now that we have the proper naming. In the meantime, what would you suggest for the test ? any "simple" way to run all the ones that must be run again ?

murrant:
./save-test-data.php -m sensors

PipoCanaja:
I already asked you that, and you already reply ... I have to write it somewhere I suppose if I can't remember :)

murrant:
Shoot.  I've been thinking about this and it seems like it would only apply to one os.  Is that right?
It would probably be better to just have custom polling for that. os.

murrant:
Didn't mean to close

PipoCanaja:
That's also a point. I identified PacketLight (optical devices) which has this behaviour. Other option would be to create a user_func to do the maths. What would be nice in that case would be to be able to create a user_func in the device files (instead of mixing it in the general files). 
What do you think ?

PipoCanaja:
A new PR is on the way, idea is to accept inline anonymous function into the YAML user_func: element. 
We can close this I suppose.

murrant:
you are working to hard on this one.  Just make packetlight.inc.php for the affected sensors.

PipoCanaja:
I would like to try to stick with yaml and the inline function works well. 
Let me know how you evaluate it. This would have the advantage of being useful in more OS than the offset approach of this PR. 
And I am not in a hurry with the packet light devices.

PipoCanaja:
PR #10121

PR #10054, opened 2019-04-01
Reduce default thread count on small systems

brianegge:
Default script options kills my raspberrypi.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/snmp-scan.py
+++ b/snmp-scan.py
@@ -28,7 +28,7 @@ import argparse
 from argparse import RawTextHelpFormatter
 import json
 from collections import namedtuple
-from multiprocessing import Pool
+from multiprocessing import Pool, cpu_count
 from os import path, chdir
 from socket import gethostbyname, gethostbyaddr, herror, gaierror
 from subprocess import check_output, CalledProcessError
@@ -57,7 +57,7 @@ class Outcome:
 
 
 VERBOSE_LEVEL = 0
-THREADS = 32
+THREADS = min(cpu_count(), 32)
 CONFIG = {}
 EXCLUDED_NETS = []
 start_time = time()


brianegge:
It's a "run twice" script. The first time you run it your machine fails, the second time you remember to specify a thread count. 

Agreed that process is not really CPU bound. It's a memory issue more than CPU, but estimating the memory required and memory available is more difficult. A better estimate can be done with:
```mem = psutil.virtual_memory
  free = mem.free + mem.cached
  threads = free / _memory_per_fork```
But you have to know the memory needed per fork. 

The command line switches only help before you have an issue. If you know in advance you are scanning a large network and have the resources, then it's easy to scale up. Otherwise the default would be safer and more stable on most systems.

murrant:
I'm not denying this is a pain.  I just think you need a solution that doesn't punish most of our users with slower runtimes.  Many won't look to see that there is even a tunable and just wait for it to run.

PipoCanaja:
May be you should match the RaspberryPi (or a little bit better, ARM CPUs) directly in the script, and only apply this thread limitation in that case. I cannot see any more "generic" way for a formula that would not punish bigger CPUs by just looking at standard values in cpuinfo.

murrant:
He said the main issue is memory, not cpu actually.

laf:
I'd vote to keep this at a high value to cover the majority of users.

Could add a config.php option to allow users to set a default?

brianegge:
How many cores do most users have? I have 28-32 on my servers but only 4 on my rpi. 

I could change this to estimate based on your available memory. 

-Brian

murrant:
Yes, changing this to check the actual thing that constrains it would help a lot.

murrant:
Feel free to re-open if you can come up with a good way to limit the default thread count.

PR #10056, opened 2019-04-01
Added PPPoE Sessions Graph for Mikrotik

neszt:
Added PPPoE Sessions Graph for Mikrotik. Cli command to get session count: /interface pppoe-server print count-only

Coding style and implement idea from 9f553af71e2f4bea70a3a9a1002ecaaa0969d765.

Tested on Mikrotik Firmware version 6.23-6.44, but should work with all firmware versions.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/includes/graphs/device/routeros_pppoe_sessions.inc.php
+++ b/html/includes/graphs/device/routeros_pppoe_sessions.inc.php
@@ -1,0 +1,19 @@
+<?php
+
+$rrd_filename = rrd_name($device['hostname'], 'routeros_pppoe_sessions');
+
+require 'includes/graphs/common.inc.php';
+
+$ds = 'pppoe_sessions';
+
+$colour_area = '99cc99';
+$colour_line = '00cc00';
+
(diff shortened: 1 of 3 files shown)


PipoCanaja:
Hi @neszt 
Thanx for your contribution. I tested this and it works as expected on the device I have here (CHR, 6.42.1). 

The only thing you might be interested in is to store the value in the DB, for alerting purposes, cause you cannot alert on values that are only graphed with RRD files. 
You could re-implement this in YAML in ```includes/definitions/discovery/routeros.yaml``` and use the type "count", and you would then get the value stored in the DB.

If you don't need any alerting, your implementation is perfectly fine.

neszt:
Hello, @PipoCanaja,

Thank you for your quick answer. I'm glad to hear you tested it and works as expected.

I don't need any alerting.

Regards,

PipoCanaja:
So it looks good to me ! Let's merge it.

PR #10059, opened 2019-04-02
Added initial support for teltonika

tomarch:
Hi,

First PR for support TELTONIKA products.
Works for RUT95X which is LTE router.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hi @tomarch 
Thanx for your contribution. 
Could you please rename the mib files to match the MIB official name ?

tomarch:
This is the name given by Teltonika.
Both file refer to "TELTONIKA-MIB".
Which name can i set ?

PipoCanaja:
Only 1 file can have a name so yes, you need to remove the not necessary one (it won't be used by snmpwalk anyway). 
And for devices using the 2nd one, you have to create a 2nd OS and a 2nd mib directory.

tomarch:
Ok, thanks.

PipoCanaja:
Bonjour ! 
Looks good. Here is the test data. Let's run the tests this and if all good, will be ready to merge. 
++

tomarch:
seems to be all good. 
Thx

PipoCanaja:
rutos.yaml is not good. 

Please rerun ```./script/save-test-data.php -o rutos ``` and commit the resulting file.

PipoCanaja:
Bonjour,
It seems that, for some reason, when you run tests, it does not match the correct OS on the data. Are you running the test in the correct branch of your dev server? 
Basically, you have to fix the json file and check on TravisCI result, which has to pass. 
++

spencerbutler:
@tomarch This can be a tricky one to figure out. I typically run my tests like [this](https://gist.github.com/spencerbutler/01c3d5891d7e2791fb3d3df9cabceb88#file-run-lnms-tests-sh) to avoid the "data doesn't match!" issue. Most of the time, it is cache/os_defs.cache ruining your game. Good luck, and hop on discord with any questions.

murrant:
This happened because the binary png files don't come through in patches...

PR #10070, opened 2019-04-04
Refactor Api transport to use Guzzle (and new variables syntax)

PipoCanaja:
The goal is to see how we can update API from using: 
```
%hostname
%title
```
to Laravel Blades {{ $xxx }}

Then, code was adapted to Guzzle and URL parameters were split (which solved a lot of urlencode issues)

Then, Laravel Blades were removed (security issue) but ney syntax kept. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/Alert/Transport/Api.php
```diff
@@ -52,10 +53,23 @@ private function contactAPI($obj, $api, $method)
     {
         $method = strtolower($method);
         list($host, $api) = explode("?", $api, 2);
+        $api = trim($api);
+
         foreach ($obj as $k => $v) {
-            $api = str_replace("%" . $k, $method == "get" ? urlencode($v) : $v, $api);
+            $api = str_replace("{{ $" . $k . " }}", $method == "get" ? urlencode($v) : $v, $api);
```
PipoCanaja: This is an adaptation of the old str_replace but using new syntax for blades
- looks similar to the other Blade based templates but is not at all one. 
- allow proper urlencode to take place for GET

LibreNMS/Alert/Transport/Api.php
```diff
@@ -52,10 +53,23 @@ private function contactAPI($obj, $api, $method)
...
+        }
+
+        try {
+            $api = view(['template' => $api], $obj)->__toString();
```
PipoCanaja: This is the Blades way of doing it, but this one does not urlencode in case method is GET

LibreNMS/Alert/Transport/Api.php
```diff
@@ -100,7 +114,7 @@ public static function configTemplate()
             ],
             'validation' => [
                 'api-method' => 'in:GET,POST',
-                'api-url' => 'required|url'
+                'api-url' => 'required|string' //we need to allow special chars & spaces if GET
```
PipoCanaja: We need to change this cause api-url would not pass validation with spaces in it.

murrant:
I think you need a slightly different approach.
Treat each key/value pair individually.
Also, try Guzzle or something instead of curl ;)

PipoCanaja:
So I do it, if we have a GET, on the url provided by the user AND the keys AND the values. Then I do the replacement. Sounds like a plan ! 
I'll have a look at guzzle. Seems that it is already available so nothing to change in composer part.

PipoCanaja:
OK. Now we have the Blade variant which processed only the values of each keys. The config of the API changed. 
It does break all the existing setups, but allows all the power of the templates to occur on every single parameter. 

With the following parameters : 
```
Method: GET
URL: http://myServer/test.php
Options: 
  msg1={{ $title }}
  msg2={{ $title }}

```
We get this in the logs of the WWW server receiving the API
`1.2.235.36 - - [17/Apr/2019:16:07:28 +0200] "GET /test.php?msg1=Alert%20for%20device%20snmpsim7%20-%20Devices%20up%2Fdown&msg2=Alert%20for%20device%20snmpsim7%20-%20Devices%20up%2Fdown HTTP/1.1" 200 100 "-" "-"
`

Works the same with post with proper URL encode.

PipoCanaja:
@murrant 
What do you think ? 
When the data processing part will validated, I'll check the Guzzle part.

PipoCanaja:
Added the Breaking-change tag cause the current API config will not work OK after this change. 
Other option would be to create a new API transport and delete the old after people do the migration. 
@librenms/reviewers Please advise on this.

LibreNMS/Alert/Transport/Api.php
PipoCanaja: Blades are now in use with the new config options.

LibreNMS/Alert/Transport/Api.php
PipoCanaja: solved now.

LibreNMS/Alert/Transport/Api.php
PipoCanaja: solved as well

PipoCanaja:
@murrant Guzzle is now in use for both GET and POST. Guzzle takes care of all urlencoding that is needed. Basic auth is also added now. 
Open points : 
- Should I create a new transport or upgrade the existing API transport as done in this PR ? That would avoid the "breaking change" aspect and would give time to users to change from old to new API.

laf:
Would it not be easier to just post all the alert data as a json encoded values or just as key=value in the post data so people can access whatever they want programatically?

PipoCanaja:
I am not sure I get it right. You would send all the alert data to the url provided ? The problem is that the user does not necessarily control the code at the other end. 
Like if we use it to send text messages via some telco provider api, I need to feed the exact variables this provider expects. That’s why I ended up with this key=value structure (inspired from another transport)

laf:
> The problem is that the user does not necessarily control the code at the other end.

Good point.

PipoCanaja:
From a feature point of view, it seems that this new version covers the 'old' ApiTransport  and adds more capabilites. 

Before merging, we must decide:
- Do we keep it with current name or do we create a new API Transport to allow users to move from old to new ?
- Define the text to inform the users (whatever solution we choose, they need to know)

murrant:
As stated before, I think we should not be using Blade for this.  But some template engine that cannot execute any php code.

PipoCanaja:
What would be the reasons to forbid running any php code here ? I don't have any usecase, but I like the fact that I have this possibility, and I also like the fact that it has exactly the same capabilities as the alert templates etc etc.

murrant:
Security.  Changing the template engine is not something that can be done easily.  I'd rather only do it once.

PipoCanaja:
Ok. I'll revert to a lighter variable matching, but using the {{ }} syntax. And ensure that characters are properly escaped. That will be safer.

PipoCanaja:
Got back to the original matching with str_replace, but newer {{ $xxx }} syntax. Works well with both GET and POST requests, urlencode is entirely left to Guzzle, using a list of URL options. Even tried funny characters like accents and it seems to behave correctly. 

That should make Api transport usable again, and more reliable.

LibreNMS/Alert/Transport/Api.php
```diff
@@ -25,42 +26,57 @@
...
-        list($host, $api) = explode("?", $api, 2);
-        foreach ($obj as $k => $v) {
-            $api = str_replace("%" . $k, $method == "get" ? urlencode($v) : $v, $api);
+        $host = explode("?", $api, 2)[0]; //we don't use the parameter part, cause we build it out of options.
+
+        //get each line of key-values and process the variables;
+        foreach (preg_split("/\\r\\n|\\r|\\n/", $options) as $current_line) {
+            list($u_key, $u_val) = explode('=', $current_line, 2);
+            foreach ($obj as $p_key => $p_val) {
+                $u_val = str_replace("{{ $" . $p_key . ' }}', $p_val, $u_val);
```
murrant: Do we need to escape the values?  I don't think so, they should be escaped by guzzle later...

LibreNMS/Alert/Transport/Api.php
```diff
@@ -25,42 +26,57 @@
...
         $method = $this->config['api-method'];
-        return $this->contactAPI($obj, $url, $method);
+        $auth = [$this->config['api-auth-username'], $this->config['api-auth-password']];
+        return $this->contactAPI($obj, $url, $options, $method, $auth);
     }
 
-    private function contactAPI($obj, $api, $method)
+    private function contactAPI($obj, $api, $options, $method, $auth)
     {
         $method = strtolower($method);
```
murrant: $request_opts and $query are never initialized

LibreNMS/Alert/Transport/Api.php
```diff
@@ -25,42 +26,57 @@
...
-        list($host, $api) = explode("?", $api, 2);
-        foreach ($obj as $k => $v) {
-            $api = str_replace("%" . $k, $method == "get" ? urlencode($v) : $v, $api);
+        $host = explode("?", $api, 2)[0]; //we don't use the parameter part, cause we build it out of options.
```
murrant: $request_opts and $query are never initialized

LibreNMS/Alert/Transport/Api.php
PipoCanaja: Nope, Guzzle does it so I only str_replace the values here, no escaping.

LibreNMS/Alert/Transport/Api.php
PipoCanaja: Corrected

murrant:
Looks good, just needs some documentation and notification :)

PipoCanaja:
Doc is already added, with example. And notification was added just now to the rss file (means that when the user sees it, it is already too late and they may have missed alerts. Should we push it the other way so they get it before merge ? ).

PipoCanaja:
@murrant Anything else to prepare for this PR ? Been out of town for a while, but on my way back to work now :)
I had a doubt on how we notify ? (do we notify before merge or do we merge AND notify at the same time ? )

LibreNMS/Alert/Transport/Api.php
```diff
@@ -83,11 +104,30 @@ public static function configTemplate()
...
+                    'name' => 'api-auth-password',
+                    'descr' => 'Auth Password',
+                    'type' => 'password',
                 ]
             ],
             'validation' => [
                 'api-method' => 'in:GET,POST',
-                'api-url' => 'required|url'
+                'api-url' => 'required|url',
+                'api-options' => 'required|string'
```
murrant: api-options probably should not be required

murrant:
I'm not so sure you need a notification.  Do people that already have it defined still work?

PipoCanaja:
> I'm not so sure you need a notification. Do people that already have it defined still work?

Unfortunately, yes :D With the right settings, current code can send text messages to a compatible provider.

LibreNMS/Alert/Transport/Api.php
PipoCanaja: Correct, I'll change that.

murrant:
If you want to notify ahead of merging.  (which I think we should) Add it to the RSS feed on the librenms.org site.
If you want to notify when/after it is merged (for monthly it will be when they update).  Add it to the RSS here.

PipoCanaja:
Ok @murrant . You can now review the notification at https://github.com/librenms/librenms.github.io/pull/117 and commit it. I announced the merge for #10070 on Wednesday. If you feel this is too short, feel free to change the notification.

murrant:

https://community.librenms.org/t/new-guzzlehttp-does-not-honor-proxy-settings/9908/1

PR #10072, opened 2019-04-05
Add Gotify Transport

ToeiRei:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Gotify.php
+++ b/LibreNMS/Alert/Transport/Gotify.php
@@ -48,24 +48,15 @@ class Gotify extends Transport
     {
         switch ($obj['severity']) {
             case "critical":
-                $data['priority'] = 1;
-                if (!empty($api['options']['sound_critical'])) {
-                    $data['sound'] = $api['options']['sound_critical'];
-                }
+                $data['priority'] = 10;
                 break;
             case "warning":
-                $data['priority'] = 1;
-                if (!empty($api['options']['sound_warning'])) {
-                    $data['sound'] = $api['options']['sound_warning'];
-                }
+                $data['priority'] = 5;
                 break;
         }
         switch ($obj['state']) {
             case 0:
                 $data['priority'] = 0;
-                if (!empty($api['options']['sound_ok'])) {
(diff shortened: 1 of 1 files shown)


LibreNMS/Alert/Transport/Gotify.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+        $gotify_server = $this->config['gotify_server'];
+        $gotify_token  = $this->config['gotify_token'];
+        return $this->contactGotify($obj, $gotify_server, $gotify_token);
+    }
+
+    private function deliverAlertOld($obj, $opts)
+    {
+
+        $this->contactGotify($obj, $gotify_server, $gotify_token);
```
PipoCanaja: This function uses $gotify_server and token without defining them. 
But as this is the "legacy" way of sending alerts (without the transport structure) I would say we could just avoid this function and just return false (should not be used anymore). 
@murrant : What do you think ?

murrant: Yeah, a lot of older code was structured that way because it was modified. (I started a PR to change all that, but never sent it).  You can delete deliverAlertOld and any use of it.

laf:
@ToeiRei Would you mind adding some documentation on how to use this for people then we get the code cleaned up and merged in.

ToeiRei:
@laf There isn't much documentation to add:
The URL of the server needs to be added, then in Gotify you'd add an app and fill in its token here.

laf:
> @laf There isn't much documentation to add:
> The URL of the server needs to be added, then in Gotify you'd add an app and fill in its token here.

@ToeiRei Maybe but it's still needed :)

LibreNMS/Alert/Transport/Gotify.php
ToeiRei: I used some existing transport and it had it. I'm completely open to corrections here

Jellyfrog:
@ToeiRei any news?

ToeiRei:
tbh, I dropped gotify and went to use some webhooks on my home automation as gotify caused problems here.

PR #10074, opened 2019-04-06
Use $_SERVER['HOST_NAME'] instead of $_SERVER['SERVER_NAME']

Thermi:
Use $_SERVER['HOST_NAME'] instead of $_SERVER['SERVER_NAME'] to make the webinterface… reachable over several domain names. This didn't work previously, even if the HTTP server was configured with several server names that pointed at librenms.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/defaults.inc.php
+++ b/includes/defaults.inc.php
@@ -80,12 +80,12 @@ $config['rrd_rra'] .= ' RRA:LAST:0.5:1:1440 ';
 // $config['rrdcached']    = "unix:/var/run/rrdcached.sock";
 
 // Web Interface Settings
-if (isset($_SERVER['HOST_NAME']) && isset($_SERVER['SERVER_PORT'])) {
-    if (strpos($_SERVER['HOST_NAME'], ':')) {
+if (isset($_SERVER['HTTP_HOST']) && isset($_SERVER['SERVER_PORT'])) {
+    if (strpos($_SERVER['HTTP_HOST'], ':')) {
         // Literal IPv6
-        $config['base_url'] = 'http://['.$_SERVER['HOST_NAME'].']'.($_SERVER['SERVER_PORT'] != 80 ? ':'.$_SERVER['SERVER_PORT'] : '').'/';
+        $config['base_url'] = 'http://['.$_SERVER['HTTP_HOST'].']'.($_SERVER['SERVER_PORT'] != 80 ? ':'.$_SERVER['SERVER_PORT'] : '').'/';
     } else {
-        $config['base_url'] = 'http://'.$_SERVER['HOST_NAME'].($_SERVER['SERVER_PORT'] != 80 ? ':'.$_SERVER['SERVER_PORT'] : '').'/';
+        $config['base_url'] = 'http://'.$_SERVER['HTTP_HOST'].($_SERVER['SERVER_PORT'] != 80 ? ':'.$_SERVER['SERVER_PORT'] : '').'/';
     }
 }
 
--- a/includes/defaults.inc.php
+++ b/includes/defaults.inc.php
@@ -80,12 +80,12 @@ $config['rrd_rra'] .= ' RRA:LAST:0.5:1:1440 ';
 // $config['rrdcached']    = "unix:/var/run/rrdcached.sock";
 
 // Web Interface Settings
-if (isset($_SERVER['SERVER_NAME']) && isset($_SERVER['SERVER_PORT'])) {
-    if (strpos($_SERVER['SERVER_NAME'], ':')) {
+if (isset($_SERVER['HOST_NAME']) && isset($_SERVER['SERVER_PORT'])) {
+    if (strpos($_SERVER['HOST_NAME'], ':')) {
         // Literal IPv6
-        $config['base_url'] = 'http://['.$_SERVER['SERVER_NAME'].']'.($_SERVER['SERVER_PORT'] != 80 ? ':'.$_SERVER['SERVER_PORT'] : '').'/';
+        $config['base_url'] = 'http://['.$_SERVER['HOST_NAME'].']'.($_SERVER['SERVER_PORT'] != 80 ? ':'.$_SERVER['SERVER_PORT'] : '').'/';
     } else {
-        $config['base_url'] = 'http://'.$_SERVER['SERVER_NAME'].($_SERVER['SERVER_PORT'] != 80 ? ':'.$_SERVER['SERVER_PORT'] : '').'/';
+        $config['base_url'] = 'http://'.$_SERVER['HOST_NAME'].($_SERVER['SERVER_PORT'] != 80 ? ':'.$_SERVER['SERVER_PORT'] : '').'/';
     }
 }
 


includes/defaults.inc.php
```diff
@@ -80,12 +80,12 @@
 // $config['rrdcached']    = "unix:/var/run/rrdcached.sock";
 
 // Web Interface Settings
-if (isset($_SERVER['SERVER_NAME']) && isset($_SERVER['SERVER_PORT'])) {
-    if (strpos($_SERVER['SERVER_NAME'], ':')) {
+if (isset($_SERVER['HOST_NAME']) && isset($_SERVER['SERVER_PORT'])) {
```
murrant: $_SERVER['HOST_NAME'] is not a valid variable, so you have effectively deleted all this code afaict.  What do the different variables contain on your server?

Thermi: Right, sorry. It seems I didn't quite pay attention when copying my tested code over to my cloned repo. I used HTTP_HOST, not HOST_NAME.

murrant:
I wonder how this will work when a reverse proxy is in play?

Thermi:
$_SERVER['HTTP_HOST'] contains the Host header's value. The HTTP server that serves librenms' pages via e.g. a CGI backend needs that header to find the right configuration of its own vhosts. So now the only question is if access without a Host header works or is supported by librenms anyway or if this change breaks that, if it was supported.

Thermi:
Currently, with $_SERVER['SERVER_NAME'], using several server names for a librenms vhost doesn't work. It seems $_SERVER['SERVER_NAME'] resolves to the first element of the list by default for some reason. Thus, all URLs created using base_url point to the first, different server name. If that name isn't accessible for some reason, the elements using that base_url can't be loaded.

murrant:
I mean, it is dubious that LibreNMS tries to set the base_url at all... and at worst a security risk.

Thermi:
Evidently, in my setup with 1.50, it does, because accessing librenms via the second server name doesn't work with the existing code. With my changes, it does.

murrant:
@Thermi You could just set $config['base_url'] in your config.php file as intended.

But I mean I can't figure out why some one would have added code automatically setting base_url in the first place.

Thermi:
If I did that, then I get exactly the problem that I try to solve: Librenms won't work correctly, because the addresses that are generated using base_url point to domains that the client can't access (because they're in, for example, a private domain space).

Thermi:
To fix this, either this code needs to be merged, or all code using base_url needs to be altered to produce relative paths only.

Thermi:
And anyway, with the wrong URLs in the code, the browser won't load those resources because of security policies preventing cross loading of resources.

murrant:
That's true, I'm just thinking about this code in general since this comes in to focus.

(Also, I don't understand why you couldn't set base_url correctly)

Thermi:
(Because I want to access librenms using two different FQDNs.)

murrant:
I see, not sure I would recommend that.  Usually I just use loopback NAT or separate DNS records to solve internal access.  But that is not relevant to this PR.  Thanks for explaining.

Thermi:
Why would you not recommend that? The problem is that the HTTP server with PHP and this code together won't produce valid URLs referencing the graphics and graphs. No amount of NAT or different DNS records (which is what I do here anyway for this librenms installation. In fact, it is the very source of the problem, together with this code here). No amount of NAT would prevent this, also. In my case, I have a management network and a public network. Hosts on one of them aren't necessaribly able to access the other.

murrant:
I'm fine with this code, but I would like some others to test it. (especially ones with peculiar setups)

murrant:
@Thermi no, that doesn't fix the issue with the code.  That fixes the need for two domain names :)

In the long run, the code should be made to run with urls that don't contain the domain/ip unless base_url is explicitly specified.  This short term fix is fine with me and will fix the issue when people don't properly set server_name in their webserver config and the urls to css, etc are wrong.

Thermi:
Then only cases where the Host header is not set need to be tested. Previously, when $_SERVER['SERVER_NAME'] was used, then at least that variable was always set, because the HTTP server set it. But is that particular use case supported anyway?

Well, the current code in the repo would work if the HTTP server always sent the right SERVER_NAME that matches the HTTP client's indicated Host (in the Host header), but at least nginx 1.14.2 /doesn't/ do that correctly and seems to send the first server_name of the vhost instead. What could work instead, would be the use of two completely different vhosts with one server_name each. I guess then that would work with nginx correctly. I don't know about Apache. We could work around that case in this branch, too, by checking if SERVER_NAME equals HTTP_HOST, if HTTP_HOST is set. Setting HTTP_HOST to IP addresses is valid though, so any client, except probably curl invoked from the command line, does that by default anyway.

laf:
Need to keep in mind that `HTTP_HOST` is user controlled. Not that I think we rely on the value for anything security related it's still worth understanding - do we need to allow users to specify whitelist of domains/ips?

Thermi:
No, we don't need to do that. The web server already matches HTTP Host headers to vhosts and thus controls access. Other than that, it's not supposed to be a security mechanism anyway. Right now, the web UI simply doesn't show graphics, but still allows all AJAX calls (by virtue that this is a bug, not a feature) if the base_url points to another domain than what is visited by the browser.

laf:
> No, we don't need to do that. The web server already matches HTTP Host headers to vhosts and thus controls access.

Only if it's not a default vhost on the box. A default vhost will accept anything afaik

Thermi:
Great, if it's not configured and thus the server name isn't equal to the Host header, the UI is broken anyway.

laf:
> Great, if it's not configured and thus the server name isn't equal to the Host header, the UI is broken anyway.

Maybe but we're trying to avoid anything being broken hence the discussion around this. We've merged things in before without thinking through the full impact and been burned.

Thermi:
> Maybe but we're trying to avoid anything being broken hence the discussion around this. We've merged things in before without thinking through the full impact and been burned.

I tested it. It is broken. I ran this code for the last 3 days and everything still works. I also checked all the code that uses it and as far as I can tell, the change is fine.

Up to now, the points brought up were nought because they weren't pertaining anything that works right now.  The comment in `/tests/config/config.test.php` explains that a "wrong" base_url makes the UI unusable. By using the Host header to build the URLs that use base_url, all those built URLs always reference the site that the web browser has requested, not a third page. When using the server name, the server name in the web server's config is used, which can be different from the Host header. Thus, in that case, all URLs built using the base_url would be wrong and the browser would not load those.

laf:
> I tested it. It is broken. I ran this code for the last 3 days and everything still works. I also checked all the code that uses it and as far as I can tell, the change is fine.
> 
> Up to now, the points brought up were nought because they weren't pertaining anything that works right now. The comment in `/tests/config/config.test.php` explains that a "wrong" base_url makes the UI unusable. By using the Host header to build the URLs that use base_url, all those built URLs always reference the site that the web browser has requested, not a third page. When using the server name, the server name in the web server's config is used, which can be different from the Host header. Thus, in that case, all URLs built using the base_url would be wrong and the browser would not load those.

I totally get that - stop getting defensive, this is merely us asking questions to avoid issues

Thermi:
Hello,

What is the status on this?

Kind regards

Noel

murrant:
No one has confirmed that your changes are correct.  I plan to keep removing base_url usage from the code base.

PR #10087, opened 2019-04-10
added initial support for cyberpower-ups

tigerdjohnson:
DO NOT DELETE THIS TEXT

#### Please note

Screenshot of device after changes:
Screenshot_2019-04-10 192 168 127 132 - LibreNMS

I'm new to github, so hopefully this goes smoothly... :)

#### Testers


includes/definitions/cyberpower.yaml
```diff
@@ -1,13 +1,13 @@
...
+text: 'Cyberpower PDU'
 type: power
 icon: cyberpower
 over:
     - { graph: device_current, text: Current }
     - { graph: device_voltage, text: Voltage }
 discovery:
     - sysObjectID:
-        - .1.3.6.1.4.1.3808
+        - .1.3.6.1.4.1.3808.1.1.3
```
PipoCanaja: This is probably not a good idea. You should exclude ```1.3.6.1.4.1.3808.1.1.1``` from the existing cyberpower (because you handle it into the new cyberpower-ups.yaml) and leave the all the other products to be matched by cyberpower.yaml.

PipoCanaja:
Hello @tigerdjohnson 
Thanx for your first contribution ! I added a change request. The idea is that the existing "cyberpower" description YAML file should continue match all the Cyberpower devices, except the one you create as cyberpower-ups. Doing so will ensure that all the other products keep getting their Cyberpower logo etc etc.

To exclude, you could do this, which means "start with .1.3.6.1.4.1.3808" AND "does not start with .1.3.6.1.4.1.3808.1.1.1". Please not that the ```'-'``` are very important to acheive the AND. 
```
discovery:
    -
      sysObjectID:
          - .1.3.6.1.4.1.3808
      sysObjectID_except:
          - .1.3.6.1.4.1.3808.1.1.1
```

includes/definitions/cyberpower.yaml
tigerdjohnson: @PipoCanaja  Ah, excellent point. I'll revert my previou changes and add the exception

includes/polling/os/cyberpower-ups.inc.php
```diff
@@ -0,0 +1,18 @@
...
+//serial   = UPS serial number as reported by RMCARD
+//version  = Firmware revision of RMCARD
+
+$cyberpower_oids = [
+    'hardware' => '.1.3.6.1.4.1.3808.1.1.1.1.1.1.0',
+    'version'  => '.1.3.6.1.4.1.3808.1.1.1.1.2.4.0',
+    'serial'   => '.1.3.6.1.4.1.3808.1.1.1.1.2.9.0',
+];
+
+$returned_oids = snmp_get_multi_oid($device, $cyberpower_oids, '-OUQn', 'CPS-MIB');
```
PipoCanaja: You could use the textual form of the OID here, instead of the numerical.

includes/discovery/sensors/voltage/cyberpower-ups.inc.php
```diff
@@ -0,0 +1,41 @@
...
+    $current = $value[1]/$item['divisor'];
+    $current_oid = $value[0];
+
+    if (!empty($oids)) {
+        d_echo($oids);
+        $oids = trim($oids);
+        if ($oids) {
+            echo $item['type'].' '.$item['mib'];
+        }
+        discover_sensor($valid['sensor'], 'voltage', $device, $current_oid, $current_oid, $item['type'], $item['descr'], $item['divisor'], 1, null, null, null, null, $current);
```
PipoCanaja: This part of the discovery seems to be doable via the YAML file as far as I see it. Did you try this approach (and revert it for a reason ?)

includes/discovery/sensors/load/cyberpower-ups.inc.php
```diff
@@ -0,0 +1,30 @@
...
+    $current_oid = $value[0];
+
+    if (!empty($oids)) {
+        d_echo($oids);
+        $oids = trim($oids);
+        if ($oids) {
+            echo $item['type'].' '.$item['mib'];
+        }
+        discover_sensor($valid['sensor'], 'load', $device, $current_oid, $current_oid, $item['type'], $item['descr'], $item['divisor'], 1, null, null, null, null, $current);
+    }
```
PipoCanaja: This part of the discovery seems to be doable via the YAML file as far as I see it. Did you try this approach (and revert it for a reason ?)

includes/discovery/sensors/voltage/cyberpower-ups.inc.php
tigerdjohnson: @PipoCanaja I just realized this is all happening because the UPS is not responding to snmpbulkwalk polls. Setting nobulk: true in cyberpower.yaml resolves the issue w/o need for further modifications. We should close out this PR. :)

tigerdjohnson:
Support already exists for CyberPower UPS. Thought support was initially not there, but turned out remote management cards have bugged snmpbulkwalk. Can disable bulkwalk in cyberpower.yaml to resolve the issue w/o need for additional changes.

murrant:
@tigerdjohnson you can disable that in config.php to make sure you don't break updates.

PR #10089, opened 2019-04-10
Correct api syntax, Update Logs.md

stevenpinewebair:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/API/Logs.md
+++ b/doc/API/Logs.md
@@ -11,8 +11,8 @@ Input:
 
   - start: The page number to request.
   - limit: The limit of results to be returned.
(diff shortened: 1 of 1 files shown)


murrant:
Funnily enough, the docs format used to work.  Looks like some over-zealous sql injection protection broke it.  It could be made to accept any date format MySQL supports again, should we do that?

stevenpinewebair:
Personally, I am using the timestamp as an approximation for finding the corresponding log. Ideally the api alert id endpoint would give me the log id which i could query directly.

if there is a better way to do this, please let me know.

PR #10094, opened 2019-04-11
Adva SNMP Trap Handlers

h-barnhart:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Adds some traps handles for Adva Ethernet devices. Includes unit tests for the handlers.


LibreNMS/Snmptrap/Handlers/AdvaAccThresholdCrossingAlert.php
```diff
@@ -0,0 +1,53 @@
...
+     * Data is pre-parsed and delivered as a Trap.
+     *
+     * @param Device $device
+     * @param Trap $trap
+     * @return void
+     */
+    public function handle(Device $device, Trap $trap)
+    {
+        $raw = $trap->getRaw();
+        Log::event($raw, $device->device_id, 'trap', 2);
```
murrant: Dumping the entire trap into the eventlog in addition to the event below might be too much noise.  (could this be debug code?)

murrant: Dumping the entire trap into the eventlog in addition to the event below might be too much noise. (could this be debug code?)

LibreNMS/Snmptrap/Handlers/AdvaAttributeChange.php
```diff
@@ -0,0 +1,239 @@
...
+{
+    /**
+     * Handle snmptrap.
+     * Data is pre-parsed and delivered as a Trap.
+     *
+     * @param Device $device
+     * @param Trap $trap
+     * @return void
+     */
+    public function handle(Device $device, Trap $trap)
```
murrant: This function is very long and hard to read.  Moving some of the code to private functions would help a lot.  Like the first if statement calling:
```php
if ($trap->findOid('CM-SYSTEM-MIB::sysLog')) {
   $this->handleSyslog($trap);
} elseif ($trap->findOid('CM-SYSTEM-MIB::aclEntry')) {
```

LibreNMS/Snmptrap/Handlers/AdvaAccThresholdCrossingAlert.php
h-barnhart: I'm was wanting to submit this now under the idea of something is better than nothing, then circle back and improve it later. If you prefer I can remove it from this pull request and add it at a later time. I'd like to get the other traps included as soon as possible as coding out all cases for this trap might take some time.

This is an interface threshold trap that fires on a number user defined thresholds. For example it will send a trap if x number of 64 byte packets are received on an interface or if the interface is up but not passing any traffic. There are 48 thresholds that are monitored at 15 minute and 1 day intervals. So this particular handler has 96 possible cases. Additionally this is only for "access port" interfaces, there is a similar handler for "network port" interfaces. Hopefully I can do one and do a find and replace on the OID names, but it will take a bit to work.

How would you like to proceed?

LibreNMS/Snmptrap/Handlers/AdvaAttributeChange.php
h-barnhart: Will do.

h-barnhart: Ok, I'll go ahead and do that for all the sections. Some of the sections I pulled back the amount of information into a generic "this was modified message". Splitting out into more functions will allow me or someone else go back and make the traps more informative again if desired.

LibreNMS/Snmptrap/Handlers/AdvaAccThresholdCrossingAlert.php
h-barnhart: Ok, I should probably have my coffee, read my code, then respond. Yep, there was debugging still in there. Going to fix that.

h-barnhart: Fixed the debug and added the requested functionality.

LibreNMS/Snmptrap/Handlers/AdvaAccThresholdCrossingAlert.php
```diff
@@ -43,11 +44,97 @@ class AdvaAccThresholdCrossingAlert implements SnmptrapHandler
...
         $ifName = $trap->getOidData($trap->findOid("IF-MIB::ifName"));
+        $thresholdMessage = $this->handleThreshold($trap);
+        Log::event("$ifName $thresholdMessage threshold exceeded for $interval", $device->device_id, 'trap', 2);
+    }
+    public static function handleThreshold($trap)
+    {
         $threshOid = $trap->getOidData($trap->findOid("CM-PERFORMANCE-MIB::cmEthernetAccPortThresholdVariable"));
-        Log::event("$ifName threshold exceeded for $interval. Threshold OID is $threshOid.", $device->device_id, 'trap', 2);
+        switch (true) {
+            case strpos($threshOid, 'CM-PERFORMANCE-MIB::cmEthernetAccPortStatsUAS') === 0:
```
murrant: FYI, we have a function called str_contains() it is a little nicer than strpos :)

h-barnhart: Made requested changes.

h-barnhart: Not sure why the changes you made are causing errors. Functionally it all looks the same, just a few changes to where things are called.

h-barnhart: I this is some kind of scope issue that I don't understand (new to php). If I call getThresholds from withing the handle() method and pass it to getThresholdMessage method it works. Would that be acceptable?

h-barnhart: Found and fixed the issue.

LibreNMS/Snmptrap/Handlers/AdvaNetThresholdCrossingAlert.php
```diff
@@ -0,0 +1,125 @@
...
+    public function getThresholdMessage($thresholdOid)
+    {
+        foreach ($this->getThresholds() as $oid => $descr) {
+            if (str_contains($thresholdOid, $oid)) {
+                return $descr;
+            }
+        }
+        return 'unknown';
+    }
+    public function getThresholds()
```
murrant: This seems to be duplicated

h-barnhart: The function name is the same, but the array of values returned is not. One is for Ethernet access ports and the other is for Ethernet network ports. Really the only difference between the two arrays are three characters in the keys that identify the trap as coming from an access port (cm**Acc**PortQosShaperStatsBT) or a network port (cm**Net**PortQosShaperStatsBT). I think I can refactor it down though.

PR #10105, opened 2019-04-14
Billing: show all ports except loopback

erdems:
- Modified ifType to include all except software loopback, to cover kernel interfaces like LAG, RETH, bridge.

- Modified / fixed redirection after bill deletion.

#### Please note

#### Testers

### DIFF
--- a/includes/html/pages/bill/actions.inc.php
+++ b/includes/html/pages/bill/actions.inc.php
@@ -9,7 +9,7 @@ if ($_POST['action'] == 'delete_bill' && $_POST['confirm'] == 'confirm') {
 
     echo '<div class=infobox>Bill Deleted. Redirecting to Bills list.</div>';
 
-    echo "<meta http-equiv='Refresh' content=\"2; url='bills/'\">";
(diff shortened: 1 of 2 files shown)


laf:
Would this make sense to get the ifDescr and os values from the DB and run each port through `is_port_valid()`?

erdems:
ifDescr sounds a good idea, too. I couldn't think of an interface with description / pattern set matching the billing but not carrying traffic.

laf:
Ace, if you update the PR to remove the port checks in the sql query and run each port through `is_port_valid()` then this should be good.

Jellyfrog:

erdems:
Sorry I hadn't realized this was pending on me.

I would recommend simply going with the initial suggestion which is to include all except software loopback to be able to capture non-Ethernet interfaces with the relevant description that can be parsed.
Does that make sense?

Jellyfrog:
ping @laf

laf:
If you use `is_port_valid()` then users can control what gets excluded.

murrant:
Easiest way is to check deleted=0 for ports.

murrant:
Closing again, pending submitter update

PR #10114, opened 2019-04-16
Fping:  added fping option to support retries for normal host checks during polling.

InsaneSplash:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
You need to be careful with this.

3 count, 3 retries, .5s interval, .5s timeout
can result in 5.5s of waiting for pings when a device is down instead of the 1.5s now.

This could result in blowing up some people's polling windows.

You should probably default retries to 0 or 1.

I'm also curious if fping still reports retries properly for the performance graph.

murrant:
Closing, no response.  Just respond to re-open :)

PR #10121, opened 2019-04-17
WIP - Inline user_func for PacketLight support

PipoCanaja:
New attempt to solve the PacketLight devices using constant offsets in their MIB replies, by allowing inline anonymous functions in YAML : 
```
                    oid: edfaConfigTable
                    value: edfaAdminOutputPower
                    num_oid: '.1.3.6.1.4.1.4515.1.9.1.1.1.13.{{ $index }}'
                    descr: '{{ $ifDescr }} Admin Output Power'
                    index: 'edfaAdminOutputPower-{{ $index }}'
                    user_func: '($value/10)-30'
                    group: '{{ $edfaFreeDescription }}'
```

If the user_func is not a 'standard' function, the user_func expression is eval'ed in an attempt to create an anonymous function, in an exception handler, and the resulting function (if any) is called with $value; 

```
                    if (isset($user_function) && !function_exists($user_function)) {
                        //We try to eval() it. This is not user entered value.
                        try {
                            $func = eval('return function($value){return (' . $user_function . ');};');
                        } catch (ParseError $e) {
                            $user_function = null;
                            d_echo('Error: Invalid user_func.');
                        }
                        if (is_callable($func)) {
                            $value = $func($value);
                        }
                    }
```

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hello @murrant,
This code will probably fail (user_func is not matching the validation now) but I would like your opinion on this variant. 
Only basic tested occured so far but it works nicely on the 3 devices I have.

PipoCanaja:
(by the way, this PR is not clean, and is mixing 2 changes, so it is not meant to be merged, only a WIP here)

PipoCanaja:
Bug identified in the discovery code. Specific bug PR opened on it. This current one will be continued when bug will be fixed.

murrant:
The problem with this is, I really hate eval()...

PipoCanaja:
> The problem with this is, I really hate eval()...

Who does not ;) ? I expected this but hoped that eval() would be tolerated here cause the string we call eval() on is not "user_provided", YAML file is static here. 

If we decide to avoid eval(), What would you think of this kind of arithmetic processor : 
https://packagist.org/packages/mossadal/math-parser

murrant:
Again, my suggestion is to not use yaml for these sensors.

PipoCanaja:
So you would code discovery AND polling for those ? OK then. I'll take that route.

PR #10126, opened 2019-04-19
fix call to shortDisplayName on null in MuninPluginController

CirnoT:
MuninPluginController did not ask for device_id in baseQuery, hence later it was impossible to construct valid Device model causing subsequent call to shortDisplayName in formatItem to be on NULL.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Nice fix

PR #10127, opened 2019-04-19
Handle edge case in graph view for Munin plugins

CirnoT:
Munin plugins have set param in Graph Controller as 'device=x&plugin=y', which works fine for graph.php call, however it does not work as link to graph page, which expects 'device=x/plugin=y'. This commit adds str_replace in blade view for graph to ensure that '&' is replaced by '/' in param.

The reason why this logic is in view instead of controller is that Munin is the only one that uses this particular type of param 'hack' and creating separate variable in controller to fix what is essentially a 'hack' for single case seems unnecessary.

If this is not acceptable, I will update this PR to create separate variable in controller and then use @isset statement in blade template to decide whether it should use $param or separate variable for link, however in my opinion it would be harder to understand.

This fixes behavior reported on https://community.librenms.org/t/dashboard-munin-graph-links/3626

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Widgets/GraphController.php
+++ b/app/Http/Controllers/Widgets/GraphController.php
@@ -160,15 +160,16 @@ class GraphController extends WidgetController
         }
 
         $type = $this->getGraphType();
-        $param = '';
+        $params = [];
 
         if ($type == 'device') {
-            $param = 'device='.$settings['graph_device'];
+            $params[] = 'device='.$settings['graph_device'];
         } elseif ($type == 'application') {
-            $param = 'id='.$settings['graph_application'];
+            $params[] = 'id='.$settings['graph_application'];
         } elseif ($type == 'munin') {
             if ($mplug = MuninPlugin::find($settings['graph_munin'])) {
-                $param = 'device='.$mplug->device_id.'&plugin='.$mplug->mplug_type;
+                $params[] = 'device='.$mplug->device_id;
+                $params[] = 'plugin='.$mplug->mplug_type;
             }
         } elseif ($type == 'aggregate') {
             $aggregate_type = $this->getGraphType(false);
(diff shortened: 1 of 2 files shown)


murrant:
Yeah, I don't think fixing it like this is preferred.

FYI, the reason it doesn't work is because the ampersand (&) is escaped to &amp;

CirnoT:
No, the & is not escaped in this case, but URL of ```https://librenms/graphs/device=7&plugin=sensors_power/type=munin_graph/from=1555689592``` is invalid as value of ```sensors_power/type=munin_graph/from=1555689592``` does not exist for 'plugin'. Similarly, ```graphs/device=7/type=munin_graph/from=1555689667&plugin=sensors_power``` is invalid too it seems, the only working one would be ```https://librenms/graphs/device=7/type=munin_graph/from=1555689667/plugin=sensors_power``` and the order of parameters is irrelevant.

CirnoT:
I've changed how graphs are rendered in blade by adding support for multiple parameters via array. Because it seems that Blade doesn't support no-whitespace tag like Twig that I am familiar with, the spacing looks weird in template now but there isn't any other way that I can see that would make Blade recognize @foreach tag and at the same time allow for spacing that won't be rendered in HTML as spaces in URL.

CirnoT:
It looks like the changes in PHP file caused old issue on Code Climate to reappear, hope that won't be a problem. 

Do let me know if this is okay or you would prefer to use ```@php``` (or possibly ```<?=``` as it works better with one-liners) to use ```implode``` to join array elements instead of ```@foreach```.

resources/views/widgets/graph.blade.php
```diff
@@ -1,9 +1,13 @@
 <div class="dashboard-graph">
-    <a href="graphs/{{ $param }}/type={{ $graph_type }}/from={{ $from }}">
+    <a href="graphs/
+@foreach($params as $param){{ $param }}/@endforeach
```
murrant: Wouldn't implode be nicer than foreach?
`{{ implode($params, '/') }}`

app/Http/Controllers/Widgets/GraphController.php
```diff
@@ -160,15 +160,16 @@ public function getView(Request $request)
         }
 
         $type = $this->getGraphType();
-        $param = '';
+        $params = array();
```
murrant: please use short array syntax $params = [];

resources/views/widgets/graph.blade.php
CirnoT: Thanks. I didn't know you can utilize PHP functions in Blade like that.

CirnoT:
Fixed requested changes. Using ```implode``` now and squashed commits.

PR #10050, opened 2019-03-31
Python dispatcher service v2

murrant:
Ping support and a lot of refactoring.
Most code for each queue is in a specific queue manager class now.
Now uses queue that restrict duplicate entries, making things simpler

Will need notification, as it requires Redis 5.x and  3.x of the Redis plugin to support ordered sets.

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #10138, opened 2019-04-23
Python service: Reconnect to MySQL server when connection is lost

jplitza:
Currently, the Poller service simply hangs when it looses the connection
to the MySQL server. With this patch, it pings the server before making
a request, which will automatically reconnect if the connection died. If
the new connection cannot be established, a second try is done after 10
seconds.

Whether 10 seconds and only retrying once obviously is up for debate, and could probably even be made configurable. So feedback is appreciated.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/__init__.py
+++ b/LibreNMS/__init__.py
@@ -1,8 +1,8 @@
 import threading
 
-from logging import critical, info, debug, exception
+from logging import critical, warning, info, debug, exception
 from math import ceil
-from time import time
+from time import time, sleep
 
 from .service import Service, ServiceConfig
 from .queuemanager import QueueManager, TimedQueueManager, BillingQueueManager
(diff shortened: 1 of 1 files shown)


murrant:
Looks good, my only concern is how much time the ping takes.  The dispatcher service hits the database fairly quickly.

jplitza:
I'd say that if you notice the delay of a ping to the database, you probably already have other problems. In our setup with ~600 devices on 5 pollers an 60-second-polling, the Poller Perf graph didn't (is that the right metric here?) didn't increase at all.

murrant:
@jplitza probably but I was thinking about this more.  Would it be more correct to release the master role if you can't connect to the DB.  Then when master is re-acquired (possibly by the same node) it should try to connect to mysql.

jplitza:
I have to admit, I didn't dig into the details of the poller service at all. Does only the master need and establish the MySQL connection?

I made this patch because we had the problem that when the MySQL server died/went away/rebooted/whatever, we had to restart the poller services. We restarted them all, but maybe we only had to restart the master?

murrant:
There is some stats reporting for each poller that has a connection.  Maybe we need to combine your fix and my idea.

jplitza:
If every poller has/needs a MySQL connection, I don't see why we should touch the poller status. Also, the loss of the MySQL connection most likely is due to the server going away (or network problems). I wouldn't expect the other pollers to still have a working connection.

laf:
Yes all pollers need a mysql connection at present.

murrant:
Integrated handling the db going away into #10050
Would be great to have extra testing before that is merged.

PR #10141, opened 2019-04-24
Relax validation for smseagle hostname

petracvv:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

When trying to setup the smseagle transport, the validator was checking for a vaild URL in the form validation. However, the transport expects a hostname not a URL because it constructs the URL itself: https://github.com/librenms/librenms/blob/5e1994a39d268382ac99bd4e0cc01fbe1fc05547/LibreNMS/Alert/Transport/Smseagle.php#L47

This pull request relaxes the validation to accept a string so IP addresses and Hostnames are accepted by the form.

### DIFF
--- a/LibreNMS/Alert/Transport/Smseagle.php
+++ b/LibreNMS/Alert/Transport/Smseagle.php
@@ -64,9 +64,9 @@ class Smseagle extends Transport
         return [
             'config' => [
                 [
-                    'title' => 'SMSEagle URL',
+                    'title' => 'SMSEagle Host',
                     'name' => 'smseagle-url',
-                    'descr' => 'SMSEagle URL',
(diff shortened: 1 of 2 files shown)


petracvv:
It might be better to rename the attribute, but I am not sure what that would entail.

PipoCanaja:
Hi @petracvv 
Renaming the attribute would kill the alerts for any user that is running smseagle now (and probably configured it before the stricter checks on the url ...). 
So I would suggest to avoid renaming it for now. 

You should still add these to the PR: 
- Change the description of the field to make it clear for user that it is a host.
- Check the documentation and adapt it if necessary.
Thanx

petracvv:
OK. I updated the property description and the docs. Let me know if I should look at anything else.

PR #10142, opened 2019-04-24
Fix services with scripts inheriting DS from previous service on detail view

CirnoT:
Ensures ```check_ds``` variable is always ```null``` on each run of loop that displays services graphs. Fixes services that have script (like check_dns) but don't set ```check_ds``` on their own inheriting ```check_ds``` from previous service that ran (for example check_icmp) and displaying additional bogus graphs as result.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


CirnoT:
Force-pushed to re-run Travis check.

PR #10144, opened 2019-04-25
Store IPv4 networks as network address and fix address search page showing networks not addresses

CirnoT:
I am not sure how LibreNMS will handle it for existing addresses, possibly there needs to be some logic added to sort them out?

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
LGTM, annoying that the discovery module never removes networks...

PR #10153, opened 2019-04-26
Fix MySQL error were prepared statement contains too many placeholders

davidmnelson:
Fixes MySQL error were prepared statement contains too many placeholders. Modifies dbBulkInsert to run array_chunk on incoming data and inserts those smaller chunks. This error is unlikely to have affected most installations, but was an issue for us getting up-to-date IP address data during discovery on our core router.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Thanks for fixing.  Not sure this is the right level for the fix, but should be ok.

PR #10158, opened 2019-04-28
Fixed Cisco MAC accounting discovery, polling and HTML templates.

deesel:
```
Component | Version
--------- | -------
LibreNMS  | 1.50-65-g6188041ba
DB Schema | 2019_02_10_220000_add_dates_to_fdb (132)
PHP       | 7.3.4
MySQL     | 10.3.9-MariaDB-1:10.3.9+maria~bionic
RRDTool   | 1.6.0
SNMP      | NET-SNMP 5.7.3
```

There were several problems with Cisco MAC accounting:

- Discovery module never created any MAC accounting entries.
- Poller then only walked through cipMacHCSwitchedBytes and cipMacHCSwitchedPkts counters, but some (older) devices support only cipMacSwitchedBytes and cipMacSwitchedPkts.
- HTML templates used period names unknown to rrdtool ('day' => '1d', 'twoday' => '2d'...)
- SQL query to get MAC accounting entries used incorrect column name to sort by default.
- rrdtool complains about empty arguments; should use quotes (empty string) in place of an argument (:: => :'':)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Can you add test data for this module?

Add this to `tests/module_tables.yaml` and follow the docs to capture snmprec and json files.

```yaml
cisco-mac-accounting:
    mac_accounting:
        excluded_fields: [ma_id, port_id, poll_time, poll_prev, poll_period]
        joins:
            - { left: ports_fdb.port_id, right: ports.port_id, select: [ifIndex] }
        order_by: ports.ifIndex, mac
```

deesel:
I had to slightly modify `tests/module_tables.yml` like this:

```yaml
cisco-mac-accounting:
    mac_accounting:
        excluded_fields: [ma_id, port_id, poll_time, poll_prev, poll_period, cipMacHCSwitchedBytes_input_rate, cipMacHCSwitchedBytes_output_rate, cipMacHCSwitchedPkts_input_rate, cipMacHCSwitchedPkts_output_rate]
        joins:
            - { left: mac_accounting.port_id, right: ports.port_id, select: [ifIndex] }
        custom_where: WHERE ports.device_id=?
        order_by: ports.ifIndex, mac
```

`*_rate` fields are actually calculated based on current time, which is obviously different each time the test is run. Also, `WHERE` clause needed to be fixed, because `mac_accounting` table has no `device_id` column.

Finally, because `cisco-mac-accounting` module uses port data, it needs to be tested together with the `ports` module.

```
./scripts/pre-commit.php --db --snmpsim -m ports,cisco-mac-accounting
```

murrant:
There is a spot where you can set dependencies, in LibreNMS\Util\ModuleTestHelper.php

deesel:
I didn't know this, thanks. Dependency is set now in LibreNMS\Util\ModuleTestHelper.php

includes/html/pages/device/port/macaccounting.inc.php
```diff
@@ -47,32 +47,32 @@
            <div style='margin: 0px 10px 5px 0px; padding:5px; background: #e5e5e5;'>
            <span class=device-head>Day</span><br />
 
-           <a href='".generate_url($link_array, array('view' => 'macaccounting', 'subview' => 'top10', 'graph' => $vars['graph'], 'sort' => $vars['sort'], 'period' => 'day'))."'>
+           <a href='".generate_url($link_array, array('view' => 'macaccounting', 'subview' => 'top10', 'graph' => $vars['graph'], 'sort' => $vars['sort'], 'period' => '1d'))."'>
```
murrant: These dates don't work.  I think it should be a timestamp.

deesel: Can you be more specific regarding of what exactly doesn't work, please? According to [RRDtool documentation](https://oss.oetiker.ch/rrdtool/doc/rrdfetch.en.html#AT-STYLE_TIME_SPECIFICATION), time offset units can be specified as abbreviations too. Generated RRDtool command looks like this:

`rrdtool graph /tmp/b6L3kMK3W97WqQp5  --alt-autoscale-max --rigid -E --start -1d --end 1558349100 --width 745 --height 300 -c BACK#EEEEEE00 -c SHADEA#EEEEEE00 ... <additional output omitted>`

...and works like a charm. I use RRDtool v1.6.0.

PR #10167, opened 2019-04-29
Add veya as bad interface for APC

CirnoT:
This seems to be somehow half-broken on AP9606, it never records any actual traffic.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/apc.yaml
+++ b/includes/definitions/apc.yaml
@@ -1,6 +1,8 @@
 os: apc
 text: 'APC Management Module'
 type: power
+bad_if:
+    - veya
 over:
     - { graph: device_current, text: Current }
     - { graph: device_voltage, text: Voltage }


laf:
> This seems to be somehow half-broken on AP9606, it never records any actual traffic.

But does on other models? If it's only one model then you should disable it on your install and leave it for others. If it's broken on all then this change would be fine.

CirnoT:
I am afraid I have no other models to test it against, but you are right - adding it to config seems like better option for now. Anyone here care to chime in and confirm whether other models have same issue/use same interface name?

murrant:
My APC PDUs have "lance" and "LOOPBACK" which all work fine.

laf:
I'll close this down until we get more feedback. I can't see any mention of veya in our test data.

PR #10173, opened 2019-05-01
Replace legacy menu with new Blade generated one

murrant:
Supports translation
Custom menu functionality ported to Blade and documented

Lots of fun with global variables...

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


laf:
I can't test every bit of this as I don't have all menus/items but these two have small issues

Screenshot 2019-05-01 at 22 42 46
Screenshot 2019-05-01 at 22 42 18

murrant:
@laf I fixed the double divider and a couple other similar issues.  But I could not reproduce the issue with custom port descriptions.

CirnoT:
@murrant The issue with custom descriptions comes from ```$config['custom_descr'][] = '';``` in `includes/defaults.inc.php`. Either change it to ```$config['custom_descr'] = array();``` or make sure that `custom_descr` is not `""` in `@foreach`

murrant:
Ok, got those fix.  I think this might be ready.  Thanks @CirnoT

CirnoT:
`includes/html/pages/apps.inc.php` is still broken and won't display anything on Overwiew (details in review on that file):
chrome_2019-05-05_15-37-48

It seems your recent changes introduced another issue with services too:
chrome_2019-05-05_15-36-27

In addition, ignored ports are still counted as disabled and down - I believe they shouldn't be.
chrome_2019-05-05_15-36-50

resources/views/layouts/menu.blade.php
```diff
@@ -151,13 +151,13 @@
                         <a href="{{ url('services') }}" class="dropdown-toggle" data-hover="dropdown" data-toggle="dropdown"><i class="fa fa-cogs fa-fw fa-lg fa-nav-icons hidden-md" aria-hidden="true"></i> <span class="hidden-sm">Services</span></a>
                         <ul class="dropdown-menu">
                             <li><a href="{{ url('services') }}"><i class="fa fa-cogs fa-fw fa-lg" aria-hidden="true"></i> All Services </a></li>
-                            @if($service_status)
+                            @if($service_counts['total'])
                                 <li role="presentation" class="divider"></li>
```
CirnoT: I think simplest fix here would be to change
```
```
to
```
@if($service_counts['warning'] or $service_counts['critical'])
@endif
```

includes/html/pages/apps.inc.php
```diff
@@ -287,7 +287,7 @@
     'device' => $device['device_id'],
     'tab'    => 'apps',
 );
-foreach ($app_list as $app) {
+foreach (\LibreNMS\Util\ObjectCache::applications() as $app) {
     echo $sep;
     if ($vars['app'] == $app['app_type']) {
```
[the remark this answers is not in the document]
murrant: Thanks, I'll check, but fyi Eloquent models implement array access.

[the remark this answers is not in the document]
murrant: The problem was the collection was nested, so I had to flatten it first.

murrant:
Thanks for testing @CirnoT 
I think I have all of those items fixed now.

LibreNMS/Util/ObjectCache.php
```diff
@@ -0,0 +1,214 @@
...
+use App\Models\Vrf;
+use Cache;
+
+class ObjectCache
+{
+    private static $cache_time = 5;
+
+    public static function applications()
+    {
+        return Cache::remember('ObjectCache:applications_list', self::$cache_time, function () {
```
Jellyfrog: Unless I'm missing something, you need to add the user id to these cache keys, else all users will get the same cached result, no matter their access level.

Jellyfrog: Maybe?

```suggestion
        return Cache::remember('ObjectCache:applications_list' . auth()->user()->id, self::$cache_time, function () {
```

murrant: Thank you @Jellyfrog 
Currently that is not an issue as this is only cached for the run duration (cache type array)  but it definitely should be.

murrant: Fixed, thanks again :)

CirnoT:
Looks good to me now.

murrant:
I can't reproduce that @CirnoT

CirnoT:
Sorry somehow I forgot how to trigger it: Make sure you have pseudowire disabled in config and Open 'Manage Users' page.

PR #10182, opened 2019-05-02
Added support for routing table collection in discovery

PipoCanaja:
Rewriting the route collection. Current stable code seems to be very limited. And the DB structure is not adapted to the newer and widely used IP-FORWARD-MIB.

- New DB table following closely the  IP-FORWARD-MIB structure. All other source of routes are "converted" to this structure for display. 
- Limitation of the number of discovered route is implemented with a new config parameter. Nobody wants the Internet routing table to be polled and stored in DB. Probably neither the device nor LibreNMS would like it anyway ... 

This code is still WIP. TODO: 
- Properly display config parameters in Settings panel
- More tests

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
To be discussed : 
- Structure of the DB new table
- Structure of the Laravel Model
- Migration from old route table to new one. The old routing code seems to be non working anyway so most probably, changing it and the table won't hurt. 

TODO: 
- Tests
- Documentation
- More interaction with VRF display

app/Http/Controllers/Table/RoutesTablesController.php
```diff
@@ -0,0 +1,188 @@
...
+        if (isset($sort['context_name'])) {
+            $query->orderBy('context_name', $sort['context_name']);
+        }
+
+        return $query;
+    }
+
+    public function formatItem($route_entry)
+    {
+        $translateProto[1] = '1-other';
```
murrant: You should define these arrays outside of the function (otherwise they get defined each time it is called)
and better to define like this:
```php
$translateProto = [
    1 => '1-other',
    ...
];
```

app/Models/InetCidrRoute.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace App\Models;
+
+class InetCidrRoute extends DeviceRelatedModel
+{
+    protected $table = 'inetCidrRoute';
+//    protected $primaryKey = 'inetCidrRoute_id';
+    public $timestamps = true;
+
+    public function getPort()
```
murrant: This should be defined as a port relationship, which means you probably want to have a port_id field on this table.

database/migrations/2019_04_22_220000_update_route_table.php
```diff
@@ -0,0 +1,47 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('inetCidrRoute', function (Blueprint $table) {
+              $table->increments('inetCidrRoute_id');
+              $table->timestamps();
+              $table->unsignedInteger('device_id');
```
murrant: Add a port_id field

includes/discovery/route.inc.php
```diff
@@ -14,154 +14,327 @@
...
+        $entryClean['inetCidrRouteNextHopAS'] = '0';
+        $entryClean['inetCidrRouteProto'] = $ipRoute['ipRouteProto'];
+        $entryClean['inetCidrRouteType'] = $ipRoute['ipRouteType'];
+        $entryClean['inetCidrRouteIfIndex'] = $ipRoute['ipRouteIfIndex'];
+        $entryClean['context_name'] = '';
+        $entryClean['device_id'] = $device['device_id'];
+        $entryClean['updated_at'] = array('NOW()');
+        $entryClean['inetCidrRouteNextHop_device_id'] = $device['device_id'];
+        if ($entryClean['inetCidrRouteNextHop'] != '127.0.0.1' && $entryClean['inetCidrRouteNextHop'] != 'fe80::') {
+            $entryClean['inetCidrRouteNextHop_device_id'] = Device::findByIp($entryClean['inetCidrRouteNextHop'])->device_id;
```
murrant: This has the possibility to return null, which will cause an error when you try to null->device_id

includes/discovery/route.inc.php
```diff
@@ -14,154 +14,327 @@
...
-    if (key_exists('vrf_lite_cisco', $device) && (count($device['vrf_lite_cisco']) != 0)) {
-        //i will only get one context of vrf, read the begin of this file
-        foreach ($device['vrf_lite_cisco'] as $vrf_lite) {
-            if (!key_exists($vrf_lite['vrf_name'], $vrfs_lite_cisco)) {
-                $vrfs_lite_cisco[$vrf_lite['vrf_name']] = $vrf_lite;
-            }
-        }
+//Get the configured max routes number
+$max_routes = 10000;
+if (null != (Config::get('routes.max_number'))) {
```
murrant: This is the really long way of writing:
```php
$max_routes = Config::get('routes.max_number', 10000);
```

murrant: default should probably match that in defaults.inc.php

includes/html/pages/device.inc.php
```diff
@@ -267,6 +267,11 @@
             $routing_tabs[] = 'cisco-otv';
         }
 
+        $device_routing_count['routes'] = @dbFetchCell('SELECT COUNT(*) FROM `inetCidrRoute` WHERE `device_id` = ?', array($device['device_id']));
```
murrant: Why are you suppressing errors here?

routes/web.php
```diff
@@ -82,6 +82,7 @@
             Route::post('device', 'DeviceController');
             Route::post('eventlog', 'EventlogController');
             Route::post('fdb-tables', 'FdbTablesController');
+            Route::post('inetCidrRoute', 'RoutesTablesController');
```
murrant: probably just call the route "routes"

murrant:
Looks pretty good overall.  I don't think you have to migrate from the old table.  It isn't historical, so nothing will really be lost.

includes/html/pages/device.inc.php
PipoCanaja: I copied paste the code from VRFs a few lines above ... did not even notice the @ while doing it I suppose.

includes/discovery/route.inc.php
PipoCanaja: better now.

database/migrations/2019_04_22_220000_update_route_table.php
PipoCanaja: done

includes/discovery/route.inc.php
PipoCanaja: done

app/Http/Controllers/Table/RoutesTablesController.php
PipoCanaja: I moved it as static in the class.

includes/discovery/route.inc.php
```diff
@@ -14,154 +14,330 @@
...
+            '`inetCidrRoute_id` = ?',
+            array($k)
+        );
+        echo '-';
+        d_echo($delete_row_data[$k]);
+    }
 }
+
+foreach ($update_row as $upd_entry) {
+    dbUpdate(
```
murrant: Using Eloquent would save a lot of code, up to you if you want to convert it as it will change a decent amount.

PipoCanaja: I don't feel comfortable yet with Eloquent. Took me some time just to use the port relation without crashing with 100 lines of Exception :) I'll keep (testing) this code for now.

murrant: no problem

PipoCanaja:
Hello,
I would say that I am done for first version. I tried with very limited success to implement filtering. But the collection, display and sort are running OK, on a few devices (Huawei, Cisco IOS, Cisco IOSXE). 
I don't have any lab device available for collecting test data right now, but the route module is probably not tested anyway for the moment. 
So from my point, this PR moved from WIP to 'To Be Reviewed' ;)

PipoCanaja:
Removed the Breaking change cause the old table did not keep any history anyway.
@murrant: Should I remove the route table right away in the migration scripts ? The route table is not used anymore after this PR as far as I found.

PipoCanaja:
@murrant How should I proceed for tests, as the route module is disabled by default :) ?

murrant:
./scripts/save-test-data.php -o whatevs -m route

PipoCanaja:
> ./scripts/save-test-data.php -o whatevs -m route
> Oh and you'll have to enable it for the snmprec collection in config.php.

@murrant The snmprec part is OK but the save-test-data.php does not want to do fill the JSON for some reason. At least, Travis does not do it either so both are equal :)

murrant:
I'll check it.

murrant:
Alright:
module_tables.yaml
```
route:
    inetCidrRoute:
        excluded_fields: [device_id, created_at, updated_at]
        order_by: inetCidrRouteDest
```

Don't store the next hop device_id it takes a lot of mysql queries and is just as easily queried when the data is displayed to the user.

Any reason to keep the old table?  I'd say drop that and rename this one "route", or modify route so matches this structure.

PipoCanaja:
All right, I will have a look this week, I've been off for a while but back now.

PipoCanaja:
Hello @murrant 
New test machine, so back on tracks for this PR.  I removed the old route table, and the next_hop device ID as well. Little bit of codeclimate and working on snmpsim tests now. At least the code itself behave correctly in my lab after those changes.
Will update when Travis is happy.

PipoCanaja:
Done

PipoCanaja:
Hi @murrant 
Took time but it is now behaving correctly. You can choose between all routes (with history) or routes from last discovery. 

Looks like it is ready for review and merge if OK with @librenms/reviewers . 

Capture d’écran 2019-08-23 à 09 28 04

Bye

murrant:
@PipoCanaja Hey after thinking about this a bit.  I want to keep the existing table structure as much as possible.  Just add the new fields and map the new oids to the old fields.  I know this is a bit of work, but I think it will be better.

Also, would like to convert this to the class base style of module, but I can do that if you can change the db structure as described above.

PipoCanaja:
Hi @murrant 
Got a lot delayed on this one. Now my dev machine is gone, need to reinstall one. But will get back to this.
In term of DB, OK to get back to "route" instead of "inetCidrRoute" table, and create all now columns I need in "route". 
But the names of the column, I would like to keep them as they are, because they follows the inetCidrRoute logic, it makes more sense and the inetCidrRoute is already ipv4/ipv6 capable. Legacy mib data (or any other "route" source we would include later), would then mapped to inetCidrRoute.

PipoCanaja:
> @PipoCanaja Hey after thinking about this a bit. I want to keep the existing table structure as much as possible. Just add the new fields and map the new oids to the old fields. I know this is a bit of work, but I think it will be better.

First try in this direction. Still a WIP, tests must be fixed, etc etc but I have a working lab again, including the changes in the Config handling. 

> Also, would like to convert this to the class base style of module, but I can do that if you can change the db structure as described above.
This part is on hold for the moment

PipoCanaja:
@murrant I now keep the "route" table instead of creating a new name. It still works well in my env, and it is rebased to current master, including the JS localisation files. Probably good time for a new review :)

murrant:
LGTM, just update mix-manifest.json with `./lnms translation:generate`

PipoCanaja:
All right, translation updated, @murrant

murrant:
:1st_place_medal:

murrant:

sajiby3k:
I need to submit some debug about this. Like for example fortigate I can see the routing table by calling discovery.php, but there is no sql insert operation happening. For Hp comware switches getting fatal error.

PR #10184, opened 2019-05-03
CBSD VM discovery

VVelox:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Still working on the extend portion, but this is all good.

Need to look into adding graphing and the like for max ram, ram usage, cpu usage, and current number of procs.

### DIFF
--- a/includes/discovery/cbsd-vminfo.inc.php
+++ b/includes/discovery/cbsd-vminfo.inc.php
@@ -46,7 +46,7 @@ if ($device['os'] == 'freebsd') {
          * Check whether the Virtual Machine is already known for this host.
          */
         if (dbFetchCell("SELECT COUNT(id) FROM `vminfo` WHERE `device_id` = ? AND `vmwVmVMID` = ? AND vm_type='cbsd'", array($device['device_id'], $vm)) == 0) {
-            $vmid = dbInsert(array('device_id' => $device['device_id'], 'vm_type' => 'cbsd', 'vmwVmVMID' => $vm, 'vmwVmDisplayName' => mres($vmwVmDisplayName), 'vmwVmGuestOS' => mres($vmwVmGuestOS), 'vmwVmMemSize' => mres($vmwVmMemSize), 'vmwVmCpus' => mres($vmwVmCpus), 'vmwVmState' => mres($vmwVmState)), 'vminfo');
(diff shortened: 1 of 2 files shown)


VVelox:
Worth nothing vmwVmVMID is always set to zero as CBSD does not produce a numeric ID really. Well there is, but this is the process ID of the main bhyve proc running it and it will change between VM restarts because of that.

includes/discovery/cbsd-vminfo.inc.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    $cbsd_keys=array_keys($cbsd[data]);
+
+    foreach ($cbsd_keys as $vm) {
+        $vmwVmDisplayName = $cbsd{data}{$vm}{jname};
+        $vmwVmGuestOS     = $cbsd{data}{$vm}{vm_os_type};
+        $vmwVmMemSize     = $cbsd{data}{$vm}{vm_ram};
+        $vmwVmState       = $cbsd{data}{$vm}{status};
+        $vmwVmCpus        = $cbsd{data}{$vm}{vm_cpus};
```
PipoCanaja: You need to write all hash as $myHash{'key'} :)

PipoCanaja:
You still have quite a few keys to updates, and you also use '{' instead of '[' to enclose the keys. Could you change that to a more standard syntax ?

includes/discovery/cbsd-vminfo.inc.php
```diff
@@ -0,0 +1,92 @@
...
+    /*
+     * Fetch information about Virtual Machines.
+     */
+    try {
+        $cbsd=json_app_get($device, 'cbsd-bls', 1);
+    } catch (JsonAppException $e) {
+        # This is just empty and exists to catch it not existing.
+    }
+
+    $cbsd_keys=array_keys($cbsd[data]);
```
PipoCanaja: another one :)

VVelox: Doh! You got to this first.

VVelox:
@PipoCanaja Howdy! Sorry about the slow reply on this. Thanks for noticing the slightly out of date comment chunk there. :)

Also I think this is alll good now. Was just giving this another look over and not seeing anything that does not meet the current standards off hand.

murrant:
This requires an snmp extend script? You probably need some docs for it.

VVelox:
@murrant Actually going to be shelving this for a bit. I really want to put more time into actually making pulling this info actually useful for something.

Currently we can for VMware and libvirt and it is not really used. This would just add a third we don't really do anything with this with.

Need to begin actually implementing actually graphing and the like based on this.

PR #10185, opened 2019-05-03
Better services graphing support

CirnoT:
- This moves services graph code from `device/` to `service/` and removes unused code (`service/availability.inc.php`). This has no real changes to end-user except ensuring that when viewing graph, proper title is shown:

After:
chrome_2019-05-03_17-35-12
Before:
chrome_2019-05-03_17-35-49

- Second commit adds proper support for services in dashboard graph widget. Previously it was piggy-backing on generic device graph, which made it impossible to select specific service for device (first service would be always used). This adds same level of support for services as present with Munin Plugins, where you can select service to be graphed.

chrome_2019-05-03_17-38-27

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Looks good but.  I think it could cause some issues if people are linking to existing graphs called device_service.  Either find a way to redirect that or we can just send a user notification probably.

CirnoT:
I think sending notification would be preferred over adding additional logic for redirecting (which may cause issues later on if we decide to have service graph under device for something else than Nagios services).

murrant:
Could you create a notification on https://github.com/librenms/librenms.github.io ?

CirnoT:
Done at https://github.com/librenms/librenms.github.io/pull/116

PR #10186, opened 2019-05-03
Cast alert ID to string for PD API

supertylerc:
This is a simple one.  The PagerDuty API expects the Alert ID to be a string, and currently it is not.  It seems that it may have been a change on PagerDuty's end to start enforcing the type as this _used_ to work, but it suddenly broke one day and I don't think it was because of LibreNMS.

PagerDuty docs reference: https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hi @supertylerc 
Thanx for the fix !

PR #10188, opened 2019-05-04
Removed broken POE graphing code

PipoCanaja:
Hello

Currently, there is no way to poll POE values and graph them with standard MIB. The only standard info we can get is "do we have POE available and active" but this information is not used in LibreNMS so far. 
The removed code did use some specific Cisco MIBS for the values, which obviously do not return any value for non Cisco devices. 
This PR is removing the code to avoid unnecessary polling of useless values. And also, currently, unnecessary creation of empty RRD (waste of space). 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/ports/port-poe.inc.php
+++ b/includes/polling/ports/port-poe.inc.php
@@ -59,24 +59,4 @@ if (($device['os'] == 'vrp')) {
         data_update($device, 'poe', $tags, $fields);
         echo 'PoE(IOS) ';
     }//end if
-} else {
-    //This is the legacy code, to be tested against devices. This code looks terribly broken. There is
-    //most probably no device that can show anything out of this ...
-
-    if ($this_port['dot3StatsIndex'] && $port['ifType'] == 'ethernetCsmacd') {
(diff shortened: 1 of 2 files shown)


includes/polling/ports/port-poe.inc.php
```diff
@@ -59,24 +59,4 @@
         data_update($device, 'poe', $tags, $fields);
         echo 'PoE(IOS) ';
     }//end if
-} else {
-    //This is the legacy code, to be tested against devices. This code looks terribly broken. There is
-    //most probably no device that can show anything out of this ...
-
-    if ($this_port['dot3StatsIndex'] && $port['ifType'] == 'ethernetCsmacd') {
```
PipoCanaja: This condition is really broken. A lot of devices do have dot3StatsIndex set and will then create a completely empty RRD ...

murrant:
I'll take a look.  I think some devices support PoE info with a state sensor (off, 24v, 48v, etc)

PipoCanaja:
Then we would probably need different RRDs for them, cause the current structure is very bound to the Cisco implementation.

PR #10196, opened 2019-05-07
Updated the install docs so composer runs and fixed nginx php permissions

bugoff:
Composer requires the COMPOSER_ALLOW_SUPERUSER flag to be set when you run it as root.
We start the nginx process with php-fpm, so i set the user to librenms. This way we don't have any problems with permissions.

#### Please note

#### Testers

> Not sure if this is also necessary if i only updated the docs?


CirnoT:
Any specific reason why to change PHP user/group? Current configuration should work with ACL and `usermod -a -G librenms www-data` (which is one of first steps during installation instruction)

bugoff:
1: Because the installation instructions don't work.
2: The proper/secure way to run php as the user who runs the process.

doc/Installation/Installation-Ubuntu-1604-Apache.md
```diff
@@ -17,8 +17,7 @@ path: blob/master/doc/
 #### Install LibreNMS
 
     cd /opt
-    composer create-project --no-dev --keep-vcs librenms/librenms librenms dev-master
-
+    export COMPOSER_ALLOW_SUPERUSER=1  && composer create-project --no-dev --keep-vcs librenms/librenms librenms dev-master
```
murrant: This just disables the warning...  The warning is appropriate.
If we want to get rid of the warning, we could use git clone and composer install as separate steps.  Running the second as the librenms user.

doc/Installation/Installation-Ubuntu-1604-Nginx.md
```diff
@@ -52,14 +52,17 @@ Ensure date.timezone is set in php.ini to your preferred time zone.  See http://
     vim /etc/php/7.0/fpm/php.ini
     vim /etc/php/7.0/cli/php.ini
 
+    vi /etc/php/7.0/fpm/pool.d/www.conf
+    Change the user and group from www-data to librenms
```
murrant: Would it be best to make a pool just for librenms?

doc/Installation/Installation-Ubuntu-1604-Nginx.md
```diff
@@ -52,14 +52,17 @@ Ensure date.timezone is set in php.ini to your preferred time zone.  See http://
...
+
     phpenmod mcrypt
     systemctl restart php7.0-fpm
 
 #### Configure NGINX
 
     vim /etc/nginx/conf.d/librenms.conf
 
-Add the following config, edit `server_name` as required:
+Add the following config, edit `server_name` as required, or add _ if you want this vhost to match everything:
```
murrant: There is nothing special about _, anything that is unresolvable will work.  Could make the default in librenms.conf something unlikely to resolve since many people don't notice they need to edit server_name.
Another item might be to make LibreNMS not user server_name internally so it doesn't break links...

doc/Installation/Installation-Ubuntu-1604-Nginx.md
CirnoT: Yeah, if anything it would be better to create separate pool just for LibreNMS just so that default one doesn't run as librenms when user installs something else on same server.

bugoff: I suspect that people who run this on a shared server won't be needing the install manual. Other config is an option, but what about the default config? Do you remove it? Do you leave it and have people run a php pool which doesn't do anything? So using "resources"?

doc/Installation/Installation-Ubuntu-1604-Nginx.md
bugoff: Then +1 for librenms not using server_names, if you run it on an internal environment you can't use the ip anymore. This can indeed be removed, as it only breaks librenms when this is the only line you have. What is the reason for librenms auto guessing the url on which it should configure? Why don't you let people just use every hostname they can think of?

doc/Installation/Installation-Ubuntu-1604-Apache.md
bugoff: So it's oke to show a bigass warning, even if there is an option to disable it and make the installer look a little bit nicer?

CirnoT: Silencing it is probably not the best course of action. @murrant idea would be better if you want to get rid of it. Further, docs state:

> It also disables automatic clearing of sudo sessions, so you should really only set this if you use Composer as super user at all times like in docker containers

which is clearly not the case here with LibreNMS

doc/Installation/Installation-Ubuntu-1604-Nginx.md
CirnoT: Install manual is always relevant - you don't expect people to just magically know know that they need to point `/api/v0` to `/api_v0.php`, don't you? Further, I would raise argument that less knowledgeable people would be more willing to install it on shared server, and people who 'know stuff' setting it on separate VM/LXC container. I don't see leaving default pool as an issue really.

murrant: One long term item is we need to have PHP create files as the librenms user.  Because there will be files shared between the frontend and backend.  The backend always runs as the librenms user.

doc/Installation/Installation-Ubuntu-1604-Nginx.md
murrant: It can run fine without server_names, just more likely for the user to misconfigure.

bugoff: Librenms uses the servername to redirect you to. If you fill in a servername librenms uses this as a url for the webpage.  In my test env i don't have a hostname pointed to the server, so i have to modify the base url.

murrant: How does this work then?
image

bugoff: So, i reinstalled librenms again.

Clicking on the alerts tab redirects me to librenms.example.com, clicking on the settings redirects me to librenms.example.com.

murrant: A lot of the issues with this are fixed by #10227

bugoff: @murrant 👍

doc/Installation/Installation-Ubuntu-1604-Nginx.md
bugoff: So, do i change the docs to create a new pool ?

doc/Installation/Installation-Ubuntu-1604-Apache.md
bugoff: Do i change the docs to git clone + composer install? or do i just set a message in the docs saying: "it's oke if you see a bigass warning"

doc/Installation/Installation-Ubuntu-1604-Nginx.md
murrant: Sure

murrant:
Feel free to re-open. Not sure any of this is still applicable.

PR #9537, opened 2018-12-12
Escape SNMP & IPMI shell commands

murrant:
hopefully doesn't break anything
Mostly issues with snmp oids and options containing spaces.  Try to remove all of those.

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #10197, opened 2019-05-07
Fix: netstats-udp, tcp and ip rrd's not updating

slashdoom:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/netstats/netstats-ip.inc.php
+++ b/includes/polling/netstats/netstats-ip.inc.php
@@ -26,7 +26,6 @@ if (!starts_with($device['os'], ['Snom', 'asa'])) {
     $data = snmp_getnext_multi($device, $oids, '-OQUs', 'IP-MIB');
 
     if (is_numeric($data['ipOutRequests']) && is_numeric($data['ipInReceives'])) {
-
         $rrd_def = new RrdDefinition();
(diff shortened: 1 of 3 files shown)


slashdoom:
Seems to have broken with PR #9537.  Attempt to fix.

slashdoom:
CLA is signed.

CirnoT:
You have to click link to recheck CLA.

CirnoT:
These changes make no sense to me. Does it actually work for you? All it did for me was to generate empty RRD file with no actual data.

murrant:
The fix the the correct approach (LibreNMS now escapes all snmp arguments).  Need to use an array so each oid gets escaped individually instead of all together.

Just the data isn't formatted as expected after the change.  I'll update netstats-ip.inc.php to show what I think is the correct fix.

slashdoom:
Thanks all.  Updates pushed and seems to working/graphing.

slashdoom:
That makes sense.  So I've changed back to checking $data and changed the checks to is_numeric().  I also moved the if statement up so it doesn't run the for loop if the checks fail.

murrant:
Thanks @slashdoom :+1:

PR #10202, opened 2019-05-09
Added aggregate config option to Billing 95th percentile calculations

llarian0:
This is very small change but it is required for our environment and I'm hoping to be able to upgrade without continually re-merging these files.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Billing-Module.md
+++ b/doc/Extensions/Billing-Module.md
@@ -49,7 +49,7 @@ To change the default option to Aggregate,
 add the following the `config.php`:
 
 ```php
-$config['95th_default_agg'] = 1;  // Set aggregate 95th as default
+$config['billing']['95th_default_agg'] = 1;  // Set aggregate 95th as default
 ```
 
 This configuration setting is cosmetic and only changes the default selected option 
--- a/doc/Extensions/Billing-Module.md
+++ b/doc/Extensions/Billing-Module.md
@@ -41,12 +41,16 @@ daily cleanup tasks.
 
 For 95th Percentile billing, the default behavior is to use the highest of the input 
 or output 95th Percentile calculation.
+
 To instead use the combined total of inout + output to derive the 95th percentile,
+This can be changed on a per basis by setting 95th Calculation to "Aggregate".
+
+To change the default option to Aggregate,
 add the following the `config.php`:
 
 ```php
-$config['billing_aggregate_95th'] = 1;  // Combine in + out throughput before calculating 95th
+$config['95th_default_agg'] = 1;  // Set aggregate 95th as default
 ```
 
-Note that the aggregate option changes how billing summaries are calculated and stored in the database.
-Changing this setting will make any values already stored for the current billing cycle inaccurate.
+This configuration setting is cosmetic and only changes the default selected option 
+when adding a new bill
--- a/doc/Extensions/Billing-Module.md
+++ b/doc/Extensions/Billing-Module.md
@@ -23,7 +23,7 @@ Edit `/etc/cron.d/librenms` and add the following:
 
 Create billing graphs as required.
 
-## Options
+## Data Retention
 
 Billing data is stored in the MySQL database, and you may wish to purge the detailed 
 stats for old data (per-month totals will always be kept).  To enable this, add the 
@@ -36,3 +36,17 @@ $config['billing_data_purge'] = 12;     // Number of months to retain
 Data for the last complete billing cycle will always be retained - only data older than
 this by the configured number of months will be removed.  This task is performed in the
 daily cleanup tasks.
+
+## 95th Percentile Calculation
+
+For 95th Percentile billing, the default behavior is to use the highest of the input 
+or output 95th Percentile calculation.
+To instead use the combined total of inout + output to derive the 95th percentile,
+add the following the `config.php`:
+
+```php
+$config['billing_aggregate_95th'] = 1;  // Combine in + out throughput before calculating 95th
+```
+
+Note that the aggregate option changes how billing summaries are calculated and stored in the database.
+Changing this setting will make any values already stored for the current billing cycle inaccurate.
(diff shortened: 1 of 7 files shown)


includes/billing.php
```diff
@@ -123,6 +123,21 @@ function getLastMeasurement($bill_id)
     return ($return);
 }//end getLastMeasurement()
 
+function get95thagg($bill_id, $datefrom, $dateto)
+{
+    $mq_sql           = "SELECT count(delta) FROM bill_data WHERE bill_id = '".mres($bill_id)."'";
```
murrant: mres() is a do nothing function, please use parameterized queries;
like `dbFetchRows('SELECT * FROM devices WHERE device_id=?', [$device_id])`

includes/billing.php
```diff
@@ -170,12 +187,17 @@ function getRates($bill_id, $datefrom, $dateto)
     $data['rate_95th_in']  = get95thIn($bill_id, $datefrom, $dateto);
     $data['rate_95th_out'] = get95thOut($bill_id, $datefrom, $dateto);
 
-    if ($data['rate_95th_out'] > $data['rate_95th_in']) {
-        $data['rate_95th'] = $data['rate_95th_out'];
-        $data['dir_95th']  = 'out';
+    if( $config['billing_aggregate_95th'] == 1 ) {
```
murrant: We are trying to convert all $config calls to Config::get()

murrant:
Overall looks fine to me, but shouldn't this be a per-bill option?

includes/billing.php
llarian0: That makes sense.  I just cargo culted the existing billing module, but I can refactor it to use your current standards easily enough.  I'll submit a new pull request when that's done.

PipoCanaja: @llarian0  : Or you can update this one directly :)

llarian0: Ah, thanks.  New-ish to public github stuff.

murrant:
`Should this be a per-bill option?` is more important than converting to Config.  Any response to that?

llarian0:
> `Should this be a per-bill option?` is more important than converting to Config. Any response to that?

I'm sorry, I thought I responded to that.  Yes, it probably should be.  I'd guess most environments are all or nothing, but for inclusion in upstream per-bill makes more sense.  The 95th percentile direction column could probably be used for that, if its set to "agg" the poller combines the values, otherwise is uses the existing behavior of setting the direction to in/out and using the max value for the billing cycle.  I'll take a look at that a little more closely later this week to see if there are any pitfalls there, but that saves adding another DB column that probably isn't necessary.

murrant:
That sounds like a good approach.  Funny you can't use "both" because it is varchar(3)...

llarian0:
> That sounds like a good approach. Funny you can't use "both" because it is varchar(3)...

Yeah, I noticed that.  I considered "tot" instead of "agg"

murrant:
agg is fine, you can show whatever in the web ui ;)

murrant:
Looks good, thank for fixing all the SQL injection issues ;)  You can use [] instead of array(), we plan to convert them all at some point. There was still a function naming inconsistency (pre-existing).

This isn't that easy for me to test, but I will try.

doc/Extensions/Billing-Module.md
```diff
@@ -36,3 +36,21 @@ $config['billing_data_purge'] = 12;     // Number of months to retain
...
+or output 95th Percentile calculation.
+
+To instead use the combined total of inout + output to derive the 95th percentile,
+This can be changed on a per bill basis by setting 95th Calculation to "Aggregate".
+
+To change the default option to Aggregate,
+add the following the `config.php`:
+
+```php
+$config['billing']['95th_default_agg'] = 1;  // Set aggregate 95th as default
```
murrant: I think it would be better to allow all three types to be set as the default in the config.
`$config['billing']['95th_default'] = 'agg';`

llarian0: There's only really 2 types.  Max of in/out, or aggregate total.  

The existing codebase sets direction to in or out based on the highest value, so I'm setting that field to 'agg' (which obviously doesn't change) to override that behavior.  

For the standard functionality, a new bill gets set to 'in' by default and then the billing-calculate script changes it based on real values once there's sufficient data to set a max direction for that billing cycle.

That said, the config knob you suggest would probably make the config more readable, so its not a bad idea.  That'd be a pretty simple change

murrant: Sounds good, yeah, somehow there only be two slipped my mind when I wrote this. The in/out makes sense as you don't know which direction is "Down" or "Up".

mctaguma:
Hi all,

I've found a problem in your code, causing Traffic Bill graphs to fail to load, with the following error (and stack trace):
[2019-05-29 11:48:13] production.ERROR: Too few arguments to function getRates(), 3 passed in /opt/librenms/includes/html/graphs/bill/bits.inc.php on line 6 and exactly 4 expected {"userId":6,"email":"user.email@domain.com","exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Too few arguments to function getRates(), 3 passed in /opt/librenms/includes/html/graphs/bill/bits.inc.php on line 6 and exactly 4 expected at /opt/librenms/includes/billing.php:169)
[stacktrace]
#0 /opt/librenms/includes/html/graphs/bill/bits.inc.php(6): getRates('8', '20190528114500', '20190529114500')
#1 /opt/librenms/includes/html/graphs/graph.inc.php(48): include('/opt/librenms/i...')
#2 /opt/librenms/html/graph.php(29): require('/opt/librenms/i...')
#3 {main}
"}

This error is caused by https://github.com/librenms/librenms/blame/0ae1f30f7dbcadd3e97002e5bdaa9215a3eaa7d4/includes/billing.php#L169.

It would appear that "function getRates($bill_id, $datefrom, $dateto)" was changed to "function getRates($bill_id, $datefrom, $dateto, $dir_95th)", without updating "$rates = getRates($vars['id'], $datefrom, $dateto);" at https://github.com/librenms/librenms/blame/master/includes/html/graphs/bill/bits.inc.php#L6.

Can someone possibly fix this? Thanks!

mctaguma:
Thanks! Apologies for not suggesting a specific fix, not familiar enough with the code.

llarian0:
> Thanks! Apologies for not suggesting a specific fix, not familiar enough with the code.

Not a problem, I have no idea how I missed that.  Fixed in my branch.

PR #10207, opened 2019-05-10
tiny script to discover & poll one or more devices

erdems:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Thanks for the script.  I'm not sure how useful it would be as most of the time I only one to do one or the other and it is just little extra typing to do both.  We are generally trying to reduce the number of little scripts and unify everything behind lnms too.

Feel free to tell me why I'm wrong ;) and Thanks!

PR #10210, opened 2019-05-12
Collect Appliance serial number from Sophos device

corsoblaster:
Add serial number read. Added on polling session instead of discovery because the S/N changes in case HA functionality is enabled, and there is a switch between active and standby.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/cyberoam-utm.inc.php
```diff
@@ -26,6 +26,7 @@
 $cyberoam_data = snmp_get_multi_oid($device, ['applianceModel.0', 'cyberoamVersion.0'], '-OQUs', 'CYBEROAM-MIB');
 $hardware = $cyberoam_data['applianceModel.0'];
 $version  = $cyberoam_data['cyberoamVersion.0'];
+$serial   = $cyberoam_data['applianceKey.0'];
```
PipoCanaja: Hello,
It seems that you do not load applianceKey at all in the patch you submitted. May be you created the PR out of copy/pastes, but them you probably missed one :)  
Could you correct the snmp_get_multi_oid, validate it in your environnement, provide the tests and submit it again ? 
Thanx

laf:
If you can update the test data that would be ace. Should be as simple as

https://docs.librenms.org/Developing/os/Test-Units/#additional-module-support-or-test-data

corsoblaster:
Test files uploaded

PR #10001, opened 2019-03-19
Fix transport options when edited on Windows

murrant:
Windows puts line returns as \r\n, when parsing LibreNMS previously left the \r, which could cause issues
Centralize the code so we just have one place to fix.
Try to fix SMSEagle, it had the options wrong (copied from another transport)

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #10221, opened 2019-05-13
removing supurious http, allowing sms eagle transport to work

charlesnwyble:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Smseagle.php
+++ b/LibreNMS/Alert/Transport/Smseagle.php
@@ -44,7 +44,7 @@ class Smseagle extends Transport
             'to' => implode(',', $opts['to']),
             'message' => $obj['title'],
         ];
-        $url    = 'http://' . $opts['url'] . '/index.php/http_api/send_sms?' . http_build_query($params);
+        $url    = '' . $opts['url'] . '/index.php/http_api/send_sms?' . http_build_query($params);
         $curl   = curl_init($url);
 
         set_curl_proxy($curl);


murrant:
Thanks for the PR.

I looked at this a bit and it makes sense if you want to use https.  However the input field is labeled "Host" in the ui, so that needs to change.  You should also set the validation type to url and include code to add http:// to existing ones so they don't break.

charlesnwyble:
I think you mis-understand the issue. The complete URL is already in the database. The hard coded http I removed is spurious. Since the code already pulls the complete URL from the database, the hard coded http is spurious. 

Sorry to repeat myself above, I just want to communicate the nature of my fix as clearly as possible. 

The UI code when setting up the transport forces entry of either http or https. This is fine, makes sense etc. It stores the full URL in the database. Sure makes sense. Since the full / complete URL is in the database, no need for protocol to be hard coded in the code.

charlesnwyble:
I spent a couple days chasing this down. Tracing through UI/transport code, database, json decoding etc. I performed each step of the process manually in isolation to fully debug the issue. In a previous version of the code, someone had just copy/pasted the playsms values, so the query itself was wrong. I see that is now fixed, so that's good. 

Please try to fully understand the issue, and the amount of work that went into tracking it down.  I performed a very thorough investigation and troubleshooting process to arrive at this conclusion. The hard coded value is spurious, simply remove it. These things happen, it's alright. Beauty of the open source process, anyone can come along and submit a fix.

charlesnwyble:
Yep I see the fix was pretty recent to fix the transport copy/paste error:

commit ac1cd1296ba0d3d1cbca11ec93afbe7431234779
Author: Tony Murray <murraytony@gmail.com>
Date:   Wed Mar 27 07:18:02 2019 -0500

    Fix transport options when edited on Windows (#10001)

    Windows puts line returns as \r\n, when parsing LibreNMS previously left the \r, which could cause issues
    Centralize the code so we just have one place to fix.
    **Try to fix SMSEagle, it had the options wrong (copied from another transport)**

murrant:
I understand and appreciate the amount of work to track down a bug.  (trust me)

But the fix needs to:
A. not fix a misconfiguration by the user
B. not break existing settings
C. try to prevent user misconfiguration

When I added the transport via the webui, this is what I get in my db:
image

So, adding http:// looks correct.

charlesnwyble:
I hadn't run git pull on my production instance. With the latest code (and adding the transport fresh) it's working with the leading http in the code.

murrant:
@charlesnwyble I think there still needs to be a change because this is allowed in the menu:

image

But then the transport won't work.

charlesnwyble:
Right. That's the exact bug I was experiencing. With the "old" code 

 git log -1
commit 8926d7cfe6e32606f4b5bcee64ea76a3f1ae7e84 (HEAD -> master, origin/master, origin/HEAD)
Author: Tony Murray <murraytony@gmail.com>
Date:   Wed Apr 10 16:33:25 2019 -0500

I was forced to put the protocol prefix. With the new code, it's not required (but still allowed). So the right/allowed version has to be picked by the team and implemented.

murrant:
Ah, I see.

I still think something like this needs to be done: https://github.com/librenms/librenms/compare/master...murrant:smseagle-allow-url

charlesnwyble:
It's really a design decision to make. I think forcing the protocol and storing value in the DB and removing the hard coded protocol from code is the way to go. Hence my pull request. It would revert back to the old behavior. For now (to allow me to run daily.sh) I just added the transport without protocol.

murrant:
Yeah, I agree.  But you can't just break the people that are working :)  I'll send those changes as a PR.

murrant:
https://github.com/librenms/librenms/pull/10223

PR #10231, opened 2019-05-15
ArubaOS - fix client count polling, add ap count polling.

twilley:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/arubaos.yaml
```diff
@@ -3,8 +3,10 @@ text: ArubaOS
 type: wireless
 icon: aruba
 over:
-    - { graph: device_arubacontroller_numaps, text: 'Number of APs' }
-    - { graph: device_arubacontroller_numclients, text: 'Number of Clients' }
+    - { graph: device_wireless_ap-count, text: 'AP Count' }
+    - { graph: device_wireless_clients, text: 'Client Count' }
+    - { graph: device_arubacontroller_numaps, text: 'Number of APs (Legacy)' }
```
murrant: Is this worthwhile to leave?  Maybe we should have a check if the rrd exists before displaying the overview graph?

PipoCanaja: @murrant That would be nice indeed !

kkrumm1:
I did a test on this with my controller. Worked great!

includes/definitions/arubaos.yaml
murrant: I took a look and it is not feasible with the current code. :(  Would take some major re-writing of graphing.

PR #10239, opened 2019-05-20
add tripplite ups support for battery charge, est. runtime, output st…

peelman:
…ate, output current, and temperature.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


peelman:
Screen Shot 2019-05-20 at 8 44 06 AM

murrant:
Hmm, I wonder if that will break any existing devices (with old firmware)

peelman:
It didn't have any ill effects on our TrippLite ATSes or PDUs (which use the same Object ID, perversely), whether they were the old style SNMPWEB or the WebLX cards.  Pretty limited test pool, admittedly.  

For the record, the unit we're monitoring is a giant 3-phase 60kW behemoth.

murrant:
Let's get some test data in the system so we know a little better :)

peelman:
Working on this but my dev box is waaaay out of date, so its going to be a bit before I can get it up to speed and then have time to go back and generate all the test stuff.

murrant:
You can capture the snmprec files in production easily.
1. Apply patch: ./scripts/github-apply 10239
2. Capture Data: ./scripts/save-test-data -h <hostname>
3. Remove patch (optional)
4. upload snmprec
5. I generate the test json for you :D

includes/definitions/poweralert.yaml
```diff
@@ -1,10 +1,16 @@
...
 type: power
+mib_dir:
+    - poweralert
 over:
     - { graph: device_current, text: Current }
+    - { graph: device_voltage, text: Voltage }
+    - { graph: device_runtime, text: Runtime }
 icon: tripplite
-rfc1628_compat: true
+rfc1628_compat: false
```
murrant: does this remove sensors from existing devices?

laf:
@peelman Any update about the backwards compatibility comment from murrant?

peelman:
right now i do not. i can flip that back on mine and see.

peelman:
> @peelman Any update about the backwards compatibility comment from murrant?

Ok; so I have kicked it back on for my install and I'm remembering why I turned it off.

It gave a few additional sensors out of the box, including Bypass Voltage, Input Voltage, Bypass current, Output current, and battery run time.  The problem I ran into is that the LX controllers (I think I don't have a fleet of these to test with) didn't respond to the voltage endpoints and I think I shut them off while I was working to avoid confusion.  It shows 1.17kV for the Bypass and Input Voltages.  I think my intention was to flesh out the model a little more before we merged and add back in things, but things have snowballed here from a project (both quantity and complexity perspective) and I haven't been able to even look at Libre from a dev standpoint in weeks.

To confirm, it did remove sensors (voltage, current, frequency, and state) from other TrippLite devices we have in the network, like our ATSes.

The newer LX-based cards seem to use `.1.3.6.1.4.1.850.1.1.1` as their object ID, whereas the older cards use `.1.3.6.1.4.1.850.1`.  Should we add an OS for TrippLite LX (Linux-based) cards going forward?  I'm not seeing a good way to square this circle with the inconsistency of the RFC1628 bits across models.  Right now the only other rack-mount unit we have in service is showing offline; I had some people go reseat the card, but its not coming back; so I'm wondering if it'll have to be reconfigured manually.  I'd say flag this as needing further testing; what is here is working fine for us, so I don't think there's anything actively wrong with the new bits, but removing the old bits may be problematic for some users.

murrant:
Hmm, if I remember right Tripplites are a bit wonky.  The SNMP card just proxies data from the UPS, but it doesn't have any way to figure out what UPS is attached.

One option is to leave both, but that may result in duplicate sensors if the device presents both data sets.

If they use different MIBs, it might be appropriate to separate them into different OS.

murrant:

https://community.librenms.org/t/tripplite-ups-tlnet-card-support/10235/11

Bowl2much:
Is there anything I can do to help?
I have both the older WEBSNMPCARD in a SRCOOLNET12K i posted about [here](https://github.com/librenms/librenms/issues/10979) and newer card in a PDU.
Neither show any health values in LibreNMS. I also have external sensors plugged into both.
What I am really saying is I could test both the old and new at the same time.

laf:
Closing due to lack of activity.

peelman:
Sorry; I have not had any cycles to contribute on this.  I stand by my comment from August that the LX stuff probably needs to be defined separately from the old legacy bits.  I can merge some code and test if anybody else has the cycles, but I am slammed with DayJob™ stuff for the foreseeable future, and any dev cycles I can find will be focused on internal applications.

Bowl2much:
Im still here!
I would love to see some progress on this issue, but I lack SNMP and development skills to help contribute to fix these issues. I just dont know where to jump off to even start to help.
But I have these devices to test against.

murrant:
This came back up again. Looks like you cannot use yaml and need to use PHP.  That way you can fall back to UPS-MIB when the tripplite oids aren't available.

peelman:
@murrant are there examples of this somewhere in existing code?

murrant:
includes/discovery/sensors

PR #10243, opened 2019-05-21
Device groups list only related to the user permission

corsoblaster:
Modify the function GetDeviceGroups() to return the Device group list only according to the current user. This prevent to list a device group, if doesn't contain any device allowed to access.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/device-groups.inc.php
+++ b/includes/device-groups.inc.php
@@ -214,13 +214,22 @@ function GetDevicesFromGroup($group_id, $nested = false, $full = '')
 }//end GetDevicesFromGroup()
 
 /**
- * Get all Device-Groups
+ * Get all the Device groups which the decive are associated to the current user
  * @return array
  */
 function GetDeviceGroups()
 {
-    return dbFetchRows('SELECT * FROM device_groups ORDER BY name');
-}//end GetDeviceGroups()
+    $user_id = '';
+    
+    if ((LegacyAuth::user()->hasGlobalRead()) or (LegacyAuth::user()->isAdmin())) {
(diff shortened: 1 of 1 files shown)


includes/device-groups.inc.php
```diff
@@ -214,13 +214,22 @@ function GetDevicesFromGroup($group_id, $nested = false, $full = '')
...
 /**
- * Get all Device-Groups
+ * Get all the Device groups which the decive are associated to the current user
  * @return array
  */
 function GetDeviceGroups()
 {
-    return dbFetchRows('SELECT * FROM device_groups ORDER BY name');
-}//end GetDeviceGroups()
+    $user_id = '';
```
murrant: Have you tried this instead of all the code in this function?:
```
return \App\Models\DeviceGroup::hasAccess(\Auth::user())->get()->toArray();
```

corsoblaster: You code is smarter than mine, but same wrong effect.
For some reason, a normal user can see as many duplicate menu items under Devices / Device groups, as many device belongs to each group.
I tried also to add a GROUP BY to my query, but the effect is the same.
Which PHP file manages the menu composition?

murrant: app\Models\DeviceGroup.php   Looks like the code there is still using a join... which is why you get duplicates.  I'll try to fix that.

murrant: Also to fetch data for the menu is `MenuComposer.php` and the template is `menu.blade.php`

PR #10248, opened 2019-05-22
Remove DB credentials from config.php

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


Jellyfrog:
Should update the documentation in a couple of places:
- https://github.com/librenms/librenms/blob/77d94a458b0eda78a560789e70b540a8306bb1a2/doc/Extensions/Authentication.md#mysql-authentication
- https://github.com/librenms/librenms/blob/06b85f5b074b94472d0c6f6561f438321d21d0b3/doc/Support/FAQ.md#where-do-i-update-my-database-credentials
- https://github.com/librenms/librenms/blob/89fae9be1d9357e40d5b9aaff251f72f535b74d9/doc/Support/Configuration.md#database-config

Jellyfrog:
Does it fail because it doesn't set `DBTEST`? https://github.com/librenms/librenms/blob/master/config/database.php#L29

app/Http/Middleware/CheckInstalled.php
```diff
@@ -38,7 +38,7 @@ class CheckInstalled
      */
     public function handle($request, Closure $next)
     {
-        if (!file_exists(base_path('config.php')) && !$request->is('install.php')) {
+        if (!file_exists(base_path('.env')) && !$request->is('install.php')) {
             // no config.php does so let's redirect to the install
```
laf: Need to update this reference as well.

html/install.php
```diff
@@ -354,64 +354,35 @@
...
-\$config\['db_user'\] = '$dbuser';
-\$config\['db_pass'\] = '$dbpass';
-\$config\['db_name'\] = '$dbname';
-\$config\['db_socket'\] = '$dbsocket';
+DB_HOST=$dbhost
+DB_DATABASE=$dbname
+DB_USERNAME=$dbuser
+DB_PASSWORD="$dbpass"
 
-// This is the user LibreNMS will run as
```
laf: Should we still not create the config.php file (or generate and show it at least)? It provides some good standards people can update.

laf:
LGTM. Are the tests failing because the old tests used to use root as the MySQL user and this seems to now fallback to librenms?

I have tested some areas of the webui, poller, discovery, alerts, billing and api with no issues.

murrant:
Something with how php artisan dusk pulls in the environment differently...

SourceDoctor:
@murrant @laf 
not sure if it is useful to move DB credentials away from config.php into a hidden file.

also no action since 4 month. Still relevant?

murrant:
@SourceDoctor the point is to not keep the credentials in two files.  Eventually, I hope to make config.php optional.

SourceDoctor:
i'm full with you on reducing redundancy. Especially in configuration cases.
But a hidden file like '.env' doesn't seem to be a good way in my eyes.
Maybe a database.conf with only db credentials in it?

with all global options in GUI it's a good way to this goal :-)

nevertheless i think it should be still possible to handle special configurations via config.php or a similar file. Thinking about configurations which shouldn't be handled in normal cases
e.G.
```
$config['os']['netgear']['bad_snmpEngineTime'] = true;
$config['os']['apc']['bad_hrSystemUptime'] = true;
```

Jellyfrog:
`.env` is from Laravel, imo we should move everything there sooner or later :)

SourceDoctor:
@Jellyfrog working on it ;-)

Jellyfrog:
@murrant guessing this failed due to `php artisan serve --env=dusk.testing` guessing it should be `php artisan serve --env=testing` or nothing at all. On phone so can't verify

murrant:
@Jellyfrog fixed it, but I should re-test the install process to double check.

murrant:
.env is for settings that need to be set before LibreNMS is booted.
@SourceDoctor you can still set those like this `./lnms config:set os.netgear.bad_snmpEngineTime true`

SourceDoctor:
@murrant 
how about put those special stuff like:

> $config['os']['netgear']['bad_snmpEngineTime'] = true;
> $config['os']['apc']['bad_hrSystemUptime'] = true;

into Global Settings,
but how to get an actual list of os in, and maybe reduce to a list of possible keys per os only?

murrant:
We can have special checking for settings under os. all OS support the same settings. (we already have yaml schema validation for all (almost) settings.

murrant:
Incorporated into #11810

PR #10252, opened 2019-05-23
Instructions for logstash

mjducharme:
Added instructions for how to configure logstash to send syslog to LibreNMS


murrant:
Hmm, I would much rather have integration like Graylog.  This duplicates and stores logs in two locations.

PR #10261, opened 2019-05-24
enhance smart to show power_on_hours also

SourceDoctor:
Power on hours on smart will be monitored with this patch


SourceDoctor:
i think it's an useful enhancement for smart application :-)

murrant:
You can't modify the rrd defintion like that.  rrds can't be changed after creation.  The best way is to create a new rrd defintion just for that one data point.

SourceDoctor:
now new rrd for id9 enhancement will be created

SourceDoctor:
@murrant something else to change for merging?

PR #10263, opened 2019-05-25
Add MPLS Support

vitalisator:
This introduce the generic MPLS support
- New database tables for MPLS LSPs and MPLS LSP Paths
- Actually discovery and polling on NOKIA SR OS (Timetra) devices
- New routing->MPLS HTML pages

_Global MPLS view_
image

_Device MPLS view_
image

Suggestions and improvements are welcome

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/defaults.inc.php
```diff
@@ -487,6 +487,8 @@
 // Poller/Discovery Modules
 $config['enable_bgp'] = 1;
 // Enable BGP session collection and display
+$config['enable_mpls'] = 1;
```
murrant: I prefer to use the modules to enable/disable things.

vitalisator:
@murrant, have not really idea how to do "new style" or even where to start.
I would give it a try if you give me some advice :-)

Can I just put my code inside public functions poll and discover? 
But I think there is more to do...

murrant:
Basically, but you would be missing out if you don't use Eloquent to update the DB, it replaces 90% of the code...  You could convert it in steps.

Since you've added test data, I could take a crack some time.

murrant:
@vitalisator ok, I updated this to the new style (and improved the infrastructure for that a bit)  Can you test it?

I did change the migration because vRtrMplsLspPathStatTable is not in the test data so, trying to save nulls to the DB for that data caused the queries to fail.  Perhaps more or less fields should be nullable?

PR #10265, opened 2019-05-26
Fix for RouterOS LLDP discovery

mjducharme:
Only the first nine neighbors were correct on RouterOS - missing hexdec() led to the remaining neighbors being linked to the wrong ports. Ex. port 10 in hex was incorrectly linked to port 10 decimal when it should be 16 in decimal

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hi @mjducharme 
Looks good so far. Could you also generate some test data (new variant for your family of device) with a few LLDP neighbors at least ? The more feature enabled you have the better (meaning the testdata will cover a wider part of the RouterOS code). 
Thx

mjducharme:
Test data submitted

PipoCanaja:
> Test data submitted

I don't see any neighbor data in it ? Could you run a test data collection with at least one or 2 neighbors ?

mjducharme:
That device has 12 neighbors. You can see them in the .snmprec. I was surprised that they do not show up in the .json file but since I've never done this before I thought that might be normal. Are they supposed to appear in the .json file somehow?

PipoCanaja:
Most probably because the neighbors are default disabled. So they are not tested. At least the snmprec is here.

PR #10274, opened 2019-05-29
Wago automation controller support + additional 'concentration' sensor

amigne:
Support for Wago automation controller, with addition of a new 'concentration' sensor type.

Wago manufactures programmable programmable automation controllers that can be used to retrieve data from misc sensors, such as temperature, humidity, CO2 concentration, contacts, ...

This PR adds support for such Wago PFC controllers and includes a new 'concentration' sensor type for gaz/particles concentration (expressed in "ppm" unit).

The Wago PFC returns raw data in an array without any meta-information describing the type of measured value or description of the origin of the data. This has to be manually set in the config.php. This is not very clean, but this is the only solution that permits to add some meaning to the collected SNMP values.

The code in this PR is working, but it is still considered as in progress. The following has to be done before being approved and merged:
- [x] Add some meaning for 'state' type values;
- [x] Fix the Health / Overview page for displaying 'State' graphs;
- [x] Fix the Health / Overview page for displaying 'Concentration' graphs;
- [x] Add display of 'Concentration' sensors on the device Overview page;
- [x] Add test data for this new OS;
- [x] Create a documentation for the $config['os']['wago']['plc'][] settings.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
@amigne I know it might be more than you want to tackle.  But maybe LibreNMS needs a way to have sensors of "unknown" type which are later assigned a type by the user in the webui...  I know of a few more devices that have external sensors that do not report the type of sensor connected via snmp.

amigne:
> @amigne I know it might be more than you want to tackle. But maybe LibreNMS needs a way to have sensors of "unknown" type which are later assigned a type by the user in the webui... I know of a few more devices that have external sensors that do not report the type of sensor connected via snmp.

@murrant This sounds to be an easier way to fix this issue (missing of sensor type from the device) from the user point of view than playing with the configuration file. But I see a point anyway: having only the webgui as a way to set sensor types would be quite a nightmare if the user has about 50 devices with 10-15 sensors each (this is my case!). It would then need to manually edit the settings for 500-750 sensors through the webgui, which is not very convenient. If devices and sensors deployment is done with some reasonable thinking, it may be much easier to define a generic setting using the configuration file and to handle exceptions either with configuration file (using regexp), or with the webgui.

As I'm currently writing this module for my business, I propose to continue my current development based on the configuration file only... Once it is finished (and after my summer holidays are over ;-) ), I can start working on the webgui to manually define the sensor type for these devices.

murrant:
Sure, but I don't think the config setting is going to be useful for most people.  What happens when you have a single device in a different configuration?

An option would be to have a button that says apply these sensor types to other devices (when clicked it shows a selectable list with all devices of the same OS)

amigne:
> An option would be to have a button that says apply these sensor types to other devices (when clicked it shows a selectable list with all devices of the same OS)

Good idea : This is noted down for the upcoming webgui dev.

amigne:
An issue still needs to be fixed : the discovery script creates new records in the 'sensors' DB table.

amigne:
Issue fixed and successfully tested.

doc/Support/Configuration.md
```diff
@@ -715,6 +715,93 @@ $config['ipmi']['type'][] = "open";
 
 [Distributed Poller](../Extensions/Distributed-Poller.md)
 
+### Wago data collection
```
murrant: This should be under doc/Support/Device-Notes/

murrant: Also, I still wouldn't mind a general approach.

amigne: Well Tony, I'll move the doc at the right place. Thank you for the feedback.

Regarding the general approach, this is in my TODO list, but this needs more thinking to design a really generic form that could be used for this Wago device, as well as any new device. I'll start working on it as soon as I'll be back from holidays. 🌞

Jellyfrog:

PR #10285, opened 2019-05-30
Don't indicate that host is always up on uptime counter when host is ignored

CirnoT:
When host is ignored, near up/down counter, don't show green, instead show gray.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/html/pages/devices.inc.php
+++ b/includes/html/pages/devices.inc.php
@@ -318,6 +318,8 @@ if ($format == "graph") {
                         return row.uptime;
                     } else if (row.status == 'down') {
                         return "<span class='alert-status-small label-danger'></span><span>" + row.uptime + "</span>";
+                    } else if (row.status == 'ignore') {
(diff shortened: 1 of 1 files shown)


murrant:
Isn't gray for disabled?  Also, isn't ignore primarily for alerts?  ignored devices are still polled so we know if they are up or down.

CirnoT:
Current logic as it is (before this PR) always displays green, because we don't really know state of device when it's ignored (as fetched from ajax request) - it's simply "ignored" instead of either "down" or "up". Because logic is as follows: If down -> red, else green, ignored devices are always green even if they are down.

murrant:
Why wouldn't we know the state of the device when it is ignored?

CirnoT:
Technically we know state, but we don't expose it in response to this specific request. Do you think we should?

murrant:
Yeah, I don't think we treat ignored devices differently.  Ignore applies to alerts.

CirnoT:
Okay, but now it means that when host is ignored and is down, it still indicates that host is up.

PR #10287, opened 2019-05-30
Use sensor labels for overview/inventory pages, refactor some html-page related code

martijn-schmidt:
Hey folks,

TLDR: this pull request mostly contains some clean-up work to move similar code that was replicated across several html pages into functions.inc.php, and it enables new-style labels for the device overview and device entity-physical/inventory pages.

============

- Moved label span creation to get_sensor_label_color() since the same code was used repeatedly on multiple different pages.

- Removed function get_unit_for_sensor_class() since it replicated function get_units_from_sensor() feature-wise. Modified a single page that was using the removed function.

- Move state translation and corresponding database calls to the get_state_label() function:
Several html pages used similar database calls and sometimes very different methodology to determine the state label and state text before calling get_state_label(), so I moved that part of the task into the function itself instead of replicating the same code multiple times on different pages.

- Refactored the device overview page to use get_state_label() and get_sensor_label_color() functions, as a "side effect" it now uses the new style labels for the non-state health sensors as well. That makes it consistent with the device health page and the LibreNMS-wide sensor table.

- Refactored some if/else statements into ternary operators for simplicity's sake.

- Enabled state translations and new style labels for the device entity-physical/inventory page.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 10287`


murrant:
Thanks for this @martijn-schmidt , I have some conflicting work at #10298
I'm moving the descriptions and units into the language files for translation and easy organizing :)

martijn-schmidt:
Hi @murrant - I've made some new commits that resolve the merge conflict and fix two typos in the language files. :) 

Out of curiosity, I see that the calls to function `get_units_from_sensor($sensor)` are replaced with `__('sensors.' . $sensor['sensor_class'] . '.unit')` but I couldn't find the corresponding code for `__()` - I'd love to take a closer look at what it does, if you don't mind pointing in the general direction of where to find it?

martijn-schmidt:
@murrant do you need anything else for this one? Because PR #10327 which was merged today has documentation which "depends" on the fact that we replaced get_unit_for_sensor_class() with the Laravel translation code for sensor class units (eg it omits a reference to the old function that was deleted here in PR #10287).

includes/html/functions.inc.php
```diff
@@ -1610,92 +1610,57 @@ function get_device_name($device)
  * Returns state generic label from value with optional text
  */
 
-function get_state_label($state_value, $state_text_param = null)
+function get_state_label($sensor)
 {
-    switch ($state_value) {
+    $state_translation = array();
+    $state_translation = dbFetchRows('SELECT * FROM state_translations as ST, sensors_to_state_indexes as SSI WHERE ST.state_index_id=SSI.state_index_id AND SSI.sensor_id = ? AND ST.state_value = ? ', array($sensor['sensor_id'], $sensor['sensor_current']));
```
murrant: You probably want dbFetchRow() here, that way you don't have to do the silly [0] stuff below

includes/html/functions.inc.php
```diff
@@ -1610,92 +1610,57 @@ function get_device_name($device)
...
-        'waterflow'            => 'l/m',
-    );
-
-    if (!array_key_exists($class, $units_by_classes)) {
-        return '';
-    }
+    $unit = __('sensors.' . $sensor['sensor_class'] . '.unit');
 
-    return $units_by_classes[$class];
+    $sensor_current = "<span class='label $label_style'>".trim(format_si($sensor['sensor_current']).$unit)."</span>";
```
murrant: You could probably inline this variable here and just return the string

includes/html/functions.inc.php
martijn-schmidt: Updated that, thanks for the tip!

includes/html/functions.inc.php
martijn-schmidt: Updated that as requested for both the get_state_label() and get_sensor_label_color() functions!

martijn-schmidt:
I've taken care of both requested changes, thanks for the tip :)

murrant:
Now for a design question.  Why did you include the raw state value in the webui?  I'm not sure how that would be useful.

martijn-schmidt:
The overview was the only page which didn't already show the raw state value in the label:
`includes/html/pages/device/overview/generic/sensor.inc.php`

The inventory page didn't do state translation at all before and only showed a raw number, which triggered me to begin investigating this part of the code:
`includes/html/pages/device/entphysical.inc.php`

All other pages already displayed it as `state translation (raw value)` and continue to behave exactly the same way as before I refactored the code:
`includes/html/pages/device/health/sensors.inc.php`
`includes/html/pages/device/port/sensors.inc.php`
`includes/html/table/sensors-common.php`

I wanted to make things consistent and figured that more information/visibility is usually better, moreover, I don't know whether someone might be using the raw value for things which could cause issues when we take it away.

Finally, the rrdtool generated graphs don't do state translations right now, so it might be helpful to know the raw value it's currently at to see whether there have been recent changes.

murrant:
I don't think the raw value is that useful.  However, perhaps adding a tooltip that shows the raw value would be a good compromise ;)

martijn-schmidt:
As I said I'm not particularly attached to the raw value, so I've removed it. ;) In effect there's a "tooltip" already in a sense that you can see small versions of the rrd graphs when you hold your mouse over the label, so that is then good enough.

Maybe we can experiment with adding the state translation info for that sensor in the graph legend, but I'd say that's something for another PR.

PR #10307, opened 2019-06-04
Cisco UCOS Version and Feature Fix

djamp42:
The current OS UCOS encompasses many different Cisco products, this PR is attempting to get the Version and Product name under this OS. Cisco does not provide specific OID's for this information, so this is looking at the installed packages from MIB (SYSAPPL-MIB) to determine what Version and Product we are dealing with. I have tested this with many different products versions and seems to be fine. 

The nested looping of the SNMP data isn't ideal, maybe someone can give me a pointers on that one. 

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os/ucos.inc.php
```diff
@@ -11,3 +11,21 @@
  */
 
 $hardware = snmp_get($device, 'sysObjectID.0', '-Osqv', 'SNMPv2-MIB:CISCO-PRODUCTS-MIB');
+
+$version = snmp_get($device, '.1.3.6.1.2.1.54.1.1.1.1.4.1', '-Osqv');
+
+$applistraw = snmpwalk_cache_oid($device, ".1.3.6.1.2.1.54.1.1.1.1.3", array());
```
murrant: You could actually just do a regular snmpwalk: `snmp_walk($device, '.1.3.6.1.2.1.54.1.1.1.1.3', '-OQv')`
Then use text tools like str_contains() or preg_match() to check for strings.

djamp42: Fixed, Much better, thanks!

PR #9179, opened 2018-09-10
Refactor datastores to classes

murrant:
Initial unit tests.  (some may be false positive)
Old functions shimmed to classes
Added Config::forget so I could use it for tests

This requires php-sockets for many non-rrd based data stores.

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

PR #10316, opened 2019-06-06
Enhance Prometheus tags processing

blufor:
Tags provided via URLpath encoded values are now properly encoded.

Also, prometheus pushgateway can't handle `/` in tags (which is frequent in cisco interface names). This applies only when using this method (and the `%2f` tricks can't be done here either, because of Go's URL lib). All slashes are now replaced with `-`.

Proper solution would be to provide all the tags/labels in the request body with the metrics, ie:

```
INOCTETS{measurement="ports",ifName="Gi1/0/1",...} 10000
```

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/prometheus.inc.php
+++ b/includes/prometheus.inc.php
@@ -23,19 +23,19 @@ function prometheus_push($device, $measurement, $tags, $fields)
                 set_curl_proxy($ch);
                 $vals = "";
 
-								# Format Prometheus tags for inclusion with metrics in the body
-								#
-								$promtags = array('measurement="' . $measurement . '"');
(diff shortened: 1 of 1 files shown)


blufor:
So in the end, I've managed to implement the better tagging option, which handles the metric tagging nicely, including slashes :)

murrant:
Any concern about increasing the data transferred since you now print the tags for every measurement?

Note to self to update #9179

murrant:
Sending more data doesn't seem like the best solution.  If you can make sure it is not worse, let us know.

PR #10324, opened 2019-06-07
sort Device Applications in Optionbar alphabetic

SourceDoctor:
assigned applications in device are sorted alphabetic now


SourceDoctor:
hu, what's the reason for failing tests?

murrant:
Update to chrome broke the tests... re-running them now that it is resolved upstream.

PR #10326, opened 2019-06-08
MRV: temperature limits from MIB

martijn-schmidt:
I received the latest MIBs from the vendor - version 5.14.1 - and that contained nbsCmmcSlotTemperatureMin and nbsCmmcSlotTemperatureLimit which now no longer have to be guessed.

Test results for snmpsim and friends didn't change because the OS version in my chassis doesn't support the new OIDs yet, so even though the yaml discovery tries to read them it'll return a null for now. I will provide new test results after we've upgraded our chassis.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 10326`


murrant:
Looks good, but you need test data and the MIB files should be renamed to the name of the MIB.

murrant:
Thanks, if you can add an snmprec file, I can generate the json test data.

martijn-schmidt:
Hey, sorry I forgot to comment but the older firmware that's running on my MRV boxes doesn't respond to the two new OIDs yet (it just returns an empty response now). 

That's why, when I make a new snmprec and json test data file, there is no difference from the previous one which I committed during the last MRV related PR.

I'm currently trying to get my hands on an unused MRV chassis so I have a lab device which can be upgraded without a maintenance window. If that works I'll be able to collect snmprec and json test data and update the PR.

murrant:
we can just add the oids to the snmprec of the existing

PR #9667, opened 2019-01-14
Add group to sensor 1st discovery + template syntax {{ $xxx }}

PipoCanaja:
- Groups are now discovered with the 1st sensor discovery (insert) and not only on the 2nd and more (sensor update).
- Adapt the impacted testdata accordingly.
- The group name can contain `{{ $varName }}` to build it out of the precached data thanks to `YamlDiscovery::replaceValues()`.

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #10327, opened 2019-06-08
Remove guessed limits for some health sensors, documentation for sensor classes

martijn-schmidt:
Original behaviour
===================

The file `includes/discovery/functions.inc.php` contains `sensor_limit_low()` and `sensor_limit()` which both attempt to guess a sane value for sensors when no explicitly defined low_limit or high_limit can be found during discovery.

Both switch control structures used in those two functions have empty case statements which means that if one of those matches, it's going to fall through and run the code for each subsequent case until a `break` is reached. See also: https://softwareengineering.stackexchange.com/a/162582

For example, when we call `function sensor_low_limit(dbm, -13.036)` it will return the value `-12.3842` instead of `null`. That is because there will be a match at `case 'dbm':` which falls through all the way to `case 'cooling':`, where it performs `$limit = -13.036 * 0.95` before hitting a `break`.

Changed behaviour
===================

I want to remove `power_consumed` and `count` guessed low_limit and high_limit, I personally added those sensor classes in PR #9471 when I didn't understand that a switch control structure has fall-through behaviour so I can guarantee that guessing limits for those is a mistake on my behalf. It should not be there, only power_factor can have guessed limits. Apologies for the issue, I'm still a beginning programmer. :smiley: 

Furthermore, I want to remove guessed high_limit values for `current` and `power` because these are supposed to draw higher values as more devices or components are installed on for example a PDU or a chassis.

Finally, I want to remove guessed low_limit and high_limit for `dbm` sensors, there is a much too large variance in power budget on commercially available optical transceivers for there to be a sensible window where you can guess these values. admiralspark on Discord is reporting the same problem, like myself he is running some DWDM optics over a larger distance where variances in signal strength are common.

Refactoring
===================

I have removed all cases that would have returned `$limit = null` and replaced them with a default that performs a `return null` directly. This shortens the function significantly which makes codeclimate's 25-lines-per-function check happy, lowers the risk of unintentional changes in the future, and I have also left a comment to warn other beginning programmers like myself about unintentionally hitting the fall-through behaviour.

Because there is no documentation on how to add a sensor class at all right now, I've gone ahead and written some documentation to make it easier to find the availability of these guessed functions. It also saves the next guy/gal that wants to implement a sensor class some time grepping through the code, so that's positive thing as well.

I didn't reference adding the unit in `get_unit_for_sensor_class()` from `includes/html/functions.inc.php` in the new documentation since that functionality is being moved to Laravel localization files like `resources/lang/$lang/sensors.php` in one of my other pull requests - PR #10287.

Bugfix
===================
There are a lot of cases in the latest test data where the "group": null, has changed to "group: "", for sensors that were discovered with yaml as a result of a change made in PR #9667:

https://github.com/murrant/librenms/commit/8e80121c33c4912162372b7348423c6a811f8d08#diff-67aaf7dd786395ce4cbd2c3996cf1bdcR1035

The original yaml discovery code had `"group": null` as a fallback, and the discovery function `discover_sensor()` still has `$group = null` as default value if nothing else is specified, so I think it might be a bug. I have updated the test data for groups along with the bugfix that makes sure yaml discovery group processing reverts to null if nothing truthy was returned.

Running a quick `grep -ir "\"group\": \"\"," | uniq | wc -l` in the tests/data/ directory shows that around 61 test files are affected for what's currently in the master branch.

It looks like this also updates rounding for some sensor limits and current values.

Cleaning up past mistakes?
===================

Since both functions are only called during discovery OR when the existing limit is null, there are going to be a lot of stale/bogus low_limit and high_limit values in existing LibreNMS deployments. Do you have any thoughts on how to perform a one-time clean-up of the limits for all the affected sensor classes that weren't manually overridden by the user? Maybe a database migration could do the trick? As long as the clean-up isn't repeated those sensors which can pull sensible low_limit and high_limit values from SNMP will re-learn them during the next discovery run after the LibreNMS update.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 10327`


martijn-schmidt:
It looks like this PR is causing a lot of unit tests to fail due to the fact that we are no longer erroneously making up low_limit and high_limit values for certain sensor classes. I'll run `./scripts/save-test-data.php -m sensors` and commit some updated json test data.

martijn-schmidt:
There are a lot of cases in the latest test data where the `"group": null,` has changed to `"group: "",` for sensors that were discovered with yaml as a result of a change made in PR #9667:

https://github.com/murrant/librenms/commit/8e80121c33c4912162372b7348423c6a811f8d08#diff-67aaf7dd786395ce4cbd2c3996cf1bdcR1035

@murrant @PipoCanaja do you know if that is intentional and whether we should update the test data for groups along with the discovery function `discover_sensor()` which still has `$group = null` as default value if nothing else is specified?

Or is it better to update the function `discovery_process()` so yaml discovery group processing reverts to null if nothing truthy was returned? Could be as simple as appending `:? null` to the following line:

https://github.com/librenms/librenms/blob/master/includes/discovery/functions.inc.php#L1078

Running a quick `grep -ir "\"group\": \"\"," | uniq | wc -l` in the `tests/data/` directory shows that around 60 test files are affected.

martijn-schmidt:
I've been thinking about this some more.. in some cases the switch fall-through behaviour may have been used intentionally, for example current and power are directly related to each other. Of course the power draw (and therefore the current) can change during normal operation so that shouldn't be a guessable limit, but there may be others where this is a valid consideration.

The dbm sensor class also certainly shouldn't have guessable limits, optical transceivers can have very wide acceptable transmit/receive windows (for example, one of our 10G-ZR DWDM optics may according to its specifications transmit between 0 and 5 dBm and receive between -23 dBm and -6 dBm). And with rx/tx dbm signals sometimes hovering around zero, the 0.95 calculation may result in a very small guessed window which causes constant alerting.

Having added the power_consumed, power_factor, and count sensor classes myself I know for sure that only power_factor can have a sensible guessed limit, the other two are mistakes on my behalf due to not knowing about the switch fall-through behaviour. We also know the empty cases at the end of the switch that don't hit a case which contains code are not used, those would've returned null anyway. 

But I'm not sure whether the empty cases with fall-through were intentional for airflow, snr, frequency, pressure, which all use the code from the case for cooling. Looking at the git history they were added by @laf throughout 2017 as part of commits 72a0d8202a d48be5f507 592e8aca4c, so it would be nice to have some feedback on what'd be the best thing to do with those - leave alone or remove? :smiley:

includes/discovery/functions.inc.php
```diff
@@ -382,64 +382,39 @@ function discover_sensor(&$valid, $class, $device, $oid, $index, $type, $descr,
...
-        case 'current':
-            $limit = null;
-            break;
         case 'fanspeed':
             $limit = $current * 0.80;
             break;
-        case 'power':
-            $limit = null;
-            break;
-        case 'power_consumed':
```
murrant: power_consumed and power_factor are both -1 for the limit

includes/discovery/functions.inc.php
```diff
@@ -382,64 +382,39 @@ function discover_sensor(&$valid, $class, $device, $oid, $index, $type, $descr,
...
         case 'power_factor':
             $limit = -1;
             break;
         case 'signal':
             $limit = -80;
             break;
-        case 'airflow':
```
murrant: airflow, dbm, snr, frequency, pressure are all supposed to be 95% of the current value.

martijn-schmidt:
Starting with a clean batch of changes so I can undo some of the mistakes I made in the first attempt. :)

includes/discovery/functions.inc.php
martijn-schmidt: Updated description of the PR to clarify that this change is intended.

includes/discovery/functions.inc.php
martijn-schmidt: Updated PR to leave airflow, snr, frequency and pressure untouched. However, a 5% variance is not enough for guessed dbm limits so I kept that removed in the PR code. Detailed explanation in the updated PR description.

martijn-schmidt:
Should be a lot better now - after starting with a clean branch I applied the yaml group discovery fallback to `null` instead of `""` patch and collected test data so we can see the effects of that in a separate commit, and then I applied the "changes to guessed limits" patch plus collected test data again to make sure the Travis build passes.

Also updated the PR description with the key differences of before vs after, as pointed out on Discord it was not clear enough.

Finally, would love some feedback on how to approach resetting the incorrectly guessed limits on existing LibreNMS deployments without messing up manually configured limits (even if we one-off clear limits that are pulled from SNMP as bycatch, those ought to be picked back up immediately with the next discovery run).

murrant:
If you set the fields to null in the DB, I think the auto-guessing will run again.

martijn-schmidt:
> If you set the fields to null in the DB, I think the auto-guessing will run again.

Yes, that's also how I interpreted the code - however what I mean is, how do we clean up the limits for sensor classes that were incorrectly guessed? This seems to come up as a support question every now and then, it would be ideal if we could preempt the questions by taking care of the "past mistakes" by one-time updating the database upon loading this code for the first time.

murrant:
You can't, there is no way to differentiate from guessed limits and user set/adjusted limits.

murrant:
Oopsie, the merge on the docs is reverting some recent changes.

PR #10333, opened 2019-06-11
Check correct path for config in the installer

janyksteenbeek:
For older installations, install.php is still available for installations that have the configuration file in the parent directory. This PR fixes this.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/install.php
+++ b/html/install.php
@@ -80,7 +80,7 @@ if ($stage == 4) {
     }
 } elseif ($stage == 6) {
     // If we get here then let's do some final checks.
-    if (!file_exists("config.php") && !file_exists("{$librenms_dir}/config.php")) {
+    if (!file_exists("{$librenms_dir}/config.php")) {
         // config.php file doesn't exist. go back to that stage
         $msg = "config.php still doesn't exist";
         $stage = 5;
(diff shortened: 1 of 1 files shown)


janyksteenbeek:
Seems like Travis CI tests are failing because of code that is not relevant to this PR.

murrant:
Yeah, upstream issue with Chrome 75.  Another PR will fix.

html/install.php
```diff
@@ -4,7 +4,7 @@
 session_start();
 if (empty($_POST) && !empty($_SESSION) && !isset($_REQUEST['stage'])) {
     $_POST = $_SESSION;
-} elseif (!file_exists("config.php")) {
+} elseif (!file_exists("config.php") && !file_exists('../config.php')) {
```
murrant: You should not use the relative directory.
try:
```php
$librenms_dir = realpath(__DIR__ . '/..');
...
} elseif (!file_exists("$librenms_dir/config.php")) {
```

janyksteenbeek: Changed accordingly

janyksteenbeek:
Replace relative directory with absolute path  according to your review

PR #10334, opened 2019-06-11
Set Service Ignore and Disabled in UI

h-barnhart:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Added checkboxes to set ignore and disabled on services in the UI using the Add Service modal. Modified check-services.php. Old version would set service_disabled depending on device status. New version takes the value in the database into consideration before polling the service.


includes/html/modal/new_service.inc.php
```diff
@@ -13,6 +13,8 @@
 
 use LibreNMS\Authentication\LegacyAuth;
 
+include_once '/opt/librenms/includes/services.inc.php';
```
Cormoran96: Can you use Librenms install path variable. Because all install is not in /opt directory

h-barnhart: Bah, I forgot that was in there. Yep, I'll get that fixed.

PipoCanaja:
Thanx @h-barnhart

PR #10352, opened 2019-06-20
Add LibreNMS watermark to graphs

sgroom-lb:
added "options support" to my rrdtool. 
I'm sure there's a better way to do this (I've only been using Libre for about a week) but it's simple and straightforward.
Possibly add an options field to Global Settings / External Settings /  RRDTool Setup

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

I recently manually added "options support" to my rrdtool. It might be something useful for folks...
I'm sure there's a better way to do this (I've only been using Libre for about a week) but it's simple and straightforward. Possibly an addition to the WebUI rrdtool settings page

### DIFF
--- a/includes/rrdtool.inc.php
+++ b/includes/rrdtool.inc.php
@@ -103,6 +103,10 @@ function rrdtool_graph($graph_file, $options)
     /** @var Proc $rrd_sync_process */
 
     if (rrdtool_initialize(false)) {
+        $now_timestamp = date('Y-m-d H:i:s', time());  // timestamp
+        $watermark_font = ' --font WATERMARK:8:Times'; // Watermark size and font
+        $watermark_text = ' --watermark "Created by LB Water Information Technologies @ '.$now_timestamp.'"'; // watermark text
+        $options .= $watermark_font.$watermark_text; // concatenate with the existing options passed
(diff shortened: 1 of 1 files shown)


sgroom-lb:
graph_example

sgroom-lb:
Capture

includes/rrdtool.inc.php
```diff
@@ -103,6 +103,10 @@ function rrdtool_graph($graph_file, $options)
     /** @var Proc $rrd_sync_process */
 
     if (rrdtool_initialize(false)) {
+        $now_timestamp = date('Y-m-d H:i:s', time());  // timestamp
```
murrant: You don't need the second argument here.
Also, this probably isn't the best place for this, since it adds it to EVERY command.  It should probably be in the rrd_graph command or just graph.inc.php

sgroom-lb: removed redundant argument. next step will be webui option

laf:
Config option to set the text and enable it (with imho it disabled by default).

laf:
@sgroom-lb Any update to this?

murrant:
Feel free to re-open. Closing due to no response.

PR #10358, opened 2019-06-20
Added Support For Teltonika RUT2XX Devices

jozefrebjak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/discovery/rutos-2xx.yaml
```diff
@@ -0,0 +1,102 @@
...
+modules:
+  sensors:
+    pre-cache:
+      data:
+        - oid:
+            - Operator
+    temperature:
+      data:
+        - oid: Temperature
+          value: Temperature
```
murrant: You don't need value if oid is the same.

jozefrebjak: I deleted them.

includes/definitions/rutos-2xx.yaml
```diff
@@ -0,0 +1,21 @@
+os: rutos-2xx
+type: network
+group: unix
+text: "Teltonika RutOS"
+icon: teltonika
+ifXmcbc: true
+ifname: true
+mib_dir:
+  - rutos-2xx
```
murrant: The mib directory should be `teltonika` can you change that and move the mib?

includes/polling/os/rutos-2xx.inc.php
```diff
@@ -0,0 +1,32 @@
+<?php
+$hardware = trim(snmp_get($device, '1.3.6.1.4.1.48690.1.7.0', '-OQv', '', ''), '"');
```
murrant: trim() functions are not needed.

includes/definitions/rutos-2xx.yaml
jozefrebjak: done, but there is also MIB for Rut9xx, which I am not using for now, but there is a support for that and it's in a rutos folder, and there is also used TELTONIKA-MIB, but that MIB is different for RUT9XX devices, but same name, so what can we do with that ? create another folder under Teltonika for RUT2XX and another folder for RUT9XX and then move MIBs with a same name here ?

includes/polling/os/rutos-2xx.inc.php
jozefrebjak: I changed it completely to OIDs from MIB, they are didn't worked for me, but I found where was a problem.

includes/definitions/rutos-2xx.yaml
murrant: yeah, I think subfolders is the best option there.  Silly folks using the same mib name.
The cambium mibs also deal with this.

murrant:
@jozefrebjak why is this different from rutos?

jozefrebjak:
rutos is designed for RUT9XX devices, this is for 2XX which are using different MIB, but with same name. Maybe we can add 2xx to rutos and update rutos, where are missing some sensors, but how to use two different MIBs with same name in yaml and php ?

PipoCanaja:
"rutos" should probably be renamed "rutos-9xx" and this PR should create "rutos-2xx". Subfolder for both sets of mibs (teltonika/rutos-9xx and teltonika/rutos-2xx).

PipoCanaja:
@jozefrebjak You still have some issues found by travis on tests. Let us know if you need help to solve them.

jozefrebjak:
@PipoCanaja i ran test twice, but it still falling, isn't it because of custom graph ?

PipoCanaja:
> @PipoCanaja i ran tests twice, but it still falling, isn't it because of custom graph ?

According to Travis : 
``` OS rutos-2xx: Discovered sensors data does not match that found in tests/data/rutos-2xx.json ``` 

You need to re-run ./scripts/save-test-data.php for the rutos test files.

murrant:
@jozefrebjak it is because you already have the states in your db, but the code that checks if they need to be updated is case insensitive.

jozefrebjak:
@murrant thank you for help with this.

PR #10363, opened 2019-06-22
Added VRP SFPs thresholds and map entPhysical to ifIndexes

PipoCanaja:
A little bug appeared in the discover entity function, as the ifIndex was not taken into account during updates, only during initial insert.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/entity-physical/entity-physical.inc.php
+++ b/includes/discovery/entity-physical/entity-physical.inc.php
@@ -117,6 +117,7 @@ if ($device['os'] == 'saf-cfm') {
 }
 
 foreach ($entity_array as $entPhysicalIndex => $entry) {
+    unset($ifIndex);
     if ($device['os'] == 'junos') {
         // Juniper's MIB doesn't have the same objects as the Entity MIB, so some values
         // are made up here.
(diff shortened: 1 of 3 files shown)


PipoCanaja:
Testing it right now.

includes/discovery/functions.inc.php
```diff
@@ -812,6 +812,9 @@ function discover_entity_physical(&$valid, $device, $entPhysicalIndex, $entPhysi
                 'entPhysicalAlias'        => $entPhysicalAlias,
                 'entPhysicalAssetID'      => $entPhysicalAssetID,
             );
+            if (!empty($ifIndex)) {
+                $update_data['ifIndex'] = $ifIndex;
```
murrant: You need to include ifIndex every time, otherwise it won't be unset if needed.

includes/discovery/entity-physical/entity-physical.inc.php
```diff
@@ -190,6 +191,19 @@
         $entPhysicalIsFRU        = $entry['entPhysicalIsFRU'];
         $entPhysicalAlias        = $entry['entPhysicalAlias'];
         $entPhysicalAssetID      = $entry['entPhysicalAssetID'];
+
+        //VRP devices seems to use LogicalEntity '1' instead of '0' like the default code checks.
+        //Standard code is still run after anyway.
+        if (array_key_exists('1', $entry) && array_key_exists('entAliasMappingIdentifier', $entry['1'])) {
```
murrant: You could use `isset($entry['1']['entAliasMappingIdentifier'])` here, it works on nested arrays.
Also, not sure about the duplication of the below code, it feels way to complex.

Maybe:
```
if (!empty($entry['1']['entAliasMappingIdentifier']) && str_contains('ifIndex',  $entry['1']['entAliasMappingIdentifier'])) {
    $ifIndex = str_replace('ifIndex.', '', $entry['1']['entAliasMappingIdentifier']);
}
```

not sure if it does the same thing, but yeah, below code is kinda silly.  I like the part where it uses strpos fIndex so it doesn't match position 0

includes/discovery/functions.inc.php
PipoCanaja: Makes sense, of course.

includes/discovery/entity-physical/entity-physical.inc.php
PipoCanaja: Ok. I'll clean it a little bit. 

Basically, a refactor would be needed. The entAliasMappingIdentifier is dual indexed, which gives this '1' or '0' or anything else really, according to the mib. 
What I would be interested to get is the lowest index of entAliasMappingIdentifier. 

Considering my dev plateform is down now (so I have to work on prod) I won't go for the refactor today.

PipoCanaja: done with a regexp, and only for VRP for now.

includes/discovery/functions.inc.php
PipoCanaja: done

PipoCanaja:
Tests done. Still not found a way to map sensors to ifIndex via YAML ... Will probably do it the PHP way at some point, unless I find a way to add this capability to YAML. 
Until then, this patch is now ready for review.

PR #10364, opened 2019-06-22
Schleifenbauer: support for external temperature/humidity sensors

martijn-schmidt:
Just a small device update, we received an external temperature/humidity sensor from the vendor to perform some demo testing: https://www.schleifenbauer.eu/en/pdu/discover-yourpdu/#sensors

I built some code to pull data from these things with LibreNMS. Unfortunately it had to be done through PHP-based discovery because the dual-index OIDs, as well as the fact that data returned from `sdbDevSnsValue` could signify either a dry-switch contact, temperature, or humidity sensor based on the corresponding value of `sdbDevSnsType`. So it was a bit more complicated than what could be achieved through YAML.

Test data was updated, besides the new sensors you'll find that the firmware was changed from v250 to v252 and that I added a second unit to create a ring topology for the tests while it was previously a standalone unit.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 10364`


murrant:
```
        pre-cache:
            data:
                - oid:
                      - sdbDevIdSerialNumber
        temperature:
            data:
                -
                    oid: sdbDevSnsTable
                    value: sdbDevSnsValue
                    num_oid: '.1.3.6.1.4.1.31034.12.1.1.2.8.2.1.3.{{ $index }}'
                    descr: '{{ $sdbDevIdSerialNumber }} External Temperature'
                    divisor: 100
                    skip_values:
                        - op: '!='
                          oid: sdbDevSnsType
                          value: T
        humidity:
            data:
                -
                    oid: sdbDevSnsTable
                    value: sdbDevSnsValue
                    num_oid: '.1.3.6.1.4.1.31034.12.1.1.2.8.2.1.3.{{ $index }}'
                    descr: '{{ $sdbDevIdSerialNumber }} External Humidity'
                    divisor: 100
                    skip_values:
                        - op: '!='
                          oid: sdbDevSnsType
                          value: H
```
This works with a slight adjustment in dynamic_discovery_get_value()
```
Index: includes/discovery/functions.inc.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- includes/discovery/functions.inc.php	(revision f3ba8947f74eba94368b4af0c19f63b0e42767d2)
+++ includes/discovery/functions.inc.php	(date 1561323847000)
@@ -1123,6 +1123,11 @@
                 return $pre_cache[$index][$name];
             } elseif (count($pre_cache[$name]) === 1) {
                 return current($pre_cache[$name]);
+            } elseif (str_contains($index, '.')) {
+                list($left, $right) = explode('.', $index);
+                if (isset($pre_cache[$name][$left])) {
+                    return $pre_cache[$name][$left];
+                }
             }
         } else {
             return $pre_cache[$name];
```

PipoCanaja:
@murrant : I like the idea. It could save me in an issue I have on huawei switches to map sensor entity to ifIndexes. Will try to test it as soon as my new dev server is up. Old one got lost in the virtualisation maze of a lab server :| ... 
Will you open a PR on it ? Should I ?

martijn-schmidt:
I've just come back from a week of travelling for work, so I'll pick this PR back up and see if I can get Tony's idea working :) ought to be another good learning experience!

laf:
@martijn-schmidt Any update on this one?

includes/definitions/discovery/omnipcx.yaml
```diff
@@ -9,7 +9,7 @@ modules:
                     num_oid: '.1.3.6.1.4.1.637.64.4400.1.2'
                     descr: 'Pbx State'
                     state_name: pbxState
-                    index: 'pbxState'
+                    index: 'pbxState{{ $index }}'
```
PipoCanaja: Hi @martijn-schmidt 
Why did you change other YAML files than the one impacted by your change ?  PBX state for instance is a "standalone" oid, the $index value has no particular meaning here.

martijn-schmidt:
Hey @PipoCanaja - responding to your comment here as it pertains to a bit more than the specific yaml file you referenced.

The trouble with Schleifenbauer is that I'd chosen to go with PHP because I needed to do several things which aren't possible with the YAML implementation in the current librenms master code branch. Those are the following three items:

1. dual-index OID -> now possible in YAML with @murrant's idea.

2. serial-based indexes -> necessary because Schleifenbauer PDUs in a ring change to a new unit address during some operational activities, meaning you could get graph readouts from a different PDU at the previous SNMP value and upon rediscovery you may also lose your historical graph data for that sensor. With the index linked to the serial rather than the unit address, you can run a rediscovery and the graph will be picked up where it left off for that PDU.

The change in 89d3f25 makes this possible to achieve in YAML, but it also means that SNMP-obtained values are used as the index if someone happened to add a valid OID from the MIB as the yaml-index (this is the case for omnipcx and a few others OS'es). I'm currently trying to figure out how I can best tackle this new behaviour, so if you have some ideas on how to approach this they'd be welcomed.

3. Math-based entPhysicalIndex -> needed for sensor support in the inventory page, but I still have to begin work on figuring this out for YAML - right now it only allows you to combine a string with one variable, as opposed to permitting math operations and multiple variables.

PipoCanaja:
Issue here is that if you change any of the existing indexes in YAML files, all the historical data will be lost for all LibreNMS users of the sensors associated with those indexes ... So, basically, this should be avoided :)

martijn-schmidt:
> Issue here is that if you change any of the existing indexes in YAML files, all the historical data will be lost for all LibreNMS users of the sensors associated with those indexes ... So, basically, this should be avoided :)

Yep, I see.. in the case of omnipcx the resulting index for the unit test is the same after making this YAML change - as you pointed out it's a "standalone" OID and here that means the $index variable is empty -  but the unit test indexes do change for airos-af and eds because $index does contain some data for those OIDs, and that's a problem.

To be fair, it could be worked around by only calling YamlDiscovery::replaceValues if `$index != $data['oid']` but that does mean that if you ever want to have an OID-based index in the future, it always has to be written out as `'{{ $oidname }}'` which is non-default behaviour for the YAML discovery process.

martijn-schmidt:
So I found that `getValueFromData()` in `LibreNMS/Device/YamlDiscovery.php` is a 1:1 clone of `dynamic_discovery_get_value()` in `includes/discovery/functions.inc.php`.

The former is only used in `LibreNMS/Device/YamlDiscovery.php` itself, and the latter is used in `LibreNMS/Device/YamlDiscovery.php` and `includes/discovery/functions.inc.php` and `tests/FunctionsTest.php`.

Should we remove one of those functions to de-duplicate the code? If so, which one?

PipoCanaja:
Hi

I```LibreNMS/Device/YamlDiscovery.php``` should be the one. ``` includes/discovery/functions.inc.php ``` should be pointing to the function in ```LibreNMS/Device/YamlDiscovery.php``` (for safety) but any call to the legacy code structure should be changed as much as possible. 

Bye

martijn-schmidt:
So.. one last "special" exception due to inconsistency in the Schleifenbauer MIB. Most `current` sensor related values in the MIB are returned as `CentiAmpere` meaning a divisor of 100 is needed - I'm using `sdbDevInActualCurrent` and `sdbDevInMaxAmps` in the YAML discovery, for example.

But the `sdbDevCfMaximumLoad` which is the maximum rated load of the device per phase returns its value as plain `Ampere`, either 16 or 32. So when the divisor is applied to that you'll get the following output in debug mode: `(limits: LL: , LW: , W: 28, H: 0.32)`

To work around that I've added some code which also enables simple arithmetic for the limit variables, so we can "manually" raise the value by 100 using an expression before the divisor is applied to `sdbDevCfMaximumLoad`. We can also fall back to the old PHP code for `current` sensors detection if that particular change isn't liked, of course..

PipoCanaja:
Hi @martijn-schmidt 
Let's ask @murrant about ExpressionLanguage evaluation. I think that mixing ExpressionLanguage with unprotected string replacement will not be safe.

PipoCanaja:
(That being said, I would love to have this kind of flexibility in YAML file).

martijn-schmidt:
I think ExpressionLanguage is meant to be safe, or at least safer than the very undesirable alternative `eval()` would be. And the `evaluate()` call evaluates the expression without compiling it to PHP: https://symfony.com/blog/new-in-symfony-2-4-the-expressionlanguage-component

If we really want to be paranoid we could also write a nice regex to make sure we only ever hit the ExpressionLanguage call if the variable being evaluated looks like a simple arithmetic expression e.g. `starts with a number like -23.45 and is followed by one or more repetitions of (one basic arithmetic operator like */+- combined with a number like -43.18)`, but that is a bit beyond my skills at the moment.

By the way, I've outlined the reasons for touching the yaml and tests for `fusion` in detail in the commit message for 6552960, but the gist of it is that it wasn't pre-caching the OIDs used for the limits correctly. This caused some problems with the new YamlDiscovery::replaceValues() call that I was using. And then as a plot twist I found that the `fusion` vendor also needs different divisors on limits than on the value we are polling, so I was able to fix that here with the same workaround that I'd built for `schleifenbauer`.

Still need to investigate what the `netgear` unit test is complaining about, but that's something for another evening!

martijn-schmidt:
Spent some time playing with regex after all, `^-?\d+(\.\d+)?( ?[-+*\/] ?-?\d+(\.\d+)?)+$` only matches formulas which begin and end with rational numbers, and where each rational number is separated by one simple arithmetic operator. This means the ExpressionLanguage->evaluate() call will not be made if the variables were not successfully replaced, or if someone tries to do something fishy.

Anyway, with this being done and the build tests succeeding I think everything is ready for @murrant's review and hopefully a merge!

martijn-schmidt:
Made one more change so the fallback system for replaceValues() only strips the last subindex digit, the behaviour for Schleifenbauer is the same because it's only a 2-depth index but this helps me with an upcoming Stulz WIB8000 pull request which has a 3-depth index system.

PipoCanaja:
@laf @murrant 
I'll need help on that one. 
- Would we go with the Expression parsing or not ? 
  - If yes, should we put even more "security checks" on it ? 

Thanx,

martijn-schmidt:
So we had a bit more talk on this subject in Discord earlier this week, but let me ask this in an unambiguous manner: if I drop-in replace the `$expressionLanguage->evaluate()` calls with a homegrown `BasicArithmetic::evaluate_arithmetic()` function, will that assuage the concerns and allow for this PR to be merged? Would the appropriate place for `public static function evaluate_arithmetic()` be in a new `class BasicArithmetic` at `LibreNMS/Util/BasicArithmetic.php`?

Or do I need to approach it differently?

murrant:
I don't think a home grown solution would be more secure/robust/performant.

Other than security another hesitation I have is the possibility that this may cause complications down the road.  Probably need others to weigh in.

As I said on Discord.  If you want to get the device changes merged, just implement them in PHP. Then you don't have to wait for this discussion to be resolved.

Jellyfrog:
@martijn-schmidt Closing this due to inactivity, please feel free to reopen if you continue working on it!

martijn-schmidt:
@Jellyfrog I was waiting for the maintainer-level discussion to be resolved to make additional changes.. didn't see much point in rewriting everything to PHP for a short term merge if we ideally have everything in YAML for the long term. So if there is a definitive decision (yes or no are both good for me) then I will pick it back up. 

PS: as it stands now this would run into a merge conflict, certain pieces of its code were already merged in another PR. So even if the decision to go YAML is made, I'll still need to spend some time to rebase the code. But that's be fine by me.

murrant:
I'm against trying to move everything to yaml at this time, it is not worth the added complexity.

PR #10365, opened 2019-06-22
Don't indicate that host is always up on uptime counter when host is ignored

CirnoT:
Have statusName return proper status (up or down) even when device is ignored. Previously discussed at https://github.com/librenms/librenms/pull/10285

For what it's worth - these colors should be standardized somehow. On devices page gray is used for disabled devices and ignored near device name but gray near uptime never shows up (this is a bug actually, another PR will handle it). On the other hand, on Availability Map gray is for ignored and black for disabled.

Attached screenshots shows how weird these colors are
chrome_2019-06-23_00-16-10
chrome_2019-06-23_00-18-48

Check https://github.com/librenms/librenms/pull/10366 for solution I propose to fix this.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/Models/Device.php
+++ b/app/Models/Device.php
@@ -348,8 +348,6 @@ class Device extends BaseModel
     {
         if ($this->disabled == 1) {
             return 'disabled';
-        } elseif ($this->ignore == 1) {
-            return 'ignore';
         } elseif ($this->status == 0) {
             return 'down';
         } else {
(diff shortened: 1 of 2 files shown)


CirnoT:
Okay. Based on what you said @murrant I've also changed logic on device page to indicate that device is down even if it's ignored.

However if Ignore applies to alerts only, what should we do with 'Disable all alerting for this host' option under 'Alert Settings'? Deprecate and migrate it to Ignore?

PR #10366, opened 2019-06-22
Better handling disabled devices on devices list page

CirnoT:
Differentiates between ignored device (gray) and disabled device (black) and fixes disabled hosts showing always as up on uptime counter by removing uptime counter for disabled devices.

chrome_2019-06-23_00-31-33

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Table/DeviceController.php
+++ b/app/Http/Controllers/Table/DeviceController.php
@@ -140,7 +140,7 @@ class DeviceController extends TableController
     private function getLabel($device)
     {
         if ($device->disabled) {
-            return 'label-default';
+            return 'blackbg';
         }
 
         if ($device->ignore) {
--- a/includes/html/pages/devices.inc.php
+++ b/includes/html/pages/devices.inc.php
@@ -314,12 +314,14 @@ if ($format == "graph") {
                     return "<span>" + row.hostname + "</span>";
                 },
                 "uptime": function (column, row) {
-                    if (isNaN(row.uptime.charAt(0))) {
-                        return row.uptime;
-                    } else if (row.status == 'down') {
+                    if (row.status == 'down') {
                         return "<span class='alert-status-small label-danger'></span><span>" + row.uptime + "</span>";
-                    } else {
+                    } else if(row.status == 'up') {
                         return "<span class='alert-status-small label-success'></span><span>" + row.uptime + "</span>";
+                    } else if(row.status == 'disabled') {
+                        return '';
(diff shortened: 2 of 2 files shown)


murrant:
@CirnoT why do you keep giving so much weight to ignored devices?  Ignored simply means mute alerts.

I still want to see if an ignored device is up or down, I don't want to see a gray bar.

What we really need is a clear definition of disabled and ignored.  These are not defined anywhere in the docs AFAIK.  After that, we can decide how they should be displayed.

kkrumm1:
The doc for this is in the FAQ https://docs.librenms.org/Support/FAQ/#faq28

CirnoT:
I don't think you understand me @murrant. This PR is for the state indicator that indicates alert status (which ignored disables), the one on the very left near device name. It also fixes a bug that causes disabled devices to always show as online near uptime counter (the green bar near the time device is up or down).

murrant:
Solution, remove the right color and put it back to the way it was about a release ago?

CirnoT:
I don't think I follow you. You want to remove uptime counter or alert state (on the left)? I don't remember any release without them so I guess it must've been added before I got into using LibreNMS. This still doesn't fix the issue of 'ignored' and 'disabled' state being interpreted as 'up'?

CirnoT:
Okay, let's try simpler example. Here's how it works right now on device that is offline.
Ignored is checked:
chrome_2019-06-23_23-50-44
Ignored is unchecked:
chrome_2019-06-23_23-51-04

And this is how it works after both PRs are merged, same device, same state (offline).
Ignored is checked:
chrome_2019-06-23_23-51-54
Ignored is unchecked:
chrome_2019-06-23_23-51-41

Do you see the bug? When device is IGNORED and OFFLINE LibreNMS shows device as ONLINE, which is FALSE.

CirnoT:
Okay, let's try simpler example. Here's how it works right now on device that is offline.
Ignored is checked:
chrome_2019-06-23_23-50-44
Ignored is unchecked:
chrome_2019-06-23_23-51-04

And this is how it works after both PRs are merged, same device, same state (offline).
Ignored is checked:
chrome_2019-06-23_23-56-05
Ignored is unchecked:
chrome_2019-06-23_23-56-58

Do you see the bug? When device is IGNORED and OFFLINE LibreNMS shows device as ONLINE (color indicator near uptime counter [the 21 days text]), which is FALSE.

PR #10372, opened 2019-06-24
Replace color indicator near uptime counter with colored text and change color of status indicator to black on disabled devices instead of gray (matches availablity map with show ignored/disabled enabled)

CirnoT:
Replace color indicator near uptime counter with colored text and change color of status indicator to black on disabled devices instead of gray (matches availablity map with show ignored/disabled enabled).

* Brings back header color for ignored devices on device's own page (https://u.gensokyo.re/d/nCpaBasc (down) / https://u.gensokyo.re/d/Ya16kBvJ (up) / https://u.gensokyo.re/d/WVakwOxH (disabled))
* Change color of left indicator status for disabled devices from gray to black to match availability map (https://u.gensokyo.re/d/scsXvQuE)
* Remove color indicator near uptime counter on the right, instead use red text to mark downtime (https://u.gensokyo.re/d/5dGnUsx4)
* Do not display uptime counter for disabled devices (https://u.gensokyo.re/d/scsXvQuE)
* Update logic to change left counter to orange if device is within reboot warning period as previously only uptime counter indicator changed color and it is now removed (https://u.gensokyo.re/d/8xgyJhIq)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Table/DeviceController.php
+++ b/app/Http/Controllers/Table/DeviceController.php
@@ -120,7 +120,7 @@ class DeviceController extends TableController
     {
         return [
             'extra' => $this->getLabel($device),
-            'status' => $device->statusName(),
+            'status' => $this->getStatus($device),
             'icon' => '<img src="' . asset($device->icon) . '" title="' . pathinfo($device->icon, PATHINFO_FILENAME) . '">',
             'hostname' => $this->getHostname($device),
             'metrics' => $this->getMetrics($device),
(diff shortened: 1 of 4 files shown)


murrant:
Looking at this I realized there is a lot of the display processing in the backend.
Probably should remove the "extra" column, add the disabled column and pass through the status column.

murrant:
Looking into the above suggestion and it kind of snowballs.  Probably just wait to rewrite the frontend page to blade for that.

PR #10384, opened 2019-06-27
Added VPN sensors for Fortinet Fortigate

jozefrebjak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

image


murrant:
This seems reasonable but it almost feels like there should be a unified VPN section instead of generic sensors.

jozefrebjak:
@murrant  i added more sensors, will be nice to have section, but when is possible to get that ? i don't know how to add to code, but i need a state of VPN's as soon as possible, we use vpn to monitor remote sites and monitor virtual ports of vpn is waste of time, because port is up even though VPN is down. So i will be happy if we can use this for now.

PR #10388, opened 2019-06-27
Add more detail to webgui alerts

djamp42:
I add everything via IP address's, this works fine, but the alert tables only display the hostname(IP). This PR is to add a little more detail to the alert, by adding the sysname under the hostname. A couple of other places do this such as "device dependence's"

Made this a config option in case someone is already doing $config['force_ip_to_sysname'] = false;

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/html/table/alerts.inc.php
+++ b/includes/html/table/alerts.inc.php
@@ -138,7 +138,7 @@ foreach (dbFetchRows($sql, $param) as $alert) {
     } elseif ((int)$alert['state'] === 1 || (int)$alert['state'] === 3 || (int)$alert['state'] === 4) {
         if ((int)$alert['state'] === 3) {
             $msg = '<i class="fa fa-angle-double-down" style="font-size:20px;" aria-hidden="true" title="Status got worse"></i>';
-         } elseif ((int)$alert['state'] === 4) {
+        } elseif ((int)$alert['state'] === 4) {
             $msg = '<i class="fa fa-angle-double-up" style="font-size:20px;" aria-hidden="true" title="Status got better"></i>';
         }
     } elseif ((int)$alert['state'] === 2) {
--- a/includes/html/table/alerts.inc.php
+++ b/includes/html/table/alerts.inc.php
@@ -203,6 +203,7 @@ foreach (dbFetchRows($sql, $param) as $alert) {
         'rule' => '<i title="' . htmlentities($alert['rule']) . '"><a href="' . generate_url(array('page' => 'alert-rules')) . '">' . htmlentities($alert['name']) . '</a></i>',
         'details' => '<a class="fa fa-plus incident-toggle" style="display:none" data-toggle="collapse" data-target="#incident' . ($rulei) . '" data-parent="#alerts"></a>',
         'hostname' => $hostname,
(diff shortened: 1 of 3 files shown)


jozefrebjak:
This will be useful 👌

PipoCanaja:
Hi @djamp42 
Thanx for your PR. Could you document it so make it easier to find for everybody ? Would be good to have it in default.inc.php as well as in the doc.

djamp42:
@PipoCanaja After thinking about this, it really should just follow whatever $config['force_ip_to_sysname'] is set to. That setting flips the hostname and sysname in a lot of places, and now it flips it in the alert table too!

djamp42:
I also added the option for a location column to the alert table/widget, and made that clickable to the location in question.

includes/html/table/alerts.inc.php
```diff
@@ -137,7 +138,7 @@
     } elseif ((int)$alert['state'] === 1 || (int)$alert['state'] === 3 || (int)$alert['state'] === 4) {
         if ((int)$alert['state'] === 3) {
             $msg = '<i class="fa fa-angle-double-down" style="font-size:20px;" aria-hidden="true" title="Status got worse"></i>';
-        } elseif ((int)$alert['state'] === 4) {
+         } elseif ((int)$alert['state'] === 4) {
```
PipoCanaja: You have an indentation issue here.

includes/html/table/alerts.inc.php
```diff
@@ -155,7 +156,11 @@
         $severity .= ' <strong>-</strong>';
     }
 
-    $hostname = '<div class="incident">' . generate_device_link($alert, shorthost($alert['hostname'])) . '<div id="incident' . ($rulei + 1) . '" class="collapse">' . $fault_detail . '</div></div>';
+    if (Config::get('force_ip_to_sysname')) {
```
laf: You can use `format_hostname()` to do this without an `if/else` block.

djamp42: I'll have to look into this more, at first glance it's no obvious to me how it would work. I'm trying to display both the hostname AND sysname at the same time. How it is displayed on the webgui is based on what **force_ip_to_sysname** is set to, it will flip-flop the placement of hostname and sysname. The exact same behavior can be seen in the devices list page. I'll look at that code and see if i can figure it out. thanks

laf: You're correct, you might need to create another helper function unless one exists already.

djamp42: @laf After looking into this more, it seems your converting most of the pages to laravel blades? However the alert notification page hasn't been done yet? I didn't want to put a ton of work into this if your just going to convert it to a blade later. If i can get a function working it seems like it would be useful to use it everywhere hostname is displayed to get it somewhat standardized across all pages. I noticed some pages display hostname + sysname, while others just display the hostname. 

I will say this PR has helped a ton in determining what is down at a glance if your just using IPs for the hostname.

murrant: @djamp42 made the change for you ;)

PR #10392, opened 2019-06-29
Allow OS to define config highlighting (pfSense)

brownowski:
Fix display for pfsense config files.
 - use xml language formatting for pfsense os
 - don't use htmlspecialchars_decode as xml tags won't show

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/html/pages/device/showconfig.inc.php
+++ b/includes/html/pages/device/showconfig.inc.php
@@ -318,6 +318,8 @@ if (LegacyAuth::user()->hasGlobalAdmin()) {
     if (!empty($text)) {
         if (isset($previous_config)) {
             $language = 'diff';
+        } else if ($device['os'] == 'pfsense') {
+            $language = 'xml';
         } else {
             $language = 'ios';
         }
@@ -327,7 +329,7 @@ if (LegacyAuth::user()->hasGlobalAdmin()) {
         // $geshi->set_line_style('color: #999999');
         echo '<div class="config">';
         echo '<input id="linenumbers" class="btn btn-primary" type="submit" value="Hide line numbers"/>';
-        echo htmlspecialchars_decode($geshi->parse_code());
+        echo $geshi->parse_code();
         echo '</div>';
     }
 }//end if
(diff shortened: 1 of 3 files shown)


murrant:
htmlspecialchars_decode() doesn't remove tags it converts things like &quot; to "

#8847

brownowski:
Yes, but by decoding the codes which are returned by the function, the xml tags in the config file are returned to the browser as tags instead of special codes and therefore don't display properly.

brownowski:
You can probably disregard this pull request.  I don't know what, but something seems to have changed from when I was testing originally and the original code now works ¯\_(ツ)_/¯

I've updated my patch for xml highlighting of the pfsense config file, which now seems to require me running htmlspecialchars_decode on $text before sending to GeSHi.  If I don't decode the special chars before GeSHi, I only get the start and close angle brackets highlighted due to them being special char encoded.  

However, the original code now seems to show the config file with the tags correctly, albeit with no syntax highlighting.

brownowski:
There might still be something here.  If I do a diff now I'm missing the xml tags like I was before.

Pre-patch:  htmlspecialchars_decode on $geshi->parse_code()
pre-patch

Post-patch: htmlspecialchars_decode on $text pre-GeSHi 
post-patch

includes/html/pages/device/showconfig.inc.php
```diff
@@ -318,16 +318,18 @@
     if (!empty($text)) {
         if (isset($previous_config)) {
             $language = 'diff';
+        } elseif ($device['os'] == 'pfsense') {
```
laf: Should we move this to the device definition? So have `config_language: xml` in the pfsense definition then fall back to ios for all others?

murrant: config_highlighting would be my naming

laf: Even better :)

laf: @brownowski Can you implement this please?

includes/html/pages/device/showconfig.inc.php
```diff
@@ -318,16 +318,18 @@
     if (!empty($text)) {
         if (isset($previous_config)) {
             $language = 'diff';
+        } elseif (isset(Config::get('os')[$device['os']]['config_highlighting'])) {
```
murrant: Do this like this:
```php
$language = isset($previous_config) ? 'diff' : Config::getOsSetting($device['os'], 'config_highlighting', 'ios');
```

murrant:

PR #10399, opened 2019-07-01
Allow filtering of getUserlist LDAP function

ifred16:
I have an openldap directory with several thousand of accounts.
so when I try to use the ldap authentification in Librenms, it does not work.

I found this solution (add an ldap filter to limit the number of results) :

I have added $config[‘auth_ldap_Userlist_filter’] = ‘service=informatique’; in the config.php

I have modified the source code of the getUserlist() function in the file librenms/LibreNMS/Authentication/LdapAuthorizer.php like this :

$filter = ‘(’ . Config::get(‘auth_ldap_prefix’) . ‘*)’;
if (Config::get(‘auth_ldap_Userlist_filter’) != NULL) {
    $filter = ‘(’ . Config::get(‘auth_ldap_Userlist_filter’) . ‘)’;
}

Now with this modification, the ldap authentification working fine.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Authentication/LdapAuthorizer.php
+++ b/LibreNMS/Authentication/LdapAuthorizer.php
@@ -152,8 +152,8 @@ class LdapAuthorizer extends AuthorizerBase
             }
 
             $filter = '(' . Config::get('auth_ldap_prefix') . '*)';
-	    if (Config::get('auth_ldap_Userlist_filter') != NULL) {
-                    $filter = '(' . Config::get('auth_ldap_Userlist_filter') . ')';
+            if (Config::get('auth_ldap_Userlist_filter') != null) {
+                $filter = '(' . Config::get('auth_ldap_Userlist_filter') . ')';
             }
 
             // build group filter
--- a/LibreNMS/Authentication/LdapAuthorizer.php
+++ b/LibreNMS/Authentication/LdapAuthorizer.php
@@ -152,8 +152,8 @@ class LdapAuthorizer extends AuthorizerBase
             }
 
             $filter = '(' . Config::get('auth_ldap_prefix') . '*)';
-            if (Config::get('auth_ldap_Userlist_filter') != null) {
-                $filter = '(' . Config::get('auth_ldap_Userlist_filter') . ')';
+            if (Config::get('auth_ldap_userlist_filter') != null) {
+                $filter = '(' . Config::get('auth_ldap_userlist_filter') . ')';
             }
 
             // build group filter
--- a/LibreNMS/Authentication/LdapAuthorizer.php
+++ b/LibreNMS/Authentication/LdapAuthorizer.php
@@ -152,6 +152,9 @@ class LdapAuthorizer extends AuthorizerBase
             }
 
             $filter = '(' . Config::get('auth_ldap_prefix') . '*)';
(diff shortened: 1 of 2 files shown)


kkrumm1:
@ifred16 you need to sign the CLA

ifred16:
@kkrumm1 It's ok, I have signed the CLA

LibreNMS/Authentication/LdapAuthorizer.php
```diff
@@ -152,6 +152,9 @@ public function getUserlist()
             }
 
             $filter = '(' . Config::get('auth_ldap_prefix') . '*)';
+            if (Config::get('auth_ldap_Userlist_filter') != null) {
+                $filter = '(' . Config::get('auth_ldap_Userlist_filter') . ')';
```
laf: This overrides `$filter` above, is that what you wanted to do?

ifred16: Yes this overrides $filter above only if $config['auth_ldap_Userlist_filter'] = 'your ldap filter here'; exist in the config.php

That is what I wanted to do

ifred16: Hello @laf can you approve changes in my pull request ?

PipoCanaja:
Hello @ifred16 
Thanx for your PR ! Could you document your option, in order for others to find out, in the same situation, how to solve it ?

ifred16:
Hello @PipoCanaja,
Thanks for your reply.

In https://docs.librenms.org/Extensions/Authentication/#ldap-authentication, I think add : 
 
Additional options (usually not needed):
...
$config[‘auth_ldap_Userlist_filter’] = 'service=informatique'; // Replace 'service=informatique' by your ldap filter to limit the number of responses if you have an ldap directory with thousand of users     
...

Do you want I document my option in /opt/librenms/doc/Extensions/Authentication.md and I make a new commit ?

PipoCanaja:
Bonjour @ifred16 
Yes please, commit your changes directly to the documentation in that same Pull Request. 
Bye

LibreNMS/Authentication/LdapAuthorizer.php
```diff
@@ -152,6 +152,9 @@ public function getUserlist()
             }
 
             $filter = '(' . Config::get('auth_ldap_prefix') . '*)';
+            if (Config::get('auth_ldap_Userlist_filter') != null) {
```
PipoCanaja: Could you rewrite the option name with low cap instead of "U" ?

ifred16: Yes, no problem, I do It.

PipoCanaja:
Thanx @ifred16. Looks good from a reader point of view but I cannot test it as I don't have LDAP in my env. 
@laf What do you think ?

ifred16:
Hello @laf 
What do you think ?

murrant:
@ifred16 Thanks for the PR!
Looks good. As I'm looking at the docs, it looks like it may be required as part of the settings.  (Yes this problem existed before your PR).  Maybe we should separate out optional settings.  (and maybe that should be a separate PR)

PR #10410, opened 2019-07-04
Allow num_oid to use OCTET STRING indexes

PipoCanaja:
Hi,

This PR creates a new variable $index_string that contains the OID converted as OCTET_STRING. basically, it converts a string index into an OID, so it can be used for snmpgets.

Exemple for an efficient_ip appliance, which exposes DNS/DHCP statistics indexed by name : 

.1.3.6.1.4.1.2440.1.4.2.3.1.3.**3.116.99.112** = 0
.1.3.6.1.4.1.2440.1.4.2.3.1.3.**3.117.100.112** = 72

it represents : 

$ snmptranslate  '-M' '/opt/librenms/mibs:/opt/librenms/mibs/efficientip' -m all .1.3.6.1.4.1.2440.1.4.2.3.1.3.3.116.99.112 
EIP-STATS-MIB::eipDnsStatValue.**"tcp"**
$ snmptranslate  '-M' '/opt/librenms/mibs:/opt/librenms/mibs/efficientip' -m all .1.3.6.1.4.1.2440.1.4.2.3.1.3.3.117.100.112   
EIP-STATS-MIB::eipDnsStatValue.**"udp"**

In the YAML file, it is now possible to do the following for DNS stats : 
```
                    oid: eipDnsStatTable
                    value: eipDnsStatValue
                    num_oid: '.1.3.6.1.4.1.2440.1.4.2.3.1.3.{{ $index_string }}'
                    descr: '{{ $eipDnsStatName }}'
                    group: 'DNS'
```
#10413 is also necessary in order to load the COUNTER values and graph them correctly. Nobody cares of the absolut number of DNS requests (RRD GAUGE), but prefers the RRD COUNTER which derives it.
Capture d’écran 2019-07-04 à 17 03 31

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Developing/os/Health-Information.md
+++ b/doc/Developing/os/Health-Information.md
@@ -85,7 +85,8 @@ The only sensor we have defined here is airflow. The available options are as fo
   the value. If not provided willuse `oid`
 - `num_oid` (required): This is the numerical OID that contains
   `value`. This should always include `{{ $index }}`.  snmptranslate
-  -On can help figure out the number
+  -On can help figure out the number.
+  In case the index is a string, `{{ $index_string }}` can be used instead.
 - `divisor` (optional): This is the divisor to use against the returned `value`.
 - `multiplier` (optional): This is the multiplier to use against the returned `value`.
 - `low_limit` (optional): This is the critical low threshold that
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -1041,7 +1041,10 @@ function discovery_process(&$valid, $device, $sensor_type, $pre_cache)
                 d_echo("Final sensor value: $value\n");
 
                 if (YamlDiscovery::canSkipItem($value, $index, $data, $sensor_options, $pre_cache) === false && is_numeric($value)) {
-                    $oid = str_replace('{{ $index }}', $index, $data['num_oid']);
+		    $oid = str_replace('{{ $index }}', $index, $data['num_oid']);
+		    // if index is a string, we need to convert it to OID
+		    // strlen($index) as first number, and each letter converted to a number, separated by dots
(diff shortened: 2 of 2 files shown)


PipoCanaja:
PR for extended support of efficient IP devices will come after a 2nd PR adding a DERIVE support for the 'count' sensor.

murrant:
Should we be forcing it to use numeric instead?

PipoCanaja:
I can think of one corner case at least: Let's imagine some device would use a string index, but containing only digits ;) That would be bad and will hopefully never occur, but ...

PipoCanaja:
That beeing said, it would be far easier for the user who defines the YAML file if he does not have to think about which index to use, so I will be happy to implement it.

murrant:
I'm saying there is an output option on snmp commands to just show string indexes as numeric.  There is no reason to use them as strings.  It could be handy to have the information available, but not sure that is worth it.

PipoCanaja:
Got you. Issue would be the other way around. We don't know when to convert the num_index to string (unless we parse the mib or add an option to the YAML file to tell YamlDiscovery to do the conversion. 
And some devices are not providing any OID to list all the string keys. (I had the issue already). 
So I think we would be on the safe side if we keep the same default behaviour we have now.

murrant:

https://community.librenms.org/t/existing-device-with-custom-oid/9032/6

PipoCanaja:
Hi @murrant 
Re-checked this, and still think this PR is a safe way to do it : 
- No change for any existing config.
- Possible to use string-based index if needed.
If it is OK with you like this, I would love to see it merged so I can start upgrading again one of my prod instances :D 
Thnx

PipoCanaja:
I installed this patch on an instance that monitors efficientIP DNS/DHCP appliances, and it did well so far. No impact in any existing YAML files.

murrant:

PR #10413, opened 2019-07-04
WIP - Allow 'count' sensors to adapt RRD type (COUNTER, DERIVE, instead of GAUGE)

PipoCanaja:
This PR allows the 'count' sensor to use RRDs with COUNTER and DERIVE instead of GAUGE which remains the default. 
Type can be changed by prefixing the 'sensor_index' with `counter__` or `derive__`.

Both of #10410 and #10413 will allow sensors from EfficientIP devices to be loaded. 

Capture d’écran 2019-07-05 à 13 00 41

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Developing/os/Health-Information.md
+++ b/doc/Developing/os/Health-Information.md
@@ -109,8 +109,8 @@ The only sensor we have defined here is airflow. The available options are as fo
   identify this sensor. `{{ $index }}` will be replaced by the `index`
   from the snmp walk.
   For the 'count' sensor class, it is possible to prefix the index with
-  ```counter__``` or ```derive__``` to change the RRD file type from
-  default GAUGE type. 
+  ``` counter__ ``` or ``` derive__ ``` to change the RRD file type from
+  default GAUGE type.
 - `skip_values` (optional): This is an array of values we should skip
   over (see note below).
 - `skip_value_lt` (optional): If sensor value is less than this, skip the discovery.
--- a/doc/Developing/os/Health-Information.md
+++ b/doc/Developing/os/Health-Information.md
@@ -108,6 +108,9 @@ The only sensor we have defined here is airflow. The available options are as fo
 - `index` (optional): This is the index value we use to uniquely
   identify this sensor. `{{ $index }}` will be replaced by the `index`
   from the snmp walk.
+  For the 'count' sensor class, it is possible to prefix the index with
+  ```counter__``` or ```derive__``` to change the RRD file type from
+  default GAUGE type. 
 - `skip_values` (optional): This is an array of values we should skip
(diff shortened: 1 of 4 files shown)


jozefrebjak:
Fantastic (y)

jozefrebjak:
@PipoCanaja Maybe, will be nice to have an option to speficy unit. What you think ?

doc/Developing/os/Health-Information.md
```diff
@@ -108,6 +108,9 @@ The only sensor we have defined here is airflow. The available options are as fo
 - `index` (optional): This is the index value we use to uniquely
   identify this sensor. `{{ $index }}` will be replaced by the `index`
   from the snmp walk.
+  For the 'count' sensor class, it is possible to prefix the index with
```
murrant: Why are you putting this setting in the index?  That doesn't seem like the right place for it.

PipoCanaja: The other option I see is to create a new column in the sensor DB. It would definitly be a little cleaner. Any other idea ?

PipoCanaja:
> @PipoCanaja Maybe, will be nice to have an option to speficy unit. What you think ?

Hi @jozefrebjak This is fairly complicated to do right now. If we decide to go with dedicated DB field for type of RRD, I could very well add a unit field as well. We'll see how it goes.

doc/Developing/os/Health-Information.md
PipoCanaja: @murrant Should I go for the variant where I add a column (enum) with the type to the sensor table ? 
If yes, I would also try to add the unit to the sensor, so we could (at least for "count") override the unit as well.

murrant: It doesn't need to be stored in the database at all, it can just be in the yaml.

murrant:
It doesn't make as much sense for non-count sensors.

PipoCanaja:
> It doesn't make as much sense for non-count sensors.

At least for temperature, also for power that can be in dB(x) or W (for signal), and a few others where not all vendors agree on how to provide the values. So I think it would'nt be bad to have a way to change the default anyway.

doc/Developing/os/Health-Information.md
PipoCanaja: That would make it for discovery, but I also need this information during polling (so I can do the diff between previous and current value). And the YAML file is not read again during each polling session (which is good for perf issues). 
So unless the information is stored in the DB, I don't see a way.

murrant: I think it should be declared separately at least in the yaml.

PR #10416, opened 2019-07-05
Transport API - Doc for placeholders

PipoCanaja:
The available placeholders are the same as the Templates documentation but the syntax is not exactly identical.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Alerting/Transports.md
+++ b/doc/Alerting/Transports.md
@@ -68,7 +68,9 @@ API transports definitions are a bit more complex than the E-Mail configuration.
 allows to reach any service provider using POST or GET URLs (Like SMS provider, etc).
 
 The API-Option field can have the same placeholders as defined in the
-[Template-Syntax](Templates.md#syntax).
+[Template-Syntax](Templates.md#syntax) but without the 'alert->' prefix.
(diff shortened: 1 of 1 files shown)


PipoCanaja:
@laf Here is a proposal to improve the doc. I would like to avoid writing the complete list of options (verbose) but what do you think ?

laf:
> @laf Here is a proposal to improve the doc. I would like to avoid writing the complete list of options (verbose) but what do you think ?

@PipoCanaja How many are there? It would be good to indicate the list or where to find it.

PipoCanaja:
> > @laf Here is a proposal to improve the doc. I would like to avoid writing the complete list of options (verbose) but what do you think ?
> 
> @PipoCanaja How many are there? It would be good to indicate the list or where to find it.

All the ones that are already defined in the existing doc for Blades templates, except that instead of $alert->xxxxx , you get them with $xxxxx. And the exact list might even depends on the alerts ... I'll check and I'll export a list, we'll see what we get.

PipoCanaja:
@laf More details are now added. A limitation is that there is no way to parse the fault array, contact array, etc etc. Unless we decide to rewrite the blades completely. 
- Should I explicitly write that as well ?
- Should I add more examples ?

PR #10421, opened 2019-07-07
MPLS Services

vitalisator:
Extends Routing->MPLS
- SDP (Service Distribution Points)
- SDP Bindings
- Services vpls, vprn, ...
- SAP (Service Access Points)

include Discovery, Polling, HTML global and device table views
image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


app/Models/MplsSap.php
```diff
@@ -0,0 +1,49 @@
...
+    // ---- Define Relationships ----
+
+    public function binds()
+    {
+        return $this->hasMany('App\Models\MplsSdpBind', 'svc_oid');
+    }
+    
+    public function services()
+    {
+        return $this->hasMany('App\Models\MplsService', 'svc_oid');
```
murrant: This relationship won't work because svc_oid is not unique across all devices.
You need to store the unique id of the service.  (Same with binds)

vitalisator: I confess that I don't really understand for what are object relationships needed, just copied that part from app/Models/MplsLsp.php :-o
can I just omit the binds() and services() methods? I do not see that these were used somewhere.

otherwise getting and store the keys would make the polling and discovery code more complex.

murrant: no, because it isn't the code that is wrong, it is the database table structure.  Yes, it does make it more complex in the lsp code, I passed the parent objects to the child polling code so they wouldn't have to be re-fetched from the db.

vitalisator: I have updated the code but something is wrong with unit tests now.
--- Expected
+++ Actual
@@ @@
     'mpls_sdps' => Array (...)
     'mpls_sdp_binds' => Array (
         0 => Array (
-            'sdp_id' => 19
-            'svc_id' => 13
+            'sdp_id' => 2
+            'svc_id' => 1

everitime I run the
./scripts/save-test-data.php -o timos -m mpls
the Script generate new ids.
On my development environment I dont see any problems in the database tables. The Ids are consistent.

murrant:
I think this is looking good.  Just needs some testing.

rsys-dev:
LGTM

peelman:
seeing significantly longer poll times on our SAS-Ks (up to over a minute per device, from 15-30 seconds before).  Based on the timing and the affected devices, pretty sure this is the culprit.

murrant:
@peelman what do you want to do about it?  It is polling more data.

peelman:
In the cases I'm seeing, its becoming aware of 3 VRFs and 1 L3 interface.

Screen Shot 2019-08-14 at 9 07 42 AM

And that's double-to-tripling the polling time?  was just looking for some @murrant- and @laf-esque levels of scrutiny and optimization here; the device itself doesn't seem to be the bottleneck, it'll feed back the data about as fast as you can ask for it, at least on the SAS-K5s and SAS-K12s I have interrogated manually.  For the amount of data returned its a huge time penalty per device.  Just wondering if things aren't being batched, or if its poor structure on the Timetra/Alcatel/Nokia side, or if under what we're asking for, its not as quick as it is when you ask for it manually.

vitalisator:
vrf polling was changed in #9622 for NOKIA
can you see wich module consums the most polling time?
I have some NOKIA 7210 SAS-R6 switches, the overall polling time is here under 10 seconds
The MPLS Module cunsumes halt of the time.
grafik

peelman:
@vitalisator these are primarily SAS-K5's.  They don't have the capability to do MPLS.  They can just do some primitive VPLS tag pushing/popping, but that's about it.

Here's the poller modules graph for a typical SAS-K5 for us for the past 2 weeks.  Its pretty obvious when I turned off the MPLS monitoring.

graph php

peelman:
Same device, same graph, over 2 months.

graph-1 php

PR #10424, opened 2019-07-09
FS.net pdu

ig0rb:
Hi, it's my first test with github, I hope i did it in the right way...

I added a basic support for an SNMP monitored PDU, the device doesn't have a MIB file.

DO NOT DELETE THIS TEXT

#### Please note

- [*] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers


includes/definitions/fs-net-pdu.yaml
```diff
@@ -0,0 +1,13 @@
...
+text: 'FS.COM monitored pdu'
+type: power
+icon: fs
+group: fs
+over:
+    - { graph: device_voltage, text: Voltage }
+    - { graph: device_current, text: Current }
+discovery:
+    - sysObjectID:
+        - .1.3.6.1.4.1
```
PipoCanaja: Please use the complete OID here, cause .1.3.6.1.4.1 is far to wide. It covers all devices of all brands ... 

You have to use .1.3.6.1.4.1.30966 which restricts to the company that makes these PDUs ...

ig0rb: Hi, sorry for the delay, I did the changes, I hope it's ok.

Thanks a lot-

PR #10425, opened 2019-07-09
[WIP] DELL F10 sensor fixes and removing zeros, BGP polling fix

cloudprodz:
DO NOT DELETE THIS TEXT
Dell dnos 9
Fix correct optics lanes, removed 0 temperatures. 
Fix negative AS numbers to real AS numbers. 
Fix BGP peer status. 
WIP: BGP ipv6 peers
Need help for BGP polling fix. Dell does not respond to specific instances on bgp snmp entries but does return the values when snmp groups are queried. Thus implemented some kind of a bulkwalk for that, probably need to check formality. 

For some reason snmpwalk_group does not allow changing snmpwalk flags and it uses numeric responses by default. Created new function for that, maybe should change the original function to allow specifying flags?

#### Please note

#### Testers


cloudprodz:
[bgpwalk.txt](https://github.com/librenms/librenms/files/3373043/bgpwalk.txt)

includes/polling/bgp-peers.inc.php
```diff
@@ -10,6 +10,9 @@
     if (!empty($peers)) {
         if ($device['os'] == 'junos') {
             $peer_data_check = snmpwalk_cache_long_oid($device, 'jnxBgpM2PeerIndex', '.1.3.6.1.4.1.2636.5.1.1.2.1.1.1.14', $peer_data_tmp, 'BGP4-V2-MIB-JUNIPER', 'junos');
+        } elseif ($device['os'] === 'dnos') {
+            $peer_data_checki = snmpwalk_group($device, 'bgpPeerEntry', 'BGP4-MIB');
+            $peer_data_check = array();
```
PipoCanaja: You probably have a typo here. $peer_data_checki is wrong and gives you an empty array. (and you anyway empty the array after).

cloudprodz: It was intentional. The way it seemed to work was to poll each device separately if the peer_data_check variable had data. Renamed variable and created a string function with different snmp flag.

cloudprodz:
Snmpwalks and difference between flags:
[bgpwalk.txt](https://github.com/librenms/librenms/files/3376050/bgpwalk.txt)
Also seems that dellNetBgpM2 would be better to get ipv6 peers aswell but seems more difficult to implement.

includes/polling/bgp-peers.inc.php
PipoCanaja: Yep, $peer_data_check is here to know if there is at least 1 peer. If yes, we can continue and collect the list of them, parse it, and store in $peer_data what is needed. 

If you need/want to skip the check part (because you have no way to get the number of peers without getting the complete list of them for instance), then you could directly fill $peer_data  instead of creating $peer_data_checki. 

That being said, BGP4-MIB is the standard MIB so this data is already parsed, you should not have to rewrite it here, all the code is already included. If something is broken in it, it should probably be identified and fixed, instead of adding code duplication. 

What's wrong with current code ?

cloudprodz: Check the beginning of bgpwalk.txt I attached in the comment below, the current code implementation does not get any proper response from Dell since it queries specific peers and not the whole table, which does not seem to be supported by Dell. Unless this is some configuration error on my end?

So what I did was create new array for polling the whole table without specific peers and then in the peer loop part use the new array instead of polling each separately. But probably would be better to start using dellNetBgpM2 but that needs some more work.

Jellyfrog:

PR #10427, opened 2019-07-09
Fix NetAgent II battery voltage

CirnoT:
- For single phase UPSes, variable name for battery voltage was incorrect hence sensor saved in SQL had "" as OID.
- Additionally another issue was fixed where divisor for battery voltage on single-phase UPSes was set to 1, when SNMP card reports 27.0V as 270 (divisor 10).

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/discovery/sensors/voltage/netagent2.inc.php
```diff
@@ -389,8 +389,8 @@
 // BATTERY Voltage
 # Set divisor and limit ranges 1 phase UPS systems
 if ($in_phaseNum == '1') {
-    $battery_voltage_oid = '.1.3.6.1.4.1.935.1.1.1.2.2.2.0';
-    $battery_voltage1 = snmp_get($device, $battery_voltage_oid, '-Oqv');
+    $battery_voltage_oid1 = '.1.3.6.1.4.1.935.1.1.1.2.2.2.0';
+    $battery_voltage1 = snmp_get($device, $battery_voltage1_oid, '-Oqv');
```
PipoCanaja: Hi,
This does not seems right. The 1 is not at the same position in the var name ... Did you copy paste in the github GUI ? it is easier (and less error-prone) to use git in CLI ...

PipoCanaja: Should be ``` $battery_voltage1_oid ``` in both lines 392 and 393.

CirnoT: Nah, I used CLI but modified the file without really testing it as I've modified it separately. The code below https://github.com/librenms/librenms/blob/1588b75927d2137b185ab3393e83fba8db946459/includes/discovery/sensors/voltage/netagent2.inc.php#L412 references variable names with '1' in name hence I am assuming that '1' was desired by author of this code for some reason.

CirnoT: Fixed it, now both 3-phase and single phase are the same (with single-phase mirroring working 3-phase)

PipoCanaja:
Looks better now :) Could you also add some test data, to allow the tests to catch any breaking issue with this code ?

includes/discovery/sensors/voltage/netagent2.inc.php
PipoCanaja: All good.

CirnoT:
The test data is already there with required OIDs and it seems that test suite did not catch it. I've ran update on sensors module to see what difference would it make but there was none, hence I don't think I need to create new snmprec file?

PipoCanaja:
It means that the current data is does not include the new OIDs you added. This is normal and expected, only the necessary test data is collected, if you now need more than what's currently there, you need to collect it again. 
So the best would be to collect a new variant, and run save-test-data on it, and attach the snmprec and json files to the PR.
Thanx

CirnoT:
Added test data

PR #10458, opened 2019-07-24
Graylog entry matching device if source is not hostname or primary ip

rsys-dev:
Changed Graylog Table to generate proper device link as it only worked if source is hostname but not if it is ip.
Added the ability to parse Severity and Facility Levels according to rfc5424 (6.2.1). Can be activated by setting $config['graylog']['severity-names'] = 'true'; and $config['graylog']['facility-names'] = 'true';
Added the ability to match any device address to show Graylog entries at /device/device=/tab=logs/section=graylog/ even if source is not hostname or primary ip.
Fixed missing quotation marks in <script> block in includes/html/common/graylog.inc.php:46, causing graylog page only load header with empty table. Introduced in librenms#10447

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


app/Models/Device.php
```diff
@@ -141,6 +141,19 @@ public static function findByIp($ip)
 
         return null;
     }
+    
+    public static function findByHostnameOrIp($hostnameOrIp)
+    {
+        $device = Device::findByHostname($hostnameOrIp);
```
murrant: Instead of doing two sql queries, you can just check to see if the input is an IP address.
Also, on the second query, you can just return it, you don't have to check if it is null to see if you should return it or null.

`\LibreNMS\Util\IP::isValid($hostnameOrIp)`

rsys-dev: > Instead of doing two sql queries, you can just check to see if the input is an IP address.

Do you mean like this to only "findByIp" if parameter is valid ip?
```
if (IP::isValid($hostnameOrIp)) {
    $device = Device::findByIp($hostnameOrIp);
    if (isset($device)) {
        return $device;
    }
}
return Device::findByHostname($hostnameOrIp);
```

> Also, on the second query, you can just return it, you don't have to check if it is null to see if you should return it or null.

Yes you are right. Will change this

murrant:
Thanks for your pull request!

Can you explain why you have created configuration options?  Why would people want them disabled?

rsys-dev:
I thought there may be people who like the Severity and Facility numbers more than names.
I also can imagine, matching to any ip address instead of only primary address might be unwanted in some cases.
For example you monitor one of your routers with 192.168.0.1.
There may be a second device at another location you monitor via any public IP Address but also has private address 192.168.0.1. To prevent logs to be showed on both devices you can choose to not enable it.

murrant:
Just show both the number and the text?  Maybe we should use translation for the text...

The multiple IPs config makes sense to me.

rsys-dev:
> Just show both the number and the text?

Will change it to show as "NUMBER (NAME)" for both, severity and facility.

> Maybe we should use translation for the text...

You mean localisation or translate text to number?
I'm not sure if localisation is neccessary, text to number is no problem

murrant:
Yes, I mean localization.   You have those arrays of int to string, you could just put that in the language files and call `__('syslog.level.' . $level)` to get the text.

rsys-dev:
Ok, created a syslog.php lang file for english.
Also tried to parse Facility and Severity Strings to numbers to have a consitent output on the website "NUMBER (NAME)", but for facility there are 2 different representations in graylog and sadly there are some identical string representations for different facility numbers (security/authorization for 4 and 10, clock for 9 and 15). As this may result in wrong numbers reportet, I removed this for facility.
Also merged "levelName" and "facilityName" functions to "syslogPrioParser" with first parameter "facility" or "severity" and second parameter is the Value

app/Models/Device.php
```diff
@@ -141,6 +141,18 @@ public static function findByIp($ip)
 
         return null;
     }
+    
+    public static function findByHostnameOrIp($hostnameOrIp)
```
murrant: Heh, just took a look at this and this should be the entire function:
```php
public static function findByHostnameOrIp($hostnameOrIp)
{
    return Device::findByIp($hostnameOrIp) ?: Device::findByHostname($hostnameOrIp);
}
```

rsys-dev: Oh yes, you are right, didn't see that.

murrant:
@rsys-dev Do you mind if I commit some changes?

rsys-dev:
No problem

murrant:
Ok, code looks good now.  I changed it so it only translates if the value is a number. Seem ok?

Just need docs for the new config setting. doc/Extensions/Graylog.md

rsys-dev:
Seemd not to work anymore. In my test environment I have a server reporting log entries with another address than the primari and no log entry is matched to this host now.
I'll check whats going wrong

rsys-dev:
Looks good to me now

murrant:
I guess my test env wasn't complex enough to catch that.  Thanks :)

murrant:
Nice work on this btw.

vitalisator:
:+1:

PR #10490, opened 2019-08-04
Use standard cache

murrant:
Change default to file
remove usage of the "cache" folder

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/DB/Schema.php
+++ b/LibreNMS/DB/Schema.php
@@ -25,6 +25,7 @@
 
 namespace LibreNMS\DB;
 
+use Cache;
 use LibreNMS\Config;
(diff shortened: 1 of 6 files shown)


Jellyfrog:
Looks good, but I fail to see (in my head) why the test fails.

laf:
Also can't see why this would fail tests but docs need an update as it references the old cache file. `./doc/Developing/os/Initial-Detection.md`

murrant:
It fails because Laravel is not booted for the test so we cant access the config() helper...
So tests need to be reworked to support this.  That is kinda why I shelved it for a bit.

PR #10501, opened 2019-08-07
Bug: Fixed Incorrect device match in Graylog

rsys-dev:
Incorrect device match fixed. Because getAddresses(Device $device) also returned 127.0.0.1 and ::1, always matched the first device with a loopback address. Introduced in https://github.com/librenms/librenms/pull/10458

Filtered 127.0.0.1 and ::1 in return statement of getAddresses(Device $device)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/ApiClients/GraylogApi.php
+++ b/app/ApiClients/GraylogApi.php
@@ -141,15 +141,15 @@ class GraylogApi
         ]);
 
         if (Config::get('graylog.match-any-address')) {
-            $addresses = $addresses->merge($device->ipv4->pluck('ipv4_address'))
+            $addresses = $addresses->merge($device->ipv4->pluck('ipv4_address')
                 ->filter(
-                    function($address) {
+                    function ($address) {
                         return $address != "127.0.0.1";
                     }
-                ))->merge($device->ipv6->pluck('ipv6_address'))
+                ))->merge($device->ipv6->pluck('ipv6_address')
                 ->filter(
-                    function($address) {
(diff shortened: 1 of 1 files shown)


PipoCanaja:
Hi @rsys-dev 
Makes sense reading it but I have absolutely no way to test it. We'll wait for another admin to review it.

app/ApiClients/GraylogApi.php
```diff
@@ -145,7 +145,11 @@ public function getAddresses(Device $device)
                 ->merge($device->ipv6->pluck('ipv6_address'));
         }
 
-        return $addresses->filter()->unique();
+        return $addresses->filter(
```
murrant: You should only filter 127.0.0.1 from the addresses tables, not when they are hostname or ip

rsys-dev: > You should only filter 127.0.0.1 from the addresses tables, not when they are hostname or ip. So basically, move this filter up a few lines.

Changed IPv6 to uncompressed form and moved the filter up to line 143.

>  Also, try not to use double negatives it makes things hard to understand: return $address == '127.0.0.1' || $address == '0000:0000:0000:0000:0000:0000:0000:0001';

Not needed anymore but this does not the same as `return $address != '127.0.0.1' && $address != '::1';` or did I miss something? Your code will return only `127.0.0.1` and `::1`, my code will return everything **except** `127.0.0.1` and `::1`, or am I wrong?

murrant:

PR #10502, opened 2019-08-07
Ports API: Workaround for ifNames with slashes

murrant:
Basically,

/api/v0/devices/3/ports/0%2f4 -> /api/v0/devices/3/ports/ifName?ifName=0/4
/api/v0/devices/3/ports/0%2f4/port_bits -> /api/v0/devices/3/ports/ifName/port_bits?ifName=0/4

Or you can just use port_id. If ifName doesn't contain / it will still work.

Also adds port_id to the default output for the ports list.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/API/Devices.md
+++ b/doc/API/Devices.md
@@ -469,16 +469,13 @@ Output:
     "count": 3,
     "ports": [
         {
-            "ifName": "lo",
-            "port_id": 1
+            "ifName": "lo"
         },
         {
-            "ifName": "eth0",
-            "port_id": 2
+            "ifName": "eth0"
         },
         {
-            "ifName": "eth1",
-            "port_id": 14
+            "ifName": "eth1"
         }
     ]
 }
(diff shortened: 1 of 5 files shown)


murrant:

https://community.librenms.org/t/api-issue-url-encoding-issue-with-port-graph/9146/7

kate66:
This is still not working as it was previously, the API call is getting a 404.

`api.ip - - [09/Aug/2019:09:05:29 -0400] "GET /api/v0/devices/19/ports/ge-0%2F0%2F41/port_bits?from=1565269528&to=1565355928&width=1000 HTTP/1.1" 404 254 "-" "-"`

laf:
Is this backwards compatible? If not, I'd say this is a breaking change that shouldn't got in without a version bump of the API

murrant:
@laf it is a breaking change that we can't fix without breaking bc.  Slashes in routes breaks the rfc :(.  Additionally, there is a workaround for the last route segment, but we can't use that because of a collision with the other route...  So bad stuff all around.

laf:
Is this because the api is now going through Laravel I assume?

murrant:
Yes.   I might have another work around, basically subverting the Laravel routing.

murrant:
@laf ok, should restore original functionality.

murrant:

https://community.librenms.org/t/api-issue-url-encoding-issue-with-port-graph/9146/14

laf:
Awesome. Thanks for keeping it compatible. Makes it easier at work :)

murrant:

murrant:

https://community.librenms.org/t/api-issue-with-search-port/13341/1

PR #10509, opened 2019-08-09
Added Graylog to device overview and log level filter mechanism

rsys-dev:
I added Recent Graylog Entries to device overview like Recent Syslog or Events and 2 Config options to influence the behaviour of the "Recent Graylog" table on the overview page of the devices (row count and maximum log level to show).
Also added a log level dropdown to filter existing Graylog Pages (device subpage and graylog overview) and a graylog widget setting to set the maximum log level for the widget.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/html/print-graylog.inc.php
+++ b/includes/html/print-graylog.inc.php
@@ -38,14 +38,23 @@ $tmp_output = '
 <script>
 ';
 
+$rowCount = \LibreNMS\Config::get('graylog.device-page.rowCount');
+$maxLevel = \LibreNMS\Config::get('graylog.device-page.maxLevel');
+
 $tmp_output .= '
     $.ajax({
         type: "post",
         data: {
-            device: "' . (isset($filter_device) ? $filter_device : '') . '"
+            device: "' . (isset($filter_device) ? $filter_device : '') . '",
+            '. ($rowCount? 'rowCount: '.$rowCount .',' : '') .'
+            '. ($maxLevel? 'maxLevel: '.$maxLevel .',' : '') .'
         },
         url: "' . url('/ajax/table/graylog') . '",
         success: function(data){
+            if (data.rowCount == 0) {
+                $("#graylog-card").remove();
+                return;
+            }
             var html = "<tbody>";
             $("#graylog").append("<tbody></tbody>");
             $.each(data.rows, function(i,v){
--- a/includes/html/print-graylog.inc.php
+++ b/includes/html/print-graylog.inc.php
@@ -1,0 +1,67 @@
+<?php
+/*
+ * LibreNMS
+ *
+ * Copyright (c) 2014 Neil Lathwood <https://github.com/laf/ http://www.lathwood.co.uk/fa>
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ *
+ * @package    LibreNMS
+ * @subpackage webui
+ * @link       http://librenms.org
+ * @copyright  2017 LibreNMS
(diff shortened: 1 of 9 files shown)


includes/html/print-graylog.inc.php
```diff
@@ -0,0 +1,74 @@
+<?php
+/*
+ * LibreNMS
+ *
+ * Copyright (c) 2014 Neil Lathwood <https://github.com/laf/ http://www.lathwood.co.uk/fa>
```
murrant: Just inline this file in the overview/graylog.inc.php.

Also, you should not be assigning copyright to other people.  Please fix in all files.

resources/views/widgets/graylog.blade.php
```diff
@@ -31,7 +31,8 @@
             return {
                 stream: "{{ $stream }}",
                 device: "{{ $device }}",
-                range: "{{ $range }}"
+                range: "{{ $range }}",
+                loglevel: "{{ $loglevel }}",
```
murrant: Javascript json does not allow commas on the last item.

resources/views/widgets/settings/graylog.blade.php
```diff
@@ -29,6 +29,22 @@
         <input type="number" min="1" class="form-control" name="limit" id="limit-{{ $id }}" placeholder="@lang('Page Size')" value="{{ $limit }}">
     </div>
 
+    <div class="form-group">
+        <label for="loglevel-{{ $id }}" class="control-label">@lang('Log Level')</label>
+        <select name="loglevel" id="loglevel-{{ $id }}" class="form-control">
+            <option value="" disabled @if($loglevel == null) selected @endif>@lang('Log Level')</option>
+            <option value="0" @if($loglevel === 0) selected @endif>@lang('(0) Emergency')</option>
```
murrant: These should use the translation strings `syslog.severity.0`

includes/html/print-graylog.inc.php
rsys-dev: Ok, this was copy and paste as I copied print-syslog.inc.php and modified it for Graylog.
Should I add Copyright to overview/graylog.inc.php?
Every other file existed before, because of that I did not change / add any copyright, or should I change anything here?

resources/views/widgets/settings/graylog.blade.php
rsys-dev: Changed this. Did not realize (also never tried) that I can use these translation Strings there.

resources/views/widgets/graylog.blade.php
rsys-dev: You are right, changed

resources/views/widgets/settings/graylog.blade.php
murrant: You could even use a foreach so you don't have to repeat every option ;)

includes/html/print-graylog.inc.php
murrant: You may add your copyright on the files, that is up to you.  All files will be licensed under GPLv3 as per our license.  Individual file license headers are just for good measure.

resources/views/widgets/settings/graylog.blade.php
rsys-dev: I can't get this to work with the translation strings.
```
@foreach(range(0, 7) as $level)
                <option value="{{ $level }}" @if($loglevel == $level && $loglevel != null) selected @endif>({{ $level }}) @lang('syslog.severity.{{ $level }}')</option>
            @endforeach
```
Results in a select field with `(0) syslog.severity.` and so on as options,

```
@foreach(range(0, 7) as $level)
                <option value="{{ $level }}" @if($loglevel == $level && $loglevel != null) selected @endif>({{ $level }}) @lang('syslog.severity.'{{ $level }})</option>
            @endforeach
```
results in `Problem with backend`

rsys-dev:
Tested and works like expected

murrant:

PR #10465, opened 2019-07-27
Seafile Server Monitoring

SourceDoctor:
Monitoring for a Seafile Server

[About Seafile Server](https://www.seafile.com/en/home/)

PR #10354, opened 2019-06-20
Added certificate file validity check for services (#10354)

SourceDoctor:
check for certificates with following attributes
- age in days
- remaining validity in days

PR #10511, opened 2019-08-10
Allow user activation/deactivation (MySQL auth)

SourceDoctor:
with this enhancement user login can be deactivated, without deleting them

### DIFF
--- a/LibreNMS/Authentication/MysqlAuthorizer.php
+++ b/LibreNMS/Authentication/MysqlAuthorizer.php
@@ -20,7 +20,7 @@ class MysqlAuthorizer extends AuthorizerBase
         $username = $credentials['username'] ?? null;
         $password = $credentials['password'] ?? null;
 
-        $user_data = User::thisAuth()->where(['username' => $username])->value('password', 'enabled');
+        $user_data = User::thisAuth()->where(['username' => $username])->select('password', 'enabled')->first();
         $hash = $user_data->password;
         $enabled = $user_data->enabled;
 
--- a/LibreNMS/Authentication/MysqlAuthorizer.php
+++ b/LibreNMS/Authentication/MysqlAuthorizer.php
@@ -20,8 +20,9 @@ class MysqlAuthorizer extends AuthorizerBase
         $username = $credentials['username'] ?? null;
         $password = $credentials['password'] ?? null;
 
-        $hash = User::thisAuth()->where(['username' => $username])->value('password');
-        $enabled = User::thisAuth()->where(['username' => $username])->value('enabled');
+        $user_data = User::thisAuth()->where(['username' => $username])->value('password', 'enabled');
+        $hash = $user_data->password;
+        $enabled = $user_data->enabled;
 
         if (! $enabled) {
             throw new AuthenticationException($message = 'login denied');
--- a/LibreNMS/Authentication/MysqlAuthorizer.php
+++ b/LibreNMS/Authentication/MysqlAuthorizer.php
@@ -21,6 +21,11 @@ class MysqlAuthorizer extends AuthorizerBase
         $password = $credentials['password'] ?? null;
 
         $hash = User::thisAuth()->where(['username' => $username])->value('password');
+        $enabled = User::thisAuth()->where(['username' => $username])->value('enabled');
+
+        if (! $enabled) {
+            throw new AuthenticationException($message = 'login denied');
+        }
 
         // check for old passwords
         if (strlen($hash) == 32) {
--- a/sql-schema/282.sql
+++ b/sql-schema/282.sql
@@ -1,1 +1,0 @@
-ALTER TABLE `users` ADD `enabled` TINYINT NOT NULL DEFAULT '1';
--- a/sql-schema/282.sql
+++ b/sql-schema/282.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `users` ADD `enabled` TINYINT NOT NULL DEFAULT '1';
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -158,9 +158,9 @@ class User extends Authenticatable
         $this->attributes['can_modify_passwd'] = $modify ? 1 : 0;
     }
 
-    public function setEnabledAttribute($modify)
+    public function setEnabledAttribute($enable)
     {
-        $this->attributes['enabled'] = $modify ? 1 : 0;
+        $this->attributes['enabled'] = $enable ? 1 : 0;
     }
 
     // ---- Define Relationships ----
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -13,7 +13,7 @@ class User extends Authenticatable
     use Notifiable;
 
     protected $primaryKey = 'user_id';
-    protected $fillable = ['realname', 'username', 'email', 'level', 'descr', 'can_modify_passwd', 'auth_type', 'auth_id'];
+    protected $fillable = ['realname', 'username', 'email', 'level', 'descr', 'can_modify_passwd', 'auth_type', 'auth_id', 'enabled'];
     protected $hidden = ['password', 'remember_token', 'pivot'];
     protected $attributes = [ // default values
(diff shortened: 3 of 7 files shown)


sql-schema/282.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `users` ADD `enabled` TINYINT NOT NULL DEFAULT '1';
```
PipoCanaja: Hi,
Since we moved to Laravel, we cannot use the <id>.sql files. You cannot change them at all. You need to use Laravel migrations.
https://laravel.com/docs/5.8/migrations

You can have a look at this folder for existing migration classes: 
./database/migrations

SourceDoctor:
@PipoCanaja 
i changed database migration

sql-schema/282.sql
SourceDoctor: @PipoCanaja 
migration changed :-)

PipoCanaja: Looks better now :) Travis will still complain because the db schema is not up to date. 
Can be done by running : 
./script/build-schema.php
That should help Travis to be happy.

SourceDoctor: i tried it, but it doesn't work at me:

```
PHP Fatal error:  Uncaught Error: Call to undefined function LibreNMS\Util\app() in /opt/librenms/LibreNMS/Util/Laravel.php:71
Stack trace:
#0 /opt/librenms/LibreNMS/Util/Laravel.php(38): LibreNMS\Util\Laravel::isBooted()
#1 /opt/librenms/includes/init.php(96): LibreNMS\Util\Laravel::bootCli()
#2 /opt/librenms/includes/sql-schema/update.php(36): require('/opt/librenms/i...')
#3 /opt/librenms/scripts/build-schema.php(7): require('/opt/librenms/i...')
#4 {main}
  thrown in /opt/librenms/LibreNMS/Util/Laravel.php on line 71

Fatal error: Uncaught Error: Call to undefined function LibreNMS\Util\app() in /opt/librenms/LibreNMS/Util/Laravel.php:71
Stack trace:
#0 /opt/librenms/LibreNMS/Util/Laravel.php(38): LibreNMS\Util\Laravel::isBooted()
#1 /opt/librenms/includes/init.php(96): LibreNMS\Util\Laravel::bootCli()
#2 /opt/librenms/includes/sql-schema/update.php(36): require('/opt/librenms/i...')
#3 /opt/librenms/scripts/build-schema.php(7): require('/opt/librenms/i...')
#4 {main}
  thrown in /opt/librenms/LibreNMS/Util/Laravel.php on line 71
```

PipoCanaja: what does ./validate.php says ? does the daily script run regularly ? This script should run smoothly and update the misc/db_schema.yaml file

SourceDoctor: @PipoCanaja 
i run this now in my running development/testing instance of librenms, and it worked.
Script missed some laravel stuff which will be installed on configuration.

i hope now everything is fine.
by the way. I've got 2 Application enhancements since weeks as pull request waiting for merge. Is there a chance to speed this a little bit up? ;-)
Certificate validation check #10354 
Seafile Monitoring #10465

PipoCanaja: For current PR, you still need to remove the changes you did on 282.sql. These files cannot be changed anymore.

For the other PRs, we lack admins to review them, we lack testers to test them, we don't always have the right environnement to test them  etc etc. We do our best on our free time, which is usually not enough.

SourceDoctor: @PipoCanaja 
282.sql removed

SourceDoctor:
@PipoCanaja 
everything else to modify?

PipoCanaja:
LGTM. @librenms/reviewers Please have a look at this one and merge it if it is OK.

SourceDoctor:
@murrant @PipoCanaja 
something else to do here before merging?

PipoCanaja:
Hi
A little bit "disconnected" these days, sorry :) From my point of view it looks good. I leave it to @murrant for approval. 
Will still be internet blind for a few days :)

LibreNMS/Authentication/MysqlAuthorizer.php
```diff
@@ -21,6 +21,11 @@ public function authenticate($credentials)
         $password = $credentials['password'] ?? null;
 
         $hash = User::thisAuth()->where(['username' => $username])->value('password');
+        $enabled = User::thisAuth()->where(['username' => $username])->value('enabled');
```
murrant: you should get both values in one query instead of two mysql queries.

murrant:
Also, this should be supported on all auth types since you show it for all.

LibreNMS/Authentication/MysqlAuthorizer.php
SourceDoctor: combined both requests

SourceDoctor:
> Also, this should be supported on all auth types since you show it for all.

i see only this authorization Methods:

- Mysql  -> local controlled Authorization
- LDAP   -> remote controlled Authorization
- Radius -> remote controlled Authorization
- Http Auth -> local controlled Authorization

for Mysql Auth is is written.
Http Auth inherits Mysql Auth, so it's also supported.

LDAP and Radius for my intention local User Management Page should be complete disabled because of remote Authentication.

If i oversee something give me a hint, please :-)

LibreNMS/Authentication/MysqlAuthorizer.php
```diff
@@ -20,8 +20,9 @@ public function authenticate($credentials)
         $username = $credentials['username'] ?? null;
         $password = $credentials['password'] ?? null;
 
-        $hash = User::thisAuth()->where(['username' => $username])->value('password');
-        $enabled = User::thisAuth()->where(['username' => $username])->value('enabled');
+        $user_data = User::thisAuth()->where(['username' => $username])->value('password', 'enabled');
```
murrant: you need select, but if you want to support enabled in all authorizers, you'll need to check enabled in LegacyUserProvider::validateCredentials().  I'm not sure if that will invalidate cookie auth though.

SourceDoctor: fixed query :-)

i think "enabled" should be available for local Authentication (mysql based) only, because LDAP Controller does this itself.
But not sure how to check in "resources/views/user/form.blade.php" and "resources/views/user/index.blade.php" if used Authenticationmethod is 'mysql' or 'http_auth' (which means local Authentication)

So i would insert in *.blade.php an Authenticationtypefilter for (not) showing "enabled"

SourceDoctor: @murrant 
query is fixed ...

SourceDoctor:
@murrant 
enable/disable Feature now only visible in mysql Authorization mode

murrant:
The checkbox is not checked on all users.
Also, you should not be able to disable yourself (for safety reasons)

murrant:
Oh, the reason it wasn't checked for me is I didn't migrate my database when testing...

murrant:
Many places use the toggles, so in a way not having toggles also breaks consistency.  :)  LibreNMS is not the most consistent app in the world due to it's community supported nature.

SourceDoctor:
@laf @murrant 
checkboxes are now bootstraped :-)

SourceDoctor:
@laf 

> Also, would also be handy to enable that toggle switch in the user listing page so you can simply enable / disable users.

i didn't get it to run, don't know what happens on user list. But i didn't get a simple checkbox inside the table.
If you've got a tip i'll give it a last try before merging as is :-)

murrant:
I noticed the ui is only shown for mysql, but you said supports http auth too...

SourceDoctor:
@murrant 
> enable/disable Feature now only visible in mysql Authorization mode
;-)

Support for MySQL Authorization was my primary intention.
I only said somewhere upwards that the http auth class inherits from mysql auth class. :-)

for me it's ready for merging :-)

murrant:

PR #10519, opened 2019-08-12
Tests for IOS Vlans/Vtp

PipoCanaja:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
@murrant I think I understood why the JSON was not added ... the vlans code in ios uses a <community>@<vlanid> built community to access STP data for each VLAN ... That cannot make it in snmpsim, all of those request timeout, and that's exactly why it breaks now in Travis. 
May be with a lighter snmprec (less vlans) it could be possible. I'll try to get down to 2 vlans only for the testfile.

murrant:
Running it locally, it works but it is soooooooo sloooooow...  it doesn't complete within the time window.

PR #10521, opened 2019-08-14
Fix creating device groups issue

murrant:
Not being able to write the device relationships file should not be fatal

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/DB/Schema.php
+++ b/LibreNMS/DB/Schema.php
@@ -171,7 +171,7 @@ class Schema
             ];
 
             if (is_writable($cache_file)) {
-                @file_put_contents($cache_file, serialize($cache));
+                file_put_contents($cache_file, serialize($cache));
             } else {
                 \Log::error("Could not write cache file ($cache_file)!");
             }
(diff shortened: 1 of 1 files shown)


LibreNMS/DB/Schema.php
```diff
@@ -170,7 +170,11 @@ public function getAllRelationshipPaths($base = 'devices')
                 $base => $paths
             ];
 
-            file_put_contents($cache_file, serialize($cache));
+            if (is_writable($cache_file)) {
+                @file_put_contents($cache_file, serialize($cache));
```
Jellyfrog: This (the `@`) seems like a bad practice and hard to debug, it's trivial enough to fix a permission error for a sysadmin

murrant: I agree, that is why I added the is_writable check so it logs a message.  The bad thing that happens is the relationship cache is generated every time.  Which actually isn't that bad.  So instead of hundreds of confused users, I thought it made sense to just ignore not being able to write it in the webui.

@ may not be necessary, but I wasn't sure if is_writable follows selinux.  Since the downside isn't that bad, I thought it was ok.

Jellyfrog: Still I would remove the `@`

murrant:
Oops, I thought this was in 1.55

murrant:

PR #10530, opened 2019-08-16
Sort Devices and Groups in Alert Rules 'map to' droplist

SourceDoctor:
sort Devices and Groups alphabetic in
Alert Rule -> "Map To"

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 10530 `

### DIFF
--- a/includes/html/list/devices.inc.php
+++ b/includes/html/list/devices.inc.php
@@ -58,7 +58,7 @@ if (!empty($_REQUEST['limit'])) {
     $offset = 0;
 }
 
-$sql = "SELECT `device_id`, `hostname`, `sysName` FROM `devices` $query order by `hostname`";
+$sql = "SELECT `device_id`, `hostname`, `sysName` FROM `devices` $query";
 $devices = array_map(function ($device) {
     return [
         'id' => $device['device_id'],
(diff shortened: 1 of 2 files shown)


PipoCanaja:
Works well for the groups. Not so OK for devices, because they depend on force_ip_to_sysname option. 
You need to sort in PHP (and not in the SQL part) the $devices array, in which the text was already processed by format_hostname according to user preferences.

SourceDoctor:
@PipoCanaja 
sorting is moved

includes/html/list/devices.inc.php
```diff
@@ -68,4 +68,6 @@
 
 $more = ($offset + count($devices)) < $total;
 
+array_multisort(array_column($devices, 'text'), SORT_ASC, $devices);
```
laf: `text` isn't a valid column name in the `devices` table.

Also, you need to check if `Config::get('force_ip_to_sysname')` is true and use `sysName` for sorting if not, `hostname` should be used.

SourceDoctor: @laf 
"format_hostname()" returns the hostname string, and checks also 'force_ip_to_sysname'

this function will be used in an array_mapping function which returns an array of associative arrays, containing the keys 'id' and 'text'. 'text' contains the return value of "format_hostname()"
Take a look on lin 61-67

so on my views there is nothing to change
(i tested the code and it works well)

PipoCanaja: Good for me, tested all right

murrant:

PR #10532, opened 2019-08-17
hostname mouseover fix - only show location name instead of its array

SourceDoctor:
on mouseover on a device, now location name will be displayed instead the location associative array


Jellyfrog:
Will be fixed by https://github.com/librenms/librenms/pull/10526

PR #10539, opened 2019-08-20
Avoid state collisions between 'dell' and 'drac' OSes

PipoCanaja:
https://community.librenms.org/t/bug-idrac-sensor-discovery/8470

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 10539 `

### DIFF
--- a/includes/discovery/sensors/state/dell.inc.php
+++ b/includes/discovery/sensors/state/dell.inc.php
@@ -28,8 +28,7 @@ foreach ($tables as $tablevalue) {
 
     if (is_array($temp)) {
         //Create State Index
-        $state_oid = $tablevalue[2];
-        $state_name = "dell." . $state_oid;
+        $state_name = "dell." . $tablevalue[2];
         if ($state_name == 'dell.processorDeviceStatusStatus' || $state_name == 'dell.memoryDeviceStatus' || $state_name == 'dell.powerSupplyStatus' || $state_name == 'dell.intrusionStatus') {
             $states = [
                 ['value' => 1, 'generic' => 3, 'graph' => 0, 'descr' => 'other'],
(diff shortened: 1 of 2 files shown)


murrant:
LGTM, but will lose historical data.

PipoCanaja:
> LGTM, but will lose historical data.

Should not lose any historical data here. The RRD is not touched, only the state mapping, which does not impact any data.

laf:
> Should not lose any historical data here. The RRD is not touched, only the state mapping, which does not impact any data.

Are you sure? The sensor_type field from the DB is what is used to create the filename and it changes with this.

PipoCanaja:
For me, the RRD file name is created out of the discover_sensor parameters only. And discover_parameter does not have the new modified name. Only the table with the state mapping knows about the new name of the state. 
BUT: 
Seems that it does not create the match properly if both RRD name and state name are not equal :D  so I suspect it does not work. 

I'll do some tests via snmpsim to findout, cause I don't have a DRAC :)

PipoCanaja:
@murrant @laf I need to choose which of the 2 OSes will have state history broken ... Any clue except random :) ?

murrant:
I feel like DRAC is slightly more popular, we could check the stats...

PipoCanaja:
OK. This time, the '$type' part of the discover data is modified, so supposingly the historical data is lost. 
That being said, I don't have any test data for this ...

murrant:
I'm fine with merging without test data for this, but probably needs notification.

PipoCanaja:
Yep, it does. More than test data, I need testers ... I would like to be sure of the behaviour before merging ... IE we need this code to be working, and loose historical data cause RRD is not named the same :)

PipoCanaja:
Hi @murrant 
Been a little bit overloaded lately. In order to close that one, the notification would be : 
" Due to a duplicate name in different Dell MIBs, devices recognised as "dell" os in LibreNMS will loose historical data for the following OIDs : processorDeviceStatusStatus, memoryDeviceStatus, powerSupplyStatus, intrusionStatus, controllerState, arrayDiskState, virtualDiskState, batteryState. ". 
If ok, I'll issue the notification so we can merge this.
Bye

murrant:
Looks ok, maybe mention that it is possible to rename rrds or something.

laf:
@PipoCanaja Any news on the notification so we can get this out?

PipoCanaja:
Hi @laf 
PR for notification created. I'll merge it during the WE unless somebody asks not. Feel free to merge it right away if it LGTY. 
The actual PR #10539 would then be merged Sunday Nov 3rd. 
Bye

mikecentola:
Should we add the latest Dell MIB for OMSA while doing this as well, or do you think it's unrelated?

PipoCanaja:
Hi @mikecentola 
Let's create a dedicated PR for a new device support. This one is "just" targetting an issue between objets in 2 mibs that share the same name, and collision. 
Thanx

murrant:

PR #10548, opened 2019-08-22
Populate ipv6_network_id in ipv6_addresses where empty

garysteers:
This will populate the ipv6_network_id field in ipv6_addresses for any entries that were created before librenms#10544) commit a1c8fe6

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


garysteers:
Hey @lef

I hadn’t used [ ] as the rest of the file was using array so matched what I thought was the style? Is that not what the coding style guide dictates?

Gary

laf:
> I hadn’t used [ ] as the rest of the file was using array so matched what I thought was the style? Is that not what the coding style guide dictates?

We have no need to support older versions of php so can use `[]`. If the coding styles say something different then they need updating - feel free to update :)

PR #10565, opened 2019-08-28
Catch exceptions generated by alert transports

lfkeitel:
The alert.php script would completely die if a single alert
transport generated an exception. For example the API transport
if CURL couldn't connect to the api endpoint. The script
should attempt to send alerts to all working transports and not
die because a single transport fails.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/RunAlerts.php
+++ b/LibreNMS/Alert/RunAlerts.php
@@ -1,7 +1,7 @@
 <?php
 /*
  * RunAlerts.php
- * 
+ *
  * Copyright (C) 2014 Daniel Preussker <f0o@devilcode.org>
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -25,7 +25,7 @@
  *
  * Modified by:
  * @author Heath Barnhart <barnhart@kanren.net>
- * 
+ *
  */
 
 namespace LibreNMS\Alert;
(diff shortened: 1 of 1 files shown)


LibreNMS/Alert/RunAlerts.php
```diff
@@ -524,8 +524,12 @@ public function extTransports($obj)
                 $obj['msg']       = $type->getBody($obj);
                 c_echo(" :: $transport_title => ");
                 $instance = new $class($item['transport_id']);
```
murrant: Maybe include the instance creation too?

murrant:
Didn't know any transports would throw exceptions, nice change.

LibreNMS/Alert/RunAlerts.php
lfkeitel: I can do that if you want me to. I'm not sure if any existing transports have the possible of throwing on creation and API is only one I know of that could throw on delivery.

murrant: I don't think any really define __construct() right now.

lfkeitel: I can include just in case? I think the reasoning could apply to both. If a particular transport throws, for whatever reason, continue to try any other transports.

lfkeitel: And theoretically the DB could throw right? If for some reason the database was unavailable momentarily. Although in that case every transport would fail.

lfkeitel:
I moved class instantiation into the try block so both creation and delivery exceptions are caught. Let me know if there's anything left I need/should do.

murrant:

PR #10567, opened 2019-08-29
Fix testing fping output when the LibreNMS user doesn't have a valid …

BrianSidebotham:
…shell

Normally, the librenms user will not have a default shell (We don't want this user logging in to the system, etc.). But this breaks the `fping` tests because the lack of shell means we get no output from the tests. This simple pull requests sets a shell for the librenms user if the script is running as `root`.

For example:

```
[root@localhost ~]# cat /etc/passwd | grep librenms
librenms:x:991:986::/opt/librenms:/bin/false

[root@localhost ~]# su librenms -c "fping6 ::1"

[root@localhost ~]# su librenms -s /bin/sh -c "fping6 ::1"
::1 is alive
```

This results in `validate.php` running correctly for root as well as the librenms user.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Validator.php
+++ b/LibreNMS/Validator.php
@@ -251,7 +251,7 @@ class Validator
     public function execAsUser($command, &$output = null, &$code = null)
     {
         if (self::getUsername() === 'root') {
-            $command = 'su ' . Config::get('user') . ' -c "' . $command . '"';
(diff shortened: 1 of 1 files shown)


murrant:

PR #10582, opened 2019-09-05
Add support Racom Ray3

erotel:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


tests/snmpsim/ray.snmprec
```diff
@@ -3,8 +3,32 @@
 1.3.6.1.2.1.1.2|6|1.3.6.1.4.1.33555.1.1
```
murrant: Create a new snmprec file (the -v option), you've merged data for two different types of devices...

includes/discovery/sensors/temperature/ray.inc.php
```diff
@@ -10,13 +10,26 @@
  * the source code distribution for details.
  */
 
-d_echo('RAY');
+d_echo('RAY-OLD');
```
murrant: Probably labeling that as OLD is silly, just leave it as "RAY"

includes/discovery/sensors/state/ray.inc.php
```diff
@@ -2,11 +2,11 @@
...
 
-// System Status (Value : na (0) unknown, ok (1) ok, warning (2) warning, alarm (3) alarm)
-$state = snmp_get($device, "systemStatus.0", "-Ovqe", 'RAY-MIB');
-if ($state) {
+// System Status Ray1 and Ray2 (Value : na (0) unknown, ok (1) ok, warning (2) warning, alarm (3) alarm)
+$state = snmp_get($device, "systemStatus", "-Ovqe", 'RAY-MIB');
+if (is_numeric($state)) {
     //Create State Index
-    $state_name = 'systemStatus';
+    $state_name = 'systemStatus-old';
```
murrant: Do not rename existing state sensors.

Also, instead of doing two gets you could just do:
```
$state = snmp_getnext($device, "systemStatus", "-Ovqe", 'RAY-MIB');
```

erotel: well and how do I make sure `'.1.3.6.1.4.1.33555.1.1.3.1'` or ` '.1.3.6.1.4.1.33555.1.1.3.1.0'` is used in discover_sensor (), depending on which device is detected?

murrant: if you change the snmp output options, you can get the oid back too.

erotel:
obviously, this device will be a problem

erotel:
44/5000
Can anyone look at it if it's ok?

includes/polling/os/ray.inc.php
```diff
@@ -1,7 +1,40 @@
...
+
+//Ray1 and Ray2 has no index
+if (empty($hardware)) {
+    $ray_tmp = snmp_get_multi_oid($device, 'productName swVer serialNumber unitType', '-OQs', 'RAY-MIB');
+    $hardware      = $ray_tmp['productName'];
+    $version       = $ray_tmp['swVer'];
+    $serial        = $ray_tmp['serialNumber'];
+    $features      = $ray_tmp['unitType'];
+
+    $snmpdata = snmp_get_multi_oid($device, ['sysName', 'sysObjectID', 'sysDescr'], '-OUQn', 'SNMPv2-MIB');
```
murrant: This data is already fetched and in the $device array.

LibreNMS/OS/Ray.php
```diff
@@ -117,9 +136,10 @@ public function discoverWirelessRssi()
      */
     public function discoverWirelessSnr()
     {
-        $oid = '.1.3.6.1.4.1.33555.1.3.2.2'; // RAY-MIB::snr.0
         return array(
-            new WirelessSensor('snr', $this->getDeviceId(), $oid, 'racom', 1, 'CINR', null, 1, 10),
+            new WirelessSensor('snr', $this->getDeviceId(), '.1.3.6.1.4.1.33555.1.3.2.2.0', 'racom', 1, 'CINR', null, 1, 10),
+            new WirelessSensor('snr', $this->getDeviceId(), '.1.3.6.1.4.1.33555.1.3.2.2', 'racom-old', 1, 'CINR', null, 1, 10),
```
murrant: renaming the existing sensor types will cause data loss (it changes the rrd name)

murrant:
I feel like you should check something once, then change all the OIDs to either have .0 or not instead of creating new sensors (and renaming old ones)

LibreNMS/OS/Ray.php
erotel: I'll fix this

includes/polling/os/ray.inc.php
erotel: I don't understand this. If I don't use this command, where do I get the data?

PipoCanaja: @erotel This data is already polled and stored in $device. Just look at $device['sysName'], $device['sysObjectID'], $device['sysDescr'].

erotel: Not the data is not there because Ray 1 and Ray 2 have bad snmp (do not support index)

Ray 2 
`snmpwalk -c public -v1 10.133.41.27 .1.3.6.1.2.1.1.5.0
iso.3.6.1.2.1.1.5 = STRING: "Zes-Dol30-24GHz"`

Ray 3
`snmpwalk -c public -v1 10.133.51.251 .1.3.6.1.2.1.1.5.0
iso.3.6.1.2.1.1.5.0 = STRING: "husak-achp_RAy3-24L"`

murrant: Hmmmmmmmmmm, not sure the best way to fix, but I don't think this is it.  I hate bad SNMP implementations.

erotel: I pointed out the poor implementation of snmp directly to the developer, unfortunately there was no fix. Couldn't you do that in the discovery section of the os folder? If this was a major problem, it would not be implemented.

erotel: So I tested it. I propose to preserve os ray for Ray1, Ray2 and create os ray3 for Ray3. I would separate them with a file in `/includes/discovery/os` ,as ubnt has, then Ray3 could use a yaml file.

PR #10585, opened 2019-09-06
laravel key generate documentation

SourceDoctor:
add missing laravel key generation command to documentation


murrant:
The key is generated by `./scripts/composer_wrapper.php install`, which you kinda have to run to install the PHP dependencies.

SourceDoctor:
oh ok, i oversees this.
So this Pull Request seems to be obsolet

PR #10588, opened 2019-09-06
Added AIX file system, prefer over hrstorage for AIX

dlesel:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hi @dlesel 
The goal is to collect new data (new variant) with all the new OIDs that you may be polling compared to existing test-data.
And you also need to compute (./scripts/save-test-data.php) existing test data to regenerate json file, in order to identify changes between old behaviour and new one. This step is mandatory for TravisCI test suite to succeed. 
As you can see, currently, TravisCI is complaining.
Bye

dlesel:
Hi,
There no new OIDs.
Because hrstorage has issue with reporting storage greater than 8 TB, IBM advised me to use IBM-AIX-MIB.mib. librenms already has this mib with all OIDs.
i already run (./scripts/save-test-data.php) . Chages are in 2 files: tests/data/aix.json  and tests/snmpsim/aix.snmprec. I have to add emty string manuallty to submit tests/data/aix_net-snmp.json, tests/data/aix_std.json, tests/snmpsim/aix_net-snmp.snmprec and tests/snmpsim/aix_std.snmprec. Theya are identical to what you already have.
TravisCI complains because version and serial number of AIX/IBM are different. They will be different beacuse i submitted /opt/librenms/includes/polling/os/aix.inc.php. librenms AIX detection wrongly identifies IBM serial number and machine type. I corrected it and added AIX version detection with SP (AIX Service Pack).
I think TravisCI shoud be configure to ignore version and serial number in this case.
What i submitted already runs on our production librenms server for couple of weeks. It monitors AIX 6.1, 7.1, 7.2, ReadHat, Solaris.
Thanks,
david.

PipoCanaja:
Hi @dlesel 
The idea is to run save-test-data on all files (snmprec) where the compute is impacted by your PR. That is exactly what Travis does, so if you do it, then travis will do it as well and find the same result. That's exactly what you (and we) want. 
Then all the "impacted" data will be in this PR, for us to review.
Bye

dlesel:
i already did it. i completed all steps from https://docs.librenms.org/Developing/os/Test-Units/
All test data already collected and submitted.

If i run ./scripts/save-test-data.php and collect new test data then i can not submit test data because they are identical to already existing test data on github.com.

This is what i did today:

-bash-4.2$ ./scripts/save-test-data.php -o=aix
Starting snmpsim listening on 127.1.6.1:1161...
OS: aix
Module: all

Added device: 10
127.1.6.1 10 aix

..........................

Polled in 2.013 seconds
Saved to /opt/librenms/tests/data/aix.json
Ready for testing!


[root@librenms-dev librenms]# git status
# On branch AIX-file-system
# Your branch is ahead of 'origin/master' by 4 commits.
#   (use "git push" to publish your local commits)
#
nothing to commit, working directory clean
[root@librenms-dev librenms]# git push origin AIX-file-system
Everything up-to-date
[root@librenms-dev librenms]# git add tests/data/aix.json
[root@librenms-dev librenms]# git status
# On branch AIX-file-system
# Your branch is ahead of 'origin/master' by 4 commits.
#   (use "git push" to publish your local commits)
#
nothing to commit, working directory clean


I never used github.com before and may  do mistakes. You need to tell me what files do you need.

Thanks,
david.

includes/polling/os/aix.inc.php
```diff
@@ -0,0 +1,5 @@
+<?php
+$serial = snmp_get($device, 'aixSeSerialNumber.0', '-Oqv', 'IBM-AIX-MIB');
+$hardware = snmp_get($device, 'aixSeMachineType.0', '-Oqv', 'IBM-AIX-MIB');
+$aix_version = preg_split('/\s+/', $device['sysDescr']);
```
murrant: Your version code isn't working for the previous data that worked using the code in unix.inc.php

dlesel: unix.inc.php inc wrongly identifies serial and hardware. includes/polling/os/aix.inc.php corrects isuues.
i can correct unix.inc.php in separate pull request. 
Let's get rid of includes/polling/os/aix.inc.php for now. Storage is more important.

murrant: Sounds great.  I am always in favor of small focus'd PRs :)

murrant:
@dlesel I see what is going on here.  Your PR is removing hrStorage support, but that is required when AIX is using net-snmp.

dlesel:
if i allow hrSorage in AIX then it will be duplicate entries in DB table "storage"- one for hrStorage and another for aixFileSystem. Then end user must manually exclude hrStorage in config.php - $config['os']['aix']['ignore_mount_regexp'][] = "/^\/dev\//";
i do not like manual configuration in this case.

dlesel:
we do not need hrStorage with AIX. it reports wrong data after 8 TB.

murrant:
That doesn't mean you should break it for others.  You'll need to come up with another option than removing it.  Adding an OS option to disable hrstorage in your config.php per OS would be a good way.

dlesel:
Librenms is not usable to monitor AIX because of 2 major bugs in  hrstorage.inc.php. One bug shows logical volumes instead of mount points (/dev/hd4 instead of /, /dev/hd2 instead of /var and so on). Second bug shows negative number if storage is bigger than 8 TB. I do not think AIX admins using librenms for production monitoring. There are another bugs too - serial number and machine type are wrong, lmsensors not supported in AIX, mempools have the same problem as hrstorage. AIX does things differently than Linux including snmtp. We need switch librenms's AIX staff to IBM-AIX-MIB. Logically i did the right think - disabling hrstorage.inc.php by coding and adding aix.inc.php. 
Adding an OS option to disable hrstorage in config.php is not reasonable because ones you disable hrstorage all storage info will be gone from database. In any case you break.

murrant:
@dlesel I understand what you are saying and appreciate it.  But you can remove the built in snmp daemon and install net-snmp with working hrStorage.

The correct fix is not to have a switch that disables one.   The correct fix is check aixFsTableEntry first, then if that is not available, use hrStorage.

murrant:
I don't think you understand what I'm saying.  If you remove the existing snmpd daemon and install net-snmp, it will function like Linux, which clearly some admins have chosen to do otherwise the support would not have been added to LibreNMS.

dlesel:
Could you point the code where similar method has been applied. Then i better understand what needs to be done.

includes/discovery/storage/hrstorage.inc.php
```diff
@@ -43,6 +43,10 @@
             continue;
         }
 
+        if ($device['os'] == 'aix') {
```
murrant: Just change this to:

```php
if ($device['os'] == 'aix' && !empty($aix_filesystem)) {
```

murrant: Also, a comment would help, "Skip hrStorage if aixFsTable is available".

dlesel: Thank you!

murrant:

PR #10590, opened 2019-09-07
geographical map fix for "normal user"

SourceDoctor:
users in normal mode got a webui crash on geographical map because of undefined variables in sql query. Maybe a historical artifact.


murrant:

PR #10591, opened 2019-09-07
Eltek eNexus sensor improvements

vitalisator:
- additional sensors
- some cleanup in yaml
- add Current, Runtime, Voltage to device over

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hello @vitalisator 
Thanx for your PR. Could you let us know why you removed the file  tests/data/enexus_smartpacks2.json ? The goal of the test data is to test regressions and analyse changes between versions. In this case, we need to identify which (if any) sensors will change, loose historical data, etc etc). In this case, I would suggest to update it instead of deleting it. 
Thanx

vitalisator:
@PipoCanaja, I had just problems with pre-commit tests while both files were present.

PipoCanaja:
Hi @vitalisator 
The idea of test files is to check regressions. Or changes that impacts deeply the discovery OR poller. In this case, as you said, pre-commit did warn you that a change occured. Which is expected in your case, cause you are precisely changing discovery and polling. 
You need to update all the existing test-files (so pre-commit does not complain) and we get all the changed test data in the PR (so we have a chance to see if a change in this PR may break historical data, or anything like this). 
So could you please re-add the file, and update it so we can review this PR.
Thanx

murrant:
@vitalisator you are adding new sensors, but you also have removed sensors from other devices.  You need to have both in the yaml.  Only ones that are present will be discovered.

vitalisator:
> @vitalisator you are adding new sensors, but you also have removed sensors from other devices. You need to have both in the yaml. Only ones that are present will be discovered.

if you mean 
oid: load  and oid: battery
that are replaced by more granular oids: loadFuseTable and oid: batteryBankTable
that should not regress other devices.

vitalisator:
after some rework, I think we don't need to inform about changing sensors.

murrant:

PR #10596, opened 2019-09-10
add a IPMI tab to IPMI enabled devices to display assorted useful info...

VVelox:
This adds a IPMI tab to IPMI enabled devices with 5  possible pages.

- [Overview](https://i.imgur.com/O5P3FvW.png) :: This is the default landing page, showing power status, basic sensor info, and last 20 SEL entries.

- [Chassis](https://i.imgur.com/IthXzzw.png) :: Chassis status, power on hours, last restart cause, and power restore policy.

- [LAN](https://i.imgur.com/n9MTNLd.png) :: Shows the networking config and assorted information for the IPMI.

- [SEL](https://i.imgur.com/qYA1flA.png) :: Shows the entire system even log.

- [Sensors](https://i.imgur.com/A2A2RCz.png) :: Displays the sensor info with all threshold info as well.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


Jellyfrog:
Not sure how other pages does, or how to solve it nicer (cache?) but it will execute commands directly to IPMI on each request as I understand it, feels a bit shaky/slow, What happens if there is an error or timeout?

Jellyfrog:
Also, I count 8 places where you do `system()` and `exec()`  
I would suggest creating a class for ipmi under LibreNMS\Util maybe, look at Snmpsim.php for inspiration for example

murrant:
Collecting data in the webui won't work on many setups.

VVelox:
@Jellyfrog I like that idea. Will be nice for when later adding support for configuring IPMI later on.

@murrant Not really sure about that. Some most definitely, but lots of stuff in the web UI is while it is less dangerous as it is historical data, it is still general the sort of thing you would want to restrict access to from stuff not on the management subnets.

That is good point as to where collectors and web more sepperated though. Implementing additional polling for this all could be handy at some point. In which breaking out the IPMI into its own class/module/whatever PHP calls it is even a better idea.

murrant:
Honestly, the best way would be event based polling where the pollers connect to the master and it tells them when to poll what.  So you could collect data through the pollers (and it wouldn't have to be scheduled and stored).

VVelox:
@murrant Sorry, been crazy lately so a bit slow to reply.

I am a bit lost, do we actually support that as of currently? Unless I am missing something, I though this was just ran by cron in every instance. Or did something new get added and I've missed it?

murrant:
Oh no, sorry, talking about pie in the sky.  Sometimes I think a little too far forward...

SourceDoctor:
@VVelox something new here?

VVelox:
@SourceDoctor  Sorry, I really need to get back to this all. Been distracted with other projects taking up time as of late. Need to write a bunch of stuff for polling and like for this. Need to find the time to sit down and figure out the laravel and eloquent stuff.

murrant:
FYI, this will need to be ported to this after it is merged:
https://github.com/librenms/librenms/pull/11328

Jellyfrog:
Closing due to inactivity, Feel free to reopen

PR #10595, opened 2019-09-09
Markdown linting of docs

Jellyfrog:
Preparation for adding markdown linter to travis, see #10597

Before
```
    354  MD013 Line length
    172  MD033 Inline HTML
     66  MD024 Multiple headers with the same content
     44  MD009 Trailing spaces
     24  MD029 Ordered list item prefix
     24  MD002 First header should be a top level header
     17  MD031 Fenced code blocks should be surrounded by blank lines
     17  MD022 Headers should be surrounded by blank lines
     16  MD005 Inconsistent indentation for list items at the same level
     16  MD001 Header levels should only increment by one level at a time
     15  MD036 Emphasis used instead of a header
     13  MD026 Trailing punctuation in header
      9  MD032 Lists should be surrounded by blank lines
      8  MD034 Bare URL used
      7  MD046 Code block style
      7  MD012 Multiple consecutive blank lines
      7  MD010 Hard tabs
      4  MD004 Unordered list style
      2  MD006 Consider starting bulleted lists at the beginning of the line
      2 
      1  MD007 Unordered list indentation
      1  MD003 Header style
```

After:
``` 
    357  MD013 Line length
    146  MD033 Inline HTML
     66  MD024 Multiple headers with the same content
     22  MD002 First header should be a top level header
     15  MD036 Emphasis used instead of a header
     14  MD001 Header levels should only increment by one level at a time
     10  MD026 Trailing punctuation in header
      1  MD034 Bare URL used
```

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #10597, opened 2019-09-10
Add markdownlint to travis

Jellyfrog:
After merging #10595 the status of linting would look like this without any rules (ignoring all changelogs):
```
    357  MD013 Line length
    146  MD033 Inline HTML
     66  MD024 Multiple headers with the same content
     22  MD002 First header should be a top level header
     15  MD036 Emphasis used instead of a header
     14  MD001 Header levels should only increment by one level at a time
     10  MD026 Trailing punctuation in header
      1  MD034 Bare URL used
```

Running it with the rules in this PR:
```
      1  MD034 Bare URL used
``` 

Should we ignore less/more rules?
I could fix all cases of `MD001 Header levels should only increment by one level at a time` for example, but not sure if we want that.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
It seems like we would want pretty much all of those.

murrant:
@Jellyfrog For some reason, I feel like this should be failing instead of passing.  You sure it is running correctly? :D

Jellyfrog:
> @Jellyfrog For some reason, I feel like this should be failing instead of passing. You sure it is running correctly? :D

It only lints md-files you have changed in the PR, so zero for this one :)  
But this will be a bit to aggressive atm, since it doesn't support ignoring "single offenders". Have to think about this one for a bit.

PR #10607, opened 2019-09-15
Storage RRD - descriptionlength fix - percent column fix

SourceDoctor:
increase title description from 12 to 16 characters
also change columntitle vom "%age" to "Ratio" for percental usage column

### DIFF
--- a/includes/html/graphs/device/storage.inc.php
+++ b/includes/html/graphs/device/storage.inc.php
@@ -6,7 +6,7 @@ $scale_max = '100';
 require 'includes/html/graphs/common.inc.php';
 
 $iter         = '1';
-$rrd_options .= " COMMENT:'                        Size      Used    Ratio\\l'";
+$rrd_options .= " COMMENT:'                        Size      Used    % Used\\l'";
 
 foreach (dbFetchRows('SELECT * FROM storage where device_id = ?', array($device['device_id'])) as $storage) {
     // FIXME generic colour function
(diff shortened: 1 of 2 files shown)


includes/html/graphs/device/storage.inc.php
```diff
@@ -27,7 +27,7 @@
         $iter   = '0';
     }
 
-    $descr        = rrdtool_escape($storage['storage_descr'], 12);
+    $descr        = rrdtool_escape($storage['storage_descr'], 16);
```
Jellyfrog: Other graphs also seems to have 12 chars, like: https://github.com/librenms/librenms/blob/master/includes/html/graphs/storage/usage.inc.php#L19  
If we change one, we should change all?

includes/html/graphs/device/storage.inc.php
```diff
@@ -6,7 +6,7 @@
 require 'includes/html/graphs/common.inc.php';
 
 $iter         = '1';
-$rrd_options .= " COMMENT:'                    Size      Used    %age\\l'";
+$rrd_options .= " COMMENT:'                        Size      Used    Ratio\\l'";
```
Jellyfrog: Shouldn't it be "% Used", like the other graphs?

SourceDoctor: @Jellyfrog i changed it to "% Used"

includes/html/graphs/device/storage.inc.php
SourceDoctor: @Jellyfrog 
i agree.
is also changed now.

murrant:

PR #10618, opened 2019-09-20
add discovery of mempool percent warning limit

vitalisator:
- Remove hardcoded mempool_perc_warn (75%) from the database
- rise a default value up to 90% (maybe need to discus that)
- add possibility to individual warn limits in mempools discover module
- set Linux and vmware limits to: Physical mem: 99%, Virtual mem: 95%, Swap mem: 10%

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Not sure setting 75% on all devices is a good idea.  All Linux servers will essentially report 100% all the time.

vitalisator:
> Not sure setting 75% on all devices is a good idea. All Linux servers will essentially report 100% all the time.

That is because we have already this value per default in the database table

PipoCanaja:
LGTM but not tested

tomarch:
There's an issue with this PR, the custom %warm is always reset to the default.
Step to reproduce :
-change the %warm for a device (test with windows device).
-wait for the discovery
-the %warm come back to the default.

murrant:

PR #10634, opened 2019-09-27
Added support for drac power and current polling

theister-xan:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Any reason you didn't use yaml for these?

theister-xan:
Is there any way to apply logic to the yaml mechanism?

For the drac data for current and power is in the same tree ( amperageProbeReading ) with current being e.g. .1 and .2 while power is .3

The only way to check which sensor type it should be is to query another tree (amperageProbeType ) and check the type per index.

So if the yaml mechanism can do if then logic we could use it here, I could not find any hint how to do that and asked in discord. After not getting an answer I implemented it this way.

murrant:
Yeah, this is actually pretty easy to do.
amperageProbeType tells you what type of probe so, you just check that in the skip_values

So, for current:

```yaml
skip_values:
                        -
                            oid: amperageProbeType
                            op: '!='
                            value: 23
```

murrant:

PR #10637, opened 2019-09-28
Add Cisco QFP module

pobradovic08:
**Requires #10638 in order for graph authentication to work.**

Building on top of #10622, this module provides additional statistics about Quantum Flow Processor provided by `CISCO-ENTITY-QFP-MIB`. This includes:

- General info about QFP (status, number of boots, last boot time, traffic direction handling)
- Utilization
- Memory
- Packets per second
- Bits per second
- Calculated average packet size
- Calculated relative utilization per 1kpps processed

I placed it alongside the Processors page in Health because it doesn't really justify it's own section.
Module is enabled for Cisco IOS-XE and looks like this:

image

Tested on ASR1001-HX and ASR1001-X. 

#### Please note

#### Testers


pobradovic08:
Fixed missing poller statistics. I was unseting `$module` variable at the end of the poller file.

includes/html/pages/device/health/qfp.inc.php
```diff
@@ -0,0 +1,214 @@
...
+        $util_label = 'label-success';
+    } elseif ($tmp_component['utilization'] < 75) {
+        $util_label = 'label-warning';
+    } else {
+        $util_label = 'label-danger';
+    }
+
+    $graph_array = $default_graph_array;
+    $graph_array['type'] = 'qfp_util';
+    $text_descr = 'QFP Utilizatoin';
```
PipoCanaja: Typo :)

includes/html/pages/device/health/qfp.inc.php
```diff
@@ -0,0 +1,214 @@
...
+    echo "<div class='panel-body'>";
+    include 'includes/html/print-graphrow.inc.php';
+    echo "</div></div>";
+
+    /*
+     * Relative QFP utilization to packets processed
+     */
+    $graph_array = $default_graph_array;
+    $graph_array['type'] = 'qfp_relativeutil';
+    $text_descr = 'QFP Relative utilizatoin per kpps';
```
PipoCanaja: Typo :)

pobradovic08: Fixed, thanks!

PipoCanaja:
Thanx :) will review this asap :)

pobradovic08:
Thanks for the review. If you don't have physical devices, CSR1000V has (a virtual) QFP that should do just fine.

murrant:

dsgagi:
@pobradovic08 
Seems like this PR started spamming device event log with QFP component value updates on each poll. See the screenshot attached, Component 781 is the QFP on our ASR1006.

issue-screenshot

Just noticed this 30 mins ago. Didn't find the problem in code on first look. Will try again tomorrow. Maybe someone else will be faster :)

PipoCanaja:
@dsgagi All components are showing this behaviour, this is not at all specific to this PR. So may be a feature request could be to have a way to disable this event generation for some components for instance.

dsgagi:
Hi again. Busy end of the week, so I just found some time to look this up again.

@PipoCanaja @pobradovic08 
After reviewing the `component_prefs` table, I see the reasoning for event log entry generation. Sorry for jumping the gun. :)

Still, logging is maybe too excessive for attributes in the picture. At the moment (after 45 days from release cycle update), we have 40000+ event log entries per device with QFP module. The attributes are actually metrics, so it's normal for them to change on every poll cycle.

Possible solution would be to introduce a new column in `component_prefs` table to distinguish attribute types (status, metric, etc.) and not log changes to metric types (and maybe some other types).

PR #10642, opened 2019-09-30
Update snmp-scan.py to take notice of the discovery_by_ip setting

BrianSidebotham:
When using the `snmp-scan.py` script it doesn't take notice of the LibreNMS configuration setting `discovery_by_ip`. This was a surprise and lots of IP addresses were added to LibreNMS during a scan which we didn't want.

We don't want devices within LibreNMS that don't have reverse DNS entries.

This simple fix makes `snmp-scan.py` honour the `discovery_by_ip` setting which I think most people reading the documentation would expect to happen.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/snmp-scan.py
+++ b/snmp-scan.py
@@ -121,6 +121,18 @@ def scan_host(ip):
         except (herror, gaierror):
             pass
 
+        # If there is no hostname and the configuration option to add by IP is
+        # not set, we should not add this to LibreNMS
+        if hostname is None and not CONFIG.get('discovery_by_ip'):
+            try:
+                arguments = ['/usr/sbin/fping', ip]
+                check_output(arguments)
+            except CalledProcessError as err:
+                output = err.output.decode().rstrip()
+                return Result(ip, hostname, Outcome.UNPINGABLE, output)
+
+            return Result(ip, hostname, Outcome.FAILED, 'no dns entry found and discovery_by_ip is not set in config.php!')
+
         try:
             arguments = ['/usr/bin/env', 'php', 'addhost.php', hostname or ip]
             if args.ping:


snmp-scan.py
```diff
@@ -121,6 +121,18 @@ def scan_host(ip):
         except (herror, gaierror):
             pass
 
+        # If there is no hostname and the configuration option to add by IP is
+        # not set, we should not add this to LibreNMS
+        if hostname is None and not CONFIG.get('discovery_by_ip'):
+            try:
```
murrant: Why are you trying to ping it?
Also, you can't really hardcode the path of fping, it is not the same on many OS.

murrant:
I guess I didn't consider snmp_scan.py as "discovery".   FYI, some people (against advisement) run this in cron to add new devices, so this could be an unexpected change for them and take them a long time to track down.

snmp-scan.py
BrianSidebotham: Thanks for reviewing @murrant. Sure, no problem. `fping` can be abstracted. If you have `discovery_by_ip` set to `False` so you don't want to discover by IP the only output we can get back is `FAILED` - the ping test at least gets us the `UNPINGABLE` result if appropriate too.

BrianSidebotham:
@murrant `snmp_scan.py` is the only thing that can discover some of the devices we require in some environments. Not all devices will respond to `lldp`, etc.

It was a surprise and took a bit of digging from me to see that `snmp-scan.py` ignores `discovery_by_ip`. We should probably update the docs, not matter what we do - I can add that to the PR when we're clear of what we want.

I understand your concern about changing default behaviour. Perhaps it would be better to add a new argument to enable this new feature and change the docs to suit?

murrant:
Sure, if you are going to add a switch, you can just ignore the discovery_by_ip setting all-together.

snmp-scan.py
murrant: Yes, but it will check things below, so I don't see why you would need to ping it here anyway...

murrant: Maybe the result should be "SKIPPED" no need to ping...

BrianSidebotham: `UNPINGABLE` provides more information than `EXCLUDED`, but as we're here talking about functionality where we don't want to `addhost` when there's no reverse DNS in place then it is reasonable to simply return `EXCLUDED` - I'll make that change.

BrianSidebotham:
There's a difference between taking notice of the `discovery_by_ip` configuration setting and a flag that does forces not to discover by ip. There are already parts of the discovery configuration that `snmp-scan.py` is using and is documented in the docs such as mentioned in the `auto_disovery` docs: https://docs.librenms.org/Extensions/Auto-Discovery/ where this tool is mentioned.

It still makes sense for me for this tool to use this already-present configuration setting. Either way, I'll update the documentation to suit so it's known what the behaviour will be.

Do you want to just let me know which option you'd prefer @murrant ? Sounds like you'd prefer to just have this work via a new switch to the `snmp-scan.py` tool, and for that switch to make `snmp-scan.py` not add ip addresses as hosts when they do not have reverse DNS.

Does that sound right? I can modify the PR to suit.

murrant:
There is a difference one says enable this and it'll check another setting, the other says enable this and it is enabled...

I feel like the second is correct.  The setting is supposed to be "xDP autodiscovery discover by ip" not really a generic one which is why I feel it is wrong to use it in snmp-scan.py.  Because the name is terse, you have made an incorrect assumption about what the setting does.  I hope to correct this in this by having a little more context around the setting.

BrianSidebotham:
The title of this pull request is exactly the opposite of the change that we required. We want *only* those IP addresses that have a reverse DNS entry to be added. Currently if reverse DNS fails the IP address is added as a host and that is *not* what is intended, which is why we have the `discovery_by_ip` set to `false`.

This is what the original pull request did. Sorry I have not got time to get around to changing the pull request to use a command line argument instead.

Today our LibreNMS is totally screwed through auto update. LDAP is broken and after fixing that the UI is now broken too (Dashboard links do not work, etc.)

murrant:
FYI.  Most those issues are already fixed, just run ./daily.sh although no one has reported an ldap issue yet, so please let me know via discord or the community site.

murrant:
So what you actually want is "no IP fallback" aka only add devices that have valid reverse DNS.

murrant:
Sorry for all the miss-communication.  Just kind of reinforces my thoughts that `discovery_by_ip` should be no where near snmp-scan.py and only used for auto-discovery.  Maybe the other settings should go too.

BrianSidebotham:
Hi @murrant ,

That's right. Sorry I don't see the confusion with `discovery_by_ip` - but there may be a different meaning to it than I can see on the surface.

Anyway, apologies I couldn't get round to fixing this before you, we've been tied up with various issues recently. Thanks for the new option - that looks exactly what we need.

Yes, everything else is back to normal on the test build after a `daily.sh`. I'll close this PR now, thanks.

PR #10644, opened 2019-10-01
Add CISCO-IF-EXTENSION port stats for IOS/IOS-XE

pobradovic08:
Adds support for `CISCO-IF-EXTENSION` MIB that provides additional information about interface errors, similar to the info we can see with `show int x` command. Useful to detect runts/giants/overruns.

This is for `ios` and `ios-xe` devices. Graphs are shown in Port page alongside other port specific graphs.

I also moved the MIB to `mibs/cisco/`.

image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


sorano:
Nice! 👍

pobradovic08:
Ok, test failing doesn't seem to be related to changes in this PR, there just wasn't any test data collected for `ports` module before. I just collected test data on the master branch and it fails as well on the same things.

PipoCanaja:
Hi @pobradovic08 
Thanx for this PR. 
Concerning the test data, you need to create a new variant (both snmprec and JSON), and leave the existing snmprec file as it is.
Then, look at the results of the tests.
If not OK, then probably the existing iosxe.json will complain, and you then need to rerun save_test_data on the original iosxe.snmprec to recalculate it. But, as you noticed, the original iosxe.snmprec file does not include the extended data, so it will probably not trigger your changes.
That should make Travis happy, and allow us to merge this.

pobradovic08:
HI @PipoCanaja 
Thanks for the response. Unfortunately I don't think creating a new variant will solve the issue with tests. This PR builds on top of existing poller for `ports` module so I would have to save data for the whole module and than that new variant will fail validation. I'll commit the test data files as a separate variant.

Since current JSON doesn't have any data for ports module (and this PR doesn't update DB) i used current master branch (without changes in this PR) and ran `collect-snmp-data.php` for `ports` module and `save-test-data.php`. Tests failed.

So either i'm collecting test data wrong or the test for that module were broken already but we didn't know because there were no data for IOS-XE.

PipoCanaja:
Hi @pobradovic08
As the module does not change any data in the DB, then you don't need to add any test data at all, cause tests are done on the DB. 
I would suggest to start by reverting the snmprec and json files. Should allow Travis to be happy again, and we will review and merge this PR. 
Bye

pobradovic08:
Hi @PipoCanaja,

Thanks for the clarification, I didn't know that snmp data was used only for verification of DB data. Removed it.

murrant:

laf:
This has had quite a dramatic (in some cases) impact on poller performance, as would anything polling this many additional OIDs.

The images below will show anything from a sub 1 second improvement in poller times (device only took ~6 seconds anyway) to some 40+ seconds for devices with a higher latency. On large setups this has an overall significant impact. I'm going to do a PR to at least have a config option to enable/disable this and I'd really like it to be disabled by default.

image

image

image

PR #10648, opened 2019-10-03
Implement Oxidized "config search" as an API

theochita:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -73,7 +73,7 @@ Route::group(['prefix' => 'v0', 'namespace' => '\App\Api\Controllers'], function
         Route::put('rules', 'LegacyApiController@add_edit_rule')->name('edit_rule');
         Route::delete('rules/{id}', 'LegacyApiController@delete_rule')->name('delete_rule');
         Route::post('services/{hostname}', 'LegacyApiController@add_service_for_host')->name('add_service_for_host');
-        Route::get('oxidized/configsearch/{searchstring}', 'LegacyApiController@search_oxidized')->name('search_oxidized');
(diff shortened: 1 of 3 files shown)


routes/api.php
```diff
@@ -73,6 +73,7 @@
         Route::put('rules', 'LegacyApiController@add_edit_rule')->name('edit_rule');
         Route::delete('rules/{id}', 'LegacyApiController@delete_rule')->name('delete_rule');
         Route::post('services/{hostname}', 'LegacyApiController@add_service_for_host')->name('add_service_for_host');
+        Route::get('oxidized/configsearch/{searchstring}', 'LegacyApiController@search_oxidized')->name('search_oxidized');
```
Jellyfrog: Any reason why you added this under the admin middleware?  
As far as I know you don't need admin in the web interface for oxidized

theochita: Thanks for replying :)

The reason behind it was that getting to search through config files is something i would imagine only admins should have access to.

I can move it anywhere you want it to be :)

murrant: You do need to be an admin.  Many want it able to be separated, but that isn't an option yet.

laf: Maybe change this to `/config/search/{searchstring}`? We can then use `/config` to return the devices config from the api at a later stage.

theochita: Just updated it to follow your example.

theochita: Thanks for updating the doc @laf  stupid of me to miss that :(

murrant:

PR #10651, opened 2019-10-03
Store config data serialized

murrant:
This way we can store null, booleans, and more reliably.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/Models/Config.php
+++ b/app/Models/Config.php
@@ -45,33 +45,7 @@ class Config extends BaseModel
         'config_group' => '',
         'config_sub_group' => '',
     ];
-
-    /**
-     * Get the config_value (type cast)
-     *
-     * @param string $value
-     * @return mixed
-     */
-    public function getConfigValueAttribute($value)
-    {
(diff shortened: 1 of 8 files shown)


app/Models/Config.php
```diff
@@ -67,10 +72,6 @@ public function getConfigValueAttribute($value)
 
     public function setConfigValueAttribute($value)
     {
-        if (is_bool($value)) {
-            $this->attributes['config_value'] = $value ? 'true' : 'false';
-        } else {
-            $this->attributes['config_value'] = $value;
-        }
+        $this->attributes['config_value'] = serialize($value);
```
Jellyfrog: Any reason we can't use
```php
protected $casts = [
        'config_value' => 'array',
    ];
```
Or maybe I misunderstood how it works

murrant: That would work, but we want to support old values when we decode from the database.  Otherwise, we could just add a migration to convert all existing db values...

murrant:

PR #10652, opened 2019-10-04
api: add hostname search for list_devices function

sk4mi:
api: add domain type query for list_devices

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/html/api_functions.inc.php
```diff
@@ -300,6 +300,8 @@ function list_devices(\Illuminate\Http\Request $request)
         $sql = "`d`.`ignore`='0' AND `d`.`disabled`='0'";
     } elseif ($type == 'location') {
         $sql = "`locations`.`location` LIKE '%".$query."%'";
+    } elseif ($type == 'domain') {
+        $sql = "`d`.`hostname` LIKE '%".$query."'";
```
Jellyfrog: This is just a confusing way to search for domain, since you can still match hostnames.
You could do something like; `substring_index(hostname, '.', -2) = $query` to get the last 2 parts of a FQDN;  
```sql
> SELECT substring_index('librenms.foo.bar', '.', -2);
+----------------------------------------------+
| substring_index('librenms.foo.bar', '.', -2) |
+----------------------------------------------+
| foo.bar                                      |
+----------------------------------------------+
```
But it will fail with single TLD:
```sql
> SELECT substring_index('librenms.local', '.', -2);
+--------------------------------------------+
| substring_index('librenms.local', '.', -2) |
+--------------------------------------------+
| librenms.local                             |
+--------------------------------------------+
```

laf: I think searching for domain is the wrong title, it should just be search for hostname like `%hostname%`

sk4mi: I don't need just last 2 parts of fqdn.

Per example, this device list:
- device1.siteA.company.com
- device2.siteA.company.com
- device1.siteB.company.com
- device2.siteB.company.com

With given list, for example i just need the first two devices with a search.

```bash
curl -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/devices?type=domain&query=siteA.company.com
```

I´m agree with the fact that `domain` search can be confusing. But in work we really need this feature for listing device.

Can we discuss for a new name ? or the implementation is wrong for you ?

murrant: Just call it hostname search.  You can just put in `siteB.company.com` and it will find but those sites.  No need to make it specific to your use case.

sk4mi: Fixed query name to hostname, and match for pattern in hostname.

Thanks.

murrant:

PR #10658, opened 2019-10-07
Panos - ActiveSession Fix and SSLProxy Sensor

shepherdjay:
Primary contribution is adding a new SSLProxyUtilization, while I was doing this I noticed that the active session sensor on our install hasn't worked for over a year. Specifying the OID an additional level as well as the max level by OID seemed to resolve it for me. Not sure if there is a better way to fix it. I don't quite understand the index value as these values are not out of a table that I can tell but it is now reporting correctly.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/discovery/panos.yaml
+++ b/includes/definitions/discovery/panos.yaml
@@ -38,3 +38,10 @@ modules:
                     group: Sessions
                     descr: Active Sessions
                     high_limit: panSessionMax
+        load:
+            data:
+                -
+                    oid: panSessionSslProxyUtilization
+                    num_oid: '.1.3.6.1.4.1.25461.2.1.2.3.8.{{ $index }}'
+                    descr: SSL Proxy Utilization
+                    group: Sessions
--- a/includes/definitions/discovery/panos.yaml
+++ b/includes/definitions/discovery/panos.yaml
@@ -32,12 +32,11 @@ modules:
         count:
             data:
                 -
-                    oid: panSession
-                    value: panSessionActive
-                    num_oid: '.1.3.6.1.4.1.25461.2.1.2.3.{{ $index }}'
+                    oid: panSessionActive
+                    num_oid: '.1.3.6.1.4.1.25461.2.1.2.3.3.{{ $index }}'
                     group: Sessions
                     descr: Active Sessions
-                    high_limit: panSessionMax
+                    high_limit: '.1.3.6.1.4.1.25461.2.1.2.3.2.{{ $index }}'
         load:
             data:
                 -


murrant:
Can you add test data so we keep it from getting broken in the future?

includes/definitions/discovery/panos.yaml
```diff
@@ -32,9 +32,15 @@ modules:
...
-                    oid: panSession
-                    value: panSessionActive
-                    num_oid: '.1.3.6.1.4.1.25461.2.1.2.3.{{ $index }}'
+                    oid: panSessionActive
+                    num_oid: '.1.3.6.1.4.1.25461.2.1.2.3.3.{{ $index }}'
                     group: Sessions
                     descr: Active Sessions
-                    high_limit: panSessionMax
+                    high_limit: '.1.3.6.1.4.1.25461.2.1.2.3.2.{{ $index }}'
+        load:
```
murrant: This should be a count sensor.

Load is for electrical AMPs.

shepherdjay: charge, humidity, and load were the only sensors listed here https://docs.librenms.org/Developing/os/Health-Information/ that supported receiving the value as a percentage. Count appeared to expect an integer or count of how many sessions there were. 

Does a new sensor type need to be added instead?

murrant: According to the description, this is a count of the number of sessions, or am I missing something.

shepherdjay: Maybe we are talking about a different line? Description is "SSL Proxy Utilization" and is given as a percentage. The other health monitoring called "Active Sessions" is indeed a count and is under the count type.

murrant: Load is ok, for that since it is a percentage.  I assumed incorrectly it was a count :)

murrant:
You still need to add test data for this PR.  https://docs.librenms.org/Developing/os/Test-Units

shepherdjay:
Yep I'm working on getting this test data over for this, making sure it is cleared of sensitive data.

shepherdjay:
> You still need to add test data for this PR. https://docs.librenms.org/Developing/os/Test-Units

I've ran into an issue generating the test data. I have snmpsim installed but I get this when running. 

```
bash-4.2$  /opt/librenms/scripts/save-test-data.php --os panos
Starting snmpsim listening on 127.1.6.1:1161...
OS: panos
Module: all

panos: SQLSTATE[HY000] [1049] Unknown database 'librenms_phpunit_78hunjuybybh' (SQL: select * from `ipv4_addresses` where `ipv4_address` = 127.1.6.1 limit 1)
```

I'm working on spinning up another librenms instance incase something is unique here but if you know a quick fix that would be great, databases aren't my strong suit.

laf:
@shepherdjay Try running `./scripts/save-test-data.php` from within `/opt/librenms`

shepherdjay:
> @shepherdjay Try running `./scripts/save-test-data.php` from within `/opt/librenms`

No change in behavior similar error, so I re-ran the collect script followed immediately by the save and now I got a python traceback with a permission denied as the last entry. Running outside of the librenms just dumps a bunch of "ERROR at line xxxxx, value x, nextFlag" to the screen.

shepherdjay:
I've been unable to get the necessary tests - as such I will close this pull request until I can.

PR #10674, opened 2019-10-10
removed the ksort block as it was not working

dGs-:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

#### Info about the PR

Running ./discovery.php -h device_d -m ntp -d, I had some warnings : 
`Warning: Illegal offset type in /opt/librenms/LibreNMS/Component.php on line 163 & 166`

As suggested by murrant :

> $RESULT[$component_device_id] is an array, but the code tries to use that as an index...
> just delete that block of code...
> I'm guessing this $RESULT[$RESULT[ is supposed to just be this $RESULT[, but that means that the ksorts have never done anything, so they can be removed.


murrant:

PR #10683, opened 2019-10-15
Menu option for deleting orphaned RRD Files

SourceDoctor:
allows Admins to remove orphaned RRD Files.

Why could this be useful?
just an example: Smart Monitoring.
You configured it to use Serial Number as indicator.
You replace one or more harddiscs.
All RRD Files for the replaced harddiscs stay and will never again get new data, so they are orphaned. 
But everytime you take a look on a SMART Graph you will the this harddiscs, but without any data.
With this feature you can remove this RRD Files to have again a clean SMART Monitoring with values for only the harddiscs you are interested in.
The same for many other RRD Files...

By default a RRD File of at least 24hours without modification will be listed as possible orphaned.

Where in menu it is positioned?
Setting -> orphaned RRD Cleanup

### DIFF
--- a/includes/defaults.inc.php
+++ b/includes/defaults.inc.php
@@ -905,6 +905,10 @@ $config['dateformat']['mysql']['compact'] = '%Y-%m-%d %H:%i:%s';
 $config['dateformat']['mysql']['date']    = '%Y-%m-%d';
 $config['dateformat']['mysql']['time']    = '%H:%i:%s';
 
+// time difference to last RRD File modification in seconds to detect file as orphaned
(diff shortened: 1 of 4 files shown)


murrant:
This already exists :) https://docs.librenms.org/Support/Cleanup-options/

SourceDoctor:
@murrant 
dooh, and i thought i wrote such a nice new Feature ;-)

so it's obsolete.
Thx for link to documentation

murrant:
No worries, thanks for the PR :)

PR #10704, opened 2019-10-17
Support new os SNR

hrtrd:
mib file

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #10718, opened 2019-10-18
update snmp.inc.php

hrtrd:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


hrtrd:
#10704

includes/snmp.inc.php
```diff
@@ -267,7 +267,7 @@ function snmp_get($device, $oid, $options = null, $mib = null, $mibdir = null)
     $time_start = microtime(true);
 
     if (strstr($oid, ' ')) {
-        echo report_this_text("snmp_get called for multiple OIDs: $oid");
+        echo snmp_get called for multiple OIDs: $oid";
```
murrant: You should put an exit here too.  If you hit this, it is a development bug.

murrant: After reflecting, this should just be `throw new Exception("snmp_get called for multiple OIDs: $oid");`

PR #10600, opened 2019-09-12
Fix missing to-parameter in dashboard graph widget link

feuerrot:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

#### Description
If you add a generic-graph widget to a dashboard the image links to a bigger version of the graph (under `https://librenms/graphs/`). This link doesn't include a `/to=${timestamp}/` parameter in the URL.
Updating the time range in the graph view uses `submitCustomRange()` from [librenms.js](https://github.com/librenms/librenms/blob/master/html/js/librenms.js#L138) to replace the `from` and `to` parameters in the URL using regular expressions. If there is no `to` parameter in the URL, it won't be added - so unless you click on one of the predefined time ranges first or add it yourself you can't change the `to` parameter.

It might be a better idea to rewrite the `submitCustomRange()` function, but as I don't know enough about javascript, I just added the `to` parameter in the generic-graph widget code.

PR #10720, opened 2019-10-18
Fix missing to-parameter in graphRows

feuerrot:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

#### Description
This PR fixes the same problem as #10600 when graphRows() is used to generate multiple graphs of the same type (e.g. https://librenms.example.org/device/device=41/tab=graphs/).

### DIFF
--- a/LibreNMS/Util/Html.php
+++ b/LibreNMS/Util/Html.php
@@ -27,6 +27,7 @@ namespace LibreNMS\Util;
 
 use HTMLPurifier;
 use HTMLPurifier_Config;
+use Carbon\Carbon;
 use LibreNMS\Config;
 
 class Html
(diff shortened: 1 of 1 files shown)


murrant:
This smells like it should be fixed at a lower level.

feuerrot:
`graphRow()` calls `Url::generate()` to generate the link and I suppose `URL::generate()` wouldn't be the right place.

murrant:
Indeed, but to is not require in the url...  the default to on a graph is now as far a I know.

murrant:
https://github.com/librenms/librenms/blob/master/includes/html/graphs/graph.inc.php#L26

murrant:
Closing due to no response.  Feel free to respond and we can re-open.

PR #10739, opened 2019-10-23
Fix memcached always reporting "ERROR" for app_state

willhseitz:
perform json_encode() on memcached app $response data to match update_application() string parameter

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Nothing better to put in that field?  Just a json dump of the full data set (that is already stored in the db) does not seem that interesting.

willhseitz:
On first glance, it seemed the only reason the $response is passed back to update_application() is only to check is there is a PHP warning/error. I figured something like json_encode() would capture all the same data, as it is being passed as an array currently

murrant:
The intent of that field (I think) is just a simple ok/error state for the most part.

willhseitz:
Here is the only code block from update_application() that interacts with the $response parameter:

```
    if ($response != '' && $response !== false) {
        if (str_contains($response, array(
            'Traceback (most recent call last):',
        ))) {
            $data['app_state'] = 'ERROR';
        } else {
            $data['app_state'] = 'OK';
        }
    }
```

I think converting the entire array to string via json_encode() would keep this functionality. Do you think there is a smaller subset of the array that should be JSON encoded and passed instead?

murrant:
What is the contents of $data?
put `dd($data);` right before update_application()   (without your changes)

willhseitz:
here is the $data output: https://p.libren.ms/view/09682872
Looks to be the collected data from the memcached application module. The only difference between this and other application pollers is that $data is still an array when passed to update_application(), which expects a string

murrant:
Yeah you could do something like:
```php
$status = empty($data['uptime']) ? 'ERROR' : 'OK';
```

Then send that to the update_application() call

willhseitz:
I looked over over application modules again, and noticed that most return the raw snmp_get or snmp_walk data as the $response for update_application(). I've updated memcached.inc.php to be consistent with the other application pollers

willhseitz:
This is ready for review

danislav:
Is this how we decide if an application is in good health? - the app_state?
I thought to use the app_state to create alerts but am not convinced that this is the best.

willhseitz:
> Is this how we decide if an application is in good health? - the app_state?
> I thought to use the app_state to create alerts but am not convinced that this is the best.

This PR is more intended to make the memcached check consistent with other application checks, not to change the intention of the checks themselves.

SourceDoctor:
@murrant @laf 
User Pending still actual, or ready for merge?

willhseitz:
Ready for merge please

murrant:

PR #10687, opened 2019-10-15
Fix maps display only one link if multiple links are present between devices.

louis-oui:
DO NOT DELETE THIS TEXT

https://github.com/librenms/librenms/issues/10686

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #10755, opened 2019-10-28
Revert "Fix maps display only one link if multiple links are present between devices."

murrant:
Reverts librenms/librenms#10687


murrant:
Reverting this because it causes issues for some people.  I think the issues is around ARP maps.

The correct fix need to reduce ARP links to 1, but leave more intelligent discovery data intact.

PR #10756, opened 2019-10-28
Adding more default values to config_definitions

SourceDoctor:
add missing default properties to config_definitions.json

### DIFF
--- a/misc/config_definitions.json
+++ b/misc/config_definitions.json
@@ -594,8 +594,10 @@
             ]
         },
         "base_url": {
-            "type": "text",
-            "default": ""
+            "group": "system",
+            "section": "server",
+            "order": 5,
+            "type": "text"
         },
         "billing.base": {
             "default": 1000,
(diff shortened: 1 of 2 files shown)


misc/config_definitions.json
```diff
@@ -593,6 +593,10 @@
                 "mpls"
             ]
         },
+        "base_url": {
+            "type": "text",
+            "default": ""
```
murrant: This one should not actually have a default as the default is set programatically.

murrant:
Just one change.  Looks good, but just as a note this is not enough to add them to the webui, just allow them to be set via `lnms config:set`

misc/config_definitions.json
SourceDoctor: removed default value

SourceDoctor:
@murrant 
defined position in WebUI now :-)

SourceDoctor:
@murrant 
something else to do for integration?

murrant:

PR #10757, opened 2019-10-29
Convert overview page to laravel

Jellyfrog:
As a bonus we now return 404 on invalid pages.

Todo:
- [x] Delete old layouts
- [ ] Convert more layouts?

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


Jellyfrog:
Code is not the prettiest, I mostly converted 1:1 from old code

app/Http/Controllers/LegacyController.php
```diff
@@ -49,11 +49,10 @@ public function index(Request $request, Session $session)
         $vars['page'] = basename($vars['page'] ?? '');
         if ($vars['page'] && is_file("includes/html/pages/" . $vars['page'] . ".inc.php")) {
             require "includes/html/pages/" . $vars['page'] . ".inc.php";
-        } elseif (Config::has('front_page') && is_file('includes/html/' . Config::get('front_page'))) {
```
laf: I think we should continue to support custom front pages if I'm honest. Whilst the other ones we provide might not work, people might have created their own. However we should have them place the custom files in a location outside of git.

Jellyfrog:
> Works fine for me but a couple of things.
> 
>     1. Is it possible to allow people to setup custom front pages still without affecting daily updates?
> 
>     2. If the above is possible, can you document how?

Not currently no.  
We can create a folder that we gitignore where it will look for custom pages first, before looking for a method.  
Suggestions for a path?

app/Http/Controllers/OverviewController.php
```diff
@@ -0,0 +1,111 @@
...
+                Toastr::error(
+                    "Dashboard <code>#$request->dashboard</code> does not exist! Loaded <code>
+                    ".htmlentities($dashboard->dashboard_name)."</code> instead.",
+                    "Requested Dashboard Not Found!"
+                );
+            }
+        }
+
+        if (!isset($dashboard)) {
+            $new_dash = new Dashboard();
```
murrant: You could do this:
```php
$dashboard = Dashboard::create([
    'dashboard_name' => 'Default',
    'user_id' => $user->user_id,
]);
```

app/Http/Controllers/OverviewController.php
```diff
@@ -0,0 +1,111 @@
...
+
+            $new_dash->dashboard_name = 'Default';
+            $new_dash->user_id = $user->user_id;
+            $new_dash->save();
+
+            $dashboard = $new_dash;
+
+            // TODO: Is this still needed?
+            UserWidget::where('user_id', $user->user_id)
```
murrant: I think this code was to cleanup some db inconsistency.

app/Http/Controllers/OverviewController.php
```diff
@@ -0,0 +1,111 @@
...
+              ->where('dashboard_id', 0)
+              ->update(['dashboard_id' => $new_dash->dashboard_id]);
+        }
+
+        $data = $dashboard
```
murrant: $data = $dashboard->widgets;  will give you all the widgets.  Using it as a function starts a query.

app/Http/Controllers/OverviewController.php
```diff
@@ -0,0 +1,111 @@
...
+                            'size_x'=>6,
+                            'size_y'=>2,
+                            'refresh'=>60
+                        );
+        }
+
+        $bare        = $request->bare;
+        $data        = serialize(json_encode($data));
+        $dash_config = unserialize(stripslashes($data));
+        $widgets     = Widget::select('widget_id', 'widget_title')->orderBy('widget_title')->get();
```
murrant: You already fetched the widgets.

resources/views/overview/default.blade.php
```diff
@@ -0,0 +1,603 @@
...
+@endif
+<span class="message" id="message"></span>
+<div class="gridster grid">
+    <ul></ul>
+</div>
+@endsection
+@section('javascript')
+<script type="text/javascript" src="js/jquery.gridster.min.js"></script>
+@endsection
+@section('scripts')
```
murrant: This should be @push('scripts')

murrant:
People do use some of the other layouts, though not many.  Maybe a informal survey to see if it is worthwile to port any?

Are you planning on a fairly straight forward port or trying to do some improvement too?  Wish list items: drop "widgets" table, use better js grid library

Jellyfrog:
> People do use some of the other layouts, though not many. Maybe a informal survey to see if it is worthwile to port any?

What about adding a deprecated warning on all layouts except for "tiles"?  
If people care enough they should revolt then :)

> Are you planning on a fairly straight forward port or trying to do some improvement too? Wish list items: drop "widgets" table, use better js grid library (or maybe even just vue.js)

"Code is not the prettiest, I mostly converted 1:1 from old code"

Not planning on fixing it more than this for now.  
I think we can "lure" more people to contribute if it's at least "laravel code"

app/Http/Controllers/OverviewController.php
Jellyfrog: Yeah but we need to make a query, `widgets()` is actually `userWidget`, and we need `widget` column from `Widgets` table.

murrant:
Just splitting out the blade template does make things a lot cleaner.
Yes I think a deprecated warning on those frontpages would be a good idea.  Two options then: leave those pages for now (with warning) and merge this soon.  merge warning now, merge this after next release.

Jellyfrog:
Let's go for merging warning now. I'll prepare a PR.

laf:
> Let's go for merging warning now. I'll prepare a PR.

Looks like the old front pages have been deleted, I thought they were being left with a warning?

Jellyfrog:
> > Let's go for merging warning now. I'll prepare a PR.
> 
> Looks like the old front pages have been deleted, I thought they were being left with a warning?

See #10757

murrant:
@Jellyfrog you could have this code support the old pages for now and we could merge this sooner than later.  Then just remove the old pages after 1.59.

Jellyfrog:
> @Jellyfrog you could have this code support the old pages for now and we could merge this sooner than later. Then just remove the old pages after 1.59.

Then I would have to revert removing object cache, so it can wait instead

murrant:

https://community.librenms.org/t/deprecation-of-legacy-overview-pages/10072/1

arjitc:
It would pretty good to have at least "default.inc.php" converted out.

This is a really good/clean layout for anyone looking to quick-start with LibreNMS, the widget supporting (other "default" layout) is pretty good for advanced users/users who are looking to play around with the layout or users looking for something "extra", while the "default.inc.php" gives a nice and easy quick-start for users looking for alertlogs, eventlogs, ports and other alerts of devices.

Hoping that this can be included in as a alternative option to the default dashboard.

murrant:
What about a non-empty default dashboard @arjitc

arjitc:
> What about a non-empty default dashboard @arjitc

That would be a good option indeed, something which mimics the same elements as "default.inc.php" perhaps? 

This would give the end-user the flexibility of moving things around as well as give a good starting point/reference to users using LibreNMS for the first time/setup-and-go.

Jellyfrog:
I will give it a shot to convert `default.php` also

murrant:
@Jellyfrog I don't think we need default precisely, just replicate it's layout in the default instead of having the empty default...

Jellyfrog:
> @Jellyfrog I don't think we need default precisely, just replicate its layout in the default instead of having the empty default...

I'm not sure I fully understand what you mean here.

PipoCanaja:
> > @Jellyfrog I don't think we need default precisely, just replicate its layout in the default instead of having the empty default...
> I'm not sure I fully understand what you mean here.

I think the idea is to replicate default in the "default" overview page of a new user.

Jellyfrog:
@librenms/Reviewers Please have a look, especially around the new front pages.

PipoCanaja:
Hi @Jellyfrog 
Not clear what I need to test. 404 works well :) And Overview for a new or existing user comes as empty as it used to be. So it works the same for overview. 
What else should be tested ?

app/Http/Controllers/OverviewController.php
```diff
@@ -0,0 +1,178 @@
...
+
+        return view('overview.default', compact('bare', 'dash_config', 'dashboard', 'user_dashboards', 'shared_dashboards', 'widgets'));
+    }
+
+    public function simple(Request $request)
+    {
+        //TODO: All below missing D.ignore = '0' check
+        $devices_down = [];
```
murrant:
```suggestion
```

This is reassigned later

app/Http/Controllers/OverviewController.php
```diff
@@ -0,0 +1,178 @@
...
+        $ports_down = [];
+        $services_down = [];
```
murrant:
```suggestion
```

Also reassigned.

app/Http/Controllers/OverviewController.php
murrant: Go ahead and delete this query.

resources/views/overview/default.blade.php
```diff
@@ -0,0 +1,606 @@
...
+    }
+
+    function widget_dom(data) {
+        dom = '<li id="'+data.user_widget_id+'" data-type="'+data.widget+'" data-settings="0">'+
+              '<header class="widget_header"><span id="widget_title_'+data.user_widget_id+'">'+data.title+
+              '</span>'+
+              '<span class="fade-edit pull-right">'+
+
+                @if (
+                        ($dashboard->access == 1 && Auth::id() === $dashboard->user_id) ||
```
murrant: Could set this in the controller, no big deal though.

Jellyfrog: I'll leave it for now, the whole thing could be written better anyway.  
Mostly wanted to get rid of some old sql queries actually :)

Jellyfrog:
> I think we should merge this next month, Just a few non-critical changes. Any changes happen upstream that need to be merged in?

Need to look into this.

Jellyfrog:
Should be ready, but wouldn't mind if this runs in nightly a while in case I deleted one to many files :)

murrant:
That's why we wait for next month :)

murrant:

PR #10759, opened 2019-10-29
Migrate Python scripts to Python 3

deajan:
Hello,

Since Python 2.x EOL is pretty close, I updated the python 2 only scripts.
List of changes:

- Python 2 to 3 migration; keeping Python 2.7 compatibility
- Changed subprocess caller to a way more robust version
- Separated procedural code from functions
- A lot of PEP-8 cleanup
- Refactored config & db code into a library
- Removed default db argument from some functions in order to prevent corner cases
- Re-tested all code
- Merge threading code into a library

This is an initial PR in order to make LibreNMS fully Python 3 compatible.
There are a lot of things left to do in order to tidy up these codes among:

- Harden the code (undeclared variable handling, better exceptions...)
- Remove that DB class which does lack abstraction layer

Willing to take a bit of time of needed.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


deajan:
Of course, codeclimate had some fixes but created new issues since code needs to be refactored.

murrant:
Thanks for working on this @deajan   I have not reviewed it yet, but one thing to keep in mind is we still support CentOS 7 and probably have some installs on CentOS 6.  If backwards compatibility is not a part of this, it will require a large effort to notify users that polling will be affected.

deajan:
Hmm... Backward compatibility should not be so hard to obtain here.
I'll have this tested on CentOS 7/Python 2.7 too then.

librenms_library.py
```diff
@@ -0,0 +1,247 @@
+#! /usr/bin/env python
```
murrant: Can you place this in a subdirectory?  There are already some scripts in the LibreNMS directory...

deajan: Sure, I was actually working on if I could merge with the existing LibreNMS/__init__.py code (at least for the database code which is pretty much the same), but there are way too much other imports in that file already.
Once I've sorted out what I can merge, I'll move the file :)

murrant: Great.

SourceDoctor:
@deajan 
something new here?

murrant:

https://community.librenms.org/t/dropping-support-for-php-7-1-and-python-2/11851/4

deajan:
Sorry, I've let this slip away while improving my python skills.
I'll see if I can catch up and finish this PR soon.

SourceDoctor:
@deajan 
on problems feel free to ask on https://community.librenms.org/
or on chat system https://discord.com/invite/V8s5JZJ 
(there exists an extra development room for questions)

murrant:
@deajan I rebased this so github-apply will work.

- Removed some python2 support stuff, it won't be needed now.
- Found some duplciated code in services-wrapper.py
- Moved the library into the LibreNMS package

deajan:
@murrant Any reason you deleted poller-service.py ?

deajan:
I'll try to finish my PR this week, there are still two outstanding issues:
- lacks usage of DB class instead of that hacky query behavior
- code hardening (variables may be used before declaration

I'll update my test librenms instance this afternoon in order to play with that PR.

murrant:
@deajan yes, it is broken and deprecated.  I don't want it ported to python3 ;)

murrant:
@deajan I'm going to roll back your changes.  I know there is a lot of code duplication, but we need to get this merged soon and your last changes broke things.

Please start another PR with further refinements and probably start using some object oriented approach to prevent passing so many variables around.

Come contact me on Discord to discuss if you like :)

murrant:
@deajan by the way you can restore your commits to a new branch like this:
`git checkout df97af998692a4cf0c999253a8a089c86a79cfdf -b python3-love-continued`

deajan:
@murrant I tried to refactor a minimum to only avoid code duplication and remove some obvious caveats.

The way I see this, all three wrapper scripts should be merged into a single one, that indeed uses a class for distributed polling handling.
The downside would be that it would change the scripting bahavior, unless we add dummy scripts that would just call the main script with the right arguments.
What do you think about that idea ?

Another idea would be to replace threading with multiprocessing, so we can leverage all cpu cores power, but this would make inter process communication more diffucult.

murrant:
I'm down for whatever you propose, but I just needed a python3 version in master before the next release ;)

murrant:
Isn't multiprocessing the older way of doing it?  starting a process for each worker instead of a lightweight thread?

SourceDoctor:
@murrant multiprocessing and threading are existing parallel and having both advantages and disadvantages. Threading work "parallel" but work stays on the same core. Multiprocessing opens a seperate "process" so alle CPU Cores can be real used.
Personally i use multiprocessing in most cases only if workers have to do cpu intensive stuff to do, otherwise i us threading, mostly ... for short :D

murrant:
Honestly, there is no need to have the wrapper run on different cores is there?  All the work is done in a php process which can run on any core.

crazy-max:
Looks like [some scripts](https://github.com/librenms/librenms/search?l=python&p=1) have been missed? `snmp-scan.py` for example?

murrant:
@crazy-max snmp-scan.py already supports python3 ;)

murrant:

PR #10787, opened 2019-11-06
On editing device, only show assigned alert rules

SourceDoctor:
now only Alert Rules will be listed which are assigned to the edited Device
In advance Alert Rules which are assigned to Device Groups, the Device is a Member of,
are listed now also

### DIFF
--- a/includes/html/print-alert-rules.php
+++ b/includes/html/print-alert-rules.php
@@ -101,36 +101,39 @@ foreach ($result_options as $option) {
 
 echo '</select></td>';
 
-$query_device = 'SELECT alert_device_map.rule_id AS id FROM alert_device_map';
-$where_device = '';
-
-$query_device_group = 'SELECT alert_group_map.rule_id AS id FROM device_group_device LEFT JOIN alert_group_map ON device_group_device.device_group_id=alert_group_map.group_id';
-$where_device_group = '';
-
+$param = [];
 if (isset($device['device_id']) && $device['device_id'] > 0) {
-    $where_device .= 'WHERE alert_device_map.device_id=?';
-    $where_device_group .= 'WHERE device_group_device.device_id=?';
+    //device selected
+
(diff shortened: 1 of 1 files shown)


includes/html/print-alert-rules.php
```diff
@@ -101,26 +101,45 @@
 
 echo '</select></td>';
 
-$query = 'FROM alert_rules';
-$where = '';
-$param = [];
+$query_device = 'select alert_device_map.rule_id as id from alert_device_map';
```
laf: Can you UPPERCASE the query so it looks like it's a MySQL query please. I.e `SELECT alert_device_map.rule_id AS id FROM alert_device_map`

SourceDoctor: @laf 
done

SourceDoctor:
@laf 
something else to do?

PipoCanaja:
@SourceDoctor , @laf , @kkrumm1 
For the Alert Rule tab of the device properties, I would expect to see all rules that are applicable to this device, which are : 
- global rules
- rules applicable to this device
- rules applicable to any device groups this device is in

For the moment in the PR, the implementation only shows "rules applicable to this device or to the device groups the device is in"

What do you think ?

SourceDoctor:
@PipoCanaja why should there be all available rules visible?
That makes no sense to me.
If i take a look into Alert Rules on a specific Device, i expect to see only the rules which are assigned to this device. (directly or via device group matching)
So you can directly see for what you are alerted on the device you are on.

For seeing all available rules there exists Alert->Alert Rules.

SourceDoctor:
@PipoCanaja 
you are absolute right "global rules" are missing.

I'll patch add it in. :-) thx for pointing to it.

SourceDoctor:
@PipoCanaja 
fixed :-)

SourceDoctor:
@PipoCanaja 
everything else to fix?

PipoCanaja:
Tested OK. Thx @SourceDoctor

PR #10791, opened 2019-11-07
Create device groups via API

BrianSidebotham:
This PR adds POST to the devicegroups API endpoint to enable adding a device group via the API.

It has been tested and working. It returns the ID of the newly generated device group and supports the typical static/dynamic rule generation.

Testing of the rule is rudimentary, but follows the same as the `add_alert` API endpoint. It's reasonably easy to break things by messing the rule up though. Is there any better rule validation that we can do here? (And add to the `add_alert` endpoint?) If not, we can improve that later.

I would also like to add DELETE and GET methods for individual device groups so the API is more complete. We can do that in future PRs.

Please review and let me know your feedback. It should be good to go. The documentation may need a little re-formatting as I deviated slightly to include the *optional* and *required* notes in.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/API/Alerts.md
+++ b/doc/API/Alerts.md
@@ -36,7 +36,7 @@ Output:
    "alerted": "1",
    "open": "1",
    "timestamp": "2014-12-11 14:40:02"
-  },
+  }]
 }
 ```
 
@@ -140,7 +140,7 @@ Output:
    "alerted": "1",
    "open": "1",
    "timestamp": "2014-12-11 14:40:02"
-  }
+  }]
 }
 ```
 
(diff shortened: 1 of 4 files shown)


doc/API/Alerts.md
```diff
@@ -36,7 +36,7 @@ Output:
    "alerted": "1",
    "open": "1",
    "timestamp": "2014-12-11 14:40:02"
-  },
+  }]
 }
```
BrianSidebotham: These changes are due to incorrect JSON in the triple backtick code block.

murrant:
This looks good overall.

On the validation front, we are already using json schema files in other places to validate schema.  However, no schema file exists for the alert rules, so it would need to be generated.

murrant:
Yeah Laravel validation is nice and in the end you will want to combine both to fully validate: https://laravel.com/docs/5.8/validation

murrant:

PR #10792, opened 2019-11-07
Optimize DB update of ports and ports_statistics tables when polling

louis-oui:
Two optimizations to reduce useless update request in database : 
- table ports_statistics : Don't update ports stats in database when their ifOperStatus is down and their ifLastChange did not change - ports_statistics table
- table ports : Update the poll_time, poll_prev and poll_period of all ports in an unique request per device. Do not request update when iflastchange is the same.

This change optimizes the numbers of update request in database during port polling phase.

Gain for total polling time is significant when the disk activity in term of IO per second is saturated without the patch.

Significant gain in term of CPU usage

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/polling/ports.inc.php
+++ b/includes/polling/ports.inc.php
@@ -535,6 +535,20 @@ foreach ($port_stats as $ifIndex => $port) {
 
 
 echo "\n";
+
+// get last poll time to optimize poll_time, poll_prev and poll_period in table db
+$last_poll_time = 0;
+foreach ($ports as $port) {
+    if ($port['poll_time'] > $last_poll_time) {
+        $last_poll_time = $port['poll_time'];
+    }
+}
+
+$device_global_ports = [];
+$device_global_ports['poll_time']  = $polled;
+$device_global_ports['poll_prev']  = $last_poll_time;
+$device_global_ports['poll_period']  = ($polled - $port['poll_time']);
+
 // Loop ports in the DB and update where necessary
 foreach ($ports as $port) {
     $port_id = $port['port_id'];
@@ -578,10 +592,12 @@ foreach ($ports as $port) {
         if ($port_association_mode != "ifIndex") {
             $port['update']['ifIndex'] = $ifIndex;
         }
-
-        $port['update']['poll_time'] = $polled;
-        $port['update']['poll_prev'] = $port['poll_time'];
-        $port['update']['poll_period'] = $polled_period;
+        
+        if ($last_poll_time != $port['poll_time']){
+            $port['update']['poll_time'] = $polled;
+            $port['update']['poll_prev'] = $port['poll_time'];
+            $port['update']['poll_period'] = $polled_period;
+        }
 
         if ($device['os'] === 'airos-af' && $port['ifAlias'] === 'eth0') {
             $airos_stats = snmpwalk_cache_oid($device, '.1.3.6.1.4.1.41112.1.3.3.1', array(), 'UBNT-AirFIBER-MIB');
@@ -646,7 +662,9 @@ foreach ($ports as $port) {
 
         // update ifLastChange. only in the db, not rrd
         if (isset($this_port['ifLastChange']) && is_numeric($this_port['ifLastChange'])) {
-            $port['update']['ifLastChange'] = $this_port['ifLastChange'];
+            if ((int)$this_port['ifLastChange'] != (int)$port['ifLastChange']) {
+                $port['update']['ifLastChange'] = $this_port['ifLastChange'];
+            }
         } elseif ($port['ifLastChange'] != 0) {
             $port['update']['ifLastChange'] = 0;  // no data, so use the same as device uptime
         }
@@ -939,7 +957,9 @@ foreach ($ports as $port) {
 
         // Update Database
         if (count($port['update'])) {
-            $updated = dbUpdate($port['update'], 'ports', '`port_id` = ?', array($port_id));
+            if (!empty($port['update'])) {
+                $updated = dbUpdate($port['update'], 'ports', '`port_id` = ?', array($port_id));
+            }
             // do we want to do something else with this?
             if (!empty($port['update_extended'])) {
                 $updated += dbUpdate($port['update_extended'], 'ports_statistics', '`port_id` = ?', array($port_id));
@@ -954,6 +974,10 @@ foreach ($ports as $port) {
     // Clear Per-Port Variables Here
     unset($this_port, $port);
 } //end port update
+
+// Update the poll_time, poll_prev and poll_period of all ports in an unique request
+$updated = dbUpdate($device_global_ports, 'ports', '`device_id` = ? AND poll_time = ? AND disabled = 0 AND deleted = 0', array($device['device_id'], $device_global_ports['poll_prev']));
+d_echo("$updated updated");
 
 // Clear Variables Here
 unset($port_stats, $ports_found, $data_oids, $stat_oids, $stat_oids_db, $stat_oids_db_extended, $cisco_oids, $pagp_oids, $ifmib_oids, $hc_test, $ports_mapped, $ports, $_stat_oids, $rrd_def);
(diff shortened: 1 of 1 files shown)


louis-oui:
Fix for performance optimization. Useful to save IO on disk when disk is short of IO

includes/polling/ports.inc.php
```diff
@@ -764,7 +764,7 @@
         }//end if
 
         // We don't care about statistics for skipped selective polling ports
-        if (!empty($port['skipped'])) {
+        if (!empty($port['skipped']) || ($port['ifOperStatus'] == "down" && $port['ifOperStatus_prev'] == "down" && $this_port['ifOperStatus'] == "down" && $this_port['ifLastChange'] == $port['ifLastChange'])) {
```
PipoCanaja: Bonjour ! 
Looks good, even if I wonder if some of these tests are in fact redundants. But I need to have a deeper look. That beeing said, you could probably comment this a little more. I would also suggest to go with "elseif" statement : makes it a little bit more verbose, but that would separate the selective polling code from this ifOperStatus=down optimisation. 
Merci

murrant: What is the gains from this change?  (Polling time change)

louis-oui: It depends on the disk activity in term of IO per second. 
One of our server was running out of IO. **Polling total time changed from 2200s to 1500s with 320 polled equipments** (~320 IOPS write disk perf)
No significant performance improvment observed on another server with the same number of equipments but disk activity was not overchaged (~600 IOPS write disk perf)

louis-oui:
do no merge yet. I checking some other things

PipoCanaja:
@louis-oui Salut ! You can remove the "WIP" text when this PR is ready for review. That will prevent any merge to occur in the meantime. 
A plus

louis-oui:
Bonjour @PipoCanaja 
Fix is ready to be merged.

PipoCanaja:
Ok. Will run it in test environnement.

includes/polling/ports.inc.php
```diff
@@ -763,9 +781,12 @@
             }
         }//end if
 
-        // We don't care about statistics for skipped selective polling ports
         if (!empty($port['skipped'])) {
-            echo " $port_id skipped.";
+            // We don't care about statistics for skipped selective polling ports
+            echo " $port_id skipped because selective polling ports is set.";
```
murrant: You could make this a d_echo() so it is only output when debug is enabled.

includes/polling/ports.inc.php
```diff
@@ -535,6 +535,20 @@
 
 
 echo "\n";
+
+// get last poll time to optimize poll_time, poll_prev and poll_period in table db
+$last_poll_time = 0;
```
murrant: $last_poll_time = max(array_column($port, 'poll_time'));

murrant:
@louis-oui I like your optimizations, but I think they should be implemented a little differently.

The statistics optimization is good as-is.

The poll timers could be better. My thoughts:
1. Move the poll time update fields into the `count($port['update'])` check that way if the port is going to be updated, the time values are updated as well
2. record port_ids that were skipped above and update the poll times of them all at once.

includes/polling/ports.inc.php
```diff
@@ -578,10 +588,12 @@
         if ($port_association_mode != "ifIndex") {
             $port['update']['ifIndex'] = $ifIndex;
         }
-
-        $port['update']['poll_time'] = $polled;
-        $port['update']['poll_prev'] = $port['poll_time'];
-        $port['update']['poll_period'] = $polled_period;
+        
+        if ($last_poll_time != $port['poll_time']) {
```
murrant: Can this code block be deleted?

includes/polling/ports.inc.php
```diff
@@ -535,6 +534,17 @@
 
 
 echo "\n";
+
+// get last poll time to optimize poll_time, poll_prev and poll_period in table db
+$last_poll_time = max(array_column($ports, 'poll_time'));
+
+$device_global_ports = [];
```
murrant: I like it a little better when arrays are defined like this:
```php
$device_global_ports = [
    'poll_time' => $polled,
    'poll_prev' => max(array_column($ports, 'poll_time')),
    'poll_period' => ($polled - $port['poll_time']),
];
```

louis-oui: ok done

includes/polling/ports.inc.php
louis-oui: I think we need to keep that
// set $port['update'] times if different from $device_global_ports
// if $port['update'] times not set and $port['update'] is void, value would be replaced by

louis-oui:
@murrant Do you agree with new code ?

includes/polling/ports.inc.php
```diff
@@ -952,5 +979,9 @@
     unset($this_port, $port);
 } //end port update
 
+// Update the poll_time, poll_prev and poll_period of all ports in an unique request
+$updated = dbUpdate($device_global_ports, 'ports', '`port_id` IN ('.implode(',', $port_ids).')', array($device['device_id'], $device_global_ports['poll_prev']));
```
murrant: dbUpdate doesn't link IN being used like that. It will escape it in a way that can fail.  There are no placeholders, yet you specify two values.
You could just use this: `DB::table('ports')->whereIn('port_id', $port_ids)->update($device_global_ports);`

includes/polling/ports.inc.php
```diff
@@ -936,12 +954,21 @@
...
         // Update Database
         if (count($port['update'])) {
-            $updated = dbUpdate($port['update'], 'ports', '`port_id` = ?', array($port_id));
+            if (!empty($port['update'])) {
+                $port['update']['poll_time'] = $polled;
+                $port['update']['poll_prev'] = $port['poll_time'];
+                $port['update']['poll_period'] = $polled_period;
+                $updated = dbUpdate($port['update'], 'ports', '`port_id` = ?', array($port_id));
+            } else {
+                $port_ids[] =  $port_id;
```
murrant: `if (!empty($port['update'])`  is pretty much the same as `if (count($port['update']))` so this will always be true  (I do prefer the not empty notation)

This means this else is a dead code path.

includes/polling/ports.inc.php
```diff
@@ -578,10 +587,14 @@
         if ($port_association_mode != "ifIndex") {
             $port['update']['ifIndex'] = $ifIndex;
         }
-
-        $port['update']['poll_time'] = $polled;
-        $port['update']['poll_prev'] = $port['poll_time'];
-        $port['update']['poll_period'] = $polled_period;
+        
+        // set $port['update'] times if different from $device_global_ports
+        // if $port['update'] times not set and $port['update'] is void, value would be replaced by $device_global_ports
```
murrant: I still don't follow why this code block is necessary.  Isn't that the reason we build $port_ids.  (also $port_ids variable name could be more descriptive)

murrant:
@louis-oui sorry I haven't gotten back and thanks for updating the code.  Sorry I'm being so strict about these, but this is one of our more complex and critical pieces of code.  I would like to avoid making it harder to understand.

Btw, these are some really great changes you have submitted ;)

includes/polling/ports.inc.php
murrant: Can't you just update your $port_ids array?

louis-oui:
@murrant I followed your instructions

louis-oui:
Hi @PipoCanaja, I have followed @murrant guidelines. What do you want to do now ?

PipoCanaja:
Hi
Looks good. Needs to be tested.

louis-oui:
@PipoCanaja @murrant Any news on this PR ?

louis-oui:
Hello @murrant. Happy new year.

What should we do with this patch ?

louis-oui:
@PipoCanaja I am using this fix for a long time and I have followed @murrant guidelines
Is it possible that somebody else tests there is no regression when polling ?

murrant:

https://community.librenms.org/t/optimize-poller-performance-tip/10599/1

murrant:

murrant:

https://community.librenms.org/t/poller-performance-regression-in-1-59/10693/9

PR #10796, opened 2019-11-09
UI bug fix - Scroll disabled on "new rule from collection" modal

GramThanos:
How to reproduce:
- Reduce browser window height (so that view-port is smaller than 700px)
- Go to Alert Rules
- Click "Create rule from collection" button
- The new modal can be scrolled
- Click any rule's "select"
- The new modal is does not scroll, instead the hole page scrolls while the modal stay fixed

How to fix the bug:

To fix this bug, you need to wait the first modal to close (wait the animations too), before opening the second modal. This can be done by listening for the `hidden.bs.modal` event ~~(and then instantly removing the event handler by calling the `.off(event)`)~~ once.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/html/modal/alert_rule_collection.inc.php
+++ b/includes/html/modal/alert_rule_collection.inc.php
@@ -83,9 +83,12 @@ if (!Auth::user()->hasGlobalAdmin()) {
                                     dataType: "json",
                                     success: function (data) {
                                         if (data.status == 'ok') {
+                                            $("#search_rule_modal").on('hidden.bs.modal', function(event) {
+                                                loadRule(data);
+                                                $('#create-alert').modal('show');
+                                                $("#search_rule_modal").off(event);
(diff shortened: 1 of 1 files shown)


includes/html/modal/alert_rule_collection.inc.php
```diff
@@ -83,9 +83,12 @@
                                     dataType: "json",
                                     success: function (data) {
                                         if (data.status == 'ok') {
+                                            $("#search_rule_modal").on('hidden.bs.modal', function(event) {
```
Jellyfrog: Can't you use `one()` instead? No need to have the `off()` then

GramThanos: Yes!

murrant:

PR #10568, opened 2019-08-29
Device group based access

Jellyfrog:
If you have many devices it's a big burden to add access to each new device.
Idea here is to add support to also allow access to devices based on device group membership.  

There are a lot of places where this won't work currently, mostly due to the old codebase that runs raw SQL queries, in fact the only place it works at the moment is the pages under `/device/`.  

Many of the pages can be fixed with a generic copy-paste solution, some requires more work, but the real solution is to convert them all to laravel/eloquent.

I suggest we mark this as "experimental" and release it as is (when we're happy with it)
Then I can look into converting more pages.

image

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #10802, opened 2019-11-10
fix restricted application access for normal user

SourceDoctor:
normal user with no read access to a device should also have no access to it's application.
This is the fix for this.

### DIFF
--- a/includes/html/pages/apps/default.inc.php
+++ b/includes/html/pages/apps/default.inc.php
@@ -12,7 +12,7 @@ $graph_array_zoom['height'] = '150';
 $graph_array_zoom['width']  = '400';
 $graph_array['legend']      = 'no';
 
-$apps = Application::query()->hasAccess(Auth::user())->where('app_type', $vars['app'])->with('device')->get()->sortBy(function($app) {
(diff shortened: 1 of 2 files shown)


Jellyfrog:
Please use Permission class to get user devices, you can find a lot of examples here https://github.com/librenms/librenms/pull/10568

SourceDoctor:
@Jellyfrog Permissions class now used  :-)

but on using "Permissions::devicesForUser()" i'd expect for user with global read a list of all device_ids not an empty array.
devices_perms table has only access list for normal users. :-)

Jellyfrog:
Yeah, the real fix is ofc to use Eloquent model :)  
`Device::canAccess()` etc.

SourceDoctor:
this way works also fine ;-) don't panic :beers:

vicomte:
Just ran into this. Thanks for taking a look! Works for me.

murrant:
I rewrote this with Eloquent, want me to update this PR?

SourceDoctor:
@murrant if you like do so :-)

murrant:

PR #10825, opened 2019-11-16
alphabetic order of app overview

SourceDoctor:
sort applications alphabetic on overview page


Jellyfrog:
you should put sortBy() Before get(), so it's happens in sql instead of the collection

SourceDoctor:
@Jellyfrog 
i tried before doing pull request and got a crash ... this is the way it works ;-)

murrant:
It needs to be ->orderBy('app_type') if you do it in the SQL query (which you should)

murrant:

PR #10834, opened 2019-11-17
Show/hide maintenance devices

gdepeyrot:
Added WorldMap possibility to show/hide devices under maintenance

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


app/Http/Controllers/Widgets/WorldMapController.php
```diff
@@ -44,7 +44,8 @@ public function __construct()
             'init_lng' => Config::get('leaflet.default_lng', 0),
             'init_zoom' => Config::get('leaflet.default_zoom', 2),
             'group_radius' => Config::get('leaflet.group_radius', 80),
-            'status' => '0,1',
+            'status' => '0,1', // Show all devices
+            'maintenance' => '0', // Hide devices under maintenance
```
murrant: Any reason you are using textual numbers instead of just numbers?

gdepeyrot: No, I guess not.

Jellyfrog:
The logic should be moved to the sql query above, so you only retrieve the devices you want, then you don't need the php logic

gdepeyrot:
> The logic should be moved to the sql query above, so you only retrieve the devices you want, then you don't need the php logic

The sql query above already filters the device according to the selected status (Up + Down, Up or Down). AFAIU, there is no simple query to filter devices under maintenance. Is it possible to combine the isUnderMaintenance check into the same query?

Maybe I haven't understood what you are suggesting.

Jellyfrog:
> > The logic should be moved to the sql query above, so you only retrieve the devices you want, then you don't need the php logic
> 
> The sql query above already filters the device according to the selected status (Up + Down, Up or Down). AFAIU, there is no simple query to filter devices under maintenance. Is it possible to combine the isUnderMaintenance check into the same query?
> 
> Maybe I haven't understood what you are suggesting.

You're right, I forgot it comes from another table.
It is possible to make it a "scope function" like it's called in laravel, do you think it's worth it @murrant ?

PipoCanaja:
@murrant @Jellyfrog I let you continue on this one. From my side, looks good, but I am not the DB guy at all :)

murrant:
You should be able to create a scope on the Devices model by using the isUnderMaintenance query as a subquery.

gdepeyrot:
Ok, I'll have a look at how to implement that.

kkrumm1:
Any update on this?

gdepeyrot:
Unfortunately not. Been swamped with other subjects and have not been able to look into this.

SourceDoctor:
@gdepeyrot 
something new about pending changes?

gdepeyrot:
No, nothing new. Don't know when I'll have time to learn how to create the suggested scope. Right now concentrated on the survival of my company during the COVID-19 crisis...

Jellyfrog:
Closing due to inactivity, feel free to reopen

PR #10842, opened 2019-11-18
water sensor support for sentry 3/4 pdus

sjtarik:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


sjtarik:
@laf @murrant I need little help fixing the Travis CI tests. Tests expects the OS version and serial number of PDU to be exactly same in tests/data/sentry3.json and tests/data/sentry4.json. Should I just delete these fields null or is there a way to make them optional?

murrant:
You changed the data in the snmprec file, without updating the json file (which you need to do anyway since you have added sensors)

PipoCanaja:
@sjtarik 
The best solution in your case would be to leave the existing snmprec as they are (so revert them), and create new ones as variants (use  "./collect_snmp_data.php -h <device> -v myModel" to create sentry3_mymodel.snmprec , if the device is a sentry3). That way, we extend our test base instead of replacing it.

Jellyfrog:
Closing due to inactivity. Feel free to reopen

PR #10856, opened 2019-11-20
Graphical alerts update (text/icon alignment)

PipoCanaja:
The goal is to have the UserName and the alert number correctly aligned.

Before: 
Capture d’écran 2019-11-20 à 11 18 57

After: 
Capture d’écran 2019-11-20 à 11 19 29

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/css/styles.css
+++ b/html/css/styles.css
@@ -1817,10 +1817,11 @@ label {
 .badge-navbar-user {
    border-radius: 40%;
    font-size: 65%;
-   padding: 5px;
+   padding: 4px;
    position: relative;
    top: -10px;
-   left: -5px;
+   left: -11px;
+   margin-right: -14px;
 }
 .badge-default {
   background-color: #777;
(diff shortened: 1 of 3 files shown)


Jellyfrog:
Think ill prefer the old look

Cormoran96:
I don't like the old style. It gives the impression of a bug

SourceDoctor:
what about making position choosable? up(old look)/middle/down?
in ideal case per user ;-)

murrant:
@SourceDoctor we try to avoid option creep :)

murrant:
How about this?

image

Cormoran96:
And if there is 0 notification the 0 is hidden. And when we are more the number is exponent of icon. 
something like that:
Screenshot 2019-11-22 at 08 45 01

PipoCanaja:
I was thinking of moving the exponent style number back to the icon, and it really makes sense. I'll update the PR in that direction.

PipoCanaja:
As murrant suggested : 
Capture d’écran 2019-11-22 à 19 40 24

SourceDoctor:
@PipoCanaja 
nice :-)
but i personally like the idea from Cormoran96 only to see the Counter if messages are existing.
Maybe change it to this?
What do you think about it @murrant ?

PipoCanaja:
With notifications : 
Capture d’écran 2019-11-23 à 15 06 03

Without unread notifications: 
Capture d’écran 2019-11-23 à 15 06 25

murrant:

PR #10864, opened 2019-11-21
Update Oxidized.md

achrstl:
Hello,
I think this could be really useful, i searched for a long time before finding that command would work in a bare repo.
It allow to remove deleted devices that we don't have anymore.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Oxidized.md
+++ b/doc/Extensions/Oxidized.md
@@ -252,3 +252,9 @@ file content:
 ```
 git cat-file -p <object id>
 ```
+
+# Remove disabled/removed device
+
+```
+git rm --cached <object id>
+```
--- a/doc/Extensions/Oxidized.md
+++ b/doc/Extensions/Oxidized.md
@@ -254,6 +254,7 @@ git cat-file -p <object id>
 ```
 
 # Remove disabled/removed device
+If you want to purge saved config of a device that is not in LibreNMS anymore, you can run the following command: 
 
 ```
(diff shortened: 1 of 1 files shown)


murrant:
Hmm, I think you need to add a little more context here. To explain why people would run the command you listed.

We don't remove retired devices from our git repo because it is a tiny amount of storage and it is not visible in the ui or anything.

If you can add some more context, I'll be happy to merge.

achrstl:
I use the oxidized repo to search for config part.
By example if i need to quickly look where the IP 1.2.3.4 is configured: `git grep 1.2.3.4`
The output, will return me which device is configured with this IP.
Now, if an old device had this IP, i will have two results.
In my case, i have some automation script that need to match with the current network state.

Thats why i think it could be relevant to add this command.

doc/Extensions/Oxidized.md
```diff
@@ -252,3 +252,9 @@ file content:
 ```
 git cat-file -p <object id>
 ```
+
+# Remove disabled/removed device
```
PipoCanaja: As suggested by murrant, you could add something like : 
```
If you want to purge saved config of a device that is not in LibreNMS anymore, you can run the following command: 
```

PipoCanaja:
Hi @achrstl 
Looks good ! Thanx for your contribution.

murrant:

PR #10879, opened 2019-11-25
Detect if LibreNMS is installed from the official Docker image (librenms/docker#46)

crazy-max:
See librenms/docker#46

#### Please note

#### Testers


murrant:
When updating a docker image, do you just replace the whole image?

crazy-max:
@murrant Yes, indeed, LibreNMS installation folder is not persisted so I think this check is necessary to avoid any confusion when a user executes the validate command.

PR #10885, opened 2019-11-26
Adds API to fetch all inventory items for a given device

ajsiersema:
The current inventory API takes 742 API calls, 1m43 and 1Mb traffic in 2467 packets
to fetch the inventory items for a very common use case (core switch/router).
This proposal to extend the inventory API closely reflects the UI and only takes 1 API call, 0.5s and 237Kb traffic in 76 packets

I understand the reasoning that went into the concept of the current inventory API, but this proposal performs better and won't hammer the server (and database) as much.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1199,16 +1199,7 @@ function get_inventory_for_device(\Illuminate\Http\Request $request)
     $device_id = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
     return check_device_permission($device_id, function ($device_id) use ($request) {
         $params    = [];
-        $sql = 'SELECT `D`.`device_id` AS `device_id`,
-                       `D`.`hostname` AS `hostname`,
-                       `entPhysicalDescr` AS `description`,
-                       `entPhysicalName` AS `name`,
-                       `entPhysicalModelName` AS `model`,
-                       `entPhysicalSerialNum` AS `serial`,
-                       `entPhysicalHardwareRev` as `revision`,
-                       `entPhysicalFirmwareRev` as `version`
-               FROM entPhysical AS E, devices AS D
-               WHERE `D`.`device_id`=? AND D.device_id = E.device_id';
+        $sql = 'SELECT * FROM `entPhysical` WHERE device_id = ?';
         $params[] = $device_id;
         $inventory = dbFetchRows($sql, $params);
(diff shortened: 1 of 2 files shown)


Jellyfrog:
What am I missing? You get the same info, except hostname, from the current inventory endpoint

ajsiersema:
> What am I missing? You get the same info, except hostname, from the current inventory endpoint

Well, if I want to get the inventory information for a single device it now takes 742 "recursive" API calls (and database queries) and almost 2 minutes. Imagine getting the inventory information for several 100 or 1000s devices. Instead of hammering the librenms API and database we can now fetch the information at a fraction of the load. The GUI presents us with that option, the API will then provide the same option.

ajsiersema:
> What am I missing? You get the same info, except hostname, from the current inventory endpoint

From a functional viewpoint : I'd like to be able to pull inventory information for all devices (hundreds in my case) to automatically update the CMDB repository. The current API endpoint is both very slow and is hammering the server and database.

A side effect could be to base the UI code on the REST API somewhere in the future, i.e. replace the database queries in the UI code with a call to the REST API, render this information in a template based view :)

Jellyfrog:
This does not fetch all info tho, what about just doing `select *` instead?

ajsiersema:
> This does not fetch all info tho, what about just doing `select *` instead?

Not a problem. As it stands, I've roughly copied the information fetched from the code used by the UI, for a kind of feature parity.
And like mentioned above, it might also serve as a starting point for shifting the 'database code' from the UI code to the API, after which the UI uses the API (as well.)

Jellyfrog:
The current inventory endpoint does a `select *`, so I think this should do the same, especially since the route is called `all`, also, we don't want to rename the columns in the output.  
```sql
SELECT * FROM `entPhysical` WHERE device_id = ?
```
Should be enough

ajsiersema:
> 
> 
> The current inventory endpoint does a `select *`, so I think this should do the same, especially since the route is called `all`, also, we don't want to rename the columns in the output.
> 
> ```sql
> SELECT * FROM `entPhysical` WHERE device_id = ?
> ```
> 
> Should be enough

Apologies, I got a bit overwhelmed with work because a bout of a common cold left me a bit less productive. I've updated the pull request.

murrant:

PR #10887, opened 2019-11-27
python-ipaddress is required by snmp-scan.py

brianegge:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
We should probably just force the script to use python 3 at this point...

brianegge:
My install was using python 3 on a Ubuntu 19 server. 

-Brian

murrant:
python-ipaddres doesn't exist for python 3... it is built in.

brianegge:
I see. I thought the default python would be python3. I see Ubuntu 19 defaults to `Python 2.7.17rc1`. I would be in favor of the scan script using python3 via `#!/usr/bin/env python3`

PR #10895, opened 2019-12-02
Added FortiMail support

footstep86:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 10895 `


includes/discovery/mempools/fortimail.inc.php
```diff
@@ -0,0 +1,7 @@
+<?php
+if ($device['os'] == 'fortimail') {
```
PipoCanaja: Device OS will always be fortimail here, so you should remove this condition.

includes/polling/mempools/fortimail.inc.php
```diff
@@ -0,0 +1,8 @@
+<?php
+echo 'FortiMail MemPool';
+$mempool['perc'] = snmp_get($device, 'FORTINET-FORTIMAIL-MIB::fmlSysMemUsage.0', '-OvQ');
+// There is no OID for MemCapacity. Using fixed value based on "diag hardware sysinfo"
+$mempool['total'] = 4123984 * $mempool['mempool_precision'];
```
PipoCanaja: Could you run a complete SNMPWALK of your device, and search inside it if this value appears somewhere. Would be a little bit annoying to hardcode this especially when next hardware revision will change the embedded memory ...

PipoCanaja:
Hi @footstep86 
Thanx for your PR ! Please take a few minutes to accept the CLA (without which we cannot merge this change). 
A few comments are already inlined.

includes/polling/mempools/fortimail.inc.php
footstep86: Thanks for your input, but unfortunatly this value is really unavailable using SNMP. I've already opened a support ticket at Fortinet regarding that.

includes/discovery/mempools/fortimail.inc.php
footstep86: I followed the documentation here: https://docs.librenms.org/Developing/os/Mem-CPU-Information. But if this condition really is obsolete, I'll remove it from my code.

footstep86:
> Hi @footstep86
> Thanx for your PR ! Please take a few minutes to accept the CLA (without which we cannot merge this change).
> A few comments are already inlined.

How to accomplish that?

includes/discovery/mempools/fortimail.inc.php
footstep86: If I got it right, the includes/discovery/mempools.inc.php will run all available mempool-definitions against a device. The condition inside a definition takes care that only the relevant SNMP queries are executed. --> Maybe this kind of logic could be done within mempools.inc.php?

includes/polling/mempools/fortimail.inc.php
footstep86: I just received feedback from Fortinet that they won't implement this OID any time soon.

includes/discovery/mempools/fortimail.inc.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+echo 'FORTIMAIL-MEMORY-POOL: ';
+// fmlSysMemCapacity: Total physical memory (RAM) installed (KB)
+// Conversion from kilobyte (base 2) to byte (base 10): Precision = 1024
+discover_mempool($valid_mempool, $device, 0, 'fortimail', 'Physical Memory', '1024', null, null);
```
laf: You need to do the `snmp_get()` before this otherwise it will discovery a mempool for all fortimail devices (which may be genuinely possible however you shouldn't do this).

SourceDoctor:
@footstep86 
something new about pending changes?

includes/discovery/mempools/fortimail.inc.php
footstep86: I've just tried to generate the test data but I'm facing the following issue:
SQLSTATE[HY000] [1045] Access denied for user 'root'@'localhost' (using password: NO) (SQL: select * from `devices` where `hostname` = 127.1.6.1 limit 1)

SourceDoctor: as i remember some of the things has to be run in user mode and not as root
Maybe this helps:

SourceDoctor:
@footstep86 
could you take a look on to travis error, please?

includes/discovery/mempools/fortimail.inc.php
footstep86: I've managed to get the script running. But even when i start the script from within my development directory, it only scans the production directory for data files: 
- Script started:  /opt/librenmsdev/scripts/save-test-data.php -o fortimail
- Snippet of the output:
  Scanning "/usr/snmpsim/data" directory for  *.snmpwalk, *.MVC, *.sapwalk, *.snmprec, *.dump data files...
  Directory "/usr/snmpsim/data" does not exist

SourceDoctor:
@footstep86 please check the Checkbox on PR start

@librenms/reviewers 
can someone verify functionality?

includes/discovery/mempools/fortimail.inc.php
footstep86: I was finally able to generate the requested test data.

murrant:

PR #10907, opened 2019-12-05
Do not update alert timestamp when updating a triggered alert

louis-oui:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/database/migrations/2019_12_05_164700_alerts_disable_on_update_current_timestamp.php
+++ b/database/migrations/2019_12_05_164700_alerts_disable_on_update_current_timestamp.php
@@ -1,0 +1,28 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+
+class AlertsDisableOnUpdateCurrentTimestamp extends Migration
+{
+
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        \DB::statement("ALTER TABLE `alerts` CHANGE `timestamp` `timestamp` timestamp NOT NULL default CURRENT_TIMESTAMP;");
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        \DB::statement("ALTER TABLE `alerts` CHANGE `timestamp` `timestamp` timestamp NOT NULL default CURRENT_TIMESTAMP on update CURRENT_TIMESTAMP;");
+    }
+}
--- a/LibreNMS/Alert/AlertRules.php
+++ b/LibreNMS/Alert/AlertRules.php
@@ -103,7 +103,7 @@ class AlertRules
                         if (is_null($current_state)) {
                             dbInsert(array('state' => 1, 'device_id' => $device_id, 'rule_id' => $rule['id'], 'open' => 1,'alerted' => 0), 'alerts');
                         } else {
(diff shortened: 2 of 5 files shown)


louis-oui:
alerts table is configured to update timestamp whenever an entry is updated because of "on update CURRENT_TIMESTAMP" parameter

```
MariaDB [librenms]> desc alerts;
+-----------+------------------+------+-----+-------------------+-----------------------------+
| Field     | Type             | Null | Key | Default           | Extra                       |
+-----------+------------------+------+-----+-------------------+-----------------------------+
| id        | int(10) unsigned | NO   | PRI | NULL              | auto_increment              |
| device_id | int(10) unsigned | NO   | MUL | NULL              |                             |
| rule_id   | int(10) unsigned | NO   | MUL | NULL              |                             |
| state     | int(11)          | NO   |     | NULL              |                             |
| alerted   | int(11)          | NO   |     | NULL              |                             |
| open      | int(11)          | NO   |     | NULL              |                             |
| note      | text             | YES  |     | NULL              |                             |
| timestamp | timestamp        | NO   |     | CURRENT_TIMESTAMP | on update CURRENT_TIMESTAMP |
| info      | text             | NO   |     | NULL              |                             |
```

When polling change the state from 0 to 1 or 1 to 0, it is OK to change timestamp because timestamp reflects time the alert is triggered or recovered.

However, when alerts.php update alerts entries, we won't change to the change the timestamp as the alert is not new. Case alerts.php updates alerts entries are:
- state column is changed to value 3 or 4 : respectively "worse" and "better" state
- alert column is set to state value

louis-oui:
Code climate failed because of legacy code or Laravel external code

database/migrations/2019_12_05_164700_alerts_disable_on_update_current_timestamp.php
```diff
@@ -0,0 +1,28 @@
...
+{
+
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        \DB::statement("ALTER TABLE `alerts` CHANGE `timestamp` `timestamp` timestamp NOT NULL default CURRENT_TIMESTAMP;");
```
Jellyfrog: Please use the "real" migration way,
`$this->timestamp('timestamp')->useCurrent();` maybe works.

louis-oui: Thank you for the review
I got this issue https://stackoverflow.com/questions/34774628/how-do-i-make-doctrine-support-timestamp-columns
Tried 
```
        Schema::table('alerts', function (Blueprint $table) {
            $table->datetime('timestamp')->useCurrent()->change();
        });
```

Jellyfrog: I see. Annoying.

louis-oui:
Before merging, we need to make sure that date in file is last one in list database/migrations/2019_12_05_164700_alerts_disable_on_update_current_timestamp.php
Otherwise, migration won't work

louis-oui:
@Jellyfrog Please let me know what do you want to do with this PR ?

louis-oui:
@Jellyfrog any news on it ?

murrant:

https://community.librenms.org/t/do-not-change-alert-timestamp-when-acknowledge/10698/1

murrant:

PR #10916, opened 2019-12-08
Graphing Device Dependency

SourceDoctor:
Graphing Device Dependency

#### Please note

#### Testers


kkrumm1:
Can you post screenshot of this?

SourceDoctor:
@kkrumm1 @PipoCanaja 
just a reuse of "Communication Map" Graphiz Graph ;-)

Lonesome Nodes have no dependencies. They are graphed also so it can be seen they are "alone".

image

includes/html/print-dependency-map.inc.php
```diff
@@ -0,0 +1,124 @@
+<?php
+
+use LibreNMS\Config;
+
+if (!Auth::user()->hasGlobalRead()) {
+    $sql = 'WHERE device_id IN (SELECT device_id FROM devices_perms WHERE user_id='. Auth::id() . ')';
```
Jellyfrog: Please use Permission class to get user devices

Jellyfrog:
This page could be written with blade/laravel, see #10559 for an example

includes/html/print-dependency-map.inc.php
SourceDoctor: Ok, but where can i find an example to get all accessed devices at once?
I don't want to check every device in a single query for access.

Jellyfrog: Like so;
```Php
$device_ids = Permissions::devicesForUser()->toArray() ?: [0];
$sql = "WHERE `device_id` IN " .dbGenPlaceholders(count($device_ids));
$param = $device_ids;
```

However like I said in the other comment, this page should be written in laravel, and use the eloquent models functions

SourceDoctor:
> This page could be written with blade/laravel, see #10559 for an example

@Jellyfrog 
i need some input from you to get rewrite done

i created a class file in directory `app/Http/Controllers`
there to get this 1:1 adapted in need following functions from `includes/common.php`:
- shorthost
- format_hostname
- generate_device_link

so i insert following lines into class file:
> use function includes\common\shorthost;
> use function includes\common\format_hostname;
> use function includes\common\generate_device_link;

but functions are still not found ...

any ideas?

Jellyfrog:
You can do like `{!! \LibreNMS\Util\Url::deviceLink($device, $device->shortDisplayName()) !!}`

SourceDoctor:
@Jellyfrog 
rewritten to blade/laravel
but still one problem i didn't get:
In Code everything is present, but on mouseover over a node only hostname is listed but it should short device information with some RRD Graphs be seen like on other maps. 
Can't find the cause of this.

app/Http/Controllers/MapController.php
```diff
@@ -0,0 +1,133 @@
...
+    {
+        return array();
+    }
+
+    // Device Dependency Map
+    public function dependencyMap(Request $request)
+    {
+        $user = $request->user();
+
+        if (!$user->hasGlobalRead()) {
```
Jellyfrog: You can replace this entire if with just `Device::hasAccess(Auth::user())`

Jellyfrog: Also add `->with('parents')` to eagerload all parents, saves a lot of queries.

SourceDoctor: both fixed

SourceDoctor:
somebody an idea about the remaining mouseover Issue?

SourceDoctor:
@Jellyfrog @PipoCanaja 
Travis test corrupted? 

> ...
> pylint requires Python '>=3.5.*' but the running Python is 2.7.12
> ...

SourceDoctor:
ok, mouseover information runs.
But one problems remaining. Mouseover on every device type show mouseover display like it's a 'server'
even on switches PDU's and other device types

app/Http/Controllers/MapController.php
```diff
@@ -0,0 +1,126 @@
...
+use App\Models\Device;
+use Illuminate\Http\Request;
+use LibreNMS\Config;
+use LibreNMS\Util\Url;
+
+class MapController extends Controller
+{
+    public function __construct()
+    {
+        $this->middleware('deny-demo');
```
Jellyfrog: this should be removed?

app/Http/Controllers/MapController.php
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    protected function visOptions()
+    {
+        return Config::get('network_map_vis_options');
+    }
+
+    protected function nodeDisabledStyle()
+    {
+        return array('color' => array(
```
Jellyfrog: you have a mix between `array()`and `[]`, stick to the newer `[]` style

app/Http/Controllers/MapController.php
```diff
@@ -0,0 +1,126 @@
...
+    {
+        $devices = Device::hasAccess($request->user())->with('parents')->get();
+
+        $dependencies = array();
+        $devices_by_id  = [];
+
+        // Build the style variables we need
+
+        // List all devices
+        foreach ($devices as $items) {
```
Jellyfrog: maybe change it to '$device`

app/Http/Controllers/MapController.php
```diff
@@ -0,0 +1,126 @@
+<?php
```
Jellyfrog: move this file into its own namespace and folder

app/Http/Controllers/MapController.php
SourceDoctor: done

app/Http/Controllers/MapController.php
SourceDoctor: done

app/Http/Controllers/MapController.php
SourceDoctor: done

app/Http/Controllers/MapController.php
SourceDoctor: @Jellyfrog 
my intention was to create a base for all Maps, so own directory is not necessary and doubled code can be prevented.
Especially for rewriting NetworkMap Code, it only has to include another method into MapController Class

what do you think about this?

Jellyfrog: I mean create a folder in Controllers named Maps and there u put all mapcontrollers. Check widget folder for inspiration

SourceDoctor: changed :-)

SourceDoctor:
@Jellyfrog all changes are done

SourceDoctor:
@Jellyfrog something else to change?

app/Http/Controllers/Maps/DependencyController.php
```diff
@@ -0,0 +1,85 @@
...
+                    'id'    => $device['device_id'],
+                    'label' => $device->shortDisplayName(),
+                    'title' => Url::deviceLink($device, null, [], 0, 0, 0, 0),
+                    'shape' => 'box',
+                ],
+                $device_style
+            );
+
+            // List all Device Dependencies
+            $parents = $device->parents()->get();
```
Jellyfrog: You're making N+1 queries here, and not using the eager cache.  
```suggestion
            $parents = $device->parents;
```

Jellyfrog:
Probably it should be renamed DeviceDependencyController, since there are a lot of other dependencies

app/Http/Controllers/Maps/DependencyController.php
SourceDoctor: changed

SourceDoctor:
renamed

Jellyfrog:
need to rename route and blade file also

Jellyfrog:
need to rename route and blade file also

SourceDoctor:
@Jellyfrog done

djamp42:
I tried this out and because i have 6k devices it's unable to render the map, i left it for a couple of minutes, even if it does its probably not the best way to display this. I was hoping this would let me set the root node to start the dependency map drawing at.  Breaking them down by device group would be even better. It's a good start and i think graphing the dependency map is a really helpful and powerful tool.

SourceDoctor:
@djamp42 
do you have the same issue with `Overview -> Maps ->Network`, cause Code is similar.

djamp42:
> do you have the same problem with `Overview -> Maps ->Network`, because Code is similar.

Yes that page fails to load also, to be honest i never use it because seeing that many devices on one map doesn't make much sense. i only use the maps -> Device group maps.

app/Http/Controllers/Maps/DeviceDependencyController.php
```diff
@@ -0,0 +1,85 @@
...
+use App\Models\Device;
+use Illuminate\Http\Request;
+use LibreNMS\Util\Url;
+
+class DeviceDependencyController extends MapController
+{
+    // Device Dependency Map
+    public function dependencyMap(Request $request)
+    {
+        $devices = Device::hasAccess($request->user())->with('parents')->get();
```
Jellyfrog: More eager caching
```suggestion
        $devices = Device::hasAccess($request->user())->with(['parents', 'location'])->get();
```

app/Http/Controllers/Maps/DeviceDependencyController.php
```diff
@@ -0,0 +1,85 @@
...
+
+        $dependencies = [];
+        $devices_by_id  = [];
+
+        // Build the style variables we need
+
+        // List all devices
+        foreach ($devices as $device) {
+            if ($device['disabled']) {
```
Jellyfrog: `$device` is an object, please use `$device->disabled`

app/Http/Controllers/Maps/DeviceDependencyController.php
```diff
@@ -0,0 +1,85 @@
...
+                $device_style = $this->nodeDisabledStyle();
+            } elseif (! $device['status']) {
+                $device_style = $this->nodeDownStyle();
+            } else {
+                $device_style = $this->nodeUpStyle();
+            }
+
+            // List all Device
+            $devices_by_id[$device['device_id']] = array_merge(
```
Jellyfrog: Why do you need to set array key to device id when you remove it with array_values?

resources/views/map/device-dependency.blade.php
```diff
@@ -0,0 +1,40 @@
...
+
+@if($node_count)
+<div id="visualization"></div>
+@else
+<div class="alert alert-success" role="alert">@lang('No devices found')</div>
+@endif
+@endsection
+
+@section('scripts')
+<script src="js/vis.min.js"></script>
```
Jellyfrog: This line should be in `@section('javascript')`

Jellyfrog: And also use {{ asset('js/vis.min.js') }}

Jellyfrog:
> > do you have the same problem with `Overview -> Maps ->Network`, because Code is similar.
> 
> Yes that page fails to load also, to be honest i never use it because seeing that many devices on one map doesn't make much sense. i only use the maps -> Device group maps.
> 
> It would seem to me this map should also work this way, where your building the dependency map based on the devices the device group returns. I took a quick look at the code and it seems it wouldn't be too hard to get this working, I'll have to play around with it more.

The problem is the js lib getting slow with many nodes.  
@SourceDoctor maybe you can look into updating it / changing some parameters?

SourceDoctor:
I took before 2 weeks a look on project Page where vis.min.js comes from, about values and features. There seems to be a default limit to 600 Nodes. Have to play around with it some times later, especially about Engine Power.
Major Reason was lack of Feature to implement a Legend which describes Nodes and Connection States and Colors.
But this will be done from my side in another PR.
Also i'll implement here in another PR Node Highlighting. But first Code in #10943 has to be approved, also this PR.

app/Http/Controllers/Maps/DeviceDependencyController.php
SourceDoctor: @Jellyfrog did i miss something?
i found in code nowhere usage of "location"

SourceDoctor: changed

app/Http/Controllers/Maps/DeviceDependencyController.php
SourceDoctor: changed

app/Http/Controllers/Maps/DeviceDependencyController.php
SourceDoctor: It's an adaption of `Map->Network`.
I try it without, run's also fine. So i removed array key setting.

resources/views/map/device-dependency.blade.php
SourceDoctor: changed to `{{ asset('js/vis.min.js') }}`
changing to `@section('javascript')` has the effect no Map is graphed

Jellyfrog:
```blade
@section('javascript')
<script type="text/javascript" src="{{ asset('js/vis.min.js') }}"></script>
@endsection
```
Works for me.

SourceDoctor: ok,. this works on me also.
Changed to it

SourceDoctor:
@djamp42 i implemented Device Dependency Graphing for Device Groups.
Hope it helps

SourceDoctor:
@Jellyfrog for me it's ready for Merge

SourceDoctor:
now implemented:

- redesign to blade/laravel
- filtering for Device Groups
- Device Highlighting

@Jellyfrog not sure if still something to do here (after 30 commits now ;) )
for my view it's ready to merge

djamp42:
@SourceDoctor yeah i just tried this, and it's freaking awesome!! thanks alot!!!  Been playing around with it and it works perfectly fine with device groups as far as i can see.

app/Http/Controllers/Maps/DeviceDependencyController.php
```diff
@@ -0,0 +1,136 @@
...
+                    $device_parents[] = $parent->device_id;
+            };
+
+            $childs = $device->children;
+            foreach ($childs as $child) {
+                    $device_childs[] = $child->device_id;
+            };
+        }
+
+        $devices_to_show = array_merge($device_nodes, array_merge($device_parents, $device_childs));
```
Jellyfrog: This selects \*ALL* devices and then filters them with a N+1 query.  
You could probably replace the entire foreach with:
```php
$devices = Device::inDeviceGroup($group_id)
  ->hasAccess($request->user())
  ->with([
    'location',
    'parents' => function ($query) use ($request) {
      $query->hasAccess($request->user());
    },
    'children' => function ($query) use ($request) {
      $query->hasAccess($request->user());
    }])
  ->get();

$devices = $devices->merge($devices->map->only('children', 'parents')->flatten());
``` 
If I'm not missing anything.

routes/web.php
```diff
@@ -29,6 +29,13 @@
     Route::get('about', 'AboutController@index');
     Route::get('authlog', 'UserController@authlog');
 
+    // Maps
+    Route::group(['prefix' => 'maps', 'namespace' => 'Maps'], function () {
+        Route::get('devicedependency', 'DeviceDependencyController@dependencyMap');
+        Route::get('devicedependency/group={group_id}', 'DeviceDependencyController@dependencyMap');
```
Jellyfrog: Lets either use query strings for this, or maybe better, create a new function in your controller to handle this.  
One for normal dependency, one for group, and one function that renders the map.
```php
Route::get('devicedependency/group/{group}', 'DeviceDependencyController@groupDependency');
```

```php
function groupDependency(Request $request, DeviceGroup $group) {
  // get devices...
  return self::renderMap($devices);
}
```

routes/web.php
```diff
@@ -29,6 +29,13 @@
...
+        Route::get('devicedependency/group={group_id}/highlight_node={highlight_node}', 'DeviceDependencyController@dependencyMap');
```
Jellyfrog: This should be removed and use `$request->get('highlight_node')` instead

SourceDoctor: @Jellyfrog 
tried to remove both lines with variables
and in `DeviceDependencyController` changed 

```
public function dependencyMap(Request $request, $group_id = 0, $highlight_node = 0)
{
```
to
```
    public function dependencyMap(Request $request)
    {
    $group_id = $request->get('group_id') | 0;
    $highlight_node = $request->get('highlight_node') | 0;
```
but with this everytime GET arguments are in URL Page fall's back to Main Page

some Ideas?

SourceDoctor: @Jellyfrog 
Code changed

app/Http/Controllers/Maps/DeviceDependencyController.php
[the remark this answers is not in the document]
SourceDoctor: changed

routes/web.php
SourceDoctor: group and highlight_node are now classical get Parameter

app/Http/Controllers/Maps/DeviceDependencyController.php
```diff
@@ -0,0 +1,108 @@
...
+
+use App\Models\Device;
+use Illuminate\Http\Request;
+use LibreNMS\Util\Url;
+
+class DeviceDependencyController extends MapController
+{
+    protected static function deviceList($request) {
+
+        $group_id = $request->get('group') | 0;
```
Jellyfrog: `|` is a bitwise operator, you should use `||` if you want an `or`.   
In this case tho, You could just remove it

SourceDoctor: oh, yes you're right. :-)
i changed it to "||"

SourceDoctor:
@Jellyfrog better now? :-)

app/Http/Controllers/Maps/DeviceDependencyController.php
```diff
@@ -0,0 +1,107 @@
...
+                'location',
+                'parents' => function ($query) use ($request) {
+                    $query->hasAccess($request->user());
+                },
+                'children' => function ($query) use ($request) {
+                    $query->hasAccess($request->user());
+                }])
+            ->get();
+
+        return $devices->merge($devices->map->only('children', 'parents')->flatten());
```
Jellyfrog: add `->loadMissing('parents', 'location');` to this

SourceDoctor: @Jellyfrog 
Don't know how you want me to implement this.
Everywhere and with every combination of `parent`, `location`, `childrens` some Devices or Relationgraphs are missing.

Jellyfrog: Sorry it needs to be added after so it loads the data missing after the merge

SourceDoctor: do you mean to change it to:
`return $devices->merge($devices->map->only('children', 'parents')->loadMissing('parents', 'location')->flatten());`
This doesn't work at me. It throws following Error:

> production.ERROR: Method Illuminate\Support\Collection::loadMissing does not exist. {"userId":2,"exception":"[object] (BadMethodCallException(code: 0): Method Illuminate\\Support\\Collection::loadMissing does not exist. at /opt/librenms/vendor/laravel/framework/src/Illuminate/Support/Traits/Macroable.php:104)
[stacktrace]
#0 /opt/librenms/app/Http/Controllers/Maps/DeviceDependencyController.php(54): Illuminate\\Support\\Collection->__call('loadMissing', Array)
#1 /opt/librenms/app/Http/Controllers/Maps/DeviceDependencyController.php(68): App\\Http\\Controllers\\Maps\\DeviceDependencyController::deviceList(Object(Illuminate\\Http\\Request))
#2 [internal function]: App\\Http\\Controllers\\Maps\\DeviceDependencyController->dependencyMap(Object(Illuminate\\Http\\Request))
#3 /opt/librenms/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(54): call_user_func_array(Array, Array)

SourceDoctor: @Jellyfrog 
i changed it now to:
`return $devices->merge($devices->map->only('children', 'parents')->flatten())->loadMissing('parents', 'location');`
Hope it's right now :-)

murrant:

https://community.librenms.org/t/daily-sh-not-running-automatically/9377/7

murrant:

PR #10930, opened 2019-12-11
Fix raisecom fan speed sensor limits

vitalisator:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/definitions/discovery/raisecom.yaml
```diff
@@ -18,6 +18,8 @@ modules:
                     num_oid: '.1.3.6.1.4.1.8886.1.1.5.2.2.1.2.{{ $index }}'
                     snmp_flags: ['-OeQUs', '-Pu'] # workaround for underscores in mib
                     descr: 'Fan {{ $index }} Speed'
+                    low_limit: 0
```
PipoCanaja: Are you sure that a limit of 0 is good idea ? it probably means that the fan is not running at all ... 
I would put something like 100, at least to ensure the fan is not stopped  ?

PipoCanaja: And most importantly, have you checked that the MIB does not provide any values for those limits (that would be quite unusual to provide fan min/max but Xmas is coming, let's dream :) )

vitalisator: yes, sure :-) the raisecom os has different fan speed level:
SpdLevelIndex   FanDueSpeed(r/min)  TemperScale         
--------------------------------------------------------
1               0                   47                  
2               4400                57                  
3               7700                67                  
4               9350                90                  

so 0 rpm is a legitime value if the temperature is below 47°C.
Unfortunately that is not documented in mib, but one can see that in cli and docs.

The Fan failures are detected via Fan state sensor.

PipoCanaja: OK.

murrant:

PR #10943, opened 2019-12-15
Highlight Nodes on Network map

SourceDoctor:
Choose Node in Select Box and it's highlighted on Traffic Network Map

image

#### Please note

#### Testers


includes/html/print-map.inc.php
```diff
@@ -315,10 +334,25 @@
 $nodes = json_encode(array_values($devices_by_id));
 $edges = json_encode($links);
 
+array_multisort(array_column($devices_by_id, 'label'), SORT_ASC, $devices_by_id);
+
 if (count($devices_by_id) > 1 && count($links) > 0) {
 ?>
```
Jellyfrog: Maybe it should be a GET request so you can link to it

SourceDoctor: @Jellyfrog 
good idea, i'll change to it.
I'll adapt this to #10916 also, if it's merged :-)

SourceDoctor: @Jellyfrog 
on select box change URL not changing.
But on setting: `/map/highlight_node=<device_id>` Node is highlighted

includes/html/print-map.inc.php
```diff
@@ -315,10 +334,25 @@
...
 
+<form name="printmapform" method="post" action="" class="form-horizontal" role="form">
```
Jellyfrog: Becuse this needs to be a GET?

SourceDoctor: if i change it to GET:
URL will be: `/map?highlight_node=<device_id>`. In classical PHP this is valid. Here it doesn't work.
It has to be changed to `/map/highlight_node=<device_id>` which works well

SourceDoctor: don't know what to do to get it to this here

SourceDoctor: @Jellyfrog some ideas?

SourceDoctor: is solved it with javascript. Maybe later a better way will be implemented :-)

SourceDoctor:
@Jellyfrog something else to do here?

murrant:

PR #10949, opened 2019-12-16
Added sensors for outlets on Schleifenbauer devices.

jviersel:
Added sensors for outlets on Schleifenbauer devices. Also use the outlet name when configured.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


jviersel:
I do understand why codeclimate reports about "Similar blocks of code", but I don't think the readability of the code will improve if we merge the two loops into one.

Jellyfrog:
> I do understand why codeclimate reports about "Similar blocks of code", but I don't think the readability of the code will improve if we merge the two loops into one.

You can ignore codeclimate. However you do need to fix the errors travis reports.

martijn-schmidt:
Hello @jviersel - nice to see another #nlnog'er here. I have some feedback for the code as the author of PR #9471 which added initial Schleifenbauer support in LibreNMS.

Unfortunately Schleifenbauers deployed in a databus topology are not very fast when it comes to responding to SNMP requests, this is a hardware limitation. If multiple SNMP requests are made at the same time the device may start timing out until the daemon/databus recovers and that causes an incomplete discovery, potentially deleting sensors in the process.

This is not a problem in standalone Schleifenbauers where every PDU has its own ethernet uplink, but if you are discovering a ring of let's say 50 devices times 48 outlets then that's going to add a lot of extra load on the databus. In my original implementation I worked around this by only enumerating data about individual outlets `if ($sdbMgmtStsDevices == 1)` both during the [pre-cache stage](https://github.com/librenms/librenms/blob/master/includes/discovery/entity-physical/schleifenbauer.inc.php#L38) and the [discover_entity_physical stage](https://github.com/librenms/librenms/blob/master/includes/discovery/entity-physical/schleifenbauer.inc.php#L84) of the entity-physical discovery. Could you please do something similar here so that the LibreNMS will continue to function in environments where Schleifenbauers are deployed in a databus topology?

Could you please add some code in `includes/discovery/entity-physical/schleifenbauer.inc.php` to detect the relevant sensors in the entity-physical discovery when appropriate so that the inventory page remains fully featured?

I noticed that the calculated `$entPhysicalIndex` values which you are using for the outlets is conflicting with the namespace used for the inputs, you'd be able to find information about the correct namespace for outlets in the [entity-physical discovery](https://github.com/librenms/librenms/blob/master/includes/discovery/entity-physical/schleifenbauer.inc.php#L94) codebase as per the comment in the sensor discovery code.

Finally, although your implementation for `$descr` works perfectly well I think it might be cleaner to use a shorthand ternary operator there: https://stitcher.io/blog/shorthand-comparisons-in-php

All the best,
Martijn

jviersel:
Thank you all for the feedback. I'll look into this, but due to the holiday season it might take a little longer.

jviersel:
Although the tests are fine now (excluding some duplicate code warnings), the code needs some optimization as @martijn-schmidt stated. Please do not merge this into master yet.

jviersel:
Code should now be optimized. Maybe @martijn-schmidt can share his opinion on this?

martijn-schmidt:
Hey @jviersel - I see that you're always discovering the outlets for the master now, regardless of whether or not it is part of a ring, and ignoring any other units in the ring - I'd recommend going one step further and putting an `if ($sdbMgmtStsDevices == 1)` to work both during pre-cache and the inventory discovery so you wouldn't even attempt to snmpwalk the outlets if there is more than one device on the bus.

Could you please add some code in `includes/discovery/entity-physical/schleifenbauer.inc.php` to detect the relevant sensors in the entity-physical discovery when appropriate so that the inventory page remains fully featured?

I noticed that the calculated `$entPhysicalIndex` values which you are using for the outlets are conflicting with the namespace used for the inputs, you'd be able to find information about the correct namespace for outlets in the entity-physical discovery codebase as per the comment in the sensor discovery code.

Finally, although your implementation for `$descr` works perfectly well I think it might be cleaner to use a shorthand ternary operator there: https://stitcher.io/blog/shorthand-comparisons-in-php

All the best,
Martijn

jviersel:
Hi @martijn-schmidt . Yes, I indeed discover all outlets from the device, regardless of whether or not it is part of a ring. As you can see, I've added the devices ID to the end of the SNMP OID that is used for dicovery, so it should only discover outlets of the local device, no all devices on the ring.

The reason I've chosen this behavior is that we plan to use these devices with both a connection to the network and a ring (so you are to reboot the network when the network is down). Using your approach, no sensors would be discovered when the devices is connected to a ring.

I'll have a look at the other issues.

Jellyfrog:
The test data doesn't look correct.

jviersel:
> The test data doesn't look correct.

Can you please provide some additional information? What problem did experience? Any output of the failed test? Without any information, nobody will be able to solve the problem.

tests/data/schleifenbauer.json
```diff
@@ -315,7 +315,7 @@
                     "sensor_oid": ".1.3.6.1.4.1.31034.12.1.1.2.6.1.1.5.210.1",
                     "sensor_index": "SVNL00056319-L1",
                     "sensor_type": "schleifenbauer",
-                    "sensor_descr": "SVNL00056319-L1 RMS Current",
```
Jellyfrog, line 318: This doesn't look correct, this looks like a better name than `t=master`

jviersel:
The change involves using the sensor's description from the device (when available) instead of a default name. In the SNMP-data used for the tests, the description for this sensor is "t=master" so the output for sensor_desc is now matching the result.

Yes, "SVNL00056319-L1 RMS Current" probably looks much nicer, but it's up to the devices administrator to set a better description for this sensor. The test data is not incorrect because the name isn't pretty; the file must contain the output we expect the code to return.

Jellyfrog:
My bad then

murrant:

PR #10954, opened 2019-12-17
Add alert rule option to invert devices and groups "map to" list

louis-oui:
Current implementation :
image

Add this option :
image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/AlertUtil.php
+++ b/LibreNMS/Alert/AlertUtil.php
@@ -181,7 +181,7 @@ class AlertUtil
         WHERE a.disabled = 0 AND (
             (d.device_id IS NULL AND g.group_id IS NULL)
             OR (a.invert_map = 0 AND (d.device_id=? OR dg.device_id=?))
-            OR (a.invert_map = 1 AND ? NOT IN (SELECT d.device_id) AND ? NOT IN (SELECT dg.device_id))
+            OR (a.invert_map = 1 AND (d.device_id != ? OR d.device_id IS NULL) AND (dg.device_id != ? OR dg.device_id IS NULL))
         )";
 
         $params = [$device_id, $device_id, $device_id, $device_id, $device_id, $device_id];
--- a/LibreNMS/Alert/AlertUtil.php
+++ b/LibreNMS/Alert/AlertUtil.php
@@ -181,7 +181,7 @@ class AlertUtil
         WHERE a.disabled = 0 AND (
             (d.device_id IS NULL AND g.group_id IS NULL)
             OR (a.invert_map = 0 AND (d.device_id=? OR dg.device_id=?))
-            OR (a.invert_map = 1 AND (d.device_id != ? OR d.device_id IS NULL) AND (dg.device_id != ? OR dg.device_id IS NULL))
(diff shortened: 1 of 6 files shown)


louis-oui:
Code climate errors due to laravel and needs to be bypassed

louis-oui:
@PipoCanaja Feature is ready
Before merging, we need to make sur that date in file is last one in list database/migrations/2019_12_17_151314_add_invert_map_to_alert_rules.php 
Otherwise, migration won't work

louis-oui:
Bug invert not accepted at rule creation

PipoCanaja:
> Before merging, we need to make sur that date in file is last one in list database/migrations/2019_12_17_151314_add_invert_map_to_alert_rules.php
> Otherwise, migration won't work

Hi
I don't get why this date would have an impact, unless we have another change touching the same table and the 2 are touching the same columns with one being a prerequisite for the other, I would say.

louis-oui:
> Hi
> I don't get why this date would have an impact, unless we have another change touching the same table and the 2 are touching the same columns with one being a prerequisite for the other, I would say.

From what I have seen, Laravel migration process refuses to update schema if a new schema file is inserted and its date in filename is previous to last update

louis-oui:
> From what I have seen, Laravel migration process refuses to update schema if a new schema file is inserted and its date in filename is previous to last update

Seems my test was wrong for some reason. It is OK

murrant:

https://community.librenms.org/t/add-alert-rule-option-to-invert-devices-and-groups-map-to-list/10635/1

SourceDoctor:
@louis-oui 
this PR results on my production system into false Alerts:
Example Alert Rule:  Port 5 down Switch X

i got Alert for Port 5 down, but on Switch Y

rolling back this PR and everything is fine again

vitalisator:
same problem here. Alert device mapping to Device do not work anymore. Please revert first.

louis-oui:
@vitalisator @SourceDoctor sorry for the issue. It is working on my production environnent. I am checking what is going on my dev.

murrant:

https://community.librenms.org/t/device-specific-alerts-applying-to-all-devices-since-jan-15th-update/10649/7

louis-oui:
@vitalisator @SourceDoctor, It seems I have reproduced your issue. Alert rules was bugged when you devices on map to field. My production had only groups.

PR11038 add again the feature and should fix the issue. Could you test it works for you ? 
```
./scripts/github-apply 11038
```
Get an ID of a device not in map to. Do:
`./poller.php -h ID -m ports`

At the bottom of the output, you will have the matching rules. Check that the rules that wrongly match because of the bug does not match again.
```
### Start Alerts ####
Rule #1 (Not 7): Status: NOCHG
Rule #2 (All): Status: NOCHG
Rule #3 (Mapto 5,7,8): Status: NOCHG
Rule #5 (Not all): Status: NOCHG
Rule #7 (5): Status: NOCHG
Rule #8 (Not 1): Status: NOCHG
Rule #9 (1): Status: NOCHG
#### End Alerts ####
```
Revert after test:
```
./scripts/github-remove -d
```

SourceDoctor:
@louis-oui 
looks good
`Status: NOCHG` on all rules on all Switches
Thx for fixing :-)

murrant:

PR #10971, opened 2019-12-23
Fix ignore alerts has no effect

louis-oui:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/AlertRules.php
+++ b/LibreNMS/Alert/AlertRules.php
@@ -47,6 +47,11 @@ class AlertRules
             echo "Under Maintenance, Skipping alerts.\r\n";
             return false;
         }
+        //Check to see if alert ignore is set
+        if (!(AlertUtil::hasAlertIgnore($device_id))) {
+            echo "Alert ignore is set, Skipping alerts.\r\n";
+            return false;
+        }
         //Checks each rule.
         foreach (AlertUtil::getRules($device_id) as $rule) {
             c_echo('Rule %p#'.$rule['id'].' (' . $rule['name'] . '):%n ');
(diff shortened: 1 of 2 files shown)


PipoCanaja:
Alert ignore is right now done with macros, which gives flexibility (You can still create alerts that triggers on and "ignored" device. I suggest to open a post on the community forum and get some attention on this subject, to see if we go in this direction.

louis-oui:
> Alert ignore is right now done with macros

What do you mean by macros ? Is it ignoring alert rule if "AND devices.ignore = 0" on the rule ?

This behaviour is misleading

louis-oui:
We can rename current label "ignore alerts" to "tag for ignore alerts" and add a new settings "ignore all alerts"

murrant:

https://community.librenms.org/t/ignore-alert-on-device-settings-is-misleading/10501/1

louis-oui:
Hi @PipoCanaja, there is no comment on the community ? What do you think of that ?
We can rename current label "ignore alerts" to "ignore tag" and add a new settings "Disable alerts"

PipoCanaja:
This change has a big impact on the way alerting is done, we cannot just do it between Xmas and NYE without concertation, and without notice. Let's put this on hold for now.

louis-oui:
> This change has a big impact on the way alerting is done, we cannot just do it between Xmas and NYE without concertation, and without notice. Let's put this on hold for now.

Thank you for your response. I understand. That's why I suggest to keep "ignore" behaviour as it is and add a new "disable alert" setting.

Let's keep it in hold for now

louis-oui:
Replaced by https://github.com/librenms/librenms/pull/11011

Just need a wording and section change

PR #10981, opened 2019-12-28
Alternate Poller IP instead of Hostname

SourceDoctor:
Set Alternate IP to use for polling instead of resolved Hostname
If set, alternate IP will be used for polling instead of Hostname
useful for:
- monitoring Devices via other IP's
- preparing preproductive Devices in DevelopmentNetworks
- configure fix IP for loadbalanced Devices
- ...

It can be used wie IPv4 and IPv6

image

I hope to get all points where polling against a Device runs.

#### Please note

#### Testers


kkrumm1:
If this was added we would need documentation on how and what this supposed to do.

SourceDoctor:
@kkrumm1 
do you mean a Documentation file or something like to be filled?
Or do you need more Description than in initial message?

kkrumm1:
Like documentation in the doc for users to understand what this dose. I'm sure users will have questions and we need something to help explain what this dose.

SourceDoctor:
@kkrumm1 
which doc file i have to fill with? :-)

SourceDoctor:
@kkrumm1 
docs and Image updated :-)

SourceDoctor:
@kkrumm1 
test data defective?
only docs changed and Travis is failing?

kkrumm1:
Ok it’s been over 29 days and nobody had made a pep about thisLGTM

murrant:

louis-oui:
Hello @SourceDoctor 

same db request is called many times for the same polling ? Could you optimise this ?

Example : one devices polling. Called 213 times.
```
$ ./poller.php -h 88 -d &>/dev/stdout | grep overwrite_ip
DB Schema: 2019_12_28_180000_add_overwrite_ip_to_devices (156)
SQL[select `hostname`, `overwrite_ip` from `devices` where `hostname` = ? limit 1 ["hostname"] 0.27ms] 
SQL[select `hostname`, `overwrite_ip` from `devices` where `hostname` = ? limit 1 ["hostname"] 0.38ms] 
SQL[select `hostname`, `overwrite_ip` from `devices` where `hostname` = ? limit 1 ["hostname"] 0.44ms] 
SQL[select `hostname`, `overwrite_ip` from `devices` where `hostname` = ? limit 1 ["hostname"] 0.42ms] 
SQL[select `hostname`, `overwrite_ip` from `devices` where `hostname` = ? limit 1 ["hostname"] 0.47ms] 
SQL[select `hostname`, `overwrite_ip` from `devices` where `hostname` = ? limit 1 ["hostname"] 0.43ms] 
SQL[select `hostname`, `overwrite_ip` from `devices` where `hostname` = ? limit 1 ["hostname"] 0.44ms] 
SQL[select `hostname`, `overwrite_ip` from `devices` where `hostname` = ? limit 1 ["hostname"] 0.58ms] 
SQL[select `hostname`, `overwrite_ip` from `devices` where `hostname` = ? limit 1 ["hostname"] 0.46ms] 
^C
$ ./poller.php -h 88 -d &>/dev/stdout | grep overwrite_ip | wc -l
213

```

SourceDoctor:
@louis-oui 
thanks for the tip.
Does this happens directly in poller.php? Could you specify more precise where this happens?
I'm a little bit confused why exactly this PR should cause this issue, because it only enhanced existing sql statements and some functions.

louis-oui:
The trace comes from /opt/librenms/poller.php. Option '-d' debug says that following request is done many times :
`select `hostname`, `overwrite_ip` from `devices` where `hostname` = "hostname" limit 1;`
The only issue is performance. It requests many call in db to get the same value.

This comes from many calls to this function in app/Models/Device.php
```
    public static function pollerTarget($hostname)
    {
        $ret = static::where('hostname', $hostname)->first(['hostname', 'overwrite_ip']);
        if (empty($ret)) {
            return $hostname;
        }
        $_overwrite_ip = $ret->overwrite_ip;
        $_hostname = $ret->hostname;

        return $_overwrite_ip ?: $_hostname;
    }
```
As far as I know, that function was not existing before this PR. In other word, the issue were not present before.

Could you please optimise your code to set overwrite_ip in a var and not call the function too many times ?

SourceDoctor:
See what you mean. 
I'll take a look on it. Thanks.

PR #10982, opened 2019-12-29
Add Lancom OAP-321 Wireless AP

vitalisator:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/OS/Lcos.php
```diff
@@ -0,0 +1,313 @@
...
+        $data = snmpwalk_cache_oid($this->getDevice(), 'lcsStatusWlanRadiosEntryRadioChannel', [], 'LCOS-MIB');
+        $radios = $this->getCacheByIndex('lcsStatusWlanRadiosEntryIfc', 'LCOS-MIB');
+
+        $sensors = [];
+        foreach ($data as $index => $entry) {
+            $radio = $radios[$index];
+            if (isset($sensors[$radio])) {
+                continue;
+            }
+            for ($i = 0, $j = strlen($index); $i < $j; $i++) {
```
PipoCanaja: Could you please put the conversion of the index to decimal into a function that would be called from each discovery function, instead of copy/pasting it ?

Thanx

vitalisator: thank you for the hint, done.

LibreNMS/OS/Lcos.php
```diff
@@ -0,0 +1,308 @@
...
+        return implode('.', $dec_index);
+    }
+
+    /**
+     * Convert MAC String to decimal encoded string notation
+     *
+     * @param colon separated string (MAC address notation)
+     * @return decimal encoded OID string
+     */
+    private function macToDecOid($mac)
```
PipoCanaja: I think this one already exists somewhere in LibreNMS ... don't remember where at all, but I saw it (and used it) before.

vitalisator: found some places that uses similar code, but nowhere a function for that.
Maybe this is the right time to do this now, I'm not sure if a protected function in LibreNMS/OS.php or a just drop it to includes/functions.php would be the right way.
Or we leave that actually as it. :-)

PipoCanaja: Found it : 
```
/opt/librenms/LibreNMS/Util/Rewrite.php:    public static function oidMac($mac)
```
So you could use this one and remove your new Implementation. Rewrite.php is where this kind of function is added in the newer architecture. includes/functions.php will eventually be deprecated.

PipoCanaja:
LGTM

murrant:

PR #10985, opened 2020-01-01
Fixed Teltonika Rutos2xx Wirelless sensors

jozefrebjak:
Fixed Teltonika Wireless Sensors SINR, RSRP and added new RSRQ sensor.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Are you aware that it will break historical data ? for the sensors that change from SNR to SINR ?

jozefrebjak:
> Are you aware that it will break historical data ? for the sensors that change from SNR to SINR ? (historical data is only 2 month old cause this code was merged in late october I would say, but may be a notification would be a good idea here)

We can notificate, but i don't know if there is an users which are using Teltonika Rutos2xx devices, that data are available through the cloud of Teltonika also, so I don't care about historical data and I don't need them. I edited it because I want to have corrected that sensors for future. What type of notification are you thinking ? In community or discord ?

PipoCanaja:
You can add a notification in "misc/notifications.rss" with a link to this PR. If somebody wants historical data, we can tell them which RRD file they need to rename to achieve it.

PR #11001, opened 2020-01-05
rrdtool.inc.php: in rrdtool_escape(), fix invalid length

josephtingiris:
For graph description lengths < 12 (i.e. 'Load %'), rrdtool_escape() produces ...

Warning: substr_count(): Invalid length value in /opt/librenms/includes/rrdtool.inc.php on line 281

... as per ...

https://www.php.net/manual/en/function.substr-count.php

This patch will calculate a separate length of $string for use with substr_count().  The original $length is still used for str_pad().


murrant:

PR #11003, opened 2020-01-05
Bugfix the snmpsim collector: use the correct mibdir

martijn-schmidt:
The captureFromDevice() function was only able to return data for named oids when the MIB lived in the root mibdir because the mibdir parameter was omitted from its snmp function call - as such it couldn't retrieve data from most vendor-specific MIBs, which generally live in sub-directories. The missing MIB reference meant that snmp commands for the snmpsim data collector didn't return usable data, and as a result test data was incomplete for many discovery components.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 11003`


murrant:

PR #11006, opened 2020-01-07
Add host even if snmphost name not available

brianegge:
Dahau cameras either don't have this property or it returns 'none'. In either case, I can't change it. I'd prefer not to force add the host as I want the snmp check to complete.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -759,7 +759,7 @@ function createHost(
         $device['os'] = getHostOS($device);
 
         $snmphost = snmp_get($device, "sysName.0", "-Oqv", "SNMPv2-MIB");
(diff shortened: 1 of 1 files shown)


PipoCanaja:
Hello @brianegge 
We will not be able to merge this change like this. We must find a solution that correct your issue with this device without changing the way LibreNMS works.
Most probably, the idea would be to add a property in the YAML file of the OS, where you could say "ignore this check" for this particular OS.

kkrumm1:
Closing this for now. Feel free to re open If you have new information or another way.

PR #11009, opened 2020-01-08
Fixed ports_vlans bulk delete from crashing MySQL

laf:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

This should fix: https://community.librenms.org/t/vlan-discovery-crashes-mysql-database-after-update-to-v1-43/5394

We've been seeing the same thing where our MySQL instance was restarting twice an hour. So far, 1 hour later it's been stable. Will report back if we see issues.

I chose 50 as an arbitrary number. I couldn't find any data on max size for queries other than packet size which is a per install. We can increase / decrease that number if need to.

### DIFF
--- a/includes/discovery/vlans.inc.php
+++ b/includes/discovery/vlans.inc.php
@@ -91,8 +91,11 @@ foreach ($vlans_db as $domain_id => $vlans) {
 
 // remove non-existent port-vlan mappings
 if (!empty($valid_vlan_port)) {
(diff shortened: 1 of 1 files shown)


murrant:

https://community.librenms.org/t/vlan-discovery-crashes-mysql-database-after-update-to-v1-43/5394/4

laf:
Bad patch, ignore :)

PR #11012, opened 2020-01-08
ajax_search.php: Fix & amend broken $num_ports query on line 83

josephtingiris:
* When searching from the web interface, librenms produces
production.ERROR SQLSTATE: Column not found: 1054 Unknown column
'$perms_sql' in 'where clause'

* Above feeds resources/views/layouts/menu.blade.php which, because
of the failed query, has no device_ports as a value. Thus the search
results do not produce the expected 'name ... device with X port(s)'

* As well, the previous query did not exclude deleted and ignored
ports. Therefore, the value for X port(s) was incorrect and did not
correspond to the device's ports screen.

### DIFF
--- a/html/ajax_search.php
+++ b/html/ajax_search.php
@@ -80,7 +80,7 @@ if (isset($_REQUEST['search'])) {
                         $highlight_colour = '#008000';
                     }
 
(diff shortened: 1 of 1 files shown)


josephtingiris:
wrong user.name & user.email

PR #11019, opened 2020-01-09
mix-manifest.json: Updated to match current language files

josephtingiris:
* The mix-manifest.json does not match what's generated by lnms. When
new translations are generated, it causes validate to report a mismatch.


josephtingiris:
There is a directly related discussion about this PR in https://github.com/librenms/librenms/pull/11020

I've double checked.  The `mix-manfiest.json` in master is still out of sync.

https://github.com/librenms/librenms/pull/11020#issuecomment-576272041

SourceDoctor:
@josephtingiris 
mix-manifest.json will be updated while generating language file updates.
You've only an update of mix-manifest.json if this file didn't match, did you do some corrections on languages files and forgot them to add to PR?

josephtingiris:
This is more complicated.  @Jellyfrog and I were working on it last week.  mix-manifest generates differently on different machines.  That will be addressed in a different a PR.  This one can be closed.

PR #11024, opened 2020-01-10
correct and cleanup the mysql socket python code

wilreichert:
This PR fixes the mysql connect socket code and ensures constant behavior across all the python scripts that behaves as described in the  documentation - https://docs.librenms.org/Support/Configuration/.

1. if mysql is listening on a socket, you don't need to specify host or port in the connect string
2. setting `db_host` to a socket path and the port to 0 will still cause MySql-python to use TCP
3. usage of the `unix:` syntax in the host field is undocumented
4. The MySQL-python package has opinionated defaults that might warrant further documentation to avoid confusion
  - setting `localhost` as the `db_host` will result in a socket connection
  - setting `127.0.0.1` as the `db_host` will result in a TCP connection
  - setting `db_socket` will result in a socket connection even if `db_host` is set

config_to_json.php will default config[db_socket] to `null` if its not explicitly set, so it makes sense to use that as the tcp / socket connection determination.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/discovery-wrapper.py
+++ b/discovery-wrapper.py
@@ -89,22 +89,20 @@ discovery_path = config['install_dir'] + '/discovery.php'
 log_dir = config['log_dir']
 db_username = config['db_user']
 db_password = config['db_pass']
-db_port = int(config['db_port'])
 
 if config['db_socket']:
-    db_server = config['db_host']
     db_socket = config['db_socket']
 else:
     db_server = config['db_host']
-    db_socket = None
+    db_port = int(config['db_port'])
 
 db_dbname = config['db_name']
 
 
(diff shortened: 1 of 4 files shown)


Jellyfrog:
You should look into https://github.com/librenms/librenms/pull/10759

SourceDoctor:
@wilreichert 
nice on first look,
would it be possible to outsource DB Connection into a library to reduce code redundancy?

murrant:
Thanks.

The mysql client will always connect to the socket if it is available.  Also, this code will need to be updated to work with the new python3 code anyway.

PR #11036, opened 2020-01-16
Revert "Add alert rule option to invert devices and groups "map to" list"

kkrumm1:
Reverts librenms/librenms#10954


PipoCanaja:
Not so good in fact cause the migration is not "reverted" ...

PipoCanaja:
I dont know what would be the correct way of handling this in fact.

kkrumm1:
@louis-oui  

Can you check this out seems even after testing this is not working.

PR #11037, opened 2020-01-16
SNMP counters validation

ccperilla:
Validation step added before counters are updated in table 'bill_port_counters'.

Avoid wrong data to be added to the table in case the SNMP polling function return an invalid value.

check --> https://community.librenms.org/t/wrong-billing-record-in-delta/8540

Scenarios encountered (No1 resolved - 2 and 3 to discuss)

1. If both SNMP polling queries fail (64/32) value returned is nor numeric --> is assumed as 0 --> this is overcome on the first calculation but the value is stored as a legit counter. The problem arises on the next calculation where the delta will be calculated against the 0 value… high delta as result.
Solution:
If both queries failed (64b and 32b) the value returned is FALSE. In this case conditional needs to be created in poll-billing.php to evaluate this condition and avoid DB update on 'bill_port_counters' table.

2.SNMP function used getValue() and snmp_get() uses RFC-2665 starting querying for 64counter and querying 32bit if the 64 fails --> Case No1 --> this is a problem when the current counter is over 2^32 (max counter value), in this case the counter (32bit) returned is not relevant…calculation would then give a big Delta.
Comment:
Dependent on 3 to add conditional and determine which query to use. --> In our case we use 64b counters only, we commented 32b query for the time being (function getValue())? --> Added a debug in this scenario to query the 64b again and try to capture the error returned by the snmp query --> No success.

3.Poll-billing.php --> Delta calculation does not discriminate if value obtained from counter was 32 or 64 --> this is a problem when the previous counter value (variable retained in the DB is over 2^32), in this case the counter (32bit) returned is not relevant…calculation would then give a big Delta.
Comment:
Modify getValue() function to receive port counter (64 or 32) and poll that one only. However, there is not seem to be currently a value on the tables that flag this case. This could be taken when the adding process is completed maybe?

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/poll-billing.php
+++ b/poll-billing.php
@@ -105,12 +105,11 @@ foreach ($query->get(['bill_id', 'bill_name']) as $bill) {
             if (dbUpdate($fields, 'bill_port_counters', "`port_id`='" . mres($port_id) . "' AND `bill_id`='$bill_id'") == 0) {
                 $fields['bill_id'] = $bill_id;
                 $fields['port_id'] = $port_id;
-
                 dbInsert($fields, 'bill_port_counters');
             }
         } else {
             echo "WATCH out! - Wrong counters. Table 'bill_port_counters' not updated\n";
-            logfile("WATCH out! - Wrong counters. Table 'bill_port_counters' not updated");         
+            logfile("WATCH out! - Wrong counters. Table 'bill_port_counters' not updated");
         }
         ////////////////////////////////EndCountersValidation&DB-Update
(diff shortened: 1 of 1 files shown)


kkrumm1:
Let’s go ahead and merge the first the fix you made the 2nd and 3rd submit the code changes for that and we can help.

kkrumm1:
LGTM

murrant:

PR #11041, opened 2020-01-17
Don't overwrite real port ids with zeros in the FDB

cjwbath:
I have found occasionally that busy switches only return a partial dot1dBasePortTable when doing FDB discovery (includes/discovery/fdb-table/bridge.inc.php). This results in some FDB entries being given a port_id of zero and this gets written over any valid port IDs from previous runs in the database ports_fdb table.

This PR is to avoid doing that and just update the timestamp, on the assumption that the FDB entry is really still on the same port as before and this is just a transient issue reading the dot1dBasePortTable that will resolve soon.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/fdb-table.inc.php
+++ b/includes/discovery/fdb-table.inc.php
@@ -41,7 +41,7 @@ if (!empty($insert)) {
                 $new_port = $entry['port_id'];
                 $port_fdb_id = $existing_fdbs[$vlan_id][$mac_address_entry]['ports_fdb_id'];
 
-                // Sometimes new_port ends up as 0 if we didn't get a complete dot1dBasePort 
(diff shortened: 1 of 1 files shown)


PipoCanaja:
Any idea about the devices that are touched by this issue ?

PipoCanaja:
If the same issue occurs during the initial poll (1st time an entry is discovered) then the port_id will be 0 which is probably not good either ... an entry with 0 should be skipped as well during the insert phase I suppose.

cjwbath:
@PipoCanaja I've mostly seen it on HPE Comware switches with busy CPU. My main intention with this change was to avoid good FDB entries being made bad by this problem. Since a bad entry with zero port_id and no entry at all are pretty much equivalent, it's less important to stop them being inserted in the first place.

murrant:

PR #11049, opened 2020-01-19
Url.php: Allow findOsImage() to also use the first two words of $feature

josephtingiris:
* LibreNMS is not always using the bunded redhat.svg.
* Modern RHEL has /etc/os-release and distro ends up using it.
* The lsb release is correct, i.e. 'Red Hat Enterprise Linux Server 7.6'
* Previously, findOsImage() only considered the first word of $feature as a possibility for the icon.
* This patch also allows using the first two words (with the space removed)
* i.e. 'Red Hat' becomes 'redhat' and findOsImage uses the bundled redhat.svg

### DIFF
--- a/LibreNMS/Util/Url.php
+++ b/LibreNMS/Util/Url.php
@@ -382,7 +382,14 @@ class Url
 
         if ($os) {
             if ($os == "linux") {
+                // first, prefer the first word of $feature
                 $distro = Str::before(strtolower(trim($feature)), ' ');
+                $possibilities[] = "$distro.svg";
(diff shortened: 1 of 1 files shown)


kkrumm1:
Will this cause any issues with existing devices?

josephtingiris:
No.  It prefers the old logic.  It only takes it a step further when no icon is found.  It will fix icons for newer versions of Red Hat and allows for additional two word distros.

My setup has over 10k devices and I’ve been running this patch for over a week.  No issues.


murrant:

PR #11055, opened 2020-01-21
fdb-table.inc.php: Fix SQL constraint violation, 'port_id' cannot be null

josephtingiris:
* fix SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'port_id' cannot be null
* If $entry['port_id'] truly is null (on insert) then Illuminate throws a fatal error and all subsequent processing stops.
* Cisco ISO (and others) may have null ids. We still want them inserted as new
* strings work with DB::table->insert().


murrant:

PR #11059, opened 2020-01-21
.gitignore: Ignore some more commonly used files and directories

josephtingiris:
Add the following ignores to the repo root's .gitignore ... 

+arch*
+backup
+bin
+certificate
+etc
+httpd.conf*
+html/custom/*
+sbin
+tmp

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,6 @@ html/plugins/*
 !html/plugins/Test/
 includes/custom/*
 junk
-librenms-agent
 nbproject
 patches
 !/lib/yaml
--- a/.gitignore
+++ b/.gitignore
@@ -17,13 +17,21 @@ __pycache__
 *.pyc
 *.swp
 Thumbs.db
+arch*
+backup
+bin
+certificate
 config.php
+etc
+httpd.conf*
 html/css/custom/*
 html/images/custom/*
+html/custom/*
 html/plugins/*
 !html/plugins/Test/
 includes/custom/*
 junk
+librenms-agent
 nbproject
 patches
 !/lib/yaml
@@ -36,6 +44,8 @@ _ide_helper.php
 _ide_helper_models.php
 resources/views/menu/custom.blade.php
 resources/js/vue-i18n-locales.generated.js
+sbin
+tmp
 
 # Docs #
 ########


kkrumm1:
I don’t think we want to ignore librenms agent that is updated and maintained

josephtingiris:
librenms/librenms-agent is a separate repo.  That ignore is to ignore it if it’s cloned out at the same level as liberals/librenms, which I’m sure many of us do.


josephtingiris:
I removed librenms-agent.  Are the rest OK?

kkrumm1:
I think sober just a test and then we can GTM

josephtingiris:
TL;DR

The bottom line is that are a subset of (standard) sub-directories, like bin & sbin, that are very common and should be ignored because they're reserved for system administrators.   I believe we should follow the FHS whenever and wherever possible, because it works.  

...

This request is in the spirit of the Filesystem Hierarchy Standard and its guiding principles, which are defined here: http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch01.html

The use of /opt is a well-established practice.   The fact that LibreNMS documentation references && the default ***install_dir*** is **/opt/librenms**, implies respect for the FHS.

Most typically, in practice and throughout the FH standard,  ./bin, ./sbin, etc are reserved for local use and should be ignored.   The requirements for /opt and the rationale is stated here: http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s13.html

"_The directories /opt/bin, /opt/doc, /opt/include, /opt/info, /opt/lib, and /opt/man are reserved for local system administrator use. Packages may provide "front-end" files intended to be placed in (by linking or copying) these reserved directories by the local system administrator, but must function normally in the absence of these reserved directories._"

Yes, I know that example is referencing the second level hierarchy (/opt).  It's an example.  The same requirements and rationale extend from the root to the secondary, tertiary, and even deeper levels.  e.g. /opt/librenms/vendor/bin (and others).  

Having an automated deployment process that's package based and relies on standards is helpful.  It's more secure, reliable, and easier to have optional packages contained within the same hierarchy.  Everything should be in predictable locations.  

IMO, That's especially important for large installations and diverse infrastructure teams who all want to rely on a single NMS, like LibreNMS.  Some have many thousands of devices, exponentially more containers, and dozens of distributed pollers to manage.  I'm not hand jamming every install.  

In my case, and I'm sure others', I truly appreciate the value of standards.  There are environment specific bits I've added for deployments.  i.e. Custom service check binaries are installed installed in /opt/librenms/bin.  Environment, container, and architecture specific root binaries are in /opt/librenms/sbin, they may link to ../arch*, and misc configs are in etc.

LibreNMS is a great project, so I took the time to make this appeal.  I can, have, and will continue to live with the fact that validate warns because my local git contains modified files.  Meh.  I'll ignore that one way or another; grep -v is an IEEE standard.  :)  I understand some projects don't want to follow, care, or even know about the standards that LISAs expect.

I also regard cooperation as a benefit.  Post install, LibreNMS has over 140k files under /opt/librenms.  If, up front, application developers address the organization of the files and directories of their project then the distribution developers, package maintainers, and/or system administrators will be much more proficient (and happier).

Referring back to the purposes in chapter 1 ... '_for software **_and_** users to be able to predict the location of files and directories_' ... which was written more than 25 years ago.  ../tmp?  or ./junk?

I still think the main .gitignore needs to ignore even more.

What is the 1 change requested ?

Why _not_ bin, sbin, etc ??

Thanks for the feedback.

kkrumm1:
closing this for now. Jelly frog added his two cents but never responded.

Jellyfrog:
Yeah sorry.
I would also have closed it, dont clone a git repo into an existing folder structure.

PR #11063, opened 2020-01-23
Added OS definition and discovery for SIAE Alfo80HD

HostIRE:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


PipoCanaja:
Hi @HostIRE 
This OS already exists as sm-os, it matches the same sysObjectID. This PR would create a conflict, and it is not clear why you want to create a new one, if we already have one. Could you give us more details ? 
Thanx

HostIRE:
Hello there, even though it has the same sysObjectID it is actually a different OS, although it shares *some* features with the parent sm-os. This update adds the sensors for the wireless specific information for the SIAE Alfo 80HD platform, which isn't available under the existing sm-os suite. I have added an additional unique identifier specifically to discover the Alfo 80HD OS, so there *shouldn't* be a conflict, and we're using this new OS module extensively in production at the moment without a problem, both radio types (AlfoPlus and Alfo 80HD) are correctly detected and sensors for both working correctly.

As such I thought it would be a good idea to share our work with the community so everyone using this radio can benefit :)

PipoCanaja:
Is this "new" OS strictly used on this alfo 80hd family ? If in doubt, this os should be named in a more generic way that it is right now. May be "sm-alfo" for instance. 
And yes, adding support for this device is definitely a good idea, that is the all point of a community software. We just have to understand how this device is made, and how the vendor implemented the SNMP part of it, so we can make the best support in LibreNMS ...

HostIRE:
Yeah, I'm afraid this OS is unique to the Alfo80HD family, they have another device which is very close (The AlfoPlus2), but not close enough for these purposes, and it is almost completely different again to that for the AlfoPlus and don't get me started on the Alfo ALCPlus2CE!! Just take a look at the connectors they use and you'll understand (anybody hear of an M12 for ethernet???).

Anyway, it took us a LONG time to get meaningful information from these radios at all as the SNMP implementation is squiffy (editing the community name causes problems, if it is possible at all, for example), but this module makes monitoring them possible and (relatively) pain free...

PipoCanaja:
Hi @HostIRE 
Sounds good. Could you please generate again the JSON test data ? (this comes from a collision with another PR that was merged in between.)
With this done, we'll be ready to merge this code. 
Thanx

murrant:

PR #11068, opened 2020-01-25
Fixed device SNMP edit form (and better feedback)

josephtingiris:
* snmp.inc.php: fixed FIXME (need better feedback)
* Use Toastr for feedback, rather than print_message and print_error.  Normal messages will disappear after 5 seconds.  Errors will persist until closed.
* The feedback for Max Repeaters and Max OIDs is much better.  The constant false errors on save have been corrected.  They are no longer applicable and muted when SNMP is turned off. Their feedback now better shows the user what has been done.
* Rename & relabel $no_checks as $force_save (Force Save) because that's simpler and more clearly defines what the code is doing when it's turned on.
* Reorder the Force Save checkbox to be right above the Save button so it's even more clear what it does.  Force Save and Save are the only elements on the form that do *not* have a database setting.  They're now grouped together, and don't get separated if distributed poller is being used.
* More attention was given to security.  Issues existed when snmp v3 was being turned on.  Chrome would automatically fill authname and authpass with the user's login and password.  If the user wasn't paying attention then that would be saved to the database, even if the snmp v3 scan failed.  Now, nothing is saved to the database if the snmp scan fails (unless Force Save is turned on).  All $_POST and editing variables are unset after being used.
* Embedded comments about the use of === false as tests for setAttrib() and forgetAttrib()

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


kkrumm1:
LGTM

murrant:

PR #11097, opened 2020-02-03
Update rrdstep.php to use global rrd.heartbeat for all heartbeat including ping-perf

willhseitz:
Currently, when a device is polled for a certain module, and the RRD file doesn't not exist, the RRD file is created with heartbeat =  `Config::get(rrd.heartbeat)`, _regardless if it is ping-perf or not._
However, when ./scripts/rrdstep.php is ran, an RRD file for ping-perf is specifically set to use heartbeat = ( `Config::get(ping_rrd_step)` * 2).
This PR will update rrdstep.php to use the global rrd.heartbeat when updating any RRD, including ping-perf.

Here is the RrdDefinition->make() for reference:

>     /**
>      * Add a dataset to this definition.
>      * See https://oss.oetiker.ch/rrdtool/doc/rrdcreate.en.html for more information.
>      *
>      * @param string $name Textual name for this dataset. Must be [a-zA-Z0-9_], max length 19.
>      * @param string $type GAUGE | COUNTER | DERIVE | DCOUNTER | DDERIVE | ABSOLUTE.
>      * @param int $min Minimum allowed value.  null means undefined.
>      * @param int $max Maximum allowed value.  null means undefined.
>      * @param int $heartbeat Heartbeat for this dataset. Uses the global setting if null.
>      * @return RrdDefinition
>      */
>     public function addDataset($name, $type, $min = null, $max = null, $heartbeat = null)
>     {
>         if (empty($name)) {
>             d_echo("DS must be set to a non-empty string.");
>         }
> 
>         $ds = array();
>         $ds[] = $this->escapeName($name);
>         $ds[] = $this->checkType($type);
>         $ds[] = is_null($heartbeat) ? Config::get('rrd.heartbeat') : $heartbeat;
>         $ds[] = is_null($min) ? 'U' : $min;
>         $ds[] = is_null($max) ? 'U' : $max;
> 
>         $this->dataSets[] = $ds;
> 
>         return $this;
>     }
> 

#### Please note

#### Testers

### DIFF
--- a/scripts/rrdstep.php
+++ b/scripts/rrdstep.php
@@ -72,11 +72,10 @@ foreach ($files as $file) {
 
     if ($rrd_file == 'ping-perf') {
         $step = $icmp_step;
-        $heartbeat = $icmp_step * 2;
     } else {
         $step = $system_step;
-        $heartbeat = $system_heartbeat;
     }
+    $heartbeat = $system_heartbeat;
 
     $rrd_info = shell_exec("$rrdtool info $file");
     preg_match('/step = (\d+)/', $rrd_info, $step_matches);


laf:
Have you taken a look at https://github.com/librenms/librenms/pull/8821 which is the original source for this to make sure it wasn't intentional?

murrant:
This code was intentional.  Can you please explain why you think it is wrong?

murrant:
Closing for now, please let us now what you are trying to fix here.  Thanks.

PR #11114, opened 2020-02-07
Fix inventory page hostname/sysname and default generate_device_link behaviour

kedare:
Simplification of generate_device_link with the use of format_hostname.
Fix the inventory page to use use it.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/html/functions.inc.php
+++ b/includes/html/functions.inc.php
@@ -18,7 +18,8 @@ use LibreNMS\Config;
  * Compare $t with the value of $vars[$v], if that exists
  * @param string $v Name of the var to test
  * @param string $t Value to compare $vars[$v] to
- * @return boolean true, if values are the same, false if $vars[$v] is unset or values differ
+ * @return boolean true, if values are the same, false if $vars[$v] 
+ * is unset or values differ
  */
 function var_eq($v, $t)
 {
--- a/includes/html/functions.inc.php
+++ b/includes/html/functions.inc.php
@@ -175,9 +175,6 @@ function generate_device_link($device, $text = null, $vars = array(), $start = 0
     }
 
     $class = devclass($device);
-    if (!$text) {
-        $text = format_hostname($device['hostname']);
-    }
 
     $text = format_hostname($device, $text);
 
--- a/includes/html/functions.inc.php
+++ b/includes/html/functions.inc.php
@@ -18,7 +18,7 @@ use LibreNMS\Config;
  * Compare $t with the value of $vars[$v], if that exists
  * @param string $v Name of the var to test
  * @param string $t Value to compare $vars[$v] to
- * @return boolean true, if values are the same, false if $vars[$v] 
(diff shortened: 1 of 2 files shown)


PipoCanaja:
Rebased to master. Should pass tests now.

includes/html/functions.inc.php
```diff
@@ -175,7 +176,7 @@ function generate_device_link($device, $text = null, $vars = array(), $start = 0
 
     $class = devclass($device);
     if (!$text) {
-        $text = $device['hostname'];
+        $text = format_hostname($device['hostname']);
```
SourceDoctor: Did i miss something?
As i see $text will be formated directly after this if statement with
`$text = format_hostname($device, $text);`

includes/html/table/inventory.inc.php
```diff
@@ -61,11 +61,11 @@
     $sql .= " LIMIT $limit_low,$limit_high";
 }
 
-$sql = "SELECT `D`.`device_id` AS `device_id`, `D`.`hostname` AS `hostname`,`entPhysicalDescr` AS `description`, `entPhysicalName` AS `name`, `entPhysicalModelName` AS `model`, `entPhysicalSerialNum` AS `serial` $sql";
+$sql = "SELECT `D`.`device_id` AS `device_id`, `D`.`hostname` AS `hostname`, `D`.`sysName` AS `sysName`,`entPhysicalDescr` AS `description`, `entPhysicalName` AS `name`, `entPhysicalModelName` AS `model`, `entPhysicalSerialNum` AS `serial` $sql";
 
 foreach (dbFetchRows($sql, $param) as $invent) {
     $response[] = array(
-                   'hostname'    => generate_device_link($invent, shortHost($invent['hostname'])),
+                   'hostname'    => generate_device_link($invent),
```
SourceDoctor: to see complete hostname, instead of short variant of hostname you can change length oft shortend hostname for your needs in global settings here:
image

As you see, i extended the default value from 12 to 30 for myself to see complete FQDN.
So i don't see any need of changing code here.

PipoCanaja: This change is probably improving the behaviour with devices that are added by IP and where sysname is necessary. So indeed this one should be kept.

PipoCanaja: Tested it in a ip-hostname environnement. Works as expected.

includes/html/functions.inc.php
PipoCanaja: Fixed

PipoCanaja:
This patch was tested in an instance with devices added as IP, and format_hostname does what we expect from it now.

includes/html/table/inventory.inc.php
kedare: Indeed this is done to see the real hostname (from SNMP) instead of seeing the IP on an environment where all the devices are added by IP.

murrant:

PR #11117, opened 2020-02-07
Update devices last_ping along with device_perf

josephtingiris:
* As a user, I want to know when a device has long ping latency whether its SNMP is functioning or not.
* The device_perf timestamp is always updated regardless.  This leads to inconsistencies and confusion.  Why isn't devices.last_ping_timetaken updated every time too?
* As it is, last_ping timestamp for SNMP devices does not get updated when the device is 'down'. 
* If device_perf is inserted and the ping was successful then updating the devices last_ping timestamp is consistent.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2253,6 +2253,16 @@ function device_is_up($device, $record_perf = false)
         }
         $device_perf['debug'] = json_encode($trace_debug);
         dbInsert($device_perf, 'device_perf');
+
+        // if device_perf is inserted and the ping was successful then update device last_ping timestamp
+        if (!empty($ping_response['last_ping_timetaken']) && $ping_response['last_ping_timetaken'] != "0") {
+            dbUpdate(
+                array('last_ping' => NOW(), 'last_ping_timetaken' => $ping_response['last_ping_timetaken']),
+                'devices',
+                'device_id=?',
+                array($device['device_id'])
+            );
+        }
     }
     $response              = array();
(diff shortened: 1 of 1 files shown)


kkrumm1:
I see what you are saying. 

But why not just set up two alert rules one for ICMP and another for SNMP.

Screen Shot 2020-02-07 at 2 04 43 PM
and 
Screen Shot 2020-02-07 at 2 05 18 PM

josephtingiris:
Your way is way easier to detect when snmp is down, no doubt.  Thanks for sharing it.  I've added them to the alert_rules.json

Though that's not the only reason I wanted the last_ping timestamp and last_ping_timetaken to be correct.   Here's another use case ...

Regardless if snmp is up or down, how would I have a reliable rule like `devices.last_ping_timetaken is greater than 99` (i.e. so I can alert for long latency (even if snmp is down)?

josephtingiris:
I put those over here https://github.com/librenms/librenms/pull/11115/commits/b89a731f00be953610ca85c1601368023a11c93d

kkrumm1:
I get what you are saying. We should be good to merge after Travis is fixed. Thank you for added those alert rules to the collection.

PipoCanaja:
Fix fro travis is here : 
https://github.com/librenms/librenms/pull/11116
Should me merged soon.

murrant:

PR #11133, opened 2020-02-11
New device Cisco Air

mpikzink:
Add Cisco AIR-CAP2702,AP1142,2602 inclusive connected users.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


LibreNMS/OS/Ciscoair.php
```diff
@@ -0,0 +1,37 @@
+<?php
+/**
+ * Ios.php
```
laf: Can you update the comments to be Ciscoair.

includes/definitions/ciscoair.yaml
```diff
@@ -0,0 +1,50 @@
+os: ciscoair
+text: 'Cisco IOS'
```
laf: Do these devices run Cisco IOS?

mpikzink: Yes, it does.

#show version
Cisco IOS Software, C1140 Software (C1140-K9W7-M), Version 15.3(3)JD11, RELEASE SOFTWARE (fc1)

LibreNMS/OS/Ciscoair.php
mpikzink: Thanks for the information.

Jellyfrog:
@mpikzink the tests fails. Can you look into this please?

mpikzink:
@Jellyfrog Sorry for the delay. I will have a look at the fail.
@murrant Yes in pricipal it's ios but accesspoints and the ios class has no wireless discovery and poller so I wanted wo extend the ios class. Should I better add the poller to the ios class? I am not sure because wireless is not the typical thing for Cisco.

Jellyfrog:
Closing due to inactivity, feel free to reopen!

murrant:
@mpikzink It should be ios, check if the device is an air device, and then poll wireless  sensors.  see the Ios.php class

PR #11146, opened 2020-02-13
Fix bits graphs

josephtingiris:
This fixes a bug rendering bits graphs for devices with empty string values for device_traffic_desc. 

This is most often seen manifesting itself on the Overall Traffic graphs that are on the device's page.  The symptoms will be that Overall Traffic will not render at all (Error Drawing Graph) or produce a graph with no data (even when the indivudal port graphs work fine).

* That's because some interfaces are incorrectly ignored.
* They're incorrectly ignored when a device_traffic_descr has no value because Config::get() returned an empty string.
* With an empty string, preg_match('',$string) is always true, and that leads to false positives.
* The false positives lead to interfaces getting ignored when they should not be.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


josephtingiris:
This ...

image

josephtingiris:
Becomes this ...

image

includes/html/graphs/device/bits.inc.php
```diff
@@ -22,6 +22,9 @@
 
     if (is_array(\LibreNMS\Config::get('device_traffic_descr'))) {
         foreach (\LibreNMS\Config::get('device_traffic_descr') as $ifdescr) {
+            if (is_null($ifdesc) || !is_string($ifdesc) || $ifdesc == '') {
```
Jellyfrog: Shouldn't `empty()` be enough?

josephtingiris: I thought about it ... what if the ifdesc is a literal '0' ... empty would return true, another false positive.

josephtingiris:
@Jellyfrog is that an acceptable compromise?

LibreNMS/Util/StringHelpers.php
```diff
@@ -28,12 +28,23 @@
 class StringHelpers
 {
     /**
-     * Shorten text over 50 chars, if shortened, add ellipsis
+     * Check if $mixed is a true string value that's not empty
      *
-     * @param $string
-     * @param int $max
-     * @return string
```
mpikzink: This should not be deleted.

josephtingiris: Thanks.  You're right.

LibreNMS/Util/StringHelpers.php
```diff
@@ -28,7 +28,25 @@
 class StringHelpers
 {
     /**
-     * Shorten text over 50 chars, if shortened, add ellipsis
+     * Check if $mixed is a true string value that's not empty
+     *
+     * @param $mixed
+     * @return boolean
+     */
+    public static function emptyString($mixed)
```
murrant: Why do you have this complex function.  Isn't what you are looking for just `$string === ''`?

Jellyfrog:
When would this happen except for when the user adds a empty string to device_traffic_descr?

LibreNMS/Util/StringHelpers.php
josephtingiris: TL;DR

Yes, `$string === ''` is better than `empty($string) && $string != '0'` (but that's not enough).  I'll change it to that.

Why?

I was trying to compromise to get this PR merged.  I agree it doesn't need to be this complex, though don't agree empty() is the best solution because it would return true for '0'.

preg_match() only accepts strings and when $mixed or '' is passed it produces false positives.  

In LibreNMS, I've noticed there's an abundance of empty() and preg_match() without consideration for the actual return values.  I saw a StringHelper and thought it looked like a good place to have this conversation.

1) To avoid false positives, that are breaking the bits graphs (and others breaks involving the use of preg_match() without testing the return value), we should first test if the input is a string.  

2) Then, in other cases, a second test may needed to determine if the string has no value.  Again, primarily for preg_match() but may be usable elsewhere.

Either would return true, everything else is false.  That's the emptyString() method.

$string === '' alone wouldn't meet both conditions.  It would only match true if $mixed was a string.  For example, Array() would not be considered an empty string and would be false.

josephtingiris:
> When would this happen except for when the user adds a empty string to device_traffic_descr?

Or, what if I want to give my interfaces simple descriptions like 0, 1, 2, 3, etc?

murrant:
Yeah, most of the time the code doesn't care if the string is '', false, null, or 0... so empty() is fine.

If you only want empty string, use the strict comparison.  Sounds like you want this `isstring($s) && $s === ''`

josephtingiris:
> `!is_string($s) || $s === ''`

That's effectively the same thing, albeit with a PHP Notice for some conversions like Array().  What I'd put in emptyString() does that without any notices.  

That's pretty close to what I'd first committed.  Would you all rather I take out emptyString() and make each test independently, or can we go with this?

josephtingiris:
This illustrates the different approaches.

```
<?php

function _printf($counter, $type, $value, $not, $comparison) {
    printf("%s. %-10s %-31s %s %sa string with a value\n",
        $counter,
        "$type=".(is_array($value) ? count($value) : $value),
        $comparison,
        ($not === FALSE ? "IS" : "is"),
        ($not === FALSE ? "" : "NOT ")
    );
}

function emptyString($value)
{
    if (!is_string($value)) {
        return true;
    }
    if ($value === '') {
        return true;
    }
    return false;
}

$values=array();
$values[] = '0'; // this is the only true, non-empty string
$values[] = '';
$values[] = array();
$values[] = array(1,2,3);
$values[] = 0;
$values[] = NULL;
$values[] = TRUE;
$values[] = FALSE;

$counter=0;
foreach ($values as $value) {
    $counter++;
    $type=gettype($value);

    // meets objective
    $empty_string=emptyString($value);
    _printf($counter, $type, $value, $empty_string, "emptyString()");

    // meets objective
    ((!is_string($value) || $value === '') ? $empty_string=true : $empty_string=false);
    _printf($counter, $type, $value, $empty_string, "!is_string() || string === ''");

    // does not meet objective
    ((is_string($value) && $value === '') ? $empty_string=true : $empty_string=false);
    _printf($counter, $type, $value, $empty_string, "is_string() && string === ''");

    // does not meet objective
    (empty($value) ? $empty_string=true : $empty_string=false);
    _printf($counter, $type, $value, $empty_string, "empty()");

    echo "\n";
}

?>
```

Output

```
1. string=0   emptyString()                   IS a string with a value
1. string=0   !is_string() || string === ''   IS a string with a value
1. string=0   is_string() && string === ''    IS a string with a value
1. string=0   empty()                         is NOT a string with a value

2. string=    emptyString()                   is NOT a string with a value
2. string=    !is_string() || string === ''   is NOT a string with a value
2. string=    is_string() && string === ''    is NOT a string with a value
2. string=    empty()                         is NOT a string with a value

3. array=0    emptyString()                   is NOT a string with a value
3. array=0    !is_string() || string === ''   is NOT a string with a value
3. array=0    is_string() && string === ''    IS a string with a value
3. array=0    empty()                         is NOT a string with a value

4. array=3    emptyString()                   is NOT a string with a value
4. array=3    !is_string() || string === ''   is NOT a string with a value
4. array=3    is_string() && string === ''    IS a string with a value
4. array=3    empty()                         IS a string with a value

5. integer=0  emptyString()                   is NOT a string with a value
5. integer=0  !is_string() || string === ''   is NOT a string with a value
5. integer=0  is_string() && string === ''    IS a string with a value
5. integer=0  empty()                         is NOT a string with a value

6. NULL=      emptyString()                   is NOT a string with a value
6. NULL=      !is_string() || string === ''   is NOT a string with a value
6. NULL=      is_string() && string === ''    IS a string with a value
6. NULL=      empty()                         is NOT a string with a value

7. boolean=1  emptyString()                   is NOT a string with a value
7. boolean=1  !is_string() || string === ''   is NOT a string with a value
7. boolean=1  is_string() && string === ''    IS a string with a value
7. boolean=1  empty()                         IS a string with a value

8. boolean=   emptyString()                   is NOT a string with a value
8. boolean=   !is_string() || string === ''   is NOT a string with a value
8. boolean=   is_string() && string === ''    IS a string with a value
8. boolean=   empty()                         is NOT a string with a value
```

Only the first element in the array is |ε| != 0.  It's the only one that IS a string with a value.  

Every other test is NOT a string with a value.   

`emptyString()` and `!is_string() || string === ''` produce the desired results.  

Ultimately, a method like emptyString() would reduce total character count and complexity scores.

Jellyfrog:
I still don't understand when this would happen?
```php
>>> \LibreNMS\Config::get('device_traffic_descr')
=> [
     "/loopback/",
     "/vlan/",
     "/tunnel/",
     "/bond/",
     "/null/",
     "/dummy/",
   ]
```

So what it tests atm is:
```php
preg_match('/loopback/i', ...)
preg_match('/vlan/i', ...)
preg_match('/tunnel/i', ...)
preg_match('/bond/i', ...)
preg_match('/null/i', ...)
preg_match('/dummy/i', ...)
```

If the user adds a empty string to `Config::get('device_traffic_descr')`, well.. we can't really protect all things.

Or am I missing something?
What does your `\LibreNMS\Config::get('device_traffic_descr')` contain @josephtingiris ?

josephtingiris:
> I still don't understand when this would happen?

It contains that and you're right, it shouldn't happen.

In my case it happened because all of the interface descriptions in the machine matched /bond/ or /loopback/.  So, error drawing graph.

I can work around it in the config.  There's no need for this PR.

Thanks for your patience.

PR #11140, opened 2020-02-12
Alert Detail Ping Latency

SourceDoctor:
shows Latency in Alert Detail on Ping Latency alert:

before:
image

after:
image

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)
- [x] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #11149, opened 2020-02-14
Revert "Alert Detail Ping Latency"

PipoCanaja:
Reverts librenms/librenms#11140

### DIFF
--- a/includes/html/functions.inc.php
+++ b/includes/html/functions.inc.php
@@ -981,17 +981,6 @@ function alert_details($details)
             $fallback = false;
         }
 
-        if ($tmp_alerts['last_ping_timetaken']) {
-            $fault_detail .= "Latency: <a href='" . generate_url(array('page' => 'graphs',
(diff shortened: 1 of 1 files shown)


PipoCanaja:
There is no way to define when (and how) to display properly this message. 
A solution could be to generate this "short text", using templates (exactly like the alert transports). 
Each alert rule would include a template and generate the exact needed text.

SourceDoctor:
@PipoCanaja 
i recognized it myself this morning. I'll take a look on it

SourceDoctor:
@PipoCanaja should i approve this for you so it can be removed, till i find a suitable way?

PipoCanaja:
Yep, let's merge this first, and think about a better way. Basically, you have no way to know at this position in the code, what alert the user has configured. And the ping latency is available in every alert structure. 
Only building this alert description based on the alert rule would solve this, and this is a much bigger change.

PR #11154, opened 2020-02-15
Update RRDCached.md

kkrumm1:
Remove this small note as it’s confusing to new users setting up rrd caching.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/RRDCached.md
+++ b/doc/Extensions/RRDCached.md
@@ -5,10 +5,6 @@ path: blob/master/doc/
 
 This document will explain how to setup RRDCached for LibreNMS.
 
-Since version 1.5, rrdtool / rrdcached now supports creating rrd files
-over rrdcached. If you have rrdcached 1.5.5 or above, you can also
-tune over rrdcached. To enable this set the following config:
-
 ```php
 $config['rrdtool_version'] = '1.5.5';
 ```


doc/Extensions/RRDCached.md
```diff
@@ -5,10 +5,6 @@ path: blob/master/doc/
 
 This document will explain how to setup RRDCached for LibreNMS.
 
-Since version 1.5, rrdtool / rrdcached now supports creating rrd files
-over rrdcached. If you have rrdcached 1.5.5 or above, you can also
-tune over rrdcached. To enable this set the following config:
-
 ```php
 $config['rrdtool_version'] = '1.5.5';
 ```
```
PipoCanaja: Should we keep the config and the note ?

SourceDoctor: i think it could be savely removed.

even Debian Stable (Buster) has RRDTool 1.7.1 and RRDCached 1.7.1

rrdtool 1.5.5. was released [Nov 2015](https://github.com/oetiker/rrdtool-1.x/releases)

TheGreatDoc: CentOS 7 comes with 1.4.8

SourceDoctor: mmh, as i see here it's very old, but still fully maintained.
[https://en.wikipedia.org/wiki/CentOS#Versioning_and_releases](https://en.wikipedia.org/wiki/CentOS#Versioning_and_releases)

Ok i say it should stay in it.
Maybe with a hint since which Distributions Release names/versions tune over rrdcached is possible.

PipoCanaja: All right, so we need to keep this, or even better, we need to figure out the version of rrdtool (could be done in daily for instance) and set this automatically. That way, we would avoid the manual setup.

kkrumm1:
Yeah you guys are let’s keep it in.

PR #11165, opened 2020-02-19
Global search by IP address

josephtingiris:
* If a valid IP address is typed into the global search then search ip, overwrite_ip, and hostname for a match and sort the matches by IP address.
* Added IP address to all global search results, whether it's an IP address search or not.
* Cleaned up Handlebars.compile and only display data elements if they have a non-empty value.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/html/ajax_search.php
+++ b/html/ajax_search.php
@@ -51,21 +51,21 @@ if (isset($_REQUEST['search'])) {
             // Device search
             if (Auth::user()->hasGlobalRead()) {
                 if (\LibreNMS\Util\IPv4::isValid($search, false) || \LibreNMS\Util\IPv6::isValid($search, false)) {
-                    // Device ip, overwrite_ip, or hostname search by address
+                    // Device search ip, overwrite_ip, or hostname by address
                     $results = dbFetchRows(
                         "SELECT *,inet6_ntoa(`ip`) as `ntoa_ip` FROM `devices` LEFT JOIN `locations` on `locations`.`id` = `devices`.`location_id` WHERE inet6_ntoa(`devices`.`ip`) LIKE ? OR `devices`.`overwrite_ip` LIKE ? OR `devices`.`hostname` LIKE ? ORDER BY `ip`, `hostname` LIMIT ?",
                         ["$search%", "$search%", "$search%", $limit]
                     );
                 } else {
-                    // Device hostname, location, sysname, purpose, or notes search by non-address
+                    // Device search hostname, location, sysname, purpose, or notes by non-address
                     $results = dbFetchRows(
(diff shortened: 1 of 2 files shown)


josephtingiris:
image

kkrumm1:
Yeah, this seems to break Global search completely for me. Using the search by IP address.

josephtingiris:
Does a non-address search work?

Do you still see the ports, or no results at all?

Any sql errors in the librenms.log?

kkrumm1:
Cant search by IP address and name. Nothing comes up at all.

josephtingiris:
Any sql errors in the librenms.log?

josephtingiris:
The only thing added to the name search was the use of inet6_ntoa().   I’m thinking maybe this ...


https://mariadb.com/kb/en/mysql-56-function/

... are you using mariadb and if so, exactly which version??

kkrumm1:
5.5.60-MariaDB
--

josephtingiris:
Thanks.  That explains it.  Older versions of mariadb don’t support the inet6_ntoa function.  

PHP has similar functions.  Is it safe to assume that all LibreNMS installs are using at least PHP 7?

kkrumm1:
For the DB we need to maintain backward compatibility since we have not made a specific version of Maria db and above. 

For PHP the minimum is. 7.1.3

josephtingiris:
Ok,  My mistake.  I'll keep that in mind going forward.

For reference, MariaDB doesn't start supporting inet6_ntoa() until 10.0.12 .. https://mariadb.com/kb/en/inet6_ntoa/

I can do the same thing with PHP.  Should I commit the changes to this branch, or submit a separate PR?

kkrumm1:
I would make a new PR>

SourceDoctor:
@josephtingiris 
in new PR i would use ipv4_addresses table and ipv6_addresses table to search for IP Addresses. Otherwise you'll get only the DNS resolved one's and the overwritten one's.

josephtingiris:
@SourceDoctor Ok, I'll open another PR.  Yes, but the problem with the address tables is they don't have the addresses for ping only devices.  The device ip search works well enough, especially if inet6_aton() is available.  Maybe I'll combine them and see what it looks like.

Do you think it would be acceptable to add a config option to support newer versions of mariadb (with inet6_aton())?  Similar to the $config['rrdtool_version'] setting for rrdcached?

SourceDoctor:
This tables doesn't have them, that's correct. but if you search for a device by ip addresse you should also be able to search for IP Addresses which are not used for polling.
Especially if you want to know to which Device has this IP Addresse.
So please do the complete, but not such easy way, otherwise someone has to fix it to this.
All maintainers are open to help you on problems, optimizing code and questions.

Maybe later, i'd prefer the classical, all versions running, way instead of using a version switch for only this.

murrant:

murrant:

https://community.librenms.org/t/global-search-by-ip-address/11191/2

PR #11175, opened 2020-02-21
Add csv of groups a device belongs to in alert object

VirTechSystems:
Query for the groups being alerted on and build a csv string of groups a device belongs to and add it to the alert object for use in the transports.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/RunAlerts.php
+++ b/LibreNMS/Alert/RunAlerts.php
@@ -108,6 +108,14 @@ class RunAlerts
         $obj['proc']          = $alert['proc'];
         $obj['status']        = $device['status'];
         $obj['status_reason'] = $device['status_reason'];
+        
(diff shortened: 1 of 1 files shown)


PipoCanaja:
Hi @VirTechSystems 
Why would you provide a CSV instead of the data structured in a proper way ? We have loop mecanisms availiable in the templates, so serialisation (to any format, CSV being one) can always be done, if needed, in the template itself.

VirTechSystems:
I thought a CSV would be the easiest format to use.  What other way should it be structured?

PipoCanaja:
I would just provide the array (like the faults array). Then you can loop on the group array to retrieve it and display it.

laf:
Yes definitely as another array within the existing `$obj` array.

VirTechSystems:
Ok.  I'll rework it and do some testing.

SourceDoctor:
@VirTechSystems 
something new here?

murrant:
You don't need this, you can just load the groups in your alert template.

`{{ Device::find($alert->device_id)->groups()->pluck('name')->implode(', ') }}`

Just rolls right of the tongue.

PR #11187, opened 2020-02-23
Added discovery and graphing for JunOS (SRX) RPM probes

PelNet:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

Hi,

I've added discovery support for SRX RPM probes for supported devices running JunOS. I believe the changes are relatively self-explanatory, but I've added a couple of screenshots per your stated guidelines. Please note that I'm using a custom dark theme instead of the default, though the RPM part works like any other graph/element.

2020-02-23-193250_1828x454_scrot
2020-02-23-193313_982x782_scrot

Please let me know if I missed something or if you require further info/samples/etc. 

Cheers,

PelliX


includes/discovery/functions.inc.php
```diff
@@ -449,6 +450,9 @@ function sensor_limit($class, $current)
         case 'snr':
         case 'frequency':
         case 'pressure':
+        case 'probes':
+            $limit = 5;
```
Jellyfrog: This changes the limit for airflow, Snr, frequency and pressure also, I'm guessing this is not intended?

PipoCanaja:
Hi @PelNet 
Could you name this sensor "Percent" instead of "Probes", so it can be used in a more standard way? 
Probes, by itself, does not imply "percent", not like relative humidity which is always measured as percents. 
Thanx

includes/discovery/functions.inc.php
PelNet: Yes, good point. Will do.

PelNet:
> Hi @PelNet
> Could you name this sensor "Percent" instead of "Probes", so it can be used in a more standard way?
> Probes, by itself, does not imply "percent", not like relative humidity which is always measured as percents.
> Thanx

Fair enough; I'll be a little pressed for time, so it might be Wednesday before I get back to you.

Cheers,

PelliX

PipoCanaja:
No problem here, we have all the time needed :)

includes/discovery/functions.inc.php
PelNet: Fixed in latest commit, thanks :+1: 

PelliX

Rosiak:
Mentioned here:
https://community.librenms.org/t/juniper-rpm-support/501/

PelNet:
Hi,

I believe the issues mentioned are resolved in the last commit. Should I create a new PR or keep things in this one?

Cheers,

PelliX

TheCheezeIt:
When this gets merged, I will be testing it out for you guys.

Is there anything specific you guys want me to test other than just some probes?

doc/Developing/os/Health-Information.md
```diff
@@ -37,6 +37,7 @@ the values we expect to see the data in:
 | temperature                     | C                           |
 | voltage                         | V                           |
 | waterflow                       | l/m                         |
+| probes                          | %                           |
```
laf: Needs updating to Percent here.

PipoCanaja:
> When this gets merged, I will be testing it out for you guys.

Hi @TheCheezeIt 
It would be even better to test it before it is merged :) that way your feedback could be used to improve, if necessary, the PR before merging it. 
You can do it with `./scripts/github-apply 11187`

PelNet:
Hi,

I believe the issues mentioned are resolved in the last commit. Should I create a new PR or keep things in this one? I've also uploaded the requested test SNMP data from an SRX device which has RPM configured.

Cheers,

PelliX
[junos.snmprec.txt](https://github.com/librenms/librenms/files/4270236/junos.snmprec.txt)

PipoCanaja:
Can you please add the test data directly to this PR ? create a new variant (junos_rpm.snmprec and junos_rpm.json for instance) so that they do not collide with existing test-data.

murrant:

jpieta:
Heads up the divisor for this sensor is 1 for an SRX running Junos 15 and 1000000 for an SRX running Junos 18. Sharing because I didn't find this documented anywhere, perhaps changed to be able to show "five nines" reliability.

dorkmatt:
@jpieta does this mean the recent JunOS builds need some additional conditions for the divisor?

jpieta:
@dorkmatt Yes. I added a user function to junos.yaml to use a divisor of 1,000,000 if the value was greater than 100 in order to accommodate both Junos variants. 

Also, the oid calculation in functions.php wasn't working properly because it was calculating the index length including the name of the test. I added the following to functions.php as a quick fix:

function getRPMoid($index)
{
    $indexarr = explode('.',$index);
    $len = strlen($indexarr[0]);

    switch($indexarr[2]) {
        case "currentTest":
            $endindex = 1;
            break;
        case "LastCompletedTest":
            $endindex = 2;
            break;
        case "allTests":
            $endindex = 4;
            break;
        default:
            $endindex = 4;
            break;
    }
    $calcstring = $indexarr[0] . "\t" . $indexarr[1];
    $calcbin = implode(".", unpack("c*", $calcstring));
    $oid = $len . '.' . $calcbin . '.' . $endindex;
    //$oid = '.1.3.6.1.4.1.2636.3.50.1.2.1.4.' . $len . '.' . $calcbin . '.' . $endindex;

    return $oid;
}

dorkmatt:
@jpieta Would you be willing to submit your fixes? It should be possible to add a condition based on the version number for the updated calculations.

jpieta:
@dorkmatt sure thing. Here's the condition fix in /includes/discovery/functions.inc.php. I'll submit a PR this weekend.

function getRPMvalue($value)
{
	if($value > 100) {
		$value = $value / 1000000;
	}
	return $value;
}

PR #11171, opened 2020-02-20
Move language files to right position

SourceDoctor:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)
- [x] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #11194, opened 2020-02-24
Added LLDP port linking via LLDP Port IDs

sgocken:
DO NOT DELETE THIS TEXT

Added LLDP port linking via LLDP Port IDs. lldpLocPortId is stored in the ports table when port discovery happens. Added function find_port_id_by_lldp_port_id to do the matching. Original find_port_id will run if link is not found via LLDP Port ID

The language files came on for the ride due to: https://github.com/librenms/librenms/pull/11171
I don't know the right way to remove them from my pull request.

#### Please note

#### Testers


murrant:

https://community.librenms.org/t/lldp-links-not-created-due-to-not-storing-lldplocportid/10999/9

murrant:

https://community.librenms.org/t/bug-in-lldp-remote-port-lookup/10269/5

includes/discovery/ports.inc.php
```diff
@@ -9,6 +9,9 @@
 $port_stats = snmpwalk_cache_oid($device, 'ifAlias', $port_stats, 'IF-MIB');
 $port_stats = snmpwalk_cache_oid($device, 'ifType', $port_stats, 'IF-MIB');
 $port_stats = snmpwalk_cache_oid($device, 'ifOperStatus', $port_stats, 'IF-MIB');
+if (Config::get('autodiscovery.xdp') === true) {
```
PipoCanaja: Hi
It is not possible to run some xdp code here.

sgocken: Can't even see if it is enabled?

PipoCanaja:
Hi @sgocken 
As already mentionned in the forum, we can't go that route. We need to define a correct DB schema for this before actually coding it. 
Concerning the language files, you could "git checkout html/js/lang/*" from master (the exact command depends on your git configuration.

includes/discovery/ports.inc.php
PipoCanaja: Of course we can see, but it does not make sense in term of code structure. 
When you create modules to split different parts of the code, you just cannot do that. There are solutions to do this in a clean way, by storing the data in a dedicated xdp table (name to be defined) and populating it exclusively from the discovery_protocols module.

includes/discovery/functions.inc.php
```diff
@@ -1482,3 +1482,60 @@ function find_port_id($description, $identifier = '', $device_id = 0, $mac_addre
...
+ */
+function discover_lldp_loc_port_id($device)
+{
+
+    $port_stats = [];
+    $port_stats = snmpwalk_cache_oid($device, 'lldpLocPortId', $port_stats, 'LLDP-MIB');
+
+    foreach ($port_stats as $ifIndex => $snmp_data) {
+        $port = dbFetchRow('SELECT * from `ports` where `device_id` = ? AND `ifIndex` = ?', [$device['device_id'], $ifIndex]);
+        if ($port['port_id']) {
```
PipoCanaja: I don't see an "else" statement for this condition. So if the port does not exist yet in the DB, you do not store the lldp data currently.

database/migrations/2020_02_25_101056_create_ports_lldplocportid_table.php
```diff
@@ -0,0 +1,33 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('ports_lldplocportid', function (Blueprint $table) {
+            $table->unsignedInteger('device_id');
+            $table->string('lldpLocPortId');
+            $table->unsignedInteger('port_id')->nullable();
+            $table->unique(['device_id', 'lldpLocPortId']);
+        });
```
PipoCanaja: Could you also store the lldpLocChassisId. VMWare ESXi is (mis)using this field to provide the lldpLocPortId currently. This way, I could extend your code later to allow esxi neighbors to be discovered correctly.

PipoCanaja:
Hi @sgocken 
I rebased your PR to fix the language part. You have to do, from your machine : 
`git pull origin community-10999` to get the updated version.

includes/discovery/functions.inc.php
sgocken: Yeah, that was my plan, but after thinking about some other cases, I think I am going to rework that a bit.
Right now it will work if a switch module gets plugged in and adds ports to a switch, but if one is taken out, those old ports will still be there. I am going to look into how the current ports mod handles that and make some changes.

database/migrations/2020_02_25_101056_create_ports_lldplocportid_table.php
sgocken: I will add that. Don't you just love how vendors follow the standards.
I will also update the find port function to or on either of those fields.

includes/discovery/functions.inc.php
PipoCanaja: Correct, that's the point that would have come just after. Basically, every time you do LLDP discovery, you maintain an array of the entries that are in the DB and in the LLDP. And you delete any entry that is in the DB and not match in the snmp, and add/update any from the SNMP to the DB.

murrant:
Don't we already store this data? why duplicate it?

If we know lldpPortIdSubtype then we know which field to look at in the ports table.

murrant:
Closing do to inactivity, feel free to re-open.

strixdio:
I'm not sure if this is the cause of the issue I'm experiencing, hopefully someone has made some progress on it?

My current setup is: Mikrotik 5009 Local Port SFP+1 > Mikrotik CRS Remote Port SFP+1
However, in LibreNMS, the Mikrotik 5009 Local port is showing as Ether1.

I have verified in both devices that the lldp data is showing correct.

sgocken:
> 

I made this due to Nortel/Avaya switches referring to their LLDP port IDs as a MAC address. 
This was fixed in the switch software in some switch families, but others were EOLed before that fix was implemented. 

I think it you look at the discovery capture for one of these devices you should be able to see what LibreNMS sees and uses.
This is more of a forum post question. As Mikrotiks are heavily used, someone should have some input.

TafkaMax:
It seems this fix is not included? I am using mikrotik devices and see this behavior, where another ports LLDP is mapped to another port.

> > 
> 
> I made this due to Nortel/Avaya switches referring to their LLDP port IDs as a MAC address. This was fixed in the switch software in some switch families, but others were EOLed before that fix was implemented.
> 
> I think it you look at the discovery capture for one of these devices you should be able to see what LibreNMS sees and uses. This is more of a forum post question. As Mikrotiks are heavily used, someone should have some input.

TafkaMax:
> I was never able to get the fix to the point where it was accepted. I am still using my fix on a very old version.
> 
> Are you using all Mikrotik devices? How does it act in your environment?
> 
> > It seems this fix is not included? I am using mikrotik devices and see this behavior, where another ports LLDP is mapped to another port.
> > > > 
> > > 
> > > 
> > > I made this due to Nortel/Avaya switches referring to their LLDP port IDs as a MAC address. This was fixed in the switch software in some switch families, but others were EOLed before that fix was implemented.
> > > I think it you look at the discovery capture for one of these devices you should be able to see what LibreNMS sees and uses. This is more of a forum post question. As Mikrotiks are heavily used, someone should have some input.

I  am running librenms on debian 11 and I have autoupdates enabled for librenms.

I have 1 CRS switch and 1 CCR router and AP-s, but the CRS is the one that is connected to librenms most throughly. It shows the LLDP-s, but they are all over the place e.g. ether1 has the LLDP info of ether 3. I will get you the information soon.

TafkaMax:
I upgraded the CRS from 7.9.2 to 7.11.2 to get some vendor fixes hopefully.

For example:

combo 3 is connected to fortiswitch. But the librenms connects it to ether8, which is totally wrong. Same for combo 4 which is to another fortiswitch.

image

PR #11197, opened 2020-02-24
Global Settings - Mountpoint ignore options

SourceDoctor:
Mountpoints to be ignored no configurable via global Settings
image

#### Please note

#### Testers


misc/config_definitions.json
```diff
@@ -3234,10 +3246,16 @@
             "type": "array"
```
murrant: I think you can validate these are regex with:
```
            "validate": {
                "value.*": "regex"
            }
```

murrant: I think you can validate these are regex with:
```
            "validate": {
                "value.*": "regex"
            }
```

murrant: I wish there was a better rule to use...

murrant: I guess we could define a new validation:
```
        Validator::extend('is_regex', function ($attribute, $value) {
            return @preg_match($value, null) !== false;
        });
```

SourceDoctor: i only brought it to GUI :-)

SourceDoctor: > I guess we could define a new validation:
> 
> ```
>         Validator::extend('is_regex', function ($attribute, $value) {
>             return @preg_match($value, null) !== false;
>         }, __(':attribute is not a valid regular expression'));
> ```

do you want me to implement this sample? If so, where?

murrant: It goes in AppServiceProvider::bootCustomValidators()

SourceDoctor: ah ok, thx
i'll update file and config definitions in 4-5h. I'm still at work :-)

SourceDoctor: included :-)

SourceDoctor:
@murrant 
something else ?

SourceDoctor:
@murrant 
something else to do?

murrant:

PR #11209, opened 2020-02-26
Add more interface types to bad_iftype array in config_definitions.json

pepperoni-pi:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Support/Configuration.md
+++ b/doc/Support/Configuration.md
@@ -659,7 +659,7 @@ $config['bad_if_regexp'][] = '/^lo[0-9].*/';    // loopback
 ```
 
 Numerous defaults exist for this array already (see
-includes/defaults.inc.php for the full list). You can expand this list
+misc/config_definitions.json for the full list). You can expand this list
 by continuing the array.
 
 `bad_if` is matched against the ifDescr value.
--- a/misc/config_definitions.json
+++ b/misc/config_definitions.json
@@ -681,7 +681,9 @@
                 "atmSubInterface",
                 "aal5",
                 "shdsl",
-                "mpls"
+                "mpls",
+                "macSecControlledIF",
+                "macSecUncontrolledIF"
             ]
         },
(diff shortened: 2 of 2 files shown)


murrant:
Actually, we are looking to eliminate all the default entries for this setting.
Instead they should be added to the OS that needs them.

pepperoni-pi:
LibreNMS by default will discover a bunch of "Uncontrolled" and "Controlled" interfaces which do not seem to be useful in any way so I have currently just added them to the array via my config.php, but since they seem to not be useful at all I figured I should just get them added to the config_definitions.json bad_iftype array.

murrant:
The problem is that this applies to ALL OS and it is hard to remove them from the global ignore array defaults.  Please just add them to the OS definition that has the interfaces.  Thanks.

PipoCanaja:
> LibreNMS by default will discover a bunch of "Uncontrolled" and "Controlled" interfaces which do not seem to be useful in any way so I have currently just added them to the array via my config.php, but since they seem to not be useful at all I figured I should just get them added to the config_definitions.json bad_iftype array.

Hi @pepperoni-pi 
You can use this file as a model on how to add these directly to the OS concerned : 
```
librenms@librenms:~/includes/definitions$ cat nokia-isam.yaml
os: nokia-isam
text: 'NOKIA ISAM'
type: network
icon: nokia
group: nokia
bad_ifXEntry:
    - NFXS-E
over:
    - { graph: device_processor, text: 'CPU Usage' }
discovery:
    -
      sysObjectID:
        - .1.3.6.1.4.1.637.61
mib_dir:
    - nokia
bad_iftype:
    - aluGponOnu
    - bridge
    - ieee8023adLag
    - l2vlan
    - slip
    - softwareLoopback
```

pepperoni-pi:
Sweet! Thanks guys. I'll close this request and create a new one after I do some testing.

PR #11210, opened 2020-02-27
Running Alerts colorized depending on to their severity

SourceDoctor:
similar to Alert Widget, Colors should symbolize their severity for active Alerts

image

#### Please note

#### Testers


kkrumm1:
Looks great!

includes/html/print-alert-rules.php
```diff
@@ -200,9 +201,27 @@
             $status_msg = "All devices matching " . $rule['name'] . "  are OK";
         }
         if ((int) $sub['state'] === 1 || (int) $sub['state'] === 2) {
-            $ico   = 'exclamation';
-            $col   = 'danger';
-            $extra = 'danger';
+            switch ($severity) {
```
Jellyfrog: This should probably be a generic function, so we use the same classes everywhere

SourceDoctor: done :-)

Jellyfrog:
There are a couple of places that could be combined probably;
https://github.com/librenms/librenms/blob/6dfa5ed75a9ec0f8dd8cdd0fe32aaf1f6501ea0a/includes/html/table/alerts.inc.php#L135-L148

SourceDoctor:
@Jellyfrog 
maybe centralize them in an extra PR?

SourceDoctor:
@Jellyfrog 
done :)

laf:
This is just the alert rules page though isn't it. Imho I don't think we need to colour each row based on the severity the rule is set at.

SourceDoctor:
@laf 
This Alert Rule Page is also used in edit Device context. Filtered down to Alert Rules matching to the Device you are in.
It's for nice and fast seeing which Rule is actually alarming and which severity it has.
In Alert Widget Severity also has specific colors, so why in Alert Rules only read (which means critical)?

SourceDoctor:
@Jellyfrog 
something else, or ready for merge?

murrant:

PR #11226, opened 2020-02-28
Options for whitelisting interfaces to be included in Overall Traffic graph

pepperoni-pi:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Support/Configuration.md
+++ b/doc/Support/Configuration.md
@@ -442,6 +442,10 @@ device_traffic_iftype:
         },
 ```
 
+`device_traffic_descr` matches on either the ifDescr or ifName value.
+
+`device_traffic_iftype` matches on the ifType value.
+
 There are several options to re-include an interface that would otherwise be
 excluded by the default arrays:
 
--- a/doc/Support/Configuration.md
+++ b/doc/Support/Configuration.md
@@ -368,13 +368,6 @@ option would allow you to make use of the SNMP sysName instead as the
 preferred reference to the device.
 
 ```php
-$config['device_traffic_iftype'][] = '/loopback/';
-```
-
-Interface types that aren't graphed in the WebUI. The default array
-contains more items, please see misc/config_definitions.json for the full list.
-
-```php
 $config['enable_clear_discovery'] = 1;
 ```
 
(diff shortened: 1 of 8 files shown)


pepperoni-pi:
Here's the situation where this comes in handy - https://community.librenms.org/t/overall-traffic-not-including-cisco-firepower-interfaces/10753

Feature Request - https://community.librenms.org/t/include-interface-in-overall-traffic-using-ifalias-or-ifdescr-that-otherwise-would-be-excluded/10793

pepperoni-pi:
**Not Working**
not-working

**Working**
working

pepperoni-pi:
Right now this works with editing the config.php file and OS specific values for device_traffic_good_descr, but I haven't been able to get it to work yet using the /opt/librenms/includes/definitions/ftd.yaml

**Does work**
$config['os']['ftd']['device_traffic_good_descr'][] = '/adaptive/';

**Does not work**
```yaml
os: ftd
text: 'Cisco FTD'
type: firewall
icon: cisco
group: cisco
...
device_traffic_good_descr:
    - adaptive
```

pepperoni-pi:
I'll update documentation provided there's some consensus around this.

murrant:

https://community.librenms.org/t/include-interface-in-overall-traffic-using-ifalias-or-ifdescr-that-otherwise-would-be-excluded/10793/2

pepperoni-pi:
May add other device_traffic_good_* options such as device_traffic_good_iftype and device_traffic_good_ifalias. With a device_traffic_good_iftype we would be able to remove the band-aid in /opt/librenms/includes/html/graphs/device/bits.inc.php for ASAs operating in multi-context mode and also give people the option to include interfaces in the overall traffic graph that would normally be ignored without them having to submit a pull request. Instead they could just edit their config.php and be all set. Or if they do want to fix it for everyone, they could submit a PR to update the /opt/librenms/includes/definitions/*.yml file.

pepperoni-pi:
I could do those as separate PRs if that would be preferred too.

PipoCanaja:
Hi @pepperoni-pi 
Looks good.
I suggest that you apply the patch as well for ASA in multicontext mode, so you can indeed clear the hack that was in place. 
Please also add the documentation of this new option. 
Thanx

murrant:
@pepperoni-pi Thanks for all the work on this.

This PR is problematic for me because the whole bad_if/good_if situation on LibreNMS is terrible.
*1 there should not be a global bad port type definition at all
*2 bad ports should only be defined per os
*3 it should be easy for users to configure good/bad ports.
*4 there should be no concept of "good" only a blacklist.

I know this is way out of scope for you, but merging your change here only makes it harder to get to the above.

pepperoni-pi:
@murrant I see. Well, I suppose we could copy the current default arrays into the OS definition files. Wouldn't be very difficult to change. May result in a lot of PRs to remove entries from the OS definitions, but that would force the issue and possibly resolve this dilemma. Changing the match in includes/html/graphs/device/bits.inc.php wouldn't be very difficult.

pepperoni-pi:
Closed per my discussion with @murrant in the Discord #devel channel.

PR #11240, opened 2020-03-03
Modify os Ray and add Ray3 support

erotel:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


erotel:
modify os Ray new firmware includes
-SNMP non-table items OID defined in accordance with RFC (ending ´.0´)

PipoCanaja:
Hi @erotel 
Your change seems to modify the existing sensors/indexes for Ray devices. Which means that any historical data would be lost when it is applied. Have you noticed this ? 
@Martin22 Have you seen that when you tested it?

Martin22:
> Hi @erotel
> Your change seems to modify the existing sensors/indexes for Ray devices. Which means that any historical data would be lost when it is applied. Have you noticed this ?
> @Martin22 Have you seen that when you tested it?

The historical data remains, but when the OID is changed / repaired in the firmware, it is difficult to solve without losing the data. It has to be taken into account after the firmware upgrade, you can still rename the old RRD files to the new name.

PipoCanaja:
I am fine with the change from Ray to Ray3. My concern is that if we apply this patch to a LibreNMS with a Ray device (not 3) it looks like IDs are changing and rrd might need to be renamed. That's what I want to check.

Martin22:
If you do not load the latest firmware, the data will not be lost. OID changed in new firmware.

Martin22:
Only data for temperature and voltage were lost, the other data remained.

PipoCanaja:
I don't speak about firmware update at all but about this PR. This PR makes changes to Ray and Ray3. 
Let say a user has Ray devices, not Ray3. Then after this PR is merged, some details (processor, others) will be rediscovered in a different way (YAML instead of PHP etc) and it seems that index are changed. That's the issue I am trying to identify for sure, in order to notify user in case it is confirmed.

Martin22:
@PipoCanaja It will only be changed if RAY2 is upgraded to the latest firmware, if the upgrade is not performed, all data will remain.

erotel:
The latest firmware for ray and ray2 fixes a bug where the snmp was missing an index (.0).
Ray3 is a newer device and snmp has been changed since firmware 1.0.16.0.

before 1.0.16.0     `.1.3.6.1.4.1.33555.1.2.1.4.0`
after 1.0.16.0        `.1.3.6.1.4.1.33555.4.2.1.4.0`

why so is i asked developers, the answer was "just so is"

erotel:
ray and ray2 will lose non-wireless data.
With ray3, all the data is lost, but I don't think it can be resolved.
With the current librenms version, Ray3 reports as a generic device.

erotel:
???

PipoCanaja:
I don't speak about the firmware itself, but about how LibreNMS handles it. What we must ensure is that this PR does not change anything to currently deployed OS. 
- Can you confirm that this PR does not break of change anything for Ray and Ray2 devices that are not updated.
This is the expected "non regression" of LibreNMS. Avoid breaking things for users. Now if we cannot avoid to break, then we have to notify.
 
I don't own the devices, so only you can do tests, and convince the community that this PR is safe for everybody OR should be notified because breaking can't be avoided.

erotel:
If the firmware is upgraded to RAY2, only the CPU graph, internal temp and voltage are deleted.
Ray will not work at all, the manufacturer has discontinued support.
Ray3 is currently reporting as a generic device.

PipoCanaja:
So this PR cannot be merged. The goal is that nothing changes for current users with existing Ray and Ray2. It cannot be "removed" like this. The goal, to get this PR merged:
- Current Ray users do not see any regression
- Current Ray2 users do not see any regression
- Current Ray3 users see improvements

erotel:
You can then clear Ray support.
Ray is not sold and the manufacturer is no longer supported.
Ray2 also stops working in librenms after the firmware is updated.
Because Ray2 update firmware includes security fixes, I assume most users will install this firmware.
Ray3 from firmware 1.0.16.0 does not work in librenms.

PipoCanaja:
I must explain it wrong, sorry. The goal is to add things and not remove things. So whatever the behaviour is now, either we improve it or we keep it. We just have to avoid removing things. 
If you can fix this, we can merge this PR. If not, we should keep the existing code.

erotel:
I just don't get it. The current state will make it obsolete (with poor OID support) and unsupported Ray.
Well, the newer Ray2 and Ray3 devices will simply be excluded from support, just because we don't want to delete the data?
I just think it's better to delete the unsupported hardware, fix the support for the newer hardware, and add the latest hardware support.

PipoCanaja:
**We can add new support without breaking the old**. There is absolutely no reason to avoid adding new things just because we continue to support the old one. 
Sometimes, it makes it a little bit tricky but in case of a complete new implementation (Ray3) this is easy. Just add the new one and don't touch the old. 
For a modified one, it depends, it can be tricky. 

But if I have a device up and running and monitored in Libre, and somebody breaks it one day with a PR, I'll be mad. Even if it is old. Even if it is unsupported by the vendor. Even if I should have upgraded. I may have a very good reason not to do the upgrade, you cannot tell.

PipoCanaja:
If somebody else wants to work on this, feel free to open a new PR and define exactly what is added, changed, and if necessary notify users (to migrate RRDs). There is always a way to add support without breaking anything.

PR #11196, opened 2020-02-24
Additional Open Sense detection

FingerlessGlov3s:
Update opnsense.yaml for new OPNsense OID called version.
This will help with detection after the user turns on "Display Verion in OID" within OPNsense.
Currently detection won't work unless you have OPNsense in the hostname

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [X] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)
- [X] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #11251, opened 2020-03-04
Update OPNsense definition due to incorrect OID

FingerlessGlov3s:
sysObjectID is incorrect for OPNsense should be .1.3.6.1.4.1.8072.3.2.8.
This has been tested in my environment with 2 OPNsense Firewalls.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


FingerlessGlov3s:
Hey Pipo,

It could be wrong I'm not sure where @murrant got the OID from. If its got an install and that's the OID then yes we need a OR.


https://github.com/librenms/librenms/pull/11196#issuecomment-591186731

PipoCanaja:
@FingerlessGlov3s May be in the early days of OPNsense, they had the same. Anyway, we have to keep it for the actual users, cause if we don't it will break their LibreNMS, and we don't want that. So you'll have to go with the "OR" structure that I suggested above, and Travis should happily say "tests are OK". 
Then you would have to create a test file for your OPNsense instance (so that next time somebody changes the OS, it does not break your version).
And finally, we'll be good to merge :)

FingerlessGlov3s:
The OID ".1.3.6.1.4.1.12325.1.1.2.1.1" is indeed for PFsense, I checked it last night.

OID ".1.3.6.1.4.1.8072.3.2.8" is for FreeBSD identification https://oidref.com/1.3.6.1.4.1.8072.3.2.8

OID  ".1.3.6.1.4.1.8072.3.2.8" being used for OPNsense (FreeBSD based). They may have used PFsense's OID but I've checked far back as 18.7 and its still ".1.3.6.1.4.1.8072.3.2.8", which to be honest if your running older than that you should really update. As its security OS, and you should keep it somewhat up to date. I would need to go back and install older versions to check if the OID was ever ".1.3.6.1.4.1.12325.1.1.2.1.1" but at this point we're looking at 3 year old software, that is easily upgraded for free.

I was the one who created the PR to put this new dectection method in PR #11196 . I think Murrant just got confused with PFsense and copied the wrong OID, or just guessed what it might be.
"break their LibreNMS" isn't an issue because the following check for ".1.3.6.1.4.1.8072.1.3.2.3.1.2.7.118.101.114.115.105.111.110" to see if it contains OPNsense, won't work because this has only just been added in OPNsense 20.1, which was a feature requested by myself.

So I'm very sure my commit is correct 😀

Also I'm not sure what you mean by "test file" 🤔

PipoCanaja:
".1.3.6.1.4.1.12325.1.1.2.1.1" was there before your PR, so there was no reason, at that time, to remove it. 
Now, ".1.3.6.1.4.1.12325.1.1.2.1.1" is another identifier from FreeBSD, not pfSense specific. Which means that we cannot conclude ".1.3.6.1.4.1.12325.1.1.2.1.1" can only be pfSense. 

http://www.circitor.fr/Mibs/Html/B/BEGEMOT-SNMPD-MIB.php
https://www.gsp.com/cgi-bin/man.cgi?section=1&topic=bsnmpd

And we cannot punish people who kept running the old versions of opnsense, even I agree they *should* upgrade :) They might have very good reasons. 

So I kindly insist that you put the OR version, which does not at all prevent you from adding your feature, and keeps the old behaviour for other devices, thanx.

FingerlessGlov3s:
Yeah the OID ".1.3.6.1.4.1.12325.1.1.2.1.1" is probably generic classifier, I was just staying that OID is used on PFsense. 😀

".1.3.6.1.4.1.12325.1.1.2.1.1 was there before your PR" not before my original PR from the other week which I have been saying is incorrect. 
https://github.com/librenms/librenms/commit/fe9deb9b9cf407fc96b0d960994f5b69d2a5939b

I didn't get time to fully test it and it got approved. As Murrant added in the extra check so snmpget would only get run unless sysObjectID is ".1.3.6.1.4.1.12325.1.1.2.1.1", but as I said the snmpget for ".1.3.6.1.4.1.8072.1.3.2.3.1.2.7.118.101.114.115.105.111.110" only exists for OPNsense systems 20.1 and newer.

So currently as it stands the detection works as follows:
The sysDescr contains OPNsense
OR
   If the sysObjectID is ".1.3.6.1.4.1.12325.1.1.2.1.1"
     Then check to see if ".1.3.6.1.4.1.8072.1.3.2.3.1.2.7.118.101.114.115.105.111.110" contains OPNsense
END

What my commit does is change the if on the sysObjectID to ".1.3.6.1.4.1.8072.3.2.8" as ".1.3.6.1.4.1.8072.1.3.2.3.1.2.7.118.101.114.115.105.111.110" only exists in the latest version of OPNsnse where the sysObjectID  is ".1.3.6.1.4.1.8072.3.2.8"

I hope this makes better sense now. Probably just be describing the issue incorrectly.

Yes I know we shouldn't punish people for running old versions but there becomes a point where you should, but lets not get in to that. 😀

PipoCanaja:
It makes a lot more sense now, got it. 

So, next step is to correct the test data: Look at the Travis test results, and check the doc about tests so you can correct them.

PipoCanaja:
@murrant snmpget is indeed necessary, to match only OPNsense and not all freebsd. That's the all poing of the initial PR.

murrant:
Please provide actual data in the snmprec from the device...

FingerlessGlov3s:
Didn't I add the test correctly?

https://github.com/librenms/librenms/pull/11251/commits/ccb46762f34e51f8ca971703e9f884a54f090a78

PipoCanaja:
@FingerlessGlov3s Murrant is asking for a real device snmp collection as described in the doc, in order to have a full sample.

FingerlessGlov3s:
I'll do that with a new PR

PipoCanaja:
Please do it here instead. Will be more difficult to do it in 2 PRs...

FingerlessGlov3s:
How do I do that, since I deleted the source repo :-o

PipoCanaja:
I would say that you close this PR, and create a new one with all then :D

FingerlessGlov3s:
Will do give me 5 minutes hehe

FingerlessGlov3s:
Closing please see PR below

https://github.com/librenms/librenms/pull/11265

PR #11269, opened 2020-03-10
Geographical map: Show network links between locations

kedare:
This PR add an optional links view on the globe map, it take the links from the discovery protocols and use the location attached to device to build them.

It's a first version, I'm sure there are some improvements that could be done :)
(For example, I could not make successfully the SQL request using joins instead of all the mapping in the where, or make the PolyLine to cross the border of the maps (for example to link APAC to US without having the link crossing Europe instead of just going between them, but I could not find any leaflet setting to do this, not sure it is possible)

Screenshot 2020-03-10 at 16 45 47

It is using the same rule than on the network map (for color and link size)

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


SourceDoctor:
could you change default value to false

and make it changeable via global Settings?
file: resources/lang/en/settings.php

i would position it here:
Global Settings -> external -> Location Settings

includes/html/common/worldmap.inc.php
```diff
@@ -130,6 +130,123 @@
 marker.bindPopup(title);
     markers.addLayer(marker);\n";
     }
+
+    if (Config::get('network_map_show_on_worldmap', true)) {
```
SourceDoctor: , true   is not needed because default value is defined via config_definitions file

kedare:
Fixed, I added a new part in the settings related to the world map as there are probably more settings coming (and it's not something external)

However I could not get the l18n to work correctly. (See screenshot), am I doing something wrong ? Any idea ?

Screenshot 2020-03-10 at 19 25 16

kedare:
Fixed with `./lnms translation:generate`

PipoCanaja:
@kedare 
Seems that Travis detects an issue in the config json file : 
```
1) LibreNMS\Tests\YamlTest::testConfigSchema
config_definitions.json does not validate. Violations:
[config.network_map_show_on_worldmap.order] The property order is required
```

kedare:
Should be fixed now :)

kedare:
Looks like the failing CI is related to something else ?

SourceDoctor:
@kedare 
crash causes an OS test. Will be fixes soon

PipoCanaja:
@kedare Bonjour. Tests are fixed now. Bye

kedare:
Great, thanks :)

SourceDoctor:
@kedare please resolve merge conflict

kedare:
Should be ok now

murrant:

bernierherve:
Hi,
Very good functionnality.
But it works only on Overview>>Maps>>Geographical.
Could be a good idea to implement in WorldMap gidget also
Thanks in advance

PR #11278, opened 2020-03-11
Fallback to default Poller Group on delete

SourceDoctor:
On deleting a Poller Group all assigned Devices should fall back to default Poller to be polled further

#### Please note

#### Testers


murrant:
Probably could add a model event and use Eloquent to delete it so it triggers the event.  https://laravel.com/docs/5.8/eloquent#events

murrant:
See the Device model for a better example

SourceDoctor:
i'll look to integrate this into eloquent ;-)

SourceDoctor:
@murrant 
done. Could you take an eye on it? :-)

murrant:
@SourceDoctor in case you didn't know, you can use Eloquent from legacy code.

app/Http/Controllers/PollerGroupController.php
```diff
@@ -0,0 +1,19 @@
...
+
+use App\Models\Device;
+use App\Models\PollerGroups;
+use Illuminate\Http\Request;
+
+class PollerGroupController extends Controller
+{
+    public function destroy(Request $request, PollerGroups $pollergroup)
+    {
+        $this->authorize('admin', $request->user());
```
murrant: Try to make specific authorization, which can just defer to isAdmin() for now in the authorization code.  This will save us time down the road when we implement RBAC.

SourceDoctor: changed

SourceDoctor:
@murrant 
something still to do?

app/Models/PollerGroups.php
```diff
@@ -33,6 +33,21 @@ class PollerGroups extends Model
...
+     */
+    public static function boot()
+    {
+        parent::boot();
+
+        static::deleting(function (PollerGroups $pollergroup) {
+            // handle device pollergroup fallback to default poller
+            $default_poller_id = \LibreNMS\Config::get('distributed_poller_group');
+            $pollergroup->devices()->update(['poller_group' => $default_poller_id]);
+            $pollergroup->save();
```
murrant: I don't think you need to save the PollerGroup  Also, PollerGroup model class name should be singular.

murrant:

PR #8703, opened 2018-05-12
WIP Convert Ajax calls to Laravel

murrant:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [ ] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

PR #11287, opened 2020-03-14
Some Poller page cleanups

murrant:
Some queries loaded all devices into memory and the counted them, we should have sql count them if we only need a count.
Models should be singular
Use named routes for url generation
Try to keep presentation and data collection separated in blade and controller a little more
Don't use authorize if it isn't wired up.

@SourceDoctor I also noticed all the ajax calls are to legacy code ajax_table and ajax_form.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/PollerController.php
+++ b/app/Http/Controllers/PollerController.php
@@ -75,7 +75,8 @@ class PollerController extends Controller
         });
     }
 
-    private function checkTimeSinceLastPoll($seconds) {
+    private function checkTimeSinceLastPoll($seconds)
+    {
         if ($seconds >= $this->rrdstep) {
             return 'danger';
         } elseif ($seconds >= ($this->rrdstep * 0.95)) {
--- a/app/Http/Controllers/PollerController.php
+++ b/app/Http/Controllers/PollerController.php
@@ -16,7 +16,7 @@ class PollerController extends Controller
 
     public function __construct()
     {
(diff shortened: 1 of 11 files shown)


SourceDoctor:
yeah i know, Still much to do for rewriting.
Maybe create a new similar to [this](https://github.com/librenms/librenms/pull/8703) for ajax calls and going through the system like other tasks in the [Project](https://github.com/librenms/librenms/projects/4)

app/Http/Controllers/PollerController.php
```diff
@@ -5,26 +5,18 @@
...
-        'id' => 0,
-        'group_name' => 'General',
-        'descr' => ''
-    ];
-    public $defaultPollerMarker = '(default Poller)';
 
     public function __construct()
     {
-        $this->authorizeResource(PollerGroups::class, 'poller_groups'); // FIXME is this correct? not a resource anymore
         $this->rrdstep = \LibreNMS\Config::get('rrd.step');
```
SourceDoctor: could also be changed to `Config::get('rrd.step');` :)

murrant: Oops, It was, but then I changed things around and it got reverted :)

murrant:
@SourceDoctor  Yeah, I know the feeling it is like a rabbit hole.

PR #11301, opened 2020-03-17
Widget Fix - Server Stats - show DisplayName instead Device ID

SourceDoctor:
show Devicename instead of Device_id when Host selected

before:
image

after:
image

#### Please note

#### Testers

### DIFF
--- a/resources/views/widgets/settings/server-stats.blade.php
+++ b/resources/views/widgets/settings/server-stats.blade.php
@@ -30,6 +30,6 @@
 
 @section('javascript')
     <script type="text/javascript">
-        init_select2('#device-{{ $id }}', 'device', {}, '{{ $device ? $device->device_id : '' }}');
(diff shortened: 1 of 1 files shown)


murrant:
I think you have to set it in an object with {"label": "displayName", "value": 5} or something like that (pretty sure those aren't the right keys)  Try saving it without changing the value.

murrant:
Actually, thinking about it a bit.  I think most of them that deal with this prepopulate the selected value in html with an option tag.

SourceDoctor:
in database JSON is still the same.
With this PR Select list looks as it should.
in "init_select2" 4th argument is "selected" text
Seeing no problem here.

murrant:
Edit -> Save -> Edit = no device set

image

murrant:
@SourceDoctor That should fix it ;)

murrant:

PR #11317, opened 2020-03-20
Widget Top Interface - Option to ignore lo Interface

SourceDoctor:
Option to ignore Loopback Interfaces on Top Interfaces

image

#### Please note

#### Testers


murrant:
Can we just have it respect disabled ports?

SourceDoctor:
you mean to have the option to ignore disabled ports also?

SourceDoctor:
@murrant 
hope this is what you meant

app/Http/Controllers/Widgets/TopInterfacesController.php
```diff
@@ -54,11 +55,15 @@ public function getView(Request $request)
             ->select('port_id', 'device_id', 'ifName', 'ifDescr', 'ifAlias')
             ->groupBy('port_id', 'device_id', 'ifName', 'ifDescr', 'ifAlias')
             ->where('poll_time', '>', Carbon::now()->subMinutes($data['time_interval'])->timestamp)
+            ->where([['ignore', '=', 0], ['disabled', '=', 0]])
```
Jellyfrog: Use the scope. https://github.com/librenms/librenms/blob/master/app/Models/Port.php#L163

app/Http/Controllers/Widgets/TopInterfacesController.php
```diff
@@ -54,11 +55,15 @@ public function getView(Request $request)
...
             ->when($data['device_group'], function ($query) use ($data) {
                 $query->inDeviceGroup($data['device_group']);
             }, function ($query) {
                 $query->has('device');
             })
+            ->when(!empty($data['hide_lo']), function ($query) {
+                $query->where('ifType', '<>', 'softwareLoopback');
```
Jellyfrog: This should probably be made into a scope. Sounds usable

app/Http/Controllers/Widgets/TopInterfacesController.php
SourceDoctor: how do i use it?

Jellyfrog: https://laravel.com/docs/7.x/eloquent#query-scopes

app/Http/Controllers/Widgets/TopInterfacesController.php
SourceDoctor: changed :-)

app/Http/Controllers/Widgets/TopInterfacesController.php
SourceDoctor: changed :)

app/Http/Controllers/Widgets/TopInterfacesController.php
```diff
@@ -54,11 +55,15 @@ public function getView(Request $request)
             ->select('port_id', 'device_id', 'ifName', 'ifDescr', 'ifAlias')
             ->groupBy('port_id', 'device_id', 'ifName', 'ifDescr', 'ifAlias')
             ->where('poll_time', '>', Carbon::now()->subMinutes($data['time_interval'])->timestamp)
+            ->IsActive()
```
murrant: FYI, the case should be `->isActive()` for scopes.

murrant:
@SourceDoctor if we respect disabled ports, doesn't that make the loopback option unnecessary?

SourceDoctor:
if you disable a loopback Port on your system, your system will crash, so it cannot be disabled
so a loopback port is active

murrant:
That isn't what disabled means.
ifOperStatus shows port up/down
ifAdminStatus shows port disabled/enabled (by the device)
ignore means ignore for alerts (but has to be check by alert rule)
disabled means port is set to disabled in the LibreNMS webui
deleted means the port is still in the DB but pending deletion.

the widget should not show disabled, and deleted devices.  It should show ignore devices, that only applies to alerts.

Right now the bad_if stuff doesn't set disabled, but I plan to extend that to do so.  We could perhaps check is_port_valid() before displaying... might need a little extra to make things better.

app/Http/Controllers/Widgets/TopInterfacesController.php
SourceDoctor: if i change it,  'scopeisActive' is not more camelcase.
if i keep it camelcase 'isActive' cannot be found.

ideas?

SourceDoctor: changed

SourceDoctor:
@murrant 
scope updated ... hope it matches now better :-)

SourceDoctor:
ok what/where to change/move it to?

SourceDoctor:
@murrant 
could need some ideas what/how the better place would be
or how this would be better implemented.
Or did you do this with your last rewrite already?

murrant:
I implemented part of my thoughts on in the last PR, but the rest requires a re-written port management.

PR #11323, opened 2020-03-22
Database: add primary key to device_graphs

rkojedzinszky:
Percona Xtradb refused to INSERT IGNORE into a table
without a primary key.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/database/migrations/2020_03_22_1424_add_primary_key_to_device_graphs.php
+++ b/database/migrations/2020_03_22_1424_add_primary_key_to_device_graphs.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
+    This migration adds primary key for device_graphs
+
+    Percona Xtradb refused to INSERT IGNORE into a table
+    without a primary key.
+ */
+
+use Illuminate\Support\Facades\DB;
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+
+class AddPrimaryKeyToDeviceGraphs extends Migration
+{
+
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        DB::table('device_graphs')->delete();
+
+        Schema::table('device_graphs', function (Blueprint $table) {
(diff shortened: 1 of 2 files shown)


database/migrations/2020_03_22_1424_add_pkey_to_device_graphs.php
```diff
@@ -0,0 +1,48 @@
...
+{
+
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        DB::table('device_graphs')->delete();
```
Jellyfrog: This can't be right?

rkojedzinszky: Actually, that table might contain duplicates, as no constraint prevented that. Thus, clear the table, then add the primary key. Also, poller modules will fill the table upon next run. Am I right?

murrant: @rkojedzinszky you can run through and delete duplicates first.

murrant:
IMO, delete duplicates, add primary key id, add unique index on device_id and graph.

database/migrations/2020_03_22_1424_add_pkey_to_device_graphs.php
rkojedzinszky: Unfortunately, on a table without a primary key, you cannot delete just duplicates. You cannot identify them :)

murrant: Add a new column with a primary key first then.

rkojedzinszky: Unfortunately, I am not a php developer, and perhaps, a bigger refactor would be needed here, which I cannot handle. The original problem involves a table without primary key and inserting into it with `insert ignore`. This breaks librenms on percona xtradb. Perhaps, the real solution is somewhere else.

PR #11330, opened 2020-03-23
Update SMART extension documentation

raphaelyancey:
The documentation in the comments of the script (https://github.com/librenms/librenms-agent/raw/master/snmp/smart) is clearer regarding the device line format.

Also reformatted a bit and fixed the sudoers allowed commands.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Applications.md
+++ b/doc/Extensions/Applications.md
@@ -1652,9 +1652,10 @@ line with out a matched variable or # are treated as a disk.
 ```
 #This is a comment
 cache=/var/cache/smart
-smartctl=/usr/local/sbin/smartctl
-useSN=0
+smartctl=/usr/bin/env smartctl
+useSN=1
 ada0
+ada1
 da5 /dev/da5 -d sat
 twl0,0 /dev/twl0 -d 3ware,0
 twl0,1 /dev/twl0 -d 3ware,1
--- a/doc/Extensions/Applications.md
+++ b/doc/Extensions/Applications.md
@@ -1629,36 +1629,36 @@ hide_monitoring_account = With this Boolean you can hide the Account which you
 
 ## SNMP Extend
 
-1: Copy the Perl script, smart, to the desired host. `wget
-https://github.com/librenms/librenms-agent/raw/master/snmp/smart -O
-/etc/snmp/smart`
+1. Copy the Perl script, smart, to the desired host.
 
-2: Run `chmod +x /etc/snmp/smart`
+```wget https://github.com/librenms/librenms-agent/raw/master/snmp/smart -O /etc/snmp/smart```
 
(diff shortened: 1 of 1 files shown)


doc/Extensions/Applications.md
```diff
@@ -1629,52 +1629,56 @@ hide_monitoring_account = With this Boolean you can hide the Account which you
...
 ```
 #This is a comment
 cache=/var/cache/smart
-smartctl=/usr/bin/env smartctl
-useSN=1
+smartctl=/usr/local/sbin/smartctl
+useSN=0
 ada0
-ada1
+da5 /dev/da5 -d sat
```
SourceDoctor: not really agree with your changes.
please let Options on default values in example
Also keep `ada1` as second example for direct device access.

SourceDoctor:
i personally would keep `/usr/bin/env smartctl` as it is,

if smartctl is located on /usr/sbin/ , /usr/local/sbin/ depends on system.
So a change here seems to be not useful, because the next one will change it matching to his system, then.

doc/Extensions/Applications.md
raphaelyancey: What helped me the most was actually:
```
A disk line is can be as simple as just a disk name under /dev/. Such as in the config above
The line "ada0" would resolve to "/dev/ada0" and would be called with no special argument. If
a line has a space in it, everything before the space is treated as the disk name and is what
used for reporting and everything after that is used as the argument to be passed to smartctl.
```

... which describe the format very clearly. Example lines aren't clear on their own, pretty opaque to say the least.

Anyway this is copypasted from the script itself, don't you think it should match?

raphaelyancey:
> i personally would keep `/usr/bin/env smartctl` as it is,
> 
> if smartctl is located on /usr/sbin/ , /usr/local/sbin/ depends on system.
> So a change here seems to be not useful, because the next one will change it matching to his system, then.

I agree and will change it.

doc/Extensions/Applications.md
SourceDoctor: if you update Documentation, it's very welcome.
Examples should match default values.
It's correct the example in agent script should be fixed also.

Netherless the second example device 'ada1' should stay

raphaelyancey: Okay, will do.

> Netherless the second example device 'ada1' should stay

What information does `ada1` add exactly? As `ada0` is already a "direct device access" example in the same format.

SourceDoctor: ada1 is an second disc, for seeing how the next disc has to be added

raphaelyancey:
Looks like the list numbering is messed up when rendered in Markdown :/

murrant:

PR #11295, opened 2020-03-16
Fix bills api output for period=previous

Urth:
Include static fields from the bills table in the previous period output.
Check lowercase(bill_type) because the history table uses CDR and Quota.
Update SQL IF() because the bills table always uses lowercase.

Because of these changes the calculation of the fields allowed, used,
percent and overuse for period=previous is now enabled.

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)
- [x] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #11352, opened 2020-03-26
Fix bills api percentage calculation with cdr/quota of zero

Urth:
PHP allows division by zero but the api will fail when trying to encode
the response to json as reported at
https://github.com/librenms/librenms/pull/11295#issuecomment-604290389

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1382,13 +1382,21 @@ function list_bills(\Illuminate\Http\Request $request)
         if (strtolower($bill['bill_type']) == "cdr") {
             $allowed = format_si($bill['bill_cdr'])."bps";
             $used    = format_si($rate_data['rate_95th'])."bps";
-            $percent = round(($rate_data['rate_95th'] / $bill['bill_cdr']) * 100, 2);
+            if ($bill['bill_cdr'] > 0) {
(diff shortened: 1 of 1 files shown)


includes/html/api_functions.inc.php
```diff
@@ -1382,13 +1382,21 @@ function list_bills(\Illuminate\Http\Request $request)
         if (strtolower($bill['bill_type']) == "cdr") {
             $allowed = format_si($bill['bill_cdr'])."bps";
             $used    = format_si($rate_data['rate_95th'])."bps";
-            $percent = round(($rate_data['rate_95th'] / $bill['bill_cdr']) * 100, 2);
+            if ($bill['bill_cdr'] > 0) {
+                $percent = round(($rate_data['rate_95th'] / $bill['bill_cdr']) * 100, 2);
+            } else {
+                $percent = "-";
```
SourceDoctor: wouldn't it not be better to replace this `$percent = "-";` with `$percent = null;` ?

Urth: I should have. But I saw the overuse variable is set to "-" when it cannot be calculated so it made me doubt myself. (Overuse is a string with a number, percentage is a float)

murrant:

PR #11353, opened 2020-03-26
implement watchdog to librenms-service

willhseitz:
Added watchdog fork to librenms-service to check if poller log file was written to since last polling period.
Also added Redis socket_timeout to allow Redis connections to fail quicker. This is useful when calling self.restart() with a dead Redis server.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/__init__.py
+++ b/LibreNMS/__init__.py
@@ -236,13 +236,14 @@ class RedisLock(Lock):
         if redis_kwargs.get('sentinel') and redis_kwargs.get('sentinel_service'):
             sentinels = [tuple(l.split(':')) for l in redis_kwargs.pop('sentinel').split(',')]
             sentinel_service = redis_kwargs.pop('sentinel_service')
-            kwargs = {k: v for k, v in redis_kwargs.items() if k in ["decode_responses", "password", "db"]}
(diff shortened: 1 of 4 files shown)


willhseitz:
tested on my production cluster without issues. I'm curious about smaller installations with something like <10 devices and if the watchdog would trigger prematurely. may need to set the watchdog timer to 2x poll period

murrant:
Looks good.

Only concern I have is the timeout of 60s, could that be a bit short when using a cluster that may be resharding or something?  What is the default?

willhseitz:
The timer itself runs every polling period, equal to service_poller_frequency. The log file check also uses service_poller_frequency as the check for how often the log file needs to be written to. So out of the box, it would be 5 minutes.
Also, the watchdog is disabled by default, and only enabled with service_watchdog_enabled

willhseitz:
Found an issue when applying the patch to the rest of my poller cluster. I have the Redis timeout default to the alert frequency, but if alerting is disabled or set to 0, the redis timeout becomes 0 and always fails. 
My idea here was that the alerting frequency is probably the most frequent timer, and the timeout should match the lowest. That won't work if alerting is disabled on a poller

willhseitz:
Fixed this by forcing redis_timeout to 60 is alerting frequency is 0

murrant:

PR #11354, opened 2020-03-26
Delete ports via eloquent event

murrant:
Chunk delete during purge all operations so we don't use too much memory.
Also deletes all port related rrd files instead of just the main one.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/app/Models/Port.php
+++ b/app/Models/Port.php
@@ -4,6 +4,7 @@ namespace App\Models;
 
 use DB;
 use Illuminate\Database\Eloquent\Builder;
+use LibreNMS\Data\Store\Rrd;
 use LibreNMS\Util\Rewrite;
 use Permissions;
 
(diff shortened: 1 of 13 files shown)


app/Models/Port.php
```diff
@@ -12,7 +13,40 @@ class Port extends DeviceRelatedModel
...
+            $port->stp()->delete();
+            $port->vlans()->delete();
+
+            // dont have relationships yet
+            DB::table('juniAtmVp')->where('port_id', $port->port_id)->delete();
+            DB::table('ports_perms')->where('port_id', $port->port_id)->delete();
+            DB::table('links')->where('local_port_id', $port->port_id)->orWhere('remote_port_id', $port->port_id)->delete();
+            DB::table('ports_stack')->where('port_id_low', $port->port_id)->orWhere('port_id_high', $port->port_id)->delete();
+
+            \Rrd::purge($port->device->hostname, \Rrd::portName($port->port_id)); // purge all port rrd files
```
Jellyfrog: Tests fail because `$port->device` is undefined, would be cool if we could remove port RRD without involving device

murrant: Can't remove the rrd without knowing the hostname.  We do use eager loading, so it is only one extra query (per chunk).

murrant: ->device was undefined or null?  because I could see null if the device got deleted orphaning the ports.  Not sure how it would be undefined though.

Jellyfrog:
@murrant Should be merge? Want this in as many nightly as possible

includes/functions.php
```diff
@@ -453,12 +453,13 @@ function delete_device($id)
...
     dbQuery("DELETE `ipv6_addresses` FROM `ipv6_addresses` INNER JOIN `ports` ON `ports`.`port_id`=`ipv6_addresses`.`port_id` WHERE `device_id`=?", array($id));
 
-    foreach (dbFetch("SELECT * FROM `ports` WHERE `device_id` = ?", array($id)) as $int_data) {
-        $int_if = $int_data['ifDescr'];
-        $int_id = $int_data['port_id'];
-        delete_port($int_id);
-        $ret .= "Removed interface $int_id ($int_if)\n";
-    }
+
+    \App\Models\Port::where('device_id', $id)->with('device')->chunk(100, function ($ports) use ($ret) {
```
Jellyfrog, line 457: maybe add `->select(['port_id', 'device_id'])`

murrant: port label is used below.

murrant:
I believe this is ready.

murrant:

PR #11359, opened 2020-03-29
Change print-date-selector for apache alias

mpikzink:
In my setup with 
```Alias /librenms /opt/librenms/html```
It's not possible to manipulate the graphs from/to times.
The subdir is in the addressfield double on submit. This PR fix this problem.

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/html/print-date-selector.inc.php
+++ b/includes/html/print-date-selector.inc.php
@@ -2,7 +2,8 @@
 
 use LibreNMS\Config;
 
-$request_uri = rtrim(Config::get('base_url', ''), '/') . $_SERVER['REQUEST_URI'];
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: includes/html/print-date-selector.inc.php [hunk 1]
```diff
@@ -2,7 +2,7 @@
 
 use LibreNMS\Config;
 
-$request_uri = rtrim(Config::get('base_url', ''), '/') . $_SERVER['REQUEST_URI'];
+$request_uri = $_SERVER['REQUEST_URI'];
```
murrant: use the url helper `url()->current()`  This will give the proper url.

Guigui37:
I have the same problem, and this patch solved it.
But why it is still not commited to master ?

murrant:
Because this patch would break other users.

PR #11380, opened 2020-04-04
Widget Eventlog Sensors Link and Mouseover functionality

SourceDoctor:
Widget Eventlog
Now all Sensor Events are linked and have mouseover functionality

image

#### Please note

#### Testers


murrant:
@SourceDoctor FYI, there is a morph relationship set up for eventlog that will return the correct type.  Try $event->related.

SourceDoctor:
@SourceDoctor 
i know, but if i use them i have to add every single sensor_class to AppProvider.
That's anyoing and new sensors coming in future have to be added also.

so i decided to do this generic way

also i wrote a comment into code to explain my decision
https://github.com/librenms/librenms/pull/11380/files#diff-7310712ba0b8b5ffda493c7e37f6e141R99

murrant:
The way you've coded it, it will perform N queries, if you can use the relationship to eager load it will perform 1 or 2 queries.

Hmm, not sure why some use "sensor" and others use class, perhaps we could improve consistency.
```
>>> Eventlog::whereNotNull('reference')->groupBy('type')->pluck('type')->implode(', ')
=> "interface, mempool, processors, reboot, sensor, Temperature, Voltage"
```

SourceDoctor:
don't know how/where to implement your query example

i can do it this way:
in app/Http/Controllers/Table/EventlogController.php
```
} elseif (in_array($eventlog->type, [Fanspeed, Temperature])) {
 $sensor = $eventlog->related;
 sensorlink ...
```

and in app/Providers/AppServiceProvider.php
```
            ...
            'Temperature' => \App\Models\Sensor::class,
            'Fanspeed' => \App\Models\Sensor::class,
            ...
```
But all sensor types have to be declared specific. Not really elegant in my eyes.
In this example Temperature and Fanspeed will be detected as Sensor and handled.
Voltage will not handled as Sensor cause it's declared.

murrant:
I would do it more like this:

in app/Http/Controllers/Table/EventlogController.php
```php
if (isset($eventlog->related)) {
...
    } elseif (get_class($eventlog->related) == 'App\Models\Sensor') {
        $sensor = $eventlog->related;
        sensorlink ...
```
and in app/Providers/AppServiceProvider.php
```php
foreach (Sensor::getTypes() as $type) {
    $map[$type] = \App\Models\Sensor::class;
}
```

SourceDoctor:
that's nearly the way i searched for :)

SourceDoctor:
ok, tried
in debug query i see 12 requests ... 1 would be ideal ...
don't know how to make it "more static"

SourceDoctor:
@murrant 
hoping this is what you meant

app/Models/Sensor.php
```diff
@@ -56,7 +56,7 @@ public function icon()
 
     public static function getTypes()
     {
-        return self::groupBy('sensor_class')->pluck('sensor_class')->toArray();
+        return collect(self::groupBy('sensor_class')->pluck('sensor_class')->toArray());
```
murrant, line 59: Uh, you change from collection to array, then back to collection here.

murrant:
Nope, what I mean is you should have an explicit list of all sensor types.  This will be much faster than querying the database.

app/Models/Sensor.php
```diff
@@ -54,6 +54,11 @@ public function icon()
         return collect(self::$icons)->get($this->sensor_class, 'delicius');
     }
 
+    public static function getTypes()
+    {
+        return ['Voltage', 'Temperature', 'Fanspeed'];
```
SourceDoctor, line 59: @murrant 
you meant this?

murrant: Yes, but probably need all the types. :)

SourceDoctor: did some homework :D

SourceDoctor:
@murrant 
ready 4 merge?

SourceDoctor:
@murrant 
good now? :-)

app/Providers/AppServiceProvider.php
```diff
@@ -74,12 +75,16 @@ private function bootCustomBladeDirectives()
 
     private function configureMorphAliases()
     {
-        Relation::morphMap([
+        $sensor_types = [];
+        foreach (collect(Sensor::getTypes()) as $type) {
+            $sensor_types[ucwords($type)] = \App\Models\Sensor::class;
```
murrant, line 80: The event_log code uses ucfirst().

I wonder if we should change the eventlog to be lowercase.  We could use translations to have nice display names so instead of `Power_factor` > `Power Factor`

The collect() call here is not needed.

includes/polling/functions.inc.php
```diff
@@ -199,10 +199,10 @@ function record_sensor_data($device, $all_sensors)
         // FIXME also warn when crossing WARN level!
         if ($sensor['sensor_limit_low'] != '' && $prev_sensor_value > $sensor['sensor_limit_low'] && $sensor_value < $sensor['sensor_limit_low'] && $sensor['sensor_alert'] == 1) {
             echo 'Alerting for '.$device['hostname'].' '.$sensor['sensor_descr']."\n";
-            log_event("$class {$sensor['sensor_descr']} under threshold: $sensor_value $unit (< {$sensor['sensor_limit_low']} $unit)", $device, $class, 4, $sensor['sensor_id']);
```
murrant, line 202: Why did you remove this?

SourceDoctor: i removed from log event `'$sensor['sensor_descr']'`, because it's used in `type` so sensor_type can be seen now there.

app/Providers/AppServiceProvider.php
SourceDoctor: if i do it without ucwords, no sensor will be found. Reason is the niced sensor types in eventlog.related

collect is removed

includes/polling/functions.inc.php
murrant:
```
Voltage vddgfx under threshold: 0.725 (< 0.8075 )
Voltage Vcore above threshold: 1.072 (> 1.0488 )
```

If we remove the second word how can we know which voltage this is for example?

app/Providers/AppServiceProvider.php
murrant: you had no problem removing the sensor_descr from the event_log call, but don't want to change the case there?

includes/polling/functions.inc.php
murrant: I don't think you can remove it.

murrant: You could maybe try to remove text duplication if that what you mean.
Like:
```php
$descr = $class == $sensor['sensor_descr'] ? $class : "$class {$sensor['sensor_descr']}";
```

murrant:

PR #9426, opened 2018-11-07
Improve Junos state sensor discovery

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [X] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

Tested on MX and EX series, works fine.
Skip slots that are empty and pull all sensors in jnxFruTable, data from old code is preserved.

PR #11381, opened 2020-04-05
Fix routing engine discovery on standalone JunOS devices

TheMysteriousX:
In #9426, the sensor discover code was replaced with YAML.
Unfortunately it treats the FRU table as standalone OIDs, and did not
implement the filtering that was present in the original code.

This change causes LibreNMS to fully walk the FRU table, which allows us
to silence alerting on devices with an 'unknown' status. Without this
change, standalone devices have a spurious routing engine with state 1,
which triggers false harware failure alerts.

As we're changing the name of the sensors, this causes a loss of sensor
history.

This has been brought up on community a couple of times:
  https://community.librenms.org/t/juniper-qfx-routing-engine-state-alert/6152
  https://community.librenms.org/t/spurious-failed-routing-engine-from-some-junos-devices/6208

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/discovery/junos.yaml
+++ b/includes/definitions/discovery/junos.yaml
@@ -163,12 +163,12 @@ modules:
                         - { value: 2, descr: off, graph: 1, generic: 0 }
                         - { value: 3, descr: on, graph: 1, generic: 2 }
                 -
-                    oid: jnxFruState
+                    oid: jnxFruTable
                     value: jnxFruState
                     num_oid: '.1.3.6.1.4.1.2636.3.1.15.1.8.{{ $index }}'
-                    descr: '{{ $jnxFruName }}'
-                    skip_values: 2
-                    index: '{{ $index }}'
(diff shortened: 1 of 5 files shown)


TheMysteriousX:
In the FRU table, a state of 1 means "this slot is valid on some platforms, but not on this one".
A state of 2 means "this slot is valid on this platform, but it has no module physically inserted"

I've added skip values for both states, I'm not tied to that decision if that's not standard on other platforms.

TheMysteriousX:
Need someone else to verify, but the test data in `/opt/librenms/tests/data/junos_qfx5100.json` appears to be broken. Running save-test-data.php on the current head with no modifications results in changes:

  https://gist.github.com/TheMysteriousX/326795702dc0fb53c02f6d97a1745027#file-junos_qfx5100-json

None of the other sets of test data seem to be affected.

murrant:

PR #11385, opened 2020-04-06
Addhost tests

nimrof:
Added tests for addhost.php cli tool and addhost function.
Made some changes to addhost and createhost function in order to make addhost run without error when the additional parameter has default value 'array()'

There are some test in addhostCliTest.php for $device->authname that is commented out,
the tests fail, but if i read the help text correct, they should work, but i might be wrong

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
I'd like to see addhost.php go away in favor of `./lnms device:add`

tests/addhostCliTest.php
```diff
@@ -0,0 +1,201 @@
...
+
+    public function tearDown() : void
+    {
+        exec("./delhost.php ".$this->hostName." ");
+        parent::tearDown();
+    }
+
+    public function testCLIsnmpV1()
+    {
+        exec("./addhost.php -g poller_group -f -p ifIndex ".$this->hostName." community v1 111 tcp");
```
murrant, line 64: This should be `\Artisan::call('device:add', ...)`

murrant:
I really appreciate all the tests!

tests/addhostCliTest.php
nimrof: Sorry for late reply.
I will do the changes

tests/addhostCliTest.php
```diff
@@ -0,0 +1,155 @@
...
+            $this->assertNotNull($device);
+
+            $this->assertEquals($mode, $device->transport, "Wrong snmp transport (udp/tcp) ipv4/ipv6");
+            exec("./delhost.php ".$host." ");
+        }
+    }
+
+    public function testSnmpV3AuthProtocol()
+    {
+//        $modes = array('md5', 'sha', 'sha-512', 'sha-384', 'sha-256', 'sha-224');
```
nimrof, line 113: I commented out sha-512, sha-384, sha-256, sha-224 as the test will fail

nimrof:
@murrant 
Hi,
Tried using device:remove, but it printed some text to the console, so i did not use it.
I can always add a quiet -q mode to device:remove to fix it

murrant:
The problem is device:remove passes through the output of addhost.php rather than using the standard output functions, so -q won't work until it is converted to remove the device directly instead of being a passthrough.

murrant:
oh, it isn't using the testing database to remove the devices...

murrant:
Ok, this should be good.  remove is not needed as the transaction is rolled back on teardown.

That is why it always said host not found.

murrant:

PR #11388, opened 2020-04-06
Fix device creation using overwrited ip

Alex131089:
`deviceArray` was creating a device missing the `overwrite_ip` field, so it was not used by `isSNMPable`

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -525,7 +525,7 @@ function addHost($host, $snmp_version = '', $port = '161', $transport = 'udp', $
 
     // check if we have the host by IP
     $overwrite_ip = null;
-    if (!empty($overwrite_ip)) {
+    if (!empty($additional['overwrite_ip'])) {
         $overwrite_ip = $additional['overwrite_ip'];
         $ip = $overwrite_ip;
     } elseif (Config::get('addhost_alwayscheckip') === true) {
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -523,12 +523,10 @@ function addHost($host, $snmp_version = '', $port = '161', $transport = 'udp', $
         throw new InvalidPortAssocModeException("Invalid port association_mode '$port_assoc_mode'. Valid modes are: " . join(', ', get_port_assoc_modes()));
     }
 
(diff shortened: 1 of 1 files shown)


SourceDoctor:
could you retrigger checks please? Travis runs in a timeout

Alex131089:
@SourceDoctor me? How? I don't have Travis account currently.

SourceDoctor:
change a little bit .. maybe a comment or so and commit again, so Travis is triggered again

murrant:
@Alex131089 what is your reason for using Overwrite IP? just curious.

Alex131089:
@murrant simple reason, unreliable DNS records (could also be no dns connectivity but not in this case).
Btw, there's probably more places where `overwrite_ip` is overlooked, for example the HTTP / SSH / Telnet buttons should use this too when set.

murrant:
@Alex131089 We need to change the way it is implemented, it isn't right.  I have been considering removing it completely since you can just add devices by IP.

Alex131089:
@murrant it is still nice to have the hostname, which might not be accurate if taken from SNMP in some case 😅
Also system/snmp hostname or the set hostname might not be the fqdn.
I don't think there's (full) monitoring systems without hostname or ip?

murrant:
why not just set the display name then?

Alex131089:
I didn't notice the display name field, I only played with LibreNMS a few hours so far, using the Docker version, without `smtp` and `syslog-ng` images (and according to the disk usage and server load I observed, it certainly requires tuning / architecture adjustments to scale a bit, but I'm not sure `memcached` & `rrdcached` where used though, I assumed the proposed configuration made use of them).
But personally I like when tools leaves choice to the user (an exemple: Netbox imposes some fields, some some optional, and let the user create custom fields ; but that's an extreme case :) ).

murrant:
We have people monitoring 1600 or even 3000 devices.

Requiring the user to figure out a whole bunch of things to monitor a device is not the most friendly thing.  The goal is to lower the barriers to entry, but allow lots of customization.

murrant:

PR #11400, opened 2020-04-09
Update Airos.php to match with Airos 8.x.x

facuxt:
Oid changed from AirMaxQuality (not used anymore), to Airtime

The old airos 5 and 6 versions, used to have AIRMAX value reported. The new version doesnt have such calculation, instead airtime is an excelent indicator that is missing in librenms.

The ideal situation, should be to include a new section in the "wireless tab" that includes modulation and airtime in order to verify the health of the wireless link, but not exactly sure how to do that.

image

DO NOT DELETE THIS TEXT

#### Please note

#### Testers


Jellyfrog:
Thanks for your contribution, but we don't want to remove support for old airos, it should support both in that case

facuxt:
> Thanks for your contribution, but we don't want to remove support for old airos, it should support both in that case

There is any documentation i can read in order to get both sensors working? or what do you suggest me to do?

LibreNMS/OS/Airos.php
```diff
@@ -158,9 +158,11 @@ public function discoverWirelessPower()
      */
     public function discoverWirelessQuality()
     {
-        $oid = '.1.3.6.1.4.1.41112.1.4.6.1.3.1'; //UBNT-AirMAX-MIB::ubntAirMaxQuality.1
+        //$oid = '.1.3.6.1.4.1.41112.1.4.6.1.3.1'; //OLD UBNT-AirMAX-MIB::ubntAirMaxQuality.1
```
murrant, line 161: instead of removing the old AirMaxQuality, you should try to discover two sensors here.  Only ones that exist will be added.

murrant:
Modulation is a state sensor, on most devices, which is kind of unfortunate it doesn't display in the wireless section, but it works.

facuxt:
Nice @murrant !

murrant:
@facuxt can you update or add test data for this as requested?

facuxt:
> @facuxt can you update or add test data for this as requested?
i tested it but it only shows graphs for airtime, dont know if i miss a step... 
image

Then, seeing the documentation, there is available the sensor type WirelessUtilization, so maybe it would be wise to implement it... what do you suggest?

facuxt:
@murrant  i updated the file airos.php to use wirelessUtilization interface, and it seems to be working OK. I attach screenshot below
image

PipoCanaja:
Hi @facuxt 
Could you add the test data to this PR, according to the documentation : 
https://docs.librenms.org/Developing/os/Test-Units
Thanx

facuxt:
> Hi @facuxt
> Could you add the test data to this PR, according to the documentation :
> https://docs.librenms.org/Developing/os/Test-Units
> Thanx

im having this errors trying to test.
image

Nevertheless i uploaded a testfile airos.snmprec

PipoCanaja:
Hi @facuxt 
A test file was already existant for Aireos. You need to create a new one (we call it a variant) instead of touching the existing one. 
In order to do that, please revert the aireos.snmprec file as it was before, and create a new one with `./script/collect-snmp-data.php -h xxxx -v aireos8`
That will avoid breaking the existing one, and both test will be run.

facuxt:
sorry for the delay, im was really busy this past week. I reverted the test file to its original contents (airos.snmprec) and uploaded a new 'variant' one (airos_airos8.snmpred)

facuxt:
👏👏👏👏👏👏

murrant:

PR #11407, opened 2020-04-10
Use which instead of whereis

Jellyfrog:
which searches in PATH
whereis -b doesn't exist on some OSes

DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Config.php
+++ b/LibreNMS/Config.php
@@ -535,13 +535,10 @@ class Config
     public static function locateBinary($binary)
     {
         if (!str_contains($binary, '/')) {
-            $output = `whereis -b $binary`;
-            $list = trim(substr($output, strpos($output, ':') + 1));
-            $targets = explode(' ', $list);
-            foreach ($targets as $target) {
-                if (is_executable($target)) {
-                    return $target;
-                }
+            $target = trim(`which $binary`);
+
+            if (!empty($target)) {
+                return $target;
             }
         }
         return $binary;


murrant:
If I remember some platforms don't have which, which is why I used whereis.

SourceDoctor:
ok then it sounds like a chicken egg problem ...
i personally found everywhere which

murrant:
We can try and see bug reports, if I recall, it was BSD that had issue.

SourceDoctor:
i'm with it

SaaldjorMike:
Looks like this PR broke the installation @Rosiak and I have running to serve stats on https://stats.kollegienet.dk

I updated our installation to use the commit right before this change which works just fine. As soon as I use what was introduced in this PR, all our devices are seen as "down". If I use the commit before this was introduced all devices are "up" again.

We're running this on CentOS 7.

I propose that we revert this PR until we understand what broke the installation.

I'm happy to share any details needed to debug this any further, although I believe we run a rather simple CentOS installation.

SourceDoctor:
@SaaldjorMike 
is 'which' existing on your machines?

murrant:
I think I remember the problem is PATH isn't set for cron tasks

SourceDoctor:
if path is not set, system should use default path?
also in cron there are only php scripts ...

SaaldjorMike:
> is 'which' existing on your machines?

Yes.

```verbatim
which --version
GNU which v2.20, Copyright (C) 1999 - 2008 Carlo Wood.
GNU which comes with ABSOLUTELY NO WARRANTY;
This program is free software; your freedom to use, change
and distribute this program is protected by the GPL.
```

```
[root@librenms ~]# whereis -b which
which: /usr/bin/which
```

> I think I remember the problem is PATH isn't set for cron tasks

Perhaps, although a quick echo of `$PATH`, from a script invoked similar to `daily.sh` does seem to give me: `/usr/bin:/bin`

SaaldjorMike:
> also in cron there are only php scripts ...

Maybe I missed a change at some point? We have the following cron entries:
```cron
33   */6  * * *   librenms    /opt/librenms/cronic /opt/librenms/discovery-wrapper.py 1
*/5  *    * * *   librenms    /opt/librenms/discovery.php -h new >> /dev/null 2>&1
*/5  *    * * *   librenms    /opt/librenms/cronic /opt/librenms/poller-wrapper.py 16
15   0    * * *   librenms    /opt/librenms/daily.sh >> /dev/null 2>&1
*    *    * * *   librenms    /opt/librenms/alerts.php >> /dev/null 2>&1
*/5  *    * * *   librenms    /opt/librenms/poll-billing.php >> /dev/null 2>&1
01   *    * * *   librenms    /opt/librenms/billing-calculate.php >> /dev/null 2>&1
*/5  *    * * *   librenms    /opt/librenms/check-services.php >> /dev/null 2>&1
```

SourceDoctor:
cron script looks good

PR #11424, opened 2020-04-15
BGP Polling: Add error code management

kedare:
DO NOT DELETE THIS TEXT

Poll BGP errors codes from devices (Tested on IOS XR, JunOS, Arista EOS)
Display them on the routing pages

First line of status is generated from the error codes and subcodes (if supported)
The second line is that the device return as error text (if supported)

Screenshot 2020-04-15 at 11 47 35

#### Please note

#### Testers


kedare:
I am adding the missing test data

kedare:
I am still adding test data, I suppose I can just add blank data like most of the other values ? (new fields to Null) as I can't add real test data (as it looks like the cisco ios is inherited by a lot of others OS I don't have access to)

kedare:
Tests passed, I can't do much for the code climate issues however ?

database/migrations/2020_04_13_150500_add_last_error_fields_to_bgp_peers.php
```diff
@@ -0,0 +1,39 @@
...
+     *
+     *
+     *
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('bgpPeers', function (Blueprint $table) {
+            $table->dropColumn('bgpPeerLastErrorCode');
```
murrant, line 34: This should be one dropColumn() with an array of all columns to drop.

includes/functions.php
```diff
@@ -2524,3 +2524,94 @@ function oxidized_node_update($hostname, $msg, $username = 'not_provided')
...
+/**
+ * @params int code
+ * @params int subcode
+ * @return string
+ * Take a BGP error code and subcode to return a string representation of it
+ */
+function describe_bgp_error_code($code, $subcode)
+{
+    // https://www.iana.org/assignments/bgp-parameters/bgp-parameters.xhtml#bgp-parameters-3
+    $codes = [
```
murrant, line 2538: The best way to do this is to create a new language file (resources/lang/en/bgp.php) and add all these messages there, that way you can look them up like `trans('bgp.error_codes.$code")`  It will also allow translation :)

includes/functions.php
```diff
@@ -2524,3 +2524,94 @@ function oxidized_node_update($hostname, $msg, $username = 'not_provided')
...
+    }
+
+    if (array_key_exists($code, $subcodes) && array_key_exists($subcode, $subcodes[$code])) {
+        $subcode_message = $subcodes[$code][$subcode];
+    } else {
+        $subcode_message = "";
+    }
+
+    if ($subcode_message == "") {
+        return "[" . $code . " " . $subcode . "]: " . $code_message;
```
murrant, line 2613: Does including the code and subcode improve things?  I felt a little confused as to what the numbers were looking at the webui.

includes/polling/bgp-peers.inc.php
```diff
@@ -243,7 +266,7 @@
                                 // if parsing fails, leave the data as-is
                             }
                         }
-
+
```
murrant, line 269: oopsie whitespace

murrant:
Looks good, just some improvements.

It might be a good idea to include some updated test data with the new fields.  They seem to all be null.

includes/functions.php
kedare: Indeed we can remove them from the WebUI

kedare:
For the test data I could potentially add some for Arista, JunOS, and IOS-XR, but I will have to add it by hand on the test data, it's not very clear how to do that on the snmp test data ?

includes/polling/bgp-peers.inc.php
kedare: Fixed

includes/functions.php
kedare: Done

includes/functions.php
kedare: Done

database/migrations/2020_04_13_150500_add_last_error_fields_to_bgp_peers.php
kedare: Done

murrant:
Inserting additional data into existing snmprec files can be tricky.  If you do it by hand, make sure the oid is always increasing.

kedare:
Maybe it would be more interesting to let someone that has a lab environment to add new test data ? I don't have access to non-production environment and I am not allowed to dump data from production for the tests (for the risk of leaking data)

murrant:
It would be good to have at least one with test data.

You could manually add the oids, but that seems a little tedious.

zombah:
@kedare @murrant There is exist repo with snmpsim test data, i didn't checked it myself, but maybe it contain something useful for cases like this - https://github.com/etingof/snmpsim-data/tree/master/data/network/router

kedare:
I added some test data, let me know if this is enough

murrant:
Thanks @kedare , nice work :)

murrant:

murrant:

https://community.librenms.org/t/v1-63-high-cpu-on-arista-switches-with-bgp-peers-module-enabled/11947/4

PR #11428, opened 2020-04-16
Remove Laravel helpers

murrant:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport.php
+++ b/LibreNMS/Alert/Transport.php
@@ -2,6 +2,7 @@
 
 namespace LibreNMS\Alert;
 
+use Illuminate\Support\Str;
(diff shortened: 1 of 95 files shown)


SourceDoctor:
hrhr next step into full Laravel :D 
but why no `use Illuminate\Support\Str;`  and then Str::<method> in file?
Instead of using often multiple times `\Illuminate\Support\Str::<method>` in whole file?

murrant:
@SourceDoctor because I just used the IDE to inline the method and if it wasn't already imported, it didn't import it.

I could go through and add the import to more manually.

murrant:
There you go @SourceDoctor  :)

PR #11432, opened 2020-04-17
Added support for Peplink Pepwave & FusionHub

jozefrebjak:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


murrant:
Looks pretty good overall, nice work.

Might be able to improve the wireless discovery by walking `cellularSignalInfoTable` and storing it in a local property instead of walking individual oid of that table multiple times.

jozefrebjak:
> Looks pretty good overall, nice work.
> 
> Might be able to improve the wireless discovery by walking `cellularSignalInfoTable` and storing it in a local property instead of walking individual oid of that table multiple times.

Nice idea, but i don't have experience with that. There is few more sensors, but some of them are not supported for now by LibreNMS. I want to add some new wirelless Cellular sensors in few days. For now is this everything what we need. We are switching from MPLS to a solution with Peplink where we are bundling SIM Cards together in one project what we are not very happy, but we need to operate that.

LibreNMS/OS/Pepwave.php
```diff
@@ -0,0 +1,102 @@
...
+    WirelessRssiDiscovery,
+    WirelessSinrDiscovery
+{
+    public function discoverWirelessRssi()
+    {
+        $data = snmpwalk_group($this->getDevice(), 'cellularSignalRssi', 'CELLULAR');
+        $sensors = [];
+        foreach ($data as $index => $rssi_value) {
+            if ($rssi_value['cellularSignalRssi'] != '-9999') {
+                $sensors[] = new WirelessSensor('rssi', $this->getDeviceId(), '.1.3.6.1.4.1.23695.200.1.12.1.1.1.3.' . $index, 'pepwave', 'cellularSignalRssi' . $index, 'Celullar ' . ($index+1), null, 1, 1);
```
murrant, line 49: Instead of null, you need to send `$rssi_value['cellularSignalRssi']` for the current value.  This is the same for all the ones below.

includes/definitions/discovery/pepwave.yaml
```diff
@@ -0,0 +1,132 @@
...
+          group: WLC
+          index: "wlcNumApReg.{{ $index }}"
+        -
+          oid: wlcNumApOnline
+          num_oid: ".1.3.6.1.4.1.27662.4.1.1.7.{{ $index }}"
+          descr: Number of Online AP
+          group: WLC
+          index: "wlcNumApOnline.{{ $index }}"
+        -
+          oid: wlcNumAssocSta
```
murrant, line 51: This should be a wireless sensor "clients"

jozefrebjak: fixed

LibreNMS/OS/Pepwave.php
jozefrebjak: fixed

murrant:

https://community.librenms.org/t/peplink-support/3980/5

murrant:

PR #11442, opened 2020-04-20
Added functionality to set sysName override in $os.inc.php files

Martin22:
DO NOT DELETE THIS TEXT

#### Please note

#### Testers


includes/polling/os.inc.php
```diff
@@ -35,6 +35,13 @@
     set_device_location($location, $device, $update_array);
 }
 
+if ($name != $device['name']) {
```
murrant, line 38: This is the wrong check and will always update.
Also the variable should probably be `$sysName`

Martin22:
Since discovery will overwrite it, I'll try to solve it differently.

PR #11452, opened 2020-04-21
Add php-opcache settings to Performance Tuning

hanserasmus:

#### Please note

#### Testers

### DIFF
--- a/doc/Support/Performance.md
+++ b/doc/Support/Performance.md
@@ -179,7 +179,7 @@ For Apache (2.4.17 an above) set `Protocols h2 http/1.1` in the Virtualhost conf
 
 A lot of performance can be gained from setting up `php-opcache` correctly. 
 
-Memory based caching on with PHP cli will increase memory usage and slow things down and file based caching is not as fast as memory based and is more likely to have stale cache issues.
+**Note: Memory based caching with PHP cli will increase memory usage and slow things down. File based caching is not as fast as memory based and is more likely to have stale cache issues.**
 
 Some distributions allow seperate cli, mod_php and php-fpm configurations, we can use this to set the optimal config.
 
@@ -200,10 +200,10 @@ If you are having caching issues, you can clear the opcache by simply restarting
 Create a cache directory that is writable by the librenms user first:
 `sudo mkdir -p /tmp/cache && sudo chmod 775 /tmp/cache && sudo chown -R librenms /tmp/cache`
 
-Update your web PHP opcache.ini.  Possible locations: `/etc/php/7.2/cli/conf.d/opcache.ini`, `/etc/php.d/opcache.ini`, or `/etc/php/conf.d/opcache.ini` 
+Update your PHP opcache.ini.  Possible locations: `/etc/php/7.2/cli/conf.d/opcache.ini`, `/etc/php.d/opcache.ini`, or `/etc/php/conf.d/opcache.ini` 
 
 ```
-zend_extension=opcache
+zend_extension=opcache.so
 opcache.enable=1
 opcache.enable_cli=1
 opcache.file_cache="/tmp/cache/"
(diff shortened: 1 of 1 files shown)


doc/Support/Performance.md
```diff
@@ -174,3 +174,28 @@ For Nginx (1.9.5 and above) change `listen 443 ssl;` to `listen 443
...
+zend_extension=opcache
+opcache.enable=1
+opcache.enable_cli=1
+opcache.file_cache="/tmp/cache/"
+opcache.file_cache_only=0
+opcache.file_cache_consistency_checks=1
+opcache.memory_consumption=256
+```
+If you have set this on the pollers only, there is no need to restart anything. However, 
+if you have set this on the webserver, you would need to restart apache/nginx as well as
```
murrant, line 200: You don't need to restart nginx, only php-fpm

murrant:
Looks good, maybe add a comment that sometimes this will cause cached versions of PHP files to be used. (not happy with my wording for that)

hanserasmus:
Ok, but, how would one "clear" such caches (without rebooting)? Because that would be the next logical thing to add as well :-)

doc/Support/Performance.md
```diff
@@ -174,3 +174,36 @@ For Nginx (1.9.5 and above) change `listen 443 ssl;` to `listen 443
 ssl http2;` in the Virtualhost config.
 
 For Apache (2.4.17 an above) set `Protocols h2 http/1.1` in the Virtualhost config.
+
+## PHP-opcache
+
+A lot of performance can be gained from setting up `php-opcache`correctly.
```
murrant, line 180: missing space

hanserasmus: Sorry.

doc/Support/Performance.md
```diff
@@ -174,3 +174,36 @@ For Nginx (1.9.5 and above) change `listen 443 ssl;` to `listen 443
...
+opcache.file_cache="/tmp/cache/"
+opcache.file_cache_only=0
+opcache.file_cache_consistency_checks=1
+opcache.memory_consumption=256
+```
+
+When setting this on the pollers there is no need to restart anything. 
+Just note that by using file-based caching as is the functionality explained here, you might
+in extreme cases face some form of caching issue. Although LibreNMS does not use any
+long-running php services which would make above-mentioned scenario unlikely, it is not impossible.
```
murrant, line 203: This is misleading, when using file caching. It doesn't matter if the process is long-running.

doc/Support/Performance.md
```diff
@@ -174,3 +174,36 @@ For Nginx (1.9.5 and above) change `listen 443 ssl;` to `listen 443
...
+On pollers, you could reboot your machine, or a little less extreme just `rm -rf /tmp/cache` and issue
+restart the librenms service. This should again create a new directory into which OPcache will place
```
murrant, line 205: Will never need to restart the librenms service, it is python based.

doc/Support/Performance.md
hanserasmus: Should I then remove this completely?

doc/Support/Performance.md
```diff
@@ -174,3 +174,36 @@ For Nginx (1.9.5 and above) change `listen 443 ssl;` to `listen 443
...
+cache files.
+
+**Note: The file_cache method is not advised for the web servers. If you want to enable opcache on
```
murrant, line 208: This should be more prominent and note that some OS have separate php configs for cli and web.

doc/Support/Performance.md
murrant: Just this line.

Although LibreNMS does not use any
long-running php services which would make above-mentioned scenario unlikely, it is not impossible.

doc/Support/Performance.md
murrant: rm -rf /tmp/cache and restart apache or php-fpm should cover it.

hanserasmus: We are taking out file-based caching for apache or php-fpm. So it should not be there anymore at all.

murrant:
@hanserasmus I've rewritten the information as I see it.  Please go through and clarify/rewrite it further if I am not clear or you think it needs it.

murrant:
Thanks @hanserasmus I think we got it pretty good.

murrant:

PR #11465, opened 2020-04-22
Run artisan optimize after composer install

Jellyfrog:
Helps cleaning up issues with upgrades, also a small performance win.

#### Please note

#### Testers


Jellyfrog:
ping @murrant, you always have an opinion :smile:

murrant:
One thing to consider is this effectively clears the cache every night.  Maybe optimize:clear is too big of a hammer.

I was also thinking of caching routes on update.

Jellyfrog:
Yes but if you run nightly, you run nightly ;)  
view:clear is enough for laravel 6 upgrade what I've seen, but the others doesn't hurt.  
We could simply add "optimize" in the end, the only one of worry is config:cache.

murrant:
yeah config:cache should not be used unless we take precautions to help people be aware it is in effect.

Jellyfrog:
https://github.com/librenms/librenms/blob/05507526c0430816cdc0fdc1418cf7ef18be66d4/routes/web.php#L20-L22
This prevents route:cache

murrant:
@Jellyfrog yeah, we need to remove those types of statements to enable route:cache.  Should be no problem.

routes/web.php
```diff
@@ -150,11 +143,6 @@
     // demo helper
     Route::permanentRedirect('demo', '/');
 
-    // blank page, dummy page for external code using Laravel::bootWeb()
```
murrant, line 153: You can't remove this route.  You need to create a controller.

routes/web.php
```diff
@@ -52,9 +48,6 @@
 
     // old route redirects
     Route::permanentRedirect('poll-log', 'poller/log');
-    Route::get('settings/sub={tab}', function ($tab) {
```
murrant, line 55: removing this route could break some external links.  It can be changed to use a controller.

Jellyfrog: From where? It's a link to a settings page, no big deal. 
And like I wrote, it doesn't even work atm due to being processed after the real settings-route.

routes/web.php
Jellyfrog: It was added by:
https://github.com/librenms/librenms/commit/4e4ceffc0ae5750a647ec7aa946e84e952599b0c#diff-267b660c9508115406d8e6be937c61b3R62

And removed here:
https://github.com/librenms/librenms/commit/c9748b166eb0e16830eb6057e5b962a109eac428#diff-267b660c9508115406d8e6be937c61b3L62

murrant: Well that explains why every legacy page is being rendered twice!

routes/web.php
murrant: When people deep link into LibreNMS (from their internal documentation for example), changing the structure will break those links.  This is probably most common with the device link however...

I see your point, I think it got re-organized.  I'm fine with removing this.

LibreNMS/ComposerHelper.php
```diff
@@ -44,7 +44,10 @@ public static function postInstall(Event $event)
             self::setPermissions();
         }
 
+        // "optimize" doesn't clear view cache etc.
+        self::exec('php artisan optimize:clear');
         self::populateEnv();
+        self::exec('php artisan optimize');
```
murrant, line 50: Isn't "optimize" depracated?

Jellyfrog: it was added back again, https://github.com/laravel/framework/pull/23468

LibreNMS/ComposerHelper.php
```diff
@@ -44,7 +44,10 @@ public static function postInstall(Event $event)
             self::setPermissions();
         }
 
+        // "optimize" doesn't clear view cache etc.
+        self::exec('php artisan optimize:clear');
```
Jellyfrog, line 48: maybe we should do a `view:clear` only here, since `optimize` does some cleaning itself.

routes/web.php
murrant: I did some research and it is safe to remove this, sorry about that but I needed to make sure :)

.travis.yml
```diff
@@ -48,6 +48,7 @@ script:
   - php scripts/pre-commit.php -q -l
   - php scripts/pre-commit.php -q -s
   - php scripts/pre-commit.php -u --db --snmpsim --fail-fast
+  - test -z "$BROWSER_TEST" || php artisan config:clear
```
Jellyfrog, line 51: This is because dusk rewrites the .env file and it won't be picked up if it's cached.

LibreNMS/ComposerHelper.php
```diff
@@ -45,6 +45,10 @@ public static function postInstall(Event $event)
         }
 
         self::populateEnv();
+
+        // "view:cache" to clear the cached views, needed when upgrading Laravel
+        self::exec('php artisan view:cache');
```
murrant, line 50: maybe add these to composer.json instead of here?

murrant:

PR #11477, opened 2020-04-23
Only load default config when saving test data from snmpsim

TheMysteriousX:
If you're generating test data from a real device, you're probably
doing so on a production instance (or at least, that's where I
poll all of my most expensive kit from!).

This means there are likely to be a bunch of config settings that
cause test data to be generated differently between CI and your
own instance.

This change clears the if, mount and sensor filters before test
data is retrieved from SNMPSim.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Config.php
+++ b/LibreNMS/Config.php
@@ -36,13 +36,14 @@ use Log;
 class Config
 {
     private static $config;
+    private static $testMode = false;
 
     /**
      * Load the config, if the database connected, pull in database settings.
      *
      * return &array
      */
-    public static function load($defaultsOnly = false)
+    public static function load()
     {
         // don't reload the config if it is already loaded, reload() should be used for that
         if (!is_null(self::$config)) {
@@ -52,7 +53,7 @@ class Config
         // merge all config sources together config_definitions.json > db config > config.php
         self::loadDefaults();
 
-        if (!$defaultsOnly) {
+        if (self::$testMode === false) {
             self::loadDB();
             self::loadUserConfigFile(self::$config);
         }
@@ -65,6 +66,14 @@ class Config
         $config = self::$config;
 
         return self::$config;
+    }
+
+    /**
+     * Sets a flag to prevent the database and user configuration being loaded
+     */
+    public static function enableTestMode()
+    {
+        self::$testMode = true;
     }
 
     /**
--- a/LibreNMS/Config.php
+++ b/LibreNMS/Config.php
@@ -134,6 +134,24 @@ class Config
     }
 
     /**
+     * Find the default value for a configuration item
+     *
+     * @param string $key period separated config variable name
+     * @param mixed $default optional value to return if the setting is not set
+     * @return mixed
+     */
+    public static function getDefault($key, $default = null)
+    {
+        $definitions = self::getDefinitions();
+
+        if (array_key_exists($key, $definitions) && array_key_exists('default', $definitions[$key])) {
+            return $definitions[$key]['default'];
+        }
+
+        return $default;
+    }
+
+    /**
      * Unset a config setting
      * or multiple
      *
--- a/LibreNMS/Config.php
+++ b/LibreNMS/Config.php
@@ -42,7 +42,7 @@ class Config
      *
      * return &array
      */
-    public static function load()
+    public static function load($defaultsOnly = false)
     {
         // don't reload the config if it is already loaded, reload() should be used for that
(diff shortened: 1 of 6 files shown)


LibreNMS/Config.php
```diff
@@ -133,6 +133,24 @@ public static function get($key, $default = null)
...
+    /**
+     * Find the default value for a configuration item
+     *
+     * @param string $key period separated config variable name
+     * @param mixed $default optional value to return if the setting is not set
+     * @return mixed
+     */
+    public static function getDefault($key, $default = null)
+    {
+        $definitions = self::getDefinitions();
```
murrant, line 145: getDefinitions() is an expensive operation.  Doing it like this is going to parse json for every call.

murrant:
Been meaning to do something like this for awhile.

I was more thinking to load the settings without config.php would be my approach.

TheMysteriousX:
I think you're right, changing the initial config loading is probably a better approach than trying to fix it up once it's happened.

TheMysteriousX:
I think this should do the trick, given it a quick test on my live instance that was generating broken test data, and it seems to be working.

Not sure about the name "testMode" - if you have any better suggestions I can change it.

murrant:
Not bad.

What do you think about having a function loadDefaults().  That way we don't need to have a check in the normal load process.

TheMysteriousX:
Will test it in a bit, but the latest change removes the condition.

Jellyfrog:
But don't we want \*all\* modules to be enabled for snmpsim? (Especially for `.snmprec`)

TheMysteriousX:
@Jellyfrog That isn't done right now.

Without this change, the test .json files are saved with user config customisations - which also included disabled modules.

With this change, user customisations are ignored when saving the test data - i.e. it makes sure that the environment used to collect data and Travis match.

If all modules should be enabled for this, that could be done, but my feeling is that it should be.
The modules that are disabled in the default configuration are usually done so for a good reason (i.e. your iDRAC is never going to start talking BGP or sprout a wireless access point). But even in that case, you'd still want user customisations ignored when doing the collection.

@murrant I had to revise things again - at the point loadDefault() was called, Laravel isn't bootstrapped enough to actually load the config. In this iteration, I've used a PHP's equivalent of a function pointer to specify which set of config to load.

murrant:
Hmm, not working when I set `$config['bad_if_regexp'][] = '/.*/';` in my config.php

murrant:
How about just this:
```diff
--- LibreNMS/Config.php	(revision c31d58f9bbd636cf2dcbb1a55761823034a82546)
+++ LibreNMS/Config.php	(date 1587821925640)
@@ -51,8 +51,11 @@
 
         // merge all config sources together config_definitions.json > db config > config.php
         self::loadDefaults();
-        self::loadDB();
-        self::loadUserConfigFile(self::$config);
+
+        if (app()->environment() !== 'testing') {
+            self::loadDB();
+            self::loadUserConfigFile(self::$config);
+        }
 
         // final cleanups and validations
         self::processConfig();
```

TheMysteriousX:
I'll give it a try. Not sure why you're seeing that bad_if_regexp was still set, maybe it's being leaked through the legacy $config global somehow?

TheMysteriousX:
Setting app->env to testing doesn't appear to break anything (even though I've not got a .env file for 'testing', so I'll update this changeset with that, and add a prompt to the collection and pre-commit scripts noting that for correct behaviour, the app environment may need to be set to testing.

Re: bad_if_regexp, the leak is coming from the unit tests. Config::load gets called a lot.

scripts/pre-commit.php
```diff
@@ -336,6 +336,10 @@ function check_style($passthru = false, $command_only = false)
  */
 function check_unit($passthru = false, $command_only = false, $options = array())
 {
+    if (getenv('APP_ENV') !== 'testing') {
```
murrant, line 339: Or you could just `putenv('APP_ENV=testing')`

murrant:
Ok, so some tests require the database to be loaded across tests... (which is probably an error in the testing)

Without fixing the tests, this probably brings us back to something you had before.

TheMysteriousX:
I'll have a look and see how many tests are affected.

murrant:
It is just this test, because the webserver and the tests run in different processes.

murrant:
Has not solve issue where webserver tests run in a different process, so they need the db config loaded.  Feel free to re-open with a fix.

PR #11486, opened 2020-04-26
Install python requirements during daily

murrant:
install python requirements for the default python version.

#### Please note

#### Testers

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -121,7 +121,7 @@
             "Illuminate\\Foundation\\ComposerScripts::postUpdate"
         ],
         "python-requirements": [
-            "scripts/check_requirements.py || pip install -r requirements.txt"
+            "scripts/check_requirements.py || pip3 install -r requirements.txt || :"
         ]
     }
 }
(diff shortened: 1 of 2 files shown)


SourceDoctor:
nice, i like this.
But documentations should be enhanced so `python-pip` package is installed

murrant:
The main problem right now is our minimum version for redis requires python 3.  If we port the scripts to python 3, we might need the libraries for python 2 also.

SourceDoctor:
mmh... so we need 2 and 3 parallel to keep everything running?

On rewriting stuff to python3 i would ensure python code is PEP8 conform (maybe a test for it), and then rewrite Python Code to be compatible for Python 2 and 3.
After a while (with information/warning), we could switch to Python3 only and use it's full capabilities.
Hoping open PR, which is on it will be ready in near future ;-)

murrant:
true, we could just make sure to install deps for python3

murrant:
Honestly, I think we need to drop python2 support at the same time as php 7.1 @Jellyfrog

murrant:

PR #11490, opened 2020-04-26
Remove webtatic repo from CentOS install docs

hanserasmus:
I personally have no problem with them, but after a recent discussion in discord about a user asking which Repo, it got me thinking that maybe we should also stick to what the OS providers have to say? I see that (obviously) SCL and remi is on the trusted list, so shouldn't the install docs rather stick with them?

[CentOS Wiki](https://wiki.centos.org/AdditionalResources/Repositories)

I know many people are comfortable using it, and therefor still will use it and that is fine. Looking from an official point of view here, is why I made the PR. 

#### Please note

#### Testers


murrant:
The funny thing is it was only webtatic before when someone added the other two.

PR #11509, opened 2020-04-30
OS detection for Dell PowerVault MD arrays

TheMysteriousX:
I'm not 100% sure it was worth Dell implementing SNMP - the array
exposes 6 OID's outside of MIBII, and we poll both the useful ones.

At least we get an icon.

I've added two MIBs available from Dell in the MD Array resource bundle; seems neither are used by MD arrays.

#### Please note

#### Testers


TheMysteriousX:
Tests should pass if #11511 is merged.

includes/definitions/dell-powervault.yaml
```diff
@@ -0,0 +1,48 @@
+os: dell-powervault
+text: 'Dell PowerVault MD'
+type: storage
+icon: dell
+group: dell
+poller_modules:
```
murrant, line 6: FYI, setting modules in an OS prevents users from setting them globally.
The poller_modules and discovery_modules is more for enabling required modules rather than disabling everything.

As I said before, feel free to set this in your config.php

TheMysteriousX: That's fine, it's only affects polling speed - I'll remove them.

murrant:

https://community.librenms.org/t/v1-66-release-changelog-july-2020/12953/1

PR #11510, opened 2020-04-30
OS Detection for Mobileiron Core, Sentry appliances

TheMysteriousX:
This adds OS detection for MobileIron's appliances.

They're just generic linux under the hood; so generic in fact, I've had to walk the installed package list looking for the mobileiron packages to identify them.

Note for anyone with these appliances: they won't poll correctly at this time due to a rate limit of 10 packets per minute on UDP 161. I've raised this with their support desk.

#### Please note

#### Testers


includes/definitions/mobileiron.yaml
```diff
@@ -0,0 +1,33 @@
...
+    ospf: false
+    stp: false
+discovery_modules:
+    applications: true
+    bgp-peers: false
+    stp: false
+    vmware-vminfo: true
+    libvirt-vminfo: true
+processor_stacked: true
+register_mibs:
```
murrant, line 25: why are you registering mibs for another OS?

includes/definitions/mobileiron.yaml
```diff
@@ -0,0 +1,33 @@
+os: mobileiron
```
murrant, line 1: is mobleiron the best name for this OS?  it seems to just be "Core"

includes/discovery/os/mobileiron.inc.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+use Illuminate\Support\Str;
+
+// At ths time, MI don't make any customisations to the tree, so we just detect their packages
+if (Str::startsWith($device['sysDescr'], 'Linux') && Str::startsWith($device['sysObjectID'], '.1.3.6.1.4.1.8072.3.2.10')) {
```
murrant, line 6: Not enthused about this detection, but at least you have explained it.

includes/definitions/mobileiron.yaml
TheMysteriousX: As it's CentOS under the hood I took the Linux definition as a start point, so I just assumed they were needed.

Looking at the definitions though, I think they shouldn't be in the Linux definition either - this seems to be the intended OS?
https://github.com/librenms/librenms/blob/master/includes/definitions/pktj.yaml

includes/definitions/mobileiron.yaml
TheMysteriousX: The vendor doesn't give it an explicit name. MobileIron is the product line (and vendor) and Core/Sentry are the two types of appliance (core being UI/control plane, sentry being the edge).

Other systems in the same position use the vendor/product name, trendnet, raritan, opengear, etc.

I could split it into two defintions.

includes/discovery/os/mobileiron.inc.php
TheMysteriousX: Yeah, it's not ideal. I'm not sure how net-snmp decides to order the installed packages list, but it's not stable or obvious, so grabbing the whole branch of the tree is the only option.

includes/definitions/mobileiron.yaml
murrant: I think they were defined in PHP, then ported to yaml.  Yeah, they could be moved specifically to pktj

includes/definitions/mobileiron.yaml
murrant: It is fine, but we try to avoid vendor names if possible because it doesn't work very well when they have multiple product lines (os definitions)

includes/definitions/mobileiron.yaml
```diff
@@ -0,0 +1,24 @@
...
+type: appliance
+group: unix
+text: MobileIron
+ifXmcbc: true
+ifname: true
+mib_dir:
```
murrant, line 7: You might not need these other mib_dirs.  (but you need to test that first)

includes/definitions/mobileiron.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - supermicro
+    - dell
+over:
+    - { graph: device_processor, text: 'Processor Usage' }
+    - { graph: device_ucd_memory, text: 'Memory Usage' }
+    - { graph: device_storage, text: 'Storage Usage' }
+poller_modules:
```
murrant, line 14: Again, I'm not a big fan of setting modules to false in the OS definition.

includes/definitions/mobileiron.yaml
TheMysteriousX: Ok, I'll submit that as a seperate request to keep things simple.

includes/definitions/mobileiron.yaml
TheMysteriousX: The vendor allows you to deploy the appliance to your hardware if desired, so it'd be worth leaving those in just in case anyone does that.

murrant:

PR #11522, opened 2020-05-01
Send Device group membership to Pagerduty

spencerryan:
Send device group membership to pagerduty. This allows routing (event rules) on the Pagerduty side to make decisions based on group membership inside LibreNMS greatly simplifying alert rules and transports in LibreNMS.

I have tested this with a device in 0, 1 and 2 groups and the JSON gets built properly in all cases. 

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Pagerduty.php
+++ b/LibreNMS/Alert/Transport/Pagerduty.php
@@ -24,6 +24,7 @@
 
 namespace LibreNMS\Alert\Transport;
 
+use App\Models\Device;
 use GuzzleHttp\Client;
 use GuzzleHttp\Exception\GuzzleException;
 use Illuminate\Http\Request;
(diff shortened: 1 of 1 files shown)


LibreNMS/Alert/Transport/Pagerduty.php
```diff
@@ -54,12 +55,18 @@ public function deliverAlert($obj, $opts)
      */
     public function contactPagerduty($obj, $config)
     {
+        $groups = Device::find($obj['device_id'])->groups;
```
murrant, line 58: You should use `\DeviceCache::get($obj['device_id'])->groups` instead so it doesn't fetch the device from the database again.

LibreNMS/Alert/Transport/Pagerduty.php
```diff
@@ -54,12 +55,18 @@ public function deliverAlert($obj, $opts)
...
+        foreach ($groups as $group)
```
murrant, line 59: To just get an array of the group names this will work:
```php
\DeviceCache::get($obj['device_id'])->groups->pluck('name')->all();
```

spencerryan: Thanks, just pushed an update, rolled the changes into a single line using DeviceCache as suggested.

murrant:

PR #11533, opened 2020-05-02
Fix DB schema validation for MySQL 8.0.19+

sriccio:
Since MySQL 8.0.19, the integer display width is deprecated and not returned anymore.
This make the db schema validation fails.

This is an attempt at fixing the validation so it doesn't fail anymore with MySQL 8.0.19+

Sources:
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-19.html

Related community post:
https://community.librenms.org/t/report-database-schema-issues-here/945

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Validations/Database.php
+++ b/LibreNMS/Validations/Database.php
@@ -172,8 +172,6 @@ class Database extends BaseValidation
         $current_schema = dump_db_schema();
         $schema_update = array();
 
-        $mysql_version = $validator->getVersions()['mysql_ver'];
-
         foreach ((array)$master_schema as $table => $data) {
             if (empty($current_schema[$table])) {
                 $validator->fail("Database: missing table ($table)");
@@ -188,15 +186,13 @@ class Database extends BaseValidation
                     $column = $cdata['Field'];
 
                     // MySQL 8.0.19+ fix, deprecated integer width
-                    if (!strpos(strtolower($mysql_version), 'mariadb') && version_compare('8.0.18', $mysql_version, '<')) {
-                        if (preg_match('/\w*INT/i', $cdata['Type'])) {
-                            $cdata['Type'] = preg_replace('/\(\d+\)/', '', $cdata['Type']);
-                            $cdata['Type'] = preg_replace('/\s+/', ' ', $cdata['Type']);
-                        }
-                        if (preg_match('/\w*INT/i', $current_columns[$column]['Type'])) {
-                            $current_columns[$column]['Type'] = preg_replace('/\(\d+\)/', '', $current_columns[$column]['Type']);
-                            $current_columns[$column]['Type'] = preg_replace('/\s+/', ' ', $current_columns[$column]['Type']);
-                        }
+                    if (preg_match('/\w*INT/i', $cdata['Type'])) {
+                        $cdata['Type'] = preg_replace('/\(\d+\)/', '', $cdata['Type']);
+                        $cdata['Type'] = preg_replace('/\s+/', ' ', $cdata['Type']);
+                    }
+                    if (preg_match('/\w*INT/i', $current_columns[$column]['Type'])) {
+                        $current_columns[$column]['Type'] = preg_replace('/\(\d+\)/', '', $current_columns[$column]['Type']);
+                        $current_columns[$column]['Type'] = preg_replace('/\s+/', ' ', $current_columns[$column]['Type']);
                     }
 
                     // MySQL 8 fix, remove DEFAULT_GENERATED from timestamp extra columns
--- a/LibreNMS/Validations/Database.php
+++ b/LibreNMS/Validations/Database.php
@@ -214,8 +214,6 @@ class Database extends BaseValidation
                         }
                         $schema_update[] = $this->addColumnSql($table, $cdata, isset($data['Columns'][$index - 1]) ? $data['Columns'][$index - 1]['Field'] : null, $primary);
                     } elseif ($cdata !== $current_columns[$column]) {
-                        print_r($cdata);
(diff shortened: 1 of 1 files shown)


sriccio:
Yes, for this it would just need to remove the mysql version check in my commit.
Just tell me if you want me to remove the check.

Maybe in the future the usage of integer width in the `*.sql` files and `/misc/db_schema.yaml` should be removed too ?

arrmo:
FYI, not saying this is the right way to fix this or not (that's not for me to judge 😆), but I tried the fix vs. what I had been posting (https://community.librenms.org/t/report-database-schema-issues-here/945/494) ... and it works - yay!

Just to let you know.

Thanks!

murrant:
@sriccio we don't use *.sql files since a long time ago.

That is what I am saying, change the generation of /misc/db_schema.yaml to not save the integer width.

sriccio:
@arrmo Cool. Thank you for confirming it worked on your side too.

@murrant Oh okay, it's my first attempt at looking at the code and trying to bring a fix :)

My approach here was a try to bring a fix for latest MySQL versions and avoid it breaks the check for existing DB that runs on < 8.0.19. They would report width and it wouldn't match anymore a modded `db_schema.yaml` without width.

I guess if we remove width from yaml files, stripping the width from current DB results would still be needed to keep compatibility for existing installations.

It's up to you guys :)

murrant:
Just modify the dump_db_schema() function.

Codeclimate is just informational.

murrant:
You need to update the schema dumping code, not the schema checking code.

arrmo:
I must have missed that this got closed - my bad 😞. Is there a different item open to address this? Asking because my list of errors (from validate.php) keeps growing, up to 502 now ... LOL.

Thanks!

murrant:
@arrmo you need to change the code that generates the schema validation file, not the code that checks it.

arrmo:
OK, will take a look. Definitely no expert in this area 😆 

Thanks!

arrmo:
> you need to change the code that generates the schema validation file, not the code that checks it.

Yep, I see what you're referring to. To understand, does this schema dump happen during validate, or some other time? Want to know how to trigger it, to debug.

Thanks!

arrmo:
OK, made a minor change to the dump routine, ran ./scripts/build-schema.php to regenerate the yaml file (right?) ... then down to a MUCH shorter list of issues (from ~ 1100, to now about 20 😄). The few that are left look like below (all timestamp related?). Thoughts?

```
         ALTER TABLE `alert_log` CHANGE `time_logged` `time_logged` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP DEFAULT_GENERATED;
         ALTER TABLE `alerts` CHANGE `timestamp` `timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP DEFAULT_GENERATED;
         ALTER TABLE `applications` CHANGE `timestamp` `timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP DEFAULT_GENERATED on update CURRENT_TIMESTAMP;

```

Thanks!

murrant:
@arrmo Great :)  Those might actually be issues from the recent database migration?  Make a new pull request with your changes and I can check.

arrmo:
Sure, NP! It's a one line change 😄. PR submitted #11725. Please let me know what you think about the remaining few items shown by validate.php.

Thanks!

arrmo:
OK, very odd on the ones remaining. Here is what is in db_schema.yaml (for one being reported, alert_log, time_logged),
```
alert_log:
  Columns:
    - { Field: id, Type: 'int unsigned', 'Null': false, Extra: auto_increment }
    - { Field: rule_id, Type: 'int unsigned', 'Null': false, Extra: '' }
    - { Field: device_id, Type: 'int unsigned', 'Null': false, Extra: '' }
    - { Field: state, Type: int, 'Null': false, Extra: '' }
    - { Field: details, Type: longblob, 'Null': true, Extra: '' }
    - { Field: time_logged, Type: timestamp, 'Null': false, Extra: DEFAULT_GENERATED, Default: CURRENT_TIMESTAMP }
```

And from mysql,
```
mysql> show columns from alert_log;
+-------------+--------------+------+-----+-------------------+-------------------+
| Field       | Type         | Null | Key | Default           | Extra             |
+-------------+--------------+------+-----+-------------------+-------------------+
| id          | int unsigned | NO   | PRI | NULL              | auto_increment    |
| rule_id     | int unsigned | NO   | MUL | NULL              |                   |
| device_id   | int unsigned | NO   | MUL | NULL              |                   |
| state       | int          | NO   |     | NULL              |                   |
| details     | longblob     | YES  |     | NULL              |                   |
| time_logged | timestamp    | NO   | MUL | CURRENT_TIMESTAMP | DEFAULT_GENERATED |
+-------------+--------------+------+-----+-------------------+-------------------+
6 rows in set (0.01 sec)
```


Ummm ... these look to match, no? It is all timestamp ones being reported now.

And BTW, how / when does db_schema.yaml get updated? I took a "guess" to regenerate it ... not sure how it happens (automatically), and if I did the right thing 😞.

murrant:
manually

arrmo:
OK, NP - so it's the "baseline" then, and the validate.php check is to confirm if anything has gotten broken since ... do I have that right? And ./scripts/build-schema.php is the correct manual way to do this? Just trying to understand, thanks! I admit, didn't know I needed to run that - it's very likely in the documentation, and I just missed it 😞.

Any thoughts on the remaining failures? My PR addresses the (number) related items, but still those above. Could it be the Key ("MUL")?

PR #11536, opened 2020-05-02
bugfix: addhost (web) snmp v3 not used first

nimrof:
Changes the code so that the user settings from the web page is added to the front of the snmp v3 config array and not the back.
If it is not at the front, it will later either not be used or at all, be tested last

For comparison both addhost (cli) and lnms (cli) do use array_unshift
[addhost.php](https://github.com/librenms/librenms/blob/master/addhost.php#L126)
[routes/console.php](https://github.com/librenms/librenms/blob/master/routes/console.php#L95)

To reproduce the bug:
Force add new snmp v3 device.
The v3 settings added to the database will not be the same as entered in the webpage, but it will be the first default snmp v3 entry

#### Please note

#### Testers

### DIFF
--- a/includes/html/pages/addhost.inc.php
+++ b/includes/html/pages/addhost.inc.php
@@ -71,7 +71,7 @@ if (!empty($_POST['hostname'])) {
             Config::set('snmp.v3', $v3_config);
 
             $snmpver = 'v3';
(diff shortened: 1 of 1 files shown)


murrant:
Odd bug there, nice catch

murrant:

PR #11539, opened 2020-05-03
js files that needed asset function prepend

Cupidazul:
svg-dynamic-graphs url fix using asset internal function 

and other snipets are missing the asset function prepend.

This need is to add : ' hxxp(s):/xxxx   '  to all script 'src'  texts so that javascripts can be loaded correctly, thus svg-dynamic-graphs appear correctly to everyone that is using ReverseProxy sollution or are behind a Nat.

This has been tested on 'svg-dynamic-graphs', and need more tests on other .php files.
I'm not sure, that the 'asset()' function is availlable on all .php files.

But here is my contribution.

Keep up the good work. 

LibreNMS is great !

#### Please note

#### Testers

### DIFF
--- a/doc/Support/FAQ.md
+++ b/doc/Support/FAQ.md
@@ -125,6 +125,13 @@ using the [included
 snmpd.conf](https://raw.githubusercontent.com/librenms/librenms/master/snmpd.conf.example)
 file.
 
+If you are behind a Proxy or thru a NAT'd ipaddress with an Ngix, you 
+need to add this into your nginx.conf file inside the server statement:
+
+   location ~ ^\/.*\/([a-z_\-]+)\.php$   { try_files $uri $uri/ /$1.php?$query_string;     }
+
+   just before: " location /  { ... }"
+
 ## <a name="faq7"> How do I debug pages not loading correctly?</a>
 
 A debug system is in place which enables you to see the output from
--- a/doc/Support/FAQ.md
+++ b/doc/Support/FAQ.md
@@ -125,7 +125,7 @@ using the [included
 snmpd.conf](https://raw.githubusercontent.com/librenms/librenms/master/snmpd.conf.example)
 file.
 
-If you are behind a Proxy or thru a NAT'd ipaddress with an Ngix, you 
+If you are behind a Proxy or thru a NAT'd ipaddress with an Nginx, you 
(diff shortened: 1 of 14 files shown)


doc/Support/FAQ.md
```diff
@@ -125,6 +125,13 @@ using the [included
 snmpd.conf](https://raw.githubusercontent.com/librenms/librenms/master/snmpd.conf.example)
 file.
 
+If you are behind a Proxy or thru a NAT'd ipaddress with an Ngix, you
```
murrant, line 128: Typo here Ngix -> Nginx

Cupidazul: thx

murrant: Also, can you be more explicit about what scenarios this will help with?

Cupidazul: Sure, this helps when you have a NAT in front of libre NMS, in my case I'll be doing proxy also, but this is valid for any nginx environment. in apache you have MultiViews , in Nginx you do not. so this is a workaround for that.

If you have an Nginx, and you try to open /device/.../graph.php?args you cannot, and you will see a 404 error.

In this Nginx workaround, what it does is to "try" the "graph.php" in the http_root, where it is in fact. whar way Nginx stripes /device/ part and opens the right "graph.php" in the right path. If not, the 404 was a pointer that nginx was trying to find the "graph.php" inside some path /device/.../.../ that doesnt exists, for that the browser shows a 404 - file not found .

this was my troubleshooting last night, lost some time on this, and this was my fix .

keep up the good work, Librenms is awesome .

Cupidazul: Yup, performance tab is not showing...

SyntaxError: expected expression, got '<'  vis.min.js:1
ReferenceError: vis is not definedtab=performance:682:22

Cupidazul: sorry I might have been mistaken, I'm on a docker image running librenms, and still not sure what version I have...
validade shows 1.63, but i still have some files from previous versions...
ex: ${LIBRENMS_PATH}/includes/html/pages/device/performance.inc.php (should not be there)

Cupidazul: sorry I need to redo this pull request. It's no good this way. sorry

PR #11542, opened 2020-05-03
cupidazul: js files that needed asset function prepend

Cupidazul:
svg-dynamic-graphs url fix using asset internal function

and other snipets are missing the asset function prepend.

This need is to add : ' hxxp(s):/xxxx ' to all script 'src' texts so that javascripts can be loaded correctly, thus svg-dynamic-graphs appear correctly to everyone that is using ReverseProxy sollution or are behind a Nat.

This has been tested on 'svg-dynamic-graphs', and need more tests on other .php files.
I'm not sure, that the 'asset()' function is availlable on all .php files.

But here is my contribution.

#### Please note

#### Testers


murrant:
asset() is available everywhere.
Everything looks good, but I'm still dubious about the documentation update.  If you pull that into  a separate PR, I would be happy to merge the rest.

Cupidazul:
> asset() is available everywhere.
> Everything looks good, but I'm still dubious about the documentation update. If you pull that into a separate PR, I would be happy to merge the rest.

Hi Mr. murrant,

Ok. Thanks great news about asset everywhere... I think this merge will be quite helpfull for many people.
Everything is working fine here with all those fixes, I've developed a patch and applied on my server. 
Even realtime graphs work great now.
I don't see anymore major errors on the browser ff+chrome.
Anything else I'll keep you posted.

I'm not sure about that nginx workaround either, I'm doing some more tests on my side. but for now, I really need that nginx workaround...

Any needfull, call me up. Happy to help.
Cheers.

Cupidazul:
> asset() is available everywhere.
> Everything looks good, but I'm still dubious about the documentation update. If you pull that into a separate PR, I would be happy to merge the rest.

This is much better now, Iv'e been updating the code, and fixing issues on my docker environment. All is working now.

a good indicator is "/device/device=6/tab=edit/section=ports", when it doesnt list ports.

Cupidazul:
> asset() is available everywhere.
> Everything looks good, but I'm still dubious about the documentation update. If you pull that into a separate PR, I would be happy to merge the rest.

I might be running a never winning battle here, but I think these issues will occur to everyone with docker + chrome. I'm encountering several ajax options, like, delete's, updates, that do not work in chrome.

Maybe I need to troubleshoot more docker issues with chrome due to cross-site + HSTS issuues, that with newer versions of chrome has got worse and worse, due to more security limitations, which is a good thing. We need to focus on web security :

we are not setting up META: " Strict-Transport-Security: " but we should be careful, as this really breaks things, and even gets worse when we have docker solutions, proxies and NAT's.

here is a good site:
https://blog.heroku.com/using-http-headers-to-secure-your-site

Cupidazul:
> asset() is available everywhere.
> Everything looks good, but I'm still dubious about the documentation update. If you pull that into a separate PR, I would be happy to merge the rest.

I just reallized I'm on https://github.com/librenms/librenms/tree/1.63 , on my docker .

and not on the latest tree ok ? just be warned of this fact, as it may influence this merge.

Cupidazul:
Ok, after much debugging, It happened to be a simple issue with base_url inside config.php.

no paches are needed.

Sorry for this, I'll cancel this push.

Cupidazul:
cancel all this.

PR #11555, opened 2020-05-05
Do not delay alert recovery notifications

### LINKED ISSUE #11552: Alerter Applies "Delay" to recovery transports
This isn't in the documentation anywhere, and appears to be unintended, or at least usually not desired.

If an alert delay is set, the recovery notifications will also be delayed by that amount. In my opinion this isn't how it should operate, or at least be configurable (and not be the default)

I'm happy to submit a pull with the fixes, but I'm not sure how we feel about changing the behavior for existing installs, even if it's unintended.

For reference I think simply changing these two calls in RunAlerts.php:

https://github.com/librenms/librenms/blob/492ab32b6deb8f5a24dc32f4421329eba64e22f7/LibreNMS/Alert/RunAlerts.php#L408

https://github.com/librenms/librenms/blob/492ab32b6deb8f5a24dc32f4421329eba64e22f7/LibreNMS/Alert/RunAlerts.php#L426

To skip the continue loop if ```$alert['state'] == 0``` would make recovery notifications fire immediately.

spencerryan:
Skip the delay logic for recovery.

The alerter logic will delay recovery alerts for the same length as the configured delay before the initial alert triggers. This corrects that logic, sending recovery immediately.

This fixes #11552 

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/RunAlerts.php
+++ b/LibreNMS/Alert/RunAlerts.php
@@ -404,7 +404,7 @@ class RunAlerts
             $tolerence_window = Config::get('alert.tolerance_window');
             if (!empty($rextra['count']) && empty($rextra['interval'])) {
                 // This check below is for compat-reasons
-                if (!empty($rextra['delay'])) {
+                if (!empty($rextra['delay']) && $alert['state'] != 0) {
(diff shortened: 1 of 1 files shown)


murrant:

PR #11561, opened 2020-05-06
[WIP] OS Ray - Added support for new firmware

Martin22:

#### Please note

#### Testers

The OID has changed for the new firmware. Added condition for OID existence, old OID is not in new firmware.


Martin22:
Why is Travis writing a bug with Ray3 that I didn't touch?

OS ray3: Discovered wireless data does not match that found in tests / data / ray3.json
--- Expected
+++ Actual
@@ @@
  Array (
      'wireless_sensors' => Array (
- 0 => Array (...)
- 1 => Array (...)
- 2 => Array (...)
- 3 => Array (...)
- 4 => Array (...)
- 5 => Array (...)
- 6 => Array (...)
      )
  )

Martin22:
Please wait a moment. I noticed that the state, voltage and temperature are still bad. I'll fix it this week.

includes/discovery/sensors/state/ray.inc.php
```diff
@@ -2,8 +2,18 @@
 
 echo 'RAy Racom State';
 
+
 // System Status (Value : na (0) unknown, ok (1) ok, warning (2) warning, alarm (3) alarm)
 $state = snmp_get($device, "systemStatus.0", "-Ovqe", 'RAY-MIB');
+
+$cmd = gen_snmpget_cmd($device, '.1.3.6.1.4.1.33555.1.1.3.1', '-Oqv', 'RAY-MIB', 'ray');
```
murrant, line 9: why are you not using the built in functions?

if .0 is inconsistent, just use `snmp_getnext($device, "systemStatus", "-Ovqe", 'RAY-MIB')` and it will work for both

murrant: to find the oid returned, change the query flags:
```
$output = snmp_getnext($device, "systemStatus", "-Oqn", 'RAY-MIB');
if ($output !== false) {
    list($oid, $state) = explode(' ', $output, 2);
    ...
}
```

Martin22: > why are you not using the built in functions?
> 
> if .0 is inconsistent, just use `snmp_getnext($device, "systemStatus", "-Ovqe", 'RAY-MIB')` and it will work for both

If this is not the case, the following OID will be used.

Martin22:
> Please use the built in snmp functions.

Yes, I just need more time. I have other projects now, I'll get to it on Friday.

LibreNMS/OS/Ray.php
```diff
@@ -72,11 +72,11 @@ public function discoverProcessors()
...
-        $oid_rx = '.1.3.6.1.4.1.33555.1.2.1.4'; // RAY-MIB::rxFreq.0
-        $cmd = gen_snmpget_cmd($this->getDevice(), $oid_tx, '-Oqv', 'RAY-MIB', 'ray');
-        $data = trim(external_exec($cmd), "\" \n\r");
-        if (!preg_match('/(No Such Instance)/i', $data)) {
+        $oid_tx = '.1.3.6.1.4.1.33555.1.2.1.4'; // RAY-MIB::txFreq
+        $oid_rx = '.1.3.6.1.4.1.33555.1.2.1.4'; // RAY-MIB::rxFreq
+
+        $cmd = snmp_get($this->getDevice(), $oid_tx, '-Oqv', 'RAY-MIB', 'ray');
+        if ($cmd !== false) {
             return array(
```
murrant, line 80: Whey even do the snmpget at all?  The Wireless code will actually do an snmpget and verify it has data before adding the device.

Martin22: The problem is this:

Old firmware:
An OID without .0 at the end exits and returns a value.
An OID with .0 at the end exists and returns a value.

New firmware:
An OID without .0 does not exist, it returns an error.
An OID with .0 exists and returns a value.

If you suggest another solution, I will be happy to implement it.

Jellyfrog:
@Martin22 Missig matching `ray_newfw.json` etc.

LibreNMS/OS/Ray.php
murrant: So, you should always set it to .0 right?

Martin22:
Since another commit broke the OS RAY detection, I no longer have the strength to do anything with this and look for which commit caused it.

Since another request must not break a functional thing and it happened, this rule makes no sense to me.

The problem is that RAY does not follow the RFC in the old firmware and does not use .0 at the end. If someone fixes it, I will continue the preparation, but now it is not possible.

murrant:
Seems like if you just walked the base oids and use the numeric oid returned by the walk, it would be very simple to support both with the exact same code.  No worries man, if you feel up to it go ahead and reopen/submit.

PR #11563, opened 2020-05-06
Add python validation fix suggestions

murrant:

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Validations/Python.php
+++ b/LibreNMS/Validations/Python.php
@@ -28,6 +28,7 @@ namespace LibreNMS\Validations;
 use LibreNMS\Config;
 use LibreNMS\Util\Version;
 use LibreNMS\Validator;
+use Symfony\Component\Process\Process;
 
 class Python extends BaseValidation
 {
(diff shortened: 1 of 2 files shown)


LibreNMS/Validations/Python.php
```diff
@@ -65,7 +65,7 @@ private function checkExtensions(Validator $validator)
         exec(Config::get('install_dir') . '/' . $pythonExtensions . ' -v', $output, $returnval);
 
         if ($returnval !== 0) {
-            $validator->fail("Python3 module issue found: '" . ($output[0] ?? '') . "'");
+            $validator->fail("Python3 module issue found: '" . ($output[0] ?? '') . "'", 'pip3 install -r requirements.txt');
```
SourceDoctor, line 68: mmh this install modules global, don't you want to locate it in user home?

Jellyfrog: Agree, maybe it's more sane to suggest `--user`, in general I don't think we shouldn't make so much suggestions, it's up to the sysadmin.

SourceDoctor: outside of virtualenv i personaly prefer apt packed python modules ... they always install without problems :D

murrant: if the user is running it they will see feedback and make a decision (hopefully).

Also, when installing with --user the web validation always fails because it is a different user.

murrant: @SourceDoctor the problem with apt is the packages are named different things everywhere it seems.

I could add sudo to the command...

Jellyfrog: Leave it as is imo

murrant:

PR #11567, opened 2020-05-07
Ability to map librenms devicegroup into oxidized group

biwhite:
…ou map LibreNMS groups onto oxidized groups as such:

$config['oxidized']['maps']['group']['devicegroup'][] = array('match' => 'Libre Device Group 1', 'group' => 'group1');
$config['oxidized']['maps']['group']['devicegroup'][] = array('match' => 'Libre Device Group 2', 'group' => 'group2');

#### Please note

- [ X]
- [ X] 

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1342,7 +1342,7 @@ function list_oxidized(\Illuminate\Http\Request $request)
         unset($device['sysname']);
         unset($device['sysDescr']);
         unset($device['hardware']);
-	unset($device['devicegroup']);
+        unset($device['devicegroup']);
         $devices[] = $device;
     }
 
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1300,7 +1300,7 @@ function list_oxidized(\Illuminate\Http\Request $request)
         $params = [$hostname];
     }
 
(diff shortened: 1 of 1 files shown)


biwhite:
Previously, could only map groups based on hostname or os regex/matches, now can do it on device group membership as well.

murrant:
Hmm, there is some overlap here.

SourceDoctor:
a device can be member of multiple devicegroups.
How is it solved, so stuff of a system will be saved only once in oxidized?

biwhite:
Oxidized appears to support having a device in multiple groups, and then polls the device multiple times and stores a config per group.
It does however appear to only return one of these when querying the node by name via the 
`file_get_contents(Config::get('oxidized.url') . '/node/show/' . $oxidized_hostname . '?format=json'` call within `includes/html/pages/device/showconfig.inc.php`

jozefrebjak:
@murrant Will be this ever merged ? I am waiting for it.

biwhite:
Do we need to find a solution for what if a device is in multiple groups?
It can feed that into oxidized just fine, which causes Oxidized to poll the device once per group it's in, however the oxidized API link to a particular device shows only 1 of those groups. https://oxhost/node/show/nodename
Ideally we ought to be able to control which oxidized group it's polled from.  It's not great to poll the device multiple times.  Maybe a group by hostname on the join or such but it wouldn't necessarily be deterministic as to which group it would appear in?

SourceDoctor:
not sure what the the best way is. But each device should be polled once

murrant:
@jozefrebjak it has some design issues that need to be sorted.

What do you think about an option on the device group called "Send to Oxidized" and it would also need a priority value.

biwhite:
It looks like oxidized does support group in the fetch URL, 
oxidized.url/node/fetch/group/nodename
LibreNMS would need to use the map to work out what the oxidized group-name would be to produce the download config URL.

I think if we can get a 'group priority' or such attached to a device, so it's returned as part of that group, and then order by group-priority & group by device into the query, it'll return the highest priority group as the one it should be polled within

murrant:
@biwhite yes, the issue is LibreNMS has two ways of creating groups.  One for Oxidized, and one generic way.

PR #11571, opened 2020-05-08
Enabling general search for ports, devices, and more

hachpai:
Three years ago, I [suggested](https://community.librenms.org/t/port-search-query-based-on-ifalias-using-global-search/1356) a feature in the API but I was lacking of time and experience to write it. Now it's done. I exposed the generic search (used by ajax search bar) to the API.
For now, i just did it with ports (because it's really what I need). But it's trivial to extend it.
It's my quick contribution and almost my first time coding in PHP so you should expect a WIP, but working. Here's a curl example: `curl -H "X-Auth-Token: XXXX" 'http://127.0.0.1:8000/api/v0/search_ports?search=[what you want]'`

#### Please note

#### Testers


murrant:

https://community.librenms.org/t/port-search-query-based-on-ifalias-using-global-search/1356/9

murrant:
Nice, but you are re-arranging and pulling in a ton of complex code for something that only requires a single query.

```php
$value = "%$search%";
\App\Models\Port::hasAccess(Auth::user())
    ->orderBy('ifName')
    ->where('ifAlias', 'like', $value)
    ->orWhere('ifDescr', 'like', $value)
    ->orWhere('ifName', 'like', $value)
    ->get();
```

hachpai:
You're right. I tried to stick to the logic without diving deep :-)

hachpai:
By user-pending, you expect me to change for the implementation suggested by @murrant ?
Or we go for a more generic search and expose the same logic than ajax search?
It's more a kind of a multi purpose search engine and we cover all the use case in one pull.

murrant:
Yes, I suggest keeping the scope of this PR just for ports.

hachpai:
Just tested and pushed the new version. Is it ok for you?

includes/html/api_functions.inc.php
```diff
@@ -72,6 +72,19 @@ function api_not_found()
     return api_error(404, "This API route doesn't exist.");
 }
 
+function search_ports(\Illuminate\Http\Request $request)
+{
+    $search = $request->get('search');
```
SourceDoctor, line 77: what happens if search is empty or an empty string?

murrant: it searchs for %%

SourceDoctor: doesn't that mean "no filter", so all ports are searched?
not sure if it's such good on thousands of ports which will be returned as result

murrant:

PR #11582, opened 2020-05-09
Fix midnight poller data loss

TheMysteriousX:
We have a discovery job that takes around 8 minutes to complete. If it coincides with the nightly reload, the entire poller stops running for those 8 minutes.

This changes things up to eliminate this loss of data, while adding the small chance that a second job may be executed for the same device by executing the new poller process before all processes return.

It also fixes an issue where log messages would be lost. By default, stdout's buffer is not flushed when exit is called.

graph php

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/service.py
+++ b/LibreNMS/service.py
@@ -3,7 +3,6 @@ import LibreNMS
 import json
 import logging
 import os
-import psutil
 import pymysql
 import subprocess
 import threading
@@ -18,6 +17,11 @@ from time import sleep
 from socket import gethostname
 from signal import signal, SIGTERM, SIGQUIT, SIGINT, SIGHUP, SIGCHLD, SIG_DFL
 from uuid import uuid1
+
+try:
+    import psutil
+except ImportError:
+    print("psutil is not available - this is not a supported configuration and may leak process descriptors")
 
 
 class ServiceConfig:
@@ -264,6 +268,7 @@ class Service:
     poller_manager = None
     discovery_manager = None
     last_poll = {}
+    reap_flag = False
     terminate_flag = False
     reload_flag = False
     db_failures = 0
@@ -298,20 +303,14 @@ class Service:
         signal(SIGHUP, self.reload)  # capture sighup and restart gracefully
         signal(SIGCHLD, self.reap)  # capture sigchld and reap the process
 
-
-    def reap(self, signum, stackframe):
+    def reap_psutil(self):
         """
         A process from a previous invocation is trying to report its status
         """
-        if (signal(SIGCHLD, SIG_DFL) == SIG_DFL):
-            # signal is already being handled, bail out as this handler is not reentrant
-            # the kernel will re-raise the signal later
-            return
-
-        # Speed things up by only looking at direct children - grandchildren will be reaped by their parent
+        # Speed things up by only looking at direct zombie children
         for p in psutil.Process().children(recursive=False):
             try:
-                cmd = p.cmdline()
+                cmd = p.cmdline() # cmdline is uncached, so needs to go here to avoid NoSuchProcess
                 status = p.status()
                 pid = p.pid
 
@@ -322,8 +321,19 @@ class Service:
                 # process was already reaped
                 continue
 
-        # Re-arm the signal handler
-        signal(SIGCHLD, self.reap)
+    def reap_fallback(self):
+        # without psutil we must evaluate every single process
+        while True:
+            try:
+                # Request the state of the first waiting child - don't block if there aren't any
+                r = os.waitpid(-1, os.WNOHANG)
+            except OSError:
+                # No more processes
+                break
+
+            if r[0] > 0:
+                # If PID is <= 0, it means that the return code was likely collected already
+                warning('Reaped long running job, install psutil for detail')
 
     def start(self):
         debug("Performing startup checks...")
@@ -371,6 +381,16 @@ class Service:
                 if self.reload_flag:
                     info("Picked up reload flag, calling the reload process")
                     self.restart()
+
+                if self.reap_flag:
+                    try:
+                        self.reap_psutil()
+                    except NameError:
+                        self.reap_fallback()
+
+                    # Re-arm the signal handler
+                    signal(SIGCHLD, self.reap)
+                    self.reap_flag = False
 
                 master_lock = self._acquire_master()
                 if master_lock:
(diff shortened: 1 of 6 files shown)


TheMysteriousX:
I'm undecided if this approach is the best one, or if we should just SIGTERM any outstanding processes. If the service is restarted via systemd, it sends SIGTERM to all child processes.

murrant:

https://community.librenms.org/t/daily-gaps-in-graphs-around-12-1am/9523/28

murrant:
@TheMysteriousX I have been thinking about this issue.  This is the fix I don't want to do and will abort all current polling processes.

Ideally we need to allow existing poller (and other) processes to continue to execute while we restart the dispatch service.

If you can think of a reason we don't need to do that, let me know.

(let me know if I'm wrong and this does not abort poller processes)

TheMysteriousX:
It should do that.

Toy example: https://repl.it/repls/VeneratedForsakenLanserver

We restart the process every second using execl, and immediately after restart, we launch a long running process, which just delays 3 seconds then touches a file.

You should see that despite the re-execution, the subprocess survives - it's insulated by ```preexec_fn=os.setsid```: https://linux.die.net/man/2/setsid

I have found that I need to handle sigchld though - even though the processes complete, they linger as their parent process is no longer around to reap them. I also need to keep an eye on redis for a bit longer to make sure that we're not leaking there.

I have an alternate approach in mind, where we kick off the termination, fork() fork(), execl(), which may end up being cleaner than handling sigchld.

LibreNMS/service.py
```diff
@@ -413,19 +415,22 @@ def restart(self):
...
+        sys.stdout.flush()
+
+    def terminate(self, signalnum=None, flag=None):
         """
         Handle a set the terminate flag to begin a clean shutdown
         :param _unused:
         :param _:
         """
-        info("Received SIGTERM on thead %s, handling", threading.current_thread().name)
+        info("Received signal on thead %s, handling", threading.current_thread().name)
```
murrant, line 433: thread is mis-spelled here :D

murrant:
Great, this all looks good to me as soon as you are happy with it.

zombah:
@TheMysteriousX  Tested with 1.63 and it works as expected, no midnight gaps anymore, all graphs are solid. Thank you!

LibreNMS/service.py
```diff
@@ -329,16 +391,16 @@ def fetch_immediate_device_list(self):
 
             result = self._db.query('''SELECT `device_id`,
                   `poller_group`,
-                  COALESCE(`last_polled` <= DATE_ADD(DATE_ADD(NOW(), INTERVAL -%s SECOND), INTERVAL `last_polled_timetaken` SECOND), 1) AS `poll`,
-                  IF(snmp_disable=1 OR status=0, 0, COALESCE(`last_discovered` <= DATE_ADD(DATE_ADD(NOW(), INTERVAL -%s SECOND), INTERVAL `last_discovered_timetaken` SECOND), 1)) AS `discover`
+                  IF (%s < `last_polled_timetaken` * 1.25, 0, COALESCE(`last_polled` <= DATE_ADD(DATE_ADD(NOW(), INTERVAL -%s SECOND), INTERVAL `last_polled_timetaken` SECOND), 1)) AS `poll`,
```
murrant, line 394: I'm not sure this is exactly what you want for polling.  say last_polled_timetaken is the worst case (300s) and it just missed starting before the restart.  It will not poll until 375s after the service as started, skipping an interval.

I assume you are attempting to prevent double polling here.  However, the service already locks devices while they are being polled so they can't be double polled.

murrant: Discovery is a bit different, the locks are much shorter than the discovery interval, so there is a chance for an early rediscover.

TheMysteriousX: True, the locks should be sufficient. I'll leave it in for discovery.

TheMysteriousX:
I've removed the fallback code as due to the rrdtool/snmpget process churn, the main process burns 100% cpu when PSutil isn't available to filter the process table.

Fortunately, psutil has merged [wheel support](https://github.com/giampaolo/psutil/pull/1758) within the last few hours, so once that makes it into a release, it will be possible to add psutil as a dependency without needing a c compiler available.

LibreNMS/service.py
```diff
@@ -330,15 +371,15 @@ def fetch_immediate_device_list(self):
...
+                  IF(snmp_disable=1 OR status=0, 0, IF (%s < `last_discovered_timetaken` * 1.25, 0, COALESCE(`last_discovered` <= DATE_ADD(DATE_ADD(NOW(), INTERVAL -%s SECOND), INTERVAL `last_discovered_timetaken` SECOND), 1))) AS `discover`
                 FROM `devices`
                 WHERE `disabled` = 0 AND (
                     `last_polled` IS NULL OR
                     `last_discovered` IS NULL OR
                     `last_polled` <= DATE_ADD(DATE_ADD(NOW(), INTERVAL -%s SECOND), INTERVAL `last_polled_timetaken` SECOND) OR
                     `last_discovered` <= DATE_ADD(DATE_ADD(NOW(), INTERVAL -%s SECOND), INTERVAL `last_discovered_timetaken` SECOND)
                 )
-                ORDER BY `last_polled_timetaken` DESC''', (poller_find_time, discovery_find_time, poller_find_time, discovery_find_time))
+                ORDER BY `last_polled_timetaken` DESC''', (self.service_age(), poller_find_time, self.service_age(), discovery_find_time, poller_find_time, discovery_find_time))
```
murrant, line 382: too many variables here?

TheMysteriousX: Yes, surprised the linter didn't catch that one.

laf:
@TheMysteriousX Looks like this needs a rebase, can you deal with that please?

TheMysteriousX:
Will do. Still waiting on psutil to spin a release so this doesn't straight up kill anyone running without a C compiler installed.

zombah:
Tested again with psutil and gcc, works fine. Thanks

murrant:
?LGTM now @TheMysteriousX  is this ready for merge?

TheMysteriousX:
I think so, I haven't been able to test the last set of changes I made yet as work got busy.

PSutil should now be optional - i.e. if PSUtil is available, it'll do things nicely and reload right. If it's not available, the current behaviour will continue.

murrant (approved):
Tested without psutil

murrant:

velde666:
Hi there,

I am currently running LibreNMS on Version 21.1.0 and still have this issue that every night between app. 01:00 AM and app. 02:10 AM no polling happens as discovery is doing its work:

grafik

Polling starts again when the last long-lasting discovery ended:

grafik

Is there anything I can check if the bugfix is in place?

Best regards
Alex

murrant:
@velde666 make sure backups are not pausing the container.

velde666:
Hi @murrant there is no backup active at the moment to exclude that as cause of this issue.

velde666:
any ideas besides backup?

PR #11609, opened 2020-05-14
Fixes for Eaton UPSes. Agent hw/sw. Gigabit M2. MIB update.

dagbdagb:
Bit of a backstory here....

The enterprise ID of what is currently Eaton Corporation has been associated with a number of company- and UPS product names throughout the years.

In short: enterprise ID 534 is currently Eaton Corporation.
            Copyright (C) Exide Electronics 1992-98
            Copyright (C) Powerware Corporation 1999-2004
            Copyright (C) Eaton Corporation (2005-)."

In additon, Eaton bought out the Smalls Systems Business of what used to be Merlin Gerin around 2007. Merlin Gerin itself was acquired by Schneider Electric in 1992, but the brand was kept until 2009. From this, Eaton also has gear with the enterprise ID 705, which to this day is associated with Merlin Gerin.

Eaton (Powerware/Merlin Gerin) UPSes have had a number of SNMP-enabled options, but the bulk of the products in modern times have been supported by the optional 'Network MS' card, which handles all of the networking tasks. This card is used in a truly impressive range of products, spanning from the Exide era to 'modern' times. I do not know when the product was launched, but the current manual is dated 2013.

Ref. page 10 of the user manual:
https://www.eaton.com/content/dam/eaton/products/backup-power-ups-surge-it-power-distribution/power-management-software-connectivity/eaton-network-card-ms/eaton-network-card-ms-user-guide-manual.pdf

At the end of 2019, 'Network MS'  was replaced with a completely redesigned card, the 'Gigabit M2'. This is a much more capable product in terms of CPU and crypto. But this card officially supports only a subset of the products the older 'Network MS' can handle: 5SC rack or RT, 5P, 5PX, 9SX, 9PX, 93PM, 9PHD, 91PS, 93PS, 93PS Marine

Both of these cards will inherit the actual UPS model name as its sysDescr.
Both of these cards support the IETF UPS MIB '{ mib-2 33 }'
Both of these cards support the Powerware/Eaton XUPS MIB '{ enterprises(1) eaton(534) xupsMIB(1) }'
Only 'Network MS' supports the old MGE MIB '{ enterprises(1) merlinGerin(705) upsmg(1) }'

'Network MS' comes with the .1.3.6.1.4.1.705.1 sysOID
'Gigabit M2' comes with the .1.3.6.1.4.1.534.1 sysOID

Changes:
1. drop 'MGE' from 'text:' in ./includes/definitions/eaton-mgeups.yaml.
While the MIB is the old MGE UPS MIB, the actual UPS product is not necessarily an original MGE product.

2. add 'rfc1628_compat: true' to ./includes/definitions/eatonups.yaml.
I honestly do not know if there are Eaton products around supporting the XUPS mib, but *not* the IETF UPS MIB. But I do know that the opposite is true. Should I detect the M2 specifically and split it out into a separate OS?

3. Modify ./includes/definitions/eatonups.yaml to have the same minigraphs as ./includes/definitions/eaton-mgeups.yaml

4. Modify ./includes/polling/os/eaton-mgeups.inc.php to include the agent hw/sw ('features:')

5. Create ./includes/polling/os/eatonups.inc.php, and make it 'match'  ./includes/polling/os/eaton-mgeups.inc.php.

6. Update relevant MIB-files. Updated XUPS-MIB is required for reporting the agent hw/sw version on the M2, and the updated MIB-file relies on an updated EATON-OIDS. Due to this, 
EATON-OIDS now exists in three copies in the mibs-dir. (Rather than two.) Please advise.

Still issues:
a. eaton-mge reports 'Percentage Load' correctly (as a percentage). It also reports the same value * 1000 as 'Output (VA)' under Power. This is wrong and is equally wrong for three or four different Eaton UPS model families I can test with. Nor can I find anything supporting this interpretation of the value reported.

b. eatonups (but not eaton-mge) has some files under ./includes/discovery/sensors/. I do not fully understand the presence and function of these. Or why these files are required for eatonups and not eaton-mge.

#### Please note

#### Testers


murrant:
God bless for working on this.  I have explored the maze that you described here and appreciate your story :)

jozefrebjak:
I have EATON UPS model  `PW9130 2000VA-R` when I applied your PR and rediscovered UPS, only `OS Features: Network Management Card V6.00 JC`changed for me. Maybe we can clear some OIDs because I have some of them twice. For example 

```
charge | eaton-mgeups | Remaining battery capacity
charge | rfc1628 | Battery charge remaining
```

```
current | eaton-mgeups | Battery
current | rfc1628 | Battery
```

and there is more of them.

dagbdagb:
@jozefrebjak Thank you for testing!
I am certain that eaton-mgeups can be cleaned up quite a bit, as it pulls and presents some info from all three supported MIBs. I have hesitated to go there, as I am not allowed to change the RRD. If I understand @murrant and the 'real' developers correctly. 

In short: I do not fully understand how the code works, and I have the understanding that I am not allowed to change the data logged by the current code. 

Side note : your 'JC' version is 5+ years old. Have a look at http://powerquality.eaton.com/support/software-drivers/downloads/connectivity-firmware.asp
for updated fw. *Do* read releasenotes before upgrading, there are hardware limitations in play here.

dagbdagb:
I have been playing with the json file for eatonups and eaton-mgeups. I am obviously missing an essential understanding of how the unit tests work. Please advise.

dagbdagb:
This is still WIP, just being sidetracked with other stuff. Found another piece of kit supporting the XUPS MIB: '10/100Mb ConnectUPS Web/SNMP'. The list of firmware changes for this hardware also references RFC 1628. Hence, I conclude that also this hardware supports the IETF UPS MIB.

ajsiersema:
Like you already mentioned, if sysOID .1.3.6.1.4.1.534.1 only covers 'Gigabit M2' modules, is it possible to simply use most of the rfc1628 code and strip/cleanup the eatonups code ?

dagbdagb:
@ajsiersema 
I have no issue at all with you picking up the ball from here, if you are so inclined. (I have a couple of projects unrelated to Librenms I need to see completed before I can invest time in this. Too many hobbies...) 

I too would like to rely as much as possible on the generic code. Ideally, the 'Network MS' hardware could do the same, I think. There are a few bits and pieces of information I think must be pulled from the xups  or mgeups MIBs. (fw/hw version information, I believe.) If you are laser-focused on the M2 hardware, then disregard the Network MS hw. 

If you find more sensors and stuff to include, great.

As stated above, I beleive @murrant  had issues with the RRD-format being changed for a device already supported by LibreNMS. Not entirely sure about the implications of this.

dagbdagb:
I could need some input from @murrant or ...anyone about this PR.

As stated above, differentiating between Eaton UPSes and Eaton MG UPSes does not make sense. The old Network Management Card ("Network MS" AKA "Network Management Card V6.00") is used in either type, the more recent card (Network M2) in a subset of the UPSmodels which the older NMC supports + likely newer models which will not be supported by the older NMC. The NMC handles all networking functions.

sysDescr is identical for both cards, and only reflects the UPS hardware, not the NMC itself.
sysOID is not identical.

Also, both cards supports the IETF UPS MIB and XUPS MIB. The older card also supports the MGUPS MIB, and there are some minor bits of information which is better pulled from that mib for the older card.

I have (to the best of my ability) added some bits and pieces required to support the M2 better than was there from before, as well as some minor updates to the eaton-mgups bits. But there is a bit of assymetry in the support for this hardware:


There are some pre-existing files ./includes/discovery/sensors/*/eatonups.inc.php which I have not touched. These do not have an eaton-mgeups equivalent. Are these files needed?

Likewise, there is a pre-existing ./includes/definitions/discovery/eaton-mgeups.yaml which does not have an eatonups equivalent. Is this file needed?


In my librenms db, the old card has 50+ sensors, many of which clearly are redundant, while the new has 9. There is no overlap in sensor_type between the two cards. After this PR, I have a bunch of rfc1628 sensors for the new card, which allows for common alert rules. Yay. But does this run afoul with the 'no changes to pre-existing RRDs'?


I need to add a bit of code to present the M2 hw/sw version. Apart from that, I am fairly happy with what I have added. Less so with trimming and unifying some of the code.

dagbdagb:
Prior to my commit:
```
MariaDB [librenms]> select sensor_type,sensor_class,poller_type,sensor_descr,sensor_alert,sensor_current from sensors where device_id = 1110;
+-------------+--------------+-------------+----------------+--------------+----------------+
| sensor_type | sensor_class | poller_type | sensor_descr   | sensor_alert | sensor_current |
+-------------+--------------+-------------+----------------+--------------+----------------+
| xups        | current      | snmp        | Output         |            1 |              0 |
| eatonups    | charge       | snmp        | Battery Charge |            1 |            100 |
| xups        | frequency    | snmp        | Input          |            1 |             50 |
| xups        | frequency    | snmp        | Output         |            1 |             50 |
| eatonups    | power        | snmp        | Power Usage    |            1 |              0 |
| xups        | voltage      | snmp        | Battery        |            1 |             55 |
| xups        | voltage      | snmp        | Input          |            1 |            238 |
| xups        | voltage      | snmp        | Output         |            1 |            229 |
| xups        | voltage      | snmp        | Bypass         |            1 |            238 |
+-------------+--------------+-------------+----------------+--------------+----------------+
9 rows in set (0.00 sec)
```

After my commit:
```
MariaDB [librenms]> select sensor_type,sensor_class,poller_type,sensor_descr,sensor_alert,sensor_current from sensors where device_id = 1110;
+-----------------------+--------------+-------------+----------------------------------+--------------+----------------+
| sensor_type           | sensor_class | poller_type | sensor_descr                     | sensor_alert | sensor_current |
+-----------------------+--------------+-------------+----------------------------------+--------------+----------------+
| eatonups              | charge       | snmp        | Battery Charge                   |            1 |            100 |
| eatonups              | power        | snmp        | Power Usage                      |            1 |              0 |
| rfc1628               | current      | snmp        | Output                           |            1 |              0 |
| rfc1628               | current      | snmp        | Bypass                           |            1 |              0 |
| rfc1628               | charge       | snmp        | Battery charge remaining         |            1 |            100 |
| rfc1628               | frequency    | snmp        | Input                            |            1 |           50.1 |
| rfc1628               | frequency    | snmp        | Output                           |            1 |           50.1 |
| rfc1628               | frequency    | snmp        | Bypass                           |            1 |           50.1 |
| rfc1628               | load         | snmp        | Percentage load                  |            1 |              0 |
| rfc1628               | power        | snmp        | Output                           |            1 |              0 |
| rfc1628               | runtime      | snmp        | Time on battery                  |            1 |              0 |
| rfc1628               | runtime      | snmp        | Estimated battery time remaining |            1 |            999 |
| upsBatteryStatusState | state        | snmp        | Battery Status                   |            1 |              2 |
| upsOutputSourceState  | state        | snmp        | Output Source                    |            1 |              3 |
| rfc1628               | voltage      | snmp        | Battery                          |            1 |           55.5 |
| rfc1628               | voltage      | snmp        | Output                           |            1 |            229 |
| rfc1628               | voltage      | snmp        | Input                            |            1 |            239 |
| rfc1628               | voltage      | snmp        | Bypass                           |            1 |            239 |
| xups                  | current      | snmp        | Output                           |            1 |              0 |
| xups                  | frequency    | snmp        | Input                            |            1 |           50.1 |
| xups                  | frequency    | snmp        | Output                           |            1 |           50.1 |
| xups                  | frequency    | snmp        | Bypass                           |            1 |           50.1 |
| xups                  | voltage      | snmp        | Battery                          |            1 |             55 |
| xups                  | voltage      | snmp        | Input                            |            1 |            239 |
| xups                  | voltage      | snmp        | Output                           |            1 |            229 |
| xups                  | voltage      | snmp        | Bypass                           |            1 |            239 |
+-----------------------+--------------+-------------+----------------------------------+--------------+----------------+
26 rows in set (0.00 sec)

```

includes/polling/os/eatonups.inc.php
```diff
@@ -0,0 +1,10 @@
+<?php
```
Jellyfrog, line 1: Please convert this to the new os format, https://docs.librenms.org/Developing/os/Initial-Detection/#os-discovery

includes/polling/os/eaton-mgeups.inc.php
```diff
@@ -7,5 +7,5 @@
```
Jellyfrog, line 7: Please convert this to the new os format, https://docs.librenms.org/Developing/os/Initial-Detection/#os-discovery

Jellyfrog (changes requested):

murrant:
There are some pre-existing files ./includes/discovery/sensors/*/eatonups.inc.php which I have not touched. These do not have an eaton-mgeups equivalent. Are these files needed?

These are the php method of creating sensors, only needed if you cannot create the sensor purely in yaml.  (probably created before the yaml option existed)

Likewise, there is a pre-existing ./includes/definitions/discovery/eaton-mgeups.yaml which does not have an eatonups equivalent. Is this file needed?

It is ok to have sensors to in the discovery that will fail one one device type, but be present on another.  What we do want to avoid is having duplicate sensors appear to the user.

dagbdagb:
> `There are some pre-existing files ./includes/discovery/sensors/*/eatonups.inc.php which I have not touched. These do not have an eaton-mgeups equivalent. Are these files needed?`
> 
> These are the php method of creating sensors, only needed if you cannot create the sensor purely in yaml. (probably created before the yaml option existed)

Soooo.... seeing how these sensors were introduced when the hardware (M2) was first supported in LNMS, removing them could potentially be a change which introduces surprises to some users. And it changes the RRDs, I presume.


> `Likewise, there is a pre-existing ./includes/definitions/discovery/eaton-mgeups.yaml which does not have an eatonups equivalent. Is this file needed?`
> 
> It is ok to have sensors to in the discovery that will fail one one device type, but be present on another. What we do want to avoid is having duplicate sensors appear to the user.

OK, I definitely introduce duplicate sensors for the M2 cards by marking it as RFC1628-compatible. And the MS card has a load of duplicate sensors prior to my change.

I propose that I make the fixes @Jellyfrog proposed above, and after this is merged, we (you...) give users a Notification that alerts relying on non-RFC1628 sensors for Eaton UPSes should be reconfigured to do so within a reasonable timeframe.

Does that sound OK?

includes/polling/os/eatonups.inc.php
dagbdagb: @Jellyfrog I do not understand.

Discovery *is* performed with a yaml file. Polling is defined with php as everything else. Please elaborate.

Jellyfrog: these values now belong to the discovery phase, please check https://docs.librenms.org/Developing/os/Initial-Detection/#os-discovery for more info.  
Either the php way; https://github.com/librenms/librenms/blob/master/LibreNMS/OS/Qnap.php  
or the yaml way;  https://github.com/librenms/librenms/blob/master/includes/definitions/discovery/teldat.yaml

dagbdagb: Thank you. Will try again.

includes/polling/os/eaton-mgeups.inc.php
dagbdagb: done

ajsiersema:
Sorry I didn't chime in earlier. I've been away for 2 weeks. Prior to that I've been working on some code as well. Glad to see dagbdagb also started working on this issue again. I've also contacted the vendor to check if .1.3.6.1.4.1.534.1 are _only_ M2's. Unfortunately, the reply I received states this OID also covers "network-ms" and "x-slot" type cards, i.e. non-RFC1628 compatible devices :-(

dagbdagb:
> Sorry I didn't chime in earlier. I've been away for 2 weeks. Prior to that I've been working on some code as well. Glad to see dagbdagb also started working on this issue again. I've also contacted the vendor to check if .1.3.6.1.4.1.534.1 are _only_ M2's. Unfortunately, the reply I received states this OID also covers "network-ms" and "x-slot" type cards, i.e. non-RFC1628 compatible devices :-(

@ajsiersema 
Eaton may have misunderstood your question. Network MS clearly supports the XUPS OID (.1.3.6.1.4.1.534), but the sysOID of Network MS is definitely .1.3.6.1.4.1.705, which is the old MGE line of hardware. This also applies to the modbus variant of the same card, according to Eaton docs.

EATON-OIDS lists a number of devices matching .1.3.6.1.4.1.534.*something*, but the YAML here only matches on .1.3.6.1.4.1.534.1 and .2.  Not sure if this truly elliminates any and all options supporting XUPS-MIB but *not* the RFC1628 MIB.
Also note that the string 'Network MS' is not mentioned anywhere in EATON-OIDS.

And finally, all the Eaton network interface options are old as sin itself. Except the M2. The M2 is a shiny Tesla in an ocean of crappy Ladas and Fiats. So I want the M2 supported properly and have everything else sorted out on an as-needed basis. :-)

Not even sure if polling for RFC1628 sensors will break anything....

ajsiersema:
> Not even sure if polling for RFC1628 sensors will break anything....

My dilemma exactly. It would be nice if, instead of introducing more device specific code, we could rely on more vendor agnostic already present code.
I've however only been able to test this on the devices and s/w version mentioned in #11994
As we don't want to break things and the principle of least surprise should be followed if at all possible...

dagbdagb:
@ajsiersema I have pushed another set of fixes to https://github.com/dagbdagb/librenms/commit/4de46d791ab3d91b9ff0faa548d0736ac8efae83
This axes redundant sensors, relying on rfc1628 sensors for everything which can be read and presented that way.

But I believe we should limit the first set of changes to what is in *this* PR for now,

I have attempted to copy the logic from includes/definitions/discovery/eaton-mgeups.yaml for creating some non-rfc1628 sensors for eatonups. See includes/definitions/discovery/eatonups.yaml. Discovery now pulls the value XUPS-MIB::xupsBatteryAbmStatus.0 from an M2, but it never appears in the sensors table. Not sure what I am missing. :-(

dagbdagb:
Redoing this in another PR

PR #11638, opened 2020-05-19
Fix availability map downtime

louis-oui:
Down devices on availability map currently displays uptime instead of downtime on mouseover. This is a fix.

image

#### Please note

#### Testers

### DIFF
--- a/app/Models/Device.php
+++ b/app/Models/Device.php
@@ -270,6 +270,11 @@ class Device extends BaseModel
         return Time::formatInterval($this->uptime, $short);
     }
 
+    public function formatDowntime($short = false)
+    {
+        return Time::formatInterval(time() - $this->last_polled, $short);
+    }
+
     /**
      * @return string
      */
--- a/app/Models/Device.php
+++ b/app/Models/Device.php
@@ -265,14 +265,10 @@ class Device extends BaseModel
         return Permissions::canAccessDevice($this->device_id, $user->user_id);
     }
 
-    public function formatUptime($short = false)
+    public function formatDownUptime($short = false)
     {
-        return Time::formatInterval($this->uptime, $short);
-    }
-
-    public function formatDowntime($short = false)
-    {
-        return Time::formatInterval(time() - strtotime($this->last_polled), $short);
+        $time = ($device->status == 1) ? $this->uptime : time() - strtotime($this->last_polled);
+        return Time::formatInterval($time, $short);
     }
 
     /**
--- a/app/Models/Device.php
+++ b/app/Models/Device.php
@@ -267,7 +267,7 @@ class Device extends BaseModel
 
     public function formatDownUptime($short = false)
     {
-        $time = ($device->status == 1) ? $this->uptime : time() - strtotime($this->last_polled);
+        $time = ($this->status == 1) ? $this->uptime : time() - strtotime($this->last_polled);
         return Time::formatInterval($time, $short);
     }
 
--- a/app/Models/Device.php
+++ b/app/Models/Device.php
@@ -272,10 +272,7 @@ class Device extends BaseModel
 
     public function formatDowntime($short = false)
     {
-        $ret = static::where('hostname', $this->hostname)->first(['hostname', 'last_polled']);
-        if (!empty($ret)) {
-            return Time::formatInterval(time() - strtotime($ret->last_polled), $short);
-        }
+        return Time::formatInterval(time() - strtotime($this->last_polled), $short);
     }
 
     /**
--- a/app/Models/Device.php
+++ b/app/Models/Device.php
@@ -272,7 +272,10 @@ class Device extends BaseModel
 
     public function formatDowntime($short = false)
     {
-        return Time::formatInterval(time() - $this->last_polled, $short);
+        $ret = static::where('hostname', $this->hostname)->first(['hostname', 'last_polled']);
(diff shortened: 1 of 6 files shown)


includes/html/common/availability-map.inc.php
```diff
@@ -195,6 +195,7 @@
         $sql .= " ORDER BY `".$deviceOrderBy."`";
 
         $temp_output = array();
+        $updowntime = "";
 
         foreach (dbFetchRows($sql, $param) as $device) {
```
SourceDoctor, line 200: wouldn't it make more sense to do `$updowntime = "";` here to reset it's content "every round"?

includes/html/common/availability-map.inc.php
```diff
@@ -248,12 +251,12 @@
...
                     }
                     $temp_output[] = '
-                    <a href="' .generate_device_url($device). '" title="' . $device_system_name . " - " . formatUptime($device['uptime']) . '">
+                    <a href="' .generate_device_url($device). '" title="' . $device_system_name . $updowntime . '">
                         <span class="label ' . $deviceLabel . ' widget-availability label-font-border">' . $deviceState . '</span>
                     </a>';
                 }
             } else {
-                $temp_output[] = "<a href='" . generate_device_url($device) . "' title='" . $device_system_name . ' - ' . formatUptime($device['uptime']) . "'><div class='" . $deviceLabelOld . "' style='width:${compact_tile}px;height:${compact_tile}px;'></div></a>";
+                $temp_output[] = "<a href='" . generate_device_url($device) . "' title='" . $device_system_name . $updowntime . "'><div class='" . $deviceLabelOld . "' style='width:${compact_tile}px;height:${compact_tile}px;'></div></a>";
```
SourceDoctor, line 259: maybe mouseover text on down devices in red? maybe `#cc0000` ?

includes/html/common/availability-map.inc.php
louis-oui: fixed. Thanks

includes/html/common/availability-map.inc.php
```diff
@@ -217,11 +218,13 @@
                     $deviceLabelOld = 'availability-map-oldview-box-up';
                     $host_up_count++;
                 }
+                $updowntime = "" . ($device['uptime'] ? " - " : "") . formatUptime($device['uptime']);
```
SourceDoctor, line 221: wouldn't it be better to keep the GUI styling `($device['uptime'] ? " - " : "") .`  on to the link so `$updowntime` really only contains uptime/downtime?

includes/html/common/availability-map.inc.php
louis-oui: I don't know how to do it and not related to the fix

includes/html/common/availability-map.inc.php
louis-oui: no. It should also contain the name of the device

app/Models/Device.php
```diff
@@ -270,6 +270,14 @@ public function formatUptime($short = false)
         return Time::formatInterval($this->uptime, $short);
     }
 
+    public function formatDowntime($short = false)
+    {
+        $ret = static::where('hostname', $this->hostname)->first(['hostname', 'last_polled']);
```
murrant, line 275: Why are you running an sql query here?
`$this->last_polled` holds the value.

murrant: Why a seperate function from formatUptime?

louis-oui: > Why are you running an sql query here?
> `$this->last_polled` holds the value.
$this->last_polled returns null. I don't know why. Same with some other keys (last_ping...)

louis-oui: Device->formatUptime($short = false) does not take any time input argument. I cannot reuse it.

murrant: if last_polled holds null, it is because there was a select statement done excluding that value when the device model was loaded.

murrant: Device->formatUpDownTime($short = false) {
time = down ? now - last_polled : uptime
return format(uptime)

murrant: also, what do you think about putting a minus (-) sign before all downtimes?  that way you know what it is even out of context.

SourceDoctor: and i'd like it in a red font color ... so it also differs from normal uptime text.
what do you think about it @murrant

louis-oui: > if last_polled holds null, it is because there was a select statement done excluding that value when the device model was loaded. (or the device wasn't polled yet)

I will try to search for that statement

louis-oui: > Device->formatUpDownTime($short = false) {
> time = down ? now - last_polled : uptime
> return format(uptime)

OK for that

louis-oui: > and i'd like it in a red font color ... so it also differs from normal uptime text.
> what do you think about it @murrant

@SourceDoctor please tell me how to do it with href title !

murrant: you'd have to use a css popup (probably via bootstrap)

louis-oui: I've added downtime suffix
image

louis-oui: formatUpDownTime function factorization done

murrant:

murrant:

PR #11648, opened 2020-05-20
Fix test behaviour with PHP 7.4 on Fedora 32

TheMysteriousX:
I was seeing some odd test failures on my development system, running PHP 7.4.5.

The behaviour of hexdec seems to have changed:
  https://www.php.net/manual/en/migration74.deprecated.php#migration74.deprecated.core.invalid-base-characters

This could be a configuration/environment difference as it does not seem to happen on Travis, but the result is that when non-hex characters were passed to hexdec(), it would raise an exception on my test system.

"\<private\>" is special cased, as that's deliberately inserted by ```collect-snmp-data``` - this caused most of the failures I saw.
Otherwise an exception is raised if invalid data is seen, to prevent bad test data added in future. 

Only one snmprec needed to cahnged - it claimed a data type of hex, but was a plain string, which caused faulty test data to be generated. I don't know if this is user error, or if the device sent back faulty data.

#### Please note

#### Testers


murrant:
Probably should hex encode <private> for 4x fields in ./scripts/collect-snmp-data.php

TheMysteriousX:
Ah, there's the difference - snmpsim isn't appearing in my path for some reason.

Aside from ctm.snmprec everything looks ok other than the '<private>' records - what I can do is encode the '<private>' into hex to preserve the typing, which so it's not special cased.

TheMysteriousX:
I have no idea why the snmptrap test change is needed. On my development system the test fails as the event never makes it into the database.

Could be due to me running the very latest mariadb, but it sticks out because it was different from all the other tests in that module. It's now the same (and works for me).

murrant:
That fine, I think I had it different to verify it makes it to the db.  There should probably be a separate test for that.

LibreNMS/Util/ModuleTestHelper.php
```diff
@@ -487,7 +487,7 @@ private function cleanSnmprecData(&$data)
         foreach ($private_oid as $oid) {
             if (isset($data[$oid])) {
                 $parts = explode('|', $data[$oid], 3);
-                $parts[2] = '<private>';
+                $parts[2] = $parts[1] === '4' ? '<private>' : '3C707269766174653E';
```
murrant, line 490: I would change this to `$parts[1] == '4x' ? '3C707269766174653E' : '<private>'` but I'm not sure it makes a difference.

murrant:

PR #11585, opened 2020-05-11
Rewrite smokeping script to be an lnms command

TheMysteriousX:
I started off trying to make the existing smokeping script not kill smokeping if it was unable to resolve the device for any reason, and ended up rewriting it. Still to do is a documentation update, and some tests.

Smokeping's configuration format is... interesting, so I've taken a line-by-line approach to generating the configuration.

It is **not** a drop in replacement for the existing script, as it generates both probe and target blocks to support multi processing and IPv6, but can be invoked as follows to generate a compatible configuration:

```
./lnms smokeping:generate --targets --no-header --single-process
```

I need to update the docs, but all that should be needed to use it for new deployments is to remove the default smokeping probe and target stanza's.

This version will not insert unresolvable devices into the configuration:

```
...
# "junkdevice.example.org" was not resolvable and was omitted from the configuration
...
```

It will also support both IPv4 and IPv6 pollers, with multiple processes:

```
smokeping[4964]: Smokeping version 2.006010 successfully launched.
smokeping[4964]: Entering multiprocess mode.
smokeping[4964]: Child process 5011 started for probe lnmsFPing60.
smokeping[4964]: lnmsFPing60: probing 3 targets with step 60 s and offset 38 s.
smokeping[4964]: Child process 5012 started for probe lnmsFPing0.
smokeping[4964]: Child process 5013 started for probe lnmsFPing61.
smokeping[4964]: lnmsFPing0: probing 41 targets with step 60 s and offset 22 s.
smokeping[4964]: Child process 5014 started for probe lnmsFPing63.
smokeping[4964]: lnmsFPing61: probing 3 targets with step 60 s and offset 10 s.
smokeping[4964]: lnmsFPing63: probing 2 targets with step 60 s and offset 8 s.
smokeping[4964]: Child process 5015 started for probe lnmsFPing3.
smokeping[4964]: No targets defined for probe FPing6, skipping.
smokeping[4964]: No targets defined for probe FPing, skipping.
smokeping[4964]: lnmsFPing3: probing 40 targets with step 60 s and offset 4 s.
smokeping[4964]: Child process 5016 started for probe lnmsFPing2.
smokeping[4964]: lnmsFPing2: probing 40 targets with step 60 s and offset 42 s.
smokeping[4964]: Child process 5017 started for probe lnmsFPing62.
smokeping[4964]: lnmsFPing62: probing 3 targets with step 60 s and offset 49 s.
smokeping[4964]: Child process 5018 started for probe lnmsFPing1.
[...]

PR #11653, opened 2020-05-21
Enhance the SmokePing script and documentation

cjwbath:
The current SmokePing integration script/docs requires `+LibreNMS` to be manually prepended in cron jobs that call _gen_smokeping.php_, and even then it doesn't give a very nicely structured menu in SmokePing because of the top level Targets that come enabled by default.

I've tweaked the script to not add any top level menu entry (which will mean the LibreNMS device types get added in directly under standard Top level menu) unless the -t argument is given, in which case all the LibreNMS menu levels will be incremented so everything slots in nicely under a LibreNMS subheading to coexist with existing SmokePing configurations.

Also added an experimental option -g that allows organisation by Device Group instead of Type but doing that will prevent LibreNMS being able to find the the RRDs because they won't be in the expected locations. Potentially could result in device duplication in SmokePing as well.

This change should be backward compatible for people with existing setups installing the new version of the script. Documentation updated accordingly.

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Smokeping.md
+++ b/doc/Extensions/Smokeping.md
@@ -1,7 +1,7 @@
 source: Extensions/Smokeping.md
 path: blob/master/doc/
 
-# Smokeping integration
+# SmokePing integration
 
 [SmokePing](https://oss.oetiker.ch/smokeping/) is a tool which lets us
 keep track of network latency, and visualise this through RRD graphs.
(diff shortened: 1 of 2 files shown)


TheGreatDoc:
Hi,

Maybe you should take a look here https://github.com/librenms/librenms/pull/11585 so both of you work on the same PR.

I like the group idea, but it must be an option for config.php so recent link integration https://github.com/librenms/librenms/pull/11610/files#diff-9e55dd105036e39411d460a4b1a4c741 could be modified to work with both, type and group.

cjwbath:
Yes, #11585 looks better overall so I'll close this and consider any contributions I might make to that.

PR #11688, opened 2020-05-25
Since Rancid 3, the separator is ;

dupondje:
Another minor change to the gen_rancid script.

Since rancid 3.0, the separator is ; instead of :.
See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=778623

### DIFF
--- a/scripts/gen_rancid.php
+++ b/scripts/gen_rancid.php
@@ -56,7 +56,7 @@ foreach (dbFetchRows("SELECT `hostname`,`os`,`disabled`,`status` FROM `devices` 
         if ($devices['disabled']) {
             $status = "down";
         }
-        echo $devices['hostname'] . ':' . $rancid_map[$devices['os']] . ':' . $status . PHP_EOL;
+        echo $devices['hostname'] . ';' . $rancid_map[$devices['os']] . ';' . $status . PHP_EOL;
     }
 }
 echo "# EOF " . PHP_EOL;


PipoCanaja:
How does it behave it you have an older rancid ? Seems that it breaks ?

dupondje:
Correct! But hey, rancid 3.0 was released 2014 ... So time to get moving? :)

PipoCanaja:
If you consider that you are the 1st one to come with a PR, it could mean:
- People are using and trusting RANCID < v3
- People are not using RANCID :) 
As we cannot tell, you have 2 options: 
- Find a way to be backward compatible (seems difficult if not impossible)
- Notify the users of the change, schedule it in 2 weeks or so, and move LibreNMS to v3 only

dupondje:
Could add some config item that defaults to ';' for example, and allow it to change to ':' for people using old rancid?
Then there is still a notification, but at least its possible to 'unbreak' it :)

But really, as rancid 3.0 is default since Debian 9 and Ubuntu 16.04 for example, it seems logic to just change to the new style.

Jellyfrog:
Imo we should just change it

PipoCanaja:
OK. So notification, and change to new mode. At least if somebody uses the old way, he'll know why it is broken and will fix it.

laf:
Could we not just add an argument to this script so you can pass `-v 3` with no arg defaulting to < 3?

Jellyfrog (approved):

murrant:

PR #11698, opened 2020-05-25
Fix missing PoE port graphs for Cisco Catalyst 9K

ajsiersema:
The PoE port graphs are missing on Cisco Catalyst 9K devices.
Polling code assumes interfaces named *Ethernet[SLOT#]/[SUBSLOT#]/[PORT#] will have corresponding entries in the SNMP table, e.g. cpeExtStuff.3.0.1 for GigabitEthernet3/0/1
However, Cisco decided to omit the 0 of the SUBSLOT, the SNMP entry is thus  cpeExtStuff.3.1

This pull requests registers both variations in order to match both the Cat 9K devices, and any devices conforming to the existing assumption with regard to interface name<->SNMP name relation.  
The change will not break existing behavior, but only adds the Cat 9K case.
AFAIK there are no ios/iosxe devices that have both
GigabitEthernet[1-9]+/0/[0-9]+    i.e. slot/subslot/port
as well as
GigabitEthernet[1-9]+/[0-9]+        i.e. slot/port
All the slots with subslots start at 1, with possibly a single OOB interface GigabitEthernet0/0 (in slot 0.)

Tested on IOS-XE 16.9.x

#### Please note

#### Testers


ajsiersema:
Fixing codeclimate "Avoid deeply nested control flow statements" requires extensive code rewrite.
The same codeclimate issue popped up in other PR's, where the consensus seems to be that we can ignore this message in these cases.

includes/polling/ports.inc.php
```diff
@@ -334,7 +334,23 @@
...
+                    The ...EthernetX/Y/Z SNMP entries on Catalyst 9x00 iosxe
+                    are cpeExtStuff.X.Z instead of cpeExtStuff.X.Y.Z
+                    Creating both:
+                      $port_ent_to_if[slot.subslot.port]
+                      $port_ent_to_if[slot.port]
+                    will find the PoE stats for Cat9K and will also match for
+                    devices that do include the subslot (for subslot == 0)
+                    */
+                    $iface_parts = explode('/', $matches[1]);
+                    while (1) {
```
murrant, line 347: while (1) needs to change, you could honestly update the regex above and drop most of this code.

ajsiersema: I left the regexp from the original code as we still need to match ethernet interfaces and capture the slot[/subslot]/port part (to be able replace '/' by '.' later.)
Capturing repeated subgroups in PHP's PCRE would make for a hard to read regexp. Maybe that was your intention ? I'm not quite sure what you meant by 'update the regexp'.

We also still need to add both slot.subslot.port and slot.port to $port_ent_to_if, i.e. two assignments (at least). Otherwise we'd break the poller for non-cat9K devices.
We don't want to strip the subslot element if it's non-zero, i.e. we need to check that part of the interface notation.
Maybe your assumption is that there will always be either slot/subslot/port or slot/port interfaces, and not e.g. slot/subslot/subsubslot/port. The PR code doesn't make any assumptions.

ajsiersema: The while (1) is in effect a do..while, preventing a duplication of the assignment statements.
I'll see if  a variation on that theme with a for loop or some such is less offensive to the eyes :)

murrant:
```php
 if (preg_match('/^[a-z]+ethernet(\d+)(?:\/(\d+))(?:\/(\d+))?$/i', $if_descr['ifDescr'], $matches)) {
port_ent_to_if[$matches[1] . '.' . $matches[3] ?: $matches[2]] = ['portIfIndex' => $if_index];
}
```

Should be it.

ajsiersema: > Should be it. Specifically only matches XethernetX/X or XethernetX/X/X
> If cpeExtPsePortEntry can be indexed differently, this code is probably slightly wrong.

That's indeed the catch: I fully expect there to be an existing ios(xe) device that indexes XethernetA/B/C in cpeExtPsePortEntry with a.b.c. even with B == 0. Breaking the poller is bad, hence the code to add both A.B.C. and A.C to $port_ent_to_if, for arbitrary 'depths' of interface names.

However, if we're working with the assumption that there will only ever be XethernetX/X or XethernetX/X/X (i.e. two or three parts notation), we can refactor the PR as two assignment statements, like so:

```php
if (preg_match('/^\w+ethernet(\d+)\/(\d+)(?:\/(\d+))?$/i', $if_descr['ifDescr'], $matches)) {
    $port_ent_to_if[$matches[1] . '.' . $matches[2]] =
        $port_ent_to_if[$matches[1] . '.' . ($matches[3] ?: $matches[2])] =
            ['portIfIndex' => $if_index];
}
```
What do you think ?
(or split it up over two individual assignment statements)

ajsiersema: > Should be it. Specifically only matches XethernetX/X or XethernetX/X/X
> If cpeExtPsePortEntry can be indexed differently, this code is probably slightly wrong.

That's indeed the catch: I fully expect there to be an existing ios(xe) device that indexes XethernetA/B/C in cpeExtPsePortEntry with a.b.c. even with B == 0. Breaking the poller is bad, hence the code to add both A.B.C. and A.C to $port_ent_to_if, for arbitrary 'depths' of interface names.

However, if we're working with the assumption that there will only ever be XethernetX/X or XethernetX/X/X (i.e. two or three parts notation), we can refactor the PR as two assignment statements, like so:

```php
if (preg_match('/^\w+ethernet(\d+)\/(\d+)(?:\/(\d+))?$/i', $if_descr['ifDescr'], $matches)) {
    foreach (array(0=>0) + ($matches[3] ? array(1=>2):array()) as $idx) {
        unset($matches[$idx]);
        $port_ent_to_if[implode('.', $matches)] = ['portIfIndex' => $if_index];
    }
}
```
What do you think ?

murrant: Why do you expect that? Generally the table indexes are fixed.  I don't have any sample data to l look at so I can't confirm.  (btw you should add test data to this PR.

murrant: why the loop?
just add two entries into the $port_ent_to_if array

ajsiersema: Added tests. I've check with ```./lnms dev:check -o iosxe_c9400``` , they seem to pass.
Not sure about the 2nd ASA device in the JSON, did snmpsim inject it ?

ajsiersema: codeclimate once again complains about "Avoid deeply nested control flow statements". Can't be fixed without drastic rewriting all of the code.

Node analysis timed out "The command produced no output for 180 seconds. You can increase this timeout by defining the "idle_timeout" modifier.". Not something I can fix :)

murrant:
Understood, you aren't expected to fix the underlying code issues, just hopefully not make it worse ;)

murrant:
Also, `enable_ports_poe` is false by default, so I'm not sure if your code is being tested.

murrant:

murrant:

PR #11699, opened 2020-05-25
Fix dynamic group membership rule for devices.status equal 0

ajsiersema:
A dynamic group membership rule can be added with “devices.status equal 0”.
This would be very useful for e.g. Maps -> Device Group Dependencies
However, the device group membership is updated by the poller, and the poller code won’t update the device group membership for devices that are down.

This PR runs group membership updates for down and up devices, fixing e.g. network maps with device group dependencies.

#### Please note

#### Testers

### DIFF
--- a/includes/polling/functions.inc.php
+++ b/includes/polling/functions.inc.php
@@ -224,6 +224,22 @@ function record_sensor_data($device, $all_sensors)
     }
 }
 
+
+/**
+ * @param array $device The device for which to update the groups for
+ */
+function poll_update_device_groups($device)
+{
+    // Update device_groups
+    echo "### Start Device Groups ###\n";
+    $dg_start = microtime(true);
+    $group_changes = \App\Models\DeviceGroup::updateGroupsFor($device['device_id']);
+    d_echo("Groups Added: " . implode(',', $group_changes['attached']) . PHP_EOL);
+    d_echo("Groups Removed: " . implode(',', $group_changes['detached']) . PHP_EOL);
+    echo "### End Device Groups, runtime: " . round(microtime(true) - $dg_start, 4) . "s ### \n\n";
+}
+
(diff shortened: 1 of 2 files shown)


murrant:
I appreciate you extracting a function instead of just copying the code, but it would be better if you just called the code every time in the same spot.  The way you have done it makes it harder to understand/read the code.

My suggestion, but it right before the code that runs alerts.

ajsiersema:
> My suggestion, but it right before the code that runs alerts.

Hehe, that's exactly what I did in the initial version, but then decided you might like the patch better if it was made explicit that updating the groups is done in both cases.

I'll rework the PR by moving the code out of the function again.

murrant:
No need to squash, we squash when we merge to make nice tidy commits in master.

murrant:
You can see how confusing that is right?
if ($blah) {
wall of code...
do_something()
wall of code.. 
} else {
do_something()
}

ajsiersema:
> You can see how confusing that is right?

The first version I came up with moved the code into a function, in the include, then called that function from poller.php. Kinda how I would have implemented it to begin with.
Which then got me thinking, that although I thought it was very confusing that the device groups weren't always updated (principle of least surprise and all that), I wasn't there when the decision was made to place the code conditionally in the include cod. So there might have been a very good reason, and maybe it was decided to keep the poller.php as small/clean as possible. Hence, I moved all the code (back) info the include.

Anyway, thanks for your assistance and the merging ! (and all the other work on librenms)
Dynamic maps based on dynamic device groups are a nice feature.

murrant:

murrant:
@ajsiersema There was likely no decision.  Remember LibreNMS is 100% community supported. Someone probably changed something and didn't realize the rammifications.

PR #11702, opened 2020-05-26
QNAP QTS: Moved to yaml discovery

TheGreatDoc:
Sadly historical data will be lost as now index is dynamic and not hardwritted.

I've also grouped them so its easy to read

I dont have any QNAP with extension, but with this discovery it should now discover everything.

TODO in other PR: A way to know the raid status. QNAP decided to return it as STRING... And I dont know all possible states.

#### Please note

#### Testers


murrant:
index can be dictated in yaml (to some degree)
Ouch all these devices that don't support bulk request.  That REALLY slows down polling.

TheGreatDoc:
> index can be dictated in yaml (to some degree)

I know, and maybe I could save one or two, but will break a sanitized way to name the rrd sensors. So I though it will be better to _restart_ even with the historical data lost. 
After all, no storage/mem/cpu/interface data is lost, only state/temp/fan sensors.

>Ouch all these devices that don't support bulk request. That REALLY slows down polling.
Funny thing: I have 2 QNAP, one at home TS-253A and one at work TS-877XU-RP. 
The TS-253A doesnt respond to bulk but TS-877XU-RP does... Maybe split to two different definitions?

murrant:
@TheGreatDoc one way to disable bulk from the user perspective is to use SNMPv1

jozefrebjak:
@TheGreatDoc I will test it on Turbo NAS TS-469L & Turbo NAS TS-469 Pro

TheGreatDoc:
> @TheGreatDoc one way to disable bulk from the user perspective is to use SNMPv1

Yeah, but from user perspective what you will see is your device not being _fully_ supported.

I'm going to check discovery/poller times with the nobulk option and compare results.

jozefrebjak:
@TheGreatDoc It's much better now. I can see wrong disk finnaly which I know that is wrong some time with HD SMART state. Thank you. I am waiting for merge to master.

TheGreatDoc:
@murrant could be that only one var is allowed for `index:` ?

I just noticed this:

```
Sensor Added: temperature qnap diskTemperture-{{ $diskEnclosureID }}.1 Enclosure 0 Disk 1 Seagate ST4000VN008-2DR166 (SATA). 
```

murrant:
@TheGreatDoc possibly diskEnclosureID was not found.

TheGreatDoc:
@murrant It is. 
Is used in descr: `Enclosure ` **0** `Disk` 1

PipoCanaja:
Right @TheGreatDoc , could be that not all the variables are parsed in `index`, only $index and co.

TheGreatDoc:
@jozefrebjak Do you have a QNAP with extension enclosure?

I dont and I want to check how is the snmp reply with it.

TheGreatDoc:
Oh... I think I totally fucked up....

murrant:
@TheGreatDoc no worries, there is always a way back :D

TheGreatDoc:
Closed. Check https://github.com/librenms/librenms/pull/11882

PR #11709, opened 2020-05-26
Don't call clean() on the inputs to the SNMP settings form

cjwbath:
The device SNMP settings form calls `clean()` on all the data it gets. This in turn calls `strip_tags()` which breaks some values that contain angle brackets. In my case I discovered it because one of my SNMPv3 authentication passwords has angle brackets - perfectly allowable I would say, but the value appeared truncated when updated to the database and then didn't work for SNMP to the device.

Since the database updates/inserts are parameterised anyway, it is inappropriate to "sanitize" this early and I've removed it. If we're worried about HTML tags in data, it's better to use `htmlspecialchars` at the point  where the data is rendered into HTML (which I've added here).

#### Please note

#### Testers

### DIFF
--- a/includes/html/pages/device/edit/snmp.inc.php
+++ b/includes/html/pages/device/edit/snmp.inc.php
@@ -230,7 +230,7 @@ echo "
     <div class='form-group'>
     <label for='sysName' class='col-sm-2 control-label'>sysName (optional)</label>
     <div class='col-sm-4'>
-    <input id='sysName' class='form-control' name='sysName' value='".$device['sysName']."'/>
+    <input id='sysName' class='form-control' name='sysName' value='".htmspecialchars($device['sysName'])."'/>
     </div>
     </div>
     <div class='form-group'>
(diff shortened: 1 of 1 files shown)


murrant:
Looks good, but need to go through and verify we aren't leaking any unsanitized data.

murrant:
@cjwbath You should sanitize all the values displayed on this page, since you remove the incoming protection.  Otherwise it is js injection vulnerability.

cjwbath:
@murrant I think I _have_ sanitized anything that potentially came from user input. Was there something specific you saw?

murrant:
Pretty much all of them came from user or outside input in some way.

cjwbath:
OK I have added some more escaping to cover stuff that should be validated elsewhere (i.e. numbers) and stuff coming from SNMP. But the original `clean()` that I got rid of only covered the form inputs.

murrant:

murrant:

PR #11725, opened 2020-05-28
Remove int width from db schema (MySQL 8)

arrmo:
As discussed here, this is the correction to remove the width,
https://github.com/librenms/librenms/pull/11533#issuecomment-635131161

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2238,7 +2238,7 @@ function dump_db_schema($connection = null)
         foreach (DB::connection($connection)->select(DB::raw("SELECT COLUMN_NAME, COLUMN_TYPE, IS_NULLABLE, COLUMN_DEFAULT, EXTRA FROM information_schema.COLUMNS WHERE TABLE_SCHEMA = '$db_name' AND TABLE_NAME='$table'")) as $data) {
             $def = [
                 'Field'   => $data->COLUMN_NAME,
-                'Type'    => trim(preg_replace('/\([0-9]+\)/', '', $data->COLUMN_TYPE)),
+                'Type'    => preg_replace('/int\([0-9]+\)/', 'int', $data->COLUMN_TYPE),
                 'Null'    => $data->IS_NULLABLE === 'YES',
                 'Extra'   => str_replace('current_timestamp()', 'CURRENT_TIMESTAMP', $data->EXTRA),
             ];
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2238,7 +2238,7 @@ function dump_db_schema($connection = null)
         foreach (DB::connection($connection)->select(DB::raw("SELECT COLUMN_NAME, COLUMN_TYPE, IS_NULLABLE, COLUMN_DEFAULT, EXTRA FROM information_schema.COLUMNS WHERE TABLE_SCHEMA = '$db_name' AND TABLE_NAME='$table'")) as $data) {
             $def = [
                 'Field'   => $data->COLUMN_NAME,
-                'Type'    => $data->COLUMN_TYPE,
+                'Type'    => trim(preg_replace('/\([0-9]+\)/', '', $data->COLUMN_TYPE)),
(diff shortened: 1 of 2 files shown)


murrant:
it is only integer width that is affected right?

arrmo:
> it is only integer width that is affected right?

No, at least here it looks like I see issues with int, tinyint, varchar, smallint ... OK, I'm not keeping on looking. LOL.

arrmo:
BTW, a thought (dumb question?) here. The code I added works, removes the '(number)', which then lets the schema check pass, but ... I dumped the schema then from my database, and afterwards validate.php checks against that ... right? So I'm sort of checking against myself 😆. Is that what should happen? Or is the dump a one time thing, to get back in sync?

Sorry, just trying to understand.

Thanks!

murrant:
dump creates a record of the current schema.
Validate runs dump and compares the saved one against the one it just dumped.

arrmo:
OK, makes sense, thanks! Should I also add that dump (i.e. latest misc/db_schema.yaml) to this PR?

arrmo:
Found it! :-). Bit of code below, it's in LibreNMS/Validations/Database.php, inside private function checkSchema(Validator $validator),

```
                    // MySQL 8 fix, remove DEFAULT_GENERATED from timestamp extra columns
                    if ($cdata['Type'] == 'timestamp') {
                         $current_columns[$column]['Extra'] = preg_replace("/DEFAULT_GENERATED[ ]*/", '', $current_columns[$column]['Extra']);
                    }
```

I'm not sure why this is used / needed, but due to this the check removes DEFAULT_GENERATED! As this is not removed on the generation / dumping side ... the check fails, for timestamps (as I have been seeing). So, adding the code below to includes/functions.php, and then re-generating the schema ... it passes now!

```
            // MySQL 8 fix, remove DEFAULT_GENERATED from timestamp extra columns
            if ($def['Type'] == 'timestamp') {
                 $def['Extra'] = preg_replace("/DEFAULT_GENERATED[ ]*/", '', $def['Extra']);
            }
```

Do I go ahead and add this matching code to includes/functions.php, and then add this file (along with the new schema) to this PR?

Thanks!

arrmo:
OK, found where this was added! https://github.com/librenms/librenms/pull/9923/

Seems recent mysql updates added this, so the noted PR removed it ... but then after re-generating the schema (which didn't remove it) ... the timestamp issue I saw. I think I need to add this then (as above), and also update the schema file. Please just let me know - if you agree, I'll add the two of them to this PR. Then old and new schema files will both work.

murrant:
Yeah, add them, you don't need my permission to update the PR :)

Could you try changing your regex to `preg_replace('/int\([0-9]+\)/', 'int', $data->COLUMN_TYPE)`?

arrmo:
> Yeah, add them, you don't need my permission to update the PR :)

LOL. Just didn't know what files you want in the repository, vs. local. That's why I asked. Will do though!

> Could you try changing your regex to `preg_replace('/int\([0-9]+\)/', 'int', $data->COLUMN_TYPE)`?

I can, but not all the items being replaced are int ... still want to do this?

murrant:
@arrmo  your production DB has some oddities :)
I did a fresh db migration and dump and uploaded it after setting it to only scrub length from integer types.
For future you might use DB_TEST_DATABASE to specify an alternate database to for tests.  Then you can `./lnms migrate:fresh --database=testing` and `DBTEST=1 ./scripts/build-schema.php`  (note migrate:fresh erases all db data, don't do it on your prod :D)

arrmo:
> your production DB has some oddities :)

OK, odd, but one of those things I guess ... LOL. No biggie, thanks for explaining! To understand ... it looks like you updated the files (thanks!) - so do I just wait for the merge, then get back on master, and will have your changes in place?

But also, do I need to do anything to my (production) DB, or will this all get cleaned up with your updates?

Thanks again.

murrant:

PR #11734, opened 2020-05-29
Added upper limit to entSensorPrecision to prevent excessive padding

ryanheffernan:
We are experiencing frequent crashes of librenms due to a router bug in some Cat9k switches that are reporting a garbage high value for entSensorPrecision. Example:
```
iso.3.6.1.4.1.9.9.91.1.1.1.1.3.2126 = INTEGER: 1615384784
```

This causes librenms to padd the sensor value with 1615384784 zeroes which obviously fails to be written into the database and prevents the discovery of this interface. Librenms then logs this failure and fills up the disk space in /opt causing the application to crash. See the below log message and imagine the '0's keep going until the log file is 85GB long. 
```
[2020-05-23 02:47:07] production.ERROR: SQLSTATE[HY000]: General error: 2006 MySQL server has gone away (SQL: INSERT IGNORE INTO `sensors` (`poller_type`,`sensor_class`,`device_id`,`sensor_oid`,`sensor_index`,`sensor_type`,`sensor_descr`,`sensor_divisor`,`sensor_multiplier`,`sensor_limit`,`sensor_limit_low`,`sensor_current`,`entPhysicalIndex`,`entPhysicalIndex_measured`)  VALUES (snmp,voltage,9,.1.3.6.1.4.1.9.9.91.1.1.1.1.4.2141,2141,cisco-entity-sensor, - Te2/0/17 Bias Current ,10000000000000
```

This is obviously a router bug that is being reported to Cisco however I think librenms should have some prevention for this instead of simply accepting the precision value and padding it to whatever it receives. I arbitrarily selected 100k as a reasonable upper limit for how many decimal places to pad since I can't imagine any sensor with that level of precision in a realistic scenario. I don't exactly like this solution and would be open to alternate suggestions. 

#### Please note

#### Testers


murrant:
probably better to just ignore ones that contain integer max value `1615384784`

ryanheffernan:
Fair enough. I wasn't convinced that this value would be the only erroneous value reported but it's better than nothing.

ryanheffernan:
@murrant let me know if this works for you

ryanheffernan:
need to resubmit PR with different email

PR #11779, opened 2020-06-05
Convert LastSync date for oxidized config info

Negatifff:

#### Please note

#### Testers

### DIFF
--- a/includes/html/pages/device/showconfig.inc.php
+++ b/includes/html/pages/device/showconfig.inc.php
@@ -132,10 +132,10 @@ if (Auth::user()->hasGlobalAdmin()) {
         // fetch info about the node and then a list of versions for that node
         $node_info = json_decode(file_get_contents(Config::get('oxidized.url') . '/node/show/' . $oxidized_hostname . '?format=json'), true);
 
-        if (Config::get('oxidized.datetime_format') != "")
+        if (Config::get('oxidized.datetime_format') != "") 
         {
-            $node_info['last']['start']=date(Config::get('oxidized.datetime_format'),strtotime($node_info['last']['start'] . ' UTC'));
-            $node_info['last']['end']=date(Config::get('oxidized.datetime_format'),strtotime($node_info['last']['end'] . ' UTC'));
+            $node_info['last']['start']=date(Config::get('oxidized.datetime_format'), strtotime($node_info['last']['start'] . ' UTC'));
+            $node_info['last']['end']=date(Config::get('oxidized.datetime_format'), strtotime($node_info['last']['end'] . ' UTC'));
         }
         // Try other hostname format if Oxidized request failed
         if (! $node_info) {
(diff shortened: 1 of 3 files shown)


murrant:
Why is this an option and not just done for all?

Negatifff:
I did not dare to take such responsibility. provided the user with a choice.

murrant:
Not making a choice increases future maintenance costs, and increases the burden on the user to configure.

This should probably follow the user's timezone setting.

Negatifff:
this setting already looks into the user's timezone settings. optionally done because I don’t know if it will be fixed in oxidized or not. Now the author says that he will not correct it, but after some time he can change his mind. and the format is set by the user for his own convenience - for example, the Y-m-d format is unusual for me to read. I am more comfortable d.m.Y. therefore, the option suggested itself :).

murrant:
I'm not trying to be difficult I promise, but I need you to think about some things.

How many time format/zone settings should we have?

Negatifff:
maybe I did not quite understand the question. we have a date that needs to be transferred from the UTC timezone to the user's timezone. The user's timezone is specified in his php.ini. from there we will recognize her. but how to set the date and time format? in php.ini it is not. maybe the question is not to indicate the format of the date and time, but it seems inconvenient to me.

samotelf:
@Negatifff are you talking about when you go to the device config the Last Sync is in UTC? I also would like this.
To clarify the Last Sync should be the system time:

@murrant not sure if this an oxidized issue or not, but do you think you can please change it?


image

murrant:
Please use the timezone from `\config('app.timezone')`  instead of making a new setting.

Negatifff:
to be honest, it turned out to be the most mysterious patch. I added the ability to adjust the time zone and the format of the time output, but for some reason this does not work. yes, in pkhp there is a default timezone in the config - no question, this is used. but where to get the default date and time output format? I personally find it unusual to use Y-m-d - what should I do?

murrant:
Right now app.timezone follows the php setting, but in the future, it will allow user and/or browser to set the value.
As for time format, we try to use ISO 8601 for the most part, but for dateformats, we have them defined in the config as dateformat.compact or dateformat.long for example.

samotelf:
> Right now app.timezone follows the php setting, but in the future, it will allow user and/or browser to set the value.
> As for time format, we try to use ISO 8601 for the most part, but for system defined formats, we have them defined in the config as dateformat.compact or dateformat.long for example.

Thank you Murrant for your reply.   
I believe I'll wait for the future release, any idea when will it be available?
Let me also take the opportunity to appreciate all your effort.

Negatifff:
no more configs

murrant:
@samotelf LibreNMS is a community project it will be added when ever anyone deems it worth there time to add and contributes it.  I'm just trying to keep us from kicking ourselves later if/when it is added.

Negatifff:
@murrant thanks

murrant:
Thank you @Negatifff  :)

murrant:

PR #11790, opened 2020-06-08
Change Units on Noise Floor

arrmo:
I may be wrong, but I think this should just be dBm (i.e. integrated over the carrier bandwidth). The numbers look much too high to me to be on a per Hz basis.

If I am correct here, this should also be updated in the web interface.

Thanks!

#### Please note

#### Testers

### DIFF
--- a/doc/Developing/os/Wireless-Sensors.md
+++ b/doc/Developing/os/Wireless-Sensors.md
@@ -19,7 +19,7 @@ with the values we expect to see the data in:
 | errors      | count       | WirelessErrorsDiscovery      | The total bits of errored packets or bits, etc                                                  |
 | frequency   | MHz         | WirelessFrequencyDiscovery   | The frequency of the radio in MHz, channels can be converted                                    |
 | mse         | dB          | WirelessMseDiscovery         | The Mean Square Error                                                                           |
-| noise-floor | dBm/Hz      | WirelessNoiseFloorDiscovery  | The amount of noise received by the radio                                                       |
+| noise-floor | dBm      | WirelessNoiseFloorDiscovery  | The amount of noise received by the radio                                                       |
 | power       | dBm         | WirelessPowerDiscovery       | The power of transmit or receive, including signal level                                        |
 | quality     | %           | WirelessQualityDiscovery     | The % of quality of the link, 100% = perfect link                                               |
 | rate        | bps         | WirelessRateDiscovery        | The negotiated rate of the connection (not data transfer)                                       |
--- a/includes/html/functions.inc.php
+++ b/includes/html/functions.inc.php
@@ -1551,6 +1551,9 @@ function get_sensor_label_color($sensor, $type = 'sensors')
         $sensor['sensor_current'] = formatUptime($sensor['sensor_current'] * 60, 'short');
         return "<span class='label $label_style'>".trim($sensor['sensor_current'])."</span>";
     }
+    if ($sensor['sensor_class'] == 'frequency' && $sensor['sensor_type'] == 'openwrt') {
+        return "<span class='label $label_style'>".trim($sensor['sensor_current'])." ".$unit."</span>";
+    }
     return "<span class='label $label_style'>".trim(format_si($sensor['sensor_current']).$unit)."</span>";
 }
 
--- a/includes/html/graphs/device/wireless_noise-floor.inc.php
+++ b/includes/html/graphs/device/wireless_noise-floor.inc.php
@@ -5,6 +5,6 @@ $scale_max = '0';
 
 $class     = 'noise-floor';
 $unit      = '';
-$unit_long = 'dBm/Hz';
+$unit_long = 'dBm';
 
 require 'includes/html/graphs/device/wireless-sensor.inc.php';
--- a/includes/html/graphs/wireless/noise-floor.inc.php
+++ b/includes/html/graphs/wireless/noise-floor.inc.php
@@ -3,6 +3,6 @@ $scale_min = '-175';
 $scale_max = '0';
 
 $unit_long = 'Noise Floor (dBm/Hz)';
-$unit = 'dBm/Hz';
+$unit = 'dBm';
 
 include 'wireless-sensor.inc.php';
--- a/resources/lang/de/wireless.php
+++ b/resources/lang/de/wireless.php
@@ -64,7 +64,7 @@ return [
     'noise-floor' => [
         'short' => 'Grundrauschen',
         'long' => 'Grundrauschen',
-        'unit' => 'dBm/Hz',
+        'unit' => 'dBm',
     ],
     'power' => [
         'short' => 'Signalstärke',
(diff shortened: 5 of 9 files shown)


murrant:
Which sensor are you talking about?  There are many noise floor sensors, some in dbm, some in Hz.

SourceDoctor:
@armo @murrant  dBm/Hz can be correct ...
especially for a "noise floor"
if value is high it's a good indicator this value is in dBHz.
dB/Hz i have never seen in my praxis. What says the data sheet?
For for generric Noisefloor i only know dBHz not dB/Hz, so i think it should be changed to dBHz

short excursion:
dBHz - MHz = dB
which means:
spectral noisefloor - used Bandwidth = Carrior to Noise ratio
Example:
70 dBHz - 60 (MHz logarithmic) = 10dB C/N


as you see the dBHz value has to be high

arrmo:
Sorry, not quite following your comment - but may be me. Let me try to explain my thought though.

If this is really noise floor ... well, then starting from the thermal noise floor (-174 dBm/Hz), and if we assume a NF (noise figure) of ~ 10 dB (guess here a bit), then the noise floor would be roughly -164 dBm/Hz. Integrating over a 20 MHz carrier BW, this adds 73 dB for 10*log(BW) ... puts the integrated noise floor at roughly -91 dBm (not per Hz, this is integrated over BW). As the results are pretty much in this range, it seems to me that this measurement really is integrated ... so dBm, not per Hz.

BTW, also I do see from iwinfo on the router, noise reported as (example) -89 dBm.

Thoughts?

Thanks!

SourceDoctor:
your assume sounds right on reading over it.
-91dBm . -89dBm are not such far away ... especially for classical noise floor in RF equipment.
Keep in mind noise floor depends on many many many parameters, also on component temperatures, age of components, quality, RF injections into components and antenna, quality of isolation of RF parts, and so on... so a noise floor up to nearly -85dBm is quiet a normal dimension

arrmo:
Completely agree! And the 10 dB NF was a guess on my part ;-).

Also of course, this measurement is not just NF - it also detects environment noise floor (which is good).

SourceDoctor:
problem with the unit here and it's usage is ...
is it the real noise floor? so it's dBm. 
is it the noise figure of the system? so it's dBHz

would be useful to recheck where this unit is used here and recheck the datasheets of the devices.
Which potentially means correction of unit description on several devices and potential add another unit

arrmo:
Ahh, got your point now 😆.

Yes, agreed - it depends on what the / each device reports ... noise figure, floor, spectral density, etc. Hmmm ... not sure how to confirm that, any thoughts? At least for the devices I'm playing with, they are reporting integrated noise floor, but as you note - this may or may not be consistent.

Thanks!

SourceDoctor:
i appreciate your PR ... but i think there is much more to do as first thought ;-)
especially grepping through existing devices to see which of them use this unit
and looking into their datasheets if it is used correctly.
Adding new units is no problem, but it has to be used correctly, otherwise it makes no sense ...
Define there wrong units is like define Disc capacity in Mb/s or worse MB/s :D

arrmo:
Yes, agreed! And no worries if the PR needs to be rejected, I just noticed that this is "wrong" ... but for my devices. To your point, it doesn't mean that this is the case generically - I wasn't thinking that widely 😆.

Thanks!

SourceDoctor:
it's good to fix errors :) we appreciate this. if you've found a missing unit for this on to your device. Feel free to fix it.

arrmo:
Well, my devices (2 OS's) report dBm ... I guess what I don't know is what other devices are reporting - meaning, what is the generic answer? I'm OK with dBm, or dBm/Hz, if one or the other is used more widely. Do you happen to know? I admit, not sure what other devices are reporting 😞.

Thanks!

SourceDoctor:
i took a short look over code. There are several devices which use this unit. i think this has to be overlooked.

arrmo:
NP, I'm good with whatever you conclude - just noticed it, thought I'd try to correct it.

BTW, finding a few other "odd" unit related items (like frequency in kMHz) - these are minor of course, just trying to help clean up if I can. That said, I REALLY like the Wireless Sensors approach. Just took a bit of time to figure it out, some IBM issues ... 😆.

*IBM = Idiot Behind the Machine => me!

SourceDoctor:
feel free to correct everything you find 
on some things it may more to do than thought on first look, but it will be appreciated :)

for direct questions we'll help you in community.librenms.org or in discord 

to come back to this PR it would be nice if you would fix the units all over

arrmo:
Sure, will try to correct these, NP at all! I admit, I tried to run the maze of finding where the units are defined in the code .. felt like a mouse after cheese. LMAO!

Will keep hunting, NP - if you have any pointers where to look, it would be greatly appreciated.

Thanks!

SourceDoctor:
i found them with grep ;-)

arrmo:
Dang it! I tried that - hmmm, that's likely me. Will dig!

arrmo:
OK, easy fix for the GUI display ... do you want me to go ahead and submit a PR (change it to dBm?). FYI, still need to find the RRD generation - that's different.

And frequency - that's different also, that's about converting > 1000 (or 1024) to "k".

Thanks!

SourceDoctor:
fix what you find :)

murrant:
The graphs: rrdtool does automatic prefixing... and sucks at some times.  You could maybe convert the data to Hz in the graph and hopefully the graph would prefix correctly.  Or probably someway to disable the graph prefixing.

arrmo:
Actually, frequency is fine in rrd, it's just the dBm/Hz I need to fix there, will find it.

Frequency is messed up in the web GUI, Wireless > Frequency. Will hunt for that too.

Attach the updates here (to this PR), so all in one place?

Thanks!

arrmo:
OK, added some updates - found where (and corrected) the display units for noise floor. Also, updated the formatting of wireless frequency, so it doesn't try to format_si GHz to kMHz.

But ... I'm not 100% happy with the second one. I added a check similar to one that is already there. That part is fine, but ... I don't like checking the sensor_type => really, it should be for all wireless sensors, agreed? But I can't figure out how to check that. And I can't just check for a sensor_type of frequency, as then it will hit other items (like Raspberry Pi clock frequency). Thoughts?

Thanks!

PS: Yep, still to fix the units in the noise-floor rrd plots 😄.

arrmo:
OK, found where I *think* the units are set for rrd 😄.
`includes/html/graphs/device/wireless_noise-floor.inc.php
`

Tried to update the units, but I admit - not sure how to force re-generation of the rrd plots.

arrmo:
Never mind! 😆. It's in 2 places, not quite sure why, but fixed both, updated.
```
includes/html/graphs/device/wireless_noise-floor.inc.php
includes/html/graphs/wireless/noise-floor.inc.php
```

I think this is pretty clean now, just not 100% happy with the sensor check (that I noted above). Open to suggestions on a better check.

Thanks!

murrant:

murrant:

PR #11804, opened 2020-06-10
Modifies find_device_id

seros1521:
* Removed host name validation check for RFC. Some devices and OS allow you
  to set a name that contains characters that are prohibited by the RFC.
  Such devices can be added using their IP address, their sysName will be
  written to the database. If the device is already in the database, then it
  should be possible to find it, even if the name is unacceptable according
  to RFC requirements.
* Additional search options for sysName have been added using the default
  domain. This was needed to improve LLDP detection - a short name was used in
  the PDU, and the full name with the domain was stored in the sysName field
  in the database.

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -1363,10 +1363,7 @@ function find_device_id($name = '', $ip = '', $mac_address = '')
     $where = array();
     $params = array();
 
-    if ($name) {
-        $where[] = '`sysName`=?';
-        $params[] = $name;
-
+    if ($name && is_valid_hostname($name)) {
         $where[] = '`hostname`=?';
         $params[] = $name;
 
@@ -1375,13 +1372,6 @@ function find_device_id($name = '', $ip = '', $mac_address = '')
             $params[] = "$name.$mydomain";
 
             $where[] = 'concat(`hostname`, \'.\', ?) =?';
-            $params[] = "$mydomain";
-            $params[] = "$name";
-
-            $where[] = '`sysName`=?';
-            $params[] = "$name.$mydomain";
-
-            $where[] = 'concat(`sysName`, \'.\', ?) =?';
             $params[] = "$mydomain";
             $params[] = "$name";
         }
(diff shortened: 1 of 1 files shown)


includes/discovery/functions.inc.php
```diff
@@ -1363,7 +1363,7 @@ function find_device_id($name = '', $ip = '', $mac_address = '')
     $where = array();
     $params = array();
 
-    if ($name && is_valid_hostname($name)) {
```
murrant, line 1366: You should at least check that it isn't an IP or some other garbage input.

seros1521: The `hostname` field can contain an IP address if the user added a device by specifying an IP address instead of a host name. In this case, the `ip` field in the devices table will be null. The `sysName` field is defined by the device and there can be any value. In my case, sysName contains the symbol _, which is not permitted by the RFC, but is accepted by most operating systems (the person who assigned the names did not know such subtleties). When announcing itself through LLDP, the device sends its sysName, including the _ character, and the LLDP-discovery module cannot find my device in the database, although it is there.
In addition, the `find_device_id` function searches for, but does not change, data in the database. In my opinion, the search function should not present any complex requirements for the search name. If the object in the database has any non-standard name, you still need to be able to find it and work with it. Also, the search takes place on the sysName field, which the user sets at his own discretion.

murrant: worried about false positives?  sysName is not unique you could have multiple devices with the same sysName.

seros1521: It’s appropriate. Although the current implementation of non-unique sysName does not take into account. I suggest the following sequence:
1. verification by hostname - the original version, with verification through is_valid_hostname
2. ip check - original version
3. verification by sysName - verification without is_valid_hostname, with verification of the number of matches. If more than one - do not return anything.
4. mac check - original version

murrant: I would put the mac check before the sysName.

seros1521: Done.

seros1521: @murrant ping

seros1521:
@murrant ping

murrant:

https://community.librenms.org/t/v1-66-release-changelog-july-2020/12953/1

PR #11825, opened 2020-06-15
Make IPv6 discovery work on JunOS

TheMysteriousX:
Although JunOS populates ipAddressIfIndex.ipv6, it does not add prefix lengths or other metadata. This means that IPv6 discovery fails.

This change removes OIDs from the list when there is incomplete data. If the list is empty after IP-MIB is walked, it then tries IPv6 MIB.

It's not a great fix, but I'm not confident I can fix the underlying issue without a complete re-write of this module.

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/ipv6-addresses.inc.php
+++ b/includes/discovery/ipv6-addresses.inc.php
@@ -41,13 +41,18 @@ foreach ($vrfs_lite_cisco as $vrf) {
             $ipv6_prefixlen = explode('.', $ipv6_prefixlen);
             $ipv6_prefixlen = str_replace('"', '', end($ipv6_prefixlen));
 
+            if (Str::contains($ipv6_prefixlen, 'SNMPv2-SMI::zeroDotZero')) {
+                d_echo('Incomplete IPv6 data in IF-MIB');
+                $oids = trim(Str::replaceFirst($data, '', $oids));
+            }
+            
             $ipv6_origin = snmp_get($device, ".1.3.6.1.2.1.4.34.1.6.2.16.$oid", '-Ovq', 'IP-MIB');
 
             discover_process_ipv6($valid, $ifIndex, $ipv6_address, $ipv6_prefixlen, $ipv6_origin, $device['context_name']);
         } //end if
     } //end foreach
 
-    if (!$oids) {
+    if (empty($oids)) {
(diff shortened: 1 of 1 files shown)


includes/discovery/ipv6-addresses.inc.php
```diff
@@ -41,13 +41,18 @@
             $ipv6_prefixlen = explode('.', $ipv6_prefixlen);
             $ipv6_prefixlen = str_replace('"', '', end($ipv6_prefixlen));
 
+            if (Str::contains($ipv6_prefixlen, 'SNMPv2-SMI::zeroDotZero')) {
+                d_echo('Incomplete IPv6 data in IF-MIB');
+                $oids = trim(Str::replaceFirst($data, '', $oids));
```
murrant, line 46: what is the point of this? would a continue work?

TheMysteriousX: Not alone.

What I'm trying to do here is detect the case of 'IP-MIB V6 entries are broken on this device _completely_' vs 'IP-MB doesn't contain entries for _some_ addresses for $reasons', without creating a mess of conditions or storing additional state.

The existing code flips from IP-MIB to IPV6-MIB based on $oids being empty, so this block gradually empties $oids so the existing condition fires.

To use a continue here,  you'd need to store the results of ipv6_discovery_process and compare the number of positive results against the number of line breaks in $oid, then add an additional condition that the alternate code runs if the counts differ.

The whole logic of the discovery process is backwards IMO - IPV6-MIB should be looked at first, but that's an extensive refactor and there isn't the test data available to support it right now.

murrant: I'm also not a fan of all the crazy parsing of the snmpwalk output, could be done better.  I'm probably fine with this, but it is extremely hard to understand what it is doing...

TheMysteriousX: It's just awful all round.

prefixlen at this point on JunOS does not contain a prefix length, it ends up containing the whole string captured by snmpget with the zeroDotZero data type, rather than an 'oid not found', 'no object at this id' or similar error.

zeroDotZero doesn't seem to be well documented, but it seems it should be considered as a 'null' value.

$oids doesn't contain OIDs, it contains new line seperated pairs of IPv6 addresses and ifIndexes.

murrant:

murrant:

PR #11844, opened 2020-06-21
Fixed Ubiquiti Airfiber LTU retrieval

nightcore500:
* Fixed Rx/Tx modulation rate
* Fixed eth0 retrieval
* Fixed quality retrieval
* Fixed Tx/Rx rate
* Fixed distance retrieval
* Added Rx power chain0
* Added Rx power chain1
* Removed eth0 port status
* Removed duplicate Tx EIRP

#### Please note

#### Testers


nightcore500:
Anmerkung 2020-06-23 201249

murrant:
Also, create new test data for your device instead of overwriting the existing test data.

murrant:

https://community.librenms.org/t/ubiquiti-airfiber5xhd/12080/2

nightcore500:
This is the same device. There is no other version of the airfiber5xhd. Are you sure that I should add separate test data?

LibreNMS/OS/AirosAfLtu.php
```diff
@@ -37,7 +37,10 @@ public function discoverWirelessFrequency()
      */
     public function discoverWirelessDistance()
     {
-        $oid = '.1.3.6.1.4.1.41112.1.10.1.4.1.23.24.232.41.30.56.214'; //UBNT-AFLTU-MIB::afLTUStaRemoteDistance.1
+        $oid = snmp_getnext($this->getDevice(), '.1.3.6.1.4.1.41112.1.10.1.4.1.23', '-OnQ'); //UBNT-AFLTU-MIB::afLTUStaRemoteDistance
+        $oid = explode('=', $oid, 2);
+        $oid = trim($oid[0]);
+
         return array(
             new WirelessSensor('distance', $this->getDeviceId(), $oid, 'airos-af-ltu', 1, 'Distance', null, 1, 1000),
```
murrant: You have the value too, so send that to the new WirelessSensor call instead of null

LibreNMS/OS/AirosAfLtu.php
```diff
@@ -51,22 +57,32 @@ public function discoverWirelessDistance()
...
     {
-        $rx_power0_oid = '.1.3.6.1.4.1.41112.1.10.1.4.1.5.24.232.41.30.48.222'; //UBNT-AFLTU-MIB::afLTUStaRxPower0.1
-        $rx_power1_oid = '.1.3.6.1.4.1.41112.1.10.1.4.1.6.24.232.41.30.48.222'; //UBNT-AFLTU-MIB::afLTUStaRxPower1.1
-        $rx_ideal_power0_oid = '.1.3.6.1.4.1.41112.1.10.1.4.1.7.24.232.41.30.48.222'; //UBNT-AFLTU-MIB::afLTUStaIdealRxPower0.1
-        $rx_ideal_power1_oid = '.1.3.6.1.4.1.41112.1.10.1.4.1.8.24.232.41.30.48.222'; //UBNT-AFLTU-MIB::afLTUStaIdealRxPower1.1
-        $rx_power0_level_oid = '.1.3.6.1.4.1.41112.1.10.1.4.1.9.24.232.41.30.48.222'; //UBNT-AFLTU-MIB::afLTUStaRxPowerLevel0.1
-        $rx_power1_level_oid = '.1.3.6.1.4.1.41112.1.10.1.4.1.10.24.232.41.30.48.222'; //UBNT-AFLTU-MIB::afLTUStaRxPowerLevel1.1
+        $oids = array();
+        $sensors = array();
+
```
murrant, line 62: you could do a walk here and then foreach over the data to add the wireless sensors.

murrant:
Looking close, thanks for all the work to get this into shape :)

LibreNMS/OS/AirosAfLtu.php
nightcore500: If I read the whole table with snmpwalk, I would have to check in foreach if the next value is the one I want to read because the table contains some other values that are not useful. I also don't know why ubiquiti uses a table here at all since the airfiber is a pure point to point bridge. So there can only be one single entry in the table. Or do I have a mistake here?

murrant: When using some of the functions we have it will group entries together so you get nice output.  snmpwalk_group(), snmpwalk_cache_oid(), snmpwalk_cache_numerical_oid() are some of these functions.

nightcore500: The problem is that with none of the snmpwalk functions I got the data I needed. Either I'm doing something wrong or it's because of the way ubnt designed the table. For example with snmpwalk_cache_oid I get an array with the index '...).8.' which I can't work with. If I use snmpwalk_cache_oid_num I get an empty array. snmpwalk_group gives me as index the mac address of the counterpart which is also not useful in this case. Do you have an idea what goes wrong with this?

murrant: yeah, that index is why the values are wonky.   try setting the option `-Ob`  (set the full snmp_flags: `-OteQUsb`)

nightcore500: That's it! Thank you for your help :)

nightcore500:
@murrant 
The patch is ready now :)

The travis still fails due to the test data because it uses the old test data and not the current one. Do I have to define the variant somewhere that it is also used for testing or does it not matter in this case?

nightcore500:
> @murrant
> The patch is ready now :)
> 
> The travis still fails due to the test data because it uses the old test data and not the current one. Do I have to define the variant somewhere that it is also used for testing or does it not matter in this case?

@murrant 
What should I do now?

murrant:
Looks like the test json data needs to be updated.

nightcore500:
@murrant 
Ready now

murrant:
Looks good, you can fix the eth0 issue separately.

PR #11854, opened 2020-06-24
Add filter to exclude pending alerts based on alert rule delay

louis-oui:
Add optional filter to only display confirmed alerts based on alert rule delay:
- to alert history dashboard
- to alert history stats dashboard
- to active alerts dashboard
- to alert history page

How it works :

Add a boolean column "pending" on alert_log table with 1 default value. Pending boolean means whether alert rule is pending - delay has not expired.

An anomaly is detected:
- poller.php detects that anamoly matches an alert rules. It creates an alert_log - state 1 / active alert. Pending is 1 regardless the alert rule delay value
- alerts.php (that runs every minutes) checks whether alert rule delay has expired. If so, it sets pending to 0 to alert_log entry

The anomaly is recovered:
- poller.php detects that anamoly is recovered. It creates an alert_log - state 0 / recovered alert. Pending is always 1
- alerts.php checks whether alert rule delay was expired on alert_log and update pending to 0 to alert_log entry if alert rule delay was expired

Dashboards and alert history page can read alert_log delay column to filter out unconfirmed short alerts

image

image

#### Please note

#### Testers


murrant:

https://community.librenms.org/t/add-filter-for-delayed-alerts/4076/5

louis-oui:
@murrant please review this PR

murrant:
Seems like a good idea.  Perhaps it should just be status "pending".

I'd like to have a new js table that can filter/sort any column... but that is outside the scope here :)

louis-oui:
> Seems like a good idea. Perhaps it should just be status "pending".

Good idea. I will change that

murrant:
Basically, in the alerts table if `open` 1 and `alerted` is 0, this is alert pending.  You should be able to combine this with the alert log to filter that as only active alerts would be in the pending state.

I'm not sure you need a new field on the alert log table.  (Tell me why I'm wrong :D)

louis-oui:
> I'm not sure you need a new field on the alert log table. (Tell me why I'm wrong :D)

As far as I know:
- alert table gives info about active alerts 
- alert_log table gives details on alerts and historizes them.

Combinaison of alert and alert_log gives all info about active alerts but not about alert history.

We have an alert history dashboard. We need to check what was the alert. Currently, we are polluted by a lot of unconfirmed alerts.

louis-oui:
@murrant any comment on this PR ?

murrant:
The idea is good, but the implementation still needs some work.
I still don't think and additional database field is needed.
It looks like there are extra queries being made to exclude pending instead of checking that in the original query.

louis-oui:
@murrant The purpose of the PR is to optionaly exclude pending alert from alert history and active alert. To do so, "pending" is added on alert table.

The other options are:

- use alerted and open state in alert table. Result is OK for active alert but we cannot filter out on alert_history. I want also the option for alert_history
- Use details column of alert_log. NOK because I cannot do any SQL query on details column
- Compute time_logged difference between alerted and recovered state and select in SQL query only alerts on which delay has expired. Not sure it is possible in SQL. Anyway, it seems it is too complex. Additionaly, recovery record is optional on alert rule.
- Entry in alert_log only after expiration of the delay. Impossible because we need a record to know at what time the alert started

Do you have any other idea to suggest ? Personally, I am seeing no other options.

FIY, the table desc of alert_log and alert.

```
MariaDB [librenms]> desc alert_log;
+-------------+------------------+------+-----+-------------------+----------------+
| Field       | Type             | Null | Key | Default           | Extra          |
+-------------+------------------+------+-----+-------------------+----------------+
| id          | int(10) unsigned | NO   | PRI | NULL              | auto_increment |
| rule_id     | int(10) unsigned | NO   | MUL | NULL              |                |
| device_id   | int(10) unsigned | NO   | MUL | NULL              |                |
| state       | int(11)          | NO   |     | NULL              |                |
| details     | longblob         | YES  |     | NULL              |                |
| time_logged | timestamp        | NO   | MUL | CURRENT_TIMESTAMP |                |
+-------------+------------------+------+-----+-------------------+----------------+
6 rows in set (0.00 sec)

MariaDB [librenms]> desc alerts;
+-----------+------------------+------+-----+-------------------+----------------+
| Field     | Type             | Null | Key | Default           | Extra          |
+-----------+------------------+------+-----+-------------------+----------------+
| id        | int(10) unsigned | NO   | PRI | NULL              | auto_increment |
| device_id | int(10) unsigned | NO   | MUL | NULL              |                |
| rule_id   | int(10) unsigned | NO   | MUL | NULL              |                |
| state     | int(11)          | NO   |     | NULL              |                |
| alerted   | int(11)          | NO   |     | NULL              |                |
| open      | int(11)          | NO   |     | NULL              |                |
| note      | text             | YES  |     | NULL              |                |
| timestamp | timestamp        | NO   |     | CURRENT_TIMESTAMP |                |
| info      | text             | NO   |     | NULL              |                |
+-----------+------------------+------+-----+-------------------+----------------+
9 rows in set (0.00 sec)
```

murrant:
You seem to ignore the state fields on both tables, is there something I am missing?
Would introducing a new state on this field and logging an additional log entry when it changes from pending work?

Jellyfrog:

PR #11858, opened 2020-06-24
Show port speed on port graphs, optionally set scale

murrant:
Global setting: graphs.port_speed_zoom controls this
You can also set it on an individual graph by setting `port_speed_zoom=1` in the url

LINE2:
image

shows scale, downside is tiny amounts of data on a fast port won't even register visually

HRULE:

image

Allows zoomed in graphs, downside is you can't really see the line unless you are very close to the limit.  It is still in the legend though.

**Any color suggestions or leave it black?**

Dump ifHighSpeed db field it is useless.

#### Please note

#### Testers


Jellyfrog:
Maybe an option to turn it on/off and color?

peelman:
it’s a useful tool, but should definitely be optional.

pobradovic08:
+1 for making it optional. There are cases where you have for couple of 10G interfaces in a port-channel with bunch of subinterfaces (each terminating a customer connection for example). Some of them might have just few mbps but the interface speed is still reported as N x 10G.
image

murrant:
Added global setting and per graph setting.  you can set `port_speed_zoom=1` on any graph to scale it to the ifSpeed.

murrant:
should have been two PRs :/

pgnuta:
Hi All,  Firstly i want to say thanks to everyone for your ongoing hard work on this project but i have a slight beef with this feature...  This feature has been badly implemented in some ways.  It should have been made clear in the release notes and it should not have been enabled by default.  It's been a bit of a confusing time tracking down why a bunch of my VMware high speed interfaces were showing no apparent traffic.

aldemaroc:
Is it possible to disable this feature globally? We have some 10G interface with just a few Mbps and now our dashboards are useless. Also, "Zoom to Traffic" link is broken, as it appends `/%3Fport_speed_zoom=0/` to the URL instead of `/port_speed_zoom=0/` as per [includes/html/pages/graphs.inc.php#L158](https://github.com/librenms/librenms/blob/master/includes/html/pages/graphs.inc.php#L158).

pgnuta:
You can set this in your config.php

'''
$config['graphs.port_speed_zoom'] = false;
'''

murrant:
All, this has been set by default initially to get some feedback, it will likely be set to false before the release.

Thanks for all that given useful feedback.  Please switch to the monthly release if you want.  If you like to be able to give feedback on these things, please stay on the daily.

bakerds:
I think you've got a bug in line 158 of includes/html/pages/graphs.inc.php:
`        if ($vars['port_speed_zoom'] ?? Config::get('graphs.port_speed_zoom')) {
            echo generate_link('Zoom to Traffic', $vars, ['page' => 'graphs', '**?**port_speed_zoom' => 0]);
        } else {
            echo generate_link('Zoom to Port Speed', $vars, ['page' => 'graphs', 'port_speed_zoom' => 1]);
        }`

Including the ? seems to break the ability to toggle back and forth.

bakerds:
When "use stacked graphs" is turned off, should this draw a scale line on the negative Y axis as well?

murrant:
@bakerds I saw that and probably, but it makes the graph even smaller and usually traffic is asymmetric. I have no idea how to solve that and am open to suggestions.

Also, sent PR to change the default. Thanks for feedback all.

bakerds:
Maybe break it into two settings — one for input and one for output — and have a toggle for each on the graph display?

richardlawley:
Some feedback on this - removal of the ifHighSpeed caused any API calls which requested it to break in an unfriendly way:

`GET /api/v0/devices/XX/ports?columns=port_id,ifHighSpeed` would previously return the normal JSON output - since this PR it returns "1" (i.e. invalid JSON) while still giving a 200 status.  I suspect this problem (invalid ports requested) was there all along, but taking a field was previously valid has caused this to crop up (also mentioned by someone else at https://community.librenms.org/t/ifhighspeed-no-longer-on-api-call/16719).

Additionally, the [documentation page for ports](https://docs.librenms.org/API/Ports/) still lists it.

Took me a while to find this since it's just lumped in with what looks like an unrelated PR.

murrant:
@richardlawley a notification within LibreNMS was issued.  Thanks for the heads up on the docs.  Probably should have been two separate PRs.

richardlawley:
> a notification within LibreNMS was issued.

I've not seen such a notification, and it's not present in https://www.librenms.org/notifications.rss

murrant:
https://github.com/librenms/librenms/blob/master/misc/notifications.rss

christianseitz:
It's a great feature and it's good that it can be switches on and off globally.

The graph only seems to zoom to the port speed for incoming packets and not for both incoming and outgoing packets. I would have expected the zero line to be in the middle of the graph.

PR #11896, opened 2020-07-04
Support multiple daily process locking backends with distributed polling

murrant:
Using Laravel locks allows us to use redis or memcached locks for daily.sh in conjunction with distributed polling.

Uses database for locking by default.  To use redis or memcached, set this in your .env file:
`CACHE_DRIVER=redis`
or
`CACHE_DRIVER=memcached`

#### Please note

#### Testers


includes/sql-schema/update.php
```diff
@@ -108,10 +96,5 @@
         echo Artisan::output();
     }
 
-    if (isset($schemaLock)) {
-        $schemaLock->release();
-    }
-} catch (LockException $e) {
-    echo $e->getMessage() . PHP_EOL;
-    $return = 1;
+    $schemaLock->release();
```
Jellyfrog, line 99: Wouldn't this one be released even if skip =true?

alerts.php
```diff
@@ -27,33 +27,34 @@
  * @author: Heath Barnhart <hbarnhart@kanren.net>
  */
 
-use LibreNMS\Util\FileLock;
 use LibreNMS\Alert\RunAlerts;
+use LibreNMS\Util\FileLock;
```
Jellyfrog, line 31: Typo?

includes/sql-schema/update.php
murrant: You can't release a lock you don't own unless you forceRelease()

alerts.php
murrant: Legacy, I was looking at refactoring the existing locks, but went this route instead.  Thanks, removed this and one other reference.

Jellyfrog:
If I understand this right, you're implementing Laravels lock feature for the `file`-driver ?  
Couldn't we utilize Laravel 7's support for `database` instead? Seems like a fair trade-of if the user dosn't have redis/memcache?  https://laravel.com/docs/7.x/cache#atomic-locks

Jellyfrog:
Other than that, it looks nice

murrant:
@Jellyfrog haha, didn't see they added a database driver for 7.  That would be a good compromise.  I was worried about what to do with distributed poller setups as we have to set the cache driver to something that supports locking by default.

murrant:
@librenms/reviewers Ready for review/merge

app/Providers/AppServiceProvider.php
```diff
@@ -156,4 +159,17 @@ private function bootCustomValidators()
...
     }
+
+    protected function extendFileCacheDriver(): void
+    {
+        Cache::extend('file', function ($app) {
+            return Cache::repository(
+                new LockingFileStore(
+                    $app['files'],
+                    $app['config']->get('cache.stores.file.path'),
+                    $app['config']->get('cache.stores.file.permissions')
```
Jellyfrog, line 170: Doesn't exist?

Jellyfrog:
We're keeping the custom file locking but not using it if I'm reading this correct? Else we're missing docs

murrant:
I wasn't sure about the custom file locks as it is no longer using it.  I think I can rip it out.
The only reason I left it in was someone might have file set for cache, probably not a good reason and the code will sit mostly unused and bitrot.

alerts.php
```diff
@@ -28,32 +28,30 @@
...
-    echo 'ClearStaleAlerts():' . PHP_EOL;
-    $alerts->clearStaleAlerts();
-    echo "RunFollowUp():\r\n";
-    $alerts->runFollowUp();
-    echo "RunAlerts():\r\n";
-    $alerts->runAlerts();
-    echo "RunAcks():\r\n";
-    $alerts->runAcks();
-    echo 'End  : ' . date('r') . "\r\n";
+$alerts_lock = Cache::lock('alerts');
```
Jellyfrog, line 41: Missing use Cache?

daily.php
```diff
@@ -11,8 +11,6 @@
 use Illuminate\Database\Eloquent\Collection;
 use LibreNMS\Alert\AlertDB;
 use LibreNMS\Config;
-use LibreNMS\Exceptions\LockException;
-use LibreNMS\Util\MemcacheLock;
```
Jellyfrog, line 15: Missing use Cache?

alerts.php
murrant: hehe, the ide doesn't like the use statement...  leaving as-is.

murrant:
@librenms/reviewers This seems ready to merge to me, any objections?

hanserasmus:
I have tested both SQL locks and Redis locks on a CentOS 8 distributed system. One node running php 7.3 from Appstream repo and the other php74 from Remi. It worked well. No errors. Great job!

f0o:
LGTM 😅

murrant:

murrant:

https://community.librenms.org/t/mariadb-galera-cluster-general-error-1205-lock-wait-timeout-exceeded/14201/2

PR #11929, opened 2020-07-13
Add OSPF cost (TOS)

haydenseitz:
…ost)

#### Please note

#### Testers


willhseitz:
I'm not sure how to update the DB schema tests. I followed the suggestion from Travis CI: 
`If you have changed the schema, make sure you update it with ./scripts/build-schema.php`
but it looks like there were other changes beside my own

willhseitz:
also, I'm curious on anyone's thoughts for including port_id in the ospf_tos table to directly relate ospf_tos entries to the port, or to leave as-is with the ospf_port_id and only relate to entries in the ospf_ports table

Jellyfrog:
You probably need to manually rearrange the fields to match what travis is saying. Sorry for the inconvenience

willhseitz:
@Jellyfrog ready for review again

includes/polling/ospf.inc.php
```diff
@@ -138,6 +139,35 @@
...
+            ->value('ospf_port_id');
+        $tos = OspfTos::updateOrCreate([
+            'device_id' => $device['device_id'],
+            'ospf_port_id' => $ospf_tos['ospf_port_id'],
+            'context_name' => $device['context_name'],
+        ], $ospf_tos);
+
+        $ospf_tos_metrics->push($tos);
+    }
+
```
murrant, line 164: Why aren't you graphing these?

Jellyfrog:
Could you also generate test files for this?

includes/polling/ospf.inc.php
willhseitz: I tried copying the RRD code for OSPF TOS, but my rrdcached was giving an error for unknown arguments. Can you add additional datasets to an RRD file after it's created?

murrant: No, you cannot modify the datasets in an RRD.  You would need a new rrd file just for this.

willhseitz: Right. Route costing is set in the router config, so something like config versioning would let someone tracking cost over time. I don't think we need a separate RRD graph for it.

willhseitz:
> 
> 
> Could you also generate test files for this?

@Jellyfrog I went to add more data to unit tests, but I don't see any OSPF data when running `collect-snmp-data.php`. Is that because there is no discovery module for OSPF? There is also no definition for ospf in `tests/module_tables.yaml`.

Here is my output when trying to save ospf data:
```
$ php scripts/collect-snmp-data.php -h 9 -m ospf -f /tmp/ospf-snmp-data.snmprec
OS: iosxr
Module(s): ospf

Capturing Data:  sysDescr.0 sysObjectID.0 sysName.0 sysUpTime.0 sysContact.0 snmpEngineTime.0 hrSystemUptime.0

Updated snmprec data /tmp/ospf-snmp-data.snmprec

Verify this file does not contain any private data before submitting!
```

Jellyfrog:
> > Could you also generate test files for this?
> 
> @Jellyfrog I went to add more data to unit tests, but I don't see any OSPF data when running `collect-snmp-data.php`. Is that because there is no discovery module for OSPF? There is also no definition for ospf in `tests/module_tables.yaml`.
> 
> Here is my output when trying to save ospf data:
> 
> ```
> $ php scripts/collect-snmp-data.php -h 9 -m ospf -f /tmp/ospf-snmp-data.snmprec
> OS: iosxr
> Module(s): ospf
> 
> Capturing Data:  sysDescr.0 sysObjectID.0 sysName.0 sysUpTime.0 sysContact.0 snmpEngineTime.0 hrSystemUptime.0
> 
> Updated snmprec data /tmp/ospf-snmp-data.snmprec
> 
> Verify this file does not contain any private data before submitting!
> ```

This is really a question for @murrant

includes/polling/ospf.inc.php
murrant: sounds good

murrant:
It is because the module is not defined in tests/module_tables.yaml

murrant:
A good start at #12215 feel free to pull that into this PR.

willhseitz:
how do I include a PR in another PR? manually apply the diff?

willhseitz:
> 
> 
> @murrant how do I include a PR in another PR? manually apply the diff?

I did `git pull https://github.com/murrant/librenms.git ospf-capture` and applied those changes to this PR

willhseitz:
Travis failed because of the new OSPF data. shouldn't defined it in the iosxr.json be correct?

Jellyfrog:
Well It fails because of
```sql
SQLSTATE[HY000]: General error: 1364 Field 'ospfAuthType' doesn't have a default value (SQL: insert into `ospf_areas` (`device_id`, `ospfAreaId`, `context_name`, `ospfImportAsExtern`, `ospfSpfRuns`, `ospfAreaBdrRtrCount`, `ospfAsBdrRtrCount`, `ospfAreaLsaCount`, `ospfAreaLsaCksumSum`, `ospfAreaSummary`, `ospfAreaStatus`) values (453, 0.0.0.0, ?, importExternal, 1974, 3, 154, 705, 22790689, sendAreaSummary, active))
```

ospfAuthType is not being inserted and it doesn't have a default value in the database, so maybe allow it go be null or make sure it's always populated?

willhseitz:
@Jellyfrog helps to look at it with fresh eyes 😆. Looks like that error is for `ospf_areas` table.

@murrant I think this stems from adding ospf to testing - I would bet this issue happens on regular installs too, but gets ignored by mysql. The devices i'm testing with don't report an ospfAuthType field, or anything similar. Should the auth fields in each ospf table be changed to nullable?

murrant:
@willhseitz all tests run with SQL strict mode.  Yes it is more forgiving in production.
We can ether change them to nullable or set a default value on the model of ''. I probably lean slightly towards nullable.

willhseitz:
@murrant nullable sounds good. will this require a database migration for the existing ospf tables?

murrant:
Yes it does require a new migration, only make nullable the columns you think are needed.

murrant:
Yes, it is the new ospf testing, it hasn't been "tested"... I'll leave now.

willhseitz:
@murrant I'll start a new PR for the OSPF nullable migration

willhseitz:
@murrant created #12249 and pulled it into here (is that right?). not getting the error on ospfAuthType anymore. I'll keep adding updated test units for ospf data

murrant:
It won't be pulled into your local branch, but Travis testing will pull it in.  You'll need to rebase or merge this branch against master.

willhseitz:
re-merged from master, ready to check

murrant:
This is looking good on the operation, but I noticed the ospf_tos table is not captured by testing, can you add it to tests/module_tables.yaml and update the test data?

willhseitz:
I figured just because it was new/different information, but I think you're right about the 1-1 to ospf_ports. I'll see about adding columns to ospf_ports instead of creating ospf_tos

database/migrations/2020_10_30_093601_add_tos_to_ospf_ports.php
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('ospf_ports', function (Blueprint $table) {
+            $table->dropColumn('ospfIfMetricIpAddress');
```
murrant, line 33: This needs to be one dropColumn call with an array of the column names :)

murrant (changes requested):

murrant (approved):
Thanks for putting in the work

murrant:

PR #11941, opened 2020-07-18
Added option to always update an app metric value and _prev

TheGreatDoc:
Added an option to always update an application metric. This is usefull for mysql slow queries alerting.

The metric name must be the internal one, so for MySQL slow queries, it is `SQs`

NOTE: I made a new one as I had very bad experiences rebasing. At some point I should learn how to correctly do them but, meanwhile...

#### Please note

#### Testers

### DIFF
--- a/includes/polling/functions.inc.php
+++ b/includes/polling/functions.inc.php
@@ -351,7 +351,7 @@ function poll_device($device, $force_module = false)
                 echo "Module [ $module ] disabled globally.\n\n";
             }
         }
-        
+
         if (!$force_module && !empty($graphs)) {
             echo "Enabling graphs: ";
             $graphs = collect($graphs)->keys();
@@ -631,7 +631,7 @@ function update_application($app, $response, $metrics = array(), $status = '')
                     'application_metrics'
                 );
                 echo '+';
-            } elseif ($value != $db_metrics[$metric_name]['value']) {
+            } elseif ($value != $db_metrics[$metric_name]['value'] || Str::is(Config::get('applications.metric.always_update'), $metric_name)) {
                 dbUpdate(
                     array(
                         'value' => $value,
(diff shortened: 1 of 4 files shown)


includes/polling/functions.inc.php
```diff
@@ -631,7 +631,7 @@ function ($carry, $metric_group) use ($metrics) {
                     'application_metrics'
                 );
                 echo '+';
-            } elseif ($value != $db_metrics[$metric_name]['value']) {
+            } elseif ($value != $db_metrics[$metric_name]['value'] || Str::is(Config::get('applications.metric.always_update'), $metric_name)) {
```
Jellyfrog, line 634: Shouldn't it be in_array?

TheGreatDoc: It works with `Str::is` (I've tested with 1 and 2 elements in the array)

TheGreatDoc:
> You should just add _pref to the specific fields you want in the mysql app. instead of blanket enabling it everywhere.

People may want to have this in other metrics (including me) or dont want it anywhere and want to work with those metrics like now.

I made it an option. Maybe I could create and *Advanced* subsection or something for those kind of specific configurations

murrant:
@TheGreatDoc I don't think you are understanding what I am saying.

I mean it should be up to the developer of the application to decide which values if any have a _prev value.

There should be no need for the user to make a choice.

TheGreatDoc:
But users monitor their Network/servers in their way. I want to let them do It as they like.

And to be honest, that will require more code from my side :p 

Its ok if you dont like and want to close the PR, I will find a way to alert on those metrics.

murrant:
_prev is fine, but it is not a one-size fits all solution.  (Me dreaming, it would be great to alert of rrd/datastore data)

murrant:
Closing this for now.  Again, the application code is 100% capable of adding _prev fields on a per-field basis.  Adding them blanket to all, is not what we want I think.

PR #11954, opened 2020-07-21
Revert "Show SysName in Availability Widget"

kkrumm1:
Reverts librenms/librenms#11953

seems to break dashboard


SourceDoctor:
i assume its a base_url configuration issue ... not this here ...
had this "no dashboard" issue also and solved it in replacing "http:/myhost/" with "/"
@murrant did a fix this day ... maybe this causes  the issue

kkrumm1:
you are right sir :)

PR #11966, opened 2020-07-28
Added new authalgo support for SNMPv3

hanserasmus:
Some manufacturers are moving towards SNMPv3 where SHA-224/256/384/512 is the minimum allowed. These devices cannot be supported in LibreNMS at the moment. Trying to add support for it.

### Caveats

There are 2 conditions that need to be met for these capabilities to work (both need to be met):
- net-snmp 5.8 minimum
- openssl 1.1 minimum

It seems that these two packages are the standard installs on ~~**Ubuntu 18.04**~~ **Ubuntu 20.04** and**CentOS8**.

**EDIT**: Ubuntu 18.04.4 with standard repositories only supports up to snmp 5.7.3. However, in the install documentation Ubuntu 20.04 is the Ubuntu version of choice and according to [this](https://packages.ubuntu.com/search?keywords=snmp) site 5.8 is the default for Ubuntu 20.04.

Maybe a test when the user clicks on any of the new settings to verify that these two packages are on the correct versions? 

#### Please note

#### Testers


hanserasmus:
Thank you to @SourceDoctor for guidance and the DB migrations file.

I know this PR will have lots of discussion whether this is a good idea or not, and I respect the decision from the maintainers fully. I want to add that if it seems that this PR will be accepted, I will write documentation to explain the minimum requirements for this to work as well. I am fully aware that this might create confusion with users, but to me it seems it is just a matter of time before this requirement will become minimum standards.

Jellyfrog:
@librenms/reviewers what you say, as we don't \*require* minimum version for net-snmp/openssl,   
should we just add a notice to `addhosts`-page about minimum required for these new algos?

SourceDoctor:
is there an elegant way to check if requirements are resolved, and if not to hide those additional option?
didn't find one on a short look

hanserasmus:
Please see **Edit** in first post.

murrant:

https://community.librenms.org/t/aes256-support-for-snmp/11085/4

hanserasmus:
I need to first also add support for updated crypto algorithms AES-192 and AES-256. Will do it ASAP.

includes/functions.php
```diff
@@ -862,6 +862,33 @@ function snmp2ipv6($ipv6_snmp)
...
+/**
+ * checks if System is SNMPv3 SHA2 Capable for Auth Algorithms (SHA-224,SHA-256,SHA-384,SHA-512)
+ * @return boolean
+ */
+function snmpv3_sha2_capable()
+{
+    $netsnmp_minimum = '5.8.0';
+    $openssl_minimum = '1.1.0';
+
+    $netsnmp_ver = explode(' ', str_replace('version: ', '', rtrim(shell_exec(
```
Jellyfrog, line 875: Cant you use `version_info()` instead of running the same command again?

includes/html/pages/addhost.inc.php
```diff
@@ -241,6 +241,12 @@
               <select name="authalgo" id="authalgo" class="form-control input-sm">
                 <option value="MD5" selected>MD5</option>
                 <option value="SHA">SHA</option>
+                <?php if (snmpv3_sha2_capable()) {?>
```
Jellyfrog, line 244: I'm no fan of hiding a feature because a "hidden" dependency requirement is not met.  
Maybe set them as disabled and  print info text if the requirement is not met?

misc/config_definitions.json
```diff
@@ -4948,7 +4955,10 @@
                 }
             ],
             "validate": {
-                "value.*.authalgo": "in:MD5,SHA"
+                "value.*.authalgo": "in:MD5,SHA,SHA-224,SHA-256,SHA-384,SHA-512"
+            }
+            "validate": {
```
Jellyfrog, line 4960: woopsie

includes/html/pages/addhost.inc.php
```diff
@@ -254,8 +260,12 @@
             <label for="cryptoalgo" class="col-sm-3 control-label">Crypto Algorithm</label>
             <div class="col-sm-9">
               <select name="cryptoalgo" id="cryptoalgo" class="form-control input-sm">
-                <option value="AES" selected>AES</option>
-                <option value="DES">DES</option>
+                <option value="DES" selected>DES</option>
```
Jellyfrog, line 263: intended change setting DES as default?

hanserasmus: OCD about how it will look. AES will be the one with "more options"? If that makes sense?

hanserasmus:
All credit for commit [630f9f9](https://github.com/librenms/librenms/pull/11966/commits/630f9f9706e0f7980107eb8bbcbedec75fb77701) goes to @SourceDoctor . Thank you for this!

misc/config_definitions.json
Jellyfrog: @hanserasmus isnt this wrong? dual `validate`

hanserasmus: Don't think so?. We need to test for both SHA2 capabalities, as well as new crypto-algorithm capabilities.

murrant: You need to put them both in the same validate (otherwise you only get one as it overwrites the other key)

includes/common.php
```diff
@@ -821,10 +821,37 @@ function version_info($remote = false)
...
+    )))[1];
 
     return $output;
 }//end version_info()
 
+/**
+ * checks if System is SNMPv3 SHA2 Capable for Auth Algorithms (SHA-224,SHA-256,SHA-384,SHA-512)
+ * @return boolean
+ */
+function snmpv3_sha2_capable()
```
murrant, line 835: This doesn't feel like the right way to do this.  Is there a possibility to have other SSL libraries than openssl?
Also, this is not a lightweight function and you call it many times.
Your version compare is incorrect and it will cast the strings to floats.  you need to use version_compare()

hanserasmus: @murrant from [here](https://fossies.org/diffs/net-snmp/5.7.3_vs_5.8/configure-diff.html) it seems that test for OpenSSL themselves? But I am not 1000% sure.

image

Will fix the validation.

hanserasmus:
Thanks again @SourceDoctor for [4e0b371](https://github.com/librenms/librenms/pull/11966/commits/4e0b37144c4fdae2e8d103cc240c031ee682dd3d)

hanserasmus:
Thanks @SourceDoctor for [430cca](https://github.com/librenms/librenms/pull/11966/commits/430cca71d27071a04df0aa956022984d85094c82)

database/migrations/2020_07_27_00522_alter_authalgo_column.php
```diff
@@ -0,0 +1,33 @@
...
+        });
+    }
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('devices', function (Blueprint $table) {
```
murrant, line 28: you can just leave down empty.

includes/common.php
```diff
@@ -81,7 +82,7 @@ function graylog_severity_label($severity)
  * @param array $command
  * @return null|string
  */
-function external_exec($command)
+function external_exec($command, $device)
```
murrant, line 85: You can avoid this by using `DeviceCache::primary()` in the log event call below (if there is no primary device, it will send null, which is ok)

murrant: Also, adding an argument without a default value means you need to update all the calls to that function to include the new argument.

includes/common.php
murrant: $ snmpget --help 2>&1 |grep PROTOCOL
  -a PROTOCOL           set authentication protocol (MD5|SHA|SHA-224|SHA-256|SHA-384|SHA-512)
  -x PROTOCOL           set privacy protocol (DES|AES)

Let net-snmp do the checks, and that way we don't need to call openssl at all.  Just parse the output.

hanserasmus:
Having some trouble getting to my dev environment due to appliance failures. Will get back to applying changes to this code as soon as I can.

includes/html/pages/addhost.inc.php
```diff
@@ -241,7 +241,14 @@
               <select name="authalgo" id="authalgo" class="form-control input-sm">
                 <option value="MD5" selected>MD5</option>
                 <option value="SHA">SHA</option>
+                <option value="SHA-224"<?=snmpv3_sha2_capable()?:" disabled"?>>SHA-224</option>
```
murrant, line 244: This function calls a shell script (slow) would be better to store it in a temporary variable and just access that instead of calling it 4 times.

includes/html/pages/device/edit/snmp.inc.php
```diff
@@ -359,7 +359,16 @@
     <select id='authalgo' name='authalgo' class='form-control'>
     <option value='MD5'>MD5</option>
     <option value='SHA' ".($device['authalgo'] === 'SHA' ? 'selected' : '').">SHA</option>
+    <option value='SHA-224' ".($device['authalgo'] === 'SHA-224' ? 'selected' : '').(snmpv3_sha2_capable()?:" disabled").">SHA-224</option>
```
murrant, line 362: use temp variable here too.

murrant:
Database migration is missing, did you forget to add it or still need to create it?

includes/html/pages/device/edit/snmp.inc.php
```diff
@@ -358,11 +358,15 @@
     <div class='col-sm-4'>
     <select id='authalgo' name='authalgo' class='form-control'>
     <option value='MD5'>MD5</option>
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
hanserasmus: I had already fixed it, and resubmitted? The last submission I did addressed the conflicts where manual intervention was required.

murrant (approved):

murrant:

PR #11975, opened 2020-07-29
Add type hinting to devicecache

Jellyfrog:
Prevents sql query from executing as string.

Before:
```select * from `devices` where `device_id` = '2' limit 1```
After:
```select * from `devices` where `device_id` = 2 limit 1```

As a bonus it also prevents sql queries with `device_id = null`, small performance win.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Cache/Device.php
+++ b/LibreNMS/Cache/Device.php
@@ -46,7 +46,7 @@ class Device
      *
      * @param int $device_id
      */
-    public function setPrimary($device_id)
+    public function setPrimary(int $device_id)
     {
         $this->primary = $device_id;
     }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: LibreNMS/Cache/Device.php [hunk 1]
```diff
@@ -46,7 +46,7 @@ public function getPrimary() : \App\Models\Device
      *
      * @param int $device_id
      */
-    public function setPrimary($device_id)
+    public function setPrimary(int $device_id)
```
murrant, line 49: Shouldn't this one be ?int too?

Jellyfrog: Not really since it makes no sense it being null i suppose?

Jellyfrog: Also, we don't want this to be null? The other function is a feature. Here I would kinda call it a bug? Better to be strict then

murrant: Oopsie, got that mixed up with something else :)

murrant:

murrant:

https://community.librenms.org/t/ajax-table-php-throwing-error/12994/7

PR #11977, opened 2020-07-29
Add default values to cast

Jellyfrog:
Makes the SQL query correctly select id columns as int instead of string for most ajax table requests

Not a performance issue I think, but if it was the other way (string to int) we could missed indexed searches.

#### Please note

#### Testers


murrant:
@Jellyfrog what is the benefit of this?

Jellyfrog:
Because the sql query ends up treating the integers as Strings, which in mysql seems to not have any effect tho. But might in orders.   
And it it was a indexed column that was a string "1" for example, and you searched for `column=1` (integer) it would have missed the indexed search.  
Might be a problem in other databases, not sure. 
Tldr; to treat integers as integers

murrant:

PR #11994, opened 2020-08-03
Fixes missing items for Eaton M2 management card

ajsiersema:
System overview items like Hardware, Version and Serial and sensors like Temperature and Humidity sensors where missing for UPSes equipped with the Eaton M2 management card.
The M2 is RFC 1628 compatible and doesn't support MG-SNMP-UPS-MIB/upsmg.
Use XUPS-MIBS instead.

Tested on the:
9PX 5000i
9PX 6000i
5P 1150
5PX 1500
(v1.7.5)

Also see:
https://community.librenms.org/t/force-sw-hw-information-from-another-mib/11417

Please review.
I'd lie to add "rfc1628_compat: true"., which should reduce eatonups specific code, and also provide items like "Runtime" (Runtime remaining) and "Load" (device_load). Adding it together with the current eatonups code will make some items appear twice (e.g. Frequency/Bypass).

If we can rely on sysObjectID's .1.3.6.1.4.1.534.1|2 to _only_ match M2s, the rfc1628 code could replace some (most?) eatonups specific code, like e.g. includes/discovery/sensors/charge/eatonups.inc.php
(maybe the rfc1628 code was added after the eatonups code ?)

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/discovery/eatonups.yaml
+++ b/includes/definitions/discovery/eatonups.yaml
@@ -1,0 +1,19 @@
+mib: EATON-OIDS:EATON-EMP-MIB:XUPS-MIB
+modules:
+    sensors:
+        temperature:
+            data:
+                -
+                    oid: xupsEnvironment
+                    value: xupsEnvRemoteTemp
+                    num_oid: '.1.3.6.1.4.1.534.1.6.5.{{ $index }}'
+                    descr: 'Environment'
+                    index: 'xupsEnvRemoteTemp.{{ $index }}'
+        humidity:
+            data:
+                -
+                    oid: xupsEnvironment
+                    value: xupsEnvRemoteHumidity
+                    num_oid: '.1.3.6.1.4.1.534.1.6.6.{{ $index }}'
+                    descr: 'Environment'
+                    index: 'xupsEnvRemoteHumidity.{{ $index }}'
--- a/includes/definitions/eatonups.yaml
+++ b/includes/definitions/eatonups.yaml
@@ -8,6 +8,7 @@ over:
     - { graph: device_frequency, text: Frequencies }
 mib_dir:
     - ups
+    - mge
 discovery:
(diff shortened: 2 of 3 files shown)


includes/definitions/discovery/eatonups.yaml
```diff
@@ -0,0 +1,19 @@
+mib: EATON-OIDS:EATON-EMP-MIB:XUPS-MIB
```
murrant, line 1: only the XUPS-MIB is used here.

includes/definitions/eatonups.yaml
```diff
@@ -8,6 +8,7 @@ over:
     - { graph: device_frequency, text: Frequencies }
 mib_dir:
     - ups
+    - mge
```
murrant, line 11: Why did you add the mge mib directory, that is used in a different OS.

includes/polling/os/eatonups.inc.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+$hardware = trim(snmp_get($device, 'upsIdentModel.0', '-OQv', 'UPS-MIB'), '" ');
```
murrant, line 3: trim() is not needed, this should code should be implemented in the new style.  See LibreNMS\OS\Qnap.php

dagbdagb:
See #11609 for where I got sidetracked. Plus an explanation for the madness branded as Eaton.

ajsiersema:
> See #11609 for where I got sidetracked. Plus an explanation for the madness branded as Eaton.

I honestly searched the git repo first, I can't imagine how I have missed this one.

I think we're both trying to achieve the same, can we merge our efforts, how can I help ?

Jellyfrog:
Thank you for your contribution but we're closing this due to inactivity.  
Feel free to reopen if you plan on finishing it!

PR #11995, opened 2020-08-03
Increase default RRD data retention of MIN, MAX, and LAST to match AVERAGE

pobradovic08:
Changed the default RRD RRA definition so that `MIN`, `MAX` and `LAST` CDPs have the same values for row count as `AVERAGE`.
Currently `AVERAGE` has the following definitions:

| PDP/CDP | Rows | Data retention |
|:-----------|:------|:-----------------|
| 1 | 2016 | 7 days |
| 6 | 1440 | 30 days |
| 24 | 1440 | 120 days |
| 288 | 1440 | ~4 years |

While `MIN` and `MAX` have:

| PDP/CDP | Rows | Data retention |
|:-----------|:------|:-----------------|
| 1 | **720** | **60h** |
| 6 | 1440 | 30 days |
| 24 | **775** | **~65 days** |
| 288 | **797** | **2.2 years** |

Which doesn't make much sense. This is most visible when graphing something like temperature that uses `MIN`, `AVERAGE`, and `MAX` on the same graph, resulting in mixed resolution data sets and missing `MIN` and `MAX` for old data like in the image below:
image

I changed the `MIN`, `MAX` to be the same as `AVERAGE` so the newly created graphs don't have this issue.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Data/Store/Rrd.php
+++ b/LibreNMS/Data/Store/Rrd.php
@@ -57,9 +57,9 @@ class Rrd extends BaseDatastore
         $this->rra = Config::get(
             'rrd_rra',
             'RRA:AVERAGE:0.5:1:2016 RRA:AVERAGE:0.5:6:1440 RRA:AVERAGE:0.5:24:1440 RRA:AVERAGE:0.5:288:1440 ' .
-            ' RRA:MIN:0.5:1:720 RRA:MIN:0.5:6:1440     RRA:MIN:0.5:24:775     RRA:MIN:0.5:288:797 ' .
-            ' RRA:MAX:0.5:1:720 RRA:MAX:0.5:6:1440     RRA:MAX:0.5:24:775     RRA:MAX:0.5:288:797 ' .
-            ' RRA:LAST:0.5:1:1440 '
+            ' RRA:MIN:0.5:1:2016 RRA:MIN:0.5:6:1440     RRA:MIN:0.5:24:1440     RRA:MIN:0.5:288:1440 ' .
+            ' RRA:MAX:0.5:1:2016 RRA:MAX:0.5:6:1440     RRA:MAX:0.5:24:1440     RRA:MAX:0.5:288:1440 ' .
+            ' RRA:LAST:0.5:1:2016 '
         );
         $this->version = Config::get('rrdtool_version', '1.4');
     }
--- a/misc/config_definitions.json
+++ b/misc/config_definitions.json
@@ -4550,7 +4550,7 @@
             "type": "integer"
         },
         "rrd_rra": {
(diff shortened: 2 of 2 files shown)


murrant:
I don't know much about RRA, can you show a fixed graph?

pobradovic08:
Because this would apply only to new graphs i can't use previously collected data to show the difference. I can maybe show some differences in a day or so after new data has been collected , but for all of the stuff i mentioned would take a lot longer.

This changes how many points (rows) we store for each aggregation type. Number of rows is number of these CDPs (Consolidated Data Points) kept, which defines the overall timeframe we can store the data for for specific resolution. For example:

With 5 min step interval, for `AVERAGE` we store 2016 PDP (Precision Data Point) values covering 7 days (2016*5min) and these are not aggregated. We also have additional three CDPs that are made out of:
- 6 PDPs, each representing 30 min average, total of 30 days (1440 * 30 min)
- 24 PDPs, each representing 2h average, total of 120 days (1440 * 2h)
- 288 PDPs, each representing 1 day average, total of ~4 years (1440 * 1 day)

This means that if we graph the interval up to last 7 days we will use highest resolution data (5min samples), if we graph up to last 30 days we will use 30 min samples, and so on...

But for `MIN` and `MAX` the higest resolution data is stored only for 720 * 5min = 60 hours (not 7 days) so if we for example graph last 5 days we would use 5 min data for DS using `AVERAGE` (because there is data for 7 days), and 30 min data for `MIN` and `MAX` as we don't have highest resolution data anymore after 60h, only aggregated one. I see no reason to have different number of rows.

Hope I explained it OK.

SourceDoctor:
@pobradovic08 
is a convertion for existings RRDs possible?

pobradovic08:
@SourceDoctor I don't think there's an easy way. There is a [rrdtool resize](https://oss.oetiker.ch/rrdtool/doc/rrdresize.en.html) that can alter the number of rows but it doesn't do it inline, it creates the new `resize.rrd` file. Also the number of rows to add/remove is relative to the existing ones so we would need to know the current number of rows, as they might be different for custom settings.

I will try to make a script similar to `scripts/rrdstep.php` as I want to modify RRDs on my production server anyway, but I don't know when I will have the time to look into that.

@murrant Here are two graphs, first created with old RRD def, second one with changed:
sla-before
sla-after
You can see on the first one that min and max lines (gray) are lower resolution compared to average (red).

murrant:
I don't think conversion of existing RRDs would be a good idea.  It will be a little confusing as to why one graph has better resolution as compared to another.

How much will this increase the size of each RRD do you think? (looking at port rrds for example)

murrant:
Maybe a tool wouldn't be too bad of an idea, but we definitely should not do it automatically.

pobradovic08:
Difference in size should be proportional to the increase of number of datapoints (rows), around 25%, which is consistent with size diff of couple of RRDs i compared (before and after).

murrant:

PR #11996, opened 2020-08-03
Add option to use SNMP for pinging

tpboudreau:
This contribution adds an option to perform all device pings using SNMP rather than ICMP (fping).  This can be desirable when LibreNMS is deployed in a Kubernetes cluster that disallows executables from escalating their privilege via the setuid bit on ping/fping, or from having the required Linux capabilities enabled on the ping/fping executable.

The option is enabled for an installation by setting the new boolean configuration value 'use_snmp_ping' to 'true'.  When enabled, all pinging (discovery, polling, etc) is simulated using the snmpstatus tool.  A second new boolean configuration value, 'record_snmp_ping_rtt', controls whether or not the round trip time for snmpstatus pings are recorded for the device.  The number of retry attempts and timeout parameter are set through the new 'snmp_ping_retries' and 'snmp_ping_timeout' configuration settings.

(I'm in the process of developing some tests, so this is WIP for now.)

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Sping.php
+++ b/LibreNMS/Sping.php
@@ -1,0 +1,146 @@
+<?php
+/**
+ * Sping.php
+ *
+ * -Description-
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2020 Tony Murray
+ * @author     Tony Murray <murraytony@gmail.com>
+ */
+
+namespace LibreNMS;
+
+use App\Models\Device;
+use Log;
+
+class Sping
+{
+    private $oid;
+    private $timeout;
+    private $retries;
+
+    /**
+     * Create a new SNMP ping instance.
+     */
+    public function __construct()
+    {
+        $this->oid = "1.3.6.1.2.1.1.1.0";
+        $this->timeout = (Config::get('snmp_ping_timeout', 5)) * 1000000;
+        $this->retries = Config::get('snmp_ping_retries', 2);
+    }
+
+    /**
+     * Run sping against a device and collect stats.
+     *
+     * @param object $device
+     * @return object
+     */
+    public function sping($device)
+    {
+        $response = [
+            'result' => (boolean)false,
+            'last_ping_timetaken' => (float)0.0,
+            'db' => [
+                'xmt' => (int)1,
+                'rcv' => (int)0,
+                'loss' => (float)100.0,
+                'min' => (float)0.0,
+                'max' => (float)0.0,
+                'avg' =>(float)0.0
+            ]
+        ];
+
+        $version = $device['snmpver'];
+        $target = Device::pollerTarget($device['hostname']);
+
+        $rtt = (boolean)false;
+        if (strpos($version, '2c') !== false) {
+            $rtt = $this->sping2c($device, $target);
+        } elseif (strpos($version, '3') !== false) {
+            $rtt = $this->sping3($device, $target);
+        } else { // version 1
+            $rtt = $this->sping1($device, $target);
+        }
+
+        if ($rtt !== false) {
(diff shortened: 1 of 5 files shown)


tpboudreau:
I'm AFK this week, so the earliest I'll be able to respond to feedback is the week of 8/10.

murrant:
When I tested snmpping, it looks like it requires additional setup and priveleges.

Error in packet.
Reason: noAccess

Also, why call an external bash command.  You should just call it from within PHP.  Also, if you are requesting that some snmp device icmp pings another device, and you measure that time for that aren't you also measuring the time of the snmp request+icmp ping time, not just the icmp ping time?

murrant:
Additionally, did you know you can disable the icmp check?

tpboudreau:
> Looks interesting.

Thanks for the quick feedback, sorry for the delayed response.

> 
> When I tested snmpping, it looks like it requires additional setup and privileges.
> 
> Error in packet.
> Reason: noAccess
> 
Did you mean that you tried to run the *sping* (not snmpping) bash script and got those errors?  The sping script takes a single bar delimited string of values as an argument.  It's not really meant to be a general purpose tool, but I'm still able to run it successfully from the command line against devices with snmpd running (invoked something like `./sping 'udp|test-device-zero|161|2c|librenmsCommunity'` for a v2c device).  I have never seen those error messages when using sping.

> Also, why call an external bash command. 

I implemented sping as an external bash script so that the control-flow would be as near as possible to a drop-in replacement for fping.

> Also, if you are requesting that some snmp device icmp pings another device, and you measure that time for that aren't you also measuring the time of the snmp request+icmp ping time, not just the icmp ping time?

I'm not sure I follow this.  The sping script doesn't call ping or fping -- it runs snmpstatus as a substitute for fping.  The RTT for an SNMP request/response is longer than for ICMP, but it's not double counting anything.

tpboudreau:
> Additionally, did you know you can disable the icmp check?

I'm not a regular LibreNMS user, but it seems if you disable ICMP pinging, your device is always shown as down and several historical graphs are not populated.

murrant:
@tpboudreau no reason for it to be that way other than no one cared enough to submit a change to make icmp disabled better.

Sorry, I mis-understood this PR a bit.  I don't think sping as a droppin replacement for fping is the best approach unless you want to do it without making changes to LibreNMS upstream.

snmpstatus is probably not the best tool to measure rtt, it doesn't measure rtt itself, and it can pull in a variable amount of data to answer and add arbitrary time.  You probably just want something like snmpget and sysDescr.

tpboudreau:
I've made changes along the lines you suggested.  I think the code is RFR.  (The diff is a bit misleading in places: the code associated with ordinary fping probing has been relocated in places, but should be functionally unchanged.)

murrant:
Getting better, but why the new config setting and not just use icmp_check / override_icmp_disable?
We don't generally use the php snmp functions, any reason you have chosen to use them here?
Try to think of ways to minimize code changes in other areas.

tpboudreau:
Sorry for the delay in getting back, I was detained on other projects.

> Getting better, but why the new config setting and not just use icmp_check / override_icmp_disable?

I'll use the icmp_check if required, but I think this is probably not ideal for end-users.  First, it would change the meaning of an existing option from "turn off pinging" to "choose snmp over icmp for pinging".  Second, it seems that it eliminates one of the three useful options: "no pinging", "standard (icmp) pinging", and "alternative (snmp) pinging".

> We don't generally use the php snmp functions, any reason you have chosen to use them here?

I avoided using the pre-existing snmp_get() function (includes/snmp.inc.php) both because (1) it has side effects (like recording statistics) that seemed like they should be left unaffected by this ping-related change, and because (2) it operates by creating an external process (via external_exec()) which skews the timings as you alluded to in a previous comment.

If you think that using the snmp_get() function is still preferable, I'll make the necessary changes, please let me know.

In the meantime, I'll fix the conflicts that have arisen while I wait to hear back.

murrant:
What if instead of adding more snmp requests to emulate icmp (which it can't), we record the performance of snmp requests we send and add that to a new rrd file?

tpboudreau:
Thanks for the suggestion.  I agree that reporting the snmp RTT's for the ping RTT's is not ideal, I expect some users will choose the new option not to report it.

But could you elaborate on why you think using the success/failure of an snmp call is not a reasonable substitute for ping success/failure (when ping is not available) for up/down tests and device discovery?  Is there another protocol you think is better suited?  

I'm trying in this PR to allow discovery and polling proceed without fping, while making as few changes as possible to the overall system architecture.  I'm not sure just tracking existing snmp requests accomplishes that, but I could be missing something.  

Thanks again.

murrant:
But up/down and discovery already works without ping.

What you are trying to add is RTT and loss tracking information.

stackedsax:
Hey there @murrant, just curious, how can we configure discovery/device add/poller (for up/down) to work without ping?  Ping looks pretty central to that feature, but maybe there's something I haven't seen yet.

murrant:
Set this to on:
image

PR #12017, opened 2020-08-18
Adding OOB Management to the device type definitions

m4rcu5:
Signed-off-by: Marcus van Dam <marcus@marcusvandam.nl>

In #5744 the initial support for Opengear devices was added, introducing the `OOB Management` device type. However this device type is not included in the list of `device_type` definitions, leading to the artefact, that when editing a device of type `OOB Management`, you will always inadvertently change its type to `server` as the type drown-down menu is not populated correctly.

In this commit I am adding the type `OOB Management` to the list of device types.

**note 1:** There seem to be a fair amount more of these cases, where there is a device type defined in the definitions, but not in the default `device_types`. I compiled the following list. **bold** is included right now.
- **appliance**
- camera
- **collaboration**
- encoder
- **environment**
- environmental (single device, could be a typo?)
- **firewall**
- **loadbalancer**
- management
- **network**
- 'OOB Management' (this pr)
- **power**
- **printer**
- proxy
- sensor
- **server**
- **storage**
- timing
- **wireless**
- **workstation**

**note 2:** The `icon` property seems to be populated in all the defaults, however the image cannot be found anywhere in the source. I just went along with it and did the same.

If you'd like, I can extend the PR to fix the other types as well. Possibly merging some which seem to be misclassified.

#### Please note

#### Testers

### DIFF
--- a/misc/config_definitions.json
+++ b/misc/config_definitions.json
@@ -908,6 +908,11 @@
                     "type": "environment"
                 },
                 {
+                    "icon": "oob-management.png",
+                    "text": "OOB Management",
+                    "type": "OOB Management"
+                },
+                {
                     "icon": "loadbalancer.png",
                     "text": "Load Balancers",
                     "type": "loadbalancer"


murrant:
I don't think the icons are used.
Should either populate them all (except the one-off) or just make the dropdown automatically add an item whatever it is currently set to.
Icons are not used, they were at one time.

m4rcu5:
> Should either populate them all (except the one-off) or just make the dropdown automatically add an item whatever it is currently set to.

That would indeed make more sense. I will close this PR in favour of #12031.

PR #12025, opened 2020-08-21
Fix incorrect eth0 status for Ubiquiti AirFiber 5XHD

nightcore500:
Ubiquiti uses separate OIDs for reading the ethernet port status. When reading out via IF-MIB the device reports that eth0 is down. This leads to eth0 being reported as "down" on every discovery. 

#### Please note

#### Testers


Jellyfrog:
Test fails

nightcore500:
I am currently in contact with Ubiquiti because the problem is probably a bit deeper. It seems to be temporarily not possible to read the afLTUethConnected OID.

Jellyfrog:
if youre polling new OIDs you need to generate new snmprec also, else tests wont match

nightcore500:
@Jellyfrog 
But I do not read out a new OID. I read the same OID which is already read at every polling. The OID is also present in both snmprec's.

nightcore500:
@murrant 
I also had this thought. It seems to concern only the 1.4.0 variant. I will take a closer look at it tomorrow.

Jellyfrog (approved):

murrant:

PR #12028, opened 2020-08-24
update bdcom support (mempool, software version, serial number, hw version)

hrtrd:

#### Please note

#### Testers


hrtrd:
@Jellyfrog help pls

includes/polling/os/bdcom.inc.php
```diff
@@ -1,13 +1,23 @@
 <?php
```
Jellyfrog, line 1: This should be ported to the new format while you're at it. See https://github.com/librenms/librenms/blob/master/LibreNMS/OS/Qnap.php#L34 or similar

Jellyfrog:
Probably some new entries for the test data also?

includes/polling/os/bdcom.inc.php
hrtrd: hm, ill try it for new os, today. if I succeed, I will transfer all my OS

includes/polling/mempools/bdcom.inc.php
```diff
@@ -1,26 +1,20 @@
...
+$perc = snmp_get($device, ['.1.3.6.1.4.1.3320.3.6.10.1.12.0'], '-OQv', 'NMS-CHASSIS');
 
-$mempool['free']  = ['bdcomMemoryPoolFree.0'];
-$mempool['used']  = $memory_pool['bdcomMemoryPoolUsed.0'];
-$mempool['total'] = $mempool['free'] + $mempool['used'];
-$mempool['perc']  = $mempool['used'] / $mempool['total'];
+if (is_numeric($perc)) {
+    $mempool['perc'] = $perc;
+    $mempool['used'] = $perc;
+    $mempool['total'] = 100;
```
murrant, line 18: memory is meant to be in bytes, not percent...  This will show up as 100 bytes of memory.

hrtrd: bdcom >2019year
#sh mem-util 
 mem util threshold is 90%
 current mem util is 27%
 Status:Normal

Jellyfrog: You sure there is not a some OID containing the value in bytes?

hrtrd: yes. bdcom changed the policy, and now all their systems are brought to the same form. This data is also displayed in the original monitoring software.

PipoCanaja: @hrtrd Not a single place (inventory, entity-mib, etc etc) where an absolute value for the total memory would be provided ?

hrtrd: Screenshot_1356
only this (screenshot). Show usage in percent.  the vendor switched to the new mib libraries. We use now in / opt / librenms / mibs / bdcom already deprecated

PipoCanaja:
In the screen capture, @hrtrd , we can see nmsnvRAMUsed and nmsnvRAMSize. So I would suggest to check those, and you can probably re-create all the missing values from it. 
Bye

hrtrd:
> In the screen capture, @hrtrd , we can see nmsnvRAMUsed and nmsnvRAMSize. So I would suggest to check those, and you can probably re-create all the missing values from it.
> Bye

Nope. This oids only for BDCOM NMS app, for CPE memory/

includes/polling/os/bdcom.inc.php
murrant: @hrtrd actually, better to move it to Yaml now.  See docs: https://docs.librenms.org/Developing/os/Initial-Detection/#yaml-based-os-discovery

Jellyfrog:
@hrtrd maybe remove mempool for now so the rest could be merged?

PR #12029, opened 2020-08-25
Add functionality to use snmp-scan.py to addto specific poller group

nathanshiaulam:

I'm adding functionality to specify what poller group snmp-scan.py will add devices to. If a user calls `snmp-scan.py` with `-g {POLLER_GROUP}`, all devices will be added to the specified poller group. Otherwise, all devices will be added to the first poller group specified under `distributed_poller_group`, and 0 of neither are provided.

#### Please note

#### Testers

### DIFF
--- a/snmp-scan.py
+++ b/snmp-scan.py
@@ -45,7 +45,7 @@ class Outcome:
     TERMINATED = 6
 
 
-POLLER_GROUP = 0
+POLLER_GROUP = '0'
 VERBOSE_LEVEL = 0
 THREADS = 32
 CONFIG = {}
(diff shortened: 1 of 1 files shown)


murrant:
Lint check failed... other than that looks ok.

daichaoyang:
Got it



发自我的iPhone


------------------ Original ------------------

nathanshiaulam:
Sounds good. Not sure why the build is breaking- looked at the stack trace and doesn't seem related, but I might've missed something

nathanshiaulam:
@murrant any clue on how to resolve the lint failure? been poking around but still not sure how it's related/how to resolve

murrant:
The testing error seems unrelated :(

But when I tested I got this:
```
./snmp-scan.py 
Traceback (most recent call last):
  File "./snmp-scan.py", line 178, in <module>
    POLLER_GROUP = args.group or CONFIG.get('distributed_poller_group').split(',')[0]
AttributeError: 'int' object has no attribute 'split'
```

nathanshiaulam:
That's weird, it works just fine for me. Did you set `distributed_poller_group` in your config.php as a string or an int?

nathanshiaulam:
Double checked/confirmed that it does work when `distributed_poller_group` is set to a string/fails if it's an int.

@murrant should I cast the value to a string in case someone inputs an int? Not sure if that's an edge-case I should handle since the value should be a string

nathanshiaulam:
@murrant just casted the input to a string in case the user inputs an int

murrant (approved):

murrant:

PR #12036, opened 2020-08-27
WIP: Option to only check for duplicates within poller groups

nathanshiaulam:
This PR does two things:

- Check for overwrite_ip when looking for dupe hostnames. The overwrite_ip should be deprecated soon, but we should add this check in the meantime since auto-adding devices can re-add the dupe devices if it's given a hostname that doesn't resolve to the overwrite_ip.
- Allow for dupe host checks to be poller group specific. In a distributed poller system with remote pollers in separate networks, you can have multiple devices that share the same private IP but aren't actually related (sitting behind NAT for instance). When auto-adding devices from these pollers, the system should be able to distinguish which device is actually a dupe within the same network. 

For more context: https://community.librenms.org/t/only-check-duplicate-ips-within-poller-group-when-auto-adding-devices/13212/2

#### Please note

#### Testers


murrant:

https://community.librenms.org/t/only-check-duplicate-ips-within-poller-group-when-auto-adding-devices/13212/3

nathanshiaulam:
To deal with the complexity errors pointed out by codecliimate, I might refactor the code to include a new flag in config.php called `is_remote_poller`. If the flag is set to `True`, it will only check for duplicate hosts within the poller group, since the poller could be pulling data from a private network with its own hostname resolution/private IPs. Moving this to WIP

nathanshiaulam:
Still getting the complexity error for `device_has_ip` but not sure how to simplify it anymore. Any thoughts here? 

Also the `$ip` variable name was already there, so I'm leaving it for now.

murrant:
Codeclimate is only for suggestions.  But we can't mark it as optional on github.

murrant:
Please don't add the setting.  Poller group always exists whether using distributed polling or not, so you can just check poller group always.  Also there is no such thing as remote poller officially in LibreNMS.

nathanshiaulam:
@murrant Sounds good, I'll revert the config setting. In that case, should I make it always check for dupe hosts within the poller group? I was thinking about that, but I was worried about the case where a user would want to check for dupe hosts system-wide, regardless if it was in the same poller group or not.

nathanshiaulam:
Worked through this a bit more and I think the new desired functionality is to provide a optional flag called `within-poller-groups` (could use suggestions on the name) in `addhost.php` which would allow adding to take in a list of poller groups to check for duplicate IPs. This would enable LibreNMS to poll multiple private subnets that share an address space/is required because of the `device_has_ip` check.

nathanshiaulam:
@murrant Updated to take in a list of poller groups and removed the config change

nathanshiaulam:
@murrant bumping this

addhost.php
```diff
@@ -197,7 +205,7 @@
     }//end if
 
     try {
-        $device_id = addHost($host, $snmpver, $port, $transport, $poller_group, $force_add, $port_assoc_mode, $additional);
+        $device_id = addHost($host, $snmpver, $port, $transport, $poller_group, $force_add, $port_assoc_mode, $additional, $within_poller_groups);
```
murrant, line 208: You should pass this as a key in the $additional array.

includes/functions.php
```diff
@@ -1483,24 +1486,35 @@ function fix_integer_value($value)
...
             ->where('ipv4_address', $ip)
             ->with('port.device')
-            ->first();
+            ->get();
     }
+    if (isset($ip_addresses)) {
+        $ip_addresses = $ip_addresses->filter(function ($ip_address) {
+            return isset($ip_address->port);
+        });
+        if (! empty($within_poller_groups)) {
```
murrant, line 1509: You should do this in the query instead of in php.

nathanshiaulam: Was able to move the port existence check into the query, but when I did a double join on `ports` and `devices` along with the `with('port.device')` clause to check for thee poller group, I was able to get the correct IP address (when I print it I see both `port` and `device` in the JSON), but `$ip_address->port->device` gives me null.

Here's the old query with both port and poller group checks:
```
        $ip_addresses = \App\Models\Ipv4Address::query()
            ->where('ipv4_address', $ip)
            ->whereNotNull('ipv4_addresses.port_id')
            ->join('ports', 'ipv4_addresses.port_id', '=', 'ports.port_id')
            ->whereNotNull('ports.device_id')
            ->join('devices', 'ports.device_id', '=', 'devices.device_id')
            ->when(! empty($within_poller_groups), function($query) use ($within_poller_groups) {
                    return $query->whereIn('devices.poller_group', $within_poller_groups);
            })
            ->with('port.device')
            ->first();
```

addhost.php
nathanshiaulam: Done

nathanshiaulam:
@murrant Should be ready for review

nathanshiaulam:
@murrant bump again, fingers crossed it's good to go this time but lmk if there's anymore feedback

includes/functions.php
murrant: The joins break the eloquent loading.  You can check a relationship like this: (may not be 100% correct)
```
->whereHas('port.device.groups', function ($query) use ($within_poller_groups) {
    $query->whereIn('device_groups.id', $within_poller_groups);
})
```

murrant:
One thing that bothers me a little bit is all the additional variables on functions.  Not sure if it can be helped.  If you can think of a way to reduce that, kudos.

also, you can change the isset() things to one line `$var = $arr['key'] ?? [];`

includes/functions.php
nathanshiaulam: Got it, thanks. Since the join breaks eloquent loading can I keep the poller group check outside of the query?

nathanshiaulam:
> One thing that bothers me a little bit is all the additional variables on functions. Not sure if it can be helped. If you can think of a way to reduce that, kudos.
> 
> also, you can change the isset() things to one line `$var = $arr['key'] ?? [];`

Sounds good, I'll update parts where I use isset(). Yeah it doesn't seem like an ideal solution- I feel like an alternative would be to add groupings to these poller groups (like a poller network but tbd on naming) and create an interface in the UI. In that case, if a poller group belongs to a poller network, it would only look for dupes within that. That seems like a much larger refactor/would still require additional params in functions.php though, so not much added benefit there beyond making it more explicit.

nathanshiaulam:
^bump @murrant

haydenseitz:
wouldn't adding duplicate IPs have side effects on how networks and links are stored? for example, having multiple 192.168.0.0/24 networks being monitored in the same librenms instance could allow incorrect port relationships being displayed on the device ports pages

nathanshiaulam:
> wouldn't adding duplicate IPs have side effects on how networks and links are stored? for example, having multiple 192.168.0.0/24 networks being monitored in the same librenms instance could allow incorrect port relationships being displayed on the device ports pages

I don't think I've observed that in the pollers that I've been running since each address is unique in its own poller group/the network is split up by poller groups.  For instance, I have two separate pollers in separate networks monitoring a device at 100.64.0.1, and all the port/address data is linked up properly on the ports page. 

I'm not too familiar with how links are stored, so I can dig in deeper to make sure there aren't any side-effects or edge cases. Could you give me a few pointers on where to start? If not I'll grep around.

haydenseitz:
here's where network_id is used to match ports https://github.com/nathanshiaulam/librenms/blob/72559490408020ec30bdabb5be2161ba8bc3ec43/includes/html/print-interface.inc.php#L191

you could probably update the queries related to `enable_port_relationship` to also check for matching poller groups

nathanshiaulam:
> here's where network_id is used to match ports https://github.com/nathanshiaulam/librenms/blob/72559490408020ec30bdabb5be2161ba8bc3ec43/includes/html/print-interface.inc.php#L191
> 
> you could probably update the queries related to `enable_port_relationship` to also check for matching poller groups

Got it thanks will look into it. FWIW every device I have polled on separate pollers sharing the same IP address all have their ports and addresses assigned properly- not sure why this is the case though. It's possible it's due to hostname matching, since they don't share the same hostnames. If that's the case, I don't think this would be an issue since hostnames still have to be unique across the different pollers.

nathanshiaulam:
@haydenseitz Wait just to make sure we're on the same page, are we talking about the data populated on the `/device/device={DEVICE_ID}/tab=ports/view=basic/` page?

nathanshiaulam:
> @haydenseitz Wait just to make sure we're on the same page, are we talking about the data populated on the `/device/device={DEVICE_ID}/tab=ports/view=basic/` page?

Just looked through the code, and it's definitely filtering by device_id, so I don't think any changes need to be made here.

> here's where network_id is used to match ports https://github.com/nathanshiaulam/librenms/blob/72559490408020ec30bdabb5be2161ba8bc3ec43/includes/html/print-interface.inc.php#L191
> 
> you could probably update the queries related to `enable_port_relationship` to also check for matching poller groups

Just looked through the code and it's already filtering by device ID in the query, so I don't think any changes need to be made here:

```
        foreach (dbFetchRows("SELECT `ipv4_network_id` FROM `ipv4_addresses` WHERE `port_id` = ? AND `ipv4_address` NOT LIKE '127.%'", [$port['port_id']]) as $net) {
            $ipv4_network_id = $net['ipv4_network_id'];
            $sql = 'SELECT I.port_id FROM ipv4_addresses AS A, ports AS I, devices AS D
                WHERE A.port_id = I.port_id
                AND A.ipv4_network_id = ? AND D.device_id = I.device_id
                AND D.device_id != ?';
```

murrant:
You can only have multiple devices with the same IP if you are using hostnames.  You cannot do this when adding devices by IP.

nathanshiaulam:
> You can only have multiple devices with the same IP if you are using hostnames. You cannot do this when adding devices by IP.

That makes sense. Either way, since the ports page filters by device ID and ports have a reference to the `device_id`, I don't think the mismatch @haydenseitz brought up will be an issue/we shouldn't need to add anything extra to the query.

haydenseitz:
If there was a device at 192.168.1.5 on LAN 1 and another device 192.168.1.6 on LAN 2 (two separate 192.168.1.0/24 networks), wouldn't the port association code show that those devices are in the same network? I not sure if there would be separate entries for 192.168.1.0/24 in the `ipv4_address` table

nathanshiaulam:
> If there was a device at 192.168.1.5 on LAN 1 and another device 192.168.1.6 on LAN 2 (two separate 192.168.1.0/24 networks), wouldn't the port association code show that those devices are in the same network? I not sure if there would be separate entries for 192.168.1.0/24 in the `ipv4_address` table

In the code snippet you sent, I think it's filtering by the device_id as well, so it wouldn't show incorrect addresses on the page:
```
    if ($port_details && Config::get('enable_port_relationship') === true) {
        // Show which other devices are on the same subnet as this interface
        foreach (dbFetchRows("SELECT `ipv4_network_id` FROM `ipv4_addresses` WHERE `port_id` = ? AND `ipv4_address` NOT LIKE '127.%'", [$port['port_id']]) as $net) {
            $ipv4_network_id = $net['ipv4_network_id'];
            $sql = 'SELECT I.port_id FROM ipv4_addresses AS A, ports AS I, devices AS D
                WHERE A.port_id = I.port_id
                AND A.ipv4_network_id = ? AND D.device_id = I.device_id
                AND D.device_id != ?';
```
I'm also observing this in my current deployments with this code running. I have multiple devices at the same address in different LANs, and the ports page doesn't show incorrect relationships because it filters by device ID.

nathanshiaulam:
> If there was a device at 192.168.1.5 on LAN 1 and another device 192.168.1.6 on LAN 2 (two separate 192.168.1.0/24 networks), wouldn't the port association code show that those devices are in the same network? I not sure if there would be separate entries for 192.168.1.0/24 in the `ipv4_address` table

Mb, I think I found the issue. I'll push up a fix later today/tomorrow after I confirm it's resolved. For some reason it was only showing incorrect links on some devices even though the addresses were the same. Thanks for catching that!

nathanshiaulam:
I think this change might be more complicated than adding a check for the device's poller group. @murrant brought up that I'm basically trying to create "sites" with a group of poller groups. Because I check for duplicate IPs within a group of poller groups, I'll need to add that same check for the port relationships. However, because the PR doesn't explicitly store sites anywhere/just takes in a set of poller groups when adding devices, I don't have a way of knowing which set of poller groups to check. 

In my own deployments running this code, I'm currently creating makeshift sites with device groups containing all the pollers that belong to that site (i.e. basically an OR across all `device.poller_group`s in that site). For scanning, I have a custom script that queries the device group a specific poller belongs to and runs `snmp-scan.py` against those poller groups. It seems like we definitely need something more explicit to keep it consistent throughout. 

One option is to continue using device groups and add a flag to specify that it's a site, but that seems like a hack. The benefit there is that a lot of functionality we'd want with a "site" already exists for device groups (e.g. dependency graphs, separate views to monitor devices). Will try to think of something more explicit/move this to a WIP til I figure out a better way to do this.

murrant:
Closing this for now... sorry it got so complex.

murrant:
Feel free to re-submit or post here for us to reopen.

nathanshiaulam:
Any chance we can keep this open? Bit of an odd request, but I'm currently using it in our workflow by using github-apply. Not sure if that works for closed ticket IDs as well- also planning on re-visiting this sometime later in the month when I have more bandwidth

murrant:
It should still work.  You can always get the diff from your local branch too.  Because this needs a serious rework, I'm less inclined to leave it open.  Will be happy to open it when you get back to it.

PR #12040, opened 2020-08-29
service watchdog - add systemd watchdog for resiliency

bofh80:
I had 2 poller failures, both fired watchdog restarts, yet failed to poll and log errors.
systemctl restart librenms fixed both. I assume shutdown will lead to a service restart (restart=always is set on the service)

#### Please note

#### Testers


mcineran:
My probes are restarting around once an hour with this patch. Which while seems odd, is better than randomly dying after 8. or 12 hours.  Not 100% sure why it would fire so often, testing welcome, but works for me.

bofh80:
I sorted out problem with Redis and my strange service restarts have fixed themselves. This functioned as required and after a BARK, service watchdog restart, the logs still did not show any polling, then the systemd watchdog kicked in and restarted the service.
I should probably adjust the log messages to suggest if this has occurred there may be other underlying problems.

LibreNMS/service.py
```diff
@@ -605,9 +605,12 @@ def logfile_watchdog(self):
...
             return
 
-        if logfile_mdiff > self.config.poller.frequency:
-            critical("BARK! Log file older than {}s, restarting service!".format(self.config.poller.frequency))
-            self.restart()
+        if self.config.watchdog_enabled:
+            if logfile_mdiff > self.config.poller.frequency:
+                critical("BARK! Log file older than {}s, restarting service!".format(self.config.poller.frequency))
+                self.restart()
+            else:
```
murrant, line 612: The info message in the else does not line up with it does.  (It can only be reached if watchdog is enabled)

bofh80: Does that look better?

bofh80:
sorry this went wrong on the git rebase. i'll start a new branch.

PR #12044, opened 2020-08-31
WIP dbUpdate add transaction support

bofh80:
First attempt at adding transaction support to avoid SQL lnterrupted and deadlocks. (i think)
I haven't tested this as yet, looking for comments first.

#### Please note

#### Testers

### DIFF
--- a/includes/dbFacile.php
+++ b/includes/dbFacile.php
@@ -210,8 +210,9 @@ function dbUpdate($data, $table, $where = null, $parameters = [])
     }
 
     try {
-        $result = Eloquent::DB()->update($sql, (array)$data);
-
+        Eloquent::DB()->transaction(function () {
+            $result = Eloquent::DB()->update($sql, (array)$data);
+        }, 5};
         recordDbStatistic('update', $time_start);
         return $result;
     } catch (PDOException $pdoe) {
--- a/includes/dbFacile.php
+++ b/includes/dbFacile.php
@@ -212,7 +212,7 @@ function dbUpdate($data, $table, $where = null, $parameters = [])
     try {
         Eloquent::DB()->transaction(function () {
             $result = Eloquent::DB()->update($sql, (array)$data);
-        }, 5};
+        }, 5);
         recordDbStatistic('update', $time_start);
         return $result;
     } catch (PDOException $pdoe) {
--- a/includes/dbFacile.php
+++ b/includes/dbFacile.php
@@ -210,7 +210,7 @@ function dbUpdate($data, $table, $where = null, $parameters = [])
     }
 
     try {
-        $result = Eloquent::DB()->transaction(function () {
+        $result = Eloquent::DB()->transaction(function () use($data, $sql){
             Eloquent::DB()->update($sql, (array)$data);
         }, 5);
         recordDbStatistic('update', $time_start);
--- a/includes/dbFacile.php
+++ b/includes/dbFacile.php
@@ -210,8 +210,8 @@ function dbUpdate($data, $table, $where = null, $parameters = [])
     }
 
     try {
-        Eloquent::DB()->transaction(function () {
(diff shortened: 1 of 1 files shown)


includes/dbFacile.php
```diff
@@ -210,8 +210,9 @@ function dbUpdate($data, $table, $where = null, $parameters = [])
     }
 
     try {
-        $result = Eloquent::DB()->update($sql, (array)$data);
-
+        $result = Eloquent::DB()->transaction(function () {
```
Jellyfrog, line 213: Missing `use($data, $sql)`

Jellyfrog:
I really doubt we want to use transactions for every single insert

mcineran:
I see this repeated in the logs:
```
[2020-08-31 19:10:39] production.ERROR: SQLSTATE[70100]: <<Unknown error>>: 1317 Query execution was interrupted (SQL: UPDATE `ports` set `ifInOctets`=306459307,`ifInOctets_prev$
#1 /opt/librenms/includes/polling/functions.inc.php(315): include('/opt/librenms/i...')
#2 /opt/librenms/poller.php(142): poll_device(Array, false)
#3 {main}  
[2020-08-31 19:10:39] production.ERROR: SQLSTATE[40001]: Serialization failure: 1213 Deadlock found when trying to get lock; try restarting transaction (SQL: UPDATE `ports_stati$
#1 /opt/librenms/includes/polling/functions.inc.php(315): include('/opt/librenms/i...')
#2 /opt/librenms/poller.php(142): poll_device(Array, false)
#3 {main}
```
For various updates. from the docs, this seemed like the obvious to do:
https://laravel.com/docs/7.x/database#database-transactions

So is there another way to avoid or at least *help* with the above errors ?

includes/dbFacile.php
mcineran: Do I just need to add that line in?
$result = Eloquent::DB()->transaction(function () {
    use($data, $sql)
Or does it need to be in function() or after that call? (apologies my php is not strong)

bofh80: I believe i've updated this correctly, but now it's working on PHP 7.2 and 7.4.  But failing on PHP 7.3.
Before the change it was the other way around.

bofh80: Fixed, thanks! that sorted the travis bulds.

Jellyfrog:
I never seen that. I think you should post your issue on the forum instead. 
Also your log is cut off.

mcineran:
Apologies it seems I should have mentioned a couple of things. I'm runing this with a galera backend. 2 hosts, each with a poller, each pointing at their own host. (with other stuff done for redis memcached etc).
So I assume this is the cause, but I need the pollers and hosts to be HA.
When I run with this code i do still see errors:
```
[2020-09-01 22:08:01] production.ERROR: There is already an active transaction (SQL: UPDATE `alerts` set `alerted`=1 WHERE rule_id = 22 && device_id = 62)#0 /opt/librenms/LibreNMS/Alert/RunAlerts.php(483): dbUpdate(Array, 'alerts', 'rule_id = ? && ...', Array)
#1 /opt/librenms/alerts.php(53): LibreNMS\Alert\RunAlerts->runAlerts()
#2 {main}  
```
What I'm unsure of at this point (i've been reading the docs) is if this alert fired once (as per the retries) logged the alert, then succeeded, so no further errors. OR if it logs the above when the retries are exhausted, and should repeat X amount of retries in the logs (the same event repeated X times).

From the docs it seems that this might be logged on each attempt. But if anyone can claify? :)

mcineran:
Yeah, just to confirm if anyone comes across this before, due to my redis backend. Fixed that and this issue went away.

murrant:

https://community.librenms.org/t/mariadb-galera-cluster-general-error-1205-lock-wait-timeout-exceeded/14201/2

PR #12058, opened 2020-09-03
Python log mysql connection errors and harden config retrieval

deajan:
This small patch enables logging of database connection problems.
Also hardenes subprocess execution calls.

It's the beginning of a larger series of patches I initially wrote for librenms python3 compat.
The `command_runner` function will be used later in snmp-scan, poller-wrapper, discovery-wrapper and services-wrapper.py.

Underlying motivation:
I had problems with poller-wrapper.py not working. Nothing in the logs.
I understood that I had mysql connection problems due to the fact that I had an accent `é` in my mysql password (this is another bug, but not librenms related).
So instead of only printing the errors, it's adivsable to log them.

Also when executing php to read the config file, it's better to use a more reliable way to deal with subprocesses.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/library.py
+++ b/LibreNMS/library.py
@@ -26,6 +26,8 @@ except ImportError:
 logger = logging.getLogger(__name__)
 
 # Logging functions ########################################################
+# The following logging functions are
+# (C) 2019-2020 Orsiris de Jong under BSD 3-Clause license
 
 FORMATTER = logging.Formatter('%(asctime)s :: %(levelname)s :: %(message)s')
 
@@ -103,6 +105,7 @@ def logger_get_logger(log_file=None, temp_log_file=None, debug=False):
                 _logger.warning('Failed to use log file [%s], %s.', log_file, err_output)
     return _logger
 
+# End of Logging functions #################################################
 
 # Generic functions ########################################################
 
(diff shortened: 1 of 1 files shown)


deajan:
The command_runner function works on both linux and windows, but produces some linter errors on linux since it uses subprocess members that only exists on windows, hence the last pylint ignore statements I added.

I use this function for a couple of years now under a lot of different environments.
As for codeclimate, the function is complex because it handles a lot of different cases, but it wouldn't make sense to divide it into subfunctions.

deajan:
@murrant Would you mind taking a look ? This is the first PR of the code refactoring that I wanted to do earlier with the Python3 support. There is a "heavy" executer function that wraps subprocess, but simplifies a lot of error handling. Thanks.

murrant:
@deajan Is the "command_runner" function pasted from somewhere?  Do you need to give code credit/copyright?
Also, I don't think LibreNMS will run on Windows any time soon and if it does, it won't be using python at that point.

All for more logging.

deajan:
I wrote the "command_runner" function a year ago and improved it since then (elder version of my work here https://stackoverflow.com/a/62772886).
I am going to put that one on PyPI toghether with a privilege elevator function (UAC & sudo) later, with a BSD licence.
I can add my copyright note of course.

As for the windows support, I understand librenms is a linux project. I only pointed out why I put a linter statement in the function since I don't want to strip of the windows support which would force me to maintain two different versions of the command, just to avoid a linter warning.

deajan:
Any news here ? I'd need this PR since it's the base for my python wrapper rewrite.

murrant (approved):

deajan:
@murrant I've rewritten the code as separate module in https://github.com/netinvent/command_runner which has:
- python 2.7 up to 3.8 testes
- code coverage
- code climate

I'd like to create a refactor this PR so the `command_runner` function uses the module available on PyPI.
Could you confirm that automatic updates also update python dependencies ?

Thanks.

murrant:
Not all installs have PIP available, so no.

PR #12088, opened 2020-09-11
New OS: Ekinops

h-barnhart:
New device OS: Ekinops Optical Networking.

The Ekinops have a custom SNMP implementation, so please keep that in mind when reviewing the code. Ekinops uses a chassis form factor with various line cards for specific tasks. Ekinops does not pool all OIDs available in the chassis into a single tree to query. Instead you must either 1) set a particular OID to the slot number you want to query via SNMP or 2) Modify the SNMP community string and append the slot number. As neither of these functions are currently available in LibreNMS this initial code pull will only include information available from the management card using the base SNMP community string. Additional functionality will be added over time.

This PR will include collecting ports, line cards, and line card state. It also modifies the port ifDescr and ifAlias to make them easier to read in the UI.

Edit:

This PR also adds trap handlers for log events (Mgnt2TrapNMSEvent) and alarms (Mgnt2TrapNMSAlarm).

Chassis tested by me include a C600CH and C200H.

#### Please note

#### Testers


h-barnhart:
Looks like TravisCI is tripping on my ifDescr changer. I can get rid of the changer if its not acceptable. Without it the interface names are really long:

EKINOPS/C600HC/8/PMROADMC/Express4(PORT_Number 4   )

PipoCanaja:
Hi @h-barnhart 
It seems that your test data is not matching this PR. Could you try to run again save-test-data.php and commit the resulting file ?

h-barnhart:
What I provided includes my code for modifying the interface ifAlias and IfDescr, which doesn't appear to take effect during the data gathering stage. My code for changing those values is in includes/polling/ports/os/ekinops.inc.php. Is there another method I should use or just drop the code for changing the ifDescr and ifAlias?

h-barnhart:
Ok, I had to put my code that modifies ifDescr and ifAlias under the discovery code path as well as the poller. I realize this is code duplication and if the discovery and polling subsystems were OOB it could be avoided. If there is a better way let me know.

Beyond that I think it should checkout now as the snmpsim data will match what is actually collected during discovery.

h-barnhart:
Forgot to add my snmpsim data, only updated the json

includes/polling/os/ekinops.inc.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
Jellyfrog, line 1: This file should be converted to the new-ish format, check here for more info, https://docs.librenms.org/Developing/os/Initial-Detection/#os-discovery

Jellyfrog, line 1: This file should be converted to the new-ish format, check here for more info, https://docs.librenms.org/Developing/os/Initial-Detection/#os-discovery

Jellyfrog (changes requested):

h-barnhart:
Figured out the issue with the SNMP test data. First of all, some of the bad-if settings were preventing all my interfaces from being discovered, so they were in the snmpsim but not the json. Second I was running the save-test-data script on my test server and uploading the JSON to my dev server. I tried uploading just the snmpsim and then running save-test-data from the dev server and now unit tests are passing on my dev install. I'm going to let the checks run on GitHub and go from there.

Just an FYI, on my dev install some other OS's were failing unit tests, nothing I can do about that.

@Jellyfrog If travis checks out I'll go back and re-do the code you pointed out.

includes/polling/os/ekinops.inc.php
murrant: Should be able to do it all in yaml now. (use regex to parse mgnt2RinvHwPlatform)

includes/discovery/sensors/pre-cache/ekinops.inc.php
```diff
@@ -0,0 +1,34 @@
...
+ * Traps when Adva objects are created. This includes Remote User Login object,
+ * Flow Creation object, and LAG Creation object.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2020 KanREN, Inc
+ * @author     Heath Barnhart <hbarnhart@kanren.net>
+ */
+
+$pre_cache['ekishelf'] = snmpwalk_cache_multi_oid($device, 'mgnt2GigmBoardTable', [], 'EKINOPS-MGNT2-MIB', null, '-OQUbs');
```
murrant, line 32: Why do you have  prefetch?  I can't see where it is used.

h-barnhart: Its used for sensor discovery

h-barnhart: The line below it doesn't need to be there, though.

h-barnhart:
The fail in Travis is for a different OS.

PipoCanaja:
@h-barnhart Could you rebase your PR, a big change was merged earlier today.

includes/discovery/sensors/pre-cache/ekinops.inc.php
murrant: this prefetch doesn't work there and all it does is fetch the data twice and only use it once.  (also the sensor code doesn't need prefetch it is extremely smart about how it fetches the data)

h-barnhart: I'll pull it from m

h-barnhart:
Do I need to make the changes from styleci?

tests/data/ekinops.json
```diff
@@ -0,0 +1,10572 @@
+{
+    "os": {
+        "discovery": {
+            "devices": [
+                {
+                    "sysName": "<private>",
+                    "sysObjectID": ".1.3.6.1.4.1.20044",
+                    "sysDescr": "Ekinops 360 Platform, C600HC, Release 6.3.900",
+                    "sysContact": null,
+                    "version": " Release 6.3.900",
```
Jellyfrog, line 10: Maybe want to use trim() to get rid of the spaces

h-barnhart: fixed

Jellyfrog:
Do we really need all these MIBs? I'm only seeing 2 used

h-barnhart:
I can pull the mibs that aren't needed right now, its more than two when you look at dependencies and the sensors.

Jellyfrog (approved):

Jellyfrog:
Nice work!

murrant:

PR #11331, opened 2020-03-23
Database: add primary key to device_graphs

rkojedzinszky:
Percona Xtradb refused to INSERT IGNORE into a table
without a primary key.

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)
- [ ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #12106, opened 2020-09-18
schema: add missing primary keys

rkojedzinszky:
Add primary keys to tables where no primary and unique keys were defined. This is needed for Percona Xtradb to accept writes to such tables. Similar PR: https://github.com/librenms/librenms/pull/11331

#### Please note

#### Testers

I found that the following tables had no primary and unique keys defined:
```
bill_perms
bill_ports
devices_perms
entPhysical_state
ipv4_mac
juniAtmVp
loadbalancer_vservers
ports_perms
processes
transport_group_transport
```

So all tables have been extended with an `id` field and a primary key.

Can this change break anything?

### DIFF
--- a/database/migrations/2018_07_03_091314_create_bill_perms_table.php
+++ b/database/migrations/2018_07_03_091314_create_bill_perms_table.php
@@ -13,6 +13,7 @@ class CreateBillPermsTable extends Migration
     public function up()
     {
         Schema::create('bill_perms', function (Blueprint $table) {
+            $table->bigIncrements(id);
             $table->unsignedInteger('user_id');
             $table->unsignedInteger('bill_id');
         });
--- a/database/migrations/2018_07_03_091314_create_bill_perms_table.php
+++ b/database/migrations/2018_07_03_091314_create_bill_perms_table.php
@@ -13,7 +13,7 @@ class CreateBillPermsTable extends Migration
     public function up()
     {
         Schema::create('bill_perms', function (Blueprint $table) {
-            $table->bigIncrements(id);
+            $table->id();
             $table->unsignedInteger('user_id');
             $table->unsignedInteger('bill_id');
         });
--- a/database/migrations/2018_07_03_091314_create_bill_ports_table.php
+++ b/database/migrations/2018_07_03_091314_create_bill_ports_table.php
@@ -13,7 +13,7 @@ class CreateBillPortsTable extends Migration
     public function up()
     {
         Schema::create('bill_ports', function (Blueprint $table) {
-            $table->bigIncrements(id);
+            $table->id();
             $table->unsignedInteger('bill_id');
             $table->unsignedInteger('port_id');
             $table->boolean('bill_port_autoadded')->default(0);
--- a/database/migrations/2018_07_03_091314_create_bill_ports_table.php
+++ b/database/migrations/2018_07_03_091314_create_bill_ports_table.php
@@ -13,6 +13,7 @@ class CreateBillPortsTable extends Migration
     public function up()
     {
         Schema::create('bill_ports', function (Blueprint $table) {
+            $table->bigIncrements(id);
             $table->unsignedInteger('bill_id');
             $table->unsignedInteger('port_id');
             $table->boolean('bill_port_autoadded')->default(0);
--- a/database/migrations/2018_07_03_091314_create_devices_perms_table.php
+++ b/database/migrations/2018_07_03_091314_create_devices_perms_table.php
@@ -13,6 +13,7 @@ class CreateDevicesPermsTable extends Migration
     public function up()
     {
         Schema::create('devices_perms', function (Blueprint $table) {
+            $table->bigIncrements(id);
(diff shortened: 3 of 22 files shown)


database/migrations/2020_09_18_1800_add_missing_primary_keys.php
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
```
PipoCanaja, line 90: Hi
It seems that you did not implement the "reverse" part of the migration. The idea is to be exactly back to the initial status after reverting a migration php file. This is one of the reason we can do "live" updates in a safer way ...

PipoCanaja (dismissed):

database/migrations/2020_09_18_1800_add_missing_primary_keys.php
rkojedzinszky: @PipoCanaja That is true. I found that my earlier pull request which originally contained a reverse migration step was changed by a developer, to not contain that anymore, and I've just followed that way. My earlier PR was #11331, and the file I've created then now dont have a reverse step too:

https://github.com/librenms/librenms/blob/master/database/migrations/2020_03_24_0844_add_primary_key_to_device_graphs.php

How to proceed now?

PipoCanaja: Right. @murrant : What do you suggest ? You removed the down() code for compatibility with sqlite ?

rkojedzinszky: @PipoCanaja @murrant ping

rkojedzinszky: @PipoCanaja @murrant please comment my work.

PipoCanaja: @rkojedzinszky Let's have @murrant opinion on this. 

- Aside of this, you could sign the CLA (this is a prerequisite).
- You should also rebase your patch, as a lot of changes took place in the last few days. This will most probably solve the error in TravisCI.

murrant: we don't really use down and sqlite cannot remove indexes without recreating the table.

murrant:
@rkojedzinszky why do you have the hasColumn checks?  If you are concerned about the migration failing mid migration, you should create one migration for each table affected.

database/migrations/2020_09_18_1800_add_missing_primary_keys.php
rkojedzinszky: > @rkojedzinszky why do you have the hasColumn checks? If you are concerned about the migration failing mid migration, you should create one migration for each table affected.

@murrant I've updated the patch. Basically I've just used my earlier modified migration step as a template here. Here the migration should run without errors, so no checks are needed. Do you like this way, or should I split into separate migration steps?

murrant: It is 11 tables and we have been burned by a bad migration before.  Probably best to split it up.  Thanks.

murrant (approved):

murrant:
Ah yes, SQLite doesn't allow adding these.  Probably need to add them to the original table creation and check if they exist before adding them in the migration.

rkojedzinszky:
> Ah yes, SQLite doesn't allow adding these. Probably need to add them to the original table creation and check if they exist before adding them in the migration.

Ok, so now what?

murrant:
Normally a no-no, but edit the original table creation migrations and add this field.
Then modify these migrations to pass even if the field already exists.

Sorry this has been such a pain, thanks for putting in the work.

rkojedzinszky:
> Normally a no-no, but edit the original table creation migrations and add this field.
> Then modify these migrations to pass even if the field already exists.
> 
> Sorry this has been such a pain, thanks for putting in the work.

I've rebased and pushed a new commit. Please have a look at it.

Richard

database/migrations/2018_07_03_091314_create_bill_perms_table.php
```diff
@@ -13,6 +13,7 @@ class CreateBillPermsTable extends Migration
     public function up()
     {
         Schema::create('bill_perms', function (Blueprint $table) {
+            $table->bigIncrements(id);
```
murrant, line 16: `id` should be `'id'`

murrant: Also, there is a shortcut for this:
```php
$table->id();
```

murrant (approved):
Looks good

rkojedzinszky:
Thanks for helping on this. I dont know php and the framework this deep. Thanks for merging.

murrant:
Thanks for sticking with it

murrant:

PR #12107, opened 2020-09-18
Add Service Templates

bofh80:
You create a Service Template and assign it to multiple Devices or  Device Groups.

Services (as specified by the template)  will be created for all Devices or Device Groups.

You can do this with the function buttons, or enable the discover_services_templates: true, to have it run and apply on discover intervals.

This is similar to how nagios works.( You assign checks to 'host groups' and then assign devices to host groups)

All code seems to work fine. But please see my comment below about the db schema and index.

Service Templates:
stv2

Add / Edit Service Template:
sted

Updated Services:
4srv

Updated Add Service:
4srvadd

Updated Edit Service:
5srvedit

#### Please note

#### Testers


bofh80:
So currently the travis build is failing on 
```
--- Expected
+++ Actual
@@ @@
             9 => Array (...)
             10 => Array (...)
             11 => Array (...)
-            12 => Array (...)
         )
         'Indexes' => Array (...)
     )
```
In my previous attempts to fix the travis builds i added and then commented the following in the database migration:
`            #$table->index(['services_template_device_group_id_index','device_group_id']);`
The db schema includes:
  Indexes:
```
    PRIMARY: { Name: PRIMARY, Columns: [service_template_id], Unique: true, Type: BTREE }
    services_template_device_group_id_index: { Name: services_template_device_group_id_index, Columns: [device_group_id], Unique: false, Type: BTREE }
```
I found this for the 'services' db migrations:
`database/migrations/2020_04_06_001048_the_great_index_rename.php:            DB::statement('ALTER TABLE services DROP INDEX service_host, ADD INDEX services_device_id_index(device_id);');`

I assume the current error might due to needing me uncomment the index, but produces other errors at that seem to duplicate services_template in the name.

Can anyone suggest the fix for me? (PR's to my branch are welcome.)

murrant:

https://community.librenms.org/t/new-feature-service-templates-adieu-nagios/13515/1

murrant:
@bofh80 I went through and fixed up issues for you so the pages at least come up. (`git pull` to pull them to you dev box)  You'll still need to go through them and make it all work.  Fixed some name inconsistencies and changed the route a bit.

murrant:
@bofh80 are you using the correct approach for assigning templates?  Device groups are dynamic and can change.  What happens when two device groups overlap?  When are templates applied?

Should a service template be changed or removed if the device group membership changes?
Should we just have a button that allows you to manually apply a service template to a device or device group?

bofh80:
@murrant Yes. This is how nagios works.
'Device groups are dynamic and can change'.  Yes. And services will be added and removed appropriately assuming you have auto discover enabled.
'What happens when device groups overlap'. Nothing. If a device is a member of more than 1 device group (expected), then templates from all assigned device groups will be applied.
'Should a service template.. . . ' Yes, if auto discover templates is enabled.
'Should we have a button ?' .   I've got 4 of them already. (Apply a template, Remove a template, Edit / Delete).
The first two.

bofh80:
And with the explanation above, the idea is, a new host is automatically discovered. automatically added to relevant device groups, then finally, has all the appropriate service checks applied.

bofh80:
Stuck again, i've made a few changes, tried a few things, going nowhere in a hurry.

In app\Models\ServiceTemplate.php:
```
public static function getServiceTemplate($query)
    {
        return $query->where([
            ['id', '=', $query],
        ]);
    }
```
In includes/services.inc.php:
```
function discover_service_template($service_template = null)
{
    $services_template = ServiceTemplate::getServiceTemplate($service_template);
```
Error:
```
[2020-10-17 10:19:14] development.ERROR: Call to a member function where() on string {"userId":1,"exception":"[object] (Error(code: 0): Call to a member function where() on string at /opt/librenms/app/Models/ServiceTemplate.php:36)
[stacktrace]
#0 /opt/librenms/includes/services.inc.php(131): App\\Models\\ServiceTemplate::getServiceTemplate('2')
#1 /opt/librenms/includes/html/forms/discover-service-template.inc.php(21): discover_service_template('2')
#2 /opt/librenms/html/ajax_form.php(26): include_once('/opt/librenms/i...')
#3 {main}
"} 
```

Also attempting to edit or browse to the url services/templates/1/edit etc produces a 403 forbidden (laravel message)

Also
if (! DB::raw('COUNT(service_id)')->from('services')->
Code Analyser : "The method ``from()`` does not exist on ``Illuminate\Database\Query\Expression``."

app/Models/ServiceTemplate.php
```diff
@@ -0,0 +1,58 @@
...
+    ];
+
+    // ---- Helper Functions ----
+
+    public function getCompositeKey()
+    {
+        return $this->id . '-' . $this->device_group_id;
+    }
+
+    public static function getServiceTemplate($query)
```
Jellyfrog, line 34: You don't need a scope function for a simple query. Use ::find($id) instead

includes/services.inc.php
```diff
@@ -108,13 +109,66 @@ function delete_service($service = null)
...
+    remove_service_template($service_template);
+
+    return dbDelete('service_templates', '`id` =  ?', [$service_template]);
+}
+
+function discover_service_template($service_template = null)
+{
+    $services_template = ServiceTemplate::getServiceTemplate($service_template);
+    $status = null;
+    foreach (DB::table('device_group_device')->where('device_group_id', $services_template[0]['device_group_id'])->pluck('device_id') as $device) {
```
Jellyfrog, line 133: Please use eloquent queries

includes/services.inc.php
```diff
@@ -108,13 +109,66 @@ function delete_service($service = null)
...
+        foreach (DB::table('services')->where('service_template_id', $service_template)->where('device_id', $device)->where('service_template_changed', '!=', $services_template[0]['changed'])->pluck('service_id') as $service) {
+            $update = ['service_desc' => $services_template[0]['desc'], 'service_ip' => $services_template[0]['ip'], 'service_param' => $services_template[0]['param'], 'service_ignore' => $services_template[0]['ignore'], 'service_disabled' => $services_template[0]['disabled'], 'service_template_id' => $services_template[0]['id'], 'service_name' => $services_template[0]['name'], 'service_template_changed' => $services_template[0]['changed']];
+            edit_service($update, $service['service_id']);
+            log_event("Updated Service: {$services_template[0]['name']} from Service Template ID: {$services_template[0]['id']}", $device, 'service', 2);
+            $status = 1;
+        }
+        if (! DB::raw('COUNT(service_id)')->from('services')->where('service_template_id', $service_template)->where('device_id', $device)) {
```
Jellyfrog, line 140: Service::where('service_template_id', $service_template)->where('device_id', $device)->count()

includes/services.inc.php
```diff
@@ -108,13 +109,66 @@ function delete_service($service = null)
...
+    }
+}
+
+function remove_service_template($service_template = null)
+{
+    if (! is_numeric($service_template)) {
+        return false;
+    }
+    $status = null;
+    foreach (DB::table('services')->where('service_template_id', $service_template)->pluck('service_id') as $service_id) {
```
Jellyfrog, line 160: Service::where('service_template_id', $service_template)->delete()

app/Models/ServiceTemplate.php
bofh80: removed function and replaced call with ::find . Thanks!

includes/services.inc.php
bofh80: I've updated this to what i think is correct. :)

includes/services.inc.php
bofh80: replaced, thank you!

includes/services.inc.php
bofh80: replaced, thanks!

bofh80:
Thanks for the review/fixes @Jellyfrog :+1:  I think i've corrected it as requested. I am no longer receiving any error logs when using the Discover / Remove buttons. (Well, a normal red warning saying nothing was changed).
Now i need to figure out why the discovery function isn't working as expected...
As for the problems above I believe that's all sorted except the 403 issue on edit.

bofh80:
The last function left in includes/services.inc.php is discover_service_template.
The Create works, but I need to fix checkbox's i think. And all the fields seems to be required, and i don't why.
Apply button works but reports no changes.
Remove button doesn't work. I moved to this a 'route' and function in app/Http/Controllers/ServiceTemplateController.php but I think that needs looked at.
Edit still produces a 403 error.
Delete doesn't work last I checked.

Scrutinizer / Inspection - Still a couple of reports for 'permissions' that I can't fix. Other than that it looks clear....

bofh80:
With regards to the Remove 'route'.  I'm not sure I understand the replationship between the web form service-templates/index.blade.php with the POST remove function(or even the DELETE function), and the ServiceTemplateController.php remove / delete function. Are they seperate or ?

app/Http/Controllers/ServiceTemplateController.php
```diff
@@ -0,0 +1,201 @@
...
+    /**
+     * Display a listing of the resource.
+     *
+     * @return \Illuminate\Http\Response|\Illuminate\View\View
+     */
+    public function index()
+    {
+        return view('service-template.index', [
+            'service_templates' => ServiceTemplate::orderBy('name')->get(),
+            'device_groups' => DeviceGroup::orderBy('name')->get(),
```
Jellyfrog, line 29: Add a relationship in the models between ServiceTemplate and DeviceGroup, then change to;
```suggestion
            'device_groups' => DeviceGroup::with('serviceTemplates')->orderBy('name')->get(),
```

resources/views/service-template/index.blade.php
```diff
@@ -0,0 +1,164 @@
...
+                                <th>@lang('Remote Host')</th>
+                                <th>@lang('Description')</th>
+                                <th>@lang('Modified')</th>
+                                <th>@lang('Ignored')</th>
+                                <th>@lang('Disabled')</th>
+                            </tr>
+                            </thead>
+                            <tbody>
+                                @foreach($service_templates as $service_template)
+                                    @if($service_template->device_group_id == $device_group->id)
```
Jellyfrog, line 46: With the changes above;
```suggestion
                                @foreach($device_group->serviceTemplates as $service_template)
```

bofh80:
@Jellyfrog I don't understand how the relationship call is supposed to work. It doesn't like 
```
public function serviceTemplates()
    {
        return $this->belongsToMany(\App\Models\ServiceTemplate::class, 'device_group_id');
    }
```

bofh80:
I corrected that back to the others and seems to have worked.
+        return $this->belongsToMany(\App\Models\ServiceTemplate::class, 'device_group_device', 'device_group_id', 'device_id');
Can you confirm that completes the changes?

bofh80:
The view doesn't appear to be correct. It's not showing the templates correctly. It's listing templates under multiple device groups.

bofh80:
@Jellyfrog so after some reading, it seems I might need to implement a new table for the device relationship? device_group_service_template with service_template_id and device_group_id ? Or am I completely misunderstanding? 
I tried various iterations of DeviceGroup::class, **'device_groups**', 'id', 'device_group_id'    as that seemed to make more sense,but no dice.

Jellyfrog:
No since you already have device_group_id in your service model you can use Belongsto().
And in device group model you can use hasMany

Jellyfrog:
I'll recommend reading https://laravel.com/docs/8.x/eloquent-relationships#one-to-many

bofh80:
@Jellyfrog that appears to have worked. Thanks very much. I was wondering about device / device_group relationships.

Do I need to implement a service_templates_perms table ? (i just reverted it as part of the other changes so I can easily pop it back in).

LibreNMS/Services.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /**
+     * List all available services from nagios plugins directory
+     *
+     * @return array
+     */
+    public static function list()
+    {
+        $services = [];
+        foreach (scandir(Config::get('nagios_plugins')) as $file) {
```
Jellyfrog, line 38: add a is_dir() check before, else you get an error if the path doesnt exist.

app/Http/Controllers/ServiceTemplateController.php
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param \App\Models\ServiceTemplate $template
+     * @param \App\Models\Device $device
+     * @return bool
+     */
+    public function storeservice(ServiceTemplate $template, Device $device)
```
murrant, line 94: This should be ServiceController::store()

haydenseitz:
this is awesome! looking forward to this

one thing I noticed is `service_changed` is being set to 0 when applied via a template. adding a service directly via a device page does update `service_changed` to current unix timestamp

bofh80:
@haydenseitz Thanks again for testing my code! :) .

I think  service_changed is only relevant once a service has been checked, and it will only stay on 0 until it's checked for the first time.
But i'm happy to track that down and fix it if someone thinks it's required. (I do believe I'd need to add it as a default for the database entry)

haydenseitz:
Here's the INSERT query when adding a service (not service template) 
```php
$insert = ['device_id' => $device['device_id'], 'service_ip' => $ip, 'service_type' => $type, 'service_changed' => ['UNIX_TIMESTAMP(NOW())'], 'service_desc' => $desc, 'service_param' => $param, 'service_ignore' => $ignore, 'service_status' => 3, 'service_message' => 'Service not yet checked', 'service_ds' => '{}', 'service_disabled' => $disabled];
```
Note `service_changed` is set to NOW(). You are right about the sql schema default - that field defaults to 0, which I think is wrong, but not the worst thing

bofh80:
@haydenseitz, it does default to 0 as part of my PR, as adding services from Service Template doesn't use the usual add_service call.
However, what I noticed (previously) was this made it really easy to spot services that have been added but not checked.
I was actually thinking of doing the reverse, and removing the NOW() from add_service.

Now in your list of 100's of services  you can see the ones that have been added (or terribly, never checked properly).

murrant:
Hey @bofh80 got some testing in:
* Error when you have no device groups.
* "Disable polling and alerting" is covered up by the button, you can change the bootstrap columns on the two toggles to fix this.
* Why do we pick a device group upon creation and not application?  What if I want to apply the same set of services to multiple groups?  What if I delete a device group associated with a template?

Other than that, it works great.  After these are addressed, I can do a code review.

bofh80:
i could just stuff it in the db as json apparently, instead of the extra table, but i'm not sure. I think i'll sleep on it.

bofh80:
> Hey @bofh80 got some testing in:
> 
> * Error when you have no device groups.
Fixed this so you can create a template and leave it assigned to nothing at all.

> * "Disable polling and alerting" is covered up by the button, you can change the bootstrap columns on the two toggles to fix this.
I just turned moved from no-wrap to wrap. It seemed simpler, looks okay i guess, and keeps everything all nice and lined up.

> * Why do we pick a device group upon creation and not application?
Now you can target device(s) or device group(s), which can be set to dynamic and target away. (i'm not sure the dynamic rules  are correct for targeting a device group, but I have no idea what to do about that)

> *  What if I want to apply the same set of services to multiple groups?  
Now you can apply service(s) to multiple devices or groups as required, and if the device is targeted multiple times by the same service only 1 will appear. (ie you won't get duplicates of the same check, assuming service_template_id matches).

> * What if I delete a device group associated with a template?
I added a [commit](https://github.com/librenms/librenms/pull/12107/commits/cdf9a4d201bd90c01c360df18b782bca21813ee8), i'm not sure it's still relevant or even correct after all the following changes, might be a better way to do it, i could look at what happens when i remove a device that's assigned to a device group . . .
 
> Other than that, it works great. After these are addressed, I can do a code review.
I finally got phpcs running locally (as the online one has apparently given up on me.) i can't see anything outstanding but it's a lot harder to look at in vscode. .

For anyone following, there was some additional conversation in discord that resulted in significant code enhancement for this. Now would be a *really* good time to test again to make sure I haven't introduced new and wonderful bugs, there are probably a few...... Reviews are welcome . . very welcome :)

ottorei:
For some reason I cant seem to be able to define device rules when using dynamic device type and static device group. I would like to use existing groups and just select all hosts from that group using hostname != null on the device rule. Static + Static seems to be working.

The error it gives when trying to select dynamic devices is this:
The drules field is required when dtype is dynamic.

murrant (approved):

rgelber:
I'm eagerly awaiting for this to be in master.

bofh80:
Error:
Illuminate\Database\QueryException: SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP 'service_templates_device_service_templates_id_foreign'; check that column/key exists (SQL: alter table `service_templates_device` drop foreign key `service_templates_device_service_templates_id_foreign`)

Migration:
https://github.com/librenms/librenms/pull/12107/files#diff-fc0d16a8d18f1fac8311b38497718e03b8908af04cd974c8f84127514c83e7fe

Schema:
https://github.com/librenms/librenms/pull/12107/files#diff-1f5c3c8a2fc231d774bb8ed7b29ca4743442b73d6acc29cf64bb2cb3dae132fe

I'm not sure how to fix this. I haven't got my dev box up atm to check the db with the migrations run. Anyone have an idea?

murrant (changes requested):
The schema error indicates you have an error in the down migration.

app/Http/Controllers/Select/ServiceController.php
```diff
@@ -18,8 +18,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
  * @link       http://librenms.org
- * @copyright  2018 Tony Murray
- * @author     Tony Murray <murraytony@gmail.com>
+ * @copyright  2020 Anthony F McInerney <bofh80>
```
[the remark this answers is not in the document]
bofh80: Yes, I noticed this the other day :D  . I had only made a copyright change to try and kick the travis build, and it was supposed to be done on the file below :+1:

app/Policies/ServiceTemplatePolicy.php
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Determine whether the user can view the service template.
+     *
+     * @param \App\Models\User $user
+     * @param \App\Models\ServiceTemplate $template
+     * @return mixed
+     */
+    public function view(User $user, ServiceTemplate $template)
+    {
+        return $this->viewAny($user) || Permissions::canAccessServiceTemplate($template, $user);
```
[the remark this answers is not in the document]
bofh80: I doubt i did anything intentionally, (sorry it's been 2+ months since i looked at this code properly). I imagine i might have used something else as an example or template, feel free to suggest a change if it doesn't seem right?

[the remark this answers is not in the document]
murrant: My suggestion is to remove per-user permissions for now as they don't add much.  At some point in the future, it will be possible to add fine grained user permissions.  If this exists as is the migration will be more complicated.

murrant (approved):
Design seems good to me.  We'll merge this and fix any issues that are discovered afterwards.

Issue #5660, opened 2017-01-29
Memory Health/Occupation Linux Machines: cached memory is not considered free/available


zxyz:
In LibreNMS unter **Health - Memory** the `Physical Memory` and `Virtual Memory`  all my Linux VMs Memory is marked **red**, due to a very high occupation. The VM named **monitor** below shows the following output:

> monitor	Physical memory      7.650GB / 7.800GB           151.520MB	98%
> monitor  Virtual memory        7.650GB / 7.800GB         151.520MB	98%

This is correct, when buffered and cached memory is regarded as occupied.

`free -h` on **monitor** shows:
> `                   total           used          free         shared   buff/cache   available`
> `Mem:           7.8G        1.8G        152M         86M        5.9G            5.6G`

So actually, **monitor** is quite happy with its memory. I would suggest to no use the `free` value but the `available` value for indication (on Linux machines).

Thanks for LibreNMS!

(latest LibreNMS, ubuntu 16.04)

laf:
Most likely a duplicate of #3179 please read it and see

zxyz:
oh, yes it is a dupicate. It flipped through my search. sry for that.

laf:
The other issue is closed as well. Does it answer your question as we've not changed how we do memory graphs since then afaik

zxyz:
@laf thanks for asking. It answered my question but I think I disagree. I think  I'll comment in the other issue.

Issue #3179, opened 2016-03-07
Linux - Memory usage confusion on Front page.


eikik87:
Firstly, I'm not the biggest linux know how man. But I always thought there was something wrong with my librenms server ( Ubuntu ). Cause it kept showing 95% - 100% physical memory usage. But behind in the shadow of librenms, I can see that it is just the cached memory being used in the RAM. Like Linux machines do, they use the RAM in a smart way. But I didn't know that. But maybe there is a way to display this is more decent way if it is Linux machine. Instead of showing Physical. It would show information more useful to alarm admins if there is something wrong. 

![image](https://cloud.githubusercontent.com/assets/13901674/13570461/413aea94-e466-11e5-81ae-b935bf0924a1.png)

murrant:
Hmm, this would be a different graph I think.  I know what I would call "free" memory: "Available Memory", but what would you call the "used" memory to make it clear?

geordish:
I was talking to @eikik87 regarding this issue on IRC, and I agree with him.

The data is already collected, and displayed in a different graph.

![graph](https://cloud.githubusercontent.com/assets/1226821/13575648/eabc9f48-e481-11e5-8df8-344ef26a38ac.png)

I believe what should actually be displayed as available is the second line marked as - Sh, Bu, Ca.

paulgear:
I think it would be a bad idea to change these graphs to indicate something other than what the SNMP data is returning.  That is one of the key principles we work from - unless the data is clearly wrong and can't be fixed at the source, we shouldn't mess with it.

With respect to Linux memory usage, not all of the memory marked as shared, buffers, and cache is available.  e.g. Sometimes memory is shared because there are multiple copies of an executable running and they can share identical code pages.  A certain minimal number of buffers & cache is necessary for reasonable I/O performance also.

The answer to misunderstanding the relationship of the different memory figures is education, not changing the graph.

murrant:
I agree with @paulgear on this one.

f0o:
Agree with @paulgear

laf:
Please feel free to re-open / comment

PR #12110, opened 2020-09-19
[WIP]: Memory: read memSysAvail from snmpd

ibigbug:
There has already been several discussions on Github and LibreNMS community regarding the memory usage graph

https://github.com/librenms/librenms/issues/3179
https://github.com/librenms/librenms/issues/5660

https://community.librenms.org/t/making-physical-memory-graphs-valid-and-worthwile/10333/9
https://community.librenms.org/t/supporting-memavailable-on-linux/13171?u=appleseed

based on this [comment](https://github.com/librenms/librenms/issues/3179#issuecomment-197155806), the "Available" memory couldn't be graphed due to lack of support from snmpd.

Now the feature has been [added](https://github.com/net-snmp/net-snmp/pull/167) to net-snmp, I think it makes sense to use the new "Available Memory" data to draw the memory graphs.

see also https://github.com/net-snmp/net-snmp/pull/167#issuecomment-678905462

Putting WIP as the commit was checked into net-snmpd code base but not yet released and the release date is not know yet. see: https://github.com/net-snmp/net-snmp/pull/167#issuecomment-683793493

Also I believe we need backward compatibility for this change - please suggest how it can be done best :)

#### Please note

image

image

image

#### Testers


murrant:

https://community.librenms.org/t/supporting-memavailable-on-linux/13171/16

Jellyfrog:
Have a look at travis, something is not working as intended, https://travis-ci.com/github/librenms/librenms/jobs/387338375

ibigbug:
@Jellyfrog it's breaking test due to lack of backward compatibility.

is there a function to test a particular OID/MIB supported by a agent to create a proper if/else ?

Jellyfrog:
I'm not really sure if there is a better way, fetch both and check if `memSysAvail` is available?

ibigbug:
> I'm not really sure if there is a better way, fetch both and check if `memSysAvail` is available?

Would fetching non-exists OID cause exception if so maybe just a simple try catch? otherwise only fetch them in batch and do a if-else

Jellyfrog:
> > I'm not really sure if there is a better way, fetch both and check if `memSysAvail` is available?
> 
> Would fetching non-exists OID cause exception if so maybe just a simple try catch? otherwise only fetch them in batch and do a if-else

Nah it doesnt as far as I know

murrant:

https://community.librenms.org/t/making-physical-memory-graphs-valid-and-worthwile/10333/18

ibigbug:
it seems fetching non-exist OID is returning nothing

```
SNMP['/usr/bin/snmpget' '-v2c' '-c' 'COMMUNITY' '-OQUs' '-m' 'UCD-SNMP-MIB' '-M' '/opt/librenms/mibs' 'udp:HOSTNAME:161' 'memTotalSwap.0' 'memAvailSwap.0' 'memTotalReal.0' 'memAvailReal.0' 'memSysAvail.0' 'memTotalFree.0' 'memShared.0' 'memBuffer.0' 'memCached.0']

memSysAvail.0: Unknown Object Identifier (Sub-id not found: (top) -> memSysAvail) 
```

and `var_dump($snmpdata[0]);` gives: `NULL`

any better way to do it?

murrant:
You should instead add this into the snmp_get_multi, then check if it is set, otherwise fallback to the existing behavior.

murrant:
I was thinking about this a bit more and it might be better as an additional field.

ibigbug:
> I was thinking about this a bit more and it might be better to add an additional field.

does that mean there would be a new line added to the generated memory graph?

ibigbug:
> You should instead add this into the snmp_get_multi, then check if it is set, otherwise fallback to the existing behavior.

This is exactly what I did locally and the output was found from the "Capture Debug Information"

it seems when added non-exists field, the snmp_get_multi returns nothing at all and no chance to check the output

murrant:
you mean you don't get the other oids?  It should still return them, but not memSysAvail.

I think it is because the code is doing the stupid list into variables thing.

Just check `isset($snmpdata[0]['memSysAvail'])` and it should tell you if it returned data or not.

includes/polling/ucd-mib.inc.php
```diff
@@ -134,16 +134,11 @@
 
 $snmpdata = snmp_get_multi($device, ['memTotalSwap.0', 'memAvailSwap.0', 'memTotalReal.0', 'memAvailReal.0', 'memSysAvail.0', 'memTotalFree.0', 'memShared.0', 'memBuffer.0', 'memCached.0'], '-OQUs', 'UCD-SNMP-MIB');
 
-if (isset($snmpdata[0]['memSysAvail'])) {
+if (is_array($snmpdata[0])) {
     [$memTotalSwap, $memAvailSwap, $memTotalReal, $memAvailReal, $memSysAvail, $memTotalFree, $memShared, $memBuffer, $memCached] = $snmpdata[0];
```
murrant, line 138: I suggest just removing the temporary variables and accessing the array by index instead.

ibigbug: sounds good to me.

one follow up question, in the case there is the additional memSysAvail field, how do we make the change to use it in the graph. I've take a look includes/html/graphs/device/ucd_memory.inc.php and it doesn't seem to have condition check, will it cause error adding a non-exist field to rrd_options?

mibs/UCD-SNMP-MIB
```diff
@@ -629,6 +629,21 @@ memUsedRealTXT OBJECT-TYPE
          pages from other uses of physical memory."
     ::= { memory 17 }
 
+memSysAvail OBJECT-TYPE
```
murrant, line 632: You need to update the whole mib, just adding the one oid won't work because CounterBasedGauge64 is not imported.

includes/polling/ucd-mib.inc.php
murrant: You can add it to the end of the rrd defintion and I think it will still update the over values fine.  Then only newly created mibs will get the value.

includes/html/graphs/device/ucd_memory.inc.php
```diff
@@ -14,6 +14,7 @@
 $rrd_options .= " 'DEF:ashared=$rrd_filename:shared:AVERAGE'";
 $rrd_options .= " 'DEF:abuffered=$rrd_filename:buffered:AVERAGE'";
 $rrd_options .= " 'DEF:acached=$rrd_filename:cached:AVERAGE'";
+$rrd_options .= " 'DEF:sysavail=$rrd_filename:sysavail:AVERAGE'";
```
ibigbug, line 17: @murrant please take a look at this file change, I have zero experience in rrd options :)

ibigbug: getting this ERROR: No DS called 'sysavail' in 'pve.tl.home/ucd_mem.rrd' from the `show RRD command`

murrant: because your ucd_mem.rrd doesn't contain a sysavail data set ;)  You haven't added it yet.  And to merge this, we'll need it to not error in this case.

includes/polling/mempools/hrstorage.inc.php
```diff
@@ -11,12 +11,25 @@
 }
 
 $entry = $storage_cache['hrstorage'][$mempool['mempool_index']];
-
 $mempool['units'] = $entry['hrStorageAllocationUnits'];
-$mempool['used'] = ($entry['hrStorageUsed'] * $mempool['units']);
+$has_avail = false;  // sysAvail based calculation when available
```
ibigbug, line 15: the unit test is still complaining after tying with this flag, any ideas? @murrant

includes/html/graphs/device/ucd_memory.inc.php
ibigbug: right, I was expecting `data_update($device, 'ucd_mem', $tags, $fields);` would update the `.rrd` which didn't.

could you please guide how to make the sysavail available in that `.rrd` ?

includes/polling/mempools/hrstorage.inc.php
murrant: $has_avail will not be set anywhere...
Ok, i think this is a reasonable work-around. Make a new RRD if sysAvail exists, check if the rrd exists in the graph and then add it to the graph.

murrant: orrr, simply replace "free" in the rrd with available if it exists... let me ask around on this one.

ibigbug: sure, I feel we are almost there..

murrant:
So, I've been re-writing mempools support and I can't figure out why memSysAvail is needed.
Findings:
HOST-RESOURCES-MIB support was not logging buffer, cached, and shared memory.
UCD-MIB was already storing those items

Once we have buffer and cached, we can calculate available: free + buffer + cached
We can also calculate unavailable ram: used - buffer - cached.
the UCD graph has some semblance of those calculations already

murrant:
I take it back.  Available != free + buffers + cached.  That is an approximation.

ibigbug:
Yep, it also helps in the Memory graph where it's usually almost 99% "used" which contains "Available" memory.

if we split that out, the memory graph and alert can also leverage on the new number.

murrant:
Fixed in #12282

PR #12127, opened 2020-09-21
Refresh in widget settings

Negatifff:

#### Please note

#### Testers


Negatifff:
and need db correction:
alter table users_widgets modify column refresh smallint(4) NOT NULL DEFAULT '60';

but I don't know how do this in patch

bofh80:
in:
misc\db_schema.yaml  (edit the existing schema for that table)
and:
database\migrations\newfile (copy one existing that fits the purpose and modify ).

That's if i understand correctly. :) I just checked other PR's for the same / similar.

app/Http/Controllers/Form/WidgetSettingsController.php
```diff
@@ -45,7 +45,10 @@ public function update(Request $request, $widget_settings)
             ]);
         }
 
+        $refresh=(!is_numeric($widget_settings["refresh"]) || (int)$widget_settings["refresh"] < 0 || (int)$widget_settings["refresh"] > 32767 ? 60 : $widget_settings["refresh"]);
```
murrant, line 48: This should be done in a validation rule above.

app/Http/Controllers/Form/WidgetSettingsController.php
```diff
@@ -45,7 +45,10 @@ public function update(Request $request, $widget_settings)
...
+        unset($widget_settings["refresh"]);
```
murrant, line 49: might as well unset `_token` too

app/Http/Controllers/Form/WidgetSettingsController.php
murrant: 'settings.refresh' => 'int|min:1|max:86400',

murrant:
I suggest not using the "refresh" field, just put it in the settings json.
I also noticed an error in many widgets (the reason you had to set refresh in many widgets)
Inside the getSettings() function they should call parent::getSettings()

app/Http/Controllers/Form/WidgetSettingsController.php
Negatifff: ok, thanks. but I can't understand - I set not int value and after pressing save button I see empty red alert in up right corner. what I haven't do yet?

murrant:
You won't have an empty refresh issue if you call parent::getSettings()

Negatifff:
sorry, but I do not quite understand where to do it

Negatifff:
@murrant thanks

murrant (approved):

murrant:

PR #12130, opened 2020-09-21
Fix to make vmware port discovery match port polling.

dagbdagb:
Ref. https://community.librenms.org/t/suppress-vmware-nic-change-logs/9888/2

The port polling on vmware got som magic going on, which makes the interface name/description appear 'changed' whenever the port discovery happens every 6 hours. This causes log spam for every esxi host. See above referenced forum post for an explanation.

This fix silenced our logs, but I do find it in bad taste to push an os-specific kludge into generic code like this. 
Is there a better way?

#### Please note

#### Testers


murrant:
This is the wrong fix.  Set port association mode to ifName

PR #12158, opened 2020-09-26
Use high precision OIDs for APC UPS sensors

CirnoT:
Use high precision OIDs for APC UPS sensors.

I have looked at the git history to understand why low precision OIDs were used, but in the end I have traced them being unchanged ever since initial code was written before the fork, with no comments or description as to why.

Note that the divisor for these is 10, which is still lower than what UPS reports internally (divisor 100), but it is way better (especially concerning battery voltage) than 1V precision we had before. There doesn't seem to be any other OID on PowerNET-MIB that would provide higher precision sadly.

#### Please note

#### Testers


PipoCanaja:
Hi @CirnoT 
Can we be sure that all APC devices support both the Low and the High prec OIDs ? If yes, then we are clear to remove the Low. If no, then we'd need to try the high prec, and if they are not available, fall back to low prec.

CirnoT:
I do not have any other units to test with.

I'd say we can be reasonably sure that all units should support it, as internally the UPS does report 10mV precision for battery voltage (it is requirement for UPS to even work properly), and that can be traced back to even ones as old as 1996 where PowerChute plus over RS232 was used to report that data.

murrant:
@CirnoT how would you feel about changing these to yaml?

CirnoT:
Not familiar or up-to-date with all the recent changes in LibreNMS

murrant:
basically all the fields in the discover_sensor have 1-1 mappings.  You also tell it what table to walk and what the numeric oid will be.

CirnoT:
The code seems to also loop over some OIDs for multi phase UPSes and also try to do similar things with PDUs, can this also be achieved somehow with YAML?

murrant:
so, what it seems to do is if there are multiple phases, use one oid, otherwise use another.  You can do this via the skip_values field checking upsPhaseNumInputPhases.1

Jellyfrog:
Thank you for your contribution but we're closing this due to inactivity.  
Feel free to reopen if you plan on finishing it!

CirnoT:
The code is very much ready, I don't exactly have time to rework it for YAML because I believe that it is outside of scope of what this PR wants to achieve.

Jellyfrog:
@CirnoT could you rebase this to fix merge conflicts?

Jellyfrog (approved):

PipoCanaja:
@CirnoT Seems that we already have issues here. We definitely need to keep existing sensors for existing users and add new sensors in //.

murrant:
We should probably revert this until it is fixed...

PR #12188, opened 2020-10-05
service watchdog - add systemd watchdog for resiliency

bofh80:
I had 2 poller failures, both fired watchdog restarts, yet failed to poll and log errors.
systemctl restart librenms fixed both. I assume shutdown will lead to a service restart (restart=always is set on the service)

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 12188`


murrant (commented):
This looks good to me, but I haven't tested.

Does systemd.daemon import make the code fail on systems it is not available on? (bsd for example)

mcineran:
The latest commits should take care of those systems without systemd.

This does create a dependency on python3-systemd (Ubuntu package name)

Also, I wonder if the librenms.service should be left alone and create a new one called librenms-watchdog.service maybe? (and they can copy the appropriate one).

As,  if you run this systemd service without python3-systemd it will restart every 2 minutes. (because of the change to the service file itself).

bofh80:
The latest commits should take care of those systems without systemd.

This does create a dependency on python3-systemd (Ubuntu package name)

Also, I wonder if the librenms.service should be left alone and create a new one called librenms-watchdog.service maybe? (and they can copy the appropriate one).

As, if you run this systemd service without python3-systemd it will restart every 2 minutes. (because of the change to the service file itself). 

nb: sorry for the double post noise.

bofh80:
I've updated the docs to include installing python3-systemd. I was hoping to lower the frequency from 2 minutes to say 30 seconds, in the unit file. I can do some testing, but i was wondering...
Is that sensible with regards to the code? (i'm not sure how often it will actually notify, multiple times a second or far less?)

willhseitz:
@bofh80 nice, this will be much better than checking with just the python dispatcher.
I applied this git patch to my dev instance and installed python3-systemd via apt, but the poller is restarting every 2 minutes. I assume that the systemd module is not being imported correctly. Should the systemd module be installed via pip and the librenms user?
FYI running Ubuntu 20

willhseitz:
Also, regarding your question about how often to notify systemd, when the watchdog code was first written, I used the poller frequency as default interval for the watchdog check the log file. This was mostly for the (probably rare) situation where a librenms instance only had a few devices, and they starting polling late. If at least one device is polled within the polling interval, the watchdog would restart the service.

Now, because we are just having the dispatcher process check in with systemd, we could have 2 different timers: one to check in with system at a much lower interval (every 10 seconds?), and another to check the polling log file every polling interval. What do you think?

bofh80:
@willhseitz, every 10 seconds or so sounds good, happy that idea, where in the code was the interval set?
I'll double check the patch again now, I did have to redo the patch at one point because a rebase went horribly wrong, I might have missed something, but i thought i tested it.

bofh80:
@willhseitz as per the pip comment, I also just use apt-get to install it system wide.

LibreNMS/service.py
```diff
@@ -285,11 +291,8 @@ def __init__(self):
         self._lm = self.create_lock_manager()
         self.daily_timer = LibreNMS.RecurringTimer(self.config.update_frequency, self.run_maintenance, 'maintenance')
         self.stats_timer = LibreNMS.RecurringTimer(self.config.poller.frequency, self.log_performance_stats, 'performance')
-        if self.config.watchdog_enabled:
-            info("Starting watchdog timer for log file: {}".format(self.config.watchdog_logfile))
-            self.watchdog_timer = LibreNMS.RecurringTimer(self.config.poller.frequency, self.logfile_watchdog, 'watchdog')
-        else:
-            info("Watchdog is disabled.")
+        info("Starting watchdog timer for log file: {}".format(self.config.watchdog_logfile))
+        self.watchdog_timer = LibreNMS.RecurringTimer(self.config.poller.frequency, self.logfile_watchdog, 'watchdog')
```
willhseitz, line 295: @bofh80 here is where the watchdog_timer is set to self.config.poller.frequency

willhseitz:
> 
> 
> @willhseitz as per the pip comment, I also just use apt-get to install it system wide.

I confirmed that systemd.daemon is being imported , but the systemd service is still restarting

bofh80:
@willhseitz, I've updated the code as per your suggestions. it does look much cleaner like this. I haven't test it yet. How does it look? (as per your issues above, are you actually polling at the default 5 minute interval?)

bofh80:
@willhseitz I've tested the code cleanly against 1.69 (master seems to have a broken change). It appears to work fine. Can you confirm?

willhseitz:
> are you actually polling at the default 5 minute interval?
My dev environment is set to poll at 1 minute

> I've tested the code cleanly against 1.69
I'm on the daily channel, so not a "clean" install. I'll keep testing and see if I find anything helpful

haydenseitz:
@bofh80 tested with your new changes and it's working as expected, no incorrect restarts

bofh80:
@haydenseitz thanks very much for testing. I also set everything to poll at minute intervals currently. What do you think the ideal setting should be for the watchdog to restart the service? 30 seconds, more?

haydenseitz:
I havent worked with systemd service stuff before, so I wouldn't know. Googling around shows others setting it to 10-30 seconds. 
https://www.medo64.com/2019/01/systemd-watchdog-for-any-service/
https://unix.stackexchange.com/questions/500115/read-systemd-watchdogsec-from-environmentfile

I set WatchdogSec to 30s and left the 10 second notify, it seems to be working fine

haydenseitz:
Here's the INSERT query when adding a service (not service template) 
```php
$insert = ['device_id' => $device['device_id'], 'service_ip' => $ip, 'service_type' => $type, 'service_changed' => ['UNIX_TIMESTAMP(NOW())'], 'service_desc' => $desc, 'service_param' => $param, 'service_ignore' => $ignore, 'service_status' => 3, 'service_message' => 'Service not yet checked', 'service_ds' => '{}', 'service_disabled' => $disabled];
```
Note `service_changed` is set to NOW(). You are right about the sql schema default - that field defaults to 0, which I think is wrong, but not the worst thing

bofh80:
@murrant I have moved the systemd.service changes into a new file. This means that unless people explicitly copy the librenms-watchdog.service instead the code will have NO effect.

So it should be pretty safe to merge with a bit of testing.

doc/Installation/Installation-CentOS-7-Apache.md
```diff
@@ -16,7 +16,8 @@ yum install epel-release
 ```
 
 ```
-yum install git cronie fping jwhois ImageMagick mtr MySQL-python net-snmp net-snmp-utils nmap python-memcached rrdtool policycoreutils-python httpd mariadb mariadb-server unzip python3 python3-pip
+yum install git cronie fping jwhois ImageMagick mtr MySQL-python net-snmp net-snmp-utils nmap python-memcached rrdtool policycoreutils-python httpd mariadb mariadb-server unzip python3 python3-pip
```
murrant, line 19: We generally do not update retired documents.

misc/librenms-watchdog.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
```
murrant, line 1: This is a good compromise.

doc/Installation/Installation-CentOS-7-Apache.md
bofh80: Removed those updates.

misc/librenms-watchdog.service
bofh80: Yeah, hopefully this should avoid any issues introducing the code.
Thanks for the review!

ottorei:
I have been testing this on a distributed installation with 4 pollers and have not seen any issues yet.

twelch24:
Testing on a ubuntu 20.04.1 system, python3.8.5, 3 pollers and ~1500 devices has revealed no issues with this. I have not yet seen a service crash with it though to require an actual restart. I am using librenms-watchdog.service.

twelch24:
I have seen some service restarts now with this patch. Seems to work as it should - NO more graph gaps.

PR #12192, opened 2020-10-05
Ircbot fix

robje:

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/IRCBot.php
+++ b/LibreNMS/IRCBot.php
@@ -292,6 +292,7 @@ class IRCBot
 
     private function sendAlert($sendto, $severity, $alert)
     {
+	$sendto = explode(' ', $sendto)[0];
         $this->ircRaw('PRIVMSG ' . $sendto . ' :' . $severity . trim($alert['title']));
         if ($this->config['irc_alert_short']) {
             // Only send the title if set to short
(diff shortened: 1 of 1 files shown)


murrant:
Are you sure this isn't already fixed in master?  Because there was a bug in the 1.68 release.

robje:
> Are you sure this isn't already fixed in master? Because there was a bug in the 1.68 release.

Nope, this was still borken after 1.68 release.

A better solution might be to use sendMessage instead of ircRaw. Though this solves the problem for me (tm)

murrant:
@Olen thoughts?

Olen:
He is right.  I just have not used a key on my channel, so it has never been an issue here.
But as the channel + key is stored as a single, space seperated entry in the config:
`irc_alert_chan = [ '#channel key', '#channel_wouthout_key']` it makes sense to splt on the space and use only the first part as the channel name in the alerts. An irc-channel can never have a space in its name anyway.

Another option would be to restructure the config to so the key can be stored as a seperate entity.  Something along the lines of
`irc_alert_chan_keys['#channel'] = 'key'` 
And then add a check on join to see if there is a key defined, but I don't think it gains much value.

Jellyfrog (approved):
@robje Tips, next time explain what it actually fixes, belive it or not, we don't know every single thing about the code :)

robje:
> @robje Tips, next time explain what it actually fixes, belive it or not, we don't know every single thing about the code :)

sure, looking back I see I missed adding some explanation. Somehow I skipped reading the first sentence.

murrant:

PR #12263, opened 2020-10-27
Update to Composer 2

Jellyfrog:
No need to roll our own checks anymore since this is now built into Composer 2
Also limit the update check to once a week, it's a bit aggressive now :)

#### Please note

#### Testers


Jellyfrog:
Examples;
>Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 7.3.0". You are running 7.2.33-1+ubuntu18.04.1+deb.sury.org+1.
  
>Composer detected issues in your platform: Your Composer dependencies require the following PHP extensions to be installed: zip

murrant:
This looks good, but I haven't tested.
One thing is this will only check when ./daily.sh is run.  Previous code checks on application startup.  Not sure if we care about that.

Jellyfrog:
> This looks good, but I haven't tested.
> One thing is this will only check when ./daily.sh is run. Previous code checks on application startup. Not sure if we care about that.

No, it's in the autoloader. The above messages was taken from the site.

doc/General/Updating.md
```diff
@@ -22,7 +22,7 @@ you can do so by running the following commands:
 ```bash
 cd /opt/librenms
 git pull
-composer install --no-dev
+./scripts/composer_wrapper.php install --no-dev
 ./build-base.php
```
murrant, line 26: This line could use an update too :D

scripts/composer_wrapper.php
```diff
@@ -55,8 +55,13 @@
...
 
-    // self-update
-    passthru("$exec self-update -q" . $extra_args);
+    // If older than 1 week, try update
+    if (time() - filemtime($install_dir . '/composer.phar') > 60 * 60 * 24 * 7) {
+        // self-update
+        passthru("$exec self-update --quiet --2" . $extra_args);
+    }
+
+    @touch($install_dir . '/composer.phar');
```
murrant, line 64: Shouldn't this be inside the if statement?

murrant:
When testing, composer install results in error output, but daily.sh shows success.

scripts/composer_wrapper.php
Jellyfrog: Yes.

doc/General/Updating.md
```diff
@@ -23,7 +23,7 @@ you can do so by running the following commands:
 cd /opt/librenms
 git pull
 ./scripts/composer_wrapper.php install --no-dev
-./build-base.php
+php includes/sql-schema/update.php
```
murrant, line 26: oh, I guess people could still be on an age old version, but at least we could remove the. build-base.php script.

Jellyfrog: I'll do some following up cleanup later

Jellyfrog:
> When testing, composer install results in error output, but daily.sh shows success.

What error? Due to Composer 2?  
That it continues is known and a feature, it's so the daily cleanup runs anyway (we should split those into it's own job a sunny day)

Jellyfrog:
I added a commit to make the script at least exit with the right code, now it should *show* as FAIL (but it will continue)

murrant (approved):
They dialed back the default behavior to exclude extensions very recently.
Added change to re-enable that.
https://github.com/composer/composer/issues/9412

Testing went well otherwise.

murrant:

PR #12300, opened 2020-11-06
Adding Fortigate HA checks

epacke:
We discovered that one of our firewall nodes in a cluster had failed but it was not detected by LibreNMS.

This PR aims to fix this issue by using the following logic:
1. Is the monitored device clustered in any way?
2. If so, make sure the device count in fgHaStats array is not 1

**Credits to Elias in discord for these:**
Also adding node state (master, backup and standalone) together with config sync state to ensure that the cluster config has been synced properly.

#### Please note

#### Testers


epacke:
It is a bit unclear to me how to write the tests. I followed the instructions here:

Ran the fortigate os checks and it passed but I don't really see any other unit tests that I can look at to copy. Haven't written PHP code in ~12 years so going through the whole unit test framework seems like a bit of an insurmountable task without further guidance.

epacke:
### Cluster with sync problems
image

From the Fortigate web interface:
image

### Cluster with a failed member
image

From the web interface (note how the system itself reports the device to be synchronized)
image

### A functioning clustered device:
image

image


### A standalone device
Nothing to show because the cluster state polling is not done.

includes/polling/sensors/state/fortigate.inc.php
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @link       http://librenms.org
+ * @copyright  2020 Net Entertainment AB
+ * @author     Patrik Jonsson <patrik.jonsson@gmail.com>
+ */
+if ($device['os'] == 'fortigate') {
+    if (in_array($sensor['sensor_type'], ['clusterState', 'haSyncStatus'])) {
+        // Fetch the cluster members and their data
+        $fgHaStatsEntryOid = '.1.3.6.1.4.1.12356.101.13.2.1.1';
+        $haStatsEntries = snmpwalk_cache_multi_oid($device, $fgHaStatsEntryOid, [], 'FORTINET-FORTIGATE-MIB');
```
murrant, line 28: You are using the MIB here, you might as well use the textual oid instead of the numeric oid.

murrant:
Code looks good overall,you could possibly utilize the user_func on sensors to created this as a yaml sensor with just a bit of php code to transform the data to a state value.  (I'm fine with it either way, but yaml is easier to maintain)

You don't need to write tests for modules, just capture data.  First you need to run `./scripts/collect-snmp-data.php -h <HOSTNAME> -v 1500d`  (it is probably ok to add the data to an existing variant in this case, if you think a different one would be better, change the -v value) then after the snmprec is updated.  Just run `./scripts/save-test-data.php -o fortigate -v 1500d` to update the test data.

includes/polling/sensors/state/fortigate.inc.php
epacke: I tried that first but got `false` as a result of the walk. Will try again.

epacke:
> Code looks good overall,you could possibly utilize the user_func on sensors to created this as a yaml sensor with just a bit of php code to transform the data to a state value. (I'm fine with it either way, but yaml is easier to maintain)
> 
> You don't need to write tests for modules, just capture data. First you need to run `./scripts/collect-snmp-data.php -h <HOSTNAME> -v 1500d` (it is probably ok to add the data to an existing variant in this case, if you think a different one would be better, change the -v value) then after the snmprec is updated. Just run `./scripts/save-test-data.php -o fortigate -v 1500d` to update the test data.

I think I'll leave it as is if it's ok with you but I'm curious if there any documentation that covers the user_func? Might be handy the next time. I'm an F5 guy myself and I might spend some time writing additional checks for them later on.

includes/polling/sensors/state/fortigate.inc.php
murrant: try `fgHaTables`

murrant:
https://docs.librenms.org/Developing/os/Health-Information/#advanced-health-discovery

Only problem is it only passes the value returned by snmp.  It might be appropriate to add a second argument including the sensor.

murrant:
@epacke you forgot to git add the snmprec file :)

epacke:
The final output of `./scripts/save-test-data.php -o fortigate -v 1500d` says:

```Saved to /opt/librenms/tests/data/fortigate_1500d.json
Ready for testing!
```

The snmprec file does not seem to have updated. Did I miss something?
```
/opt/librenms/tests $ ls -alh ./snmpsim/fortigate_1500d.snmprec
-rw-r--r-- 1 nobody nogroup 1.1M Nov  3 01:16 ./snmpsim/fortigate_1500d.snmprec
```

epacke:
Maybe something funky with my docker setup. I'll try running it in our Libre installation.

epacke:
There's like 23000 lines of data in that file and it's riddled with stuff I should not leave here on Github... I assume the rec file is the simulated SNMP data and the JSON is the expected output and it's gonna take me weeks to clean them.

epacke:
Might be able to do it with a script. Hang on, don't give up on me. :)

epacke:
Started but realized that the current data you guys have will be destroyed since my operation merged. Also, it was much tougher than I thought to anonymize the data. Just spent a few hours on it and the JSON is easy enough to anonymize but making the snmprec match was just a pain in the rear.

Suggesting three possible options forward:
1. Injecting the necessary oids into the SNMP rec file and adding the new data to the json file
2. Skip the new test data
3. We keep the added code as a local patch in our company and close this PR :(

What do you think?

Jellyfrog:
Make another snmprec , (the `-v` option), and use the `-m` option to only include the module you need, less to filter

Jellyfrog:
@epacke snmprec data needs to match the format specified, see https://docs.librenms.org/Developing/os/Test-Units/#snmprec-format for more info.  

Pro tip, You should only update the `.snmprec` file and then run the save-test-data.php script to generate a new json file

epacke:
Saw your tip just now. I'll let Travis shame me one more time. Then I'll try again. :)

epacke:
Heard that we have a common acquaintance btw (Johan S). Small world.

epacke:
Finally. Do you want me to squash or will you do it via Github?

murrant (approved):
LGTM, nice work :)

murrant:

PR #12308, opened 2020-11-08
Fix disabling of built-in commands when using ./lnms

Jellyfrog:
It wasn't working correctly after Laravel upstream changes

#### Please note

#### Testers


murrant:
It doesn't seem this is working right either.  lnms list shows a lot of Laravel commands still.

murrant:
https://github.com/nunomaduro/laravel-console-summary maybe?

Jellyfrog:
I Know, it's due to what I wrote in the comment. I'll look into that package.

Jellyfrog:
I think we should remove all commands *not* defined by us,
If a user needs to do more advanced stuff, they are own their own and should use artisan
```
$ ./lnms

  LibreNMS  1.69-8-gaefe5d466

  USAGE: artisan <command> [options] [arguments]

  scan                     Scan the network for hosts and try to add them to LibreNMS
  update                   Update LibreNMS and run maintenance routines

  config:get               Get configuration value
  config:set               Set configuration value (or unset)

  dev:check                LibreNMS code checks. Running with no options runs all checks

  device:add               Add a new device
  device:remove            Remove a device
  device:rename            Rename a device, this can be used to change the hostname or IP of a device

  poller:alerts            Check for any pending alerts and deliver them via defined transports
  poller:billing           Collect billing data
  poller:billing-calculate Run billing calculations
  poller:discovery         Discover information about existing devices, defines what will be polled
  poller:ping              Check if devices are up or down via icmp
  poller:poll              Poll data from devices as defined by discovery
  poller:services          Update LibreNMS and run maintenance routines

  smokeping:generate       Generate a configuration suitable for use with smokeping

  user:add                 Add a local user, you can only log in with this user if auth is set to mysql
```

murrant:
I agree.  But, do we want to hide the commands (and allow them to be run still) or prevent them from being used through the lnms cli??

Jellyfrog:
This still allows commands to be runned, I think it's fair enough

murrant (approved):
one tiny fix :) looks good

murrant:
`USAGE: artisan <command> [options] [arguments]` -> `USAGE: artisan <command> [options] [arguments]`

murrant:

PR #12321, opened 2020-11-17
Added devices unpolled alert rule template

Senetus:
Added devices unpolled alert template

#### Please note

- [x ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

### DIFF
--- a/misc/alert_rules.json
+++ b/misc/alert_rules.json
@@ -516,7 +516,7 @@
     "name": "QNAP NAS has a bad SMART disk status"
   },
   {
-    "rule": "UNIX_TIMESTAMP(NOW())- UNIX_TIMESTAMP(devices.last_polled) > (select config_value from config WHERE config_name=\"rrd.heartbeat\")",
+    "rule": "UNIX_TIMESTAMP(NOW())- UNIX_TIMESTAMP(devices.last_polled) > 300",
     "name": "Device unpolled"
   }    
 ]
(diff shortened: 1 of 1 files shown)


misc/alert_rules.json
```diff
@@ -514,5 +514,9 @@
   {
     "rule": "devices.os = \"qnap\" && sensors.sensor_type = \"diskSmartInfo\" && sensors.sensor_current != \"0\"",
     "name": "QNAP NAS has a bad SMART disk status"
-  }
+  },
+  {
+    "rule": "UNIX_TIMESTAMP(NOW())- UNIX_TIMESTAMP(devices.last_polled) > (select config_value from config WHERE config_name=\"rrd.heartbeat\")",
```
murrant, line 519: This needs a default value (coalesce) because rrd.heartbeat is not always set in the database.  Also, it will not take into account if this is set in config.php, which is probably ok.

Senetus: Ah, good point. I grabbed it first from the poller_cluster_stats but that wouldn't work with cron I think. Is there a place in the db where the polling interval is always set? Or is it just better to "hardcode" the interval?

murrant: there is no place in the db, which is why you need to use coalesce and set the second value to 300.  that way if set in the DB (which should be the case) it will override the 300.

murrant (approved):
Looks good, thanks!

Senetus:
Aaah there is such a mysql function, didn't even know that :D Thanks!

murrant:

PR #12330, opened 2020-11-19
Install docs addition

p4k8:
Added nginx route config for API to make it work right after install.

#### Please note

#### Testers

### DIFF
--- a/doc/Installation/Install-LibreNMS.md
+++ b/doc/Installation/Install-LibreNMS.md
@@ -237,6 +237,9 @@ Feel free to tune the performance settings in librenms.conf to meet your needs.
          location ~ /\.(?!well-known).* {
           deny all;
          }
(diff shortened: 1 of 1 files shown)


murrant (changes requested):
Why are you adding this?  It is absolutely not needed. `api_v0.php` is not needed and only exists to allow old installs to continue to function.  It is an alias of index.php.

p4k8:
Without this change, API calls on a fresh install get redirected to login page. Problem isn't mine alone, as I also seen mentions of this in Discord.
So some kind of change is still needed, I will research it later and edit PR accordingly.

murrant:
what webserver and webserver layout are you using?

murrant:
I did notice when you fail auth it redirects to login page, it didn't used to do that...

PR #12360, opened 2020-11-26
Add support for SCS KS air-conditioning Devices

Torch09:
WIP: Adds support for SCS KS air-conditioning Devices.

These Devices are using a Carel Controller (CAREL cpCO) but not the pCOWeb-Card.

#### Please note

#### Testers


Torch09:
I have some questions remaining before I can finish the pull request, hope someone of you with more experience in this field can help me.

1. 
I had to write a separate MIB-File because it uses the Carel Enterprise ID, but has other OIDs which are not defined in the Carel MIB. Because the beginning of the OIDs does overlap with the Carel-Vendor-MIB, I don’t know if I have to look out for something to not break the original MIB. Does the new MIB look ok or do I have to modify something?

2.
I added the device to LibreNMS and the discovery works fine, only the last OID returns:
“SNMP['/usr/bin/snmpbulkwalk' '-v2c' '-c' 'COMMUNITY' '-OteQUs' '-Ih' '-m' 'SCS-ks-series-MIB' '-M' '/opt/librenms/mibs:/opt/librenms/mibs/carel' 'udp:HOSTNAME:161' 'status-collective-fault.0']
status-collective-fault.0 = No more variables left in this MIB View (It is past the end of the MIB tree)”

If I run a “snmpget” it returns the right number. Is that a problem with my MIB-File, or is this a problem on the device itself?

Thanks for your help!

Torch09:
The same happens with Numeric OIDs. 
It seems like this is a problem with the SNMP service running on the device.
I'll see if this can be fixed on the Device. Every other values is returned correctly.

Jellyfrog:
Wondering if scs-ks-series is such a good OS name? Maybe skip series or name it carel cpco or something?

Also, is there no Oid for hardware, version, etc?

Torch09:
Sure, I can do that. 😄 
I would name it scs-ks, cpco would be confusing with the "real" cpco + pco Web hardware.

AFAIK, yes.
It returns the OIDs from the cpco controller and there are no OIDs for hardware, ver, etc.
The returned values are mostly Network + Basic things like Name, Location, etc.

Torch09:
Finished the renaming.
Now the devices are called scs-ks.

I've added the Interface Modules, forgot to include these in the definitions.

Jellyfrog (approved):

PR #12366, opened 2020-11-28
First attempt to add Printers Counters

zozolito:
Add support for Konica-Minolta BizHub basic Counters


zozolito:
sorry, i'm newbie

PR #12371, opened 2020-12-01
KYOCERA, KONICA-MINOLTA and HP JetPrinter Counters

zozolito:
Please excuse me if I miss some protocol.

Just wanted to add my first contribution.

Theses files are to add KYOCERA, KONICA-MINOLTA and HP JetPrinter counters support.

Feel free to check and comment this work.
Greetings.

image


zozolito:
@murrant, @PipoCanaja 
Hi guys, may you show me how to go further than this.
Files comittee works very well for me, but maybe my childish style writing code is not good for the project.
Don't want to put a mess in the awesome of librenms.

Thank you.

PipoCanaja:
Hi @zozolito 
You can start by checking the Continuous Intergration tests below, that are failing right now. Just code style rules for the project.

zozolito:
> Hi @zozolito
> You can start by checking the Continuous Intergration tests below, that are failing right now. Just code style rules for the project.

ok thank you, got some weird issues with github always add one empty line at the end of the file. Probably because I edit files directly on the web 🤷‍♂️😁

btw it looks good.
Thank you for your help 🤙

zozolito (commented):
Sorry for the mess, this is my first commit 😬
I don't even know if this is the place to write down what I'm going to write. I won't apologize anymore after that line 😅

You'll see a lot of oids in the comments, they're here for memory and to show where it was taken from the snmpwalk. As I have only a limited number of different models, some of you may have more Oids to add.

Also in the konica.inc.php, lines in the array have been commented, because I found them in double. As I know vendors don't add things in double for no reason, I let theses lines in the code just in case someone is in trouble and want to try alternate oids for some counters.

Thank you for allowing me to add my contribution to this project 😁

Jellyfrog:
You should convert this to the new yaml based format, see for example;
https://github.com/librenms/librenms/blob/f2f169ae78ab6e50794c1cfe5776c75894aa5c9a/includes/definitions/discovery/smartax.yaml#L100-L115
And more info here; https://docs.librenms.org/Developing/os/Health-Information/

Jellyfrog (changes requested):
You should convert this to the new yaml based format, see for example;
https://github.com/librenms/librenms/blob/f2f169ae78ab6e50794c1cfe5776c75894aa5c9a/includes/definitions/discovery/smartax.yaml#L100-L115
And more info here; https://docs.librenms.org/Developing/os/Health-Information/

The `group` function will come in handy here to group some of these values together under a common title

Jellyfrog:
Also, cant you get MIBs for these?

zozolito:
I tried the "yaml" way before, but was unable to find Konica Minolta bizhub MIBs, so I had to go the "php" way. Then I wrote the same code for other printers I have here.

murrant:
I'm a bit on the fence about whether these are fine as generic "count" sensors or whether they should be printer specific data.  (also, the "toner" poller module needs to be renamed printer because it polls more than toner)

joseUPV:
We are interested in this Pull Request and can do tests too.

murrant:
@joseUPV ideally these are added to the "toner" poller modules which needs to be renamed to "printer".  And given custom display in the UI.

murrant:
@zozolito Thought about this a little more and it is ok to use the count sensor.  Sure it would be great to have a printer specific section with nice graphs, but this is fine.

To get this ready to merge please add the group variable to the counters (this will put them in sections in the webui) and use snmp_get_multi_oid() instead of many individual snmp_get calls.

murrant:
Closing due to inactivity, feel free to update and reopen.

Issue #12205, opened 2020-10-11
"Sub-directory Support" does not accept top-level (root) URL


candlerb:
Summary: when configured following the sub-directory support instructions, paths like `/librenms/devices` and `/librenms/validate` work fine, but the top level path `/librenms/` gives a 404 error.

To reproduce:

* I created a fresh Ubuntu 20.04 lxd container
* I installed LibreNMS git head (2793585c) exactly as per the [Install-LibreNMS](https://docs.librenms.org/Installation/Install-LibreNMS/) page, using Apache rather than Nginx.
* I went through the database setup pages, then validate, then added `localhost` device.  LibreNMS is working fine at this point.
* I then modified the configuration as per [Sub-directory Support](https://docs.librenms.org/Extensions/Sub-Directory/)

```
--- /etc/apache2/sites-available/librenms.conf.old	2020-10-11 17:40:42.714705144 +0000
+++ /etc/apache2/sites-available/librenms.conf	2020-10-11 17:47:59.622870362 +0000
@@ -1,8 +1,10 @@
+AllowEncodedSlashes On
+Alias /librenms /opt/librenms/html
+
 <VirtualHost *:80>
   DocumentRoot /opt/librenms/html/
   ServerName  librenms.example.com

-  AllowEncodedSlashes NoDecode
   <Directory "/opt/librenms/html/">
     Require all granted
     AllowOverride All
```

```
--- /opt/librenms/html/.htaccess.old	2020-10-11 17:45:44.213581077 +0000
+++ /opt/librenms/html/.htaccess	2020-10-11 17:45:51.393649471 +0000
@@ -20,7 +20,7 @@
 </IfModule>

 RewriteEngine on
-RewriteBase /
+RewriteBase /librenms
 RewriteCond %{REQUEST_FILENAME} !-f
 RewriteCond %{REQUEST_FILENAME} !-d
 RewriteCond %{REQUEST_URI} !\.(js|ico|txt|gif|jpg|png|css|php)
```

```
--- /opt/librenms/.env.old	2020-10-11 17:45:25.301400911 +0000
+++ /opt/librenms/.env	2020-10-11 17:45:58.561717747 +0000
@@ -5,5 +5,5 @@
 DB_USERNAME=librenms
 DB_PASSWORD=password

-#APP_URL=
+APP_URL=/librenms/
 NODE_ID=5f83401f8b0fa
```

```
--- /opt/librenms/config.php.old	2020-10-11 17:44:58.193142622 +0000
[...]

PR #12391, opened 2020-12-10
 WIP: Fix "Sub-directory Support"

mpikzink:
This changes works im my Apache env.

1. The Double prefix from #12205 @candlerb is fixed with the changes in Url.php

2. I had several problems with leading "/" in href.

3. The root-Level URL from #12205 is fixed with web.php

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Util/Url.php
+++ b/LibreNMS/Util/Url.php
@@ -269,7 +269,7 @@ class Url
     {
         $vars = array_merge($vars, $new_vars);
 
-        $url = url(Config::get('base_url', true) . $vars['page'] . '');
+        $url = url($vars['page']);
         unset($vars['page']);
 
         return $url . self::urlParams($vars);
--- a/LibreNMS/Util/Url.php
+++ b/LibreNMS/Util/Url.php
@@ -306,7 +306,7 @@ class Url
             $urlargs[] = $key . '=' . urlencode($arg);
         }
 
-        return '<img src="' . url('graph.php') . '?' . implode('&amp;', $urlargs) . '" style="border:0;" />';
+        return '<img src="' . 'graph.php' . '?' . implode('&amp;', $urlargs) . '" style="border:0;" />';
     }
 
     public static function graphPopup($args, $content = null, $link = null)
--- a/LibreNMS/Util/Url.php
+++ b/LibreNMS/Util/Url.php
@@ -269,7 +269,7 @@ class Url
     {
         $vars = array_merge($vars, $new_vars);
 
-        $url = url($vars['page']);
+        $url = url($vars['page'] . '');
         unset($vars['page']);
 
         return $url . self::urlParams($vars);
--- a/LibreNMS/Util/Url.php
+++ b/LibreNMS/Util/Url.php
@@ -309,33 +309,33 @@ class Url
         return '<img src="' . url('graph.php') . '?' . implode('&amp;', $urlargs) . '" style="border:0;" />';
     }
 
-    public static function graphPopup($args)
+    public static function graphPopup($args, $content = null, $link = null)
     {
         // Take $args and print day,week,month,year graphs in overlib, hovered over graph
         $original_from = $args['from'];
         $now = CarbonImmutable::now();
 
-        $graph = self::graphTag($args);
-        $content = '<div class=list-large>' . $args['popup_title'] . '</div>';
-        $content .= '<div style="width: 850px">';
+        $graph = $content ?: self::graphTag($args);
+        $popup = '<div class=list-large>' . $args['popup_title'] . '</div>';
+        $popup .= '<div style="width: 850px">';
         $args['width'] = 340;
         $args['height'] = 100;
         $args['legend'] = 'yes';
         $args['from'] = $now->subDay()->timestamp;
-        $content .= self::graphTag($args);
+        $popup .= self::graphTag($args);
         $args['from'] = $now->subWeek()->timestamp;
-        $content .= self::graphTag($args);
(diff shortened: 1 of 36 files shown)


ThomasCr:
Hi, there is still a Bug with "Sub-directory Support" after your changes.
When I go to Ports view and select "All-Ports" and hower on tehat page over the port col the grafs are not shown.

grafik

resources/views/locations.blade.php
```diff
@@ -87,7 +87,7 @@
                 url: "{{ url('/ajax/table/location') }}",
```
ThomasCr, line 87: /ajax looks also hard-coded

mpikzink:
> Hi, there is still a Bug with "Sub-directory Support" after your changes.
> When I go to **Ports** view and select **All-Ports** and hower on that page over the **Port** column the grafs are not shown.
> 
> grafik
> 
> Same thing on **Overview** -> **IPv4 Address**, when I hover over the Interfaces
> 
> grafik
> 
> There is the subdirectory lost. - All other pages / graphs (also the mouse over ones) are working.
> 
> Also the double /librenms/librenms is still present on devices and ports view...? Do we need to flush caches somewhere?
> What I already tested is: `rm storage/framework/views/*.php`
> 
> I found also this with: `grep -R -e '/devices' -e '/ports' . | grep -v -e routes-v7 -e mib -e '/doc'` from the install directory
> 
> ```
> diff --git a/includes/html/dev-groups-overview-data.inc.php b/includes/html/dev-groups-overview-data.inc.php
> index 4a450ac0f..1cc23f603 100644
> --- a/includes/html/dev-groups-overview-data.inc.php
> +++ b/includes/html/dev-groups-overview-data.inc.php
> @@ -18,7 +18,7 @@ if (count($device_groups)) {
>                          <div class="col-sm-12">
>                          <?php foreach ($device_groups as $group) { ?>
>                              <span style="margin: 8px;">
> -                                <a href="/devices/group=<?=$group['id']?>" target="_blank"><?=$group['name']?></a>
> +                                <a href="devices/group=<?=$group['id']?>" target="_blank"><?=$group['name']?></a>
>                              </span>
>                          <?php } ?>
>                          </div>
> ```
> 
> there are also some more things in api.php and BasicApiTest.php which are also not looking so well.

murrant:
Can you add `dd(url()->current(), url('/'), url(''));` to a page and tell me what it looks like?

mpikzink:
Screenshot_2020-12-16_16-20-45
I think url() / Laravel is not the problem. I think the problem coms just with hard coded links in html. My last commits didn't change thinks in my setup.

mpikzink:
Subdir_ok
Without leading / everything is fine.

murrant:
So, seems like the fix would be to convert all urls to use the url() helper (route() is preferred for full laravel pages).

I was curious because I'm pretty sure the leading slash is preferred when using the url() helper.

mpikzink:
Subdir_missing
But with leading / the subdirectory will not be included!?

murrant:
Only when using raw urls.  As shown by your output the url helper knows what to prepend to the url correctly.  It should not matter either way because the url helper will put a slash in between the root and path every time.

mpikzink:
> So, seems like the fix would be to convert all urls to use the url() helper (route() is preferred for full laravel pages).
> 
> I was curious because I'm pretty sure the leading slash is preferred when using the url() helper.

OK, I can change the hrefs i found to route(/...) but how to use route() in page which are not converted to Laravel blade? Can I think I can't use {{ }}directly.

murrant:
Look at the name column from this command `php artisan route:list` to see all routes that can use the route helper.  Then call like this for example `route('device-groups.index')`

For others, just wrap the url() function around the url.

`{{ url('blah') }}` is only in blade templates and is almost the same as `<?php echo url('blah'); ?>` in php files.

mpikzink:
OK, thanks.  Can you show me where the popover links are created? I searched for it but either I can not find the place or something is cached and I do not know how to renew it.

murrant:
generate_port_link() has not been ported to use the url helper yet...

LibreNMS/Util/Url.php
```diff
@@ -225,13 +225,7 @@ public static function sensorLink($sensor, $text = null, $type = null, $overlib
...
     public static function deviceUrl($device, $vars = [])
     {
-        $routeParams = [is_int($device) ? $device : $device->device_id];
-        if (isset($vars['tab'])) {
-            $routeParams[] = $vars['tab'];
-            unset($vars['tab']);
-        }
-
-        return route('device', $routeParams) . self::urlParams($vars);
+        return self::generate(['page' => 'device', 'device' => is_int($device) ? $device : $device->device_id], $vars);
```
murrant, line 228: This is going to cause issues, above code has a workaround for a bug with legacy code calling it.

murrant:
@mpikzink I suggest smaller changes across multiple pull requests.  Will be much easier to make progress.

murrant:
Closing due to inactivity, feel free to reopen.

arrmo:
FYI, I definitely do still have this issue (just to let you know). The "hack" I have in place here works fine ... but I'm not sure it's ready for prime time. I don't see it causing other issues, but I may not be seeing some of the side effects. That's why I haven't tried pushing it (PR) ... but I can, no issue here. Thoughts?

Thanks!

murrant:
@arrmo issues are known. One problem is this PR is too big so a few issues held up known things that do fix stuff.  If you have the time, make multiple smaller PRs to fix issues.

PR #12353, opened 2020-11-25
GitHub Actions

crazy-max:
This PR is the beginning of a best-effort to migrate to GitHub Actions as discussed on Discord since travis-ci.org [will be officially closed down completely on December 31st, 2020](https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom?e=%5BUNIQID%5D). But today [travis-ci.com requires a limited number of credits](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing).

> ...we will upgrade you to our trial (free) plan with a 10K credit allotment (which allows around 1000 minutes in a Linux environment).

ATM I've added a simple workflow to generate mkdocs documentation without disruption of the actual implementation with the Travis build:

* Doc generated with a simple Dockerfile. In the future we could remove `scripts/deploy-docs.sh`.
* Push generated documentation to `gh-pages` branch on this repo.
* We could remove [librenms-docs/librenms-docs.github.io](https://github.com/librenms-docs/librenms-docs.github.io) and only rely on this repo for a better integration and without sharing a Personal Access Token.

You can generate and test doc locally with the following commands:
  * Create mkdocs Docker image: `docker build -t mkdocs -f ./doc/Dockerfile .`
  * Launch local doc website (http://localhost:8000): `docker run --rm -it -p 8000:8000 -v $(pwd):/docs mkdocs`

Live example of the build process and generated doc available on my fork: https://github.com/crazy-max/librenms/runs/1452595127?check_suite_focus=true

image

https://crazymax.dev/librenms/

PR #12035, opened 2020-08-26
Update travis to use both Ubuntu 20.04 and 18.04

Jellyfrog:
This gives us testing of both MySQL 5.7 and 8
Also bumps python to 3.6 and 3.8

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)
- [x] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #12392, opened 2020-12-11
GitHub Actions dev:check ci

crazy-max:
Follow-up #12353 

Web check failed without any log:

```
Running web check... 
failed (0.62s)
```

Git error right after `php artisan dev:check -vv ci`, don't know why:

```
fatal: ambiguous argument 'master': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Running PHP lint check... success (20.91s)
Running Python lint check... success (5.82s)
...
```


Jellyfrog:
It fails because the branch master is not available. Guessing you have to change it to origin/master

.github/workflows/test.yml
```diff
@@ -0,0 +1,143 @@
...
+      - 'doc/**'
+      - 'mkdocs.yml'
+
+env:
+  APP_ENV: testing
+  DB_NAME: db
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
Jellyfrog, line 28: We want to be able to use different version here. 
Check https://github.com/librenms/librenms/pull/12035

Tldr is that we want to test different net-snmp and mysql versions

crazy-max: If you just want to test other releases of mysql you just have to change the version of the mysql Docker service (or create a new one with your desired release). This way we are sure of the current MySQL release we want to use: https://github.com/librenms/librenms/pull/12392/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R46

f0o: I agree with SQL versions supplied via the services tag.

However net-snmp version changes shouldnt really matter much although we should keep track of the LTS versions of ubuntu just to be sure that compatibility is ensured.

murrant: I'm fine with using a new version, just aware that some test data will fail.

murrant:
If you can supply a list of changed files in the `FILES` env variable, the dev:check command will not need to run the git command to find the differences.

murrant (approved):
LGTM, removed the extra test units run, added a couple visual things.

murrant:
@crazy-max @Jellyfrog thanks for getting this going :D

PR #12397, opened 2020-12-12
Add ID to Device Table List and to Device Dependencies Table List + Shorten ifname in Device: Recent Events.

Cupidazul:
3d1eebf:
Add 'device_id' information to Device Table List and to Device Dependencies Table List. This is usefull because sometimes we need to know what device is the last device added do the db.

       modified:   app/Http/Controllers/Table/DeviceController.php
       modified:   includes/html/pages/device-dependencies.inc.php
       modified:   includes/html/pages/devices.inc.php

d6660bc:
Shorten interface name inside device page, inside event-short list, on some devices this interface text is huge, i.e.: teldat (ex. interface name: 'Switch Port interface: Group 1, Port 4; Associated interface: ethernet0/0')

       modified:   includes/html/print-event-short.inc.php

### DIFF
--- a/includes/html/pages/device-dependencies.inc.php
+++ b/includes/html/pages/device-dependencies.inc.php
@@ -29,12 +29,12 @@ require_once 'includes/html/modal/manage_host_dependencies.inc.php';
     <table id="hostdeps" class="table table-hover table-condensed table-striped">
         <thead>
             <tr>
-                <th data-column-id="deviceid" data-visible="false" data-css-class="deviceid">No</th>
+                <th data-column-id="id" data-formatter="deviceid" data-type="int" data-sortable="true" data-visible="true">Id</th>
                 <th data-column-id="hostname" data-type="string" data-css-class="childhost" data-formatter="hostname">Hostname</th>
                 <th data-column-id="sysname" data-type="string" data-visible="false">Sysname</th>
                 <th data-column-id="parent" data-type="string" data-css-class="parenthost" data-formatter="parent">Parent Device(s)</th>
                 <th data-column-id="parentid" data-visible="false">Parent ID</th>
-                <th data-column-id="actions" data-searchable="false" data-formatter="actions">Actions</th>
+                <th data-column-id="actions" data-sortable="false" data-searchable="false" data-formatter="actions">Actions</th>
             </tr>
         </thead>
         <tbody>
@@ -80,6 +80,9 @@ var grid = $("#hostdeps").bootgrid({
         },
         "hostname": function(column, row) {
             return '<a href="device/device='+row.deviceid+'/" class="list-device">'+row.hostname+'</a><br />'+row.sysname;
+        },
+        "deviceid": function(column, row) {
+            return row.deviceid;
         },
         "parent": function(column, row) {
             if (row.parent == 'None') {
--- a/includes/html/pages/device-dependencies.inc.php
+++ b/includes/html/pages/device-dependencies.inc.php
@@ -29,7 +29,7 @@ require_once 'includes/html/modal/manage_host_dependencies.inc.php';
     <table id="hostdeps" class="table table-hover table-condensed table-striped">
         <thead>
             <tr>
-                <th data-column-id="id" data-formatter="deviceid" data-type="int" data-sortable="true" data-visible="true">Id</th>
+                <th data-column-id="id" data-type="int" data-sortable="true" data-visible="true">Id</th>
                 <th data-column-id="hostname" data-type="string" data-css-class="childhost" data-formatter="hostname">Hostname</th>
                 <th data-column-id="sysname" data-type="string" data-visible="false">Sysname</th>
                 <th data-column-id="parent" data-type="string" data-css-class="parenthost" data-formatter="parent">Parent Device(s)</th>
@@ -80,9 +80,6 @@ var grid = $("#hostdeps").bootgrid({
         },
         "hostname": function(column, row) {
             return '<a href="device/device='+row.deviceid+'/" class="list-device">'+row.hostname+'</a><br />'+row.sysname;
-        },
-        "deviceid": function(column, row) {
-            return row.deviceid;
         },
         "parent": function(column, row) {
             if (row.parent == 'None') {
--- a/app/Http/Controllers/Table/DeviceController.php
+++ b/app/Http/Controllers/Table/DeviceController.php
@@ -53,12 +53,13 @@ class DeviceController extends TableController
             'disable_notify' => 'nullable|in:0,1',
             'group' => 'nullable|int',
             'poller_group' => 'nullable|int',
+            'device_id' => 'nullable|int',
         ];
     }
 
     protected function filterFields($request)
     {
-        return ['os', 'version', 'hardware', 'features', 'type', 'status' => 'state', 'disabled', 'disable_notify', 'ignore', 'location_id' => 'location'];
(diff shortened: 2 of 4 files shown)


includes/html/pages/device-dependencies.inc.php
```diff
@@ -81,6 +81,9 @@
         "hostname": function(column, row) {
             return '<a href="device/device='+row.deviceid+'/" class="list-device">'+row.hostname+'</a><br />'+row.sysname;
         },
+        "deviceid": function(column, row) {
```
Jellyfrog, line 84: this is not needed if youre not formatting the data

Cupidazul: Ok, thanks.

PipoCanaja:
Hi @Cupidazul 
Could you make this ID hidden by default ? So it can be made visible if needed by the user.

Cupidazul:
> Hi @Cupidazul
> Could you make this ID hidden by default ? So it can be made visible if needed by the user.

We have that in the Basic view. I don't know how the Detailed view works in these "defaults" options, buts seams to me that we have all itens preselected on the Drop-down, I'll add-in some prints to show what I mean.

Cupidazul:
image
image

includes/html/pages/device-dependencies.inc.php
```diff
@@ -29,12 +29,12 @@
     <table id="hostdeps" class="table table-hover table-condensed table-striped">
         <thead>
             <tr>
-                <th data-column-id="deviceid" data-visible="false" data-css-class="deviceid">No</th>
+                <th data-column-id="id" data-formatter="deviceid" data-type="int" data-sortable="true" data-visible="true">Id</th>
```
Jellyfrog, line 32: Same here. No point in adding a formatter that does nothing

Jellyfrog: And we don't want to show device id by default. It's not very useful info 99% of the times

Cupidazul: Sure, thanks for that.

includes/polling/applications/rrdcached.inc.php
```diff
@@ -40,13 +40,21 @@
     $sock = fsockopen($device['hostname'], 42217, $errno, $errstr, 5);
 
     if (! $sock) {
+        d_echo("\nNo Socket to rrdcached server " . $device['hostname'] . ":42217 try to get rrdcached from SNMP\n");
```
Jellyfrog, line 43: These changes should move to its own pull request

Cupidazul: You're right @Jellyfrog.
I dont know how to do that, thats why it came up to this Pull... I'm not 100% confortable with git. But I'll try to do just that, so that this RRDCached CHG has its own Pull.
Thank you for your comments and advice.

includes/html/pages/device-dependencies.inc.php
Jellyfrog: You should fix all comments in the same pull request

includes/polling/applications/rrdcached.inc.php
Cupidazul: done. thx.

includes/html/pages/device-dependencies.inc.php
Cupidazul: > And we don't want to show device id by default. It's not very useful info 99% of the times

Actually, this "ID" is very important, for the Admin to aquire last ID inserted.
If we have many devices on librenms, and many people addin devices do monitor, the Admin needs to easily check those last devices inserted into librenms to check for configuration best-practices, as for example, if devices have been "type" updated, if Snmp "Port Association Mode" and max-repeaters are correct, If the Ports have been disabled, etc...

murrant:
Just a thought, wouldn't created date (inserted column in the db) be better to show what you are looking for?

Cupidazul:
> Just a thought, wouldn't created date (inserted column in the db) be better to show what you are looking for?

Yeah, Thats a great idea, I'll check into that.
Another idea: shoudn't, all db fields should be availlable for the user in this table ? (I would include ID also, we need the ID for other purposes: manual debug and discovery. p.s: I know the id is on the URL, but most users could not get that far.)

murrant:
I usually find the ID by mousing over the device link and looking at the url.  Sure the options for more fields would be acceptable, but we don't want to crowd the default listing.

Cupidazul:
> > Just a thought, wouldn't created date (inserted column in the db) be better to show what you are looking for?
> 
> Yeah, Thats a great idea, I'll check into that.
> Another idea: shoudn't, all db fields should be availlable for the user in this table ? (I would include ID also, we need the ID for other purposes: manual debug and discovery. p.s: I know the id is on the URL, but most users could not get that far.)

Looking at the db (devices table), I'm not sure, we should have a 'first_discovered_date', is this the 'inserted' (field) ?
Looking at some examples of devices in my db, I dont think the 'inserted' is the actual first discovered date...

Because, all other date fields, are much more recent : last_polled / last_discovered / last_ping

Cupidazul:
> I usually find the ID by mousing over the device link and looking at the url. Sure the options for more fields would be acceptable, but we don't want to crowd the default listing.

Sure, yes, just having them on the drop-down-list should be enough.

murrant:
inserted is when it was first added.  last_x was when the device was most recently polled/discovered/pinged.

Would be nice if we persisted table view settings, but that might have to wait until after someone implements a new table library for LibreNMS.  The current one is no longer supported.

Cupidazul:
> inserted is when it was first added. last_x was when the device was most recently polled/discovered/pinged.
> 
> Would be nice if we persisted table view settings, but that might have to wait until after someone implements a new table library for LibreNMS. The current one is no longer supported.

I think we are using jquery.bootgrid now, right? we have some other good options... (https://www.sitepoint.com/jquery-tables/) but I don't dislike the current one...
Why do you say the current one is no loger supported ? Maybe we just need to add some options in the user preferences (My Settings), and either "cache" the options the user selects, or enable him to fix the default fields in Basic and Detailed grids.

murrant:
> Why do you say the current one is no longer supported ?
Last release was in 2015.

Cupidazul:
> > Why do you say the current one is no longer supported ?
> > Last release was in 2015.

ok.

Cupidazul:
> > Why do you say the current one is no longer supported ?
> > Last release was in 2015.

Cupidazul:
I need to release my VM for future LibreNMS work, thus, I need to know if this pull will fo forward, or should be discarded.

Jellyfrog:
The remaining issues are not fixed yet. 
Also its no problem on working on multiple pull requests at the same time

Cupidazul:
> The remaining issues are not fixed yet.
> Also its no problem on working on multiple pull requests at the same time

thanks, I'll do that, but I need to test your reviews to check if all is working as expected. We dont want to break anything here. This UI is very important and should be robust.

Jellyfrog:
> > The remaining issues are not fixed yet.
> > Also its no problem on working on multiple pull requests at the same time
> 
> Thanks, I'll do just that, starting by fixing the issues, but I need to test your reviews to check if all is working as expected. We dont want to break anything here. This UI is very important and should be robust.
> 
> I've done this pull in my master thread, I should have started a different branch, I'm learning as I go here, and I thank you guys for you patience.
> 
> Thx.

Hi. No problem. Thanks for fixing!

murrant (approved):

PR #12398, opened 2020-12-15
Fixes issues with binding and authenticating users in nested groups

paddy01:
Signed-off-by: Patrik Forsberg <paddy@paddyonline.net>

After having a few issues with authenticating users with our Active Directory I created these modifications and they seem to be working as intended.
In the function bind it returns the result from $bind_result to the requesting module if it is true - which returned "null" pre-patch making active directory bind fail.
In the function userInGroup a bit of code that seem obsolete has been cut away and return value is now derived from ldap_get_entries in the initial search - which seems to be doing what the obsolete code was doing and by doing it again it actually broke authentication.

Using PHP 7.3 and not 7.0 might have something to do with it.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Authentication/ActiveDirectoryAuthorizer.php
+++ b/LibreNMS/Authentication/ActiveDirectoryAuthorizer.php
@@ -91,10 +91,10 @@ class ActiveDirectoryAuthorizer extends AuthorizerBase
         }
 
         // special character handling
-        $ldap_find_chr = array();
+        $ldap_find_chr = [];
         $ldap_find_chr[] = '(';
         $ldap_find_chr[] = ')';
-        $ldap_replace_chr = array();
+        $ldap_replace_chr = [];
         $ldap_replace_chr[] = '\(';
         $ldap_replace_chr[] = '\)';
 
--- a/LibreNMS/Authentication/ActiveDirectoryAuthorizer.php
+++ b/LibreNMS/Authentication/ActiveDirectoryAuthorizer.php
@@ -119,10 +119,10 @@ class ActiveDirectoryAuthorizer extends AuthorizerBase
         $entries = ldap_get_entries($connection, $search);
 
         if ($entries['count']) {
-            return boolval(1);
+            return true;
         }
 
-        return boolval(0);
+        return false;
     }
 
     public function getUserlevel($username)
--- a/LibreNMS/Authentication/ActiveDirectoryAuthorizer.php
+++ b/LibreNMS/Authentication/ActiveDirectoryAuthorizer.php
@@ -91,8 +91,8 @@ class ActiveDirectoryAuthorizer extends AuthorizerBase
         }
 
         // special character handling
-        $ldap_find_chr[]  = "(";
-        $ldap_find_chr[]  = ")";
+        $ldap_find_chr[] = "(";
+        $ldap_find_chr[] = ")";
         $ldap_replace_chr[] = "\(";
         $ldap_replace_chr[] = "\)";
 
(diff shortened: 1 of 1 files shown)


murrant:
The return $bind_result makes little sense to me because no calls to the bind function expect a return value.
Also, it looks like after you removed that code from the userInGroup function it seems to only check if the group exists... which is not a good thing.

paddy01:
> The return $bind_result makes little sense to me because no calls to the bind function expect a return value.

I'm not sure if it is only the test tool that needs it or not but without returning the value the test tool returns " Could not bind to AD, you will not be able to use the API or alert AD users " but with the return of $return_value (which would be the same as returning true) " AD bind success " - so _something_ is using the return value.

> Also, it looks like after you removed that code from the userInGroup function it seems to only check if the group exists... which is not a good thing.

You could be correct.. I never dug deep enough to look closer to what userInGroup actually does .. but I am sure that a user with the wrong group is not allowed and it is used to determine which level the user will get, I have two different groups where one get level 5 and one 10 and even though the user is part of both groups the user gets the "higher" level. And if the user is not part of any allowed groups the user gets rejected.
So after checking a bit even tho it no longer verify that the user is part of the group it verifies that the user is part of the correct DN, which is a plus. Other parts verify that the user is part of the correct group it seems, the name for userInGroup might be wrong after the change tho...

murrant:
@paddy01 I got around to testing this.  Mostly it worked great, but I found one issue.  When you are in multiple approved groups, your final access should be the highest access of all groups you are in.

murrant (changes requested):
@paddy01 I got around to testing this. Mostly it worked great, but I found one issue. When you are in multiple approved groups, your final access should be the highest access of all groups you are in.

paddy01:
> @paddy01 I got around to testing this. Mostly it worked great, but I found one issue. When you are in multiple approved groups, your final access should be the highest access of all groups you are in.

@murrant actually my patch is faulty and probably opens some security issue. I found the real issue and will update my poll request with a real fix for the issue. I've re-instated the group check but kept the return result in the bind function.
As for the use of highest level if a user is part of multiple groups is, from what I can understand in the code, by design .. in the function "getUserlevel" the line " max($userlevel, $level['level']) " would pick the highest possible level among those the user is part of - quite logical but that is the reason.

LibreNMS/Authentication/ActiveDirectoryAuthorizer.php
```diff
@@ -90,7 +90,16 @@ protected function userInGroup($username, $groupname)
             throw new AuthenticationException();
         }
 
-        $group_dn = $result[0]['dn'];
+        // special character handling
+        $ldap_find_chr = [];
```
murrant, line 94: please don't define an array like this.  Do this:
```php
        $ldap_find_chr = ['(', ')'];
        $ldap_replace_chr = ['\(', '\)'];
```

LibreNMS/Authentication/ActiveDirectoryAuthorizer.php
```diff
@@ -90,7 +90,16 @@ protected function userInGroup($username, $groupname)
...
+        $ldap_find_chr[] = '(';
+        $ldap_find_chr[] = ')';
+        $ldap_replace_chr = [];
+        $ldap_replace_chr[] = '\(';
+        $ldap_replace_chr[] = '\)';
+
+        // fix group_dn based on above character handling
+        $group_dn = str_replace($ldap_find_chr, $ldap_replace_chr, $result[0]['dn']);
```
murrant, line 102: Additionally, just use this:

`$group_dn = addcslashes($result[0]['dn'], '()');`

murrant (approved):

PR #12406, opened 2020-12-23
Update functions.inc.php

rk4an:
I think left and right text is inverse. Example in page with all processor, percent_used is at the right instead of left


Jellyfrog (approved):
Seems right, Thanks!

murrant:
Thanks, a casualty of a refactor :)

PR #12414, opened 2020-12-31
Make LibreNMS recognize Schleifenbauer model DPM27/E with existing OS definition

martijn-schmidt:
This pull request is intended to make LibreNMS recognize the DPM27/E energy meter:

https://www.schleifenbauer.eu/wp-content/uploads/2019/04/Datasheet_DPM27E.pdf

It runs the same OS as the other Schleifenbauer hardware that is already supported by LibreNMS, but for some reason the vendor decided to use a sysObjectID `.1.3.6.1.4.1.31034.11.3` instead of `.1.3.6.1.4.1.31034.11.1` which meant that LibreNMS didn't recognize the OS of the device without updating the definitions in `includes/definitions/schleifenbauer.yaml` first.

I collected additional SNMP data and regenerated the JSON file as per the instructions from the [documentation](https://docs.librenms.org/Developing/os/Test-Units/), but only the former changed - the latter stayed exactly the same even if I deleted the original file before running `./scripts/save-test-data.php -o schleifenbauer` to recreate it. Maybe I did something wrong? Feel free to point out my mistake(s), if any..

Since I was working on the OS anyway I also took the liberty of uploading the most recent version of the MIB files, going from release 244 to release 262.

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 12414`


Jellyfrog:
you should run the test data generation scripts with `-v` to specify a variant, maybe `-v dpm27e`

martijn-schmidt:
Thanks for the tips @Jellyfrog, I've reverted the commit with the previous snmpsim data (e.g. two pieces of data in one snmpsim file) and re-collected snmpsim data + generated test output with the `-v dpm27e` as you suggested. Should be good to merge now, once the tests are all done.

Jellyfrog (approved):

PR #12423, opened 2021-01-07
support newer SyncServer

neg2led:
I have a Symmetricom/Microsemi/Microchip (what a lineage) SyncServer S350 running firmware version 2.70.11. 

It reports a sysObjectId of `.1.3.6.1.4.1.9070.1.2.3.1.5` rather than the `.1.3.6.1.4.1.39165.1.6` currently searched for in `microsemitime.yaml`

I have added `.1.3.6.1.4.1.9070.1.2` as this resolves to enterprises.symmetricom.symmNetworkManagement.symmSnmpManagement which is used by formerly-Symmetricom-now-Microsemi timesync products to identify themselves.

I have also slightly reformatted the .yaml to remove the curly braces around the sysObjectId array/list in keeping with a few other definition files' layouts.

#### Please note

- [x] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it. (it does not)

#### Testers


Jellyfrog:
Don't forget to use `--variant`

neg2led:
OK, so I hadn't actually read the relevant bit of the doco there, my bad. Updated test data.

```librenms@skynet:~$ ./lnms dev:check unit --db --snmpsim
Only checking os: microsemitime
Running unit check...
Starting snmpsim listening on 127.1.6.2:1162...
Refreshing database...done
PHPUnit 9.4.2 by Sebastian Bergmann and contributors.

........                                                            8 / 8 (100%)

Time: 00:05.576, Memory: 76.50 MB

OK (8 tests, 1665 assertions)
success (18.07s)
librenms@skynet:~$```

Seems happier now, let's see what CI thinks...

**Comment by Jellyfrog** [comment]
You should extract more fields, version for example, see https://docs.librenms.org/Developing/os/Initial-Detection/#os-discovery

Delete the test data before generating it again after

**Comment by neg2led** [comment]
OK, have added the SYMM-SMI mib as provided by Microsemi/Symmetricom, added extraction of version/hardware model/serial number, and the ntpSysMode state sensor available in SYMM-SMI

I'd like to add some more of the information reported in this MIB, but have two issues. First, most of the state values in this MIB are strings, (despite being purely numeric) but LNMS wants numeric alues for state - I tried to add in the 'tymingSource' value but being Not An Int it did not parse.

Second, is there a cleaner way to integrate the NTPd state (dispersion, offset etc) into this? Normally on a regular old Linux box running ntpd (which is essentially what these are) we'd use the application SNMP extend shellscript, but that's not an option here.

Captured output of `snmpwalk -v2c -cCOMMUNITY -OUQ -m ALL -M '/opt/librenms/mibs/microsemi:/opt/librenms/mibs' 10.61.1.24 .1.3.6.1.4.1.9070`:

```
SYMM-SMI::ntpSysLeap.0 = noWarning
SYMM-SMI::ntpSysStratum.0 = 2
SYMM-SMI::ntpSysPrecision.0 = -20
SYMM-SMI::ntpSysRootDelay.0 = "0.000237"
SYMM-SMI::ntpSysRootDispersion.0 = "0.015576"
SYMM-SMI::ntpSysRefID.0 = "10.61.1.20"
SYMM-SMI::ntpSysRefTime.0 = "3819612619.56339"
SYMM-SMI::ntpSysPoll.0 = 8
SYMM-SMI::ntpSysPeer.0 = 335625482
SYMM-SMI::ntpSysPhase.0 = "0.000896"
SYMM-SMI::ntpSysFreq.0 = "-32.489960 PPM"
SYMM-SMI::ntpSysError.0 = "No error"
SYMM-SMI::ntpSysClock.0 = "Thu Jan 14 22:42:04 2021"
SYMM-SMI::ntpSysSystem.0 = "Symmetricom SyncServer"
SYMM-SMI::ntpSysProcessor.0 = "X86 Processor"
SYMM-SMI::ntpSysNotrust.0 = 1
SYMM-SMI::ntpSysPktsReceived.0 = 486
SYMM-SMI::ntpSysMode.0 = client
SYMM-SMI::ntpSysVersion.0 = "ntpd 4.2.4p8@1.1607-o Thu Mar 10 18:46:30 UTC 2016 (1)"
SYMM-SMI::tymingStatus.0 = "Good"
SYMM-SMI::tymingSource.0 = "4"
SYMM-SMI::tymingTime.0 = "Thu Jan 14 22:42:04 2021"
SYMM-SMI::tymingVersion.0 = "2.103"
SYMM-SMI::tymingFlyPeriod.0 = 0
SYMM-SMI::gpsPosition.0 = "1 0 0 0 0 1 0 0 0 0 0"
SYMM-SMI::gpsUTCOffset.0 = 0
SYMM-SMI::gpsHealth.0 = "Receiver Health: 2"
SYMM-SMI::gpsSatlist.0 = "0"
SYMM-SMI::gpsMode.0 = "Receiver Mode: Survey"
SYMM-SMI::etcVersion.0 = "2.83.2"
SYMM-SMI::etcSerialNbr.0 = "1126A67238"
SYMM-SMI::etcModel.0 = "S350"
SYMM-SMI::etcUpgrade.0 = "Update value: 0"
SYMM-SMI::etcUpgradeServer.0 = "www.symmetricom.com"
SYMM-SMI::etcAlarmString.0 = ""
```
(my unit does not presently have GPS lock)

**Comment by Jellyfrog** [comment]
> I'd like to add some more of the information reported in this MIB, but have two issues. First, most of the state values in this MIB are strings, (despite being purely numeric) but LNMS wants numeric alues for state - I tried to add in the 'tymingSource' value but being Not An Int it did not parse.

You use the `desc` to match a string like so;
`{ value: 1, descr: yes, graph: 1, generic: 0 }`

File: tests/data/microsemitime.json [hunk 1]
```diff
@@ -0,0 +1,1830 @@
...
+                    "sysName": "<private>",
+                    "sysObjectID": ".1.3.6.1.4.1.39165.1.6",
+                    "sysDescr": "Linux SyncServer 2.6.10_mvl401-pc_target.Gryphon.2.10.2009.1201126 #18 Thu Mar 10 10:31:17 PST 2016 i686",
+                    "sysContact": null,
+                    "version": null,
+                    "hardware": null,
+                    "features": null,
+                    "os": "microsemitime",
+                    "type": "timing",
+                    "serial": null,
```
Jellyfrog, line 15: you need to generate new test data, as you can see here your version/serial/hardware is empty

neg2led:
I am bad at Git, please do not judge me

Anyway, I've fixed test data and now that I've wrapped my head around how all that works a little bit better, added a few more state sensors while I was at it. Also added a count sensor, which is... maybe not the right place for that value to go and I should probably remove it.

I also added a small php file that translates some of the silly choices Symmetricom/Microsemi made when designing their MIB (why are some of these things strings instead of ints? why is there a value that's a fixed string prepended to an int? *why?*) - I made an attempt at doing this with `user_func:` but did not get very far, so please let me know if I am Doing It Wrong here.

I want to come back and add proper support for these devices in the NTP poller module, but that's proving to be a bit more of a task than I thought it might & merits its own separate PR.

includes/polling/sensors/state/microsemitime.inc.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+// Workaround for Microsemi/Symmetricom SyncServer SNMP behaviour
+// Several values that should be returned as integers are returned as strings,
+// sometimes with extra text prepended.
+if ($class == 'state') {
```
Jellyfrog, line 6: you should change this to
if ($device['os'] == 'microsemitime') {

murrant: Sensors only loads if the os and class matches, I don't think any if statement is needed here...

murrant: Also, you sure the won't work with yaml?  the state sensor code tries to convert strings to values.

neg2led: Apologies for the delay in replying here. @murrant you're right, it's not necessary at all since this should only be run if the OS and sensor type already match.

That said, taking another look at the other files in this folder, what @Jellyfrog has suggested appears to be convention; save for dhcpatriot.inc.php, they all check for an OS match, I guess as a final sanity check?

It does sort of work with yaml, but the crappy SNMP MIB that they've implemented makes the result it spits out kind of cryptic, e.g.:

```
gpsHealth OBJECT-TYPE
    SYNTAX		DisplayString (SIZE(1..80))
    MAX-ACCESS	read-only
    STATUS		current	
    DESCRIPTION 
        "This is the GPS� receiver health status defined as
        follows:
        ======================================================
        0 = Receiver Down       The Hardware Clock can't 
                                communicate with the receiver.

        1 = Unknown Mode        An undefined mode of the GPS
                                receiver.

        2 = Acquiring Signal    The receiver is attempting to
                                track a GPS signal.

        3 = Bad Geometry        The geometry of the tracked
                                satellites is unsatisfactory for
                                a position solution.

        4 = Propagate Mode      A position estimation mode used
                                in highly dynamic environments.

        5 = 2d Solution         The receiver is able to perform 
                                position fixes for latitude and 
                                longitude but does not have 
                                enough satellites for altitude.

        6 = 3d Solution         The receiver is now able to 
                                perform position fixes for 
                                latitude, longitude and altitude.

        7 = Position Hold       Position fixes are no longer 
                                attempted, and the user entered 
                                or surveyed position is used.

        8 = Time Valid          The receiver has valid timing 
                                information from GPS satellites 
                                (including current leap second 
                                information).  This is the final 
                                state for all configured GPS modes.
        ======================================================"
```

This is a DisplayString, and (for example) returns `Receiver Health: 2`. Yes, I can match on that description with YAML, but since the same YAML field is used to define what string to match on as well as what string to display, I'm unable to map the returned value to its actual description (in this case, `Acquiring Signal`) for display in the web interface,

The same applies to tymingSource, the only difference being that they don't prepend another string to that one's return value; likewise, it's much more useful to display `External 1PPS` than `16`; It would be nice if the YAML state sensor format allowed to match on one string and display another.

If there's a better way to handle this (or a way to handle it solely using YAML) I'm quite happy to change my approach, but this seems to be what's been done for a few other devices with similar flaws.

murrant: In this case the code also extracts the first integer seen in a string.  So, you should still just be able to use a regular state sensor.

neg2led: H-uh. Well. Would you look at that, so it does. Let me go clear this file out, then...

Jellyfrog (approved):

PR #12428, opened 2021-01-11
apc epdu

zerrac:
- Update apc PowerNet Mib from https://www.se.com/fr/fr/product/SFPMIB436/powernet-mib-v4.3.6/
- Add definitions to get votage, current and power from APC easy PDUs (https://www.apc.com/shop/aw/en/products/Easy-PDU-Switched-ZeroU-16A-230V-20-C13-4-C19-IEC309/P-EPDU1116S).

#### Please note


includes/definitions/discovery/apc-epdu.yaml
```diff
@@ -0,0 +1,36 @@
+mib: PowerNet-MIB
+modules:
+    os:
+        hardware:
```
murrant, line 4: This doesn't need to be an array.  Don't these need the mib in the oid to work?
PowerNet-MIB::ePDUDeviceStatusModelNumber

includes/definitions/apc-epdu.yaml
```diff
@@ -0,0 +1,12 @@
+os: apc-epdu
```
murrant, line 1: Is it worthwhile to have a different OS?

includes/definitions/discovery/apc-epdu.yaml
zerrac: my bad, i fixed it

includes/definitions/apc-epdu.yaml
zerrac: maybe i missed something but i think apc os use completely different values in the same mib. My PDUs have no info for the oids used by apc os.

Jellyfrog:
@zerrac missing the .json from https://docs.librenms.org/Developing/os/Test-Units

zerrac:
@Jellyfrog  done

Jellyfrog (approved):

PR #12430, opened 2021-01-11
Rrdcached SNMP: Add poller feature for RRDCached SNMP to query remote agent.

Cupidazul:
Add poller feature for RRDCached SNMP to query remote agent instead of falling back to local unix socket if no connection to agent:42217/tcp is availlable. 
Documentation update is needed to include rrdcached extend snmp agent similar to memcached extend.
Today rrdcached fallback to local unix socket if remote agent is not availlable in port 42217, and in this case, we should try to fetch stats info from SNMP, similar to how memcached works.

       modified:   includes/polling/applications/rrdcached.inc.php

Documentation Update included.

### DIFF
--- a/doc/Extensions/Applications.md
+++ b/doc/Extensions/Applications.md
@@ -128,6 +128,7 @@ by following the steps under the `SNMP Extend` heading.
 1. [PureFTPd](#pureftpd) - SNMP extend
 1. [Raspberry PI](#raspberry-pi) - SNMP extend
 1. [Redis](#redis) - SNMP extend
+1. [RRDCached](#rrdcached) - SNMP extend
 1. [SDFS info](#sdfs-info) - SNMP extend
 1. [Seafile](#seafile) - SNMP extend
 1. [SMART](#smart) - SNMP extend
@@ -1649,6 +1650,24 @@ SNMP extend script to monitor your Redis Server
 
 ```
 extend redis /etc/snmp/redis.py
+```
+
+# RRDCached
+
+SNMP extend script to monitor your RRDCached via snmp
+
+## SNMP Extend
+
+1: Download the script onto the desired host. `wget
+   https://raw.githubusercontent.com/librenms/librenms-agent/master/snmp/rrdcached
+   -O /etc/snmp/rrdcached`
+
+2: Make the script executable: `chmod +x /etc/snmp/rrdcached`
+
+3: Edit your snmpd.conf file (usually `/etc/snmp/snmpd.conf`) and add:
+
+```
+extend rrdcached /etc/snmp/rrdcached
 ```
 
 # Seafile
--- a/doc/Extensions/Applications.md
+++ b/doc/Extensions/Applications.md
@@ -1657,10 +1657,10 @@ extend redis /etc/snmp/redis.py
 Install/Setup:
 For Install/Setup Local Librenms RRDCached: Please see [RRDCached](RRDCached.md)
 
-Capable of collecting stats via:
-1: Local unix socket specified by rrdcached config setting
-1: Local or remote connection to port 42217 specified by rrdcached config setting
-2: Monitor thru snmp with Snmp Extend:
+Will collect stats by:
+1: Connecting directly to the associated device on port 42217
+2: Monitor thru snmp with SNMP extend, as outlined below
+3: Connecting to the rrdcached server specified by the `rrdcached` setting
 
 SNMP extend script to monitor your (remote) RRDCached via snmp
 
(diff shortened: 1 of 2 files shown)


doc/Extensions/Applications.md
```diff
@@ -1651,6 +1652,24 @@ SNMP extend script to monitor your Redis Server
 extend redis /etc/snmp/redis.py
 ```
 
+# RRDCached
+
+SNMP extend script to monitor your RRDCached via snmp
```
murrant, line 1657: This documentation is incomplete.  Monitoring the rrdcached server that LibreNMS uses needs to be described.

murrant:
I guess this is for when you can't connect to the rrdcached server directly from the librenms poller?

doc/Extensions/Applications.md
Cupidazul: Ok, you mean adding in this part, some information regarding local rrdcached socket, right? Ok, I'll add it in.

I think we should add some info here also, right?
https://docs.librenms.org/Extensions/RRDCached-Security/

Regarding security, it is preferrable to use Snmp rather than opening up 42217 directly... right, do you agree?

Do you want me to add in some text there too?

Cupidazul:
> I guess this is for when you can't connect to the rrdcached server directly from the librenms poller?

Yes thats right, this is when we dont have direct access to port 42217 or we didn't open it in the webserver config...

the listen part of:
https://docs.librenms.org/Extensions/RRDCached-Security/

doc/Extensions/Applications.md
murrant: Honestly, if you are monitoring the rrdcached server LibreNMS is using, it does not make any sense to use SNMP to me.
You are already connecting to 42217 for the polling process.  For security, I would just lock it down to the IPs of the pollers via firewall rules.  I am not sure the point of using nginx to proxy, you cannot encrypt the stream or add authentication, so it is basically acting as a firewall and maybe adding overhead.  Perhaps you gain some logging, but I'm unsure.

Cupidazul: Hi, @murrant, I agree with you for localhost.

But this CHG is for polling a remote rrdcached service on another server, where we dont have access to rrdcached thru port 42217 for any reason.

I think that, librenms shouldn't just fallback to local unix socket, especially for cases of remote servers service polling, we should add Snmp query to remote 'extend' rrdcached service.

This was my case, I have librenms on docker, and I have rrdcache on another server that I want to monitor, but I didn't want to open the port for security reasons, what I did was, extend snmp in that server, but without this code, I was simply, monitoring local rrdcache, It took me a while to figure this out, and understand that the rrdcached graphs for the other server, were equal to LibreNMS rrdcached 'localhost' graphs.

The atual code, forces us to query rrdcached of localhost, so the information is not from the remote, and may lead to confusion and it's a wrong information...

murrant:
Yeah, sounds fine.  But you need to update the docs because they make it seem like your use case (which is pretty niche) is actually how everyone should configure the rrdcached app. 99% of users are just going to monitor the rrdcached server that LibreNMS uses itself.

Cupidazul:
> Yeah, sounds fine. But you need to update the docs because they make it seem like your use case (which is pretty niche) is actually how everyone should configure the rrdcached app. 99% of users are just going to monitor the rrdcached server that LibreNMS uses itself.

Ok, sure, I'll update the docs a bit more.
I think that, if any user has memcached on another server, and enables rrdcached 'application', will in fact need this.
99% of users that monitor librenms rrdcache, will enable the application for localhost.

An idea:
Maybe we should differentiate (opt-in or even detect),  somehow, what is the monitoring definition, for local applications, and for remote applications. Users should not see local rrdcached-librenms graphs in 'remote' servers, this may need to confusion.
Information that is local, should be fetched locally, and, information that is remote, should be fetched remotely, and never thru local unix-socket.

murrant:
Agreed, but that is how applications work.

doc/Extensions/Applications.md
```diff
@@ -1651,6 +1652,39 @@ SNMP extend script to monitor your Redis Server
...
+
+Install/Setup:
+For Install/Setup Local Librenms RRDCached: Please see [RRDCached](RRDCached.md)
+Securing RRCached: Please see [RRDCached Security](RRDCached-Security.md)
+
+Local Polling:
+For local Polling the 'RRDCached' Application (librenms-localhost) the usual is either to:
+
+1: Use unix:socket as described: [RRDCached](RRDCached.md)
+2: or Monitor thru port listen 42217 as described: [RRDCached Security](RRDCached-Security.md)
```
murrant, line 1665: Why are you mentioning RRDCached Security doc so much?  Preferred setup for listening on 42217 is covered in the base RRDCached documentation.

Cupidazul: no reason. just for users that need that option. want me to remove those 4 lines ?
thx.

murrant: Just remove them.  As i said, RRDCached Security doc is not very useful for improving rrdcached security, especially since it is a more complicate configuration.

Cupidazul: ok

murrant (approved):

PR #12431, opened 2021-01-11
Processor Graphs - use RRD's last recorded values instead of averages

ottorei:

#### Please note

#### Testers

Add an optional mode to draw graphs using RRD's recorded LAST-value instead of AVERAGE. Option can be enabled from the UI and is disabled by default.

last_values


murrant:
You've got slightly the wrong approach here.  You would never want to change this globally.
It should be graph specific.  Each graph should have a best way to show it's data.

ottorei:
> 
> 
> You've got slightly the wrong approach here. You would never want to change this globally.
> It should be graph specific. Each graph should have a best way to show it's data.
> 
> Like for example sometimes it makes sense to display average and max.

What kind of approach would you suggest? Separate options for different type of graphs?

murrant:
If you want to keep using the generic_multi_line.inc.php base graph, make an option to change the consolidation function.

In the file where the graph is generated, set the option to change the CF to AVERAGE, MINIMUM, MAXIMUM, and LAST as desired.

includes/html/graphs/device/processor_separate.inc.php
```diff
@@ -27,4 +27,10 @@
 
 $nototal = 1;
 
+if (Config::get('webui.graph_processor_last_values')) {
```
murrant: Getting warmer, remove the option and just set `$use_last_values = true;`

murrant (approved):

murrant:
Hmm, after thinking about this a bit.  I am not sure what the point is.  For non-aggregated data (5 min per point) average == last.
For aggregated data, showing the last basically discards a lot of the data.  If you want to see max and min values for aggregation, it should be show in addition to the average.

PR #12437, opened 2021-01-15
Enable secure cookies

tgurr:
Security analyzers (for example Greenbone) report an issue with LibreNMS:

```
SSL/TLS: Missing `secure` Cookie Attribute:
The host is running a server with SSL/TLS and is prone to information
disclosure vulnerability.
```

This PR attempts to address this issue.

Without this change:
librenms_secure_cookie_off

With this change:
librenms_secure_cookie_on

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -174,7 +174,7 @@ return [
     |
     */
 
-    'secure' => env('SESSION_SECURE_COOKIE'),
+    'secure' => env('SESSION_SECURE_COOKIE', true),
 
     /*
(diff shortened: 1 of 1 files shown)


murrant:
You should set `SESSION_SECURE_COOKIE=true` in your .env file.

Are you sure this won't break http only installs?

Jellyfrog:
Yes this is sadly not a good idea to set as default option.
@tgurr feel free to make a pull request documenenting it. Or maybe extending the installer script and set it there

PR #12438, opened 2021-01-15
changed default param max_rows to increase widget count on dashboard

alakiza:
In a large network, there is a need to create large dashboards, so the limit of 15 lines is not enough.

#### Please note

#### Testers


Jellyfrog:
Can't you set max_rows where we use it instead?

murrant:
Yes, very welcome change, but can we do it without modifying the source?  Someone might upgrade it and overwrite your change.

alakiza:
Reasonable. I will try to make it possible to edit this parameter via config.php

murrant:
It doesn't need to be a config.php setting.  Just set `max_rows` here:
https://github.com/librenms/librenms/blob/6d2a8867dc7a2efe92fdf8773b320308e12bf3e9/resources/views/overview/default.blade.php#L183

murrant (approved):
;)

PR #12441, opened 2021-01-17
Add planet-pdu os device

paulierco:
Add planet-pdu os device

#### Please note

#### Testers


includes/definitions/discovery/planet-pdu.yaml
```diff
@@ -0,0 +1,70 @@
+mib: DIGIPDU-MIB
+modules:
+    os:
+        sysDescr_regex: '/PLANET (?<hardware>\S+) .*(?<version>Ver\S+)?/'
```
Jellyfrog, line 4: this doesnt seem to match, if you look at your test data here; https://github.com/librenms/librenms/pull/12441/files#diff-3e25df1fe03d17609c1d8f8f6ec089f9bb18c4496225e0c51f9c530dd3969de5R8-R11 the values are detected as null, probably because the sysdesc doesnt start with `PLANET`

paulierco: That's true. I looked into the other planet OS and just did a copy/paste without checking after. I removed that part so should be fine now.

Jellyfrog: Why not fix it instead? The data is available. Don't forget to update the tests after.

paulierco: Updated. How does it look now ?

murrant (approved):

PR #12453, opened 2021-01-22
Improving readability and layout of dark theme dashboards

efelon:
- lighten the text color in widget tables such as syslog or eventlog (it was black on grey)

before | after
--- | ---
image | image

- remove the useless horizontal scroll bars for widgets alert history and alert history stats (always displayed and scrolls only few pixels)

order | image
--- | ---
before | image
after | image

The screenshots are from the additional custom css "kibanafeel" but **tested as well the with default dark theme**.
I tested the changes on the latest firefox and chrome browsers.

#### Please note

#### Testers

### DIFF
--- a/html/css/dark.css
+++ b/html/css/dark.css
@@ -7183,6 +7183,7 @@
   .device-overview>.panel-body>.row:nth-child(odd) { background-color: #4f565d; }
   .device-overview>.panel-body>.row:hover {background-color: #686d73;}
   .gridster .gs-w{
+    color: var(--color-light2);
   	background: #353a41;
   }
   a.list-device{
@@ -7210,4 +7211,7 @@
   }
   .vis-legend-text {
     color: black;
+  }
+  .widget_body {
+    overflow-x: hidden;
   }


murrant:
I'm not sure overflow hidden is the right fix for the scrollbar showing on the bottom.
It looks like it just needs the margins fixed.

efelon:
I found that setting padding to 1em (see below) would also remove the scrollbars, but also reduce the content of every widget a little bit.
```
.widget_body {
    padding: 1em;
}
```
And I was not able to find a widget which required the horizontal scrollbars of the widget body at all. When you reduce the width of a widget enough, the content itself would display the scrollbar (e.g. Alert history stats, Syslog or Eventlog). On those the `overflow-x: hidden;` has no effect at all.

efelon:
Your comment got me looking in the HTML (which I wanted to prevent to only work with CSS) of the Alert widgets. And indeed there is a strange row before the "alertlog_65-header". The header should be the first in the widget (as in Syslog):

image

If I remove (or reduce by 5px) the margin right the scrollbar disappears as well.

efelon:
Feel free do decline this PR and I will create an issue out of the scroll bars and a new PR for only the readability.

PR #12461, opened 2021-01-25
ISIS-adjacency polling support

ottorei:
Adds a very basic ISIS-protocol support with the functionality to poll configured adjacencies. This information may be used to trigger alarms when the device loses any of the configured adjacencies. Also adds a simple table for device and routing contexts that can be used to view the general status of the topology. Also, links are created on the UI where possible - requires that the device's ifIndexes are saved into the DB.

Each polled adjacency is saved into the DB during the first time it is seen. When polled again, the old record is matched and updated by local ifIndex + device_id as a unique identifier. If the port has ISIS configuration but there has never been any adjacents, this saves only the ifIndex and marks it's status as down.

==> Tested with Juniper MX-80, MX-240 and Cisco ASR9K-series routers. Seems to poll all configured adjacencies

- [x] Basic polling support (== Adjacencies and their status can be polled to some extent)
- [x] Adjacencies can be saved into the DB
- [x] Basic tables in device and routing menus
- [x] Module can be enabled from polling options
- [x] Translations for settings
- [x] Alert fault listing (functions.inc.php) by adjacency IP - could also generate links if the adjacency is found in NMS
- [x] Routing menubar link
- [x] Add alert template for adjacency status check
- [x] Device menubar link
- [x] Cleanup options from poller module (missing as in extra entries that exist in the DB but do not exist on the poll. These entries should be purged from the DB)
- [x] Code style fixes
- [x] Generate links on routing page pages where possible (point interfaces and / or IPs to the devices in question if they exists on the NMS)
- [x] Remove entries from the DB when the device is deleted
- [x] Data formatting
- [x] Database schema and migrations
- [x] Convert pollert module to new style
- [x] Test cases

#### Testing on your instance: 
This patch can be tested with : 
`./scripts/github-apply 12461 `

#### Screenshots
Menu entry on main page:
isis_routing_menu

Routing menu from main page:
isis_routing_view

Routing entries on device page:
isis_device_page

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 12461 `


app/Http/Controllers/Device/Tabs/RoutingController.php
```diff
@@ -38,6 +39,7 @@ public function __construct()
         $device = DeviceCache::getPrimary();
         $this->tabs = [
             'ospf' => $device->ospfInstances()->count(),
+            'isis' => IsisAdjacency::where('device_id', $device->device_id)->count(),
```
Jellyfrog, line 42: use relation instead

database/migrations/2021_25_01_0127_create_isis_adjacencies_table.php
```diff
@@ -0,0 +1,39 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('isis_adjacencies', function (Blueprint $table) {
+            $table->increments('id');
+            $table->integer('device_id');
+            $table->integer('port_id');
```
Jellyfrog, line 18: missing index on both

includes/html/pages/device/routing/isis.inc.php
```diff
@@ -0,0 +1,57 @@
...
+        </thead>';
+
+foreach (IsisAdjacency::where('device_id', $device['device_id'])->get() as $adj) {
+    if ($adj->isisISAdjState == 'up') {
+        $color = 'green';
+    } else {
+        $color = 'red';
+    }
+
+    $interface_name = Port::query()->where('port_id', $adj->port_id)->first()->ifName;
```
Jellyfrog, line 32: creating N+1 queries, use `with()` on your IsisAdjacency query.

includes/html/pages/routing/isis.inc.php
```diff
@@ -0,0 +1,62 @@
...
+    // If ifIndex does not exist for example, because of SNMP interfaces filtering on the device
+    $interface_name = '';
+
+    if ($adj->isisISAdjState == 'up') {
+        $color = 'green';
+    } else {
+        $color = 'red';
+    }
+
+    $interface_name = Port::query()->where('port_id', $adj->port_id)->first()->ifName;
```
Jellyfrog, line 37: same as above

includes/polling/isis.inc.php
```diff
@@ -0,0 +1,91 @@
...
+echo "\nTotal adjacencies: " . $adjacency_count;
+
+echo PHP_EOL;
+
+unset(
+    $adjacencies_poll,
+    $adjacencies,
+    $isis_data,
+    $tmp_adjacencies,
+    $tmp_time,
```
Jellyfrog, line 90:
```suggestion
    $tmp_time
```

Jellyfrog (changes requested):

database/migrations/2021_25_01_0127_create_isis_adjacencies_table.php
ottorei: $table->index(['device_id', 'port_id']) should do this?

includes/functions.php
```diff
@@ -353,6 +353,10 @@ function delete_device($id)
     dbQuery('DELETE `ipv4_addresses` FROM `ipv4_addresses` INNER JOIN `ports` ON `ports`.`port_id`=`ipv4_addresses`.`port_id` WHERE `device_id`=?', [$id]);
     dbQuery('DELETE `ipv6_addresses` FROM `ipv6_addresses` INNER JOIN `ports` ON `ports`.`port_id`=`ipv6_addresses`.`port_id` WHERE `device_id`=?', [$id]);
 
+    //Remove IsisAdjacencies
+    \App\Models\IsisAdjacency::where('device_id', $id)->delete();
+    \App\Models\IsisAdjacency::where('device_id', $id)->delete();
```
murrant, line 358: Duplicate line

includes/html/functions.inc.php
```diff
@@ -936,6 +936,17 @@ function alert_details($details)
             $fallback = false;
         }
 
+        if ($tmp_alerts['isisISAdjIPAddrAddress']) {
```
murrant, line 939: What is the reason for this special case?

includes/html/pages/device/routing/isis.inc.php
murrant: `IsisAdjacency::where('device_id', $device['device_id'])->with('port')->get()`

and then `$adj->port->ifName`

includes/functions.php
ottorei: Removed the other one

includes/polling/isis.inc.php
```diff
@@ -0,0 +1,91 @@
+<?php
```
murrant, line 1: I would prefer if this were a new style poller module.  See the modules in `LibreNMS/Modules`

murrant (dismissed):

includes/html/functions.inc.php
ottorei: To format the alert to show only the adjacents and not general polling time or device's uptime when expanded. See the screenshot.
isis_special_case

includes/html/pages/device/routing/isis.inc.php
ottorei: Fixed

includes/polling/isis.inc.php
ottorei: Changed to use the new style

ottorei (commented):
Review changes

murrant:
Also one more thing needed for all new modules is testing support.  You will need to add an entry to `tests/module_tables.yaml` then capture and generate test data for at least 1 device.

ottorei:
Basic operations are working now. The schema needs to be changed to also save the ifIndex as a fallback in case the port_id can not be found for example due to filtering.

includes/html/functions.inc.php
murrant: Hmm, this feels like the wrong place for that, but I don't have any help beyond that, sorry.

murrant:
I'm not sure handling when the port is missing is a worthwhile case to code around.   Maybe send an event log when that occurs is enough.

ottorei:
> 
> 
> I'm not sure handling when the port is missing is a worthwhile case to code around. Maybe send an event log when that occurs is enough.

I was thinking this too at first but after testing with lab gear I found that there might be some cases where the port_id can not be found. This is particularly an issue with Juniper gear with SNMP-filters. By also saving the ifIndex the adjacencies can be reliably linked to each device in these cases. This problem only occurs if there is no port data of the device's ports available.

includes/html/functions.inc.php
murrant: I suggest  moving this code to a separate PR, it is going to be a blocker for this PR.

misc/config_definitions.json
```diff
@@ -4286,6 +4286,13 @@
             "default": true,
             "type": "boolean"
         },
+        "poller_modules.isis": {
+            "order": 331,
```
murrant, line 4290: 331 does not seem alphabetical.

LibreNMS/Modules/Isis.php
```diff
@@ -0,0 +1,164 @@
...
+        $isis_codes['3'] = 'L1L2';
+        $isis_codes['4'] = 'unknown';
+
+        if ($os instanceof Junos) {
+            // Do not poll loopback interface
+            unset($circuits_poll['16']);
+        }
+
+        // Loop through all configured adjacencies on the device
+        foreach ($circuits_poll as $circuit => $circuit_data) {
```
murrant, line 80: Instead of this huge foreach loop, take a look at using the SyncsModels trait.  See https://github.com/librenms/librenms/blob/master/LibreNMS/Modules/Mempools.php#L57 for an example.

LibreNMS/Modules/Isis.php
```diff
@@ -0,0 +1,164 @@
...
+     * @param OS $os
+     */
+    public function poll(OS $os)
+    {
+        // Get device object
+        $device = $os->getDevice();
+        $device_id = $os->getDeviceId();
+
+        // Poll all ISIS enabled circuits from the device
+        $circuits_poll = $os->getCacheTable('ISIS-MIB::isisCirc', 'ISIS-MIB');
```
murrant, line 61: getCacheTable() makes the data available outside this module by storing it on the OS.  Is that needed?

includes/html/functions.inc.php
ottorei: OK, I will remove this and make a separate PR for this later.

murrant:
@ottorei this is failing because of an upstream change.  You can either update your branch and re-capture test data or update the json files by hand.

LibreNMS/Modules/Isis.php
ottorei: Good point, that is not actually needed.

LibreNMS/Modules/Isis.php
ottorei: All circuits would still need to be polled every time since the adjacency is lost when the configured adjacency goes down for a while. The only way to detect these is to first poll all configured circuits and then poll all adjacencies for these circuits. The circuit will always return the correct value if there is an ISIS-configuration on it but the adjacencies table does only return values if the adjacency is up at the moment the polling is done. That is why I am saving all values and updating them based on the ifIndex of the device. That way the missing and downed adjacencies can be updated and they also retaing useful information such as IfIndex and system id while they are down.

mibs/junos/ISIS-MIB
```diff
@@ -0,0 +1,4581 @@
+ISIS-MIB DEFINITIONS ::= BEGIN
```
PipoCanaja, line 1: Hello
ISIS-MIB is a standard MIB, it must be located in the mibs folder, not in the ,mibs/junos folder. And in fact, is is already in the mibs folder, and seems to be the same version. Please check and remove the mibs/junos one.
Thanx

PipoCanaja (changes requested):
Mib file

mibs/junos/ISIS-MIB
ottorei: For some reason the ISIS-polling does not seem to work at all when this file is removed. Do I need to specify the location for those generic MIBs?

ottorei: Found the issue - there seems to be an old MIB-file with possible incorrect OIDs in the junos-folder. Deleting that seems to resolve this issue.

LibreNMS/Modules/Isis.php
Jellyfrog: After implementing SyncsModels in https://github.com/librenms/librenms/pull/12605, i can totally recommend it ^^

LibreNMS/Modules/Isis.php
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        // Loop through all configured adjacencies on the device
+        foreach ($circuits_poll as $circuit => $circuit_data) {
+            if (is_numeric($circuit)) {
+                echo "\nAdjacency found on ifIndex: " . $circuit;
+                $port_id = (int) $device->ports()->where('ifIndex', $circuit)->value('port_id');
+
+                if ($circuit_data['isisCircPassiveCircuit'] != '1') {
+                    //var_dump($adjacencies_poll[$circuit]['isisISAdjState']);
```
Jellyfrog, line 87: Debug code

includes/html/pages/device/routing/isis.inc.php
```diff
@@ -0,0 +1,63 @@
...
+    //dd($adj);
+    if ($adj->isisISAdjState == 'up') {
+        $color = 'green';
+    } else {
+        $color = 'red';
+    }
+
+    //dd($adj);
+    $interface_name = $adj->port->ifName;
+    //dd($interface_name);
```
Jellyfrog, line 38: Debug code

includes/html/pages/device/routing/isis.inc.php
```diff
@@ -0,0 +1,63 @@
...
+            <th>Adjacent</th>
+            <th>System ID</th>
+            <th>Area</th>
+            <th>System type</th>
+            <th>State</th>
+            <th>Last uptime</th>
+          </tr>
+        </thead>';
+
+// FIXME: Can refer to ports, production.ERROR: Call to undefined relationship [Port] on model [App\Models\IsisAdjacency]
```
Jellyfrog, line 25: Still relevant?

LibreNMS/Modules/Isis.php
```diff
@@ -0,0 +1,166 @@
...
+        // Poll all ISIS enabled circuits from the device
+        $circuits_poll = snmpwalk_group($device_array, 'ISIS-MIB::isisCirc', 'ISIS-MIB');
+
+        // Poll all available adjacencies
+        $adjacencies_poll = snmpwalk_group($device_array, 'ISIS-MIB::isisISAdj', 'ISIS-MIB');
+        $adjacencies = collect();
+        $isis_data = [];
+
+        // Translate system state codes into meaningful strings
+        $isis_codes['1'] = 'L1';
```
murrant, line 70: Please initialize $isis_codes.  The best way is to just set it to a fully formed array.

includes/html/pages/device/routing/isis.inc.php
ottorei: Not anymore, I'll push updates probably tomorrow.

LibreNMS/Modules/Isis.php
ottorei: Removed debug code

includes/html/pages/device/routing/isis.inc.php
ottorei: Removed old commend

LibreNMS/Modules/Isis.php
ottorei: Initialized

ottorei (commented):
Deleted debug code and unnecessary mib-file

murrant:
Just a little more cleanup from the lint failures.

murrant:
Can you have discovery check `ISIS-MIB::isisSysAdminState` or something, then store that as a component or device attribute and check it in  the poller to see if we need to poll ISIS.  That way we can enable this for all OS.

Instead of end() you could use \Illuminate\Support\Arr::last()

ottorei:
> 
> 
> Can you have discovery check `ISIS-MIB::isisSysAdminState` or something, then store that as a component or device attribute and check it in the poller to see if we need to poll ISIS. That way we can enable this for all OS.
> 
> Instead of end() you could use \Illuminate\Support\Arr::last()

Where should this information be stored? Is there already a place for this?

murrant:
> > Can you have discovery check `ISIS-MIB::isisSysAdminState` or something, then store that as a component or device attribute and check it in the poller to see if we need to poll ISIS. That way we can enable this for all OS.
> > Instead of end() you could use \Illuminate\Support\Arr::last()
> 
> Where should this information be stored? Is there already a place for this?

I gave you two options  `Component` or `DeviceAttrib`.

ottorei:
> 
> 
> Can you have discovery check `ISIS-MIB::isisSysAdminState` or something, then store that as a component or device attribute and check it in the poller to see if we need to poll ISIS. That way we can enable this for all OS.
> 
> Instead of end() you could use \Illuminate\Support\Arr::last()

I added a discovery process for adding this as a component

> 
> 
> > > Can you have discovery check `ISIS-MIB::isisSysAdminState` or something, then store that as a component or device attribute and check it in the poller to see if we need to poll ISIS. That way we can enable this for all OS.
> > > Instead of end() you could use \Illuminate\Support\Arr::last()
> > 
> > 
> > Where should this information be stored? Is there already a place for this?
> 
> I gave you two options `Component` or `DeviceAttrib`.

Added a simple component check.

murrant (approved):

PR #12463, opened 2021-01-25
Cisco enhanced cellular

Schouwenburg:
Added extra cellular sensors for Cisco IOS/IOS-XE.

image
image
image

#### Please note

#### Testers


Jellyfrog:
Since you have a different device than the current test data, please generate new tests by utilizing `--variant` flag

loopodoopo:
Tested this PR. Looks very good, this would be a great addition, would be greate to see in the next release :)

murrant:
@Schouwenburg Looking at this, it seems like band and channel are just ways of showing frequency.  Items should always be stored as their base units.  If the display should be different, it should be at the display level, not the storage level.

Another option is to implement them as state sensors (or both)

Schouwenburg:
Hi Tony,

Technically you are right. Band and channel in the end all come down to a
frequency. The thing is that it is way more complicated them in example
WiFi where you have the limited concept of bands (2.4Ghz, 5Ghz) or
microwave (upper 6, lower 6). In mobile you have the different technology
versions gprs, hsdpa, hsdpa+, lte, lte-a, 5G and most bands in each
technology map to other frequencies, so band 5 is not always the same
frequency. Also the mapping of channels is different. I think that is the
reason why it is all abstracter to bands and channels so humans can
understand it at a glance.

Yes these could be sensors, but in the end they all could be sensors right.
The thing is that you loose the overview where you can see the relation
between quality parameters, handover in cells, bands etc. That is the great
benefit of having them on one page. When analyzing connectivity issues it’s
all about these relations. Our NOC constantly scans them and we have weekly
meetings where our third line specialists discuss and analyse individual
endpoint performance and try to make logic out of the numbers. If it would
be needed to jump to different screens all the time for each endpoint, that
would not be a great experience. We already have the challenge with a
separate window to watch the location in time (which is by the way
unexplored territory for LibreNMS and might be a future enhancement).

That’s my thoughts on this.I understand my goals will probably differ a bit
from yours. I want to supply a great tool for our use case, you want a
generic solution that will not become  a mess over time because of all
additions that have been made. But in the end it is important that the tool
makes sense for the user and that relevant data can be found on a logical
place. In this case I feel I made the right choice, but hey, it’s your call
:-). Keep up the great work.

Regards,

Janno



Op di 2 feb. 2021 om 06:04 schreef Tony Murray <notifications@github.com>

murrant:
I am sure you have valid use cases for you.  But in the end one of my goals as a maintainer is to reduce code that only benefits one user and reduce code duplication.

If you look at other wireless devices anything that is one of a set group of items is always a state sensor.  Look at modulation for example.  I think band definitely falls in the state sensor group and band could go either way (or even both ways).  All WiFi equipment should convert channels to frequency for example.

As for having one place to see everything you are looking for.  An often requested feature is reporting ;)

Location over time does sound cool :)

Schouwenburg:
I was also thinking about a feature for custom grouping certain relevant
measurements together on user/system defined pages, but it seems you call
that reporting. We do that already with rrd’s from cacti on a portal, so we
could consider something similar with rrd’s from Libre.

Back to the PR, just let me know how you want it, I think you know my
preferred setup and reasoning, so it comes down to you now.

Regards,

Janno

Op di 2 feb. 2021 om 07:34 schreef Tony Murray <notifications@github.com>

murrant:
FYI, you can already include almost any graph on a dashboard.

I outlined where I think the sensors should go in my previous comment :)

includes/discovery/sensors/count/cisco.inc.php
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @package    LibreNMS
+ * @subpackage webui
+ * @link       http://librenms.org
+ * @copyright  2021 LibreNMS
+ * @author     LibreNMS Contributors
+*/
+
+$tables = [
+    ['num_oid' => '.1.3.6.1.4.1.9.9.661.1.3.1.1.1.',    'oid' => 'c3gImsi',                   'state_name' => 'c3gImsi',                    'mib' => 'CISCO-WAN-3G-MIB', 'descr' => 'IMSI'],
```
PipoCanaja, line 19: Hi
IMSI and IMEI are not something you want to store in an RRD. You should use it in the description of other sensors, or as a group name, in order to see it without wasting ressources to store it with historical data every 5 minutes.
Nearby cells are completely OK, they should remain in Count sensor.

includes/discovery/sensors/count/cisco.inc.php
```diff
@@ -0,0 +1,38 @@
...
+];
+
+foreach ($tables as $tablevalue) {
+    $temp = snmpwalk_cache_multi_oid($device, $tablevalue['oid'], [], $tablevalue['mib']);
+    $cur_oid = $tablevalue['num_oid'];
+    $state_name = $tablevalue['state_name'];
+    foreach ($temp as $index => $entry) {
+        //Discover Sensors
+        $descr = ucwords($temp[$index][$tablevalue['descr']]);
+        if ($state_name == 'ffgfggf' || $state_name == '345tyh' || $state_name == 'dgdfgdd' || $state_name == 'gfgfgf') {
```
PipoCanaja, line 31: This code here seems interesting :) you probably forgot to cleanup :) 
State_name doesn't mean anything here anyway, this is not a state sensor. You should completely get rid of it.

PipoCanaja (dismissed):

includes/discovery/sensors/count/cisco.inc.php
Schouwenburg: Thanks, indeed needed a cleanup. fixed it locally, will be fixed in next push.

includes/discovery/sensors/count/cisco.inc.php
Schouwenburg: @PipoCanaja thanks for the feedback. IMEI is indeed mostly static, IMSI graphing has in our environment great use as they can change more often. Nevertheless I understand your feeling about these sensors. The thing is that I do think the presentation now is very clean, and your suggestions to safe some cycles is to do the presentation in a dirty way and put data on places which are not made for it. I sincerely do not think that would be the right way to go. In general I think what we are mostly missing is a sensor of type string that will not be graphed. Would be a really nice addition and would also be a clean way of presenting other data than numerical values. As this would be a core enhancement, i don't think I should work in that direction. I will think a bit more about an alternatives for IMEI and IMSI.

Schouwenburg:
> If you look at other wireless devices anything that is one of a set group of items is always a state sensor. Look at modulation for example. I think band definitely falls in the state sensor.

@murrant As per your request, band has been removed as a Wireless sensor and became a count sensor.

includes/discovery/sensors/count/cisco.inc.php
PipoCanaja: @Schouwenburg I understand your need, but RRDs are definitely not the way to go, as you know. We don't currently have an ideal solution to store historical data for a sensor outside an RRD. Eventlog is the only existing structure that would allow you to keep history, and the data itself could be stored in components. They are completely flexible.
Only current IMSI/IMEI would be in the SQL db, and the historical data would be in the eventlog (at every poll, if new value is different from current one, you generate an eventlog in order to keep track).

Schouwenburg: @PipoCanaja Ok, as per your request IMSI and IMEI are now removed as sensors and added to entity-physical.

Schouwenburg:
I think this PR is now ready for a new evaluation.

murrant:
I think this is acceptable to me.  One small tweak (that will require updated data)
Instead of including `Modem 0 on Cellular0` in the name, set that to the group for state and count sensors.

Schouwenburg:
> I think this is acceptable to me. One small tweak (that will require updated data)
> Instead of including `Modem 0 on Cellular0` in the name, set that to the group for state, count, and temp sensors.

Done.

murrant (approved):

includes/discovery/sensors/state/cisco.inc.php
```diff
@@ -117,6 +122,63 @@
...
+            $states = [
+                ['value' => 1, 'generic' => 3, 'graph' => 0, 'descr' => 'unknown'],
+                ['value' => 2, 'generic' => 0, 'graph' => 0, 'descr' => 'roaming'],
+                ['value' => 3, 'generic' => 0, 'graph' => 0, 'descr' => 'home'],
+            ];
+        } elseif ($state_name == 'c3gGsmSimStatus') {
+            $states = [
+                ['value' => 1, 'generic' => 3, 'graph' => 0, 'descr' => 'unknown'],
+                ['value' => 2, 'generic' => 0, 'graph' => 0, 'descr' => 'ok'],
+                ['value' => 3, 'generic' => 0, 'graph' => 0, 'descr' => 'not inserted'],
```
Cupidazul, line 170: Note that this is not right:

.1.3.6.1.4.1.9.9.661.1.3.2.1.6.<idx> =  c3gGsmSimStatus  : refers to :

Current service type: 
unknown(1) - service type is unknown 
invalid(2) - no service 
circuitSwitched(3) - circuit switched service 
packetSwitched(4) - packet switched service 
combined(5) - combination of circuit and packet switched service

source:
http://oidref.com/1.3.6.1.4.1.9.9.661.1.3.2.1.14

I've detected this when testing in a real device:
My LibreNMS shows "SIM Status" = 3 = not inserted .

Router SIM is OK:  telnet info:

Modem Security Information
==========================
Active SIM = 0
SIM switchover attempts = 0
Card Holder Verification (CHV1) = Enabled
SIM Status = OK
SIM User Operation Required = None
Number of CHV1 Retries remaining = 3

Schouwenburg: Hi @Cupidazul, thank you for your feedback and great to see others value the enhanced cellular information. I will take a look at your comment and will look how that oid would turn out in my environment, and yes, I also have these 899's and a bunch of other cellular devices available. What would your suggestion be? Adding the extra status information as a sensor and making the output of the c3gGsmSimStatus better by looking at the status of other variables like IMSI and change "not inserted" to "ok" if there is clearly a sim inserted? My guess is that the faulty reporting is IOS dependent as I have not seen the wrong reporting myself.

Cupidazul: Hi @Schouwenburg , yes I use these OID's frequently on other systems, It's awesome, someone finally got this working in LibreNMS. I have other two device types I usually collect info, Huawei and TELDAT, actually I have all the information about MIB's , Status, RSSI and other values, your work is a very nice starting point. Congrats to you! Very Nice!

Here are my suggestions:
1. We should create a new discovery/poller/menu type, like 'gsm' or 'celullar', instead of 'wireless', this deserves something better.
2. As we can see, in this case we are talking about SIM status, and we could try to adapt for each device type, in this case the 899's have another information status for the atual OID (yes the router has a SIM sinserted, for sure, as we can see in the telnet output).
3. We should divide the information about GSM, each in their groups, SIM Info , Cell Info , Service Info , Roaming Info, etc...

And, Yeah I know what you mean, unfortunately, I've seen sometimes, from one device to another, and even with different IOS's things might change.

But in this paticular MIB  the actual code states:   .1.3.6.1.4.1.9.9.661.1.3.2.1.6. = c3gGsmSimStatus 
 ['num_oid' => '.1.3.6.1.4.1.9.9.661.1.3.2.1.6.',    'oid' => 'c3gGsmSimStatus',                      'state_name' => 'c3gGsmSimStatus',              'mib' => 'CISCO-WAN-3G-MIB',                'descr' => 'SIM status'],

We might need to double check this, as, I've found this information, that looks to contradict our actual code[:](http://www.circitor.fr/Mibs/Html/C/CISCO-WAN-3G-MIB.php)

c3gGsmCurrentRoamingStatus | 1.3.6.1.4.1.9.9.661.1.3.2.1.6
-- | --
This object indicates whether the modem is in the home network or is roaming.
OBJECT-TYPE |   |   |  
  | INTEGER | unknown(1), roaming(2), home(3)

---

c3gGsmSimStatus | 1.3.6.1.4.1.9.9.661.1.3.5.1.1.2
-- | --
SIM status. Indicates whether the SIM is present or removed from the SIM socket, and its current status.
OBJECT-TYPE |   |   |  
  | INTEGER | unknown(1), ok(2), notInserted(3), removed(4), initFailure(5), generalFailure(6), locked(7), chv1Blocked(8), chv2Blocked(9), chv1Rejected(10), chv2Rejected(11), mepLocked(12), networkRejected(13) |  


Keep Up this nice Work! yup I'm using it ! its awesome.
Cheers!

Cupidazul: Hi @Schouwenburg , reminder.

Cupidazul: Corrected in ([#12724](https://github.com/librenms/librenms/pull/12724))

PR #12471, opened 2021-01-27
API: Allow logs to be filtered by min/max id

kedare:
Allow the from/to parameters on the various logs API endpoints to be used to filter by id (get all the logs after/before this event id) instead of datetimes

(I did not tested it yet, I am building my dev environment to do so, this may take some time so feel free to take a look if it looks okay)

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -2235,18 +2235,22 @@ function list_logs(Illuminate\Http\Request $request, Router $router)
         $query = ' FROM eventlog LEFT JOIN `devices` ON `eventlog`.`device_id`=`devices`.`device_id` WHERE 1';
         $full_query = 'SELECT `devices`.`hostname`, `devices`.`sysName`, `eventlog`.`device_id` as `host`, `eventlog`.*'; // inject host for backward compat
         $timestamp = 'datetime';
+        $id_field = 'event_id';
     } elseif ($type === 'list_syslog') {
         $query = ' FROM syslog LEFT JOIN `devices` ON `syslog`.`device_id`=`devices`.`device_id` WHERE 1';
         $full_query = 'SELECT `devices`.`hostname`, `devices`.`sysName`, `syslog`.*';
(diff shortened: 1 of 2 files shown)


includes/html/api_functions.inc.php
```diff
@@ -2264,12 +2268,20 @@ function list_logs(Illuminate\Http\Request $request, Router $router)
...
+            $query .= " AND $id_field >= ?";
+        } else {
+            $query .= " AND $timestamp >= ?";
+        }
         $param[] = $from;
     }
 
     if ($to) {
-        $query .= " AND $timestamp <= ?";
+        if (is_numeric($to)) {
```
murrant, line 2280: What if to or from is a timestamp?  How can you tell the difference?

kedare: I am not aware you can actually pass numerical timestamps ? As the database fields are datetime (I always used dates as string as parameter in this case)

kedare: Tested locally and currently it looks like numeric timestamps are ignored.

SourceDoctor:
for myself it's not just clear why i want to filter by specific Row Id
Could you give an example for what it is usefull?

kedare:
For my example I have developed a Stackstorm integration for LibreNMS that watch for eventlog/errorlog (I query every 15 seconds the api endpoint to get the last 15s of events),
Currently I am using the timestamp but I have a few issue :
- It's TZ dependent
- If I switch to using the max(date) from the last request +1s to not get the same last entry), I would probably lose potential events from this second
- Having to urlize date as string in the URL is not the best thing (lot of escaping)

So I wanted to be able to say, take all the events after the last event I got from the previous request.

This is the code I am currently using https://github.com/kedare/stackstorm-librenms/blob/master/sensors/lib/sensors.py

murrant (approved):
Interesting, LGTM.

kedare:
Thanks :)

PR #12473, opened 2021-01-28
Add cache busting support to gh action

Jellyfrog:
Since github action lacks a way to clear the cache, this allows to set the secret `CACHE_VERSION` to invalidate the current cache

#### Please note

#### Testers


murrant:
If you need this doesn't that mean the cache key is incomplete?

Jellyfrog:
> If you need this doesn't that mean the cache key is incomplete?

It is. But it's tricky to fix since we can't calculate it in a good way. Either way it's good to be able to invalidate it.

murrant:
Agreed, just hope we don't have to use it often.

PR #12483, opened 2021-01-30
Add application xmrig

yrebrac:
LibreNMS application for XMRig (monero miner) 
Accompanied by python snmp extend script in https://github.com/librenms/librenms-agent/pull/347
Tested on live installation
Style checks applied
Example of each graph below
xmrig_hashrate_example
xmrig_jobtime_example
xmrig_shares_example
xmrig_threads_example

#### Please note

#### Testers


yrebrac:
It's easy enough to write a unit test but I don't see any existing unit tests to base off:
`for prefix in `ls includes/polling/applications | cut -d. -f1`; do find tests/ -iname '$prefix*'; done`
Is it actually a requirement for applications? If so, the results should simply match what the comments for function json_app_get() say, so actually a generic test might do. Some applications seems to optionally use this or a legacy method though, so it seems like a generic test would need some more thought how it would be called, and that's getting outside the scope of this PR. What do you recommend?

Jellyfrog:
No need to write anything, did you read https://docs.librenms.org/Developing/os/Test-Units

yrebrac:
Ok figured it out I think. added the .snmprec

yrebrac:
The pending CLA problem is back. How to fix permanently? Looks like there is one commit I keep trying to amend with the wrong author.

yrebrac:
commit email problems, rebranching and will open new PR

murrant:
@yrebrac rebase -i and squash the commits with the bad email (or all of them)

PR #12413, opened 2020-12-31
Add alert_details script for easier toubleshooting

theochita:
This script will make alert troubleshooting easier by providing a simple way to output alert specific details.

Please give a short description what your pull request is for

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [ ] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [ ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #12484, opened 2021-01-30
webui: Add button to show verbose alert details in the alert, alert-log webui

theochita:

#### Please note

#### Testers

### DIFF
--- a/includes/html/common/alert-log.inc.php
+++ b/includes/html/common/alert-log.inc.php
@@ -83,6 +83,7 @@ $common_output[] = '
             <th data-column-id="hostname">Device</th>
             <th data-column-id="alert">Alert</th>
             <th data-column-id="severity">Severity</th>
+            <th data-column-id="verbose_details" data-sortable="false">Details</th>
         </tr>
         </thead>
     </table>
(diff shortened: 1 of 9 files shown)


theochita:
Based on Murrants comment in https://github.com/librenms/librenms/pull/12413

theochita:
image

image

includes/html/common/alert-log.inc.php
```diff
@@ -83,6 +83,7 @@
             <th data-column-id="hostname">Device</th>
```
Jellyfrog, line 83: All files have the wrong permissons

Jellyfrog (changes requested):

includes/html/common/alert-log.inc.php
theochita: should have fixed the permissions now.

TheMysteriousX:
This change allows unprivileged users to view credentials used for device access - is this intended?

theochita:
> This change allows unprivileged users to view credentials used for device access - is this intended?
When i copy a successful call to the ajax_form type "alert-details" and then logout and try again i get Unauthorized.

It is correct that it is not doing any user permission checks itself, normal users would be able to use it if they have the device permissions, giving them access to the community string so could be a potential source of leaks.

Making it global admin only is a possibility.

theochita:
I made the use of this feature require admin permissions as per what @TheMysteriousX pointed out, better safe than sorry.

Jellyfrog:
The tests fails

includes/html/common/alert-log.inc.php
```diff
@@ -37,6 +37,10 @@
     'Warning' => 5,
 ];
 
+if (Auth::user()->hasGlobalAdmin()) {
+    $admin_verbose_details = '<th data-column-id="verbose_details" data-sortable="false">Details</th>';
```
murrant, line 41: This only prevents the link from showing up, the details are still in the html below.  You'll need to add a check down there too.

theochita (commented):
all tests passing.

theochita:
Anything else i can do?

murrant (approved):
I'd prefer not to add legacy frontend code, but lgtm.

Jellyfrog (approved):

PR #12495, opened 2021-02-04
Added feature that will replace variables in the body field

tehEGO:
Added feature that will replace variables in the body field of the Transport Api.php

This adds the ability to markup variables similar to the header and options fields via the {{ $variable }} syntax that walks the whole object. It is useful for crafting the body as json or whatever format your endpoint it but also passing the lookup of results of the object, alert, and fault information.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Api.php
+++ b/LibreNMS/Alert/Transport/Api.php
@@ -43,7 +43,7 @@ class Api extends Transport
     {
         $request_opts = [];
         $request_heads = [];
-	    $query = [];
+        $query = [];
 
         $method = strtolower($method);
         $host = explode('?', $api, 2)[0]; //we don't use the parameter part, cause we build it out of options.
@@ -76,9 +76,9 @@ class Api extends Transport
         */
         function iterateVariables($obj, &$subject)
         {
-            foreach( $obj as $key => $value ) {
+            foreach ($obj as $key => $value) {
                 // iterate variables if it is an array recursivly
-                if ( is_array($value)) {
+                if (is_array($value)) {
                     iterateVariables($value,$subject);
                 } else {
                     // replace expected templated markup of {{ $variable }}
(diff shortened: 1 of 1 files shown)


murrant:
Would it be better to just supply the variables to blade?

tehEGO:
Sorry I am new to most of this. I am supplying the variable names to the body ui but there is no code that translates the variables? Do you have some way I can already write variables in body text field of Api.php for this to work? Thank you.

tehEGO:
I am testing Laravel template in body field of Api.php as suggested will report back.

tehEGO:
I have put the templated syntax into the body field of the Api.php and it is not evaluated, the other sections header and options do pass through looking for the templated syntax for resolution.

tehEGO:
I iterate the whole object tree to access alerts AND Faults.

byCarlsJr:
Break the function out of the contactAPI method into it's own method in the class

`
    private function contactAPI($obj, $api, $options, $method, $auth, $headers, $body)
    {
		// recursively iterate $obj attributes. Json encode/decode used to typecast into native array
		$this->iterateObjectAttributes(json_decode(json_encode($obj),true),$body);
	}

	/**
	* Recursively interates $obj's attributes into $subject for text replacement
	* as {{ $value }} format
	* @param array $obj       - Object array to iterate
	* @param string &$subject - String with text to be replaced
	* @return void
	*/
	private function iterateObjectAttributes($obj,&$subject)
	{
		foreach( $obj as $key => $value )
		{
			if( is_array($value) )
			{
				$this->iterateObjectAttributes($value,$subject);
			} else {
				$subject = preg_replace("/\{\{\s*\\\${$key}\s*\}\}/i",$value,$subject);
			}
		}
	}
`

murrant:
Here is where I am at with this PR.  I like the functionality, but the implementation is wrong.  Using preg_replace is not the right way.  We already have code that parses {{ }} and fills in variables.  So effectively this is double processing the template and will likely lead to bugs and edge cases.

Look here to get started: https://github.com/librenms/librenms/blob/master/LibreNMS/Alert/Template.php#L74

murrant:
Closing due to lack of activity.

PR #12499, opened 2021-02-05
Added additional function called good_if_regexp to allow whitelist of…

philr2:
… intefaces using regular expressions.

This should also be used with a wildcard match on all interfaces using bad_if_regexp.

e.g the following will only allow ens interfaces on linux.
$config['os']['linux']['bad_if_regexp']='/.*/'
$config['os']['linux']['good_if_regexp']='/^ens[0-9]+$/'

e.g the following will only allow physical and aggregated interfaces on IOSXR.
$config['os']['iosxr']['bad_if_regexp']='/.*/'
$config['os']['iosxr']['good_if_regexp']='/^TenGigE[0-9].*/'
$config['os']['iosxr']['good_if_regexp']='/^GigabitEthernet[0-9].*/'
$config['os']['iosxr'']['good_if_regexp']='/^FastEthernet[0-9].*/'
$config['os']['iosxr']['good_if_regexp']='/^Bundle-Ether[0-9].*/'

e.g the following will only allow physical and aggregated major interfaces on JUNOS, but not their sub interfaces
$config['os']['junos']['bad_if_regexp']='/.*/'
$config['os']['junos']['good_if_regexp']='/^(gx)e-[0-9]+\/[0-9]+\/[0-9]+$/'
$config['os']['junos']['good_if_regexp']='/^ae[0-9]+$/'

#### Please note

#### Testers

### DIFF
--- a/doc/Developing/os/Settings.md
+++ b/doc/Developing/os/Settings.md
@@ -66,9 +66,12 @@ bad_iftype: # ifType (substring)
     - sonet
 good_if: # ignore all other bad_if settings ifDescr (substring, case insensitive)
     - virtual
-bad_ifoperstatus # IfOperStatus (substring, case insensitive)
+good_if_regexp: # ifDescr (regex, case insensitive). 
+	- "/ens[0-9]+/"
+bad_ifoperstatus: # IfOperStatus (substring, case insensitive)
     - notPresent
 ```
+For Whitelisting on Interface ifDescr see [Whitelisting intefaces](../../Support/Configuration.md#whitelisting-interfaces).
 
 ### Controlling interface labels
 
--- a/doc/Support/Configuration.md
+++ b/doc/Support/Configuration.md
@@ -655,6 +655,10 @@ definitions (includes/definitions/_specific_os_.yaml) can contain
 bad_if\* arrays, but should only be modified via pull-request as 
 manipulation of the definition files will block updating. 
 
+The default filtering mode is to blacklist based on ifDescr. 
+Whitelisting interfaces is also possible. 
+See [Whitelisting intefaces](../Support/Configuration.md#whitelisting-interfaces) later in this section for more information.
+
 Examples:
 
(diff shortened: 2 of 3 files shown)


murrant:
:/ piling more complex on this overly complex system does not seem ideal

philr2:
I'm unclear on which part is complex to you?
Do you mean Librenms is overly complex or the functions that parse the discovery of the interface?

I would argue that filtering on interface discovery is a requirement for an NMS. My addition is to allow for more parsing of the interface discovery process. At the moment it allows black listing. Black listing all uninteresting interfaces is not a good practice. White listing is something that makes more sense to me in some circumstances.

murrant:
I agree, filtering is great, but it should not be hard.
Right now LibreNMS has 7 negative filtering options and 2 positive filtering options.

philr2:
Right well where to from here?
My addition should be self explanatory. Its 5 lines of code. Unless I'm missing something very obvious, I honestly can't fathom why you think this is any harder than what already present.

philr2:
OK so I've done a bit more digging around. Would you think the best way to whitelist interfaces is to discover all and then use "Enable selected port polling"? to only poll the interfaces?

murrant:
If I had to throw something out there:
```
if_filter.direction=whitelist|blacklist
if_filter.filters = [
'ifName' => ['/Gigabit/'],
'ifSpeed' => [1000, 1000000]
]
```

Jellyfrog:
@philr2 tests fails

philr2:
OK so I've pushed another commit. This time with some documentation around it. I kept to the existing filtering functions as much as I could while including a new if_filter_mode definition.

murrant:
Getting warmer, we have a way to deprecate variables:
https://github.com/librenms/librenms/blob/master/LibreNMS/Config.php#L432

murrant:
Closing due to inactivity, feel free to reopen

PR #12502, opened 2021-02-05
Add update_device_attrib api

TridTech:
Add an endpoint to the API for modifying attributes such as override_icmp_disable, and override_sysContact_bool.  Supports changing multiple attributes at a time.

Adds the route to the new endpoint and the documentation to support it.

#### Please note

#### Testers

### DIFF
--- a/doc/API/Devices.md
+++ b/doc/API/Devices.md
@@ -1264,6 +1264,51 @@ Output:
 ]
 ```
 
+### `update_device_attrib`
+
+Update devices attributes in the database.
+
+Route: `/api/v0/attribs/:hostname`
+
+- hostname can be either the device hostname or id
+
+Input (JSON):
+
+- attrib_type: The attribute type within the database (can be an array of fields)
+- attrib_value: The attribute value to update the column with (can be an array of data))
+
+Examples:
+
+```curl
+curl -X PATCH -d '{"attrib_type": "override_icmp_disable", "attrib_value": "false"}' -H 'X-Auth-Token: YOURAPITOKENHERE' https://librenms.org/api/v0/attribs/localhost
+```
+
+Output:
+
+```json
+[
+    {
+        "status": "ok",
+        "message": "Device Attrib override_icmp_disable field has been updated"
+    }
+]
+```
+
+```curl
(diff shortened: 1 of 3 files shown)


includes/html/api_functions.inc.php
```diff
@@ -1844,6 +1844,34 @@ function rename_device(Illuminate\Http\Request $request)
...
+    // use hostname as device_id if it's all digits
+    $device_id = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
+    $data = json_decode($request->getContent(), true);
+    if (empty($data['attrib_type'])) {
+        return api_error(400, 'Device Attrib field to patch has not been supplied');
+    }
+    if (is_array($data['attrib_type']) && is_array($data['attrib_value'])) {
+        if (count($data['attrib_type']) == count($data['attrib_value'])) {
+            for ($x = 0; $x < count($data['attrib_type']); $x++) {
+                if (dbUpdate(['attrib_value' => $data['attrib_value'][$x]], 'devices_attribs', '`device_id`=? and `attrib_type`=?', [$device_id, $data['attrib_type'][$x]]) < 0) {
```
murrant, line 1859: Please don't use the legacy db commands.  https://laravel.com/docs/8.x/eloquent
The Device model also has attrib helpers on it.

doc/API/Devices.md
```diff
@@ -1264,6 +1264,51 @@ Output:
...
+Update devices attributes in the database.
+
+Route: `/api/v0/attribs/:hostname`
+
+- hostname can be either the device hostname or id
+
+Input (JSON):
+
+- attrib_type: The attribute type within the database (can be an array of fields)
+- attrib_value: The attribute value to update the column with (can be an array of data))
```
murrant, line 1278: Why not use key value pairs instead of two separate arrays?

includes/html/api_functions.inc.php
TridTech: To be honest, I started with a copy of the update_device API, and adapted it from there.  The device helpers for attribs is a better spot to work from.  I'll look at re-writing for that.

doc/API/Devices.md
TridTech: The device api uses a field: name, data: value format.  I figured it was best to stay in the style of the existing api.

TridTech:
Cancelling/closing for the time being until I have time to refactor to newer methods.

PR #12520, opened 2021-02-10
Airtime graphing for Aerohive 

tandrews13:
Looking for feedback:

Added code to existing HiveosWireless.php to poll 4 airtime oids (ahRadioTxAirtime and ahRadioRxAirtime for wifi0 and wifi1).

Polling creates two rrd files with two data sources each.  Also added two custom graphs that use CDEFs to graph Total Airtime as a percentage.  Still monitoring graphs against periodic AP observations to ascertain reasonableness of graphic representations.

Also important:  The AH-INTERFACE_MIB lists ahRadioTxAirtime and ahRadioRxAirtime as Counter64.  However, running snmpwalk on those oids shows that they are Counter32.  So I had to edit the mib to get this to work.

I defined the graphs in config,php ... not sure where they should be defined to be useful for others.

#### Please note

#### Testers


murrant:
config_defintiions.json is where you would define the graph.

I don't think this is the correct place. If you don't have this in the wireless sensor section like other devices, most people will not find it and get the benefit.

We already have wireless utilization that is a percent.  The main problem is that is defined as a guage.  We do have an example of a counter type, see wireless_errors.


Options.
1. Add a new wireless type that defines airtime utilization as a counter.  (I don't like this one as much)
2. Adapt the wireless sensor code so it can accept type (guage or counter) on a case by case basis and use the standard wireless utilization.

tandrews13:
> config_defintiions.json is where you would define the graph.
> 
> I don't think this is the correct place. If you don't have this in the wireless sensor section like other devices, most people will not find it and get the benefit.
> 
> We already have wireless utilization that is a percent. The main problem is that is defined as a guage. We do have an example of a counter type, see wireless_errors.
> 
> Options.
> 
> 1. Add a new wireless type that defines airtime utilization as a counter.  (I don't like this one as much)
> 2. Adapt the wireless sensor code so it can accept type (guage or counter) on a case by case basis and use the standard wireless utilization.

Thank you.  I see some other issues with my CDEF math as well.  

I want to go back and keep working on this.  Should I withdraw my pull request?

murrant:
Your choice.  If you mark it as draft we won't attempt to review it.

tandrews13:
Thanks for the reply.  I need to spend some time looking at your
recommendations.

I would very much like to contribute to this project.  LibreNMS is
awesome.  It would be nice to have a brief conversation with you or someone
to understand how some of these pieces fit together, but I understand that
you are busy (aren't we all!).  I'm learning quite a bit, so I'll continue
to plug away as I have time to make changes in a way that will be useful to
others.

murrant:
@tandrews13 We really do appreciate every contribution!  However as you mentioned time is short.  I've considered doing a livestream or something that could be recorded.  Feel free to ask questions on Discord.  I may not answer right away, but I try to respond to everything.

PR #12521, opened 2021-02-10
GPS coordinates from device

murrant:
in yaml or php
define for epmp, airos, and airos-af-ltu
quirk for airos bug with - in the middle of the number

#### Please note

#### Testers


app/Models/Device.php
```diff
@@ -324,15 +324,33 @@ public function getAttribs()
...
+        if (! $location->location) {
+            $this->location_id = null;
+
+            return;
+        }
+
+        $coord = array_filter($location->only(['lat', 'lng']));
+        if (! $this->relationLoaded('location') || optional($this->location)->location !== $location->location) {
+            if (! $location->exists) {
+                $location = Location::firstOrCreate(['location' => $location->location], $coord);
```
Jellyfrog, line 345: cant we just do `save()` here?

app/Models/Device.php
```diff
@@ -324,15 +324,33 @@ public function getAttribs()
...
-        if ($location_text) {
-            $location = Location::firstOrCreate(['location' => $location_text]);
+        if (! $location->location) {
+            $this->location_id = null;
+
+            return;
+        }
+
+        $coord = array_filter($location->only(['lat', 'lng']));
+        if (! $this->relationLoaded('location') || optional($this->location)->location !== $location->location) {
```
Jellyfrog, line 343: I must be missing something here,
in what cases would the location relation be loaded already? and why would it matter for this?

app/Models/Device.php
```diff
@@ -324,15 +324,33 @@ public function getAttribs()
...
+            if (! $location->exists) {
```
Jellyfrog, line 344: when would it exist? since you create it with `new Location` it never exists?

murrant: Correct, but people can implement custom location fetching in the OS PHP.  This simple check avoids an sql query in that case.  Could remove it, but I think it is better to have it there than not.

app/Models/Device.php
murrant: If custom os code loads the relation, we want to use that instead of fetching it from the DB.
Or if the new location is already saved to the DB, we can just use that.

murrant: I tend to think of code that normal contributors will write as a barrier.  Anything can happen outside that barrier :)

PR #12538, opened 2021-02-16
Use detected rrdtool_version as fallback

crazy-max:
Follow-up librenms/docker#166 librenms/docker@ab027d7

Currently [Validation of RRD](https://github.com/librenms/librenms/blob/20b4215204d2494430195e5bac306fc3b2d1fc47/LibreNMS/Validations/Rrd.php#L47-L48) specify that we can comment `rrdtool_version` config so I assume it will use the one installed but it will [fallback to `1.4` instead](https://github.com/librenms/librenms/blob/20b4215204d2494430195e5bac306fc3b2d1fc47/LibreNMS/Data/Store/Rrd.php#L63) which can't handle new features.

This PR allows to use detected rrdtool version as fallback instead of a static value.

### DIFF
--- a/LibreNMS/Data/Store/Rrd.php
+++ b/LibreNMS/Data/Store/Rrd.php
@@ -60,7 +60,9 @@ class Rrd extends BaseDatastore
             ' RRA:MAX:0.5:1:2016 RRA:MAX:0.5:6:1440     RRA:MAX:0.5:24:1440     RRA:MAX:0.5:288:1440 ' .
             ' RRA:LAST:0.5:1:2016 '
         );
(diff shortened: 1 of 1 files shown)


Jellyfrog:
No this function takes way to much time to run.

crazy-max:
@Jellyfrog Yes indeed. I see [Validator.getVersion method allows to cache the result](https://github.com/librenms/librenms/blob/01019cfc020447dfbb38e09b73cb9191d87392e1/LibreNMS/Validator.php#L232-L249). Do you think we could use it here?

Jellyfrog:
Closing in favor of #12539

PR #12549, opened 2021-02-22
Add BGP discovery and polling for Dell OS10 devices

waddles:
Adds support for BGP stats on Dell OS10 devices that use Dell Enterprise MIBs (DELLEMC-OS10-BGP4V2-MIB) instead of standard BGP4-MIB.

The DELLEMC-OS10-BGP4V2-MIB is almost exactly the same as the ARISTA-BGP4V2-MIB so the polling changes are basically a copy/paste with prefix replacements.

The discovery code is also very similar to the existing code for `timos` devices. I am not sure if there are any Dell devices that support multiple VRFs so I wrote it to support multiple, falling back to the default `1` if there are no vrfs discovered.

#### Please note

#### Testers


Jellyfrog (approved):
Thanks

PR #12554, opened 2021-02-23
Notifications : display sensor state textual value

PipoCanaja:
Display state textual value in notification list.
Capture d’écran 2021-02-23 à 10 40 07

Text was `Current Value: 2 (state)` before patch

#### Please note

#### Testers

### DIFF
--- a/includes/html/functions.inc.php
+++ b/includes/html/functions.inc.php
@@ -963,10 +963,13 @@ function alert_details($details)
         }
 
         if ($tmp_alerts['sensor_id']) {
-            $details = 'Value: ' . $tmp_alerts['sensor_current'] . ' (' . $tmp_alerts['sensor_class'] . ')<br>  ';
             if ($tmp_alerts['sensor_class'] == 'state') {
+                // Give more details for a state (textual form)
                 $details = 'State: ' . $tmp_alerts['state_descr'] . ' (numerical ' . $tmp_alerts['sensor_current'] . ')<br>  ';
-            }
+            } else {
+                // Other sensors
+                $details = 'Value: ' . $tmp_alerts['sensor_current'] . ' (' . $tmp_alerts['sensor_class'] . ')<br>  ';
+            ]
             $details_a = [];
 
             if ($tmp_alerts['sensor_limit_low']) {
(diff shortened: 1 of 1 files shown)


SourceDoctor:
how does it look before?

PipoCanaja:
It was `Current Value: 2 (state)` before patch. So the textual value of the state was not visible at all. Now it is (and the int value is still there in case needed).

SourceDoctor:
Screenshot ? :-)
for better imagination

PipoCanaja:
Don't have any unfortunately :)  The best I can provide is the line of text that changed

Text before: 
`Current Value: 2 (state)`
Text after: 
`State: needManualSync (value 2)`

Capture after (so you know where to change the text)
Capture d’écran 2021-02-23 à 10 40 07

includes/html/functions.inc.php
```diff
@@ -963,7 +963,10 @@ function alert_details($details)
         }
 
         if ($tmp_alerts['sensor_id']) {
-            $details = 'Current Value: ' . $tmp_alerts['sensor_current'] . ' (' . $tmp_alerts['sensor_class'] . ')<br>  ';
+            $details = 'Value: ' . $tmp_alerts['sensor_current'] . ' (' . $tmp_alerts['sensor_class'] . ')<br>  ';
```
SourceDoctor, line 966: why not this in a else part of the if clause ?

PipoCanaja: For a while, the style check did not like else statements. Don't know if it now stopped complaining about this. But I avoid else statements when easily avoidable. 
But both are completely equivalent for sure.

SourceDoctor: those check is not more included :-)
"else" is/was fine everytime, even if a style checker was against it ;-)

PipoCanaja: True, but that variant is also fine :) got used to it :) So unless you want me to remove it ... 

To solve this case in a proper way, I still have in my todo list to use blades and define this piece of text in the alert list directly from them. Each rule would then be able to adapt how it is displayed in the list, with all but only the necessary details.  But not today :)

SourceDoctor: yeah changing stuff to blade is a big task :-)
nethertheless i would recommend it to have it this way:

if ($tmp_alerts['sensor_class'] == 'state') 
   detail = ...
else
   detail =


change it and you'll get my approve
deal? ;-)

PipoCanaja: That's a negotiation... 
But agreed :D 

Was lazy to do a `git checkout` to change :)

SourceDoctor (approved):
LGTM

PR #12563, opened 2021-02-25
Nokia sap improvements

Channing-Jiang:
This pull request is for Nokia sap improvements.

1, Sum the ingress and egress traffic.
2, Calculate and show the VLAN values when the port is in access without encapsulation or Dot1q or QinQ mode.

Previous page:
Screenshot from 2021-02-25 20-40-04

After the changes:
Screenshot from 2021-02-25 20-35-04

3, add the title to sap graph
Previous page:
Screenshot from 2021-03-03 16-13-51
 After the changes:
Screenshot from 2021-03-03 16-13-59

4, add the sap API for the sap port's graph

#### Please note

#### Testers


Jellyfrog:
This seems to change the output, so can you please generate new test data?

Jellyfrog:
@craig-nokia any input?

craig-nokia:
> @craig-nokia any input?

Patch is good when you look at individual routers sap encapsulation.
http://localhost/device/63/routing/proto=mpls/view=saps/

But for the global sap view with all the routers, it is still showing the wrong encapsulation.
For the http://localhost/routing/protocol=mpls/view=saps/

I assume that updates to includes/html/pages/routing/mpls.inc.php should be applied.

Channing-Jiang:
@craig-nokia Thanks, forgot to update the global sap view. Just done that.

craig-nokia:
> @craig-nokia Thanks, forgot to update the global sap view. Just done that.

Patch is good to go.  

But, I don't understand why the tests failing are for a simple update to a webpage display.

includes/html/pages/device/routing/mpls.inc.php
```diff
@@ -469,10 +469,19 @@
...
         }
 
+        $outerEncapValue = (($sap['sapEncapValue'] % 65536) == 4095 ? '*' : ($sap['sapEncapValue'] % 65536));
+        $innerEncapValue = (intval($sap['sapEncapValue'] / 65536) == 4095 ? '*' : intval($sap['sapEncapValue'] / 65536));
+        $innerEncapValue = ($innerEncapValue == '0' ? '' : ".$innerEncapValue");
+        $showSapEncapValue = "$outerEncapValue$innerEncapValue";
+
+        if ($showSapEncapValue == 0 and $showSapEncapValue != '*') {
+            $showSapEncapValue = '';
+        }
```
Jellyfrog, line 479: Since this now exists in two places, please move it to a function

includes/html/pages/device/routing/mpls.inc.php
```diff
@@ -469,10 +469,19 @@
             $operstate_status_color = 'danger';
         }
 
+        $outerEncapValue = (($sap['sapEncapValue'] % 65536) == 4095 ? '*' : ($sap['sapEncapValue'] % 65536));
+        $innerEncapValue = (intval($sap['sapEncapValue'] / 65536) == 4095 ? '*' : intval($sap['sapEncapValue'] / 65536));
+        $innerEncapValue = ($innerEncapValue == '0' ? '' : ".$innerEncapValue");
+        $showSapEncapValue = "$outerEncapValue$innerEncapValue";
+
+        if ($showSapEncapValue == 0 and $showSapEncapValue != '*') {
```
Jellyfrog, line 477: use `&&` instead of and

craig-nokia:
Reapplied the patch and looks good.

Channing-Jiang:
@Jellyfrog Thanks for the advise and review. I did another patch for the sap graph page to show the correct title, should I open another pull request or just push it to here?

Jellyfrog:
You're adding a new database field but no sh schema migration created for it.
But I'm not so sure we should add this field since it seems we can just calculate it on the fly

Jellyfrog (changes requested):
You're adding a new database field but no sh schema migration created for it.
But I'm not so sure we should add this field since it seems we can just calculate it on the fly

Channing-Jiang:
@Jellyfrog Yes, with this new database field, I am able to get the readable VLAN value for not only the encapsulation display but also the sap graph title and the API.

I am ok to just use the function to calculate it and function to reverse calculate the value, but it means that we need to run the functions all the time.

craig-nokia:
Let me know when you want me to test.  I have around 175 devices with ~2500 SAPs.   IMHO, using the database field would be faster.  But, who am I to know what is best.

murrant:
You need to rebase this and I agree that the display calculation is trivial and doesn't seem like it should add much to web page display.   We are generally more concerned about polling time than web loading time.  Polling usually happens more frequently.

Channing-Jiang:
@murrant Sorry, just saw this, so rebase those stuff without adding a new database field right?

murrant:
@Channing-Jiang I merged, so you don't need to rebase any more.  Just add the database migration.  (and probably still needs a review after that)

Jellyfrog:

PR #12566, opened 2021-02-25
Modernize pdf support

Oirbsiu:
Adding support for Larvel dompdf - https://github.com/barryvdh/laravel-dompdf as suggested by jellyfrog.
 

#### Please note

#### Testers


app/Http/Controllers/PDFController.php
```diff
@@ -0,0 +1,81 @@
...
+namespace App\Http\Controllers;
+
+use Illuminate\Http\Request;
+use DOMPDF;
+use LibreNMS\Config;
+use Illuminate\Support\Facades\Route;
+use LibreNMS\Util\DynamicConfig;
+
+require '/opt/librenms/includes/init.php';
+require_once '/opt/librenms/includes/html/functions.inc.php';
```
Jellyfrog, line 12: these are not allowed, if you need functions from the legacy code you need to port them to modern classes.
There are many examples of that inside that file, see for example;
https://github.com/librenms/librenms/blob/6c3dcfd871064337a603844158f54aa7c8a9d47c/includes/html/functions.inc.php#L61-L64

app/Http/Controllers/PDFController.php
```diff
@@ -0,0 +1,81 @@
...
+		// this is the URL used in the button "Create PDF" - probably can be done another way but this works.... 
+		$route	= '/dompdf/generate/'. $clean_path;
+
+		$header_logo = \LibreNMS\Config::get('dompdf.header_logo');
+		$footer_logo = \LibreNMS\Config::get('dompdf.footer_logo');
+		
+		// Validate that the view exists 
+		if (view()->exists($view)) {
+			// Run report and return JSON object - saved to $json
+			include_once "/opt/librenms/includes/html/reports/$clean_path.pdf.inc.php";
```
Jellyfrog, line 29: not allowed

includes/html/reports/bills.pdf.inc.php
```diff
@@ -0,0 +1,96 @@
+<?php
```
Jellyfrog, line 1: this needs to be converted to a class or so, and use eloquent for sql queries

Jellyfrog (changes requested):

app/Http/Controllers/PDFController.php
Oirbsiu: Thanks for looking -
If I omit lines 11 & 12 these, I get this error
#Error
Call to undefined function dbFetchRows()

this was the only way i could get the database queries to work in the file referenced at line 29 
(in this case its inlcudes/html/reports/bills.pdf.inc.php)

which you say below is not allowed :(
how can I access dbFetchRows - what have I missed?

app/Http/Controllers/PDFController.php
Oirbsiu: should I run the Database query here instead?

app/Http/Controllers/PDFController.php
murrant: No queries should use the dbFacile functions.  See https://laravel.com/docs/8.x/eloquent   For example for bills see App\Models\Bill

Jellyfrog:
Top of my mind;
- This should be a http controller with routes
- Need to replace the functionality of `html/pdf.php` and `includes/html/reports/...`
- Remove `html/pdf.php` and `includes/html/reports/`
- Remove `tecnickcom/tcpdf` dependency

Oirbsiu:
I had to start again with my changes as i kinda broke it. 
Anyway, managed to get this working but ill admit that I am far from understanding the right way to do things.

What Ive done is updated the existing PDF code based on the examples in  /opt/librenms/includes/html/reports

added new routes 
(get) pdf/generate/{view}
(get) pdf/preview/{view}
(get) pdf/alerts{params}
(post) pdf/Alerts

in /opt/librenms/includes/html/reports is have added bills - this is basically replicating the existing code for displaying current and past bills. only saves to PDF. 

Its not perfect as I have some weird formatting happening from page 2 related to the progress bar.

in order to get this working using eloquent etc, I had to improvise a little - this is where my knowledge is limited - I took the what is in includes/common.php and copied this to LibreNMS/Common/Functions.php

Its not perfect I know and will need help correcting my knowledge gaps but its working for me.
 
Only other question that I have is  how do i merge this request with the work here - https://github.com/Oirbsiu/librenms/tree/adding-dompdf

murrant:
This request is against your master branch, so options are limit.  You might need to close this and open a new one from your adding-dompdf branch.  (reference this PR there so we can follow the conversation)

Oirbsiu:
Im not sure if what I have done is correct - still learning how to use GIT. Sorry.
When I created this PR I kinda made a mess of things so I ended up deleting my clone and starting again. 
I did all my changes and did what I thought was right and pushed these to a branch I named 'adding-dompdf' 
I have now just merged these on the clone i made. so i think all my changes are in https://github.com/Oirbsiu/librenms

These changes have not been picked up here so I guess I need to close this PR and create a new one.

PR #12575, opened 2021-03-01
Add no-db-cleanup argument in daily.sh

Negatifff:
I am using distributed pollers. and one database. each poller maintains the current librenms code and cleans the database after each update. if many pollers are updated, everyone will clean up the database. running the daily update script with this key will avoid such situations. but we must remember that the database should still be cleaned with a daily update, but now it will be possible to do it only from one place.

### DIFF
--- a/daily.sh
+++ b/daily.sh
@@ -342,7 +342,7 @@ main () {
                 status_run 'Updating SQL-Schema' 'php includes/sql-schema/update.php'
                 status_run 'Cleaning up DB' "$DAILY_SCRIPT cleanup"
             ;;
-            post-pull)
+            post-pull|no-db-cleanup)
                 # re-check dependencies after pull with the new code
                 check_dependencies
 
@@ -366,7 +366,9 @@ main () {
                 # List all tasks to do after pull in the order of execution
                 status_run 'Updating SQL-Schema' 'php includes/sql-schema/update.php'
                 status_run 'Updating submodules' "$DAILY_SCRIPT submodules"
-                status_run 'Cleaning up DB' "$DAILY_SCRIPT cleanup"
+                if [ "$arg" != "no-db-cleanup" ] ; then
+                    status_run 'Cleaning up DB' "$DAILY_SCRIPT cleanup"
+                fi
                 status_run 'Fetching notifications' "$DAILY_SCRIPT notifications"
                 status_run 'Caching PeeringDB data' "$DAILY_SCRIPT peeringdb"
             ;;


murrant:
@Negatifff what is the downside of the cleanup being run on every poller?

Negatifff:
> @Negatifff what is the downside of the cleanup being run on every poller?
the most harmless is the lack of meaning. more seriously - a procedure performed N times instead of one. the removal procedure. before writing the answer, I made a test bench of two virtual machines on which pollers are deployed. update script is run simultaneously on both pollers. the poller who first reached the database cleaned the tables. cleaned them for 15 minutes. all these 15 minutes the second poller was waiting to be allowed to clean the same table that the first poller was cleaning. that is, to do unnecessary work. what for?
I now have 16 pollers running and have only covered a third of my network. I will need approximately 50 pollers. it turns out that a daily update will entail 50 database cleanups. why, if you can only do one?

PipoCanaja:
You have a point. 
A safer way would be that:
- All pollers keep trying to do the DB cleaning but only the 1st one that starts it actually does it. 
- All others will skip the task instead of waiting and doing it after.
- A minimum interval could be defined to ensure that no other cleanup occurs too early. That way, if your pollers are not running daily.sh at the same time of the day, you can still ensure only a limited amount of DB cleaning will occur. 20 hours could be a good default value for this.

Negatifff:
> You have a point.
> A safer way would be that:
> 
> * All pollers keep trying to do the DB cleaning but only the 1st one that starts it actually does it.
> * All others will skip the task instead of waiting and doing it after.
> * A minimum interval could be defined to ensure that no other cleanup occurs too early. That way, if your pollers are not running daily.sh at the same time of the day, you can still ensure only a limited amount of DB cleaning will occur. 20 hours could be a good default value for this.

and then a reasonable question arises - why is the daily update combined with cleaning the database?

murrant:
@Negatifff only first cleanup should take any significant amount of time, the rest should be instant or lock time out and skip. 15 minutes is long, are you sure you don't have an issue?

I would much rather see bug fixes if what I stated is not true.

Negatifff:
> @Negatifff only first cleanup should take any significant amount of time, the rest should be instant or lock time out and skip. 

I don't want to sound boring, but why is it? Why do you need to perform more than one action if only one will be executed anyway? 
I would suggest doing database cleanup from one place. 

> 15 minutes is long, are you sure you don't have an issue?

I hope so. I have:
- the default setting for storing alert logs is 365 days
- table with 6 million records

deleting records from this table takes quite a long time.

murrant:
I hear what you are saying, but these are my two sides of this.
Increase possibility of LibreNMS daily cleanup not being run by accepting this or let the clean up run multiple times deleting 0-5 records for each subsequent cleanup.

I would be very happy to see improvements in this area as long as it doesn't increase the likelyhood of daily cleanup not being run.  When daily is not run LibreNMS tanks and it is a bad experience.

One simple solution would be to reduce the lock timeout.

If you wanted on your instance, you could remove daily.sh from cron and add the  commands to manually update and run daily.php -f for each of the things daily.sh does.

Again, we already have too many people who blow up their LibreNMS instance by not running daily.sh come for support.

Negatifff:
actually my solution is not the best either. The best solution is to separate the update and cleanup of the database.
and start cleaning the database with a separate script in cron. I read that some users disable or do not use the update script, but this is another reason for entities separation.

murrant:
I don't disagree with that

Negatifff:
> I don't disagree with that

ok

PR #12577, opened 2021-03-01
Developer device simulation

murrant:
lnms dev:simulate
handy helper to start snmpsim with test data and optionally add and remove a device to LibreNMS

#### Please note

#### Testers

### DIFF
--- a/app/Console/Commands/DevSimulate.php
+++ b/app/Console/Commands/DevSimulate.php
@@ -1,0 +1,131 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Console\LnmsCommand;
+use App\Models\Device;
+use Illuminate\Support\Str;
+use LibreNMS\Util\Snmpsim;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Process\Process;
+
+class DevSimulate extends LnmsCommand
+{
+    protected $name = 'dev:simulate';
+    protected $developer = true;
+    /**
+     * @var Snmpsim
+     */
+    protected $snmpsim = null;
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Simulate devices using test data';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->addArgument('file', InputArgument::OPTIONAL);
+        $this->addOption('multiple', 'm', InputOption::VALUE_NONE);
+        $this->addOption('remove', 'r', InputOption::VALUE_NONE);
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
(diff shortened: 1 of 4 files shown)


app/Console/Commands/DevSimulate.php
```diff
@@ -0,0 +1,131 @@
...
+        // set up removal shutdown function if requested
+        if ($this->option('remove')) {
+            $this->queueRemoval($device->device_id);
+        }
+    }
+
+    private function queueRemoval($device_id)
+    {
+        if (function_exists('pcntl_signal')) {
+            pcntl_signal(SIGINT, function () {
```
Jellyfrog, line 107: dont you need `declare(ticks = 1);` for this to work?

app/Console/Commands/DevSimulate.php
```diff
@@ -0,0 +1,131 @@
...
+                exit(); // exit normally on SIGINT
+            });
+        }
+
+        register_shutdown_function(function () use ($device_id) {
```
Jellyfrog, line 112: we should _always_ shutdown snmpsim on exit also

app/Console/Commands/DevSimulate.php
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        $this->snmpsim = new Snmpsim();
+        $snmprec_dir = $this->snmpsim->getDir();
+        $listen = $this->snmpsim->getIp() . ':' . $this->snmpsim->getPort();
```
Jellyfrog, line 52: this should be configurable

murrant: It never has been before.  I feel like it could be detrimental to have it configurable, if you add multiple devices it won't work if you use different IPs and ports.

app/Console/Commands/DevSimulate.php
murrant: when the variable is deleted, the process should be terminated

app/Console/Commands/DevSimulate.php
murrant: Honestly, no.  We don't need to react to the ctrl-c immediately.  It worked in testing.

app/Console/Commands/DevSimulate.php
Jellyfrog: I maually change it sometimes, so i think it should be configurable

Jellyfrog:
Do we want to print stdout also?

app/Console/Commands/DevSimulate.php
Jellyfrog: Seems it does, old one didnt.

murrant:
@Jellyfrog I'm going to merge this as is.  If you want to add additional CLI options, please do :)

app/Console/Commands/DevSimulate.php
murrant: `./scripts/save-test-data.php --snmpsim` does not allow that to be changed... this is a replacement for that.

PR #12582, opened 2021-03-04
Add Packet Buffers as memory for PanOS

shepherdjay:
I've added the packet buffers and descriptors dataplane for PanOS.

#### Please note

- [x ] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [x ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers


shepherdjay:
This first attempt at this is unsatisfactory and am looking for input. The primary concern I have is while SNMP reports the value as an integer (as part of the hrStorageTable) ultimately as a user I would want the value and high_limit as a percentage.

This can be seen via other disks that show up natively. I'm uncertain though how to properly modify this for that type of behavior where it can be considered a load percentage.

murrant:
@shepherdjay should this be considered memory?

shepherdjay:
> @shepherdjay should this be considered memory?

I do think it is probably closer to a type of memory over storage or the count sensor as I've implemented here. I tried looking for examples from other network gear to see if anyone had implemented packet buffers and what type but I was unsuccessful in my search.

If I modify this to implement it as a custom mempool should I attempt to do that via the yaml file or since `Librenms\OS\Panos.php` already exists should I be editing that file to implement `discoverMempools` and `pollMempools`

murrant:
Except this isn't even a custom mempool it just uses the standard HR MIB.  Just need to update HostResources.php to include this in the heuristic which determines which ones are memory vs storage.

shepherdjay:
> Except this isn't even a custom mempool it just uses the standard HR MIB. Just need to update HostResources.php to include this in the heuristic which determines which ones are memory vs storage.

@murrant - I guess I had some trepidation about making a change that could affect more than the target device os. Still I took a crack at a working implementation and checked my search strings against the snmprec files available for unexpected matches. I didn't find any unexpected.

I attempted to allow for partial matches in this as the exact description regarding the slot and dataplane numbers I believe will be different depending on the model palo alto deployed and how many slots are filled out. It didn't make sense to me to clutter a shared heuristic with a laundry list of OS Specific descriptions but I can change that if desired.

murrant:
No need to do a substring search, if others need  to be added in the future, we can update them.
Can you add test data so I can see what it looks like?

shepherdjay:
@murrant - I have added some test data for the 5200 series in non-vsys mode. The existing test data is just for standard mode and vsys mode for all panos. This is probably correct but I wasn't clear if I should be replacing that or adding new so for now I just created a separate variant. I can overwrite the old versions from my lab unit if preferred.

murrant:
@shepherdjay Hey, I thought of a much better way to implement this.

In Panos.php add the function:
```php
protected function memValid($storage) {
    if (parent::memValid($storage) {
        return true;
    }

    return true;  // insert appropriate logic here instead of true
}
```

shepherdjay:
@murrant 

I ended up implementing the custom logic and then just calling the parent function. I'm not sure which you would prefer, the if logic or this. Does appear to be working though and I'll squash these down.

LibreNMS/OS/Panos.php
```diff
@@ -133,4 +138,17 @@ public function pollOS()
             $this->enableGraph('panos_activetunnels');
         }
     }
+
+    protected function memValid($storage)
+    {
+        if ($storage['hrStorageType'] == 'hrStorageOther') {
+            foreach ($this->validNetBufferMemory as $netBufferMemory) {
```
murrant, line 145: Awesome, looking great.  Just one more tweak:
Str::contains()  accepts an array of needles, so we don't need  a loop.

shepherdjay:
Thanks @murrant - Is there anything further needed to merge this?

PipoCanaja (changes requested):
Hi @shepherdjay 
Concerning tests, it would be better to keep the existing snmprec (may be rename it with model name in it) and add you own new snmprec. Make it clearer what changed for existing test data. And then run save-test-data.php on all 3 of them. We would end up with the 2 new test files using old existing snmprec and your completely new test file, making a total of 3 here.

shepherdjay (commented):
I reverted to the old snmprec test data (the one I generateed is not substantially different as all palos' platforms run the same os type) - and resaved the test data. panos.json changes appears large as the previous test data file had sensors first then os... it regenerated them in the opposite order. On diff though only change to test data for both files is the expected mempools added.

PipoCanaja (approved):
LGTM, Thanx @shepherdjay

PR #12594, opened 2021-03-08
Use high precision OIDs for APC UPS sensors

CirnoT:
Based on #12158, keeps previous index for sensors and tries to use old low-precision OIDs if high precision ones are not available.

---

Note, that this keeps low-precision index on sensors even when high-precision is detected; only OID and divisor are updated when high precision OID is used. This is done so that previous data is not removed (sensor is recognized by its unique index). 

The alternative is to update index too, but this will cause LibreNMS to remove old sensor along with its historical data. Do we want to do that (it would be one-time change for now and would additionally happen every time user would upgrade/downgrade between firmware revisions)?

---

#### Please note

#### Testers


Jellyfrog:
What we really want is for this to be in yaml format.
It's basically as easy as adding;
```yaml
-
    oid: atsInputFrequency
    num_oid: '.1.3.6.1.4.1.318.1.1.8.5.3.2.1.4.{{ $index }}'
    index: 'atsInputFrequency.{{ $index }}'
    descr: 'Input Feed'
    divisor: 1
```
to `includes/definitions/discovery/apc.yaml`

CirnoT:
Why won't you do this yourself if it's so easy then? I have no issues running modified variant for myself if help is not welcome here.

kkrumm1:
I agree - I have been kinda quiet with this project due to JellyFrog's responses. Jellyfrog is a smart dude and helps a lot! But one thing you can work on is your responses to people. It's not very welcoming and condescending at times.

Jellyfrog:
Here is a complete example of what youre doing but in yaml

```yaml
                -
                    oid: atsInputFrequency
                    num_oid: '.1.3.6.1.4.1.318.1.1.8.5.3.2.1.4.{{ $index }}'
                    index: 'atsInputFrequency.{{ $index }}'
                    descr: 'Input Feed'
                    divisor: 1
                    skip_value_lt: 0
                -
                    oid: atsOutputFrequency
                    num_oid: '.1.3.6.1.4.1.318.1.1.8.5.4.2.1.4.{{ $index }}'
                    index: 'atsOutputFrequency.{{ $index }}'
                    descr: 'Output Feed'
                    divisor: 1
                    skip_value_lt: 0
                -
                    oid: upsHighPrecInputFrequency
                    num_oid: '.1.3.6.1.4.1.318.1.1.1.3.3.4.{{ $index }}'
                    index: 'upsHighPrecInputFrequency.{{ $index }}'
                    descr: 'Input'
                    divisor: 10
                -
                    oid: upsHighPrecOutputFrequency
                    num_oid: '.1.3.6.1.4.1.318.1.1.1.4.3.2.{{ $index }}'
                    index: 'upsHighPrecOutputFrequency.{{ $index }}'
                    descr: 'Output'
                    divisor: 10
                -
                    oid: upsAdvInputFrequency
                    num_oid: '.1.3.6.1.4.1.318.1.1.1.3.2.4.{{ $index }}'
                    index: 'upsAdvInputFrequency.{{ $index }}'
                    descr: 'Input'
                    divisor: 1
                    skip_values:
                        -
                          oid: upsHighPrecInputFrequency
                          op: 'exists'
                          value: true
                -
                    oid: upsAdvOutputFrequency
                    num_oid: '.1.3.6.1.4.1.318.1.1.1.4.2.2.{{ $index }}'
                    index: 'upsAdvOutputFrequency.{{ $index }}'
                    descr: 'Output'
                    divisor: 1
                    skip_values:
                        -
                          oid: upsHighPrecOutputFrequency
                          op: 'exists'
                          value: true
```

if yo uwant to keep the history data, set `index` to the same value (I think)

PipoCanaja:
Hi guys,

If PHP allows the code to be completely backward compatible, meaning it won't break existing sensors and add high-precision OIDs for devices having them, that would be a good beginning. 

That being said, it is indeed doable in YAML, it even got simpler now that you don't even have to specify numerical OIDs anymore. Even possible to do it and be backward compatible (a little bit trickier, I agree). 

So for this time, let start with PHP, and do it well so we don't break a 2nd time APC users. And for next time @CirnoT, YAML is the way to go, simpler, shorter, in 95% of the devices. And we keep PHP for the 5% remaining. At the end of the day, YAML is less work for the initial dev, but also and more importantly for the ease of maintenance of the code.

Bye

CirnoT:
>Here is a complete example of what youre doing but in yaml

Frequency does indeed look simple, but look at Voltage and temperature...

CirnoT:
Again, before this is merged, I'd like to make sure how we should proceed: do we want to consider high/low precision as separate sensors or not? This will decide whether the index of sensor should differ between them and whether historical data is kept between these two.

Jellyfrog:
> Again, before this is merged, I'd like to make sure how we should proceed: do we want to consider high/low precision as separate sensors or not? This will decide whether the index of sensor should differ between them and whether historical data is kept between these two.

Input @murrant ?

murrant:
They should be separate sensors.  If the high precision oids exist, don't use the low precision sensors.

PR #12603, opened 2021-03-09
Install options guidance

dagbdagb:
Added some text to help prospective new users decide which installation option fits their particular situation and skillset better.

#### Please note

#### Testers

### DIFF
--- a/doc/Installation/index.md
+++ b/doc/Installation/index.md
@@ -1,15 +1,48 @@
 source: Installation/index.md
 path: blob/master/doc/
 
-# Installing LibreNMS
+# Installing LibreNMS - 3 main routes
+
 
 ## Docker
+
+Choose docker if you are somewhat familiar with docker and you just want to give the 
+official docker image a quick spin for a smoketest of LibreNMS. Or, if you are very familiar 
+with both docker and LibreNMS and want to build a custom installation.
+
+There may be simpler avenues to learn about Docker than trying to learn LibreNMS and 
+Docker at the same time.
 
 An official LibreNMS docker image based on Alpine Linux and Nginx is available 
 on [DockerHub](https://hub.docker.com/r/librenms/librenms/). Documentation can 
 be found on the [Github repository](https://github.com/librenms/docker).
 
-## Manually
+Do note that the official image (at time of writing) is respun once a month, and that 
+the native daily update mechanism in LibreNMS is disabled.
+
+
+## Virtual Machine Images
+
+These are also good for a smoketest or installs not intended to scale to 1000s of devices. 
+Daily updates should be possible with these.
+
+VirtualBox images you can use to get started:
+
+- [Virtual Machines](Images)
+
+
+## Manual install
+
+Choose manual install if you have tested either the docker image or the VM and decided that :
+
+a) You like what you see so far.
+
+and
+
+b) You need something which scales to more devices, or you want daily updates, or 
+you don't feel like learning docker right now.
+
+Or for any other reason you see fit. You are not obligated to use Docker or a VM first.
 
 If you want to install manually then we have some documentation which should make it easy.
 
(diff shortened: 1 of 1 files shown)


Jellyfrog:
Hello
Thanks for your interest in improving the docs!
for me these changes undermines the docker container as a toy which shouldn't be used. 
And basically only manual installations (with nightly updates) are the way to go, which is not really true.

dagbdagb:
Presenting the installation options as 'equal', is to oversell the docker image. In my opinion, of course. Sorry if this rubs you the wrong way. 

It is not my intention to undermine anything, only to help new users to have a good experience with LibreNMS. I assert that users inexperienced with docker or librenms is better served with a non-docker install for their production setup, as it removes one layer of complexity. They may also find more people able to help in the Discord channel.

I did write "Or, if you are very familiar with both docker and LibreNMS and want to build a custom installation.". So, no. I do not see it as a toy. It is a tool, and to use it properly you need a bit of knowledge and experience.

Happy to discuss alternative ways to express this.
Do you have other aspects of the various installation options which you think could help users make an informed choice?

murrant:
Honestly:
VM: for test/small/toy installs.  Don't use in production.
Manual install: generally results in the best result
Docker: for users that prefer docker images.

dagbdagb:
That is a fair summary of what I said, I think.

murrant:
The words you have written don't really convey that.  It seems like:
1. it suggests that you almost never use docker
2. you should not do a standard install unless you have tried vm or docker, but then later there is text to contradict that.
3. seems to imply one or more of the installs can not be updated... odd. monthly and daily updates are supported on all installation methods.  (of course the docker images don't use the build in updater)

I am definitely for any improvements we can make but what are the goals?
What order should the install option be listed?
What order do we prefer users use to install?
What methods don't we prefer users use to install?
What problems do certain install types have?
What benefits?

dagbdagb:
I do get the sense I am offending someone, and I can wholeheartedly say it was not my intention. The current text provides no guidance whatsoever to prospective users, and I think this is a flaw.

So, I am not trying to bash docker. I am merely suggesting that LibreNMS in a docker setting may not be a great choice if you know anything about either. And I believe some of us agree on that? 

What I try to convey is this:
- if your have experience with docker, great. Docker works well. 
I do believe it is entirely fair to make a note of the (current) built-in latency of updates witht the docker option?

- If you don't have experience with docker, a manual install works well. It is the the option which scales best
With a possible exception for custom docker setups . 

- If this is a very small system, or just a test drive, the VM is a great option.


I'll try to come up with a more detailed version, covering the points you make @murrant . As for:
* What order do we prefer users use to install?
* What methods don't we prefer users use to install?

... I am not sure it is our business to express *that*. It depends on the user, and it depends on the use case.

murrant:
I understand what you are trying to achieve, just that the text seemed to include some bias, been trying to point that out.

I personally would not use the VM image for any production installs.  I would suggest a regular install, but I do understand some people are afraid of the linux cli.

dagbdagb:
Got busy writing other stuff.  I'll get to this.

kkrumm1:
Too many people think virtual images are great for production environments. I personally think they should be pulled.

dagbdagb:
Heh. I find virtualization great. Although a single vm for all functions only takes you so far.

As for containers: I have now seen, at close range, a customer (trying to) implement LibreNMS in Kubernetes in Azure. I am afraid this exposure renders me entirely too biased to update this text as I intended to do. In short, I cannot in good conscience write a text which displays all three installation options as equal. They are not.

If we want to dissuade people from running the premade VM in production, why not just nerf it a bit? I do believe in 'fail fast'. Being able to discover, test and render a verdict on  a new tool (for example, LibreNMS) between morning coffee and lunch qualifies. Having a prebuilt VM enables this.

murrant:
I think the virtualization images can be good for a certain crowd with little linux experience and very small networks.

murrant:
Closing due to inactivity

PR #12605, opened 2021-03-11
Re-implement Printer as a class based module

Jellyfrog:

#### Please note

#### Testers


LibreNMS/Interfaces/Polling/PrinterPolling.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+namespace LibreNMS\Interfaces\Polling;
+
+interface PrinterPolling
+{
+    /**
+     * @return
+     */
+    public function pollPrinter();
```
murrant, line 30: should send it the models that need to be polled.

Jellyfrog: These files are actually left-overs from the none implementation of os-overrides, should I implement it or delete them? :)

murrant: ah, go ahead and delete.  Easy to re-add if needed.

murrant:
I was thinking about the state sensors.
It kind of solves the issue if this module is called "PrinterSupplies"  Then the state sensors should not be here.

database/migrations/2021_03_11_003713_rename_printer_columns.php
```diff
@@ -0,0 +1,96 @@
...
+class RenamePrinterColumns extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('printer_supplies', function (Blueprint $table) {
```
murrant, line 16: You can do these all in the same Schema::table block

includes/html/pages/device/overview/toner.inc.php
```diff
@@ -4,7 +4,7 @@
 
 $graph_type = 'toner_usage';
 
-$supplies = \App\Models\Toner::query()->where('device_id', $device['device_id'])->get()->groupBy('toner_type');
+$supplies = \App\Models\PrinterSupply::query()->where('device_id', $device['device_id'])->get()->groupBy('printer_type');
```
murrant, line 7: ->groupBy('printer_type');  this doesn't make much sense.  why did you prefix the fields with printer_?

murrant (commented):
We will need a hook in config to populate printer-supplies module discovery and poller options  from the existing toner options.

database/migrations/2021_03_11_003713_rename_printer_columns.php
Jellyfrog: Nope, sqlite doesnt support it.

includes/html/pages/device/overview/toner.inc.php
Jellyfrog: it does makes sense since the data is grouped by it :)

murrant (approved):
Looks great.

PR #12613, opened 2021-03-12
WebUI - Display interface errors per second instead of accumulated

antonio-jose-almeida:
Hello,
I had a lot of difficulty in identifying the interfaces that were generating errors in my network because the value that was shown in the table was the accumulated value. I tried to identify a way to change the code to show the delta (I even considered using the rate but I think that to identify errors the delta value is more indicated)

librenms_changes

#### Please note

#### Testers


murrant:
Like it but the name is not good.  Delta from what?  Need to be able to convey that in the title.

antonio-jose-almeida:
Hi, 
I've change the name to be more verbose. Now makes more sense?

PipoCanaja:
Hi @antonio-jose-almeida 

Really like the idea. Current display is not very usable. 

I would probably go a little further here. 
- Leave the description as `Errors In` and `Errors Out`
- Display the delta value, but computed in a "per second" (or "per minute") value
- Add a Unit to make it clear

Cause the delta value you display, is taken between 2 polls. By default it is 5 minutes. But some users use different intervals (I use 1 minutes on some servers). In order to compare, dividing this "delta" value with the poller interval would give you "errors per second". Per minute might even be a little bit easier for the user, on interfaces with a low amount of trafic (and errors)

antonio-jose-almeida:
Hello,
I tried to implement it as suggested. I think I never saw the "Eps" reference for errors per second but it was the acronym that occurred to me.

eps

PipoCanaja (approved):
LGTM

PipoCanaja (approved):
LGTM

PipoCanaja:
Thanx @antonio-jose-almeida !

synnack:
Is there a button to toggle this behaviour or have the old Errors In/Out columns back as extra columns? It's really difficult to sort by interfaces that have error peaks (but not the current polling interval) with the current behaviour.

Issue #12427, opened 2021-01-11
librenms-service.py poller crash


ntaken:
Hello!

Got this crash on the poller part of the service. After having runned librenms-service.py for little more than one hour.

Jan 11 10:36:41 srv-net101 librenms-service.py[1348043]: Poller_0-2(INFO):Completed poller run for 161 in 4.58s
Jan 11 10:36:41 srv-net101 librenms-service.py[1348043]: Traceback (most recent call last):
Jan 11 10:36:41 srv-net101 librenms-service.py[1348043]:   File "/opt/librenms/librenms-service.py", line 48, in <module>
Jan 11 10:36:41 srv-net101 librenms-service.py[1348043]:     service.start()
Jan 11 10:36:41 srv-net101 librenms-service.py[1348043]:   File "/opt/librenms/LibreNMS/service.py", line 404, in start
Jan 11 10:36:41 srv-net101 librenms-service.py[1348043]:     sleep(self.config.master_resolution)
Jan 11 10:36:41 srv-net101 librenms-service.py[1348043]:   File "/opt/librenms/LibreNMS/service.py", line 550, in reap
Jan 11 10:36:41 srv-net101 librenms-service.py[1348043]:     handler = signal(SIGCHLD, SIG_DFL)
Jan 11 10:36:41 srv-net101 librenms-service.py[1348043]:   File "/usr/lib/python3.8/signal.py", line 48, in signal
Jan 11 10:36:41 srv-net101 librenms-service.py[1348043]:     return _int_to_enum(handler, Handlers)
Jan 11 10:36:41 srv-net101 librenms-service.py[1348043]: TypeError: 'int' object is not callable
Jan 11 10:36:42 srv-net101 librenms-service.py[1348043]: Poller_0-5(INFO):Completed poller run for 117 in 4.12s
Jan 11 10:36:42 srv-net101 librenms-service.py[1348043]: Poller_0-18(INFO):Completed poller run for 169 in 6.33s
Jan 11 10:36:42 srv-net101 librenms-service.py[1348043]: Poller_0-12(INFO):Completed poller run for 385 in 4.30s
Jan 11 10:36:42 srv-net101 librenms-service.py[1348043]: Poller_0-8(INFO):Completed poller run for 100 in 4.49s
Jan 11 10:36:43 srv-net101 librenms-service.py[1348043]: Poller_0-13(INFO):Completed poller run for 298 in 4.23s
[...]

PR #12593, opened 2021-03-08
Simplify process reaping

rkojedzinszky:
Fixes #12427

Eliminated signal handler installation/deinstallation process.

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [x] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [ ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #12616, opened 2021-03-14
Potential fix for #12427

TheMysteriousX:
When testing, check that:
* the process restarts overnight with no polling gaps
* the main process does not run at 100% load
* there are no 'defunct' processes

If anyone is running a poller on Windows, macOS or BSD, it'd be interesting to see if this is portable.
#### Please note

#### Testers

### DIFF
--- a/LibreNMS/__init__.py
+++ b/LibreNMS/__init__.py
@@ -1,4 +1,5 @@
 import os
+import signal
 import subprocess
 import threading
 import timeit
@@ -18,6 +19,12 @@ def normalize_wait(seconds):
     return ceil(seconds - (time() % seconds))
 
 
+def reset_signals():
+    #https://bugs.python.org/issue38435
+    #https://bugs.python.org/issue32985
+    signal.signal(signal.SIGCHLD, signal.SIG_DFL)
+
+
 def call_script(script, args=()):
     """
     Run a LibreNMS script.  Captures all output and throws an exception if a non-zero
@@ -35,8 +42,8 @@ def call_script(script, args=()):
     base_dir = os.path.realpath(os.path.dirname(__file__) + "/..")
     cmd = base + ("{}/{}".format(base_dir, script),) + tuple(map(str, args))
     debug("Running {}".format(cmd))
-    # preexec_fn=os.setsid here keeps process signals from propagating (close_fds=True is default)
-    return subprocess.check_call(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT, start_new_session=True, close_fds=True)
+    # preexec_fn=reset_signals ensures we don't receive signals from children (close_fds=True is default, but may become false in a future release)
+    return subprocess.check_call(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT, preexec_fn=reset_signals, start_new_session=True, close_fds=True)
 
 
 class DB:
--- a/LibreNMS/__init__.py
+++ b/LibreNMS/__init__.py
@@ -36,7 +36,7 @@ def call_script(script, args=()):
     cmd = base + ("{}/{}".format(base_dir, script),) + tuple(map(str, args))
     debug("Running {}".format(cmd))
     # preexec_fn=os.setsid here keeps process signals from propagating (close_fds=True is default)
-    return subprocess.check_call(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT, preexec_fn=os.setsid, close_fds=True)
+    return subprocess.check_call(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT, start_new_session=True, close_fds=True)
 
 
 class DB:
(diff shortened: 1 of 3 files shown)


TheMysteriousX:
s/fix/workaround/, as this is possibly a runtime bug.

rkojedzinszky:
I did not check the code through, is that the only place where children are created? If yes, then perhaps there is no need to handle SIGCHLD, as subprocess module should clean up everything after a process exists. If there are other places where children can be created, then indeed a reaping process is needed.

TheMysteriousX:
If SIG_CHLD is unhandled zombie processes are created when the service reloads.

murrant:
@rkojedzinszky @TheMysteriousX any consensus on a fix?  I am not knowledgeable enough to judge.

TheMysteriousX:
After finding this docbug ( https://bugs.python.org/issue32985 ) I've cracked the issue that required the detailed signal handling in the first place.

When forking off for the subprocess, `restore_signals` (a default parameter) restores only the signals that the intepreter itself modified, i.e. `sigint`, `sigterm` and not `sigchld`.

This results signals still being received for children - I used psutil to filter these out. With the commit I'm about to push, subprocesses should be fully and completely isolated, and all deadlocks eliminated without the need for the custom reap method, and the process should restart cleanly with no zombies left around from previous invocations.

Unfortunately, the fix is only good until 3.13 at latest, as per https://bugs.python.org/issue38435 but it seems like the Python dev's are aware of the issue.

rkojedzinszky:
> If SIG_CHLD is unhandled zombie processes are created when the service reloads.

Then perhaps the reload logic should be refactored. It is usually not recommended to mix threads and exec() calls.

rkojedzinszky (commented):
Can you provide test scripts for these? I suspect that resetting signals affect the child processes, not their parents, and not preventing their parents receiving SIGCHLD upon exit.

rkojedzinszky:
> After finding this docbug ( https://bugs.python.org/issue32985 ) I've cracked the issue that required the detailed signal handling in the first place.
> 
> When forking off for the subprocess, `restore_signals` (a default parameter) restores only the signals that the intepreter itself modified, i.e. `sigint`, `sigterm` and not `sigchld`.
> 
> This results in signals being received for children - I used psutil to filter these out. With the commit I'm about to push, subprocesses should be fully and completely isolated, and all deadlocks eliminated without the need for the custom reap method, and the process should restart cleanly with no zombies left around from previous invocations.

Still dont understand what kind of deadlocks may arise.

> 
> Unfortunately, the fix is only good until 3.13 at latest, as per https://bugs.python.org/issue38435 but it seems like the Python dev's are aware of the issue.

murrant:
Hmm, I have a thought, what if we just let systemd restart it?  However this does mean if you run the script by hand it will just exit when maintenance runs.

TheMysteriousX:
We could strip out even more code.

The initial implementation of the service just got restarted by the process supervisor, but that was decided to be undesirable as it'd abort polls/discovers in flight.

It'd just be a case of stripping out the remaining signal code, removing the args from check_call that prevent signal propogation and removing the restart action and any subsequent dead code, if you want to do that instead.

rkojedzinszky:
At least for short term a solution would be nice to be merged. I can say, we are affected, and it seems there are other installations too where pollers stop randomly. It seems that my really short patch solves the issue, but of course, in the long term, exec() should be avoided, and let some process manager do this job.

murrant:
Merged #12593 for the release.  I think we still need to look at this.

murrant:
My thoughts:
1. Move poller finishing in to php code (lock release, etc)
2. Exit parent process without waiting
3. Children are orphaned onto pid 1
4. supervisor restarts python process, and queue jobs.
5. locks are not expired for polling processes that have not exited, so new jobs won't be queued
6. orphans exit and release locks after updating last_polled (pid 1 reaps orphans)
7. jobs are queued normally at next interval

TheMysteriousX:
That'd work - you'll need to add 'KillMode=process' to the unit file otherwise systemd cleans up the entire cgroup regardless of parentage. I think it should also work with other inits too - docker may need some thought/changes, I don't know how the poller has been implemented there (does it even use the service?).

You can either implement the restart using Restart=always in the unit and just have the poller terminate itself as it does now (not portable to other inits) or using a timer (still not portable but timers can be replaced with cron scripts).

You'll still need 'start_new_session=True' from this changeset, and to remove the same blocks of code (plus a bit more).

rkojedzinszky:
> My thoughts:
> 
> 1. Move poller finishing in to php code (lock release, etc)
> 2. Exit parent process without waiting
> 3. Children are orphaned onto pid 1
> 4. supervisor restarts python process, and starts queuing jobs.  (or other master is still queuing jobs)
> 5. locks are not expired for polling processes that have not exited, so new jobs won't be queued
> 6. orphans finish polling (updating last_polled), release locks, and exit (pid 1 reaps orphans)
> 7. jobs are queued normally at next interval

Is this all needed because of some automatic update mechanism? What if the actual update process is left to the user?

murrant:
@rkojedzinszky that is already possible.  Auto update is optional, but I would say a very large percentage of users have auto updates enabled.

Jellyfrog:

PR #12619, opened 2021-03-15
Add CPU and Mem for Teldat Devices

Cupidazul:
Just adding CPU em Memory information for Teldat devices...

Mib's used:

:cpu:
teldatCPUBusy5sec		= 1.3.6.1.4.1.2007.4.1.2.1.2.1.1.1 = 5 segs util%
teldatCPUBusy1min	= 1.3.6.1.4.1.2007.4.1.2.1.2.1.1.2 = 1 min util%
teldatCPUBusy5min	= 1.3.6.1.4.1.2007.4.1.2.1.2.1.1.3 = 5 mins util%

:memory:
:class: system:
telProdNpMonSistemMemTotal			= 1.3.6.1.4.1.2007.4.1.2.1.1.26 = "Total system memory in bytes."

:class: cache:
telProdNpMonSistemMemTotalcache		= 1.3.6.1.4.1.2007.4.1.2.1.1.27 = "Total cache memory in bytes."
telProdNpMonSistemMemFreecache		= 1.3.6.1.4.1.2007.4.1.2.1.1.28 = "Available free cache memory in bytes."

:class: shared:
telProdNpMonSistemMemTotalnoncache	= 1.3.6.1.4.1.2007.4.1.2.1.1.29 = "Total non cache memory in bytes."
telProdNpMonSistemMemFreenoncache	= 1.3.6.1.4.1.2007.4.1.2.1.1.30 = "Free non cache memory in bytes."

During the development, as per https://docs.librenms.org/Developing/os/Mem-CPU-Information/, I've seen that LibreNMS is not able to plot all memory information related to Flash used/free , so I've opted for now to show these results (note: some devices may return "zero": telProdNpMonSistemMemTotalnoncache = telProdNpMonSistemMemFreenoncache = 0, not sure if this is a bug for now.).

For more information:
https://github.com/librenms/librenms/pull/12397

note: there are new mibs included in this pull.

files:
(CHG) includes/definitions/discovery/teldat.yaml
+mibs/teldat/TELDAT-MON-CPU-MIB (:cpu:)
+mibs/teldat/TELDAT-MON-CommonInfo-MIB (:memory:)
+mibs/teldat/TELDAT-SW-STRUCTURE-MIB (:memory: pre-req.)

Thx.
Cupidazul

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

[...]

PR #12621, opened 2021-03-15
Switch parsing order of <field>_template and <field>_regex

Jellyfrog:
This allows for a rudimental "else" in yaml with template support.

For example; This would match the first oid that exists,
either "oid1" or "oid2/oid3":

```yaml
    serial:
        - MIB::oid1
        - MIB::oid2
        - MIB::oid3
    serial_template: '{{ MIB::oid1 }} {{ MIB::oid2 }}/{{ MIB::{{ MIB::oid3 }} }}'
    serial_regex: '/(?<serial>[^ ]+)/'
```

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/OS/Traits/YamlOSDiscovery.php
+++ b/LibreNMS/OS/Traits/YamlOSDiscovery.php
@@ -67,15 +67,16 @@ trait YamlOSDiscovery
 
         foreach ($oids as $field => $oid_list) {
             if ($value = $this->findFirst($data, $oid_list, $numeric)) {
-                // extract via regex if requested
-                if (isset($os_yaml["{$field}_regex"])) {
-                    $this->parseRegex($os_yaml["{$field}_regex"], $value);
-                    $value = $device->$field;
+                $device->$field = $value;
+
+                if (isset($os_yaml["{$field}_template"])) {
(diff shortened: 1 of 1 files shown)


Jellyfrog:
ping @murrant

Jellyfrog:
Refs: https://github.com/librenms/librenms/pull/12619#discussion_r594714725

murrant:
This breaks logic used in a lot of definitions.  Why would your example require an order change?

Jellyfrog:
> This breaks logic used in a lot of definitions. Why would your example require an order change?

Where? And why not? I think the logical is backwards today, why would you do templating *after* regex matching?
As I see it you want to combine your fields, then match on them.
It would either select `oid1` or `oid2/oid3` (combined)

I had a similar use case for vmware but solved it in another way

murrant:
The template is only needed for oids mostly.  the sysDescr_regex is more of a fallback if those do not exist.
Maybe I'm not understanding the use case correctly.

Jellyfrog:
The point is to be able to combine oids and then match on them. sometimes a serial number can be split into multiple oids, or features can be from multiple fields

murrant:
Isn't that how it works already?

murrant:
Not needed at this time.  Seems to swap natural parsing order to enable tricks.

PR #12629, opened 2021-03-18
VRP - Fix SSID Client count

PipoCanaja:
This PR fixes information that was misplaced:
Back when I added support for VRP wireless, only "per radio and per SSID" count was available. Resulting in a huge amount of sensors. And at that time no aggregation of wireless sensors was possible. So at the end of the day, it was impossible to visualize the amount of clients per SSID at all. 

Now, both aspects improved: 
- We now have "per band and per SSID" which mean, today, twice as much data as expected for SSID client Count. So one sensor for 5Ghz and one for 2.4 Ghz.
- We can aggregate those using wireless-sensors (creating a 3rd "Total" sensor for each SSID)

The number of clients per radio is of course still available in the AccessPoints screen.

#### Please note

#### Testers


LibreNMS/OS/Vrp.php
```diff
@@ -404,28 +404,52 @@ public function discoverWirelessApCount()
...
-                        $a_index_oid . '.' . $r_index . '.' . $s_index,
-                        'Radio:' . $r_index . ' SSID:' . $ssid['hwWlanVapProfileName'],
-                        $ssid['hwWlanVapStaOnlineCnt']
-                    );
-                }
+
+        //Clients (STA) for 2.4Ghz and 5Ghz
+        $sta5gTable = $this->getCacheTable('hwWlanSsid5gStaCnt', 'HUAWEI-WLAN-VAP-MIB', 3);
+        $sta2gTable = $this->getCacheTable('hwWlanSsid2gStaCnt', 'HUAWEI-WLAN-VAP-MIB', 3);
+        $staTable = array_merge_recursive($sta5gTable, $sta2gTable);
```
murrant, line 411: snmpwalk_group() can accept an input array so you don't have to merge them.  Also, it doesn't look like this data is used for any other sensor, so no need to cache them.

PipoCanaja: Makes sense. Will fix that.

LibreNMS/OS/Vrp.php
```diff
@@ -404,28 +404,52 @@ public function discoverWirelessApCount()
...
+
+        //Map OIDs and description
+        $oidMap['hwWlanSsid5gStaCnt'] = '.1.3.6.1.4.1.2011.6.139.17.1.2.1.3.';
+        $oidMap['hwWlanSsid2gStaCnt'] = '.1.3.6.1.4.1.2011.6.139.17.1.2.1.2.';
+        $descrMap['hwWlanSsid5gStaCnt'] = '5 Ghz';
+        $descrMap['hwWlanSsid2gStaCnt'] = '2.4 Ghz';
+
+        foreach ($staTable as $ssid => $sta) {
+            //Convert string to num_oid
+            $numSsid = strlen($ssid) . '.' . implode('.', unpack('c*', $ssid));
```
murrant, line 421: If you use the -Ob option when you snmpwalk the data, you don't need to change the data at all, it is already in numeric oid format.

PipoCanaja: But I need the text as well, in order to build the description. So either I convert text to OID or the opposite.

LibreNMS/OS/Vrp.php
PipoCanaja: Hum, snmpwalk_group only works with one OID, not an array. So the merge is still mandatory. But at least caching will be avoided.

murrant: nope, like this (changed to snmpwalk_cache_oid because group does not allow specifying snmp flags):
```php
        $staTable = snmpwalk_cache_oid($this->getDeviceArray(), 'hwWlanSsid', [], 'HUAWEI-WLAN-VAP-MIB', null, '-OQUsb');
        $staTable = snmpwalk_cache_oid($this->getDeviceArray(), 'hwWlanSsid2gStaCnt', $staTable, 'HUAWEI-WLAN-VAP-MIB', null, '-OQUsb');
        $staTable = snmpwalk_cache_oid($this->getDeviceArray(), 'hwWlanSsid2gStaCnt', $staTable, 'HUAWEI-WLAN-VAP-MIB', null, '-OQUsb');

```

murrant: See how passing the same array inserts them into it.  Much better than a recursive merge.

PipoCanaja: `        $staTable = snmpwalk_cache_oid($this->getDeviceArray(), 'hwWlanSsid', [], 'HUAWEI-WLAN-VAP-MIB', null, '-OQUsb');`
But `hwWlanSsid` is the index. It can't have it in the result of any snmpwalk. So this line returns an empty array. Only the hwWlanSsid2gStaCnt and hwWlanSsid5gStaCnt give a result. So no matter what, I'll need to do the translation to get num_to_string or string_to_num. 

So at the end, here are the 2 working variants:
```
        $sta5gTable = snmpwalk_group($this->getDeviceArray(), 'hwWlanSsid5gStaCnt', 'HUAWEI-WLAN-VAP-MIB', 3);
        $sta2gTable = snmpwalk_group($this->getDeviceArray(), 'hwWlanSsid2gStaCnt', 'HUAWEI-WLAN-VAP-MIB', 3);
        $staTable = array_merge_recursive($sta5gTable, $sta2gTable);
```
and 
```
        $staTable = snmpwalk_cache_oid($this->getDeviceArray(), 'hwWlanSsid2gStaCnt', [], 'HUAWEI-WLAN-VAP-MIB', null, '-OQUs');
        $staTable = snmpwalk_cache_oid($this->getDeviceArray(), 'hwWlanSsid5gStaCnt', $staTable, 'HUAWEI-WLAN-VAP-MIB', null, '-OQUs');
```

The second one avoids the merge_recursive, but does caching we don't need.

One way of the other, this table is not expected to be 1000 entries long. It is only the doubled SSID table, because of the 2 bands. And even in big networks, you don't have 1000 SSIDs on a single controller. I am not sure caching and/or merge will ever be an issue here.

murrant: All index values must have an oid you can access from.  Look at the snmp table, it is a member of the table.  Does it return nothing from the captured data or the actual device?  Because it would not be in the captured data.

PipoCanaja: Documentation confirms the behaviour. The `hwWlanSsid` is the index, and cannot be walked by itself. So no other choice than extracting (and converting) the index from num to str or the opposite.

OID | Object Name | Syntax | Max Access | Description | Implemented Specifications
-- | -- | -- | -- | -- | --
1.3.6.1.4.1.2011.6.139.17.1.2.1.1 | hwWlanSsid | OCTET STRING | **not-accessible** | This object indicates the table index. | This object is implemented as defined in the corresponding MIB file.
1.3.6.1.4.1.2011.6.139.17.1.2.1.2 | hwWlanSsid2gStaCnt | Integer32 | read-only | This object indicates the number of users at 2.4G frequency band. | This object is implemented as defined in the corresponding MIB file.

murrant: Fun.

PipoCanaja: Annoyingly fun :)

murrant (approved):

PR #12630, opened 2021-03-18
Modernize pdf support (#12566)

Oirbsiu:
Trying again to submit pull request for DOMPDF support

originally created as PR #12566

#### Please note

#### Testers


murrant:
@Oirbsiu looks like you have a decent amount of work in this.
Your method for handling legacy functions is not right.  We don't want to just copy code into a class and call that.
look at format_si / Number::formatSi for an example of what to do.  Code moved into a class should not call legacy functions.

Also, you may need to make something like a billing repository class to hold billing related logic.

Oirbsiu:
@murrant thanks for the feedback - my reasoning behind simply copy and paste is partly due to limited understanding of how LibreNMS is put together, my knowledge of PHP, and I didnt want to suggest a change that ends up removing/breaking a feature that another user is using.
I will have a go at these the best I know how. :)
Thanks

murrant:
Well, I did you a favor and removed a chunk of functions, so it is easy to see they have been migrated.
If you are concerned about not breaking existing code, you make a wrapper function with the same name and signature as the old function that just calls the new function in object oriented code.
We also don't want a generic Functions.php class, it is better to group things together.

Oirbsiu:
Thank you. this is where my knowledge is starting to be on the thin side - very new to git and I cant see the bits that you changed (you said you removed a chunk of functions). 
I tired to merge the "master" with my local master - this I think I have done. but I see from the checks above, it doesn't like what I did.

murrant:
It might be tricky to get this to where you need to be. It might be a good idea to take the changes you made and start a new branch.  Sorry this is being difficult.  You are doing great :)

murrant:
Closing for now, would be happy to see this work finished.  Feel free to resubmit!

PR #12640, opened 2021-03-22
Fixed polling and health issues for 9001 and NCS devices, added suppo…

### LINKED ISSUE #12591: Cisco NCS exclusion for Cisco XR dBm workaround
Cisco ASR9K devices have previously needed a work around to obtain correct dBm values from their transceivers.
This is because the ASR9K devices only return these values as a wattage value, rather than a dBm value.

The pull request below was added to fix these in XR devices.
https://github.com/librenms/librenms/pull/1661
https://github.com/librenms/librenms/issues/1641

The code in question is here, it builds a "fake sensor" to map the watt values as dBm results after a conversion

https://github.com/librenms/librenms/blob/master/includes/discovery/sensors/cisco-entity-sensor.inc.php#L201

Since this pull request, some of the newer XR devices (Namely, the NCS series) does report correct dBm values, what now happens is that this conversion logic is now applied to these values that are already correct and returns a very incorrect value.

image

Now, i've tested excluding devices with NCS in the "hardware" field, however I wanted to start a discussion about how this should best be approached.

Would it be better to:

1. Statically exclude XR devices that do provide correct dBm values from this logic to convert watts to dbm
2. Change the current logic to statically include XR devices that do not return dBm vlaues (potentially breaking some users setups while we determine what devices these are, currently i only believe ASR9K routers do this)
3. try and add a check somehow to see what values the devices return and apply the correct logic (unsure how we would approach this)

Currently i've tested excluding NCS devices in the logic, as i know those return the correct values and that has turned out correctly below.

image

**PLEASE READ FIRST**

**Please do not post bugs or ask questions or feature requests here**

**For help and support you can use our [Discord server](https://t.libren.ms/discord) or our [community site](https://community.librenms.org/c/help).**

To create a new device request please use this link: https://github.com/librenms/librenms/issues/new?template=New-Device.md

WillIrvine:
This PR fixes the following items

Polling Fixes - An extra lookup has been added to the polling cycle to determine if the original value was a mW value. This ten only applies the logic to these sensors and does not change the rest. fixes #12591

- Adjusted to account for both traditional dBm sensors, and fake sensors created for some iosxr devices from mW values
- Adjusted to account for a floating precision value on seperate XR devices rather than a statically set value

Discovery Fixes - 

- Changed the divisor logic to more accurately determine the divisor value. 
- Fixes a non set "Critical" threshold from overwriting a "Major" Threshold
- Added a preg_match to fit ASR9901 devices

As a note to users with ASR9001 devices, if your returned values are 10 times higher than they should be, contact cisco and request the patch for the following bug.

https://bst.cloudapps.cisco.com/bugsearch/bug/CSCvs33097

#### Please note

#### Testers


murrant:
I think you should fix this in discovery with user_func and delete the custom polling file.

WillIrvine:
Thanks for the input @murrant. Switched to using "user_func" and removed polling file.
looks like the tests that are failing are unrelated and in the current master branch

1) LibreNMS\Tests\OSDiscoveryTest::testOS with data set "freshtomato" ('freshtomato')
Test file: freshtomato.snmprec

Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'freshtomato'
+'linux'

/opt/librenms/tests/OSDiscoveryTest.php:113
/opt/librenms/tests/OSDiscoveryTest.php:77

2) LibreNMS\Tests\OSDiscoveryTest::testAllFilesTested
Not all snmprec files were checked: Array
(
    [0] => freshtomato
)

murrant:
@WillIrvine Looks better :)   But, you seem to have a badif filter applied when you captured data and the CI tests have none.  You'll need to disable them and re-create the json files.

WillIrvine:
Hi, @murrant . This looks like its now failing on an older test "tests/data/iosxr_asr9010.json" as its now picking up dBm sensors that previously didn't discover. 

Whats the best approach here as i don't have this type of device to rebuild the test data on.

murrant:
Luckily you don't need a device to update the json. Just run save-test-data.php with the correct os and variant info (if no variant, use '')

I did notice that the sensor_current values match, just the divisor and user_func etc are changed, which means your changes are correct :)

murrant (approved):

PR #12642, opened 2021-03-23
Misc cleanups

Jellyfrog:
Should only be cosmetic changes.
A bit preparing work for static analysis tools. 

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Widgets/WidgetController.php
+++ b/app/Http/Controllers/Widgets/WidgetController.php
@@ -60,14 +60,11 @@ abstract class WidgetController extends Controller
     public function __invoke(Request $request)
     {
         $this->show_settings = (bool) $request->get('settings');
+        $settings = $this->getSettings();
 
         if ($this->show_settings) {
             $view = $this->getSettingsView($request);
-        }
-
-        $settings = $this->getSettings();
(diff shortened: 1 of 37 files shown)


app/Http/Controllers/Widgets/WidgetController.php
```diff
@@ -60,14 +60,11 @@ public function getSettingsView(Request $request)
     public function __invoke(Request $request)
     {
         $this->show_settings = (bool) $request->get('settings');
+        $settings = $this->getSettings();
 
         if ($this->show_settings) {
             $view = $this->getSettingsView($request);
-        }
```
murrant, line 66: I think the intention here was to avoid fetching settings when they are not needed.
However, it doesn't look like the code succeeds at that, perhaps it was changed after the fact.

murrant:
Looks mostly good, a few are riskier than others.

app/Http/Controllers/Widgets/WidgetController.php
Jellyfrog: Yeah I thought the same first, but seems it just loads it anyway

murrant (approved):

PR #12655, opened 2021-03-26
Add lint Github Action

murrant:

#### Please note

#### Testers

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,6 +20,10 @@ jobs:
       - name: Lint Code Base
         uses: github/super-linter@v3
         env:
+          VALIDATE_BASH: true
+          VALIDATE_BASH_EXEC: true
+          VALIDATE_PHP_BUILTIN: true
+          VALIDATE_PYTHON_PYLINT: true
           VALIDATE_ALL_CODEBASE: true
           DEFAULT_BRANCH: master
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


Jellyfrog:
I think this will be too much.
Larastan we should prepare for tho.

murrant:
Larastan only check PHP code.  We need to lint python and bash code, too.   I tried to run against all files, but apparently, this means run all check against all files... way too much.

Jellyfrog:
I'm aware what Larastan does and I still think the rest is almost unnecessary for this.

>I tried to run against all files, but apparently, this means run all check against all files... way too much.

It doesnt mean run all against all, it means run against all files not not just changed;

>Will parse the entire repository and find all files to validate across all types. NOTE: When set to false, only new or edited files will be parsed for validation.

And without disabling a lot of checks it will just be a mess, https://github.com/librenms/librenms-agent/blob/master/.github/workflows/linter.yml#L27-L32

murrant:
It wasn't that bad, mostly just bash scripts we never touch with issues: https://github.com/librenms/librenms/runs/2206789857?check_suite_focus=true
Also, the whitelist is better than blacklist, as we only want a few validations.

Jellyfrog:
I think you should enable black also

murrant:
Black looked a little aggressive with its style enforcement, not sure I care about style in python.

Jellyfrog:
It got built in fixer, so we don't need to care really, but i can submit a PR with that after this.

Jellyfrog (approved):

PR #12678, opened 2021-03-30
Enable PHPStan linter

Jellyfrog:
Couldn't add it via super-linter due to some errors with missing php extensions

#### Please note

#### Testers


Jellyfrog:
Any thoughts on what level to set?

Right now with level 3, its a modest 20 errors on the ignore list, 5 of them actual real bugs(!) 
Level 5 generates only 140 more errors.
Level 6 is all about type hinting, which also might be a good idea to start enforce (2565 more errors tho)

Jellyfrog:
I added a commit to bump it to level 5

murrant:
Level 6 seems too high.  Would be great if we can merge this with almost no blacklist.

Jellyfrog:
Well. Each merge we do contributes to the ignore list :) so I think we should merge this as is. And work on fixes after

murrant:
Should start settling down considering release is in a week.

Jellyfrog:
@murrant I would like to merge this now before too many new errors trickle in :)

murrant (approved):

murrant:
Going to have errors all over ;)

PR #12683, opened 2021-03-30
Remove snmp2ipv6

murrant:
Fixes a bug too.

#### Please note

#### Testers


Jellyfrog:
It seems you forgot to update the tests :)

murrant:
Yeah, then, I was like, why isn't local address there...

tests/data/vrp_ce12804-withvrf.json
```diff
@@ -24928,9 +24928,9 @@
                     "bgpPeerIdentifier": "192.168.189.96",
                     "bgpPeerRemoteAs": 64512,
                     "bgpPeerState": "established",
-                    "bgpPeerAdminStatus": "stop",
-                    "bgpPeerLastErrorCode": null,
-                    "bgpPeerLastErrorSubCode": null,
+                    "bgpPeerAdminStatus": "",
```
Jellyfrog, line 24931: do we want this to be empty strings or null?

PipoCanaja: This file does not include the OID we need for this AdminStatus. I'd need to collect again snmp with current code.  Cannot do it currently, I don't have a lab device available, and handcleaning prod data is very time consuming (and usually result in a "false" file breaking tests ...)

PipoCanaja: Hum, the value is correctly populated during poll, I need to do some tests with a real device to clarify this. But I don't have an easy way to run this in prod currently (too many patches to apply over my prod instance running stable).

Jellyfrog (approved):

tests/data/vrp_ce12804-withvrf.json
murrant: @PipoCanaja best way to test is to temporarily add the device to your development install (assuming it can reach the device)... then remove the device after testing.

PipoCanaja: This test file is behaving funny. Testing on my local machine (with DBTEST=1, on a freshly cleaned test DB) gives a result which is not the same as the tests here on github ... drives me crazy. I even tried to recreate it all together, same behaviour.

murrant: @PipoCanaja did you see your highlighted change is not in the test files anymore :)

vincentbernat:
This change is not really well explained.

On Juniper, their implementation of the BGP4-V2-MIB is buggy as they don't encode the index with an implicit length. This has been reported and fixed some time ago but they cannot fix it by default since it would break for many people relying on the bug for their own needs. So, they put it behind an option (`set protocols bgp snmp-options emit-inet-address-length-in-oid`). 

As most people won't put this option, unless LibreNMS tries to detect the situation, I think this is safer to keep the current hack.

murrant:
@vincentbernat would be great if we have some test data as our tests did not catch this.

vincentbernat:
I am surprised, because the data you have for Juniper should have caught this case. For example, you have (in https://github.com/murrant/librenms/blob/9d4e052c6a2621bf7339f4bcdbe2526e5a178f4b/tests/snmpsim/junos_mx.snmprec):

```
1.3.6.1.4.1.2636.5.1.1.2.1.1.1.11.0.2.253.0.0.40.0.1.0.1.0.0.0.0.0.1.0.4.2.253.0.0.40.0.1.0.1.0.0.0.0.0.1.0.5
```

The IP addresses are not correctly encoded and we cannot use the MIB to decode them:

```
 15:19 ❱ snmptranslate -m +BGP4-V2-MIB-JUNIPER  -M+$PWD 1.3.6.1.4.1.2636.5.1.1.2.1.1.1.11.0.2.16.253.0.0.40.0.1.0.1.0.0.0.0.0.1.0.4.2.16.253.0.0.40.0.1.0.1.0.0.0.0.0.1.0.5
BGP4-V2-MIB-JUNIPER::jnxBgpM2PeerRemoteAddr.0.ipv6."fd:00:00:28:00:01:00:01:00:00:00:00:00:01:00:04".ipv6."fd:00:00:28:00:01:00:01:00:00:00:00:00:01:00:05"
```

The OID needs to be fixed by prefixing the IP by their length. Add `16` before each IP:

```
 15:19 ❱ snmptranslate -m +BGP4-V2-MIB-JUNIPER  -M+$PWD 1.3.6.1.4.1.2636.5.1.1.2.1.1.1.11.0.2.16.253.0.0.40.0.1.0.1.0.0.0.0.0.1.0.4.2.16.253.0.0.40.0.1.0.1.0.0.0.0.0.1.0.5
BGP4-V2-MIB-JUNIPER::jnxBgpM2PeerRemoteAddr.0.ipv6."fd:00:00:28:00:01:00:01:00:00:00:00:00:01:00:04".ipv6."fd:00:00:28:00:01:00:01:00:00:00:00:00:01:00:05"
```

The use of the above option fixes that, but it seems your code doesn't work either for IPv6 (but it's OK for IPv4). I suppose you don't parse IPv6 strings encoded as octet strings. Unfortunately, as I say, I don't think you can expect most users to fix use that option, so you need to continue to have to use the OID instead of the symbolic names.

murrant:
@vincentbernat odd, I thought the new code did that.

PR #12694, opened 2021-03-31
Fix OS sensor array return

Jellyfrog:

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/OS/AirosAfLtu.php
+++ b/LibreNMS/OS/AirosAfLtu.php
@@ -49,6 +49,8 @@ class AirosAfLtu extends OS implements
                 new WirelessSensor('distance', $this->getDeviceId(), '.1.3.6.1.4.1.41112.1.10.1.4.1.23.' . $index, 'airos-af-ltu', 1, 'Distance', $entry['afLTUStaRemoteDistance'], 1, 1000), //UBNT-AFLTU-MIB::afLTUStaRemoteDistance
             ];
         }
+
+        return [];
     }
 
     /**
--- a/LibreNMS/OS/Routeros.php
+++ b/LibreNMS/OS/Routeros.php
@@ -109,7 +109,6 @@ class Routeros extends OS implements
     {
         $data = $this->fetchData();
 
-        $sensors = [];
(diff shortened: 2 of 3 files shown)


LibreNMS/OS/Routeros.php
```diff
@@ -112,19 +112,21 @@ public function discoverWirelessFrequency()
         $sensors = [];
         foreach ($data as $index => $entry) {
             if ($entry['mtxrWlApFreq'] == null) {
-                return $this->discoverSensor(
+                $sensors[] = $this->discoverSensor(
```
Jellyfrog, line 115: Ok, these changes are wrong, because `discoverSensor()` loops and return an sensor array itself.  
I fail too see the logic of the code tho.
ping @murrant, you're more clever than me

LibreNMS/OS/AirosAfLtu.php
```diff
@@ -44,11 +44,12 @@ public function discoverWirelessDistance()
     {
         $oids = snmpwalk_cache_oid($this->getDeviceArray(), 'afLTUStaRemoteDistance', [], 'UBNT-AFLTU-MIB', null, '-OteQUsb');
 
+        $sensors = [];
         foreach ($oids as $index => $entry) {
-            return [
-                new WirelessSensor('distance', $this->getDeviceId(), '.1.3.6.1.4.1.41112.1.10.1.4.1.23.' . $index, 'airos-af-ltu', 1, 'Distance', $entry['afLTUStaRemoteDistance'], 1, 1000), //UBNT-AFLTU-MIB::afLTUStaRemoteDistance
```
murrant, line 49: I think this was by design to take the first entry only.

murrant:
I think what you fail to see is both of these are point-to-point links.  So they only have one result in the cases where they return.

Jellyfrog:
> I think what you fail to see is both of these are point-to-point links. So they only have one result in the cases where they return.
> 
> Probably the correct fix is to add `return [];` at the bottom.

Now I get it :upside_down_face:

murrant (approved):

PR #12696, opened 2021-04-01
BDCOM update support

hrtrd:

#### Please note

#### Testers


hrtrd:
what happens?

Jellyfrog:
Id suggest revert the changes on bdcom files. And add new ones with --variant specified.

includes/definitions/discovery/bdcom.yaml
```diff
@@ -1,13 +1,127 @@
...
+                 - ponIfIndex
+                 - gponOltPonPortPortActiveOnuNum
+                 - gponOltPonPortPortInactiveOnuNum
     processors:
         data:
             -
-                oid: bdpmCPUTotal5min
-                num_oid: '.1.3.6.1.4.1.3320.9.109.1.1.1.1.5.{{ $index }}'
-                index: 0
+                oid: nmscardCPUUtilization
```
murrant, line 29: According the the tests, it is using HOST-RESOURCES-MIB instead of this...  Maybe a mib failure?

PipoCanaja:
Yep, @hrtrd 
Could you restore the original bdcom test files and create a new variant for your tests ?
(so we end up with both the old and the new test files pair) ?
You also need to add your new MIB files to this PR, that why tests cannot find the same results as on your dev machine.

hrtrd:
yes, that's right, I completely forgot...

includes/definitions/discovery/bdcom.yaml
PipoCanaja: Seems better now with mibs added. But still wondering if the bdpmCPUTotal5min should be kept aside of nmscardCPUUtilization or replaced. 
@hrtrd Are you sure nmscardCPUUtilization works on any device/version ? wouldn't it make sense to keep both in the yaml file?

murrant (approved):
LGTM, nice work.

PR #12702, opened 2021-04-03
fix empty sensor_classes

si458:
fix empty sensor_classes from being shown in menu and headers
this is caused by IPMI inserting blank sensors into DB with no sensor_class for some reason?
so the menu and headers seem to be showing an empty option?

#### Please note

**BEFORE:**
goofed
**AFTER:**
fixed

#### Testers

### DIFF
--- a/LibreNMS/Util/ObjectCache.php
+++ b/LibreNMS/Util/ObjectCache.php
@@ -74,7 +74,7 @@ class ObjectCache
     public static function sensors()
     {
         return Cache::remember('ObjectCache:sensor_list:' . auth()->id(), self::$cache_time, function () {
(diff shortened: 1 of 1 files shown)


Jellyfrog:
Do you know what device creates this? And in that case could you try capute some data for us with

./scripts/collect-snmp-data.php -h *ip* -f /tmp/ipmi.snmprec -v ipmi

si458:
I will have a look tomorrow for you, catch up on discord

the blank sensors only appear after I enable ipmi polling for our hp servers 

I don't poll the snmp of the ipmi ip itself as it would be another device to poll and list

murrant:
Indeed, it would definitely be better to fix the broken sensor :)

si458:
this can be closed as it should be fixed by #12709

PR #12689, opened 2021-03-31
Attempt to get some apps working with rrdcached.

djamp42:
I posted this request under feature requests, but figured i would have a go at it myself, i confirmed this works with the Docker app and rrdcached. 
https://community.librenms.org/t/support-remote-rrdcached-for-some-applications/15275

Please give a short description what your pull request is for

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [X] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [X] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #12704, opened 2021-04-05
Rrdtool array class

djamp42:
Some applications grab an array of rrd filenames to display links on the webgui and display the rrd graph data. This does not work over rrdcached, this pull request is to fix that.. I only have docker fixed in this pull request, but can fix the rest if this looks good. 

Here is the orginal PR that i tried.
https://github.com/librenms/librenms/pull/12689

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Util/RrdTool.php
+++ b/LibreNMS/Util/RrdTool.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace LibreNMS\Util;
+
+use App\Models\Device;
+use LibreNMS\Config;
+use LibreNMS\Util\Rewrite;
+use LibreNMS\Data\Store\Rrd;
+
+
+class RrdTool
+{
+    /**
+     * Get array of all rrd files for device
+     *
+     * @param array $device device for which we get the rrd's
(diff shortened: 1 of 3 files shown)


LibreNMS/Util/RrdTool.php
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Get array of all rrd files for device
+     *
+     * @param array $device device for which we get the rrd's
+     * @return array $rrdfilearray array of rrd files for this host
+     */
+    public static function getRrdFiles($device)
+    {
+        if (Config::get('rrdcached')) {
+            $rrdcachedcmd = sprintf('%s list /%s --daemon %s', Config::get('rrdtool', 'rrdtool'), $device['hostname'], Config::get('rrdcached'));
```
murrant, line 21: You should send rrdtool commands to the listening instance rather than spawn a new process.
(Unless I am miss-remembering and the process is not always available)

djamp42: yeah it looks like it would make sense to put it in LibreNMS/Data/Store/Rrd.php so i'll rework this.

PR #12706, opened 2021-04-05
Add feature to capture a full snmprec

Jellyfrog:
Good for adding device support when you don't have access to it directly. 
Or when discovery/poller causes errors when capturing normally

#### Please note

#### Testers


Jellyfrog:
Suggestion for other name of the argument?

`--full` Maybe

murrant:
--full would be good.  Make a note NOT to submit this to the repo, we don't need to be adding multi-megabyte snmprec files :D  Also, explain why you would want to use that option in the help text.

murrant (approved):

PR #12719, opened 2021-04-07
Update syslog.php

Fireball2k15:
Some syslog-sendes use their FQDN as sender. If you put the whole FQDN in your config.php you can not access this entry. The Function Config::get uses dots as an array seperator.

This pull request fixes the problem described above.
Allready checked an have this running on my own site.

#### Please note

#### Testers

### DIFF
--- a/includes/syslog.php
+++ b/includes/syslog.php
@@ -53,9 +53,16 @@ function process_syslog($entry, $update)
     }
 
     $entry['host'] = preg_replace('/^::ffff:/', '', $entry['host']);
-    if ($new_host = Config::get("syslog_xlate.{$entry['host']}")) {
-        $entry['host'] = $new_host;
+
+    $axlate = Config::get('syslog_xlate');
+    if(true == is_array($axlate)){
+      foreach ($axlate as $xname => $xlate) {
+        if ( $xname == $entry['host']){
+           $entry['host'] = $xlate;
+        }
+      }
     }
+    
     $entry['device_id'] = get_cache($entry['host'], 'device_id');
     if ($entry['device_id']) {
         $os = get_cache($entry['host'], 'os');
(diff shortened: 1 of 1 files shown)


includes/syslog.php
```diff
@@ -53,9 +53,16 @@ function process_syslog($entry, $update)
     }
 
     $entry['host'] = preg_replace('/^::ffff:/', '', $entry['host']);
-    if ($new_host = Config::get("syslog_xlate.{$entry['host']}")) {
-        $entry['host'] = $new_host;
+
+    $axlate = Config::get('syslog_xlate');
+    if (true === is_array($axlate)) {
+        foreach ($axlate as $xname => $xlate) {
```
murrant, line 59: why loop?   just do `$entry['host'] = $axlate[$entry['host']] ?? $entry['host'];`

murrant: you could even inline the $xlate variable

includes/syslog.php
```diff
@@ -166,6 +173,12 @@ function process_syslog($entry, $update)
 
         unset($os);
     }//end if
+    elseif (true == Config::get('syslog_debug')) {
+        // For finding unknowen Hosts
+        $fd = fopen('/opt/librenms/logs/Debug.log', 'a');
```
murrant, line 178: don't open a new file , just `Log::<level>()`

includes/syslog.php
Fireball2k15: ok, letzt take your code - it's more efficient.

includes/syslog.php
Fireball2k15: I tryed to use 
Log::debug("Can't process host: " . $entry['host']);
but I can not see the outpout in the Debug File.

murrant: It goes to the standard `logs/librenms.log` then.

PR #12735, opened 2021-04-09
RRD Graph optimization

murrant:
Do not use temporary files to generate graphs
Don't start up a process to pipe commands to, just run the command
Error image improvements

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Data/Store/Rrd.php
+++ b/LibreNMS/Data/Store/Rrd.php
@@ -28,9 +28,11 @@ use Illuminate\Support\Str;
 use LibreNMS\Config;
 use LibreNMS\Data\Measure\Measurement;
 use LibreNMS\Exceptions\FileExistsException;
+use LibreNMS\Exceptions\RrdGraphException;
 use LibreNMS\Proc;
 use LibreNMS\Util\Rewrite;
 use Log;
+use Symfony\Component\Process\Process;
 
 class Rrd extends BaseDatastore
 {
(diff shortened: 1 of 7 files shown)


includes/html/graphs/graph.inc.php
```diff
@@ -24,13 +26,12 @@
...
-$graph_type = (isset($vars['graph_type']) ? $vars['graph_type'] : Config::get('webui.graph_type'));
-
 $period = ($to - $from);
-$base64_output = '';
 $prev_from = ($from - $period);
 
-$graphfile = Config::get('temp_dir') . '/' . strgen();
+$graph_image_type = $vars['graph_type'] ?? Config::get('webui.graph_type');
+$rrd_options = '';
+$auth = false;
```
Jellyfrog, line 34: Row 41-49 is now dead code

Jellyfrog: And many more :)
Debug code?

murrant: Negative.  $auth is set in this file `require Config::get('install_dir') . "/includes/html/graphs/$type/auth.inc.php";`  yes this is old silly code.

Jellyfrog: 😐🔫

LibreNMS/Data/Store/Rrd.php
```diff
@@ -481,25 +483,26 @@ public function purge($hostname, $prefix)
...
-
-            $output = implode($this->sync_process->sendCommand($cmd));
-
-            d_echo("<p>$cmd</p>\n<p>command returned ($output)</p>");
-
-            return $output;
-        } else {
-            return 0;
+        $cmd = $this->buildCommand('graph', '-', $options);
+        $process = Process::fromShellCommandline(Config::get('rrdtool') . ' ' . $cmd);
```
Jellyfrog, line 496: Since we're not doing any redirection or so, why aren't we building an array as argument, and use the "normal" command? So we don't have to worry about escaping and such?
Or can `$options` contain that? 

> If you need to use stream redirections, conditional execution, or any other feature provided by the shell of your operating system, you can also define commands as strings using the fromShellCommandline() static factory.
> 
> Each operating system provides a different syntax for their command-lines, so it becomes your responsibility to deal with escaping and portability.

murrant: Because I don't want to change 20,000 lines of code to fix every single graph.  The way graphs were designed in Observium is to concatenate rrdtool options onto a string and run that.

Jellyfrog:
(Triggering tests)

murrant:
You can't get me with your tricky tests :stuck_out_tongue:

murrant:
Just re-tested this and it works.

PR #12736, opened 2021-04-09
webui - Services bootstrap enable + status

PipoCanaja:
Adding bootstrap checkbox to enable/disable service polling, and display graphically if alerts are enabled or disabled.

Before: 
Capture d’écran 2021-04-09 à 14 38 42

After:
Capture d’écran 2021-04-09 à 14 38 25

#### Please note

#### Testers


includes/html/forms/update-service-status.inc.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+/*
+ * LibreNMS
+ *
+ * Copyright (c) 2014 Neil Lathwood <https://github.com/laf/ http://www.lathwood.co.uk/fa>
```
murrant, line 6: Would prefer not to add legacy code.  Also, this is the wrong copyright.
See, app/Http/Controllers/PortController.php for an example

PipoCanaja: Agreed.
I could modify the includes/html/forms/create-service.inc.php which also updates, to make it a little bit more generic and allow update of one variable without providing all of them (as it is required now). Would you prefer that ?

PipoCanaja: Last commit is getting rid of the new ajax_form file, and makes the original "create_service" versatile. Can now handle updates of any field that comes in the request. I had to take care of the checkboxes that are not sent when not checked. 
Let me know which version you prefer.

murrant (approved):
It feels a little odd that you can disable the service on the list page, but not disable alerting.

Other than that, LGTM.  Can I volunteer you to port this page to Laravel later? :D

PipoCanaja:
@murrant I thought having both editable might be confusing. Disabling Alterting is something I rarely do, compared to disable the service itself. So I only improved the visual displayed of the boolean (including a hint that if the device is disabled, it won't obviously generate alerts even if alerts are still enabled). 
And I cannot promise for the `laravelisation` of the page :) I am a little bit uncomfortable with bigger (and deeper) changes. But never say never :)

PipoCanaja:
rerun tests

PR #12746, opened 2021-04-12
Make docker app work with rrdcached

djamp42:
Okay opening up another pull request on the docker over rrdcached issue, i have moved the methods in the \LibreNMS\Store\Rrd class and tested with both rrdcached and without and both load docker app just fine now. The only thing that is still a question for me is the sync vs async for sending the command to rrdtool. I forced it to sync from what the commend said. 

If this is approved i can open another pull requests and fix all the other apps that are broken with rrdcached. 

Prior pull requests here
https://github.com/librenms/librenms/pull/12704
https://github.com/librenms/librenms/pull/12689

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Data/Store/Rrd.php
+++ b/LibreNMS/Data/Store/Rrd.php
@@ -494,7 +494,6 @@ class Rrd extends BaseDatastore
         $offset = substr_count($app_name, $separator);
 
         foreach ($rrdfile_array as $rrd) {
-
             if (str_contains($rrd, $pattern)) {
                 $filename = basename($rrd, '.rrd');
                 $entry = explode($separator, $filename, 4 + $offset)[3 + $offset];
--- a/LibreNMS/Data/Store/Rrd.php
+++ b/LibreNMS/Data/Store/Rrd.php
@@ -384,6 +384,8 @@ class Rrd extends BaseDatastore
         if ($command == 'last' && $this->init(false)) {
             // send this to our synchronous process so output is guaranteed
             $output = $this->sync_process->sendCommand($cmd);
+        } elseif ($command == 'list') {
+            $output = $this->sync_process->sendCommand($cmd);
         } elseif ($this->init()) {
             // don't care about the return of other commands, so send them to the faster async process
             $output = $this->async_process->sendCommand($cmd);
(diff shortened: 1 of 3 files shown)


LibreNMS/Data/Store/Rrd.php
```diff
@@ -442,6 +444,65 @@ public function buildCommand($command, $filename, $options)
...
+            $pattern = sprintf('%s-%s-%s-%s', 'app', $app_name, $app_id, $category);
+        } else {
+            $pattern = sprintf('%s-%s-%s', 'app', $app_name, $app_id);
+        }
+
+        foreach ($rrdfilearray as $rrd) {
+            
+            if (str_contains($rrd, $pattern)) {
+                $filename = basename($rrd, '.rrd');
+                $entry = explode($separator, $filename, 4 + $offset)[3 + $offset];
```
Jellyfrog, line 496: Offset doesn't exist

LibreNMS/Data/Store/Rrd.php
```diff
@@ -384,6 +384,8 @@ private function command($command, $filename, $options)
         if ($command == 'last' && $this->init(false)) {
             // send this to our synchronous process so output is guaranteed
             $output = $this->sync_process->sendCommand($cmd);
+        } elseif ($command == 'list') {
```
murrant, line 387: This should be combined with the above if statement.
```php
if (in_array($command, ['last', 'list']) && $this->init(false)) {
```
The second statement makes sure the sync process is started.

LibreNMS/Data/Store/Rrd.php
```diff
@@ -442,6 +444,67 @@ public function buildCommand($command, $filename, $options)
...
+     */
+    public function getRrdFiles($device)
+    {
+        if ($this->rrdcached) {
+            $filename = sprintf('/%s', $device['hostname']);
+            $rrd_files = $this->command('list', $filename, '');
+            // Command output is an array, create new array with each filename as a item in array.
+            $rrdfilearray = preg_split('/\s+/', trim($rrd_files[0]));
+        } else {
+            $rrddir = self::dirFromHost($device['hostname']);
```
murrant, line 462: Should be `$rrddir = $this->dirFromHost($device['hostname']);`

Rrd is a singleton that stays loaded in memory after it is started.   The external alias \Rrd is a static way to access that singleton.

LibreNMS/Data/Store/Rrd.php
```diff
@@ -442,6 +444,67 @@ public function buildCommand($command, $filename, $options)
...
+     * @param string  $app_name name of app to be searched
+     * @param string  $category which category of graphs are searched
+     * @return array $rrdfilearray array of rrd files for this host
+     */
+    public function getRrdApplicationArrays($device, $app_id, $app_name, $category = null)
+    {
+        $entries = [];
+        $separator = '-';
+
+        $rrdfilearray = self::getRrdFiles($device);
```
murrant, line 484: `$rrdfilearray = $this->getRrdFiles($device);`

LibreNMS/Data/Store/Rrd.php
```diff
@@ -442,6 +442,67 @@ public function buildCommand($command, $filename, $options)
...
+     * @param array $device device for which we get the rrd's
+     * @return array $rrdfilearray array of rrd files for this host
+     */
+    public function getRrdFiles($device)
+    {
+        if ($this->rrdcached) {
+            $filename = sprintf('/%s', $device['hostname']);
+            $rrd_files = $this->command('list', $filename, '');
+            // Command output is an array, create new array with each filename as a item in array.
+            $rrdfilearray = preg_split('/\s+/', trim($rrd_files[0]));
```
Jellyfrog, line 458:
```suggestion
            $rrdfilearray = explode("\n", trim($rrd_files[0]));
```

Jellyfrog: The last line will be  `"OK u:0.01 s:0.01 r:0.02"` , maybe we should drop it?

Jellyfrog:
I added sorting (so both ways return the result in the same way)

murrant (approved):

f0o:
@Jellyfrog @murrant Just noticed that this (specifically https://github.com/librenms/librenms/pull/12746/files#diff-fb57ae1ea0f72f33f98641088606fd88c0f3b1f774eccd2828af80d39ee50803R456) fails with rrdcached 1.4.8 - Unsure when `list` operation was introduced but seeing how rrdcached 1.4.x is supported and https://github.com/librenms/librenms/blob/master/LibreNMS/Data/Store/Rrd.php#L68 defaults to 1.4x it might befenit from a strict version check

murrant:
@f0o not sure what to do about it, since we don't know when rrdlist was introduced...

rrdtool version is detected at runtime now, so it should never fall back to 1.4

f0o:
@murrant what I mean is that anything calling `getRrdApplicationArrays` will return empty array if the installation uses rrdcached and is not on a very recent version. It might be fine with actually removing that whole rrdcached thing unless somebody knows which version introduces `list`. I can say 1.4.x doesnt but 1.7.x does.

murrant:
Aren't all supported os on 1.7+?

PR #12758, opened 2021-04-14
Misc cleanup

Jellyfrog:

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Maps/DeviceDependencyController.php
+++ b/app/Http/Controllers/Maps/DeviceDependencyController.php
@@ -154,7 +154,8 @@ class DeviceDependencyController extends MapController
             $devices_by_id = $this->highlightDevices($devices_by_id, $this->parentDeviceIds);
         }
 
-        array_multisort(array_column($device_list, 'label'), SORT_ASC, $device_list);
+        $device_list_labels = array_column($device_list, 'label');
+        array_multisort($device_list_labels, SORT_ASC, $device_list);
 
         $group_name = DeviceGroup::where('id', '=', $group_id)->first('name');
         if (! empty($group_name)) {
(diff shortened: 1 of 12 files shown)


app/Http/Controllers/Widgets/GlobeController.php
```diff
@@ -102,7 +102,7 @@ public function getView(Request $request)
                     $location->lat,
                     $location->lng,
                     $location->location,
-                    $count ? (1 - $up / $count) * 100 : 0, // percent down
```
Jellyfrog, line 105: `$count` is "verified" on line 100 already

app/Models/Device.php
```diff
@@ -155,7 +155,7 @@ public function isUnderMaintenance()
                     $query->where('alert_schedulables.alert_schedulable_id', $this->device_id);
                 });
 
-                if ($this->groups) {
+                if (count($this->groups)) {
```
Jellyfrog, line 158: `if ($this->groups)` is always true since it's a collection

app/Http/Controllers/Maps/DeviceDependencyController.php
```diff
@@ -154,7 +154,8 @@ public function dependencyMap(Request $request)
             $devices_by_id = $this->highlightDevices($devices_by_id, $this->parentDeviceIds);
         }
 
-        array_multisort(array_column($device_list, 'label'), SORT_ASC, $device_list);
```
Jellyfrog, line 157: While it seems to work anyway, array_multisort require a variable passed by reference

app/Models/Device.php
murrant: $this-groups->isNotEmpty() would probably be better then :)

PR #12764, opened 2021-04-19
Corrected Chassis Over Temp oid for state indexes

craig-nokia:
Updated the discovery timos.yaml file to correct the state_translations table for the over temperature alarm.

#### Please note

#### Testers


includes/definitions/discovery/timos.yaml
```diff
@@ -82,7 +82,7 @@ modules:
                         - { value: 8, generic: 0, graph: 0, descr: GreenBlink }
                         - { value: 9, generic: 2, graph: 0, descr: RedBlink }
                 -
-                    oid: tmnxChassisTable
+                    oid: tmnxChassisOverTempState
```
PipoCanaja, line 85: Hi
What was wrong with the initial oid ? Looks good according to the MIB file at least.

craig-nokia: See if I can explain it...

The tmnxChassisTable has 10 different states for alarms and in the same table, the tmnxChassisOverTempState has only 2 states.  When discovery happens, it overwrites the values 1 and 2 in the state translation table.

Before:  The Critical LED would say "Overtemp"
image
With the patch, the correct alarm text is displayed "Red :
image

PipoCanaja: Got it.
The issue should be solved in a different way. 
Just keep current OID and VALUE, and add this statement to ensure the state is correct for this entry. : 
` state_name: tmnxChassisOverTempState `
Basically, any state from this table without a `state_name` defined should be corrected.

PipoCanaja (changes requested):

PipoCanaja:
Could you check the current YAML file ?

craig-nokia:
Thanks PipoCanaja!   Adding the state_names to the oid: tmnxChassisTable is good to go for the timos.yaml file.

PipoCanaja (approved):
LGTM

PR #12765, opened 2021-04-19
Added metric for MySQL Slave Lag (secs)

rasssta:
Added metric for MySQL Slave Lag (in seconds) to be able to build Alert Rules when slave lag value is greater than 0 (or X). It's possible to add graphs and other metrics as well, but the priority for me is to be able to build alert rules based on the "Slave Lag" metric.

#### Please note

#### Testers

### DIFF
--- a/includes/polling/applications/mysql.inc.php
+++ b/includes/polling/applications/mysql.inc.php
@@ -96,6 +96,7 @@ $mapping = [
     'CSt'     => 'c5',
     'CUe'     => 'c3',
     'CUMi'    => 'c9',
+    'SlLa'    => 'br',
 ];
 
 $data = explode("\n", $mysql);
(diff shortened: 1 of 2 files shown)


Jellyfrog (approved):

murrant:
Changing the RRD definition does not work :)

rasssta:
That's correct, my bad. However, the RRD part could actually be removed in total as the main use is to get the metric in to the DB (and not display it in the Web UI with RRD data). Any suggestions on how to proceed?

murrant:
No way to proceed without breaking some people now.

Matiasjrt:
> No way to proceed without breaking some people now.

Hi Murrant,
I think this change has broken my historical mysql rrd. This is possible?
I have checked the output of the ./poller.php -h <HOST> -m applications, everything seems to be fine but when I go to check the .rrd it is not being updated. I have tried renaming the app-mysql-51153.rrd to app-mysql-51153.rrd.bkp and when a new one is created automatically, it works correctly. 
I have compared the new to the old, from what I have been able to see with the rrdtool info, the DS SlLa [index = 79] has been added but when changing the definition of the .rrd it seems not to be compatible.
Is there any way to manipulate the old .rrd without this DS to make it compatible? Or should the old information be discarded and keep the new one?

rrdupdate
preupdate
beforeWA

PR #12766, opened 2021-04-19
Adjust Alcatel-Lucent aos7

paulierco:
I had to adjust few things after movement to yaml from php

#### Please note

#### Testers


paulierco:
Err, why are tests failing ?

PipoCanaja:
And you can run the tests locally on your server: 
`./lnms dev:check -o aos7`

Jellyfrog:
> You have to re-generate all existing test-data (`./scripts/save-test-data.php -o aos7 -v ""` and all other variants) (yes the "" empty variant exists :) ) and commit them here.

It's better to rename the test data to include the variant tho :)

paulierco:
Very strange. 

On my productive server I do not have any errors from ./lnms dev:check.
On my dev server I have two errors.

There were 2 failures:

1) LibreNMS\Tests\Unit\LocationTest::testDnsLookup
--- Expected
+++ Actual
@@ @@
 Array (
-    'lat' => 51.50354111111111
-    'lng' => -0.12766972222222223
 )

/opt/dev/alcateldev/tests/Unit/LocationTest.php:136

2) LibreNMS\Tests\Unit\SqliteTest::testMigrationsRunWithoutError
could not find driver (SQL: PRAGMA foreign_keys = ON;)



Could not run migration ?) on SQLite

/opt/dev/alcateldev/tests/Unit/SqliteTest.php:47

ERRORS!
Tests: 2278, Assertions: 4345, Errors: 7, Failures: 2, Skipped: 194.
failed (47.55s)

Jellyfrog (dismissed):

Jellyfrog:
Euhm... Well for sqlite your missing the php extension. For the rest i don't really know. It differes with uppercase, do you develop on Windows or so?

paulierco:
No, ubuntu 18.04. I hate windows.

murrant:
@paulierco my guess is you have a state already in the state_indexes table and it will not update because something is case insensitive.  Go delete it from your state_indexes table and it will create a new one with the correct case.

paulierco:
Nope, it doesn't work. I will do another installation these days, probably i'm missing something.

includes/definitions/discovery/aos7.yaml
```diff
@@ -121,7 +121,7 @@ modules:
                 -
                     oid: chasControlModuleTable
                     value: chasControlSynchronizationStatus
-                    num_oid: '.1.3.6.1.4.1.6486.801.1.1.1.3.1.1.1.1.4.{{ $index }}'
+                    num_oid: '.1.3.6.1.4.1.6486.801.1.1.1.3.1.1.1.1.5.{{ $index }}'
```
PipoCanaja, line 124: According to the mib, it seems to be still the wrong OID : 
1.3.6.1.4.1.6486.800.1.1.1.3.1.1.1.1.6 | chasControlSynchronizationStatus

paulierco: Fixed. Add some other changes. Still the same. 

> OS aos7: Discovered ports data does not match that found in tests/data/aos7.json

This is something new to me. I tried to fix only some sensors. Can be that something was actually broken before?

includes/definitions/discovery/aos7.yaml
```diff
@@ -161,7 +167,8 @@ modules:
                         - { value: 1, generic: 1, graph: 1, descr: notRunning }
                         - { value: 2, generic: 0, graph: 1, descr: running }
                 -
-                    oid: chasEntPhysOperStatus
+                    oid: chasEntPhysicalEntry
```
murrant, line 170: Why are you walking the table now? doens't seem like you use anything else below.  Also please use chasEntPhysicalTable instead of chasEntPhysicalEntry

paulierco: This is even worse. I did it to see if this is changing the test results in a positive way. 
chasEntPhysicalTable  is bigger than chasEntPhysicalEntry. 

I will stick with `oid: chasEntPhysOperStatus`.

paulierco:
Lesson learned: don't touch the old tests.
What i did? I re-download current tests and collect new tests and push the change. My old tests were wrong.

paulierco:
I will work in the next days to migrate AOS6 to yaml. Currently is the same old way as AOS7 was, php.

includes/definitions/discovery/aos7.yaml
murrant: chasEntPhysicalEntry chasEntPhysicalTable should have the same output because the table references the entry....

murrant (approved):

PR #12779, opened 2021-04-22
Device: updated Nexus (nxos) os information and test data

Torch09:
Adds a discovery yaml for Nexus Devices.

Currently on every discovery the hardware, version and serial are overwritten with wrong values.
For example:
Hardware: N3K-C3064PQ-10GX -> ciscoModules.3.1.3.1163
Serial: FOCabcdef12 ->
OS Version: 7.0(3)I7(8) ->

after that, the poller sets the values to the right ones on the next poll.

Now the values are discovered with the right OIDs. 
These are the same values which are used by the poller (nxos.inc.php)

Added Test-Data for Nexus N3K-3064pq

#### Please note

#### Testers


PipoCanaja:
@Torch09 
In fact, Cisco HW version and serial are discovered in a specific (non standard) way. So you can remove the YAML alltogether and keep what is currently (correctly) discovered/polled. 

Thanx for the test data, we need some for these devices.

Torch09:
@PipoCanaja 
So it should replace the values on every discovery and the nexus discovery yaml isn't needed?

PipoCanaja:
Hi @Torch09 
I misread your post. What must be checked is that this YAML file works in all case compared to current code (which run in the poller only, and results in this "flap" of the HW, Ver and Serial.
Current poller code is : 
`includes/polling/os/nxos.inc.php`
And we see that a preg_replace takes place for the HW : `    $hardware = str_replace(' Chassis', '', $data[149]['entPhysicalName']);`. So flapping would completely disappear only if this is also done in YAML.

Then, next step is to run again the `save-test-data.php -o nxos -v '' ` to fix current nxos test data.

And you'll be good to go.

murrant (dismissed):
I'm not sure these are universally the correct values...

murrant (dismissed):
You forgot to delete the legacy nxos.inc.php file, fixed that for you :)

murrant (approved):

PR #12787, opened 2021-04-23
Add juniper rpm support

geg347:
Hello dear LibreNMS Community,

In this pull request I add support for Juniper RPM the same way it is already working for Cisco SLA.

So I created a discovery module juniper-rpm. It fills the same SQL table slas with a mock sla_nr field for compatibility (the real "key" for searching data is owner + tag). The code is querying the DISMAN-PING-MIB with OID for pingCtlEntry field (1.3.6.1.2.1.80.1.2.1).

I created the poller modules juniper-rpm that querying the DISMAN-PING-MIB with OID for pingCtlEntryfield (to get status), with OID for pingResultsEntry field and the JUNIPER-PING-MIB with OID for jnxPingResultsEntry field.
I did the same as Cisco SLA by creating a first RRD file with only the Round Trip Time. And I added support for IcmpEcho and IcmpTimeStamp Juniper RPM types in RRD file sla-<sla_nr>-<type>. The extra MIB values that I put in this second RRD are MInRttUs, MaxRttUs, StdDevRttUs, ProbeResponses and ProbeLoss. (if needed it is posiblle to add other value).

I adapted the code to display graphs. So I created graph and page files for those 2 RPM types. I am currently displaying 2 simple grapsh for RTT and for PacketLoss in percentage. (here again you can add whatever you want).

I tried to integrate properly this feature by declaring the modules, adding them in junos.yaml device file. I also declared the Juniper RPM types.

If I missed some part like the documentation, maybe the unit test (also not done for Cisco SLA) or if you see errors or possible improvements in those changes feel free to tell me!

I hope this feature will be useful!! I did it to completely migrate fro Observium to LibreNMS in my enterprise.
Have a nice week-end!

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [ ] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [ ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #12799, opened 2021-04-27
New module add juniper rpm support + reimplementation of cisco-sla module

geg347:
Hello dear LibreNMS Community,

In this pull request I added support for Juniper RPM and I reimplemented the Cisco SLA existing module.
It is following my previous Pull Request ([Add juniper rpm support #12787](https://github.com/librenms/librenms/pull/12787)).

So I created a new discovery module called Slas. This module follow the new way of defining module.
The Juniper RPM part fills the same SQL table slas with a mock sla_nr field for compatibility (the real "key" for searching data is owner + tag). The code is querying the DISMAN-PING-MIB with OID for pingCtlEntry field (1.3.6.1.2.1.80.1.2.1).

I created the corresponding poller modules Slas (now declared in the same file as the discovery part) that querying the DISMAN-PING-MIB with OID for pingCtlEntry field (to get status), with OID for pingResultsEntry field and the JUNIPER-PING-MIB with OID for jnxPingResultsEntry field.
I did the same as Cisco SLA by creating a first RRD file with only the Round Trip Time. And I added support for IcmpEcho and IcmpTimeStamp Juniper RPM types in RRD file sla-<sla_nr>-. The extra MIB values that I put in this second RRD are MInRttUs, MaxRttUs, StdDevRttUs, ProbeResponses and ProbeLoss. (if needed it is posiblle to add other value).

I adapted the code to display graphs. So I created graph and page files for those 2 RPM types. I am currently displaying 2 simple graphs for RTT and for PacketLoss in percentage. (here again you can add whatever you want).

I tried to integrate properly this new module and to delete the old cisco-sla one. It may need some changes with the way II select the device (if loop by os or os_group).

If I missed some part like the documentation or if you see errors or possible improvements in those changes feel free to tell me!

I was about to collect test data when I saw existing data and snmpsim files for junos_rpm, could I use them for the unit test?
Is it possible to have help for implementing those tests, please? I never did unit test..

I hope this feature will be useful!! I did it to completely migrate fro Observium to LibreNMS in my enterprise.

#### Please note

#### Testers


geg347:
I don't understand the two last errors with the continuous-integration test...
For me it is the same lines.

murrant:
Sounds amazing @geg347 I'll try to look it over later.

murrant:
Could be a line return in there or something.  Try downloading and applying the diff from styleci.  Or ru `lnms dev:check style`

murrant:
FYI, to fix the test failure, you need to add sla to misc/os_schema.json.
We also probably need to convert anyone having the old cisco slas module enabled in their config to this.  This can be done in LibreNMS/Config.php with deprecatedVariable()

geg347:
So I tried to convert old cisco-sla to slas new one the same way Jellyfrog did when he reimplemented Printer Supplies in LibreNMS/Config.php. Could you tell me if it is enough or if I missed something, please?

For the test failure, I tried to replace cisco-sla by slas but I am sure it is not enough. Could you tell me what to do please? I am not (yet) familiar with Unit Test.

Thank you!

Jellyfrog:
> FYI, to fix the test failure, you need to add sla to misc/os_schema.json.
> We also probably need to convert anyone having the old cisco slas module enabled in their config to this. This can be done in LibreNMS/Config.php with deprecatedVariable()

One idea is to enable test capture for sla (or is it already?) in a separate PR and then generate tests for it *before* merging this, then we will see if this pull request will affect the current data

murrant:
Might be hard to convert test data from before, but could be possible.  I apologize @geg347 I have not read the code yet and was kind of thinking ahead.  Glad you looked at printer supplies PR, that is a good example.

geg347:
Hello,
I added collected data from my virtual Juniper VMX router and define module slas in tests/module_tables.yaml the same way printer_supplies. Then I executed the checks with :
```lnms dev:check unit -m slas```
And I got OK (2 tests, 7 assertions) success
I also tried for JunOS with : 
```lnms dev:check unit -o junos```
And I got OK (28 tests, 2433 assertions) success
Is it enough for Unit Test, please?

**Comment by PipoCanaja** [comment]
Hi @geg347 
Currently testing this successfully on Cisco device (don't have any Juniper).
Looks very good indeed, Thx !

**Comment by geg347** [comment]
> Hi @geg347
> Currently testing this successfully on Cisco device (don't have any Juniper).
> Looks very good indeed, Thx !

Hello @PipoCanaja,
OK glad to hear this.
No problem!

File: LibreNMS/Modules/Slas.php [hunk 1]
```diff
@@ -0,0 +1,614 @@
...
+    public function discover(OS $os)
+    {
+        $device = $os->getDeviceArray();
+
+        if ($device['os_group'] == 'cisco') {
+            $this->discoverSlas($device);
+        } elseif ($device['os'] == 'junos') {
+            $this->discoverRpms($device);
+        }
+        // } else if ($device['os'] == 'huawei')
```
PipoCanaja, line 48: That's for me, I guess 👍

PipoCanaja (dismissed):
LGTM

LibreNMS/Modules/Slas.php
geg347: Yes you said you need the feature for Huawei. It is probably not the right way to target Huawei's device though. I let you implement this part.

geg347:
> LGTM

Thank you for checking!

LibreNMS/Modules/Slas.php
```diff
@@ -0,0 +1,614 @@
...
+     * Try to keep this efficient and only run if discovery has indicated there is a reason to run.
+     * Run frequently (default every 5 minutes)
+     *
+     * @param OS $os
+     */
+    public function poll(OS $os)
+    {
+        $device = $os->getDeviceArray();
+
+        if ($device['os_group'] == 'cisco') {
```
murrant, line 65: The best way to implement this is inheritance.  Look at some of the other modules.
There is an interface on the OS class that is called to poll something.  If the OS needs to override the default discovery/polling, just override that method in that OS's class file.

When multiple os use the same oids, we can use a trait (such as HostResources.php) to share the code.

PipoCanaja: @geg347 : you can check how Mempools module does with the inheritance. That way, dedicated OS code ends up in OS class.

geg347: Alright I will check that tomorrow.
For Juniper RPM the mib used in discovery seems standard and could work for other vendors. But for polling there are 3 mibs with one specific for Juniper.
I will check Mempools, thanks!

LibreNMS/Modules/Slas.php
```diff
@@ -0,0 +1,614 @@
...
+     * @param OS $os
+     */
+    public function cleanup(OS $os)
+    {
+        $os->getDevice()->printerSupplies()->delete();
+    }
+
+    private function discoverRpms($device)
+    {
+        $slas = snmp_walk($device, 'pingMIB.pingObjects.pingCtlTable.pingCtlEntry', '-OQUs', '+DISMAN-PING-MIB');
```
PipoCanaja, line 89: Hi @geg347 
DISMAN is indeed a standard mib, you are right. Meaning other vendors could support it. In that case, instead of moving DISMAN code to Juniper class, it would make sense to keep it here. 
Cisco code, on the other end, should move to specific Cisco OS classes. 
And Huawei (VRP) is also using proprietary so I'll use the class model as well.

In term of discovery and polling : 
- Discovery should try to call OS and standard (DISMAN) 
- Polling should poll only whatever was discovered -> best performance.

geg347:
Hello,
So I tried to implement the discovery part with inheritance. 

I found a file LibreNMS/OS/Shared/Cisco.php so I put the discovery code for Cisco in it.

I put the DISMAN discovery part in LibreNMS/OS/Traits/HostResources.php as @murrant adviced.

I need to do continue with the poller part in LibreNMS/OS/Junos.php and LibreNMS/OS/Shared/Cisco.php

But it was quite difficult to understand how it works so if you can give me advice it will be nice, please.
First should I push my modifications in this branch (and so  this Pull Request) or should create a third Pull Request, please?

PipoCanaja:
Hi @geg347 
Yes, pushing the changes here is the good way to go. No need to create a new one. Will try to review the PR tomorrow. 
Bye

geg347:
Hello,
Thanks for checking this!
I don't know how to fix the 2 PHP Static Analysis errors left. With the implementation of inheritance the 2 functions pollSlas and retrieveJuniperType are not defined if the class is not Junos or Cisco. Should I implement these functions elsewhere and keep them empty, please?
Have a nice evenning!

PipoCanaja:
> Hello,
> Thanks for checking this!
> I don't know how to fix the 2 PHP Static Analysis errors left. With the implementation of inheritance the 2 functions pollSlas and retrieveJuniperType are not defined if the class is not Junos or Cisco. Should I implement these functions elsewhere and keep them empty, please?
> Have a nice evenning!

Hello @geg347 
The idea here is to use "interfaces" of the Object model. 

https://en.wikipedia.org/wiki/Object_model

Basically, the idea is that ALL classes instances must implement what is mandatory for their type. So Junos.php class must implement all OS class elements. But then, how can you implement a new module only for some classes ? By using interfaces. So Junos.php will "implement" the Interface "SlaDiscovery" and the interface "SlaPolling" which will respectively define the discovery and the polling code.

So an interface is a way to add a common set of features to a class, which can later be used in a general way. 

`~/LibreNMS/Interfaces/Polling`and `~/LibreNMS/Interfaces/Discovery`

geg347:
Hello @PipoCanaja,
Thank you for your advices.
So I tried to use Interfaces, is it good like this, please?
Also I movedt the discoverSlas function in Junos class. But it can work with other vendors. I just don't know where to put this code. I tried in HostResources but it was not working.

LibreNMS/Interfaces/Polling/SlaPolling.php
```diff
@@ -0,0 +1,35 @@
...
+namespace LibreNMS\Interfaces\Polling;
+
+interface SlaPolling
+{
+    /**
+     * Poll Sla data for Sla in database.
+     *
+     * @param Collection $slas
+     */
+    public function pollSlas(Collection $slas);
```
Jellyfrog, line 34: missing a `use whatever/path/it/might/be/Collection`

geg347: Yep thanks I have just fixed this.

geg347:
Hello,
Do you think this PM can be integrated by the 15th of June?
Do not hesitate to challenge me if some changes/improvements are needed.

murrant:
@geg347 that index was wrong...  I've gone with crc32

murrant (approved):

geg347:
Hello, thank you very much for your job on this!
I saw your improvements and it looks really good. Specially the way you improved the Poller to query only the needed variables with snmpwalk_group. The performances will be way better.
Also adding the rtt in database seems to be a good idea. It's almost the only variable we actually need (except the probesResponses and sentProbes for the Packetloss).
Thanks again and have a great day!

PR #12805, opened 2021-04-28
Revert "RRD Graph optimization"

murrant:
Reverts librenms/librenms#12735


murrant:
Some sort of problem with rrdcached.

PR #12811, opened 2021-04-29
Remove debug globals

murrant:

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Util/Debug.php
+++ b/LibreNMS/Util/Debug.php
@@ -1,0 +1,74 @@
+<?php
+/*
+ * Debug.php
+ *
+ * -Description-
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @package    LibreNMS
+ * @link       http://librenms.org
+ * @copyright  2021 Tony Murray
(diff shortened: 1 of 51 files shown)


app/Http/Controllers/DeviceController.php
```diff
@@ -11,8 +11,11 @@
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Auth;
 use LibreNMS\Config;
+use LibreNMS\Util\Debug;
 use LibreNMS\Util\Graph;
 use LibreNMS\Util\Url;
+use const false;
+use const true;
```
Jellyfrog, line 18: What does these do?

includes/common.php
```diff
@@ -427,27 +425,6 @@ function del_dev_attrib($device, $attrib_type)
     return DeviceCache::get((int) $device['device_id'])->forgetAttrib($attrib_type);
 }
 
-/*
- * convenience function - please use this instead of 'if ($debug) { echo ...; }'
- */
-if (! function_exists('d_echo')) {
```
Jellyfrog, line 433: This is used a bit here and there

LibreNMS/Util/Debug.php
```diff
@@ -0,0 +1,137 @@
...
+use Illuminate\Database\Events\QueryExecuted;
+use LibreNMS\DB\Eloquent;
+use Log;
+
+class Debug
+{
+    private static $debug = false;
+    private static $verbose = false;
+
+    public static function set($debug = true, bool $silence = false): bool
```
Jellyfrog, line 38: Why not typehint debug?

app/Http/Controllers/DeviceController.php
murrant: stupid intellij put those in there...  I'll remove them.

includes/common.php
murrant: @Jellyfrog it is a duplicate function. There is still an implementation.

LibreNMS/Util/Debug.php
murrant: I did at first, but we are supplying this function to an on boot startup, and the first argument sent to the function is an application.  So, the options are change that, make a specific function, or make this less strict.   I went with the later and added the cast below.

Jellyfrog (approved):
I'm all for removing globals

PR #12825, opened 2021-05-03
Alcatel-Lucent Aos7 no bulk fix

paulierco:

#### Please note

#### Testers


PipoCanaja:
Hello @paulierco 
Why did you write a new _nobulk function ? The original is already using "_nobulk" if you configure it in YAML.

paulierco:
Because...that will put everything in nobulk. I discovered that this increased the polling time for some of our core switches from 100s to 500s. 


There is only one state sensor that doesn't work with bulk. Damn Alcatel. 
I copy the other function to create only a nobulk function just for that small sensors. Probably is dirty and ugly.

The other way would be just to drop that state sensor but there is no way to have nobulk globally for OS.

image

PipoCanaja:
Hi @paulierco 
Another nicer option would be to extend YAML syntax to support this nobulk setting on a "per-sensor" basis. This is not currently possible but could be achieved with some changes in `function gen_snmpwalk_cmd` definition in `includes/snmp.inc.php`.

PipoCanaja:
And for next time, you need to explain the change you are suggesting in the Pull Request description. This allows the reviewers to understand what you are trying to achieve (and why)

paulierco:
> 
> 
> Hi @paulierco
> Another nicer option would be to extend YAML syntax to support this nobulk setting on a "per-sensor" basis. This is not currently possible but could be achieved with some changes in `function gen_snmpwalk_cmd` definition in `includes/snmp.inc.php`.

Not sure how to do that. If someone can add this to YAML, i can modify the change and move it back.

paulierco:
> 
> 
> And for next time, you need to explain the change you are suggesting in the Pull Request description. This allows the reviewers to understand what you are trying to achieve (and why)

You're right. My mistake here. I will take care of this next time. Sorry.

murrant (changes requested):
Instead of duplicating code, you should set nobulk temporarily: `Config::set('os.aos7.nobulk', true)`  then `Config::forget('os.aos7.nobulk')`

Also, do not use pre-fetch when it the data is only used in one place.  Pre-fetch makes things less optimal in this case.

This could be added to yaml snmp options if desired.

paulierco:
> Instead of duplicating code, you should set nobulk temporarily: `Config::set('os.aos7.nobulk', true)` then `Config::forget('os.aos7.nobulk')`

This should be done [here ](https://github.com/librenms/librenms/blob/bb39b4ad83f7348a9b51d8cab998f5923ddcfb2f/includes/snmp.inc.php#L132)?

paulierco:
I think i know how to do this, i will try to make the changes in the next two days.

paulierco:
I will reopen this after https://github.com/librenms/librenms/pull/12833 is finished.

PR #12827, opened 2021-05-04
CyberPower UPS Updates

arrmo:
1) Update to latest MIB (v2.2 => v2.9)
2) Correct scaling factors (matching updated info in MIB)
3) Add Load - confirmed to work

NOTE: Tried running script to collect data - but it's failing. Not sure why ... it seems to think the host doesn't exist, but poller.php works just fine with it.
NOTE: If Current is not supported (some UPS's) ... calculate it from Load (percentage, rated) and Voltage?

Thanks!

#### Please note

#### Testers


arrmo:
OK, I think the issue with the scripts not working is because this is not on a standard port (i.e. not on port 161). Can the scripts handle that? Not seeing a parameter to set that, but it may be me.

Thanks!

arrmo:
Yes, trying to (add tests)! But as above, the test script is failing 😞. Thinking it's because the UPS is not on port 161 ... is there a way to get the script to work on another port?

Thanks!

Jellyfrog:
What does it say? as far as I know the testing should work exactly the same as normal regarding ports

arrmo:
> What does it say? as far as I know the testing should work exactly the same as normal regarding ports

It's rather odd - could be me ... LOL. So if I poll the device,
```
./poller.php -h ups-winserver
LibreNMS Poller
Starting polling run:

Hostname:    ups-winserver
Device ID:   105
OS:          cyberpower
Assigned IP: winServer
```

Proceeds from there, no issues at all - data all valid and good. But, if I try to run the tests,
```
./scripts/collect-snmp-data.php -h ups-winserver -o cyberpower
Script to collect snmp data from devices to be used for testing.
Snmp data is saved in tests/snmpsim.

Usage:
  You must specify an existing device to collect data from.
```

Huh? I did specify an existing device (as poller.php confirms). Thoughts?

Thanks!

Jellyfrog:
You didn't specify variant, it's a required argument

arrmo:
> You didn't specify variant, it's a required argument

I think that's new, no? I don't ever recall doing that in the past ... LOL.

Sorry, but - what is a "variant"? It's not the OS or hostname, not quite sure what it is (and what to use in this case). Thanks!

Jellyfrog:
It's the model/variant. So in your case the model name of your cyberpower ups

arrmo:
> It's the model/variant. So in your case the model name of your cyberpower ups

That was it, thanks! OK, I know I didn't do that in the past ... LOL! But I do like it, good improvement. FYI, I had to use the "variant" that was reported as Hardware, from discovery.

Test Data added now, let's see if all the tests are happy 🤣.

Thanks again.

arrmo:
OK, on the test failures - I have seen that one before 😜. It's because I changed (corrected) the scaling factor. Do I need to remove the older cyberpower.json file?

Thanks!

arrmo:
Yep, that was it! Had to remove the older test data, with the old / incorrect scaling factor.

Thanks for the help!

murrant:
@arrmo it is better to correct the expected data (aka the json file) than to delete it.

arrmo:
OK, NP. It seemed like the naming has changed now - I can re-run / re-capture, then re-name the output files. Make sense?

Thanks!

murrant:
You need to specify variant with -v,  for ones without a variant, just put `-v ''`

arrmo:
> You need to specify variant with -v, for ones without a variant, just put `-v ''`

Thought that was the secret! 😄. But no joy, still only get the following from `collect-snmp-data.php`,
```
1.3.6.1.2.1.1.1.0|4|UPS Local
1.3.6.1.2.1.1.2.0|6|1.3.6.1.4.1.3808
```

But it's working fine with the poller. For example, the same device,
```
#### Load poller module sensors ####
Checking (snmp) charge Battery Capacity...
100 %
Checking (snmp) frequency Output...
0 Hz
Checking (snmp) load Load...
24 %
Checking (snmp) runtime Battery Runtime...
20.5 Min
Checking (snmp) state Battery Status...
Checking (snmp) state Battery Replace Indicator...
Checking (snmp) state Input Line Cause...
Checking (snmp) state Input Status...
Checking (snmp) state Output Status...
2 #
2 #
1 #
1 #
2 #
Checking (snmp) voltage Battery Voltage...
Checking (snmp) voltage Input Line Voltage...
Checking (snmp) voltage Output Voltage...
24 V
123 V
123 V
```

Thoughts? Very odd that the tests aren't working - with or without a blank variant.

Thanks!

murrant:
collect-snmp-data.php is currently broken :upside_down_face: #12837

arrmo:
LOL! C'mon man, I don't have much hair left to pull out 🤣.

NP, will wait for that fix, then update the test files here. Actually glad to hear this, thought I was losing my mind.

Thanks!

arrmo:
> collect-snmp-data.php is currently broken 🙃 #12837

OK, that helped - thanks! Rebased off of master (after your merge for this correction) ... now data looks better. Not quite sure why the data is failing in some cases now - I looked at the results, but not seeing the issue 😞.

Appreciate it!

arrmo:
Never mind 😜. That was me being stupid - dang it ... missed one old file hanging around (before updating test runs). Tests all clean now!

Jellyfrog:
Since the current test data is from another device, please  don't overwrite it and fix the tests files  manually if possible

arrmo:
> Since the current test data is from another device, please don't overwrite it and fix the tests files manually if possible

I can put the older files back, add a variant name to this one - NP. I admit, wasn't thinking I should do that, but I can 😄.

The issue is that the older file has incorrect scaling factors => when I tried this, then the tests fail. And I'm not sure I see a real way to fix that. Thoughts?

Thanks!

arrmo:
> Since the current test data is from another device, please don't overwrite it and fix the tests files manually if possible

OK, put the old data back in place, set mine up as a variant. Hopefully that's the intention 😄.

As I saw before, some of the legacy data fails - because the scaling factor was incorrect. This is one of the changes, to correct this, as outlined in the MIB (and also here, https://community.librenms.org/t/cyberpower-v2-9-mib/12546/3)

Thoughts?

murrant:
You have to change the specific test data.

The way we capture test data can capture both correct and incorrect values.  So if the code is wrong when the test data is captured, we need to update existing test data.  We should always manually review the data and make sure it looks righ.

arrmo:
> We should always manually review the data and make sure it looks righ.

Gotcha! Modified the single measurement (scaling factor, and corresponding result). All seems OK now.

Thanks!

murrant (approved):
Nice work!

PR #12829, opened 2021-05-04
Fix allow_unauth_graphs

nightcore500:
This patch fixes the functionality of the setting "allow_unauth_graphs" which does not work anymore since the "Rrd graph optimizations" patch. The variable $auth is already set in html/graph.php when the setting "allow_unauth_graphs" is set. But this variable gets overwritten in includes/html/graphs/graph.inc.php. With this patch the variable $auth is initialized in html/graph.php before checking if the user is allowed to view the graph.

#### Please note

#### Testers

### DIFF
--- a/html/graph.php
+++ b/html/graph.php
@@ -10,6 +10,7 @@
 use LibreNMS\Data\Store\Datastore;
 use LibreNMS\Util\Debug;
 
+$auth = false;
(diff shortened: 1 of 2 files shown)


murrant (approved):
Thanks @nightcore500 looks correct :)

PR #12842, opened 2021-05-08
Collect MAC vendors and display them

PipoCanaja:
This PR is collecting MAC OUI from https://macaddress.io/database/macaddress.io-db.json and caching them using standard cache mecanism. This occurs during daily.sh.
Values are (currently) cached for 7 days, and not refreshed before 2.5 days.

Then, webui can display the vendor aside the mac-address, like so : 
Capture d’écran 2021-05-08 à 09 34 38

This is a draft for the moment, we need to define if the caching mecanism (vs database storage) is fine, and the refresh times (I suspect this list to be fairly static so we could increase the refresh interval. )
 
#### Please note

#### Testers


Jellyfrog:
There is also https://gitlab.com/wireshark/wireshark/-/raw/master/manuf
Not sure which is best.

PipoCanaja:
I like the JSON for the ease of loading. But it is certainly bigger to DL and longer to process. We'll see, it is not a difficult change anyway.

PipoCanaja:
STP page is now supported as well : 
Capture d’écran 2021-05-09 à 17 18 50

and

Capture d’écran 2021-05-09 à 17 19 48

murrant:
@PipoCanaja This looks great. Only question I have is why disabled by default? :D

dagbdagb:
Per device FDB tables no longer loads for me? Says 'Loading...' forever.

PipoCanaja:
> @PipoCanaja This looks great. Only question I have is why disabled by default? :D


That's something I am happy to discuss :D Now that it seems the performance impact is almost null. 

Other point to discuss: timers for refreshing the DB. I don't expect it to change much (and if you have a new vendor on this list, it takes probably a few months until you actually receive a device in your network using a new vendor anyway). We should probably extend the cache time some more. (1 month and refresh after 3 weeks + random for instance).

PipoCanaja:
> Per device FDB tables no longer loads for me? Says 'Loading...' forever.

Do you have the "DNS" column visible ? This behaviour occurs when the DNS column is refreshing, causing PTR DNS requests to be sent for each IP address of your table. 
If not, then it means I have an issue with this part of the code (detection if DNS column is displayed or not).

dagbdagb:
I do not have the DNS column enabled. But per device FDB tables works now.
The 'default enabled'  bit, is that for caching only, or also for the web UI? Because I still have to manually enable it in the web UI.
Allow me to yet again give praise for this feature. Love it.

dagbdagb:
If there isn't a list of these 'frequently requested features not yet implemented' on the website, shouldn't there be one? Low(?) hanging fruit for someone itching to contribute?

Also, this particular feature (remembering views from one page view to the next) could/would technically be solved with a cookie, no?

PipoCanaja:
> If there isn't a list of these 'frequently requested features not yet implemented' on the website, shouldn't there be one? Low(?) hanging fruit for someone itching to contribute?

community forum, "feature request" tag, for now.

> Also, this particular feature (remembering views from one page view to the next) could/would technically be solved with a cookie, no?

Most probably a way to do it. Another is to store it in the DB with the user profile.

murrant:
@PipoCanaja the problem with the saved table settings is we are in need of a new JS library for tables.  The current one is basically unmaintained.  Run into choice paradox when trying to find a new one.

PR #12847, opened 2021-05-10
Filter unwanted data in Routes Ajax reply

PipoCanaja:
Currently, eloquent DB result is sent as the AJAX reply, effectively sending more column than necessary for the bootgrid table to be filled. 
For my test, 32k for reply instead of 15k after filtering only the expected columns, with this : 
```
        $var_list = ['context_name' => '',
            'inetCidrRouteDestType' => '',
            'inetCidrRouteDest' => '',
            'inetCidrRoutePfxLen' => '',
            'inetCidrRouteNextHop' => '',
            'inetCidrRouteIfIndex' => '',
            'inetCidrRouteMetric1' => '',
            'inetCidrRouteType' => '',
            'inetCidrRouteProto' => '',
            'created_at' => '',
            'updated_at' => '',
        ];
        // Lets filter only the column needed in the AJAX, and not all the DB columns that we selected
        $item = array_intersect_key($item, $var_list);
```

@murrant @Jellyfrog I am pretty sure there is a better way to do this. Let me know. Probably extending the Controller to do this by default would be the best thing to do (all TableControllers would benefit more or less from this kind of optimisation). 

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Table/RoutesTablesController.php
+++ b/app/Http/Controllers/Table/RoutesTablesController.php
@@ -169,58 +169,54 @@ class RoutesTablesController extends TableController
      */
     public function formatItem($route_entry)
     {
-        if ($route_entry->updated_at) {
-            $item['updated_at'] = $route_entry->updated_at->diffForHumans();
+        $item['updated_at'] = $route_entry->updated_at ? $route_entry->updated_at->diffForHumans() : $route_entry->updated_at;
+        $item['created_at'] = $route_entry->created_at ? $route_entry->created_at->toDateTimeString() : $route_entry->created_at;
+        $item['inetCidrRouteIfIndex'] = $route_entry->inetCidrRouteIfIndex == 0 ? 'Undefined' : $route_entry->inetCidrRouteIfIndex;
+        $item['inetCidrRouteMetric1'] = $route_entry->inetCidrRouteMetric1;
+        $item['inetCidrRoutePfxLen'] = $route_entry->inetCidrRoutePfxLen;
+        $item['inetCidrRouteDestType'] = $route_entry->inetCidrRouteDestType;
+
+        try {
+            $obj_inetCidrRouteDest = IP::parse($route_entry->inetCidrRouteDest);
+            $item['inetCidrRouteDest'] = $obj_inetCidrRouteDest->compressed();
+        } catch (\Exception $e) {
+            $item['inetCidrRouteDest'] = $route_entry->inetCidrRouteDest;
         }
-        if ($route_entry->created_at) {
-            $item['created_at'] = $route_entry->created_at->toDateTimeString();
-        }
-        if ($route_entry->inetCidrRouteIfIndex == 0) {
-            $item['inetCidrRouteIfIndex'] = 'Undefined';
-        }
-        if ($route_entry->inetCidrRouteNextHop) {
-            try {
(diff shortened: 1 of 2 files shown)


Jellyfrog:
You should extend the table controller and add a new function "select" here;
https://github.com/librenms/librenms/blob/be54aa3e6de1c516ac6e4262d6c13973aa07aa16/app/Http/Controllers/Table/TableController.php#L63-L65

so that in your controller you can add a function to only select some fields

PipoCanaja:
AJAX reply is now cleaner. Will add tooltips as well to the Route display (ipv6 cannot be displayed on smaller screens currently) and it will be OK for this one.

PipoCanaja:
Tooltips installed in the route view.
image

app/Http/Controllers/Table/RoutesTablesController.php
```diff
@@ -196,28 +194,33 @@ public function formatItem($route_entry)
...
             } else {
-                $item['inetCidrRouteNextHop'] = $item['inetCidrRouteNextHop'] . '<br>(' . Url::deviceLink($device) . ')';
+                $item['inetCidrRouteNextHop'] = $route_entry->inetCidrRouteNextHop . '<br>(' . Url::deviceLink($device) . ')';
             }
         }
+        $item['inetCidrRouteProto'] = $route_entry->inetCidrRouteProto;
         if ($route_entry->inetCidrRouteProto && $route_entry::$translateProto[$route_entry->inetCidrRouteProto]) {
             $item['inetCidrRouteProto'] = $route_entry::$translateProto[$route_entry->inetCidrRouteProto];
         }
+        $item['inetCidrRouteType'] = $route_entry->inetCidrRouteType;
```
murrant, line 214: Looking a bit better, instead of building the array ad-hoc, you can do this:
```php
return [
    'inetCidrRouteType' => $route_entry::$translateType[$route_entry->inetCidrRouteType] ?? $route_entry->inetCidrRouteType,
];
```

PipoCanaja: I tried to make it that way but with the try catch for the exceptions, it would be a mess. I could prepare all the values for the array (one variable for each returned value), and return the dynamically built array, but it would not improve readability.

In term of performance, I really think it does not make a difference nowerdays. 

I think we have a more compact and readable way of filtering these variables now, compared to before, and for sure more efficient network-wise.

murrant: Mostly, I value readability over other things.
Sometimes compactness and readability go together.

if statements are poison for readability ;)

PipoCanaja: Yep, but I really don't see how to avoid the ones that are still around now.

Jellyfrog:
Can't you add a select() up in basequery? Then it won't be unnecessary selected from the database either...

PipoCanaja:
> Can't you add a select() up in basequery? Then it won't be unnecessary selected from the database either...

We could select less than "*" but we still need to filter to remove all the columns we need to compute but don't want to display (like port_id is necessary so we can create create the link, but we don't want to send it as a column). 
So I went directly as @murrant suggested with the variant that create the $item[] empty instead of converting the copying all the data selected into $item[].

PipoCanaja:
@librenms/reviewers Anything else to change here ?

Jellyfrog (approved):

PR #12853, opened 2021-05-11
feature: Added support for Nexus PVLAN MAC in FDB

Torch09:
Hello,
We’ve had a problem where Cisco Nexus Devices doesn’t show the port on which a MAC Address is connected inside the FDB Table.
 It’s not populated if the MAC is learned from a port which is configured with private VLANs.

Normally the Switch returns the dot1dTpFdbPort which can be mapped to a physical port, for example:
```
dot1dTpFdbPort[34:e3:80:f1:7e:e0] = 4105
vlan 1988 mac 34e380f17ee0 port (4105) 11217
```

If pvlan is active, the returned port cannot be mapped to a port_id. 
The switch returns the number of the physical port, not the dot1dBasePort number which is needed for the mapping.
Example:
`dot1dTpFdbPort[34:e3:80:a7:37:10] = 44`

To get the portId the “ports”-table is queried, and the returned value is added to the portId.
For example, a port which is configured with pvlan:
```
dot1dTpFdbPort[34:e3:80:a7:37:10] = 44
Found BasePort without portID, Port: 44
Interface-Name: Ethernet1/44, trying to find a match in ports Table
SQL[SELECT * FROM `ports` WHERE `device_id` = ? AND `ifName` = ? [1816,"Ethernet1\/44"] 0.74ms]
vlan 1988 mac 34e380a73710 port (44) 11261
```

To be completely safe, this is only done if the returned $dot1dBasePort is < 55 and the device is a N3K-C30, because this is the only switch group I can test. If somebody could test this against N5k or N7k and the problem is present there, this could be extended.
The “nxos.inc.php” is a copy from “ios.inc.php” so these tests doesn’t need to be checked against non nxos devices.

How are your opinions? 
If something needs to be changed, I’ll try, PHP isn’t really my world :D

Thanks!

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/fdb-table.inc.php
+++ b/includes/discovery/fdb-table.inc.php
@@ -19,9 +19,12 @@ foreach ($sql_result as $entry) {
 $insert = []; // populate $insert with database entries
 if (file_exists(Config::get('install_dir') . "/includes/discovery/fdb-table/{$device['os']}.inc.php")) {
     require Config::get('install_dir') . "/includes/discovery/fdb-table/{$device['os']}.inc.php";
-} elseif ($device['os'] == 'ios' || $device['os'] == 'iosxe' || $device['os'] == 'nxos') {
-    //ios,iosxe,nxos are all Cisco
+} elseif ($device['os'] == 'ios' || $device['os'] == 'iosxe') {
+    //ios,iosxe are all Cisco
     include Config::get('install_dir') . '/includes/discovery/fdb-table/ios.inc.php';
+} elseif ($device['os'] == 'nxos') {
+    //nxos check
+    include Config::get('install_dir') . '/includes/discovery/fdb-table/nxos.inc.php';
 }
 
 if (empty($insert)) {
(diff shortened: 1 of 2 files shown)


Torch09:
@Jellyfrog 
I can't add any Tests. 
They are the same as before. The N3k tests where already mine and it seems like MAC related informations aren't stored in in these tests.

murrant:
If tests are already there, your fix should make them fail. If not, then the code makes no noticeable change.

Torch09:
Hello @murrant , sorry for the late reply.
The code only adds a fallback option for Nexus 3k Ports which couldn’t be to a dot1dTpFdbPort because of the returned value. It doesn’t add/remove or changes something which could be picked up by the test units. To be picked up, the “collect-snmp-data” scripts needs to add FDB data to its checks, IMHO it doesn’t do this. I couldn’t find a MAC form the Switches FDB inside the data.
The fallback is only working with data that exists inside LibreNMSs “ports” table, so there are no additional SNMP calls.

Added a picture which represent how LibreNMS currently shows Ports, which are configured with PVLANs and how its shown after the code changes. (Sorry for the MS paint magic :D )
libre_nxs_mac

If you need some data to verify something I’ll try to add them.

murrant:
It is supported, but not included by default.  Run with the `-m fdb-table` option to include the data.

Jellyfrog:

PR #12865, opened 2021-05-12
HWG WaterLeak sensor support

PipoCanaja:

#### Please note

#### Testers


murrant:
Any reason this needs to be a new OS?

includes/definitions/hwg-wld.yaml
```diff
@@ -0,0 +1,42 @@
+os: hwg-wld
+text: 'HWg WLD'
+type: environment
+icon: hwg
+mib_dir: hwg
+over:
+    - { graph: device_state, text: States }
+nobulk: true
+poller_modules:
```
murrant, line 9: Please don't do this in OS.  It blocks users global settings.

PipoCanaja: I kept whatever was existing in the original OS, but as I own one WLD device, I'll check if it is really necessary. Basically, the complete snmpwalk of the device is 10 lines long (dumb device, really, does only what it is meant to do), meaning that any discovery or poller will essentially timeout :) I think this was the reason the original device included a lot of disabled modules.

PipoCanaja:
> Any reason this needs to be a new OS?

Yes. They have the same sysObjectID but are completely different. Different MIB (but with same object names) which I patched to remove the collision. Still, the `over:` display for the hwg-ste is temperature (makes sense for a temperature sensor). This one is a water leak detector, no temperature, only states. So it looked ugly. But technically, aside of the `over:`, the 2 OS may be merged (and renamed hwg, the brand, without any specific model).

includes/definitions/hwg-wld.yaml
PipoCanaja: I cleaned hwg-wld.yaml for module restrictions. It works, but it has an impact on the poller time (from 600 ms to 1.8 s). It is definitely a waste of time because clearly the device has only 10 lines in the snmpwalk. Won't change anytime soon. 
Let me know what you prefer. I would personnaly keep the restrictions for now, and if, one day, they create a new model, we'll always be accepting a new PR to remove the restrictions.

murrant: Which modules?  Poller modules that aren't polling data should not add to the polling time.  I have tried to make that the case for most of them.

murrant:
Makes sense, fine with the separate OS, just wanted to make sure.
Really interested about the modules that add polling time...

PipoCanaja:
> Makes sense, fine with the separate OS, just wanted to make sure.
> Really interested about the modules that add polling time...

The 3 longer poller modules: 
- netstats polling takes 0,55 seconds
- ports polling takes 0,35 seconds 
- ucd-mib takes 0.13 seconds

None of them return a single OID, of course. In this device, the snmpwalk is as below: 
```
.1.3.6.1.2.1.1.1.0 = HWg-WLD
.1.3.6.1.2.1.1.2.0 = .1.3.6.1.4.1.21796.4.1
.1.3.6.1.2.1.1.3.0 = 0:17:14:47.00
.1.3.6.1.2.1.1.4.0 = XXXXXXXX
.1.3.6.1.2.1.1.5.0 = xxxx-pm7-wld-001
.1.3.6.1.2.1.1.6.0 = xxxx Cafeteria
.1.3.6.1.2.1.1.7.0 = 72
.1.3.6.1.4.1.21796.4.5.4.1.1.1 = 1
.1.3.6.1.4.1.21796.4.5.4.1.2.1 = "Cafeteria"
.1.3.6.1.4.1.21796.4.5.4.1.3.1 = 1
.1.3.6.1.4.1.21796.4.5.4.1.4.1 = "27C0A91E02590ADA"
.1.3.6.1.4.1.21796.4.5.4.1.5.1 = 215
.1.3.6.1.4.1.21796.4.5.4.1.6.1 = 0
.1.3.6.1.4.1.21796.4.5.70.1.0 = "00:0A:59:02:1E:A9"
```

murrant:
I'd suggest disabling just those 3 for now.  These are all "poller only" modules :)

PipoCanaja:
@librenms/reviewers OK with you guys ?

Jellyfrog (approved):

PR #12885, opened 2021-05-15
OS support for West Mountain 4005i DC PDU

dorkmatt:
New OS support for [West Mountain 4005i](http://www.westmountainradio.com/product_info.php?products_id=rr_4005i) DC PDU. This is a 12VDC power distribution unit that is geared for "_amateur radio, emergency communications systems and repeater site operators_".

Note that the device support itself is quite limited, this is powered by a PIC. The SNMP implementation is quite strange: IF-MIB doesn't exist at all, SNMPv2-MIB::sysServices returns a `-1` value, none of the custom MIB data is available in a table, the device serial number isn't available _(expect as a default hostname)_, etc. I've reported to the vendor suggestions to make the implementation a bit more industry standard, it's unclear if an update will be available to address these concerns.

Due to the lack of any SNMP tables, I tried to avoid the use of `{{ $index }}` values - when I did attempt to use indexes, they would either work in the poller or the discovery - but not both; thus I avoided their use entirely.

#### Please note

- [x ] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
image

#### Testers


includes/definitions/westmountainradio.yaml
```diff
@@ -0,0 +1,58 @@
...
+over:
+    - { graph: device_current, text: Current }
+    - { graph: device_voltage, text: Voltage }
+    - { graph: device_state, text: 'Power Status' }
+discovery:
+    -
+        sysObjectID:
+            - .1.3.6.1.4.1.15117.1
+poller_modules:
+    applications: false
```
Jellyfrog, line 16: Please don't disable modules if not needed due to buggy devices

mibs/westmountainradio/rr4405i.mib
```diff
@@ -0,0 +1,523 @@
+-- MIB.  Import this into your MIB Browser software.
```
Jellyfrog, line 1: Please rename this file to `RMCU`

Jellyfrog:
Hi

Your missing step 2 of the test data, see https://docs.librenms.org/Developing/os/Test-Units

includes/definitions/westmountainradio.yaml
dorkmatt: Ok, left only enabled modules - noting that this increases polling & discovery time substantially

mibs/westmountainradio/rr4405i.mib
dorkmatt: Renamed

dorkmatt:
> Your missing step 2 of the test data, see https://docs.librenms.org/Developing/os/Test-Units

Added JSON test data

mibs/westmountainradio/rr4405i.mib
Jellyfrog: Need to remove the mib extension also

includes/definitions/westmountainradio.yaml
Jellyfrog: Please Remove all, it's really only meant for buggy devices

mibs/westmountainradio/rr4405i.mib
dorkmatt: Done

Jellyfrog:
Please see failing tests at https://github.com/librenms/librenms/pull/12885/checks?check_run_id=2596005767

dorkmatt:
> Please see failing tests at https://github.com/librenms/librenms/pull/12885/checks?check_run_id=2596005767

The unit test appears to be regex for lacking `$index` in the `num_oid` discovery file. The device doesn't have any SNMP tables in their MIB, thus I assumed that indexes should be avoided. Will updating the unique OID to use `{{ $index_string }}` be acceptable _(guessing not based on the regex)_?

Sadly I'm unable to run the CI test offline, I don't know how to debug this? Strangely all the other CLI tools like poller.php, discovery.php, etc work fine to access the database.

```
$ lnms dev:check unit --db --snmpsim -o westmountainradio -vvvv
SQL[select `config_name`, `config_value` from `config` [] 1ms] 
  
SQL[select * from `graph_types` [] 0.52ms] 
  
Only checking os: westmountainradio
Running unit check... 
Starting snmpsim listening on 127.1.6.2:1162... 
PHPUnit 9.5.4 by Sebastian Bergmann and contributors.

Error in bootstrap script: PDOException:
SQLSTATE[HY000] [1698] Access denied for user 'root'@'localhost'
failed (6.76s)
```

PipoCanaja:
Hello
Please try to change the YAML for all sensors to this scheme: 
```
                -
                    oid: RMCU::analog1val-int
                    num_oid: '.1.3.6.1.4.1.15117.1.4.1.{{ $index }}'
                    index: '1-{{ $index }}'
                    descr: "Outlet 1"               
```

`$index` will be "0" for all of them, and this is how the mib defines it actually. 
Of course, you'll need to re-compute the tests.

PR #12886, opened 2021-05-15
Alcatel-Lucent aos7 LLDP Neighbors

paulierco:
LLDP Neighbors for Alcatel-Lucent Aos7 is working only for chassis ports.

If the switch has a module installed, LLDP Neighbors  is missing for module ports and doesn't show because data using dot1dBasePortIfIndex from BRIDGE-MIB is not there.

I'm pretty sure is not the greatest way to do this, i looked to some code but at least for AOS7 is working. I'm not sure if I broke something else now. 

I started this in order to get help here.

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/discovery-protocols.inc.php
+++ b/includes/discovery/discovery-protocols.inc.php
@@ -239,7 +239,6 @@ if (($device['os'] == 'routeros')) {
             } else {
                 $ifIndex = $entry_key;
             }
-	 
             if (($device['os'] == 'aos7')) {
                 $local_port_id = find_port_id($ifName, null, $device['device_id']);
             } else {
--- a/includes/discovery/discovery-protocols.inc.php
+++ b/includes/discovery/discovery-protocols.inc.php
@@ -221,21 +221,31 @@ if (($device['os'] == 'routeros')) {
                 }
             }
         }
+	if (($device['os'] == 'aos7')) {
+		$lldp_local = snmpwalk_cache_oid($device, 'lldpLocPortEntry', array(), 'LLDP-MIB');
+		$lldp_ports = snmpwalk_group($device, 'lldpLocPortId', 'LLDP-MIB');
+	} else {
 
         $dot1d_array = snmpwalk_group($device, 'dot1dBasePortIfIndex', 'BRIDGE-MIB');
         $lldp_ports = snmpwalk_group($device, 'lldpLocPortId', 'LLDP-MIB');
+	}
     }
 
     foreach ($lldp_array as $key => $lldp_if_array) {
         foreach ($lldp_if_array as $entry_key => $lldp_instance) {
-            if (is_numeric($dot1d_array[$entry_key]['dot1dBasePortIfIndex'])) {
+		if (($device['os'] == 'aos7')) { 
+			$ifName = $lldp_local[$entry_key]['lldpLocPortDesc'];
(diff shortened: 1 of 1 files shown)


includes/discovery/discovery-protocols.inc.php
```diff
@@ -221,21 +221,31 @@ function ($v) {
...
+            $dot1d_array = snmpwalk_group($device, 'dot1dBasePortIfIndex', 'BRIDGE-MIB');
+            $lldp_ports = snmpwalk_group($device, 'lldpLocPortId', 'LLDP-MIB');
+        }
     }
 
     foreach ($lldp_array as $key => $lldp_if_array) {
         foreach ($lldp_if_array as $entry_key => $lldp_instance) {
-            if (is_numeric($dot1d_array[$entry_key]['dot1dBasePortIfIndex'])) {
+            if (($device['os'] == 'aos7')) {
+                $ifName = $lldp_local[$entry_key]['lldpLocPortDesc'];
```
PipoCanaja, line 236: instead of this $ifName and the SELECT you do below, wouldn't it be possible to tweak ` $local_port_ifName = $lldp_ports[hexdec($lldp_ports_num[$key]['mtxrNeighborInterfaceID'])]['mtxrInterfaceStatsName'];` so that all the remaining code can be kept unchanged ?

PipoCanaja (changes requested):

PipoCanaja:
Hello @paulierco 
A better way would be for sure to have a class based discovery here, so the OS not matching default LLDP implementation would be less dependant and the code would be more readable. 
Your code seems to only make changes to aos7 OS, meaning that if you are 100% sure all aos7 devices are behaving better or same as before your patch, it should be OK from a functional point of view.

paulierco:
> 
> 
> Hello @paulierco
> A better way would be for sure to have a class based discovery here, so the OS not matching default LLDP implementation would be less dependant and the code would be more readable.
> Your code seems to only make changes to aos7 OS, meaning that if you are 100% sure all aos7 devices are behaving better or same as before your patch, it should be OK from a functional point of view.

Thanks for reviewing this PR. 
The current LLDP code for AOS7 works but only for chassis ports, sometimes this is like half of the ports. If there is a module installed, output from/for that module is different and LLDP will not work. 

I had to apply this to only aos7 because half of the output is missing, adding this will change the priority how to handle missing dot1dBasePortIfIndex output.

includes/discovery/discovery-protocols.inc.php
PipoCanaja: The idea would be to avoid any dbFetchRow (we are trying to get rid of them, and at least not to add any new one). So if you can get the $ifIndex right, and $local_port_id as well, then get_port_by_id will do it's magic. At the end, code is easier to read and understand and maintain :)

paulierco: So now i got rid of `dbFetchRow` and i changed only `$local_port_id`
Does this look better now from your point of view?

PipoCanaja:
Force tests re-run

includes/discovery/discovery-protocols.inc.php
PipoCanaja: yup :)

PipoCanaja:
OK @paulierco

PipoCanaja (approved):
LGTM

paulierco:
Thanks @PipoCanaja

PR #12889, opened 2021-05-17
Initial HAProxy ALOHA support

Jellyfrog:
Basic support for HAProxy ALOHA

#### Please note

#### Testers


Jellyfrog:
Unsure if I should call the os `haproxy` or `aloha`, there is both the software haproxy, and then both virtual and hardware appliances. This was only tested on a hardware version; https://www.haproxy.com/products/aloha-hardware-load-balancer/

includes/definitions/discovery/aloha.yaml
```diff
@@ -0,0 +1,48 @@
...
+                        - { value: 2, generic: 1, graph: 1, descr: DOWN }
+                        - { value: 1, generic: 0, graph: 1, descr: UP }
+
+                -
+                    oid: alServerTable
+                    value: alServerStatus
+                    num_oid: '.1.3.6.1.4.1.23263.4.2.1.3.4.1.19.{{ $index }}'
+                    index: 'alServerStatus.{{ $index }}'
+                    descr: alServerName
+                    #group: "{{ $alBackendTable }}.{{ $alServerBackendID }}"
```
Jellyfrog, line 41: I would like this to work, but its not possible at the moment.
If i pre-cache it it returns an array like;
```
alBackendName.1.2 = "stats"
alBackendName.1.17 = "api-library-search"
```
Problem is the index sent to the "translation" function is for example `1.17.4`.
`alServerBackendID` contains `17` which would be nice to use (since that's how you *should* map it)

PR #12896, opened 2021-05-18
Added Unpolled Devices rule to collection

ahmedsajid:
Another way of achieving https://github.com/librenms/librenms/pull/12321 

#### Please note

#### Testers

### DIFF
--- a/misc/alert_rules.json
+++ b/misc/alert_rules.json
@@ -525,7 +525,7 @@
   },
   {
     "name": "Unpolled Devices",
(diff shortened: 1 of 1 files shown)


murrant:
Looking at https://github.com/librenms/librenms/blob/master/app/Checks.php#L58  shows you should check isUp() for this alert.
Which can be shown here: https://github.com/librenms/librenms/blob/master/app/Models/Device.php#L402

ahmedsajid:
@murrant did you mean something like this? https://github.com/librenms/librenms/pull/12896/files#diff-5b08d454d858403bff7228f576fb69b131d72812bcb3d850b0b0cca79167c30aR523

murrant:
Negative, you should check last polled and all of these fields:
```
status
ignore
disable_notify
disabled
```

PR #12901, opened 2021-05-20
LLDP - Extend discovery lldp code to support different subtypes

PipoCanaja:
LLDP MIB allows different type of data inside 2 of the fields : **lldpRemChassisId** and **lldpRemPortId**. 
The type of data is provided via 2 other fields : 
**lldpRemChassisIdSubtype**: chassisComponent(1), interfaceAlias(2), portComponent(3), macAddress(4), networkAddress(5), interfaceName(6), local(7)
**lldpRemPortIdSubtype**: interfaceAlias(1), portComponent(2), macAddress(3), networkAddress(4), interfaceName(5), agentCircuitId(6), local(7)

This PR uses a little bit more of those subtypes indications to workaround missing or currently-not-useful LLDP entries. 
Like if PortID is not provided but ChassisID is the ifName, then use that as portID, etc etc...

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/discovery-protocols.inc.php
+++ b/includes/discovery/discovery-protocols.inc.php
@@ -300,7 +300,7 @@ if (($device['os'] == 'routeros')) {
                     $remote_device_id,
                     $remote_port_mac
                 );
-                if ($remote_port_id == 0 ) { //We did not find it
+                if ($remote_port_id == 0) { //We did not find it
                     $remote_port_name = $remote_port_name . ' (' . $remote_port_mac . ')';
                 }
                 if (empty($lldp['lldpRemSysName'])) {
--- a/includes/discovery/discovery-protocols.inc.php
+++ b/includes/discovery/discovery-protocols.inc.php
@@ -242,8 +242,15 @@ if (($device['os'] == 'routeros')) {
             foreach ($lldp_instance as $entry_instance => $lldp) {
                 // normalize MAC address if present
                 $remote_port_mac = '';
+                $remote_port_name = $lldp['lldpRemPortId'];
+                if ($lldp['lldpRemChassisIdSubtype'] == 4) { // 4 = macaddress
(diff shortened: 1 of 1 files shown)


includes/discovery/discovery-protocols.inc.php
```diff
@@ -272,39 +279,43 @@ function ($v) {
...
                     $lldp['lldpRemPortDesc'],
-                    $lldp['lldpRemPortId'],
+                    $remote_port_name,
                     $remote_device_id,
                     $remote_port_mac
                 );
-
+                if ($remote_port_id == 0) { //We did not find it
+                    $remote_port_name = $remote_port_name . ' (' . $remote_port_mac . ')';
+                }
```
PipoCanaja, line 305: If we have no port_id found, it means that the remote device and/or port is unknown from LibreNMS. So let's store both port_name and port_mac, which may help the LibreNMS user.

includes/discovery/discovery-protocols.inc.php
```diff
@@ -272,39 +279,43 @@ function ($v) {
...
                 if (empty($lldp['lldpRemSysName'])) {
                     $lldp['lldpRemSysName'] = $remote_device['sysName'] ?: $remote_device['hostname'];
                 }
-
-                if ($interface['port_id'] && $lldp['lldpRemSysName'] && $lldp['lldpRemPortId']) {
+                if (empty($lldp['lldpRemSysName'])) {
+                    $lldp['lldpRemSysName'] = $lldp['lldpRemSysDesc'];
+                }
```
PipoCanaja, line 311: Some devices do not provide a SysName. They were skipped. This patch allows to use sysDesc in that case, and allow discovery. I had EMULEX cards in servers in that case (the card is directly sending LLDP packets independently of the OS).

Jellyfrog (approved):

PR #12906, opened 2021-05-21
Improve cleanup scripts

haydenseitz:
update `lock_and_purge` shared function to do dbDelete in blocks of 1000 rows, similar to how the cleanup for `syslog` currently happens

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2049,6 +2049,7 @@ function get_device_oid_limit($device)
  */
 function lock_and_purge($table, $sql)
 {
+    // TODO: add block delete statements
     $purge_name = $table . '_purge';
     $lock = Cache::lock($purge_name, 86000);
     if ($lock->get()) {
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2045,11 +2045,11 @@ function get_device_oid_limit($device)
  *
  * @param string $table
  * @param string $sql
+ * @param int $count
  * @return int exit code
  */
-function lock_and_purge($table, $sql)
+function lock_and_purge($table, $sql, int $count = 1000)
 {
-    // TODO: add block delete statements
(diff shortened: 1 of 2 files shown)


haydenseitz:
updated `lock_and_purge` function. not sure how to implement with `lock_and_purge_query`, as if uses dbQuery instead of dbDelete.

Are there integration tests for the daily.sh script?

includes/functions.php
```diff
@@ -2056,9 +2057,16 @@ function lock_and_purge($table, $sql)
 
         $name = str_replace('_', ' ', ucfirst($table));
         if (is_numeric($purge_days)) {
-            if (dbDelete($table, $sql, [$purge_days])) {
-                echo "$name cleared for entries over $purge_days days\n";
+            if (is_numeric($count)) {
```
murrant, line 2060: FYI $count is always numeric, you said it must be an int in the parameters.

haydenseitz: good catch, removed the check

haydenseitz:
after running this a few days in production, I see the loop creates a sort of race condition on certain tables that are updated frequently (device_perf). Any of the SQL statements that use `NOW()` in the WHERE to delete old rows can potentially never return 0 rows, as there are new entries every time it runs.

I'm thinking I can update each of the SQL statements passed to the `lock_and_purge` function to either:
1. get the current time in PHP and use that in each the statement, then it's always looping over the same timestamp until all rows are deleted
2. update each statement to use CURDATE(), which will only delete records up to midnight of X days before, instead of actually X days before

I'm leaning toward option 2 - suggestions?

murrant:
2 seems like the better choice, a fixed time instead of a rolling window.

murrant:
Closing due to inactivity

PR #12932, opened 2021-06-02
Firebrick: Fix an issue which led to duplication of BGP peers.

cjsoftuk:
Fixes an issue where Firebrick BGP Peers would be duplicated on every discovery run.

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/bgp-peers/firebrick.inc.php
+++ b/includes/discovery/bgp-peers/firebrick.inc.php
@@ -97,12 +97,12 @@ foreach ($peers as $value) {
     $address = $value['bgpPeerIdentifier'];
 
     // Cleanup code to deal with 0 vs Null in the DB
-    if($vrfId === 0){
+    if ($vrfId === 0) {
         // Database says it's table 0 - which is wrong.  It should be "null" for global table
         $deleted = dbDelete('bgpPeers', 'device_id = ? AND bgpPeerIdentifier = ? AND vrf_id = ?', [$device['device_id'], $address, $vrfId]);
         echo str_repeat('-', $deleted);
         continue;
-    }else{
+    } else {
         $testVrfId = $vrfId;
     }
 
--- a/includes/discovery/bgp-peers/firebrick.inc.php
+++ b/includes/discovery/bgp-peers/firebrick.inc.php
@@ -33,17 +33,17 @@ foreach ($bgpPeersCache as $key => $value) {
     if (strlen($address) > 15) {
         $address = IP::fromHexString($address)->compressed();
     }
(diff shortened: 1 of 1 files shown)


cjsoftuk:
Hmm... I've not touched the test data - so I guess I must have broken something here......

cjsoftuk:
My local tests pass - the CI tests don't.... that's weird.

@Jellyfrog and/or @PipoCanaja - This fix is kinda important for Firebrick users (otherwise they may end up with thousands of BGP Peers showing) - but I don't actually know how it's broken in CI, but not locally..... there must be something to do with the way the DB is set up.

includes/discovery/bgp-peers/firebrick.inc.php
```diff
@@ -33,17 +33,17 @@
     if (strlen($address) > 15) {
         $address = IP::fromHexString($address)->compressed();
     }
-
-    if (isset($value['fbBgpPeerTableId'])) {
+    if (isset($value['fbBgpPeerTableId']) && $value['fbBgpPeerTableId'] !== '') {
         $bgpPeers[$value['fbBgpPeerTableId']][$address] = $value;
     } else {
-        $bgpPeers[0][$address] = $value;
+        $bgpPeers[null][$address] = $value;
```
PipoCanaja, line 39: $bgpPeers[null][whatever] does not make sense. so this part of the fix is for sure wrong.

includes/discovery/bgp-peers/firebrick.inc.php
```diff
@@ -33,17 +33,17 @@
...
     }
 }
 unset($bgpPeersCache);
 
 foreach ($bgpPeers as $vrfId => $vrf) {
     if (empty($vrfId)) {
+        $vrfId = null;
```
PipoCanaja, line 46: not necessary cause this variable is not used after the empty() test above

includes/discovery/bgp-peers/firebrick.inc.php
PipoCanaja: The idea here is that null can be stored in the DB (and means global routing table) but cannot be stored in a hash table as a key.

PipoCanaja (changes requested):

includes/discovery/bgp-peers/firebrick.inc.php
cjsoftuk: This was in use in the insert - and it was that that was causing part of the issues.  I've put a comment in accordingly.

includes/discovery/bgp-peers/firebrick.inc.php
cjsoftuk: We now use 0 in the hash table but null in the DB (and convert accordingly).

PipoCanaja (approved):
LGTM

PR #12938, opened 2021-06-09
Add OS Support for Siemens Scalance X Switching

rhinoau:
OS support for Siemens Scalance X industrial switches.

Basic device health and states are functional, may need some minor tweaks/explanation with the State graphs values of power supplies etc.

image
image

#### Please note

#### Testers


Jellyfrog:
Please rename the mib file to `SN-MSPS-SCX200-MIB` (without .mib)

includes/definitions/discovery/scalance.yaml
```diff
@@ -0,0 +1,156 @@
+mib: SN-MSPS-SCX200-MIB
+modules:
+    os:
+        sysDescr_regex: '/Siemens, SIMATIC NET, (?<hardware>SCALANCE X.+, HW: .+\d), FW:.+/'
```
Jellyfrog, line 4: You should probably split this to the part after `,` is part of `feature` instead.
I would do like this; `/SCALANCE (?<hardware>[^,]+), (?<feature>[^,]+)/`

rhinoau:
Thanks @Jellyfrog have committed suggested changes.

I have one issue with the automated inventory nesting elements when it shouldn't as below, would like to fix if can get some advice:

image

The BASEXC206-2SFP (Slot0) line should be at the top level and not under the L2 power line.  I can't see anything obvious to cause this in the snmpwalks, though this is my first time attempting such.

rhinoau:
I'm not understanding the cause of the failed test and what I may need to do to correct it: https://github.com/librenms/librenms/runs/2782123481?check_suite_focus=true

```
1) LibreNMS\Tests\OSModulesTest::testOS with data set "scalance_xc206-2sfp" ('scalance', 'xc206-2sfp', array('os', 'ports', 'processors', 'mempools', 'sensors', 'vlans'))
TypeError: preg_match(): Argument #2 ($subject) must be of type string, array given
```

includes/definitions/discovery/scalance.yaml
```diff
@@ -0,0 +1,156 @@
+mib: SN-MSPS-SCX200-MIB
+modules:
+    os:
+        sysDescr_regex: '/SCALANCE (?<hardware>[^,]+), (?<features>[^,]+)/'
+        serial: '.1.3.6.1.4.1.4329.6.3.2.1.1.3.0'
+        version: '.1.3.6.1.4.1.4329.6.3.2.1.1.5.0'
+    processors:
+        data:
+            -
+                # snMspsDiagnosticsUsageTable contains multiple disaparate items with no formal type identifiers, specific OIDs used
```
murrant, line 10: This is why your tests are failing.

rhinoau: Thanks, unfortunately I don't have the knowledge yet to resolve this, so will need additional assistance from other to learn what is occurring and how I need to resolve it. Had a couple of attempts on discord to no consensus.
My local unit tests (php 7.4.3) come out OK, so I'm yet to find a local issue I can keep testing against.

rhinoau: Thanks for some more tips in discord - that I'm probably seeing a nested array instead of a string value.

This is the table, it works for mempool in the same format - when I try to attack it as a full table I never get a correct CPU value no matter what Index I used:

snmpwalk:
```SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageIndex.0 = INTEGER: 0
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageIndex.1 = INTEGER: 1
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsagePresent.0 = INTEGER: 1
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsagePresent.1 = INTEGER: 1
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageName.0 = STRING: "CPU............................................................................................................................................................................................................................................................"
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageName.1 = STRING: "RAM............................................................................................................................................................................................................................................................"
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageStatus.0 = INTEGER: ok(1)
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageStatus.1 = INTEGER: ok(1)
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageValue.0 = INTEGER: 4
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageValue.1 = INTEGER: 59
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageHighWarning.0 = INTEGER: 0
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageHighWarning.1 = INTEGER: 90
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageHighCritical.0 = INTEGER: 0
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageHighCritical.1 = INTEGER: 98
```
snmptable:
```
SNMP table: SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageTable

 snMspsDiagnosticsUsageIndex snMspsDiagnosticsUsagePresent                                                                                                                                                                                                                                        snMspsDiagnosticsUsageName snMspsDiagnosticsUsageStatus snMspsDiagnosticsUsageValue snMspsDiagnosticsUsageHighWarning snMspsDiagnosticsUsageHighCritical
                           0                             1 "CPU............................................................................................................................................................................................................................................................"                           ok                          21                                 0                                  0
                           1                             1 "RAM............................................................................................................................................................................................................................................................"                           ok                          59                                90                                 98
```

As pointed out - the debug was walking the table and returning 2 values for CPU and Memory, as it's not a table only for CPU or only Memory - let's call it a broken table for these purposes.

```
#### Load disco module processors ####
snMspsDiagnosticsUsageValue.0 = 3
snMspsDiagnosticsUsageValue.1 = 59

#### Load disco module mempools ####
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageValue.0 = 3
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageValue.1 = 59
```
If I then hard-code processor, that works and values/graphs good.

If I hard-code Memory I never get a graph but get correct value in debug.  So assuming we have to use a table for mempool, I tried to exclude the erroneous index 0 CPU value, and I can get the graph displaying.

```
#### Load disco module processors ####
snMspsDiagnosticsUsageValue.0 = 3

#### Load disco module mempools ####
SN-MSPS-SCX200-MIB::snMspsDiagnosticsUsageValue.1 = 59
```

I still don't know how to catch these errors with local unit tests - so submitting again to see if it passes this time.

rhinoau: Turns it that was still causing issues with multiple values in mempools, no matter what I tried - left those attempts in the file, but now implemented via php file after following some other examples.

rhinoau:
This is working now by implementing processors and mempools via php to avoid walking the kitchen sink diagnostic table.

The incorrect inventory nesting is not a show-stopper, but if there is an easy fix I'm happy to learn and attempt.

rhinoau:
> The incorrect inventory nesting is not a show-stopper, but if there is an easy fix I'm happy to learn and attempt.

I've had a play and seen how this can be nudged around in includes/discovery/entity-physical/entity-physical.inc.php - though this also further confirms to me it's the device that has it wrong.
To clean it up would require specific changes for every switch model based on container/module associations that would require intimate device knowledge, and any movements by the vendor to correct it would then break it again.  I will log a bug with the vendor, and I think that's the best we can do.

Jellyfrog (approved):

rhinoau:
Siemens have formally agreed via an SR the entPhysical inventory issue is a bug and will push to R&D to fix in future firmware revisions.

PR #12952, opened 2021-06-14
Added New-OS: Alcatel-Lucent Enterprise Stellar Wireless

BennyE:
Added New-OS "Alcatel-Lucent Enterprise Stellar Wireless" (AWOS). This is working fine in my setup, although I had some trouble with the recommended "development environment".

stellar-wireless

#### Please note

#### Testers


Jellyfrog:
Sorry, still missing the json file of the test data

BennyE:
> Sorry, still missing the json file of the test data

I hope we're good now. This is my first contribution to this project and I guess you already noticed this by the number obstacles I found on my path. Thanks for your patience. :)

includes/definitions/discovery/stellar.yaml
```diff
@@ -0,0 +1,38 @@
+modules:
+    os:
+        hardware:
```
Jellyfrog, line 3: You could also extract `version`, seems it's part of sysdescr. Or might be it's own Oid.

If it's only in sysdescr you can do like;

sysDescr_regex: '/\s(?<version>[\d\.]+)/'

Jellyfrog:
> > Sorry, still missing the json file of the test data
> 
> I hope we're good now. This is my first contribution to this project and I guess you already noticed this by the number obstacles I found on my path. Thanks for your patience. :)

No worries. It's not the easiest thing to understand

includes/definitions/discovery/stellar.yaml
BennyE: There is currently no way to get the full version via SNMP. In this case it would be 4.0.3.16 (4.0.3 being the Major release; 16 being the build number; which you can't see in the output). I'm in contact with our R&D to get a MIB/OID for this in the future, but would prefer to not "read out" the improper value now.

That aside, if you'd like to have the "incomplete" version number (that may be misleading to customers) - it could be added.

Jellyfrog (approved):

PR #12962, opened 2021-06-18
Create Laravel Sessions Table

djamp42:
Laravel supports using the database to hold session information. This is creating the table to hold that information, simply migrate in. then set **SESSION_DRIVER=database** in .env

#### Please note

#### Testers

### DIFF
--- a/database/migrations/2021_06_07_123600_create_sessions_table.php
+++ b/database/migrations/2021_06_07_123600_create_sessions_table.php
@@ -1,0 +1,34 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+
+class CreateSessionsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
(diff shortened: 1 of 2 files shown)


murrant:
Probably no harm in having the table exist.  You just need to update the db_schema.yaml file for verification.

djamp42:
> Probably no harm in having the table exist. You just need to update the db_schema.yaml file for verification.

Yes i thought the same, this really has no major benefit unless you do database clustering as in my other pull request, with both this and clustering enabled.. Enabling Nginx on two librenms installations and load balancing between them seem to work perfectly when session data is in the database. Session data was getting lost when the server went offline that has the session data in a file.

murrant:
FYI, I consider redis session driver better, but depends on your infrastructure.

djamp42:
> 
> 
> FYI, I consider redis session driver better, but depends on your infrastructure.

Yes i originally had this, but i didn't have a redis cluster setup, so for what i was doing it was failing, i will eventually move back to this. 

Because it was just a table missing that was preventing the database from being used, i decided we might as well add this as i don't see the harm in it.

murrant (approved):

PR #12963, opened 2021-06-18
Support multiple db servers

djamp42:
I was testing a galara cluster and while that was working fine, librenms was still limited to a single connection point into the database. You would have to edit the .env file to point to another galara node to get the database back online. 

Larvael supports multiple database connections for read and write connections to the database. [Read-and-write-connections](https://laravel.com/docs/8.x/database#read-and-write-connections). I just put each galara/mariadb node into each of the read and write arrays. 

I had to leave the current method for backwards compatibility and it broke a couple of things, i believe lnms db command breaks. 'host' => env('DB_HOST', 'localhost'), because it broke a couple of things when it was missing. 

Here is an example of the what the .env would look like for 3 db servers in a galara cluster. 
DB_HOST=192.168.1.35
DB_HOST2=192.168.1.36
DB_HOST3=192.168.1.37
DB_DATABASE=librenms
DB_USERNAME=librenms
DB_PASSWORD=password

If you are running a mariadb server instance on the actual librenms machine or a distributed poller, you should set DB_HOST to the ip address of that machine, ie the mariadb also running on it. A couple of things like the lnms command do not failover and only use DB_HOST, so it's best to have that one set to the local machine. 

I can't find anything broken in this, it works pretty well as far as i can tell. So putting this out there to see what could be improved. 

#### Please note

#### Testers

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -75,11 +75,11 @@ return [
                 'host' => [
                     env('DB_HOST', 'localhost'),
                     env('DB_HOST_R2', ''),
-	            env('DB_HOST_R3', ''),
+                    env('DB_HOST_R3', ''),
                     env('DB_HOST_R4', ''),
                     env('DB_HOST_R5', ''),
                     env('DB_HOST_R6', ''),
-	            env('DB_HOST_R7', ''),
+                    env('DB_HOST_R7', ''),
                     env('DB_HOST_R8', ''),
                     env('DB_HOST_R9', ''),
                 ],
--- a/config/database.php
+++ b/config/database.php
@@ -116,7 +116,6 @@ return [
             ]) : [],
         ],
 
-
         'testing' => [
             'driver' => env('DB_TEST_DRIVER', 'mysql'),
             'host' => env('DB_TEST_HOST', 'localhost'),
--- a/config/database.php
+++ b/config/database.php
@@ -97,6 +97,7 @@ return [
                     env('DB_HOST_W9', ''),
                ],
             ],
+            'sticky' => env('DB_STICKY', null),
             'driver' => 'mysql',
             'host' => env('DB_HOST', 'localhost'),
             'port' => env('DB_PORT', '3306'),
--- a/config/database.php
+++ b/config/database.php
@@ -95,7 +95,7 @@ return [
                     env('DB_HOST_W7', ''),
                     env('DB_HOST_W8', ''),
                     env('DB_HOST_W9', ''),
-               ],
+                ],
             ],
             'sticky' => env('DB_STICKY', null),
             'driver' => 'mysql',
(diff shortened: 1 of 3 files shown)


murrant:
Ok, I think we need to make a separate connection profile for this in the config/database.php file.

And some docs would be nice even if they are only a short paragraph and link to the Laravel docs.

djamp42:
> Ok, I think we need to make a separate connection profile for this in the config/database.php file. Maybe called mysql_cluster ?

Yes, that will work even better i think.

> And some docs would be nice even if they are only a short paragraph and link to the Laravel docs.

No problem will add also.


> 
> 
> Ok, I think we need to make a separate connection profile for this in the config/database.php file. Maybe called mysql_cluster ?
> 
> And some docs would be nice even if they are only a short paragraph and link to the Laravel docs.

config/database.php
```diff
@@ -70,6 +70,53 @@
...
+                    env('DB_HOST_W5', ''),
+                    env('DB_HOST_W6', ''),
+                    env('DB_HOST_W7', ''),
+                    env('DB_HOST_W8', ''),
+                    env('DB_HOST_W9', ''),
+               ],
+            ],
+            'sticky' => env('DB_STICKY', null),
+            'driver' => 'mysql',
+            'host' => env('DB_HOST', 'localhost'),
```
murrant, line 102: is this line needed?  Or some LibreNMS code depending on it by mistake?

doc/Extensions/Galera-Cluster.md
```diff
@@ -0,0 +1,154 @@
...
+
+This is currently being tested, use at your own risk.
+
+LibreNMS can be used with a MariaDB Galera Cluster. This is a Multi Master cluster, meaning each
+node in the cluster can read and write to the database. They all have the same ability. LibreNMS will
+randomly choose a working node to read and write requests to. 
+
+
+For more information see
+<https://laravel.com/docs/8.x/database#read-and-write-connections>
```
murrant, line 14: You can use this link to avoid specific version link:
`https://laravel.com/docs/database#read-and-write-connections`

murrant:
Need to add it to the mkdocs.yml file.
You did not mention sticky in the docs.
It also doesn't introduce the concept that you can set different servers on different poller nodes, to keep a poller node talking to the closest mysql node(s).  Not sure if that is worthwhile to mention or just adds confustion.

config/database.php
djamp42: DB_Host is needed because things like lnms break without it, might be couple of other things aswell.

doc/Extensions/Galera-Cluster.md
djamp42: Fixed

djamp42:
Yup, added some updates regarding this.

murrant (approved):

PR #12964, opened 2021-06-18
API mac search

mpikzink:
A useful API search if you are looking for a client that is not included in LNMS. This can be used to locate the connected switchport based on the client MAC.

#### Please note

#### Testers


includes/html/api_functions.inc.php
```diff
@@ -2586,6 +2586,16 @@ function del_service_from_host(Illuminate\Http\Request $request)
     return api_error(500, 'Failed to delete the service');
 }
 
+function search_by_mac(Illuminate\Http\Request $request)
+{
+    $search = $request->route('search');
+    $ports = dbFetchRows('select ports.* from ports_fdb join ports on ports.port_id=ports_fdb.port_id join devices on devices.device_id=ports.device_id where ports_fdb.port_id in (select port_id from ports_fdb where mac_address=?) group by ports_fdb.port_id having count(ports_fdb.port_id)>0 order by count(ports_fdb.port_id) limit 1', $search);
```
Jellyfrog, line 2592: Would you mind writing this in eloquent syntax instead? It's more future proof. 
Not sure why you need to mix in device table either

includes/html/api_functions.inc.php
```diff
@@ -2598,6 +2599,23 @@ function del_service_from_host(Illuminate\Http\Request $request)
     return api_error(500, 'Failed to delete the service');
 }
 
+function search_by_mac(Illuminate\Http\Request $request)
+{
+    $search = $request->route('search');
+    $portlist = PortsFdb::select('port_id')
```
Jellyfrog, line 2605: I think this should return all ports matching the Mac Instead?
Something like
```
Port::whereHas('fdbEntries', function($q) {
   $q->where('mac_address', $search);
})->get();
```

mpikzink: I just want to get the Downlink.
e.g.
```
PortsFdb::whereIn('port_id',PortsFdb::where('mac_address', $search)->select('port_id'))->groupBy('port_id')->orderByRaw('count(port_id)')->select('port_id')->first()->port;
```

murrant: @mpikzink your code has two queries, but the inside on isn't actually executed. Also, it looks like you want to select the port, so perhaps you should start with Port:: for the query.

mpikzink: @murrat Yes I need a port. But I am looking for a port that "only" has this MAC in the FDB. Traditionally as SQL query I can only achieve this via a subquery so that I can count the remaining MAC addresses on the port in the FDB.

murrant: Remember that a MAC can be on many ports (switch downlinks).

Subquery is supported, but that is not the correct syntax

mpikzink: I know that it can be on many ports but I think if I search for a client the client will be on the port with the fewest count of other MACs. I have this (first commit) option in our setup since a year and it works very well for our client search which we use multiple times per day. I will have a look for the right syntax for laravel. 

BTW: Since my first commit I think my chosen route is also not the best place and I need to describe in the doc as well.

Jellyfrog:
I think for this endpoint we except all ports that match to be returned.
Add a filter option if you want more granular results.
Also, there should already be some function to normalize the Mac address, so users can search with different formats

LibreNMS/Util/Rewrite.php
```diff
@@ -135,6 +135,17 @@ public static function shortenIfName($name)
         return str_ireplace(array_keys($rewrite_shortif), array_values($rewrite_shortif), $name);
     }
 
+    /**
+     * Remove all allowed delmiters and convert to lower
+     *
+     * @param string $mac
+     * @return string
+     */
+    public static function normalizeMac($mac)
```
murrant, line 144: Indeed as Jellyfrog has mentioned: Rewrite::macToHex()  is the same as this, but more robust (handles macs omitting leading zeros)

mpikzink: Ups, I haven't seen it. It's much more better then my method. I will convert it to Rewrite::macToHex()

murrant (approved):
Yay, unit tests :)

PR #12968, opened 2021-06-20
Poll current connections for F5 ltm

martinberg:
Adds polling for "current connections" for LTM VS:es and LTM Pools.
Unfortunately this adds a new DS to a already existing RRD-file and therefore the RRD-file needs to be recreated.

bild

#### Please note

#### Testers

### DIFF
--- a/includes/html/graphs/device/bigip_ltm_allpm_currconns.inc.php
+++ b/includes/html/graphs/device/bigip_ltm_allpm_currconns.inc.php
@@ -2,7 +2,7 @@
 /*
  * LibreNMS module to display F5 LTM Virtual Server Details
  *
- * Copyright (c) 2016 Aaron Daniels <aaron@daniels.id.au>
+ * Copyright (c) 2021 Martin Bergström <martin@bergstr0m.se>
  *
  * This program is free software: you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by the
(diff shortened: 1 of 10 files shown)


PipoCanaja:
Hi @martinberg 
Any chance you could create a new RRD file instead of extending an existing one ?

martinberg:
> 
> 
> Hi @martinberg
> Any chance you could create a new RRD file instead of extending an existing one ?

I'm unsure how to accomplish that in a way that would make sense in case somebody wants to add more stats in the future. 
I feel that `includes/polling/loadbalancers/f5-ltm.inc.php` is the right place to loop through all objects at least but don't know how to implement it.
Do you have any suggestions or are able to put together something?

murrant:
Definitely make a new RRD.  If I could redo anything in LibreNMS, I would only put 1 DS per rrd... that way you can and and remove values without breaking everything.

martinberg:
Made a separate RRD. Not sure if this is a good way to do it but this is what I came up with.
Let me know if there is a better way :)

includes/html/graphs/device/bigip_ltm_allpm_currconns.inc.php
```diff
@@ -0,0 +1,74 @@
+<?php
+/*
+ * LibreNMS module to display F5 LTM Virtual Server Details
+ *
+ * Copyright (c) 2016 Aaron Daniels <aaron@daniels.id.au>
```
PipoCanaja, line 5: Hi
Could you please correct the copyright on files ?
Thanx

PipoCanaja (changes requested):

includes/html/graphs/device/bigip_ltm_allpm_currconns.inc.php
martinberg: Thanks! Forgot about that part

PipoCanaja (approved):
LGTM

PR #12978, opened 2021-06-22
return api error when device doesn't exist

VirTechSystems:
Function device_by_id_cache returns a hash table, which isn't null despite the device not existing.  This checks to see if the type has been populated to also determine if the device exists. 

Without this, the function always returns a "valid" device, even if one doesn't exist.

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -266,7 +266,7 @@ function get_device(Illuminate\Http\Request $request)
 
     // find device matching the id
     $device = device_by_id_cache($device_id);
-    if (! $device) {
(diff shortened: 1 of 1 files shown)


includes/html/api_functions.inc.php
```diff
@@ -266,7 +266,7 @@ function get_device(Illuminate\Http\Request $request)
 
     // find device matching the id
     $device = device_by_id_cache($device_id);
-    if (! $device) {
+    if (! $device || ! $device['type']) {
```
Jellyfrog, line 269: Device_id would be the correct field to check

Jellyfrog (approved):

PR #12989, opened 2021-06-25
Help users that did not upgrade MySQL try two

murrant:
Instead of dropping the table and re-creating.  Just try adding the index.  This way it won't fail if they somehow did the port_group_port migration first.  And still should fail if the index cannot be created.

#### Please note

#### Testers

### DIFF
--- a/database/migrations/2020_12_14_091314_create_port_groups_table.php
+++ b/database/migrations/2020_12_14_091314_create_port_groups_table.php
@@ -12,16 +12,24 @@ class CreatePortGroupsTable extends Migration
      */
     public function up()
     {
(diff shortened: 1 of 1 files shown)


Jellyfrog:
Problem is this will just keep happening with the next migration, we should block even running the migrations command instead

murrant:
No, it will keep hanging at this migration until they upgrade MySQL.  This will try to add the index and it will fail. (I have not tested that)

murrant:
We really can't block migrations as it will cause the ui to be unreachable if they are out of sync.  We could have blocked updates entirely, but that needed to be done before this migration was committed.  And so we are here.

PR #12992, opened 2021-06-30
Eaton Network MS xups sensors

dagbdagb:
This patch does two things:
(0. Add the XUPS mib to discovery definitions)
1. Add the ambient temperature sensor exposed in the the xups mib, rename the current xups temp sensor
2. Flag alerts raised in the xups mib. I choose to put this with the other (mgups) alerts under 'State Sensors'.

The reason for (2) is that I have devices where an alert *is* raised with the xups mib, but not in the mgups mib. 

The reasons for not exposing more details about the xups alerts in LibreNMS is that the onboard web UI for these devices can be kinda vague about what the alert is about ("contact Eaton for a battery check"), and I am not certain I fully understand the xups mib.

Hence I just flag an alert and the user will have to check the actual device and/or support provider.

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/discovery/eaton-mgeups.yaml
+++ b/includes/definitions/discovery/eaton-mgeups.yaml
@@ -189,27 +189,7 @@ modules:
                     states:
                         - { descr: 'No Alarm', graph: 0, value: 0, generic: 0 }
                         - { descr: 'Alarm raised', graph: 0, value: 1, generic: 2 }
-                -
-                    oid: xupsAlarmID
-                    value: xupsAlarmID
-                    num_oid: '.1.3.6.1.4.1.534.1.7.2.1.1.{{ $index }}'
-                    descr: 'XUPS Alarm Description'
-                    index: 'xupsAlarmID.{{ $index }}'
-                    state_name: xupsAlarmID
-                    states:
-                        - { descr: 'On Battery', graph: 0, value: 3, generic: 2 }
-                        - { descr: 'Low Battery', graph: 0, value: 4, generic: 2 }
-                        - { descr: 'Utility Power Restored', graph: 0, value: 5, generic: 2 }
-                        - { descr: 'Return From Low Battery', graph: 0, value: 6, generic: 2 }
-                        - { descr: 'Output Overload', graph: 0, value: 7 , generic: 2 }
-                        - { descr: 'Internal Failure', graph: 0, value: 8, generic: 2 }
-                        - { descr: 'Battery Discharged', graph: 0, value: 9, generic: 2 }
-                        - { descr: 'Inverter Failure', graph: 0, value: 10, generic: 2 }
-                        - { descr: 'On Bypass', graph: 0, value: 11, generic: 2 }
-                        - { descr: 'Bypass Not Available', graph: 0, value: 12, generic: 2 }
-                        - { descr: 'Output Off', graph: 0, value: 13, generic: 2 }
-                        - { descr: 'Input Failure', graph: 0, value: 14, generic: 2 }
-                        - { descr: 'Building Alarm', graph: 0, value: 15, generic: 2 }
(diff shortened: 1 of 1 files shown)


Jellyfrog:
Hi,

You need to update the test data accordingly

includes/definitions/discovery/eaton-mgeups.yaml
```diff
@@ -173,4 +179,17 @@ modules:
...
+                    num_oid: '.1.3.6.1.4.1.534.1.7.1.{{ $index }}'
+                    descr: 'XUPS Alarm'
+                    index: 'xupsAlarms.{{ $index }}'
+                    state_name: xupsAlarms
+                    states:
+                        - { descr: 'No Alarm', graph: 0, value: 0, generic: 0 }
+                        - { descr: 'Alarm raised', graph: 0, value: 1, generic: 2 }
+                        - { descr: 'Alarms raised', graph: 0, value: 2, generic: 2 }
+                        - { descr: 'Alarms raised', graph: 0, value: 3, generic: 2 }
+                        - { descr: 'Alarms raised', graph: 0, value: 4, generic: 2 }
```
mpikzink, line 194: why not this way?
```
         count: 
                -
                    oid: xupsAlarms
                    value: xupsAlarms
                    num_oid: '.1.3.6.1.4.1.534.1.7.1.{{ $index }}'
                    descr: 'XUPS Alarm'
                    index: 'xupsAlarms.{{ $index }}'
                    high_limit: 1
```
I personally would think about warn_limit 1 and high_limit 2 because if you have an alarm you should get a yellow warning, in the most of the times. If this grows up you have bigger problems so you should get a red state but this depends on the hardware :-). A better way would be to walk thought the underlying table in the MIB and print the details.

dagbdagb: @mpikzink Thank you for the feedback. I believe the number of alarms from your UPS is a dubious measure for the urgency implied.  :-)    

And yes, parsing the MIB and printing more details would be more informative. But a)I do not entirely trust the reported details, seeing how these alarms are reported in the web ui of the device. And b) I did not trust my understanding of the mechanics involved.  The mib speaks of alarmID and alarmEventID and one is likely the unique alarm event and the other the type of alarm. And all in all, I just want LibreNMS to flag that there is *something* to look into.

Grouping the xups alarms with the other alarm-related sensors made sense to me. Hence the some what silly 'states' stanza.

dagbdagb:
Added test data, but unsure how to make the unit tests use *my* test data. I find it highly likely that I am doing something stupid, but sometimes the obvious path is just invisible. What do?!

PipoCanaja:
> Added test data, but unsure how to make the unit tests use _my_ test data. I find it highly likely that I am doing something stupid, but sometimes the obvious path is just invisible. What do?!

You need to re-run the save-test-data on the failing files (look at the test results, eaton-mgeups_5px.json is the 1st one you need to re-compute)

This is because your patch changes how the existing tests results. So you need to re-run them (so the tests will then succeed) and the reviewer will have a chance to see the changes (and evaluate if they are safe or not for LibreNMS community)

includes/definitions/discovery/eaton-mgeups.yaml
PipoCanaja: Hi @dagbdagb 
The issue I see here is if you have, let say, more than 5 alarms ... State conversion will fail. So I agree with @mpikzink , you need to use count, set a high limit at 1 (so anything except 0 will trigger an alert) and you'll end up with exactly the same information, without the risk of failure with the states.

PipoCanaja (changes requested):

dagbdagb:
I was sort of expecting that feedback.   :-)
I think that if you have more than 4 alarms, you are bound to notice anyway. Unsure if 4 alarms even can occur simultaneously.
But OK, I'll push a change.

PR #12993, opened 2021-06-30
Bintec be.IP plus support

jul13579:
This MR adds support for the Bintec be.IP plus router. It adds the necessary MIB and discovery configuration files. Additionally, since the device reports OIDs out of order, this MR adds the boolean `out_of_order_oids` attribute for use in OS configuration files.

#### Please note

#### Testers


jul13579:
Sorry for force-push, but I missed setting the correct commit author.

mibs/bintec/bintec.mib
```diff
@@ -0,0 +1,118 @@
+-- This file is corresponding to Release 10.2.3.100 from 2018/07/18 00:00:00
```
Jellyfrog, line 1: Please rename this fle to `BINTEC-MIB`without .mib

mibs/bintec/mib-acct.mib
```diff
@@ -0,0 +1,204 @@
+-- This file is corresponding to Release 10.2.3.100 from 2018/07/18 00:00:00
```
Jellyfrog, line 1: Rename to `BINTEC-ACCTSTORE`

mibs/bintec/mib-address.mib
```diff
@@ -0,0 +1,412 @@
...
+
+
+---------------------------------------------------------------------------
+-- $RCSfile: mib-address,v $
+-- $Revision: 1.4 $
+-- $Date: 2013-08-12 11:44:47 $
+-- Author: RFC 3291
+---------------------------------------------------------------------------
+
+INET-ADDRESS-MIB DEFINITIONS ::= BEGIN
```
Jellyfrog, line 11: And this one, to `INET-ADDRESS-MIB`, and in fact, all other files :) just name them to whats before `DEFINITIONS`

jul13579: Hi! Thanks for reviewing :blush:  I renamed all files according to your suggestion.

PipoCanaja:
Hi @jul13579 
Still on the MIB side, please ensure that you remove all the standard MIB files from the bintec directory. They are already installed in ~/mibs directory and will be loaded.
Thanx

jul13579:
@PipoCanaja Removed all standard MIBs i could find and updated the PR :wink: 

Could anyone give me a hint why the unittest (8, Web, 1, mariadb:10.5) was failing? It was failing on https://github.com/librenms/librenms/blob/131f5c7e87e8f859492509f58a51ceb7568fa861/LibreNMS/Device/Processor.php#L237 expecting a string but getting an array for `$data`. I don't really understand why this test only failed on mariaDB, but not on mysql?

includes/definitions/discovery/bintec-beip-plus.yaml
```diff
@@ -0,0 +1,21 @@
+# Some values are hardcoded because snmpbulkwalk is not supported
+modules:
+    os:
+        version: BIANCA-BRICK-MIB::biboAdmSWVersion.0
+        serial: BIANCA-BRICK-MIB::biboABrdSerialNo.0.0
+        hardware: BIANCA-BRICK-MIBRES-MIB::cpuDescr.1.0
```
Jellyfrog, line 6: not sure if there is a better one, but this doesn't really seems to be a hardware model

includes/definitions/discovery/bintec-beip-plus.yaml
```diff
@@ -0,0 +1,21 @@
+# Some values are hardcoded because snmpbulkwalk is not supported
+modules:
+    os:
+        version: BIANCA-BRICK-MIB::biboAdmSWVersion.0
```
Jellyfrog, line 4: Maybe you could split this oid into version and features like so:
```suggestion
        version: BIANCA-BRICK-MIB::biboAdmSWVersion.0
        version_regex: '/^(?<version>[^ ]+) (?<features>.*)/'
```

jul13579: Good call! I added the proper regex that works with the reported string (e.g. `Bintec Be.IP Plus V.10.2.9.102 IPv6, IPSec, PBX from 2021/03/16 00:00:00`)

includes/definitions/discovery/bintec-beip-plus.yaml
jul13579: Unfortunately I was no able to find a better string representing the hardware version. I also cannot see any hardware version information being logged into the device or even using its integrated SNMP browser. With the above config, it reads "Onboard MIPS 24Kec CPU". Would you rather remove this information than having this string displayed?

Jellyfrog: Up to you if it makes sense or not!

jul13579:
With the latest changes I improved device detection and was able to get all tests running (see my master branch) :wink:

includes/definitions/discovery/bintec-beip-plus.yaml
```diff
@@ -0,0 +1,26 @@
...
+        serial: BIANCA-BRICK-MIB::biboABrdSerialNo.0.0.0
+        hardware: 
+            - BIANCA-BRICK-MIB::biboABrdPartNo.0.0.0
+            - BIANCA-BRICK-MIB::biboABrdHWRelease.0.0.0
+        hardware_template: '{{ BIANCA-BRICK-MIB::biboABrdPartNo.0.0.0 }} {{ BIANCA-BRICK-MIB::biboABrdHWRelease.0.0.0 }}'
+    mempools:
+        data:
+            -
+#                total: BIANCA-BRICK-MIBRES-MIB::memoryTotal.2.1
+#                used: BIANCA-BRICK-MIBRES-MIB::memoryInuse.2.1
```
PipoCanaja, line 15: Any reason you used the numOID here ? Textual form would be easier to read I think.

PipoCanaja (changes requested):

PipoCanaja:
Nice work @jul13579 👍  
Almost good to go.

includes/definitions/discovery/bintec-beip-plus.yaml
jul13579: Kind of, yes :sweat_smile: At that time, I could not get the commented version using the textual representation to work. I did some debugging and found the reason and also was able to work around it:

https://github.com/librenms/librenms/blob/ae2b761e940b0cfc0b66badc6b6ada99f2e16f5f/LibreNMS/OS/Traits/YamlMempoolsDiscovery.php#L93

Apparently, when using the textual representation with a static index appended, the above linked method is not capable of getting the data from the array of received SNMP data. The method e.g. receives
```php
$field = "used" // To get the amount of used memory
$index = "2.1" // The correct index I was statically appending
$yaml = array (
  "total" => "BIANCA-BRICK-MIBRES-MIB::memoryTotal.2.1"
  "used" => "BIANCA-BRICK-MIBRES-MIB::memoryInuse.2.1"
);
```
Now, in the next line:
https://github.com/librenms/librenms/blob/ae2b761e940b0cfc0b66badc6b6ada99f2e16f5f/LibreNMS/OS/Traits/YamlMempoolsDiscovery.php#L95
the requested OID is fetched from the array. After that, the OID is used to fetch the requested data from the `$this->mempools` array:
https://github.com/librenms/librenms/blob/ae2b761e940b0cfc0b66badc6b6ada99f2e16f5f/LibreNMS/OS/Traits/YamlMempoolsDiscovery.php#L96
But since the fetched OID already has the index statically appended, the above lookup fails. BTW, the `$this->mempools` array looks like this:
```php
array(
"2.1" => array (
    "BIANCA-BRICK-MIBRES-MIB::memoryTotal" => someVal,
    ...
  )
);
```
which IMO correctly represents the data received by walking over the SNMP table. The `getData` function just is not capable to work with OIDs in textual form while also appending a static index.

Since I could not find documentation that states such configuration is supported/allowed, I don't consider this a bug. I was just expecting this to work, as I could use such OIDs in different places of the YAML config without issues. Maybe you are thinking differently about it? At the end, I could work around this by configuring it to walk over the entire `memoryTable` and skipping all entries that were not of interest.

jul13579:
@PipoCanaja is there any more work needed for this PR?

murrant:
I'm not sure allowing out-of-order oids on all snmp queries is desired.  Probably should be a case-by-case basis.

jul13579:
@murrant What exactly do you mean with "case-by-case" basis? Being able to set the flag for different discovery pools/sections case-by-case? What exactly is the downside of setting the flag on a device-basis?

includes/definitions/bintec-beip-plus.yaml
```diff
@@ -0,0 +1,14 @@
+os: bintec-beip-plus
+nobulk: true
```
Jellyfrog, line 2: This is now snmp_bulk. You should rebase this pr

Jellyfrog:
Please try the  only include actually used mibs

murrant:
> @murrant What exactly do you mean with "case-by-case" basis? Being able to set the flag for different discovery pools/sections case-by-case? What exactly is the downside of setting the flag on a device-basis?

Receiving garbage data and inserting it into the database/rrd is the danger.  And I mean setting `-Cc` on only the problematic snmp query.

includes/definitions/bintec-beip-plus.yaml
jul13579: Done 👍

includes/definitions/bintec-beip-plus.yaml
```diff
@@ -0,0 +1,14 @@
...
+out_of_order_oids: true
```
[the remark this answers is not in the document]
jul13579: Done 👍

jul13579:
> Please try the only include actually used mibs

@Jellyfrog I updated the PR to include your proposed changes 😉

tests/data/bintec-beip-plus.json
```diff
@@ -37,8 +37,6 @@
                     "ifSpeed_prev": null,
                     "ifConnectorPresent": null,
                     "ifPromiscuousMode": null,
-                    "ifHighSpeed": null,
```
jul13579: For some reason, I had to remove all lines including `ifHighSpeed*` from test data, else the tests would fail (it would expect these lines not to be in the test file). Even after re-creating the test and snmprec files from scratch, these props would still be there and lead to failing tests. I searched for changes regarding these props in upstream, but was not yet able to understand the changes. If removing these props from test data is an issue, please let me know how to deal with this.

murrant:
@jul13579 sorry this has sat for awhile.  After re-reviewing.  I still think you should just set -Cc on the specific queries using snmp_flags.  Enabling -Cc on an entire OS is just too risky.

jul13579:
@murrant Thanks for your review, it is highly appreciated. I looked into the `snmp_flags` since I was not aware of it existing (it somehow also is only documented in the *Memory* section of the *CPU/Memory Information* docs. I checked where discovery with disabled global `-Cc` fails and tried to set the `snmp_flags` prop accordingly. The modules the `-Cc` is required are:
* ports
* arp-table

During discovery I get multiple `OID not increasing` errors per module. Hence I added the sections `ports` and `arp-table` to my discovery YAML, and added the `snmp_flags` prop to either of them. However this did not help, as the setting only applies in `YamlDiscovery.php` if the `data` array is also populated with OIDs.

Now, since I couldn't find any information about configuring the `ports` and `arp-table` modules in the docs and any other discovery file, I'm wondering how I am supposed to configure this correctly. Can you give me any more information on how to deal with this?

PR #12994, opened 2021-06-30
Eaton M2 EMP g2

dagbdagb:
The Eaton M2 NMC for current generation UPSes has a companion Environmental Monitoring Probe gen 2. (temp, humidity, 2x contacts). This adds minimal support for said probe. I choose to use the (simpler) xups mib over the much more detailed eaton-sensor-mib.

This PR also has an updated snmprec for a device with this probe, as well as with an updated M2 firmware version (2.0.5)

#### Please note

#### Testers


Jellyfrog:
Hi,

Please see failed tests

dagbdagb:
The failed tests reference stuff I haven't touched. 
Additionally, I have added new json- and snmprec-files, as variants. (New major firmware release + new external sensor. Unclear if these two points warrants a variant.) 

One thing I observe after my additions is that my new sensors have weird types under 'Health settings'. The two contacts are type xupsEnvironment, while temp/hum are of type eatonupsm2. Not sure if this implies a bug or a need for adjustments.

Would very much appreciate some handholding here. As time and energy permits, of course.

PipoCanaja:
> The failed tests reference stuff I haven't touched.
As soon as you touch an existing YAML OS file, you will impact any device that uses it. This is why we have regression tests. If the resulting change is acceptable, then the regression tests will be changed. If they are not acceptable, then you'll need to adapt your patch. 
So please run save-test-data on every failing regression test file (as seen in the test results).

> Additionally, I have added new json- and snmprec-files, as variants. (New major firmware release + new external sensor. Unclear if these two points warrants a variant.)
This is good practice (and required) to add new test data for every change, to ensure regression tests will also check your new addition.

includes/definitions/discovery/eatonupsm2.yaml
```diff
@@ -13,8 +13,26 @@ modules:
             data:
                 -
                     oid: xupsEnvironment
-                    value: xupsEnvAmbientTemp
```
PipoCanaja, line 16: Hi @dagbdagb 
Why did you remove this sensor. Are you sure it cannot be of any use for anybody ? Removing it here will remove it for each and every LibreNMS user after this PR is merged. In doubt, I would keep it, and also add you xupsEnvRemoteTemp,  so you end up with both.

PipoCanaja (changes requested):

includes/definitions/discovery/eatonupsm2.yaml
dagbdagb: > Hi @dagbdagb
> Why did you remove this sensor. Are you sure it cannot be of any use for anybody ? Removing it here will remove it for each and every LibreNMS user after this PR is merged. In doubt, I would keep it, and also add you xupsEnvRemoteTemp, so you end up with both.

You are right of course. The current major release does not expose the onboard temperature sensor. But the previous release did, so we need to keep this. Improved version coming up. 

Thank you for reviewing.

PipoCanaja (approved):
LGTM. Thanx for your contribution

PR #12998, opened 2021-07-02
New plugin system based on Laravel Package Development

mpikzink:
We want plugins to be full packages install-able with composer.  We want translation and all that: https://laravel.com/docs/8.x/packages

Todos
- [x] Backward compatibility
- [x] Convert to blade
- [ ] Change the documentation
- [x] Add the authentication system
- [x] Support local plugins for ease of use.  Local plugins will not support publishing routes, translation, etc like you can with a full package. 

#### Please note

#### Testers


PipoCanaja:
Hello @mpikzink 
I am not an expert at all, but for readability of code and isolation of Plugins, it would be great to have the views in the same Plugins tree. Target is that a plugin should be selfcontained.

mpikzink:
Hi @PipoCanaja thanks for the feedback. This is also my first conversion so I think I have some tweaking to do. I will play around ;-)

mpikzink:
In the Laravel documentation I can only find views stored in /resources/Views. According to that this is also the default. I understand @PipoCanaja recommendation and also like to encapsulate everything in a directory, but this contradicts the Laravel recommendation (at least as I understand it). Does anyone else have a recommendation which way we should go for the future?

murrant:
Yeah, the "plugin" design for LibreNMS is not a good thing.

mpikzink:
@murrant What do you mean? Should I go another way, or should I leave it as it is? I have tried to redo it, but so that the old.way still works because we do not know how many users have written a plugin themselves.

murrant:
No just random lamentations.  I did not look at your code, but I'm not sure if it is the correct place to focus effort.

If you can make it work and ensure existing plugins don't break I'm all for it :)

html/plugins/Test/Test.php
```diff
@@ -2,26 +2,38 @@
...
-        echo 'Example display in Port tab</br>';
-        echo '</div></div></div></div>';
+        $name = self::className() . ' Plugin';
+        $port = Port::find($port['port_id']);
+        echo view(self::viewPath(), compact('name', 'port'));
+    }
+
+    private static function viewPath()
+    {
+        return 'plugins.' . strtolower(self::className()) . '.' . debug_backtrace()[1]['function'];
```
PipoCanaja, line 32: It seems that each plugin can add its own view path, that could be done in a class constructor for instance.  
```
$path = base_path("html/plugins" . self::classname . "/ressources/views");
 \View::addLocation($path");
```

Goal is to keep the fact that a plugin is self_contained.

Of course, this view loading code should be in ~/LibreNMS/Plugins.php class, and expose the functions to the plugins themselves. That would ensure that no view collision occurs.

PipoCanaja (changes requested):

PipoCanaja:
(and yes, I am a plugin user, and have a huge dependency on them. The support team here is using a custom made plugin to do configuration and checks on network devices, changing VLANs etc etc).

html/plugins/Test/Test.php
mpikzink: @PipoCanaja thanks for help. I used a abstract class for Plugin.php can you have a look that you think about it?

PipoCanaja: Will do my best to take the required amount of time ASAP to review this. A 10 seconds look at the files seems promising ! :)

mpikzink: No rush! Whenever you find the time. For me it's never that urgent :-)

LibreNMS/Plugins/Plugin.php
```diff
@@ -0,0 +1,35 @@
...
+
+    final protected static function className()
+    {
+        return str_replace(__NAMESPACE__ . '\\', '', get_called_class());
+    }
+
+    public function menu()
+    {
+        $name = self::className();
+        echo view(self::viewPath(), compact('name'));
```
murrant, line 27: instead of using compact, just make an array:
```php
echo view(self::viewPath(), ['name' => self::className()]);
```

(and other places too)

murrant:
You are basically shimming the plugin system... I suggest you implement new functions such as:
protected $settings_view = 'blah/settings.blade.php';
settingsData(): array

You can probably come up with something better.  Put the echo part in the Plugin shim, requiring plugins to echo inside a function is terrible.

mpikzink (commented):
> You are basically shimming the plugin system... I suggest you implement new functions such as:
> protected $settings_view = 'blah/settings.blade.php';
> settingsData(): array
> 
> You can probably come up with something better. Put the echo part in the Plugin shim, requiring plugins to echo inside a function is terrible.

Sorry, I don't understand what you mean. I understand not to echo the view but I don't know how to do it in a different way.

LibreNMS/Plugins/Plugin.php
murrant: The compact() function is extremely unintuitive and people generally don't know what it is doing.  I try to avoid it in general.

mpikzink: ok, after i changed it i totally agree with you ;-)

murrant:
Put a little code in the abstract class that implements the functions for the plugin system and declare them as final(optionally).

I also changed the base view location to prevent collisions.

```php
<?php

namespace LibreNMS\Plugins;

use View;

abstract class Plugin
{
    protected $menu_view = 'menu.blade.php';
    protected $settings_view= 'settings.blade.php';
    private $prefix;
    
    public function __construct()
    {
        View::addLocation(base_path('html/plugins'));
        $this->prefix =  $this->className() . '/resources/views/';
    }

    private function className()
    {
        return str_replace(__NAMESPACE__ . '\\', '', get_class());
    }

    final public function menu()
    {
        echo view($this->prefix . $this->menu_view, $this->menuData());
    }

    final public function settings()
    {
        echo view($this->prefix . $this->settings_view, $this->settingsData());
    }
    
    public function menuData(): array
    {
        return [];
    }
    
    public function settingsData(): array
    {
        return [];
    }
}
```

murrant:
Additionally, we should allow plugins to publish a page or pages...  Currently they reside in the public html directory which means we cannot enforce authentication.

Additionally, some have wished to be able to plug in to other parts of LibreNMS such as the ports page or the device overview...  but we don't have to do anything with that.

PipoCanaja:
Hi @mpikzink 
Following this PR with a lot of attention. Looks very good so far. This PHP is over my PHP knowledge for sure, and I need help to understand the traits. How can a plugin inherit DeviceTrait for instance ? Same with the PortTrait ? That would be interesting to have them both in the Example plugin.
Thx

mpikzink:
My idea is that you only enable the hooks you want to use in your own plugin. The traits represent a kind of default, which probably need to be changed in the rarest cases.
You can easily include the traits by commenting out these lines (in Example.php)
    //use DeviceTrait;
    //use PortTrait;

I think settings and menu should be in every Plugin but this we can discuss.

My first tests look promising but I definitely need to test further with my legacy code. Unfortunately it gets a bit confusing in the include part because of the backward compatibility and I don't know yet how to convert admin.inc.php to Laravel. I'll have to drink some more coffee for that. If someone can help here I would be happy about any input.

LibreNMS/Plugins/DeviceTrait.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace LibreNMS\Plugins;
+
+use App\Models\Device;
+
+trait DeviceTrait
+{
+    protected static $device_view = 'device_overview';
```
PipoCanaja, line 9: Hi @mpikzink 
In order to allow different views to be selected, if necessary, I would suggest to replace `$device_view` protected static variable with a function `device_view($device)`. Function, that will, by default, return the 'device_overview' string, like it is now. 
If a plugin decides to override this function, the plugin will be able to call different views, for different devices, with it's own logic. 
(in my case, I have devices that my SupportTeam need to access, and some they shouldnt. Logic is based on name of the device, location, etc etc). 

```
public static function device_view($device)
    {
        return 'menu;
    }
```

Same logic would apply for PortTrait with 2 parameters (device and port).

LibreNMS/Plugins/PortTrait.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace LibreNMS\Plugins;
+
+use App\Models\Port;
+
+trait PortTrait
+{
+    protected static $port_view = 'port';
```
PipoCanaja, line 9:
```
public static function port_view($device, $port)
    {
        return 'port';
    }
```

mpikzink:
@PipoCanaja do you have some example code for this authentication part?

LibreNMS/Plugins/DeviceHook.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace LibreNMS\Plugins;
+
+use App\Models\Device;
+
+trait DeviceHook
+{
+    public static function device_overview_container($device)
```
murrant, line 9: All methods should be camel case.

PipoCanaja: This one is the historical one for plugins. Meaning changing it will be a challenge for backward compatibility and would require users to patch existing plugins, and notification will be necessary.

murrant:
Note: to install plugin without breaking updates: `composer require --no-update some/plugin`

Jellyfrog:
(Sorry wanted to trigger new tests)

mpikzink:
I cannot find the problem. Can someone help?

database/migrations/2021_07_28_102443_plugins_add_version_and_settings.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('plugins', function (Blueprint $table) {
+            $table->integer('version')->default(1);
+            $table->json('settings')->nullable();
```
Jellyfrog, line 18: Guessing this field type could be the problem?

murrant: It is, it uses utf8mb4_bin collation, which is valid. But I think I messed up my fix for that :)

Jellyfrog:
Current phpstan error is a bug/limitation in phpstan itself. Just add a ignore comment to them.

murrant:
@mpikzink code is all straightened out... did you get a chance to test implementing an actual plugin?

mpikzink:
@murrant Yay, this looks very good. I don't see problems here.

includes/functions.php
```diff
@@ -320,14 +320,14 @@ function delete_device($id)
...
     \App\Models\Port::where('device_id', $id)
-        ->with('device')
-        ->select(['port_id', 'device_id', 'ifIndex', 'ifName', 'ifAlias', 'ifDescr'])
-        ->chunk(100, function ($ports) use (&$ret) {
-            foreach ($ports as $port) {
-                $port->delete();
-                $ret .= "Removed interface $port->port_id (" . $port->getLabel() . ")\n";
-            }
-        });
+                    ->with('device')
```
Jellyfrog, line 323: Unrelated change?

mpikzink:
@murrant multiple plugins in v2 are not possible in the moment.

includes/functions.php
murrant: looks like indentation change

murrant:
@mpikzink please elaborate.

mpikzink:
Screenshot_2021-08-23_12-45-50
@murrant  It shows the menu view of the first Plugin multiple times.

mpikzink:
Multiple Plugins v2 works now thanks @murrant.

mpikzink:
> Heh. Phpstan strikes again :)

What to do?

PR #13010, opened 2021-07-07
Fixing irrelevant values rpm packetloss

geg347:

#### Please note

#### Testers

Hello again,

After weeks of use, I saw that the values of Juniper RPM packetloss was strangely too high. After debugging I found that the problem comes from the DISMAN Ping MIB values. The values currently queried are the current probe values and if a probe is still pending, we got irrelevant pingResultsProbeResponses and pingResultsSentProbes values so the calculated percentage is wrong.
So to fix this I changed the MIB to Juniper's one (jnxPingLastTestResultTable). And now only the completed RPM probe are queried and the graphs correspond to reality.
I also changed the RTT to the average RTT of the last completed probe.

The same behaviour could occur with Huawei NQA depending on the implementation.

(I also fixed a typo in Cisco pollerSla part)

Thanks for checking and have a nice end of day!


PipoCanaja:
Hi @geg347 
Thanx for the PR. You need to replace the tests files (cause you are now checking OIDs that are not part of the current test files). 
You can check the test results after this replacement is done. 
I'll check the Huawei part if the same issue is found, and adapt it if necessary.
Bye

PipoCanaja:
@geg347 
Removing snmprec data is not the way to go :) 

- For any test variant existing, you should re-run save-test-data on it, so JSON is computed again with your code.
- For at least one variant (the juniper device you own and on which you tested this current patch) you collect new snmp values (or update existing, but this can be messy) and run save-test-data on those new values. 

Thanx

PipoCanaja (changes requested):
still need test data.

geg347:
OK @PipoCanaja sorry I thought I needed to regenerate both test data and snmprec.
Is it OK like this, please?

PipoCanaja:
Your snmprec (the one that was added in the previous PR) must be updated to comply with current code. Andd then you need to regenerate all JSON that are failing (the one from previous PR as well as all others that fail)

PipoCanaja (approved):
LGTM

geg347:
OK sorry I am not so familiar with unit tests.
So following the procedure on this page : https://docs.librenms.org/Developing/os/Test-Units/,
I need to rerun `./scripts/collect-snmp-data.php -h <myhostname> -m slas -v vmx`?
And then rerun `./scripts/save-test-data.php -o junos -m slas -v vmx`?
Please.
Not sure so I am asking before submitting a 3rd attempt...

PR #13011, opened 2021-07-07
Fixed missing quotes on edit device

murdick:
**Problem:**
Form field on edit device is not wrapped in quotes. 

**Description:**
We are using a combination of spaces in the Hostname and Overwrite IP in order to have friendlier looking device names in our lists, as recommended by PipoCanaja on the forums.

We noticed that while the form accepts and stores device names with spaces, when you return to edit them, only the information before the first space is visible.

**Resolution:**
After applying this change, we are able to edit device names that have spaces in them.

libre_fixquotes

**Reference:**
https://community.librenms.org/t/device-feature-override-sysname/7757/5

#### Please note

#### Testers


murdick:
Actually it looks like a lot of things break with spaces in the names. I'll circle back to this when there is more time.

murrant:
Yeah, hostname needs to be a valid hostname.  LibreNMS does not support arbitrary device names as of now.

PR #13012, opened 2021-07-07
Two-factor UI config and status display

rhinoau:
This PR improves visibility of 2FA status across users, adds config UI and defaults, and updates doco.

Inspired by my observation of the admin experience and doco out of this forum post: https://community.librenms.org/t/too-many-two-factor-failures/16323

Changes:
1. Set config defaults for twofactor and twofactor_lock
2. Add UI config for 2FA throttle-time and help text/descriptions
3. Add Manage Users grid 2FA column and visual indicator when twofactor is globally enabled
4. Documentation to suit and general cleanup/corrections

image
image

#### Please note

#### Testers


app/Http/Controllers/UserController.php
```diff
@@ -56,6 +56,7 @@ public function index()
         return view('user.index', [
             'users' => User::orderBy('username')->get(),
             'multiauth' => User::query()->distinct('auth_type')->count('auth_type') > 1,
+            'twofactor' => UserPref::query()->where('pref', 'twofactor')->get('user_id'),
```
Jellyfrog, line 59: should use keyBy('user_id') instead so you dont need to loop it further down

resources/views/user/index.blade.php
```diff
@@ -22,6 +22,9 @@
                     @if(\LibreNMS\Authentication\LegacyAuth::getType() == 'mysql')
                     <th data-column-id="enabled" data-formatter="enabled">@lang('Enabled')</th>
                     @endif
+                    @if(\LibreNMS\Config::get('twofactor'))
```
murrant, line 25: You can just use `@config('twofactor')`  and `@endconfig` (They are custom directives LibreNMS has)

resources/views/user/index.blade.php
```diff
@@ -38,6 +41,11 @@
                             @if(\LibreNMS\Authentication\LegacyAuth::getType() == 'mysql')
                             <td>{{ $user->enabled }}</td>
                             @endif
+                            @if(\LibreNMS\Config::get('twofactor'))
+                                @if(isset($twofactor[$user->user_id]))
+                                <td>{{ $twofactor[$user->user_id]->pref }}</td>
```
murrant, line 46: I suggest you use `UserPref::getPref($user, 'twofactor')`, but I think the code should be updated to be more efficient (as below).  You should also pre-load the preferences relationship.  As in my other comment.

```php
    public static function getPref(User $user, $pref)
    {
        if ($user->relationLoaded('preferences')) {
            return optional($user->preferences->firstWhere('pref', $pref))->value;
        }

        return $user->preferences()->where('pref', $pref)->value('value');
    }
```

app/Http/Controllers/UserController.php
```diff
@@ -56,6 +56,7 @@ public function index()
         return view('user.index', [
             'users' => User::orderBy('username')->get(),
```
murrant, line 57: Update this to:
`'users' => User::with('preferences')->orderBy('username')->get(),`

and delete the 'twofactor' line below.

resources/views/user/index.blade.php
```diff
@@ -62,6 +70,11 @@
                             return '<span class="fa fa-fw fa-close text-danger"></span>';
                         }
                     },
+                    twofactor: function (column, row) {
+                        if(row['twofactor'] == 'twofactor') {
```
murrant, line 74: This will probably need a tweak after my sql changes.

Jellyfrog (approved):

resources/views/user/index.blade.php
```diff
@@ -41,11 +41,11 @@
                             @if(\LibreNMS\Authentication\LegacyAuth::getType() == 'mysql')
                             <td>{{ $user->enabled }}</td>
                             @endif
-                            @if(\LibreNMS\Config::get('twofactor'))
-                                @if(isset($twofactor[$user->user_id]))
-                                <td>{{ $twofactor[$user->user_id]->pref }}</td>
+                            @config('twofactor')
+                                @if(app('App\Models\UserPref')->getPref($user, 'twofactor'))
```
murrant, line 45: This calls the static function non-statically.

PR #13026, opened 2021-07-11
Update cleanup config document

Sea-n:
- Add missed `units` for config definition
- Change values in two documents to the actual default value
- `syslog_purge` and `route_purge` having `order=2`, change the former to `order=1`
- Sort configs in the two documents by `order`.

#### Please note

#### Testers

### DIFF
--- a/doc/Support/Cleanup-options.md
+++ b/doc/Support/Cleanup-options.md
@@ -11,15 +11,20 @@ entries. With Cleanup Options, you can stay in control.
 
 These options rely on ```daily.sh``` running from cron as per the installation instructions.
 
-Cleanup Options are set in ```config.php```
+You can adjust the config in Global Setting -> System -> Cleanup.
+
+Cleanup Options can also be set in ```config.php```
 
 ```php
-$config['syslog_purge']                              = 30;
 $config['eventlog_purge']                            = 30;
+$config['syslog_purge']                              = 30;
+$config['route_purge']                               = 10;
+$config['alert_log_purge']                           = 365;
 $config['authlog_purge']                             = 30;
+$config['ports_fdb_purge']                           = 10;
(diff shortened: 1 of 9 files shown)


doc/Support/Cleanup-options.md
```diff
@@ -14,12 +14,15 @@ These options rely on ```daily.sh``` running from cron as per the installation i
...
 ```php
-$config['syslog_purge']                              = 30;
 $config['eventlog_purge']                            = 30;
+$config['syslog_purge']                              = 30;
+$config['route_purge']                               = 10;
+$config['alert_log_purge']                           = 365;
 $config['authlog_purge']                             = 30;
+$config['ports_fdb_purge']                           = 10;
 $config['device_perf_purge']                         = 7;
-$config['rrd_purge']                                 = 90;// Not set by default
```
murrant, line 21: Why did you change these two to remove the notes that they aren't enabled by default?

Sea-n: I think the inconsistency between the default config and the document is misleading. It's clearly the units are days, not necessary to have `90` as an example. If `90` is a suggested value, I think we need to change the default config instead of the document.

For `ports_purge`, it's set to `10` instead of `true` or `false` by default. The original value and comment are misleading.

\* Although I have traced the related code, I am not 100% sure about how it works. Please let me know if something wrong/misunderstanding.

PipoCanaja: @Sea-n You could also add that these parameters can be configured directly and graphically in the Web GUI instead of the old way in config.php.

Sea-n: @PipoCanaja Thanks for the suggestion! I will add that later.

PipoCanaja (approved):
LGTM

PR #13048, opened 2021-07-16
Use non numeric snmpwalk for nxos fan trays

e-caille:
Related to the issue described here : https://community.librenms.org/t/librenms-is-not-showing-fan-status-for-nexus-devices-as-nexus5010/12949/14

When calling `snmpwalk_cache_oid()` with the snmpwalk flag `-n`, the oid for the fan trays will looks like `.1.1.xxxx`.
So this line of the function return an empty oid :
`[$oid, $index] = explode('.', $oid, 2);`

This PR replace `snmpwalk_cache_oid_num()` with `snmpwalk_cache_oid()` so that the `-n` is not set, and the explode() works.

#### Please note

#### Testers


PipoCanaja:
Bonjour @e-caille 

This patch works well against Nexus 5548. 

If you add test data, we can merge it as it is. But it would be even better to migrate it to YAML syntax : simpler, and easier to maintain.

Could you try that ? 

Merci

PipoCanaja (approved):
LGTM, Merci pour ta contribution @e-caille

PR #13056, opened 2021-07-19
Adding API route to set bgpPeerDescr field

geg347:

#### Please note

#### Testers

Hello,

To add information about the BGP session in our alerts, we need to use the bgpPeerDescr field.
To ease the configuration of this field I created a new route in the API that update this value in database.

Maybe this feature will be usefull for others so I submitted it here.
Please just tell me if it needs some improvements or if you don't want this feature in the project.

Do not forget to clear the routes cached with "php artisan route:clear".

Bye

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -604,8 +604,8 @@ function get_bgp(Illuminate\Http\Request $request)
 
 function edit_bgp_descr(Illuminate\Http\Request $request)
 {
-    $data = json_decode($request->getContent(), true);
-    if (! $data) {
+    $bgp_descr = $request->json('bgp_descr');
+    if (! $bgp_descr) {
         return api_error(500, 'Invalid JSON data');
     }
 
@@ -616,9 +616,6 @@ function edit_bgp_descr(Illuminate\Http\Request $request)
     }
 
     $peer = \App\Models\BgpPeer::firstWhere('bgpPeer_id', $bgpPeerId);
-
-    // get description
-    $bgp_descr = $data['bgp_descr'];
 
     // update existing bgp
     if ($peer === null) {
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -619,9 +619,8 @@ function edit_bgp_descr(Illuminate\Http\Request $request)
 
     // update existing bgp
     if (count($bgps) == 0) {
-        return api_error(404, "BGP peer " . $bgpPeerId . " does not exist");
-    }
-    else if (is_array($bgps) && count($bgps) == 1) {
+        return api_error(404, 'BGP peer ' . $bgpPeerId . ' does not exist');
+    } elseif (is_array($bgps) && count($bgps) == 1) {
         $bgp = $bgps[0];
 
         $update_data = [
@@ -632,7 +631,7 @@ function edit_bgp_descr(Illuminate\Http\Request $request)
             return api_error(500, 'Failed to update existing bgp');
         }
     }
-    return api_success_noresult(200, "BGP description for peer " . $bgp['bgpPeerIdentifier'] . " on device " . $bgp['device_id'] . " updated to " . $bgp_descr . ".");
+    return api_success_noresult(200, 'BGP description for peer ' . $bgp['bgpPeerIdentifier'] . ' on device ' . $bgp['device_id'] . ' updated to ' . $bgp_descr . '.');
 }
 
 function list_cbgp(Illuminate\Http\Request $request)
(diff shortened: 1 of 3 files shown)


includes/html/api_functions.inc.php
```diff
@@ -602,6 +602,39 @@ function get_bgp(Illuminate\Http\Request $request)
...
+{
+    $data = json_decode($request->getContent(), true);
+    if (! $data) {
+        return api_error(500, 'Invalid JSON data');
+    }
+
+    $bgp = [];
+    //find existing bgp for update
+    $bgpPeerId = $request->route('id');
+    $bgps = dbFetchRows("SELECT * FROM `bgpPeers` WHERE `bgpPeer_id` = $bgpPeerId LIMIT 1");
```
PipoCanaja, line 615: Hi @geg347 
Please have a look at the get_bgp function just above. A few checks are missing. You need to validate the data (if bgpPeerId is not numeric for instance) and most importantly the dbFetchRow must be injection_proof. Which is not the case here (using the question mark syntax).

PipoCanaja (changes requested):

geg347:
Hello @PipoCanaja,
So I added the check on bgpPeerId to validate it is a integer.
I also changed the SQL request in dbFetchRow but I am not sure it is injection_proof. Do you have an example of an injection_proof dbFetchRow call, please?
And should I also check/validate the bgp_descr HTTP POST variable or the json_decode function is enough?

murrant:
@geg347 FYI, all db* function bypass any SQL injection protection. (because it is a compatibility shim)  Native Laravel queries are much safer and more portable.

geg347:
Hello @murrant,
OK so the dbFetchRows function should be safe by testing if the id is numeric.
But for the dbUpdate I copy paste code from existing API function for editing bills.
I don't know how to protect this from SQL injection... Should I edit the common function dbFacile to prepare the SQL request as PDO can do? Is there some other tests possible in the edit_bgp_descr function?

geg347:
Hello,
Any update on this please?
What can I do, do you have a working example?

includes/html/api_functions.inc.php
```diff
@@ -602,6 +602,43 @@ function get_bgp(Illuminate\Http\Request $request)
...
+    }
+
+    $bgp = [];
+    //find existing bgp for update
+    $bgpPeerId = $request->route('id');
+    if (! is_numeric($bgpPeerId)) {
+        return api_error(400, 'Invalid id has been provided');
+    }
+
+    $bgps = dbFetchRows('SELECT * FROM `bgpPeers` WHERE `bgpPeer_id` = ?', [$bgpPeerId]);
```
murrant, line 619: $peer = \App\Models\BgpPeer::firstWhere('bgpPeer_id', $bgpPeerId);

includes/html/api_functions.inc.php
```diff
@@ -602,6 +602,43 @@ function get_bgp(Illuminate\Http\Request $request)
...
+
+    // get description
+    $bgp_descr = $data['bgp_descr'];
+
+    // update existing bgp
+    if (count($bgps) == 0) {
```
murrant, line 625: if ($peer === null) {

includes/html/api_functions.inc.php
```diff
@@ -602,6 +602,43 @@ function get_bgp(Illuminate\Http\Request $request)
...
+        return api_error(404, 'BGP peer ' . $bgpPeerId . ' does not exist');
+    } elseif (is_array($bgps) && count($bgps) == 1) {
```
murrant, line 627: delete this elseif we already know one was found

includes/html/api_functions.inc.php
```diff
@@ -602,6 +602,43 @@ function get_bgp(Illuminate\Http\Request $request)
...
+        $bgp = $bgps[0];
+
+        $update_data = [
```
murrant, line 630: $peer->bgpPeerDescr = $bgp_descr;

includes/html/api_functions.inc.php
```diff
@@ -602,6 +602,43 @@ function get_bgp(Illuminate\Http\Request $request)
...
+            'bgpPeerDescr' => $bgp_descr,
+        ];
+        $update = dbUpdate($update_data, 'bgpPeers', 'bgpPeer_id=?', [$bgpPeerId]);
```
murrant, line 633: if ($peer->save()) {

murrant (changes requested):

includes/html/api_functions.inc.php
geg347: I have put a last line : return api_error(500, 'Failed to update existing bgp');
I don't know if it is usefull or if it can cause some problem.

murrant (approved):

includes/html/api_functions.inc.php
murrant: Yeah, better to have more error messages than less to help the api user.

PipoCanaja:
All good then @geg347
Thx for the contribution.

PR #13069, opened 2021-07-25
endrun_updated_support

hanserasmus:
Added OS support for Endrun Sonoma Meridia II devices. Also updated original Sonoma device discovery to include OS version.

#### Please note

#### Testers


includes/definitions/endrun.yaml
```diff
@@ -19,3 +19,8 @@ discovery:
           oid: .1.3.6.1.4.1.13827.12.1.16.0
           op: 'contains'
           value: 'Sonoma_D12'
+    -
+      snmpget:
+          oid: .1.3.6.1.4.1.13827.14.1.16.0
```
PipoCanaja, line 24: Hi @hanserasmus 
You need to add 
```      sysObjectID: .1.3.6.1.4.1.8072.3.2.10```
above the snmpget line. Exactly like Tempus LX CDMA. 

And this should also be changed for Sonoma D12. 

The idea is that if (and only if) sysObjectID equals  .1.3.6.1.4.1.8072.3.2.10, then the snmpget will take place. It will ensure that this snmpget is not tried on all devices -> results in quicker discovery with less unnecessary snmpget for all other devices.

**Review by PipoCanaja — CHANGES_REQUESTED** [review]

File: includes/definitions/endrun.yaml (hunk 1 above)
Comment (hanserasmus) [reply includes/definitions/endrun.yaml:24]: Ok. Thank you @PipoCanaja  I will do so and resubmit. I see now that there is one failed test as well, will fix the indentation as well.

**Comment by hanserasmus** [comment]
Please ignore this merge. Adding the code change @PipoCanaja requested breaks the discovery of both Sonoma and Meridian devices now.

**Comment by PipoCanaja** [comment]
@hanserasmus 
Not according to the github tests. Which probably means that the test-data is very incomplete for these devices. Could you post the data for the devices that are failing now ?

**Comment by hanserasmus** [comment]
> @hanserasmus
> Not according to the github tests. Which probably means that the test-data is very incomplete for these devices. Could you post the data for the devices that are failing now ?

My problem is as follows:

If I change the line in the OS/Endrun.php to read like this:
`$info = snmp_get_multi($this->getDeviceArray(), ['gntpVersion.0', 'snmpSetSerialNo.0'], '-OQUs', 'SONOMA-MIB:MERIDIAN2-MIB:SNMPv2-MIB');`

then only the MERIDIAN devices work. So, Sonoma_D12 is picked up as endrun, but all other values except Serial is unset. If I change the line to read:
`$info = snmp_get_multi($this->getDeviceArray(), ['gntpVersion.0', 'snmpSetSerialNo.0'], '-OQUs', 'SONOMA-MIB:SNMPv2-MIB');`
then Meridian devices are discovered as generic, or if previously discovered, they get unset. So, to my knowledge when I ran the initial tests for the first submission, having the line with multiple MIBS set (SONOMA+MERIDIAN+SNMP_v2) worked. But now, if I set the line to read like that, the device details are being unset. Not sure why that would happen.

File: LibreNMS/OS/Endrun.php [hunk 2]
```diff
@@ -34,15 +34,14 @@ class Endrun extends OS implements OSDiscovery
     public function discoverOS(Device $device): void
     {
         if (Str::contains($device->sysDescr, 'Sonoma')) {
-            $info = snmp_get_multi($this->getDeviceArray(), ['gntpVersion.0', 'snmpSetSerialNo.0'], '-OQUs', 'SONOMA-MIB:SNMPv2-MIB');
+            $info = snmp_get_multi($this->getDeviceArray(), ['gntpVersion.0', 'snmpSetSerialNo.0'], '-OQUs', 'SONOMA-MIB:MERIDIAN2-MIB:SNMPv2-MIB');
```
murrant, line 37: If both contain gntpVersion or snmpSetSerialNo, it will only resolve the oid from the first listed.  It will not try one than the other.

tests/data/endrun_sonoma.json
```diff
@@ -8,7 +8,7 @@
                     "sysDescr": "Linux NTSDC3NTPPTP 3.2.2-Sonoma 6010-0064-000_v2.01 #1 PREEMPT Sun Nov 13 21:31:23 UTC 2016 armv5tel",
                     "sysContact": "<private>",
                     "version": null,
-                    "hardware": "Sonoma_D12 GPS 6010-0065-000",
+                    "hardware": null,
```
murrant, line 11: Why did we lose the hardware version here?

murrant: (probably my comment about the two MIBs above)

LibreNMS/OS/Endrun.php
murrant: Might be easier to implement this in yaml and call `parent::discoverOS($device)`  You can list both separately and it will try one then the other.

hanserasmus: All this regex stuff and yaml. this ought to be fun. Thanks, will try and see how far I get.

hanserasmus:
@murrant anything more I need to do? I see the tests fail, but not sure if it is a "true failure"? And if so, how I can fix it?

murrant:
@hanserasmus sorry for being late, you need to update the test data to remove the no-longer present fields. (and merge endrun class with upstream)

murrant:
There is also unwanted file mode (permissions) changes in this PR.

hanserasmus:
@murrant Thank you for the patience again. Hopefully all is good now?

PR #13070, opened 2021-07-26
WIP: Add support for multiple Geist watchdog sensors

steelfly33:
Geist Watchdog devices have only been able to inherently support the built-in 'internalTable' for standard temperature monitoring and alerting.  This update adds support for the correct MIB oids to support the discovery and polling of external temperature sensors probes connected to the Watchdog.

* Tested on devices running firmware version 3.4.0

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/sensors/temperature/geist-watchdog.inc.php
+++ b/includes/discovery/sensors/temperature/geist-watchdog.inc.php
@@ -46,4 +46,23 @@ foreach ($temp_table as $index => $temp_data) {
     }
 }
 
+$sensor_count = 0;
+$temp_table = snmpwalk_cache_oid($device, 'internalTable', [], 'GEIST-V4-MIB');
+foreach ($temp_table as $index => $temp_data) {
+    $current_oid = '.1.3.6.1.4.1.21239.5.1.2.1.5.' . $index;
+    $descr = $temp_data['internalName'];
+    $value = $temp_data['internalTemp'];
+    discover_sensor($valid['sensor'], 'temperature', $device, $current_oid, $sensor_count++, 'geist-watchdog', $descr, 1, 1, null, null, null, null, $value);
+}
+
+$temp_table = snmpwalk_cache_oid($device, 'tempSensorTable', [], 'GEIST-V4-MIB');
+foreach ($temp_table as $index => $temp_data) {
+    if ($temp_data['tempSensorAvail'] == 1) {
+        $current_oid = '.1.3.6.1.4.1.21239.5.1.4.1.5.' . $index;
+        $descr = $temp_data['tempSensorName'];
+        $value = $temp_data['tempSensorTemp'];
+        discover_sensor($valid['sensor'], 'temperature', $device, $current_oid, $sensor_count++, 'geist-watchdog', $descr, 1, 1, null, null, null, null, $value);
+    }
+}
+
 unset($temp_table);
--- a/includes/polling/sensors/temperature/geist-watchdog.inc.php
+++ b/includes/polling/sensors/temperature/geist-watchdog.inc.php
@@ -22,5 +22,7 @@
  * @author     Neil Lathwood <gh+n@laf.io>
  */
 if ($sensor_cache['geist_temp_unit'] === '0') {
-    $sensor_value = fahrenheit_to_celsius($sensor_value / 10) * 10;
+    $sensor_value = fahrenheit_to_celsius($sensor_value / 10);
+} else {
+    $sensor_value = $sensor_value / 10;
 }


Jellyfrog:
Hi @steelfly33 
Thank you for your contribution, would you mind looking into adding this in yaml format instead?

steelfly33:
> Hi @steelfly33
> Thank you for your contribution, would you mind looking into adding this in yaml format instead?

@Jellyfrog I saw two paths towards a solution and wasn't sure which was preferred.  WRT the yaml file, I believe that the root of the problem is that the sensor index does not get incremented for each new temperature sensor discovered and added.  When learning from the yaml file, it looks like the internal temperature sensor is found and added and then the remote temperature sensor is found and added with the same sensor index.  Can you point me towards where I might handle this via the yaml file?  I found the correct insertion point under `includes`, but did not trace the yaml parsing code.  I am finding my solution really has to delete the yaml file to prevent learning from it from updating the sensors incorrectly (I end up with two sensors of the same name and value).

murrant:
@steelfly33 this happens when you add the same type of sensors from two different tables with overlapping indexes.
Instead of allowing automatic index, use manual: `index: 'internal.{{ $index }}`  and `index: 'remote.{{ $index }}`

steelfly33:
> @steelfly33 this happens when you add the same type of sensors from two different tables with overlapping indexes.
> Instead of allowing automatic index, use manual: `index: 'internal.{{ $index }}` and `index: 'remote.{{ $index }}`

@murrant Using that syntax in the yaml leads to the same result as if there were no `index`  specified.  I still only get one temperature sensor.

```
                -
                    oid: internalTable
                    value: internalTemp
                    num_oid: '.1.3.6.1.4.1.21239.5.1.2.1.5.{{ $index }}'
                    descr: internalName
                    index: 'internal.{{ $index }}'
                -
                    oid: tempSensorTable
                    value: tempSensorTemp
                    num_oid: '.1.3.6.1.4.1.21239.5.1.4.1.5.{{ $index }}'
                    descr: tempSensorName
                    index: 'remote.{{ $index }}'
```

steelfly33:
Using the yaml file (`includes/definitions/discovery/geist-watchdog.yaml`), the problem is the first line where the mib is specified.
`mib: GEIST-V4-MIB:GEIST-MIB-V3`

The oids for the `tempSensorTable` are different between those mib versions.
```
snmptranslate -On -M /opt/librenms/mibs/geist:/opt/librenms/mibs  GEIST-MIB-V3::tempSensorTable
.1.3.6.1.4.1.21239.2.4
snmptranslate -On -M /opt/librenms/mibs/geist:/opt/librenms/mibs  GEIST-V4-MIB::tempSensorTable
.1.3.6.1.4.1.21239.5.1.4
```
For my watchdog sensor, which uses the V4 MIB, it gives no response to the V3 tempSensorTable query (`No Such Object available on this agent at this OID`) and discovery stops there for that table.  If I change the order of the MIBs in the yaml file, I can correctly detect my remote sensors, but that may break users who have V3 remote sensors.

Since I have no V3 sensors, simply removing the MIB was a quick solution:
```
--- a/includes/definitions/discovery/geist-watchdog.yaml
+++ b/includes/definitions/discovery/geist-watchdog.yaml
@@ -1,4 +1,4 @@
-mib: GEIST-V4-MIB:GEIST-MIB-V3
+mib: GEIST-V4-MIB
 modules:
     sensors:
         temperature:
```

I am open to suggestions on the best approach using the yaml file to resolve this issue to support both versions.

murrant:
> @murrant Using that syntax in the yaml leads to the same result as if there were no `index` specified. I still only get one temperature sensor.

It should not. Do you mind posting a discovery debug output?

murrant:
as for the MIB, do sensors allow you to specify the mib in the sensor?  I don't remember.

steelfly33:
> Using the yaml file (`includes/definitions/discovery/geist-watchdog.yaml`), the problem is the first line where the mib is specified.
> `mib: GEIST-V4-MIB:GEIST-MIB-V3`
> 
> The oids for the `tempSensorTable` are different between those mib versions.
> 
> ```
> snmptranslate -On -M /opt/librenms/mibs/geist:/opt/librenms/mibs  GEIST-MIB-V3::tempSensorTable
> .1.3.6.1.4.1.21239.2.4
> snmptranslate -On -M /opt/librenms/mibs/geist:/opt/librenms/mibs  GEIST-V4-MIB::tempSensorTable
> .1.3.6.1.4.1.21239.5.1.4
> ```
> 
> For my watchdog sensor, which uses the V4 MIB, it gives no response to the V3 tempSensorTable query (`No Such Object available on this agent at this OID`) and discovery stops there for that table. If I change the order of the MIBs in the yaml file, I can correctly detect my remote sensors, but that may break users who have V3 remote sensors.
> 
> Since I have no V3 sensors, simply removing the MIB was a quick solution:
> 
> ```
> --- a/includes/definitions/discovery/geist-watchdog.yaml
> +++ b/includes/definitions/discovery/geist-watchdog.yaml
> @@ -1,4 +1,4 @@
> -mib: GEIST-V4-MIB:GEIST-MIB-V3
> +mib: GEIST-V4-MIB
>  modules:
>      sensors:
>          temperature:
> ```
> 
> I am open to suggestions on the best approach using the yaml file to resolve this issue to support both versions.
> 
> EDIT: I also made the change above WRT `index`



> > @murrant Using that syntax in the yaml leads to the same result as if there were no `index` specified. I still only get one temperature sensor.
> 
> It should not. Do you mind posting a discovery debug output?

I noted (somewhat convolutedly) that this did work, but it had to be coupled with the removal of the V3 mib.

steelfly33:
> as for the MIB, do sensors allow you to specify the mib in the sensor? I don't remember.

Great question.  If they do, I am unaware of how to specify it.

murrant:
@steelfly33 planning on finishing this? Let us know to re-open.

PR #13078, opened 2021-07-29
Allow stack index other than '1.' for CiscoSB

dennypage:
When in a stack configuration, the index for each port is based on the unit number in the stack. The first unit in the stack uses index '1.', the second uses index '2.', the third '3.' and so on.

The prior code had a hard coded index of '1.' which prevented get_port_by_index_cache() lookup from finding the correct port for any port not on the first stack unit.

This resulted in the port description being missing from tables and graphs like so:

port_desc_missing

This change allows for stack indexes values other than '1.', which corrects the missing port description:

port_desc_present

#### Please note

- [ x] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers


mpikzink:
With the number of sensors it would make sense to group them by stackunit. (Via the group field)

dennypage:
> With the number of sensors it would make sense to group them by stackunit. (Via the group field)

More words please...

mpikzink:
Have a try and fill the $group variable with the Stackunit index. In the discoverSensor function. It might give us a better overview.

dennypage:
> Have a try and fill the $group variable with the Stackunit index. In the discoverSensor function. It might give us a better overview.

Taking a quick look at that, it seems to me that grouping would more appropriate as a different PR. That appears to be more of a feature enhancement, vs the index issue which is a simple defect. This PR only addresses the index defect.

mpikzink:
Yes, you we do that, it just popped into my head when I saw your screenshot... Go ahead :)

Jellyfrog:
> Please add test data so we can ensure your change is not broken in the future.
> 
> Read the docs to find out how: https://docs.librenms.org/Developing/os/Test-Units

Would you mind generating some test data for this?

dennypage:
> Would you mind generating some test data for this?

I don't have snmpsim, but I can give you an snmprec file. Do you want just the sensors module, or do you want the whole nut? Either way, it will take a bit of work to redact.

dennypage:
Here is the snmprec file generated by scripts/collect-snmp-data.php:

[ciscosb_ciscosb.snmprec.zip](https://github.com/librenms/librenms/files/6911998/ciscosb_ciscosb.snmprec.zip)

dennypage:
Thank you @murrant

Jellyfrog (approved):

PR #13094, opened 2021-08-02
Full Python code fusion / refactor and hardening

deajan:

This PR is a full refactor of all the python code, which has grown to have many duplicates over time.
This is an improvement of PR #13090 which only covers wrapper refactoring.
The main aim is to make the codebase leaner and more maintainable, while hardening the code.
It should have no user visible effect.

## Enhancements:
- All system calls are now guaranteed to:
  - Have a timeout preventing hanging (for service and wrapper)
  - Always return an exit code
  - Have interpretable results (there's always a result, regardless of OS or script failure)
  - Work regardless of Python versions of subprocess module
- discovery, service-poller and poller wrapping is now fully logged
- Retain compatibility with earlier LibreNMS implementations
- Add an log message when too old Redis is used

## Fixes:
- Removed duplicate code for
   - Database connection functionnality
   - Config fetching
- Merged discovery, service-poller and poller wrapper code into a single wrapper
- Linting fixes

**DISCLAIMER**: The code has not been enough tested yet. Please comment on the changes.
Also, I would love if someone could check the wrappers with distributed polling.

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e  `./scripts/github-apply 13094` 


deajan:
So far so good, I could test those changes on two different LibreNMS instances.
The wrapper usage was tested, the librenms service was tested.
Wrapper was tested with distributed and standard polling.
Service was tested with redis.

Code does not break anything for me ;)
Can anyone confirm ?

paulierco:
Just started to test this in a small env, just 100 devices. So far looks good. 
Should we expect better performance ?

deajan:
No performance enhancements, it's just easier to maintain code, better logs, and guarantee that nothing may hang.
Performance wise, it shouldn't really by python that is the bottleneck anyway IMO.

ottorei:
Testing this with a distributed setup of 5 servers and ~7600 devices now. Been running for half an hour now without noticeable issues.

For some reason I have had some trouble with alerting with the current Python service - sometimes the alerts are delayed like multiple days for transports only but they are instantly transported when running alerts.php from the console. I believe that something causes the alerting queue to hang and I am not sure what is causing this. Sometimes the master node seems to have changed - might have something to do with this.

deajan:
@ottirei So I guess you are using redis distributed caching and not memcached ?

As for the alert issue, this is exactly what this PR should resolve, since it will stop queue execution after one hour, which should be fairly suffficient IMO. But since it's a transient error, you might run the service with -d debug flag in order to have more info.

Thanks for having taken the time for testing ;)

murrant (approved):

deajan:
So far so good, the PR has been tested on 4 LibreNMS setups.
Is it enough to get merged ?

ottorei:
> 
> 
> @ottorei So I guess you are using redis distributed caching and not memcached ?
> 
> As for the alert issue, this is exactly what this PR should resolve, since it will stop queue execution after one hour, which should be fairly suffficient IMO. But since it's a transient error, you might run the service with -d debug flag in order to have more info.
> 
> Thanks for having taken the time for testing ;)

Yes, I am using Redis. Do you mean that if any issues were to occur with the queue worker, it would restart after 1 hour? Could this value be user configurable? Also, if the queue were to stall, could it be verified from logs? At the moment all my pollers seem to be working fine with this patch.

deajan:
The queueworker itself won't be restarted, but the scripts that get executed (`alert.php`, `billing.php`, and so on) would be stopped after an hour, handing back the control to the queueworker.
The timeout is "hand" configurable in `LibreNMS/__init__.py` at line `DEFAULT_SCRIPT_TIMEOUT = 3600`
Running with debug flag should give full execution logs of those scripts with this patch.

murrant (approved):
LGTM, but I think there is a high chance we might have to revert this one.  There are a lot of different python configurations out there.

deajan:
I'll stay tuned to fix potential issues.

paulierco:
@deajan Can you join discord and check the general channel ? There are some users that complains after this patch.

For example one has this error:

> Could not import .env - check that the poller user can read the file, and that composer install has been run recently
> ERROR: Could not execute command [['/usr/bin/env', 'php', '/opt/librenms/config_to_json.php']]: Expecting value: line 1 column 1 (char 0)
> Traceback (most recent call last):
>   File "/opt/librenms/poller-wrapper.py", line 37, in <module>
>     log_dir = config["log_dir"]
> TypeError: 'NoneType' object is not subscriptable

deajan:
@paulierco I'm on it. Strange, I really don't see how `composer install` should fix this import problem, unless it somehow rewrites `.env` which in turn makes `dotenv` python package happy.
At least, since there's a will remove `config.php` in favor of `.env`, the setup of the guy who had trouble should be fixed by now.

PR #13098, opened 2021-08-03
Rename nobulk -> snmp_bulk

Jellyfrog:
Rename option to align with the rest.
Please note it switches from a negative to positive option!

#### Please note

#### Testers

### DIFF
--- a/includes/snmp.inc.php
+++ b/includes/snmp.inc.php
@@ -129,7 +129,7 @@ function gen_snmpget_cmd($device, $oids, $options = null, $mib = null, $mibdir =
  */
 function gen_snmpwalk_cmd($device, $oids, $options = null, $mib = null, $mibdir = null, $strIndexing = null)
 {
-    if ($device['snmpver'] == 'v1' || (isset($device['os']) && Config::getOsSetting($device['os'], 'nobulk'))) {
+    if ($device['snmpver'] == 'v1' || (isset($device['os']) && Config::getOsSetting($device['os'], 'snmp_walk', true) == false)) {
         $snmpcmd = [Config::get('snmpwalk')];
     } else {
         $snmpcmd = [Config::get('snmpbulkwalk')];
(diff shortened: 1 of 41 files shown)


Jellyfrog:
Please note it switches from negative to positive option

PipoCanaja:
Hi @Jellyfrog 
Shouldn't we add a notification to this change?

Jellyfrog:
Doubt it's needed

includes/snmp.inc.php
```diff
@@ -129,7 +129,7 @@ function gen_snmpget_cmd($device, $oids, $options = null, $mib = null, $mibdir =
  */
 function gen_snmpwalk_cmd($device, $oids, $options = null, $mib = null, $mibdir = null, $strIndexing = null)
 {
-    if ($device['snmpver'] == 'v1' || (isset($device['os']) && Config::getOsSetting($device['os'], 'nobulk'))) {
+    if ($device['snmpver'] == 'v1' || (isset($device['os']) && Config::getOsSetting($device['os'], 'snmp_walk', true) == false)) {
```
murrant, line 132: snmp_walk  --> snmp_bulk

murrant (changes requested):
fix typo ;)

includes/snmp.inc.php
Jellyfrog: Argh. How did I manage that...

murrant:
We probably do need a notification or backwards compatibility.   This is a likely option for people to set in their local config for problem devices.

Jellyfrog:
Hmm I see.

Well dont know if we should do it with the deprecated-config function thingy. because we need to loop all OSs...

Notification might be enough? :)

murrant:
Yep, fine with just a notification.  You can add it to the in-tree rss or the one on the web site.

murrant (approved):

PR #13109, opened 2021-08-05
Allow more compatibility in STP port discovery/polling

PipoCanaja:
This fixes Zyxel STP discovery (using "dots" as MAC separator, and using xxx-yy for priority and port)

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/stp.inc.php
+++ b/includes/discovery/stp.inc.php
@@ -126,12 +126,12 @@ if ($stpprotocol == 'ieee8021d' || $stpprotocol == 'unknown') {
             // set port binding
             $stp_port['port_id'] = dbFetchCell('SELECT port_id FROM `ports` WHERE `device_id` = ? AND `ifIndex` = ?', [$device['device_id'], $stp_raw[$port]['dot1dStpPort']]);
 
-            $dr = str_replace([' ', ':', '-'], '', strtolower($stp_raw[$port]['dot1dStpPortDesignatedRoot']));
-            $dr = substr($dr, -12); //remove first two octets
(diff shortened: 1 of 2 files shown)


includes/discovery/stp.inc.php
```diff
@@ -126,12 +126,12 @@
             // set port binding
             $stp_port['port_id'] = dbFetchCell('SELECT port_id FROM `ports` WHERE `device_id` = ? AND `ifIndex` = ?', [$device['device_id'], $stp_raw[$port]['dot1dStpPort']]);
 
-            $dr = str_replace([' ', ':', '-'], '', strtolower($stp_raw[$port]['dot1dStpPortDesignatedRoot']));
-            $dr = substr($dr, -12); //remove first two octets
+            $dr = str_replace(['.', ' ', ':', '-'], '', strtolower($stp_raw[$port]['dot1dStpPortDesignatedRoot']));
```
mpikzink, line 129: I would prefer LibreNMS\Util\Rewrite\macToHex()

PipoCanaja: That's a point. Will check !

PipoCanaja: Hi @mpikzink 
Not really possible cause there is no easy way to identify where the Mac Address starts. 
Basically, macToHex expects a mac address only, not a string ending with a mac address as we have here. And because we don't know where the mac address starts, we can only do the string replacements, and then keep the last 12 chars. 
So this PR cannot use macToHex

murrant (approved):

PR #13117, opened 2021-08-07
Added OS Security updates (for YUM only so far)

wimg:
Adds OS Security updates, useful when managing tons of servers to avoid notification fatigue caused by non-stop updates of packages.
This way, security updates can be placed in a rule separately from standard os updates.

Will submit the required change for librenms-agent in a PR in that repo in a minute.

#### Please note

#### Testers

### DIFF
--- a/includes/polling/applications/os-updates.inc.php
+++ b/includes/polling/applications/os-updates.inc.php
@@ -9,12 +9,18 @@ $mib = 'NET-SNMP-EXTEND-MIB';
 $oid = '.1.3.6.1.4.1.8072.1.3.2.4.1.2.8.111.115.117.112.100.97.116.101.1';
 
 $rrd_name = ['app', $name, $app_id];
-$rrd_def = RrdDefinition::make()->addDataset('packages', 'GAUGE', 0);
+$rrd_def = RrdDefinition::make()
+    ->addDataset('packages', 'GAUGE', 0)
+    ->addDataset('security_updates', 'GAUGE', 0);
 
 $osupdates = snmp_get($device, $oid, $options, $mib);
+$lines = explode(',', $osupdates);
 
-$fields = ['packages' => $osupdates];
+$fields = [
+    'packages' => is_numeric($lines[0]) ? $lines[0] : null,
+    'security_updates' => is_numeric($lines[1]) ? $lines[1] : null
+];
 
 $tags = ['name' => $name, 'app_id' => $app_id, 'rrd_def' => $rrd_def, 'rrd_name' => $rrd_name];
 data_update($device, 'app', $tags, $fields);
-update_application($app, $osupdates, $fields, $osupdates);
+update_application($app, $osupdates, $fields);
--- a/doc/Extensions/Applications.md
+++ b/doc/Extensions/Applications.md
@@ -1351,6 +1351,7 @@ The application should be auto-discovered as described at the top of
 the page. If it is not, please follow the steps set out under `SNMP
 Extend` heading top of page.
 
+
 ## PHP-FPM
 
 ### SNMP Extend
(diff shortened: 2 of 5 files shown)


wimg:
librenms-agent PR : https://github.com/librenms/librenms-agent/pull/373

wimg:
This all works, except for the fact that, if you already had a graph with osupdates, it now returns an error because we try adding a column to the RRD. If anyone has a suggestion on how to fix this (other than deleting the RRD), let me know.

PipoCanaja:
> This all works, except for the fact that, if you already had a graph with osupdates, it now returns an error because we try adding a column to the RRD. If anyone has a suggestion on how to fix this (other than deleting the RRD), let me know.

Hi @wimg 
The solution is to create a new RRD for your new variable, and keep the existing RRD for the "legacy" variables. This way, the migration will be smooth.

SourceDoctor:
@PipoCanaja 
i think this goes into the way of migrating further data to existing RRDs

https://stackoverflow.com/questions/51393953/updating-values-in-an-existing-rrd-file

but i think this has to be tested before, and on success maybe we should write a generic migrations script for future uses.
This case is not the first time such feature would be useful

murrant:
@SourceDoctor I've investigated that before, that is the way of pain.  I've settled on the fact that the optimal rrd configuration is 1 DS per RRD file.

@wimg instead of adding a field to the existing rrd file, create a new rrd file.

includes/polling/applications/os-updates.inc.php
```diff
@@ -9,12 +9,18 @@
 $oid = '.1.3.6.1.4.1.8072.1.3.2.4.1.2.8.111.115.117.112.100.97.116.101.1';
 
 $rrd_name = ['app', $name, $app_id];
-$rrd_def = RrdDefinition::make()->addDataset('packages', 'GAUGE', 0);
+$rrd_def = RrdDefinition::make()
+    ->addDataset('packages', 'GAUGE', 0)
+    ->addDataset('security_updates', 'GAUGE', 0);
 
 $osupdates = snmp_get($device, $oid, $options, $mib);
+$lines = explode(',', $osupdates);
```
murrant, line 17: There is a new format for apps, that supports versioning and uses json encoding.  Allows us to support multiple versions of the client nicely.

wimg: I'm sorry, I don't understand your comment. Could you explain in a different way ?

murrant: https://github.com/librenms/librenms/blob/master/includes/polling/applications/docker.inc.php#L31

wimg: I'm sorry, I just don't understand where you're going with your sentence and how it applies to the line you highlighted. Please explain.

murrant: This line is parsing the output of the snmp extend... we have other code for that which is much safer and can handle different clients.  The main thing is using the json_app_get() function, but also you should catch and handle errors (and implement a fallback for old extend scripts that only supply one category of updates.

Unfortunately, I couldn't find any documentation for this, so the only documentation is the code:

$ grep -lr json_app_get includes/polling/applications
includes/polling/applications/gpsd.inc.php
includes/polling/applications/mailcow-postfix.inc.php
includes/polling/applications/puppet-agent.inc.php
includes/polling/applications/ntp-client.inc.php
includes/polling/applications/mdadm.inc.php
includes/polling/applications/fail2ban.inc.php
includes/polling/applications/powermon.inc.php
includes/polling/applications/backupninja.inc.php
includes/polling/applications/certificate.inc.php
includes/polling/applications/ntp-server.inc.php
includes/polling/applications/pureftpd.inc.php
includes/polling/applications/fbsd-nfs-client.inc.php
includes/polling/applications/zfs.inc.php
includes/polling/applications/seafile.inc.php
includes/polling/applications/chronyd.inc.php
includes/polling/applications/dhcp-stats.inc.php
includes/polling/applications/redis.inc.php
includes/polling/applications/powerdns.inc.php
includes/polling/applications/docker.inc.php
includes/polling/applications/fbsd-nfs-server.inc.php
includes/polling/applications/portactivity.inc.php
includes/polling/applications/ups-apcups.inc.php

murrant:
@wimg Let us know if you plan to finish this PR.  Would be happy to re-open it.

PR #13118, opened 2021-08-08
Priority filtering for syslog widget

Talkabout:
HI all,

I am currently in the process of migrating from Observium to LibreNMS. One of the main features I used in Observium was the syslog widget to show only critical logs. I was not able to achieve the same with the current LibreNMS widget so I started to dig a little bit into the implementation and finally found an approach to do exactly what I wanted. With this pull request the syslog widget receives an additional option allowing to define the priority level of log entries to be shown:

image

As this is my first change on the LibreNMS code I am pretty sure the code is still optimizable. Suggestions are very welcome!

Thanks!

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Select/PriorityController.php
+++ b/app/Http/Controllers/Select/PriorityController.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * PriorityController.php
+ *
+ * -Description-
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * @link       https://www.librenms.org
+ */
+
+namespace App\Http\Controllers\Select;
+
+use App\Models\Syslog;
+
+class PriorityController extends SelectController
+{
+    protected function searchFields($request)
+    {
+        return ['level'];
+    }
+
+    /**
+     * Defines the base query for this resource
+     *
+     * @param \Illuminate\Http\Request $request
+     * @return \Illuminate\Database\Eloquent\Builder|\Illuminate\Database\Query\Builder
+     */
+    protected function baseQuery($request)
+    {
+        return Syslog::query()
+          ->distinct()
+          ->select('level')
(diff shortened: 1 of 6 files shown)


PipoCanaja:
Hi @Talkabout 
You'll need to fix the CLA (Contributor License agreement). You'll probably have to ensure that your commits are not made by "root" but by the user you are registered in GitHub.

Talkabout:
Hi @PipoCanaja ,

yes, I am not sure why exactly this happened... Unfortunately I don't have enough experience with GIT yet to be able to change the commits. What is the best way to change the commti author? Is this possible at all?

Thanks!

Talkabout:
Hi @PipoCanaja ,

I have tried to fix the commits but instead git created new ones. I am not sure if it is possible to remove the ones done by root? Can you point me into the right direction?

Thanks!

murrant:
An easy way to fix your commits is to squash them.
`git rebase -i 3f9d16a096c2c0f9cc15fa847eefd0f159901eb7` (the last commit before you started work.  Leave the first one as pick and change the other commits you made to squash.   This will make them one new commit with all the changes.

As for the widget nice work, but I don't know if the dynamic drop down is necessary as we should have a finite known list of priorities.   It also prevents you from picking a priority that doesn't already exist (or currently if removing old log entries).  Might be best to just put a static list of priorities.  Do you agree?

Talkabout:
Hi @murrant ,

I have done as you suggested, but obviously the "old" commits are still "existing" and thus the cla requires the root user to sign. Can I remove the old commits somehow?
As per the dynamic selection: yes, you are right, it is not really required, but the SelectController is based on a query. As I am not that familar with Laravel I am not sure if the query can be build in a way that it returns a static list. If there is such a way, can you point me into the right API?

Thanks

PipoCanaja:
if the rebase is done as murrant suggested, and if you "force push", the original "root" commits should disappear. 
If not, another way of doing it is just doing another PR from a clean branch (without any "root" commits)

Talkabout:
Hi @PipoCanaja ,

thanks for your comment!

So the rebase command obviously didn't work, no idea why. Is there another thing I can try here to remove the commits from root?

If creating new PR is the only solution left, what happens with this PR? Can it be deleted? Does not make much sense to keep it, right?

Thanks!

Talkabout:
Hi all,

I am not sure what to do now. I have tried to create a new fork but there the files are already modified, so no new commit possible to pull. What is the best option for me now to make sure that this change finds it way into the main branch? Would be great to get some help here...!

Thanks!

PipoCanaja:
@Talkabout 
- You can save your modified files somewhere out of the git repository
- Start over from master branch
- Create a new fork
- Commit
- Create pull request.

Other option is to do all from the github website, as you have only 6 files : 
- go to '[https://github.com/librenms/librenms](https://github.com/librenms/librenms)'
- navigate to the first file you want to change
- edit it
- commit it (to a new branch as it will be suggested)
- create a pull request
- navigate to the pull request, in the "File changed" tab
- edit all 5 other files (and each edit will be a new commit into this PR)

PR #13130, opened 2021-08-12
enclose IPv6 literal in [brackets] for snmpget and unix-agent

kevinwallace:
This fixes SNMP polling when a target is specified as an IPv6 literal.  Before this change, IPv6 SNMP polling was only possible by specifying a DNS name that resolves to an IPv6 address.

See e.g. https://community.librenms.org/t/unable-to-add-ipv6-devices-using-snmp-udp6-but-can-snmpwalk-telnet-to-them-from-librenms/16374

Previously, for IP `2001:db8::1234` this code would generate an `snmpget` command line that looked like:
```
snmpget [...flags...] udp6:2001:db8::1234:161 [...oids...]
```
which is wrong.  Now, it generates:
```
snmpget [...flags...] udp6:[2001:db8::1234]:161 [...oids...]
```
which is correct.

---

#### Please note

#### Testers

### DIFF
--- a/includes/snmp.inc.php
+++ b/includes/snmp.inc.php
@@ -176,7 +176,7 @@ function gen_snmp_cmd($cmd, $device, $oids, $options = null, $mib = null, $mibdi
         array_push($cmd, '-r', $retries);
     }
 
-    $pollertarget = Device::pollerTarget($device);
+    $pollertarget = \LibreNMS\Util\Rewrite::addIpv6Brackets(Device::pollerTarget($device));
     $cmd[] = $device['transport'] . ':' . $pollertarget . ':' . $device['port'];
     $cmd = array_merge($cmd, (array) $oids);
 
--- a/LibreNMS/Util/Rewrite.php
+++ b/LibreNMS/Util/Rewrite.php
@@ -447,4 +447,14 @@ class Rewrite
     {
         return str_pad($num, $length, '0', STR_PAD_LEFT);
     }
+
+    /**
+     * If given input is an IPv6 address, wrap it in [] for use in applications that require it
+     * @param string $ip
+     * @return string
+     */
+    public static function addIpv6Brackets($ip): string
+    {
+        return IPv6::isValid($ip) ? "[$ip]" : $ip;
+    }
 }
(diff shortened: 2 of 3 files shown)


includes/snmp.inc.php
```diff
@@ -177,6 +177,9 @@ function gen_snmp_cmd($cmd, $device, $oids, $options = null, $mib = null, $mibdi
     }
 
     $pollertarget = Device::pollerTarget($device);
```
Jellyfrog, line 179: You should add this change to inside the `Device::pollerTarget()` function instead, and add a comment on why this is needed

kevinwallace: I intentionally didn't put it in `Device::pollerTarget()` because doing so would break its other callers (`device_is_up()`, `add_service()`, `includes/polling/unix-agent.inc.php`), which expect IPv6 literals *not* to be enclosed in brackets.  It's only `snmpget` & friends that require brackets, which is why I only put it here.

I've added a comment explaining why this is needed, and why only here.

Jellyfrog: I think some other services need it also, like unix-agent and maybe pingcheck?

kevinwallace: PingCheck (`app/Jobs/PingCheck.php`) passes it to `fping`'s stdin, which expects it without brackets.  You're right about unix-agent though; I overlooked `fsockopen`'s semantics.  I moved this into `pollerTarget` and added an optional boolean parameter to control it.  How's that?

murrant:
Instead of updating the device model function and adding a boolean flag (should be avoided), make a helper function probably on the \LibreNMS\Util\Rewrite class

```php
    /**
     * If given input is an IPv6 address, wrap it in [] for use in applications that require it
     * @param string $ip
     * @return string
     */
    public static function addIpv6Brackets($ip): string
    {
        return IPv6::isValid($ip) ? "[$ip]" : $ip;
    }
```

then just add that around the call where needed.

murrant:
@Jellyfrog yes pollerTarget absolutely does not belong on the device model... (did I ever mention I hate the "overwrite_ip" code?)

kevinwallace:
> Instead of updating the device model function and adding a boolean flag (should be avoided), make a helper function probably on the \LibreNMS\Util\Rewrite class

Done!

kevinwallace:
Squashed into a single commit and modified commit message to reflect the fact that `unix-agent` is included; PTAL.

Jellyfrog (approved):

murrant (approved):
LGTM

PR #13134, opened 2021-08-13
Priority filtering for syslog widget

Talkabout:
HI all,

I am currently in the process of migrating from Observium to LibreNMS. One of the main features I used in Observium was the syslog widget to show only critical logs. I was not able to achieve the same with the current LibreNMS widget so I started to dig a little bit into the implementation and finally found an approach to do exactly what I wanted. With this pull request the syslog widget receives an additional option allowing to define the priority level of log entries to be shown:

image

As this is my first change on the LibreNMS code I am pretty sure the code is still optimizable. Suggestions are very welcome!

Thanks!

#### Please note

#### Testers


Talkabout:
Hi @PipoCanaja, @murrant ,

second try with this pull request :) Hopefully now everything is correct.

Thanks for your help!

Talkabout:
Hi @murrant ,

as you suggested in the previous pull request I have changed the "PriorityController" to return a static list instead of a dynamic query. Not sure if the code/approach is correct or there is a better way to do this.

Thanks!

murrant:
@Talkabout thanks, although I guess I wasn't clear enough :)

You really just need a regular select without using select2 to fetch data from the backend since this is just a static list.
So you can completely delete PriorityController and related items and just list out the options in html

Talkabout:
Hi @murrant ,

you are totally right, it makes perfect sense! I was not sure if the backend logic is required for all options so I never considered that. But now looking also at other implementations (widgets) I can see, that this is quite common. I updated the code accordingly.

Thanks!

resources/views/widgets/settings/syslog.blade.php
```diff
@@ -25,6 +25,16 @@
         <label for="hidenavigation-{{ $id }}" class="control-label">@lang('Hide Navigation')</label>
         <input type="checkbox" class="form-control" name="hidenavigation" id="hidenavigation-{{ $id }}" value="{{ $hidenavigation }}" data-size="normal" @if($hidenavigation) checked @endif>
     </div>
+    <div class="form-group">
+        <label for="level-{{ $id }}" class="control-label">@lang('Priority')</label>
+        <select class="form-control" name="level" id="level-{{ $id }}">
+            @if($level)
```
murrant, line 31: why the @if($level) ?

app/Http/Controllers/Widgets/SyslogController.php
```diff
@@ -43,6 +44,8 @@ public function getSettingsView(Request $request)
 
         $data['device'] = Device::hasAccess($request->user())->find($data['device']);
 
+        $data['priorities'] = array_flip(app('translator')->get('syslog.severity'));
```
murrant, line 47: You don't need to array_flip() here, just flip $name and $val in syslog.blade.php foreach

resources/views/widgets/settings/syslog.blade.php
Talkabout: was a leftover from the previous approach, removed.

app/Http/Controllers/Widgets/SyslogController.php
Talkabout: Done as requested.

murrant (approved):
Looks good!  Thanks for seeing it through.

PR #13141, opened 2021-08-15
Dark mode improvements

facuxt:
- Minor change to internal css to modify all tables without 'table' class applied
image
- Transparency on mini graphs on ports/details
- Dark background on most popups
image
image
- Alert table without any alerts now has dark background
image

#### Please note

#### Testers

### DIFF
--- a/html/css/dark.css
+++ b/html/css/dark.css
@@ -7232,3 +7232,7 @@
   #overDiv{
     background: #353a41;
   }
+  .bootgrid-table td.loading, .bootgrid-table td.no-results {
+    background: #3e444c;
+    text-align: center;
+  }
--- a/html/css/dark.css
+++ b/html/css/dark.css
@@ -7210,6 +7210,15 @@
     color: black;
   }
 
-.expandable:hover span {
-    background-color: #2e3338;
-}
+  .expandable:hover span {
+      background-color: #2e3338;
+  }
+  table.table tr:not(:first-child){
+    background: #3e444c!important;
+  }
+  table.table tr:not(:first-child):hover{
+    background: #4f565f!important;
+  }
+  div[role=listbox]{
+    background: rgb(70, 76, 83);
+  }
--- a/html/css/dark.css
+++ b/html/css/dark.css
@@ -7222,3 +7222,13 @@
   div[role=listbox]{
     background: rgb(70, 76, 83);
   }
+  .overlib, .overlib-contents{
+    background: #353a41;
+    color: #bebfbf;
+  }
+  .overlib-text, .overlib-contents list-large{
+    color: #bebfbf;
+  }
+  #overDiv{
+    background: #353a41;
+  }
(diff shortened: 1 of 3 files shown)


Talkabout:
Hi @facuxt ,

I have tested your pr and it makes the dark layout more consistent. There is one thing missing which is to make the "Apps" overview also transparent. Currently it looks like:

image

By removing the hardcoded background color from "includes/html/pages/apps/overview.inc.php" (line 46) it looks like:

image

Maybe you can add this change to your pr also.

html/css/dark.css
```diff
@@ -7213,12 +7213,26 @@
   .expandable:hover span {
       background-color: #2e3338;
   }
-  table.table tr:not(:first-child){
```
Jellyfrog, line 7216: Could we add the class to the tables that's missing it instead?

includes/html/print-interface.inc.php
```diff
@@ -82,11 +82,11 @@
 
 if ($port_details) {
     $port['graph_type'] = 'port_bits';
-    echo generate_port_link($port, "<img src='graph.php?type=port_bits&amp;id=" . $port['port_id'] . '&amp;from=' . Config::get('time.day') . '&amp;to=' . Config::get('time.now') . '&amp;width=100&amp;height=20&amp;legend=no&amp;bg=' . str_replace('#', '', $row_colour) . "'>");
+    echo generate_port_link($port, "<img src='graph.php?type=port_bits&amp;id=" . $port['port_id'] . '&amp;from=' . Config::get('time.day') . '&amp;to=' . Config::get('time.now') . '&amp;width=100&amp;height=20&amp;legend=no&amp;bg=' . str_replace('#', '', $row_colour) . "00'>");
```
Jellyfrog, line 85: how does this affect light theme?

murrant: seems like it always makes the color 100% transparent

facuxt: Sorry for the delay, im really busy during the week, it always make the color 100% transparent its a upgrade even to the light, blue and mono theme, it looks like in all other themes, the background is always white (screenshot1), it just changes the behavior when you do hover over the port by itself... see screenshot 2


screenshot1
image

Screenshot2 (this will now be transparent so the rectangles will not be visible)
image

html/css/dark.css
facuxt: The code to set this table a class is buried inside lots of confusing php files... i could also do it, but we face the risk to not find all the use scenarios/cases in which the table is used.

facuxt:
> Hi @facuxt ,
> 
> I have tested your pr and it makes the dark layout more consistent. There is one thing missing which is to make the "Apps" overview also transparent. Currently it looks like:
> 
> image
> 
> By removing the hardcoded background color from "includes/html/pages/apps/overview.inc.php" (line 46) it looks like:
> 
> image
> 
> Maybe you can add this change to your pr also.

Done!, Thanks for the input.

facuxt:
is this OK to merge? @murrant @Jellyfrog  ?

Jellyfrog (approved):

PR #13143, opened 2021-08-16
Cleanup Port hover minigraph description when using "Interface Description Parsing"

fbourqui:
Use port_descr_descr if exists (it should if using Interface Description Parsing), if no default to std ifDescr.

Port description: Cust: Example Customer [10Mbit] (T1 Telco Y CCID129031) {EXAMP0001}
Current result: Cust: Example Customer [10Mbit] (T1 Telco Y CCID129031) {EXAMP0001}
NEW description: Example Customer

Port description: Core: core.router01 FastEthernet0/0 (Telco X CCID023141)
Current result in device-Ports (/device/X/ports): Core: core.router01 FastEthernet0/0 (Telco X CCID023141)
Current result in ports-Core (/iftype/type=core  : Corecore.router01 FastEthernet0/0 (Telco X CCID023141)
NEW description: core.router01 FastEthernet0/0

Use the same generate_port_link() in:
- Devices - Ports
- Ports - All/Errored/Ignored
- Ports - "Types"
- Ports - Customer

I could attach some screenshot, but I think my description is pretty clear.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Util/Url.php
+++ b/LibreNMS/Util/Url.php
@@ -135,7 +135,9 @@ class Url
         }
 
         $content = '<div class=list-large>' . addslashes(htmlentities($port->device->displayName() . ' - ' . $label)) . '</div>';
-        if ($port->ifAlias) {
+        if ($port['port_descr_descr']) {
+            $content .= addslashes(\LibreNMS\Util\Clean::html($port['port_descr_descr'], [])) . '<br />';
+        } elseif ($port['ifAlias']) {
             $content .= addslashes(htmlentities($port->ifAlias)) . '<br />';
         }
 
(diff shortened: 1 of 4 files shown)


Jellyfrog (approved):

ntrikoupis:
This seems to have broken the minigraph pages.

ulf-f:
Same here, this breaks my minigraph pages "Graphs: Bits (Mini) | Unicast Packets (Mini) | Non-Unicast Packets (Mini) | Errors (Mini)". If I revert the first commit of this PR, my minigraphs pages look like this:
Selection_558
Including this PR they look like:
Selection_559

MrMaus13:
This breaks our hover view;
We use a ":" in some of our descriptions and it now parses them through the interface description parser for this hover view.
We use interface description parsing for peering / transit / core ports to combine them in a single overview page.

If the description interface parser is used for mouse-overs, it breaks every description with a ":", while it should only parse descriptions with "Peering" and "Transit" and "Core" (in our case). It should leave the rest alone. Our mouseovers are now unreadable.

fbourqui:
#13286, fix the minigraph display issue

PipoCanaja:
@chasgames 
The full description would be restored by default in #13787 . It is indeed a pain for the NOC to do searches with this new behaviour.

@fbourqui Bonjour voisin vaudois! Could you let us know your usecase for the short description ? It seems that it would make sense to add a setting/config to move to short desc, but keep long desc as default. There is now a getDescription() function in the Port model, that would be the right place to do this.

Issue #13146, opened 2021-08-18
Syslog :: syslog_xlate does not work with hostnames with dots


vakartel:
### The problem

includes/syslog.php line 26: Config::get("syslog_xlate.{$entry['host']}") workes with dotless hostnames only.

at config.php I put an example from documentation plus two additional hosts with and without dots.
```
$config['syslog_xlate'] = array(
  'loopback0.core7k1.noc.net' => 'n7k1-core7k1',
  'loopback0.core7k2.noc.net' => 'n7k2-core7k2',
  'host-name' => 'hostname.domain.com',
  'librenms.domain.com' => 'localhost',
);
```
and a simple script to reproduce the problem
```
<?php
$init_modules = [];
require __DIR__ . '/includes/init.php';
use LibreNMS\Config;
$entry['host'] = $argv[1];
echo "--- ".Config::get("syslog_xlate.{$entry['host']}")." ---\n";
```
script output for all four hosts is:
```
$ php s.php loopback0.core7k1.noc.net
---  ---

$ php s.php loopback0.core7k2.noc.net
---  ---
$ php s.php host-name
--- hostname.domain.com ---

$ php s.php librenms.domain.com
---  ---
```

### Output of ./validate.php

```txt
====================================
Component | Version
--------- | -------
LibreNMS  | 21.7.0-72-g2bc5b15a3
DB Schema | 2021_08_04_102914_add_syslog_indexes (213)
PHP       | 7.4.22
Python    | 3.8.10
MySQL     | 10.3.31-MariaDB-0ubuntu0.20.04.1
RRDTool   | 1.7.2
SNMP      | NET-SNMP 5.8
====================================

[OK]    Composer Version: 2.1.5
[OK]    Dependencies up-to-date.
[OK]    Database connection successful
[OK]    Database schema correct
```

PR #13148, opened 2021-08-18
Fix issue #13146

vakartel:
fixed issue #13146 - Syslog :: syslog_xlate does not work with hostnames with dots

#### Please note

#### Testers

### DIFF
--- a/includes/syslog.php
+++ b/includes/syslog.php
@@ -53,7 +53,8 @@ function process_syslog($entry, $update)
     }
 
     $entry['host'] = preg_replace('/^::ffff:/', '', $entry['host']);
-    if ($new_host = Config::get("syslog_xlate.{$entry['host']}")) {
+    $syslog_xlate = Config::get('syslog_xlate');
+    if ($new_host = $syslog_xlate[$entry['host']]) {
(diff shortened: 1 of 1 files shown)


includes/syslog.php
```diff
@@ -53,7 +53,8 @@ function process_syslog($entry, $update)
     }
 
     $entry['host'] = preg_replace('/^::ffff:/', '', $entry['host']);
-    if ($new_host = Config::get("syslog_xlate.{$entry['host']}")) {
+    $syslog_xlate = Config::get('syslog_xlate');
+    if ($new_host = $syslog_xlate[$entry['host']]) {
```
murrant, line 57: This should like this to avoid errors:

```php
if (! empty($syslog_xlate[$entry['host']])) {
        $entry['host'] = $syslog_xlate[$entry['host']];
```

murrant:
Nice catch!

includes/syslog.php
vakartel: It works for me this way but ok, I changed it

Jellyfrog (approved):

Jellyfrog:
Thanks :partying_face:

PR #13155, opened 2021-08-20
Rewrite ISIS Adjacency discovery/polling

murrant:
hopefully much faster now, do most work in discovery and just check up/down in poller.

#### Please note

#### Testers


wtc-jeff:
I am game to give it a try, but as I have never installed a like this before, is it as easy as:

_scripts/github-apply 5926_

I am not sure if the example you gave above is the actual pr_id.

murrant:
@wtc-jeff the number is 13155   you can check it at the top of the page :)

wtc-jeff:
I have applied the update and ran the discovery.php vs one of the routers, it is now showing duplicate adjacencies on the device->routing->isis page. The original list of adjacencies has nothing under the Admin heading, the new adjacencies that were added have "on" under the Admin heading.  The new discovery did not find the down adjacencies on the router.

For testing, is there a way that I can clear the existing known adjacencies (mysql -> delete from isis_adjacencies)?  I do not currently have any alerts or monitoring tied to these items.

murrant:
@wtc-jeff yes, but it should delete them..... oh wait maybe not.  How did they get added in the first place I wonder.

murrant:
oh, I know, poller ran before discovery.

murrant:
@wtc-jeff run `./scripts/github-remove -d` and `./scripts/github-apply 13155` to get updated code.

Then run this in `php artisan tinker`
`IsisAdjacency::whereIn('id', IsisAdjacency::limit(99)->offset(1)->pluck('id'))->delete()`
This will delete all except the first one for testing. (assuming you have less than 99 :D)

Then let the poller run again and you should get no duplicates this time.

ottorei:
> 
> 
> I have applied the update and ran the discovery.php vs one of the routers, it is now showing duplicate adjacencies on the device->routing->isis page. The original list of adjacencies has nothing under the Admin heading, the new adjacencies that were added have "on" under the Admin heading. The new discovery did not find the down adjacencies on the router.
> 
> For testing, is there a way that I can clear the existing known adjacencies (mysql -> delete from isis_adjacencies)? I do not currently have any alerts or monitoring tied to these items.

Actually this is adding duplicates quite often. I am seeing over thousand entries on some devices now.

murrant:
@ottorei I've updated the code to fix that, are you running the latest version of this PR?

ottorei:
> 
> 
> @ottorei I've updated the code to fix that, are you running the latest version of this PR?

Updated now. I had to truncate the table. For some reason disabling the discovery module did not clear it.

ottorei:
When discovering a device without any existing adjacencies where the device has at least one downed adjacency, the discovery process crashes with this error:

Error in isis module. SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'isisISAdjNeighPriority' cannot be null (SQL: insert into `isis_adjacencies` (`device_id`, `ifIndex`, `port_id`, `isisCircAdminState`, `isisISAdjState`, `isisISAdjNeighSysType`, `isisISAdjNeighSysID`, `isisISAdjNeighPriority`, `isisISAdjLastUpTime`, `isisISAdjAreaAddress`, `isisISAdjIPAddrType`, `isisISAdjIPAddrAddress`) values (24, 1866, 447426, on, on, Array, , ?, 1, , ?, ))

Also, what is that Array there?

murrant:
@ottorei bad stuff happened when the adjacency was not present for the circuit :)

I couldn't test that, but your post was helpful. Should be fixed now.

murrant:
Except we probably shouldn't clear out the values when it is down.  And the redicovery code won't account for that.

murrant:
@ottorei ok, another fix now

ottorei:
> 
> 
> @ottorei ok, another fix now

All fields except admin, state and last uptime are empty in the DB.

murrant:
@ottorei @wtc-jeff I've adjusted the database schema again...  `lnms migrate:rollback` to remove previous migrations ONLY if you ran them with this patch before.

Then github-remove and github-apply again. and run lnms migrate.

ottorei:
PHP Fatal error:  Cannot declare class IsisAdjacenciesPortNullable, because the name is already in use in /opt/librenms/database/migrations/2021_25_01_0129_isis_adjacencies_nullable.php on line 7

murrant:
@ottorei fixed now.

murrant:
but there are other problems... I shouldn't code without testing locally :)

ottorei:
> 
> 
> but there are other problems... I shouldn't code without testing locally :)

SQLSTATE[22007]: Invalid datetime format: 1292 Truncated incorrect INTEGER value: ''

murrant:
@ottorei ok, migration is fixed now, sorry about that :(

murrant:
and one more time, phpstan found the bug causing adjacency data to always be blank.

ottorei:
> 
> 
> and one more time, phpstan found the bug causing adjacency data to always be blank.

Better, but local interface is still empty.

ottorei:
> 
> 
> > and one more time, phpstan found the bug causing adjacency data to always be blank.
> 
> Better, but local interface is still empty.

ifIndex is populated into the DB but port_id is still null.

> 
> 
> > and one more time, phpstan found the bug causing adjacency data to always be blank.
> 
> Better, but local interface is still empty.

My bad, had an SNMP interface filter on. Now it seems to populate all fields.

ottorei:
If the port does not exist in the db, this gives an error on the routing page. Should the portlink handle this error somehow? In my case, a bunch of devices have an SNMP-filter on interfaces so the interface does not exist in the DB. That is the reason I originally added the ifIndex to the adjacencies table instead of just using the port_id itself.

[2021-08-21 01:13:24] production.ERROR: Call to a member function getLabel() on null {"userId":5,"exception":"[object] (Error(code: 0): Call to a member function getLabel() on null at /opt/librenms/LibreNMS/U
til/Url.php:132)
[stacktrace]
#0 /opt/librenms/includes/html/pages/routing/isis.inc.php(93): LibreNMS\\Util\\Url::portLink()
#1 /opt/librenms/includes/html/pages/routing.inc.php(64): include('/opt/librenms/i...')
#2 /opt/librenms/app/Http/Controllers/LegacyController.php(54): require('/opt/librenms/i...')

murrant:
@ottorei Ok, fixed that.

ottorei:
Still having some errors:

[2021-08-21 16:01:38] production.ERROR: Class '\App\Models\IsIsAdjacency' not found {"exception":"[object] (Error(code: 0): Class '\\App\\Models\\IsIsAdjacency' not found at /opt/librenms/app/Observers/ModuleModelObserver.php:42)
[stacktrace]
#0 /opt/librenms/LibreNMS/Modules/Isis.php(62): App\\Observers\\ModuleModelObserver::observe()
#1 /opt/librenms/includes/discovery/isis.inc.php(30): LibreNMS\\Modules\\Isis->discover()
#2 /opt/librenms/includes/discovery/functions.inc.php(164): include('/opt/librenms/i...')
#3 /opt/librenms/discovery.php(117): discover_device()
#4 {main}
"}

murrant:
@ottorei Yep, and I don't think the old components do any harm.  I'm not sure of a clean way to remove them.  Maybe in the cleanup function (which is never called right now)

ottorei:
Old adjacencies which are still down are showing recovery and alert messages seemingly between discovery internal. Common to these devices is that the adjancecy does not have any data except the ifIndex since it was never polled. These seem to trigger these messages. The state was not changed between discovery / poll but it still recovered and alerted again. This happens with rule `isis_adjacencies.isisISAdjState != "up" AND macros.device_up = 1 AND isis_adjacencies.isisCircAdminState != "up"`

ottorei:
> 
> 
> Old adjacencies which are still down are showing recovery and alert messages seemingly between discovery internal. Common to these devices is that the adjancecy does not have any data except the ifIndex since it was never polled. These seem to trigger these messages. The state was not changed between discovery / poll but it still recovered and alerted again. This happens with rule `isis_adjacencies.isisISAdjState != "up" AND macros.device_up = 1 AND isis_adjacencies.isisCircAdminState != "up"`

This is now probably working as intended, not sure what was the issue here but at the moment all up and down stated adjacencies are updating correctly and no duplicate alerts are issued after yesterday. I have run discovery several times now without issues.

murrant:
@ottorei thanks for the update

PR #13159, opened 2021-08-21
Change automatic widget updates to use bootgrid when possible

niddey:
Changes automatic updates to use bootgrid if the widget has ajax enabled.
This will stop the widget from forgetting your search/filters and being scrolled to top on every update.

#### Please note

#### Testers


resources/views/overview/default.blade.php
```diff
@@ -609,57 +609,57 @@ function widget_settings(data) {
...
+        const $widget_body = $(`#widget_body_${id}`);
+        const $widget_bootgrid = $(`#widget_body_${id} .bootgrid-table`);
+        const settings = $widget_body.parent().data('settings') == 1 ? 1 : 0;
+
+        if (settings === 1) {
+            $(`#widget_body_${id} .bootgrid-table`).bootgrid('destroy');
+            $(`#widget_body_${id} *`).off();
         }
+
         $.ajax({
```
murrant, line 622: Why call the ajax method if you are just going to throw it away and refresh bootgrid?

murrant:
Nice change @niddey

resources/views/overview/default.blade.php
niddey: I was doing it for the title counter. With a quick search no other widget aside from alerts seems to have a counter in the title, so I have now changed the counter to be updated with the bootgrid and skip the other request.

murrant (approved):
LGTM, now we just need to change the other widgets to ajax :D

PR #13171, opened 2021-08-25
Validate APP_KEY

murrant:
Make sure APP_KEY matches on all hosts and hasn't changed since first validation
key:rotate command to rotate keys, only rotates validation data for now

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Util/EnvHelper.php
+++ b/LibreNMS/Util/EnvHelper.php
@@ -90,8 +90,9 @@ class EnvHelper
 
             $value = self::escapeValue($value);
 
-            if (strpos($content, "$key=") !== false) {
-                // only replace ones that aren't already set for safety and uncomment
+            if (preg_match("/^(#$key=$|$key=)/m", $content)) {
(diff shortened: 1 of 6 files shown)


Jellyfrog:
I'm not really sure what this adds.
First you need to set APP_KEY on your machines manually to the same value.
And if you *chose* to run this, it will see if it can decrypt a test value with the old key (which is no longer really valid anyway)

I'm guessing you're planing for support for encrypted data?

I think it would be more helpful if it worked like this:

```bash
$ lnms key:rotate

Are you sure? y/n
y

Run the following command on all machines:
lnms key:rotate --new-key=2f434f34f4..............
```

Maybe we can extend `KeyGenerateCommand`, to not duplicate too much code.

If it receives `--new-key` it we overwrite the key with the value instead of creating a new one.

murrant:
Correct, I'm adding the validation... the command is something else we will need eventually.  Also, I want a way for people to fix the validation if the key has changed.  We all know people like their green checkmarks.

Yes, we could generate a new key as part of the process.

PR #13186, opened 2021-08-30
Install new python dependencies during daily maintenance

deajan:
As of today, check_requirements simply checks for `PyMySQL` package, whereas `requirements.txt` lists a couple more dependencies.
This makes the checks dynamically read `requirements.txt`

WARNING: as of today, this would make `daily.sh` complain with something like the following on elder 1.5.x librenms setups:
```
[FAIL]  Python3 module issue found: '('Required packages:', ['PyMySQL', 'python-dotenv', 'redis>=3.0', 'setuptools', 'psutil'])
Package not found: psutil
```
I think that just like the  `composer install` in `daily.sh`, there should be a `pip3 install -r requirements` in `daily.sh` in order to make sure we update packages as we update LibreNMS.

I would be happy to modify the script if the LibreNMS dev team is okay with the above statement.

#### Please note

#### Testers


murrant (changes requested):
If you check the history of this file, it originally did this.

It was changed to just check PyMySQL because this is used to put users on the python2 support branch... 

They dynamic dependencies had a lot of false positives pushing users to the python2 support branch when they had no need of redis for example.

If you want to check this dynamically, it needs to be in a different file.  Also, LibreNMS no longer supports Python 2, so you don't need to add compatibility options.

deajan:
Thanks for the clarification.
How about we just have a static list when running python2 and read `requirements.txt` for python3 ?
Would that fit ?

murrant:
basically what you need to do is revert check_requirements.py and put your script in a new file.

daily.sh
```diff
@@ -136,8 +136,8 @@ check_dependencies() {
 
     python3=$(python3 -c "import sys;print(int(sys.version_info < (3, 4)))" 2> /dev/null)
     if [ "$python3" -eq 0 ]; then
-        python3 scripts/check_requirements.py > /dev/null 2>&1 || $(type -p python3) -m pip install -r requirements.txt > /dev/null 2>&1
```
murrant: Do not change daily.sh  I'm telling you it will break installs.
Only add your requirements check in validation.

daily.sh
```diff
@@ -133,14 +133,27 @@ check_dependencies() {
     local branch ver_56 ver_71 ver_72 ver_73 python3 python_deps phpver pythonver old_branches msg
 
     branch=$(git rev-parse --abbrev-ref HEAD)
-    scripts/check_requirements.py > /dev/null 2>&1 || pip3 install -r requirements.txt > /dev/null 2>&1
+
+    python3=$(python3 -c "import sys;print(int(sys.version_info < (3, 4)))" 2> /dev/null)
```
murrant, line 137: Don't check this in daily.sh, only in validation.

deajan: I am sorry, I don't get it. In th daily.sh version on master, we already have check_requirepents.py || pip install -r requirements.py
I just split into python2 mode that stays the same, and python3 that adds more checks.
Running pip as user does not require privileges and will install necessary packages.

What does it break exactly ? Sorry if this doesn't make sense to you :(

Jellyfrog:
My two cents, remove the downgrade functionally in daily.sh for python deps, keep only python version check.
We're not downgrading if composer fails, so why do it if python deps fail?

Other related ramblings;
daily.sh is a mess, and its hard to know what is a "validation" vs an "installation".
Basically we should try to do as little "possible breaking stuff" as possible in daily.sh

Here is how I see it in the future:
Laravel scheduler takes care of all maintenance tasks like database cleanup etc, basically all this stuff:
```
                status_run 'Updating SQL-Schema' 'php includes/sql-schema/update.php'
                status_run 'Updating submodules' "$DAILY_SCRIPT submodules"
                status_run 'Cleaning up DB' "$DAILY_SCRIPT cleanup"
                status_run 'Fetching notifications' "$DAILY_SCRIPT notifications"
                status_run 'Caching PeeringDB data' "$DAILY_SCRIPT peeringdb"
                status_run 'Caching Mac OUI data' "$DAILY_SCRIPT mac_oui"
```

A separate update.sh script should be created, only dealing with upgrades, in a more robust way.

1. Fetch changes without merging (ie. git fetch)
2. Check if we would fulfil the requirements of the new code (php, python, net-snmp, ... )
3. Switch to the code
4. Upgrade dependencies
5. Run migrations

murrant:
We added the dependency check because a lot of people were missing the python3 mysql package required for polling.  I'm fine with hardcoding that specific check (I don't think we can remove it yet, people still run crazy old versions)

murrant:
@deajan fixing this up because it is needed :)

Don't fail during, daily.sh, install python deps at the normal spot during composer install.  Use for validation too.

murrant:
originally, it was a dynamic check, but then it needed to only check pymsql for daily.sh.  Splitting them out updating the places that expected a dynamic check is the correct fix.

murrant (approved):

deajan:
Sorry for not being very active, currently moving my housing... no time for computers these weeks ;(

murrant:
No worries @deajan only a small change to merge this.

PR #13188, opened 2021-08-31
Full Python code fusion / refactor and hardening 2nd edition

deajan:

This PR is a full refactor of all the python code, which has grown to have many duplicates over time.
It replaces #13090, #13094 and #13125
The main aim is to make the codebase leaner and more maintainable, while hardening the code.
It should have no user visible effect hopefully.

## Enhancements:
- All system calls are now guaranteed to:
  - Have a timeout preventing hanging (for service and wrapper)
  - Always return an exit code
  - Have interpretable results (there's always a result, regardless of OS or script failure)
  - Work regardless of Python versions of subprocess module
- discovery, service-poller and poller wrapping is now fully logged
- Retain compatibility with earlier LibreNMS implementations
- Add an log message when too old Redis is used

## Fixes:
- Removed duplicate code for
   - Database connection functionnality
   - Config fetching
- Merged discovery, service-poller and poller wrapper code into a single wrapper
- Linting fixes

#### Please note

#### Testers


deajan:
So this is the second installment of my python code refactoring, that includes all minor environment related fixes too, that I discovered last time PR #13094 was in the wild.
The main culprit remained in the `command_runner` library that created too much threads, and also wasn't 100% reliable on non blocking stdout reading.
Interestingly, there doesn't seem to be a reliable way to read stdout in non blocking form which works with all earlier python versions (no async :( ), so I ended up reading stdout with a blocking read and use a thread just to make sure timeouts are enforced.

I'll love to have some feedback on that PR before going live again ;)

paulierco:
Tested in one of my small env with only 100 devices. Works fine.

LibreNMS/__init__.py
```diff
@@ -371,10 +525,20 @@ def put(self, item):
...
-        else:
-            item = self._redis.zpopmin(self.key)
+        try:
+            if block:
+                item = self._redis.bzpopmin(self.key, timeout=timeout)
+            else:
+                item = self._redis.zpopmin(self.key)
+        # Unfortunately we cannot use _redis.exceptions.ResponseError Exception here
+        # Since it would trigger another exception in queuemanager
+        except Exception as e:
```
murrant, line 535: I think you might catch other exceptions here and confuse the user with your specific message.  Can we make this more narrow?

deajan: I honestly tried to catch a more narrow exception, hence the comment.
How about I change the exception error message to the following ?
```
"BZPOPMIN/ZPOPMIN command failed: {}\nNote that redis >= 5.0 is required.".format(e)
```

ottorei:
Applied this patch now.

ottorei:
ICMP only devices are showing errors on discovery process. The device is up and can be polled from that particular poller.

`librenms-service.py[917322]: Discovery_3-1(INFO):Device 4177 is down, cannot discover, waiting 120s for retry`
However, it does not wait 120 seconds as the next log line immediately tells that the discovery is completed.

Should these devices even be discovered? Or maybe the logging could just say that discovering is disabled since the device is ICMP only?

Also, when I started the Python-process from multiple pollers at the same time, pollers had some lines with the following kind of error for random devices. One poller had 120 lines, one 70 lines etc. Total amount of devices is ~7700 and I have 5 pollers. After the initial startup I could no longer see these errors so it is probably nothing too bad? It came to my mind that maybe the startup of the service caused too many threads to spawn at the same time or take too much memory? However, I could not see any OOM-messages on dmesg. LibreNMS user had unlimited ulimits.

`Poller_5-22(ERROR):Polling device 5423 failed with exit code -15`

deajan:
@ottorei Thanks for the report.
Unfortunately I am not able to reproduce this by force adding an unexisting device.
Do you use the service or the wrapper ? Also, are you using distributed redis/memcache or local polling ?

As for the discovery, I didn't change the ICMP only behavior.
We could maybe tweak the SQL request to add a WHERE icmp_only = False or so.
Actually, I didn't change any librenms behavior at all, just code cleanups.

I have no idea where exit code -15 comes from.
System kill exit code would be 143 or 137 alikes.

Is there a chance I get to peek into a clone of one of your pollers maybe ?

LibreNMS/__init__.py
murrant: Perhaps just send the message of the original exception and append the note :)

murrant:
Perhaps this is returning the wrong thing in the discovery PHP code:
```php
    if ($device['snmp_disable'] == '1') {
        return false;
    }
```

LibreNMS/__init__.py
deajan: Just commited, should be good

deajan:
I've spent another week trying to work down some strange race conditions that only happened on pypy 3.7 (?), where the thread that checked the timeout did not finish in time for the process monitor to be informed that a timeout had happened.
Strange issue that only happened on github actions, unable to replicate locally.
Anyway, I'm pretty sure the subprocess calls work good.

@ottorei Could you update the patch and report back please ?

deajan:
Since the last 3 commits, all tests that succeed now fail with missing `PyMySQL` and `redis` python packages, but I didn't change anything related. @murrant could you please help me here ? I'm kinda lost

ottorei:
> 
> 
> I've spent another week trying to work down some strange race conditions that only happened on pypy 3.7 (?), where the thread that checked the timeout did not finish in time for the process monitor to be informed that a timeout had happened.
> Strange issue that only happened on github actions, unable to replicate locally.
> Anyway, I'm pretty sure the subprocess calls work good.
> 
> @ottorei Could you update the patch and report back please ?

Sure, I'll update my servers tomorrow with the latest commits.

deajan:
@ottorei Any news ?

ottorei:
> 
> 
> @ottorei Any news ?

Sorry for the delay, just updated today. The new code has been running a few hours now without noticeable issues. I will let it run for some time and check the journalctl logs for anything interesting.

paulierco:
I've been tested this for my dev server with 100 devices and i didn't had any issues.

murrant:
@deajan ready to try to merge it?

deajan:
@murrant I'm on the road right now.
If we schedule monday, I can keep an eye on the discord channel. Could this be okay ?

ottorei:
> 
> 
> @murrant I'm on the road right now.
> If we schedule monday, I can keep an eye on the discord channel. Could this be okay ?

I have been running this for the past few days now on a 5 server cluster with ~8000 devices. I checked the journalctl logs of the poller service on multiple servers matching error lines - did not see any noticeable issues other than planned service breaks.

Good job ;)

deajan:
@ottorei Thanks for the feedback

@murrant  I'll stick toi the discord this evening (CEST+1) which should be the time you are online usually

murrant (approved):

PR #13192, opened 2021-09-03
ZTE ZXR10 define discovery for mempool and processor

vakartel:
Add mempool and processor discovery for ZTE ZXR10

#### Please note

#### Testers


vakartel:
I added memory and processor definitions like described in documentation
https://docs.librenms.org/Developing/os/Mem-CPU-Information

But still have no success with graphing.
I rediscovered my zxr10 device dozen times, but it doesn't draw memory graph and shows "No Processors" in processor graph. :(

includes/definitions/zxr10.yaml
```diff
@@ -3,6 +3,11 @@ group: zxr10
 text: 'ZTE ZXR10'
 type: network
 icon: zte
+mib_dir: zte
```
murrant, line 6: These are the defaults, you don't need to set them.

includes/definitions/discovery/zxr10.yaml
```diff
@@ -0,0 +1,13 @@
+mib: ZXR10-MIB
+modules:
+    mempools:
+        data:
+            -
+                total: ZXR10-MIB::zxr10SystemMemSize.1
+                percent_used: ZXR10-MIB::zxr10SystemMemUsed.1
```
murrant, line 7: The mib says zxr10SystemMemUsed is in bytes not percent.  so you should use "used:" instead.
You should also remove the precision line below.

murrant:
If you can generate test data, it would help as that enables us to test locally.

includes/definitions/discovery/zxr10.yaml
vakartel: I thought so too, but:
$ snmpwalk ....... zxr10SystemMemSize
ZXR10-MIB::zxr10SystemMemSize.1 = INTEGER: 268435456
mem-size - 256M in bytes

$ snmpwalk ...... zxr10SystemMemUsed
ZXR10-MIB::zxr10SystemMemUsed.1 = INTEGER: 60
mem-used is in percent not in bytes

zxr10 device cli shows the same:
#show processor 
PhyMem: Physical memory (megabyte)

    Panel    CPU(5s) CPU(1m) CPU(5m)  PhyMem  Buffer  Memory
           
MP(M)   1      47%     48%     48%      256     0%      59.958%

vakartel: I removed the precision line and now memory draws correctly
And now memory oids collects without .1 index.

includes/definitions/zxr10.yaml
vakartel: I tried remove it and catched mib-not-found errors while run discovery.php

MIB search path: /opt/librenms/mibs
Cannot find module (ZXR10-MIB): At line 9 in (none)
zxr10SystemCpuUtility5m: Unknown Object Identifier (Sub-id not found: (top) -> zxr10SystemCpuUtility5m)
 
MIB search path: /opt/librenms/mibs
Cannot find module (ZXR10-MIB): At line 9 in (none)
ZXR10-MIB::zxr10SystemMemSize: Unknown Object Identifier 

So it must be there

vakartel:
I removed the precision line and now memory and processor collects and draws correctly!

But I can't collect test data from my zxr10 device.
My existing and running zxr10 device id is 19, so I run:
$ ./scripts/collect-snmp-data.php -h 19

and script sayed:   You must specify an existing device to collect data from.
:(

vakartel:
Tests added

includes/definitions/zxr10.yaml
```diff
@@ -3,6 +3,11 @@ group: zxr10
 text: 'ZTE ZXR10'
```
murrant, line 3: OH, here is the problem, group is zxr10 instead of zte.  There are not any other devices in the zxr10 group and no

Looking closer, the only real group is zyxel... which just adds fallback to ZYXEL-ES-COMMON for os information.

Just delete group here and leave mib_dir to zte.

You can set group to zyxel if it adds hardware/version/serial number.

includes/definitions/discovery/zxr10.yaml
murrant: Good job following your own mib ZTE :D  Looks good.

includes/definitions/zxr10.yaml
vakartel: zyxel - is a vendor
zte - is an another vendor. is's not the same to zyxel

zxr10 is zte's product line of L3 switches

in librenms there is another product line from zte called zxdsl (ADSL modems/routers I think)

So I think it needs to just change zxr10 group to zte

murrant: Oops, too early in the morning.  I agree with this fix :)

murrant (approved):

PR #13199, opened 2021-09-07
Modified SQL delete statement for vrf is null

kterobinson:
I modified the SQL insert and delete statements in both discovery and polling for bgp peers module.  We were running into an issue where peers were being inserted with vrf ID of null which isn't possible on a Nokia device.  The delete statement in the discovery code to clean up peers needed to be is NULL instead of '= null'.  I also checked during polling if the os=nokia and vrf=null, don't add.

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/bgp-peers/timos.inc.php
+++ b/includes/discovery/bgp-peers/timos.inc.php
@@ -46,6 +46,7 @@ if (Config::get('enable_bgp')) {
                 $astext = get_astext($value['tBgpPeerNgPeerAS4Byte']);
 
                 if (dbFetchCell('SELECT COUNT(*) from `bgpPeers` WHERE device_id = ? AND bgpPeerIdentifier = ? AND vrf_id = ?', [$device['device_id'], $address, $vrfId]) < '1') {
+                    echo 'dbFetchCell is less than 1';
                     $peers = [
                         'device_id' => $device['device_id'],
                         'vrf_id' => $vrfId,
@@ -70,6 +71,7 @@ if (Config::get('enable_bgp')) {
                     }
                     echo '+';
                 } else {
+                    echo 'Updating BGP peer';
                     dbUpdate(['bgpPeerRemoteAs' => $value['tBgpPeerNgPeerAS4Byte'], 'astext' => $astext], 'bgpPeers', 'device_id = ? AND bgpPeerIdentifier = ? AND vrf_id = ?', [$device['device_id'], $address, $vrfId]);
                     echo '.';
                 }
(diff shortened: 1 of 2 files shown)


includes/discovery/bgp-peers/timos.inc.php
```diff
@@ -83,11 +85,22 @@
...
+                    $deleted = dbDelete('bgpPeers', 'device_id = ? AND bgpPeerIdentifier = ? AND vrf_id is NULL', [$device['device_id'], $address]);
+                } else {
+                    $deleted = dbDelete('bgpPeers', 'device_id = ? AND bgpPeerIdentifier = ? AND vrf_id = ?', [$device['device_id'], $address, $vrfId]);
+                }
                 echo str_repeat('-', $deleted);
                 echo PHP_EOL;
             }
+
+            // deletes any duplicates with same device id, same peer address and same vrfid
+            $duplicatecount = dbFetchCell('SELECT COUNT(*) from `bgpPeers` WHERE device_id = ? AND bgpPeerIdentifier = ? AND vrf_id = ?', [$device['device_id'], $address, $vrfId]);
```
murrant, line 99: Why would there be duplicates?   This troubles me.

drshawnkwang: I will help answer as I worked on the debugging of this problem. We think that discovery of bgppeers adds rows to the database with `vrfID=NULL` . Why these row were being inserted was not found. We couldn't "catch it in the act". But we do now that it doesn't make sense for the Nokia routers to have vrfID to be NULL in the first place.

This lead to the "fix" of deleting these errant rows.

Thus in some ways this row is not a duplicate. It is "wrong data" with the same bdgPeer identifier IP address that is being cleaned up.

murrant: It should be inserting null not "NULL", the SQL is wrong... One of the downsides of using the ancient db* functions.

To actually insert null, you need to send ['NULL'] as the update parameter. (the string null inside an array)
Or you could just change these to Eloquent queries, but at point, you might as well re-write this as a modern class based module as that would be less work.

drshawnkwang: So the problem is not that the code does not correctly insert NULL, the real problem is that the polling/discovery code for bgppeering should never insert into the database a row where `vrfid` is null. (Also this id is the primary key for the `vrfs` table which the schema is "int(10) unsigned and NULL=no").

Again we don't know why these corrupt data get inserted, but the code change in the PR is to clean this up if found.  I realize it would be ideal to refactor this code, but given the constraints we are under, this bug fix is what we use and wish to contribute to the upstream project.

drshawnkwang:
> Your PR must be tested against other vendors, before being merged, to ensure it does not break things.

@PipoCanaja Regarding tests, what do you need to see? 

FTR: we are running this in our libreNMS instance (custom code) with Nokia, Juniper, Cienna, Arista, etc. so I'm fairly confident that it won't affect other vendors, but what do you need to from us in terms of testing?

includes/discovery/bgp-peers/timos.inc.php
murrant: vrf_id should be null in the bgpPeers table when the route is in the default context.

It is not clear that you understand "NULL" and NULL are not the same thing.  Can you please confirm that?
If you look in the database and it says NULL, that means it is a string and not an empty value (null).

It is acceptable to not want to refactor, but I'm having a hard time confirming your bug fix is correct.

drshawnkwang: > It is not clear that you understand "NULL" and NULL are not the same thing. Can you please confirm that?

I am man enough to know when I admit when I am ignorant: I did not understand that you were writing about two different values: a 'null' string and a NULL value. 

However, I believe the values we have in our database that are duplicates are NULL value and not 'null' string. I based this because I ran two different SQL statements, below:

```
# returns nothing
SELECT * FROM bgpPeers WHERE device_id=3697 AND bgpPeerIdentifier="10.10.10.10" AND vrf_id="null";

# returns 4 rows
SELECT * FROM bgpPeers WHERE device_id=3697 AND bgpPeerIdentifier="10.10.10.10" AND vrf_id IS NULL;
```

Your comment above "It should be inserting null not "NULL", the SQL is wrong" now makes more sense.

drshawnkwang: It seems that this line (86) from `timos.inc.php` is the problem? If `$vrfID` is a NULL value then it will not deleting anything. 

```
$deleted = dbDelete('bgpPeers', 'device_id = ? AND bgpPeerIdentifier = ? AND vrf_id = ?', [$device['device_id'], $address, $vrfId]);
```

PipoCanaja: Yes, this seems to be the culprit. 

You can solve this by checking $vrfID and doing 2 different dbDelete depending if vrfID is null or not. Just because `null` in PHP is translated to `''` (empty string) which is not equal to `NULL` in SQL ... 

This check is done in the `vrp.inc.php` bgp implementation, you can have a look at line 142 of includes/discovery/bgp-peers/vrp.inc.php.  Basically, a new variable is created to do the vrf check, and is modified depending if vrfID is null or not.

drshawnkwang: Thanks to you both for the explanations. I'll work with Kate on using the `vrp.inc.php` code as a guide to write the "clean up peers" part of `timos.inc.php` and update this PR.

drshawnkwang:
We have been running with this custom code and feel it works. We even saw a case where it removed a number of duplicate vrfs from the `bgpPeers` table. Is there anything else that is needed for this to be merged?

PR #13212, opened 2021-09-09
Added device group type check to rule validation logic for add_device_group API endpoint.

DanielMuller-TN:
A simple fix for the error I was encountering with submitting statically assigned device groups via the API. I've placed the dynamic rule logic validation within a conditional check so it only executes for device groups of type 'dynamic'. It's pretty low hanging fruit, sorry.

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1944,14 +1944,17 @@ function add_device_group(Illuminate\Http\Request $request)
         return api_error(422, $v->messages());
     }
 
+    $deviceGroup = DeviceGroup::make(['name' => $data['name'], 'type' => $data['type'], 'desc' => $data['desc']]);
+
     // Only use the rules if they are able to be parsed by the QueryBuilder
-    $query = QueryBuilderParser::fromJson($data['rules'])->toSql();
-    if (empty($query)) {
-        return api_error(500, "We couldn't parse your rule");
+    if ($data['type'] == 'dynamic') {
+        $query = QueryBuilderParser::fromJson($data['rules'])->toSql();
+        if (empty($query)) {
+            return api_error(500, "We couldn't parse your rule");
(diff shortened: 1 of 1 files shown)


includes/html/api_functions.inc.php
```diff
@@ -1944,14 +1944,17 @@ function add_device_group(Illuminate\Http\Request $request)
         return api_error(422, $v->messages());
     }
 
+    $deviceGroup = DeviceGroup::make(['name' => $data['name'], 'type' => $data['type'], 'desc' => $data['desc']]);
+
     // Only use the rules if they are able to be parsed by the QueryBuilder
-    $query = QueryBuilderParser::fromJson($data['rules'])->toSql();
-    if (empty($query)) {
-        return api_error(500, "We couldn't parse your rule");
+    if ($data['type'] == 'dynamic') {
```
murrant, line 1950: Hmm, actually static groups can have rules, they are just unused until you change the type to dynamic.  We should check `if (! empty($data['rules'])) {` instead of the type

DanielMuller-TN: @murrant - Absolutely, if the desired behaviour is to allow the definition of rules for static groups then wrapping the QueryBuilderParser::fromJson() call and accompanying api_error validation in the conditional you've suggested would be the way to go. I'll drop this pull request, test it and submit a new one with the change included.

murrant: You don't need to close this, just update it :)

murrant:
you messed up your branch a bit, try doing a rebase against upstream master.

DanielMuller-TN:
> you messed up your branch a bit, try doing a rebase against upstream master.

Yeah, I see that - the noob strikes again. Sorry.. monkey behind the keyboard. Fixing it now.

PR #13213, opened 2021-09-09
IOS/IOSXE PoE stats

martinberg:

#### Please note

#### Testers


includes/definitions/discovery/ios.yaml
```diff
@@ -0,0 +1,28 @@
...
+                    index: 'pethMainPsePower.{{ $index }}'
+                    group: "PoE"
+                    descr: "PoE Budget Total - ID {{ $index }}"
+                -
+                    oid: pethMainPseTable
+                    value: pethMainPseConsumptionPower
+                    num_oid: '.1.3.6.1.2.1.105.1.3.1.1.4.{{ $index }}'
+                    index: 'pethMainPseConsumptionPower.{{ $index }}' 
+                    group: "PoE"
+                    descr: "PoE Budget Consumed - ID {{ $index }}"
```
mpikzink, line 19: what do you think about ```setting high_limit: pethMainPsePower``` instead of as an extra sensor? I'm currently working on the PoE sensor data from Procurve and will do the same.

martinberg: Good idea! But two things came to my mind

What would happen if you added another PSU to the switch?
One PSU gives you 740W PoE Budget and adding another one gives you a total budget of 1480W.
`pethMainPseConsumptionPower` gets the consumption from each PSU and `pethMainPsePower` for the whole switch.
`pethMainPseConsumptionPower` would never reach the switch total budget due to being bound to one PSU.
Haven't confirmed this myself so needs to be tested.

If you send your Librenms data to ie Influxdb you should be able to calculate the total PoE budget. For this to be possible with IOS `pethMainPsePower` has to be a separate sensor due to missing consumed and remaining.
With IOSXE `pethMainPsePower` could be dropped as a separate sensor but I feel that consistency is to be preferred here. Or maybe a per switch total PoE budget is needed in some case I can't come up with right now.

Thanks for your input!

mpikzink: Thanks for the detailed description of your concerns. I will also need to test this for my sensors.

Making the evaluation in one with Influx or Grafana is of course a possibility. But it would be nicer to have everything directly in LNMS and thus be able to use the alerts. Not everyone has such a big setup and then they could not benefit from your work.

I don't think that the power of the PSUs is added because the MIB table is the same as Consumed. If that were the case, you would also have the problem that two sensors would be displayed with the same power (1440W). We should really test this. I will try it for HP Procurve. Since we don't have switches with two power supplies, I would have to order one first, so it will take some time.

I think my proposed way is not perfect yet, because you don't see the maximum power directly in the overview, but the display could be customizable in the future. 212/720W for example.

Translated with www.DeepL.com/Translator (free version)

martinberg: I agree that it would be nice to have it all in LNMS and hopefully it is possible to implement.

In my case the increased PoE budget is added to PSU A and no extra entry is created in as `pethMainPseConsumptionPower`.

```
#PSU A and B inserted
[martinberg@comp ~]$ snmpwalk -v 2c -c community sw01  .1.3.6.1.2.1.105.1.3.1
SNMPv2-SMI::mib-2.105.1.3.1.1.2.1 = Gauge32: 1480
SNMPv2-SMI::mib-2.105.1.3.1.1.3.1 = INTEGER: 1
SNMPv2-SMI::mib-2.105.1.3.1.1.4.1 = Gauge32: 0
SNMPv2-SMI::mib-2.105.1.3.1.1.5.1 = INTEGER: 80
[martinberg@comp ~]$ snmpwalk -v 2c -c community sw01  .1.3.6.1.4.1.9.9.402.1.3
SNMPv2-SMI::enterprises.9.9.402.1.3.1.1.1 = INTEGER: 1017
SNMPv2-SMI::enterprises.9.9.402.1.3.1.2.1 = STRING: "Switch 1 - Power Supply A"
SNMPv2-SMI::enterprises.9.9.402.1.3.1.3.1 = INTEGER: 1
SNMPv2-SMI::enterprises.9.9.402.1.3.1.4.1 = Gauge32: 0
SNMPv2-SMI::enterprises.9.9.402.1.3.1.5.1 = Gauge32: 1480000

#PSU A inserted
[martinberg@comp ~]$ snmpwalk -v 2c -c community sw01  .1.3.6.1.2.1.105.1.3.1
SNMPv2-SMI::mib-2.105.1.3.1.1.2.1 = Gauge32: 740
SNMPv2-SMI::mib-2.105.1.3.1.1.3.1 = INTEGER: 1
SNMPv2-SMI::mib-2.105.1.3.1.1.4.1 = Gauge32: 0
SNMPv2-SMI::mib-2.105.1.3.1.1.5.1 = INTEGER: 80
[martinberg@comp ~]$ snmpwalk -v 2c -c community sw01  .1.3.6.1.4.1.9.9.402.1.3
SNMPv2-SMI::enterprises.9.9.402.1.3.1.1.1 = INTEGER: 1017
SNMPv2-SMI::enterprises.9.9.402.1.3.1.2.1 = STRING: "Switch 1 - Power Supply A"
SNMPv2-SMI::enterprises.9.9.402.1.3.1.3.1 = INTEGER: 1
SNMPv2-SMI::enterprises.9.9.402.1.3.1.4.1 = Gauge32: 0
SNMPv2-SMI::enterprises.9.9.402.1.3.1.5.1 = Gauge32: 740000

#PSU B inserted
[martinberg@comp ~]$ snmpwalk -v 2c -c community sw01  .1.3.6.1.2.1.105.1.3.1
SNMPv2-SMI::mib-2.105.1.3.1.1.2.1 = Gauge32: 740
SNMPv2-SMI::mib-2.105.1.3.1.1.3.1 = INTEGER: 1
SNMPv2-SMI::mib-2.105.1.3.1.1.4.1 = Gauge32: 0
SNMPv2-SMI::mib-2.105.1.3.1.1.5.1 = INTEGER: 80
[martinberg@comp ~]$ snmpwalk -v 2c -c community sw01  .1.3.6.1.4.1.9.9.402.1.3
SNMPv2-SMI::enterprises.9.9.402.1.3.1.1.1 = INTEGER: 1020
SNMPv2-SMI::enterprises.9.9.402.1.3.1.2.1 = STRING: "Switch 1 - Power Supply B"
SNMPv2-SMI::enterprises.9.9.402.1.3.1.3.1 = INTEGER: 1
SNMPv2-SMI::enterprises.9.9.402.1.3.1.4.1 = Gauge32: 0
SNMPv2-SMI::enterprises.9.9.402.1.3.1.5.1 = Gauge32: 740000
```

martinberg: I tried adding a `high_limit` but was unsuccessful. Of all of the following the only one that actually set a limit was `high_limit: 30` or another static number.

Tried the following
``` 
high_limit: pethMainPsePower
high_limit: POWER-ETHERNET-MIB:pethMainPsePower
high_limit: 'POWER-ETHERNET-MIB:pethMainPsePower.{{ $index }}'
```
Might be a restriction for what you can input with variables or maybe somebody knows what I'm doing wrong?
I believe an index is needed in case the switches in the stack have different PoE budgets and/or the PSUs have uneven Watts

ottorei: @murrant Can these values be dynamic?

murrant: The value wither needs to be in the table you are walking or pre-fetched.

martinberg: The value I wanted to use is not in the table and when I tried pre-fetching the values I couldn't get the indexing to work.
If somebody can think of a way to solve this feel free to post a comment or fix in their own PR.
I didn't manage to set a `high_limit` for ios but I set a `low_limit` for iosxe. 

Feel free to merge this if everything looks good

martinberg:
> Isn't this a duplicate of existing poe functionality. see port-poe.inc.php

'port-poe.inc.php' seems to be for per-port. This is the total for the whole switch/stack.

murrant:
Seems like you need to re-create the test data.  Your PR includes value changes on things that should not be affected by your changes.

martinberg:
> 
> 
> Seems like you need to re-create the test data. Your PR includes value changes on things that should not be affected by your changes.

I guess you are referring to the 2960x json and the changes to `entPhysicalIndex` and some Stack Port Status.
Should I manually edit the json to not include those changes? Or is there a better way?
I don't know a way to include only the changes from the snmprec I created when I later run `./scripts/save-test-data.php -o ios -v 2960x -m sensors`.

murrant:
Do you have other code applied when collecting the data?

murrant:
Oh, the code incorrectly depends on the port module, so we need to add manually that when generating new test data.

PR #13226, opened 2021-09-10
Feature: Add opengear humidity to sensors

drshawnkwang:
Added to opengear.yaml file the OID for the environmental humidity sensor.  

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/discovery/opengear.yaml
+++ b/includes/definitions/discovery/opengear.yaml
@@ -27,7 +27,7 @@ modules:
                 -
                     oid: ogEmdHumidityTable
                     value: ogEmdHumidityValue
-                    num_oid: '.1.3.6.1.4.1.25049.17.10.1.5.1.{{ $index }}'
+                    num_oid: '.1.3.6.1.4.1.25049.17.10.1.5.{{ $index }}'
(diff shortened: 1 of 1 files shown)


includes/definitions/discovery/opengear.yaml
```diff
@@ -13,6 +13,7 @@ modules:
                     oid:
                         - ogCellModemTable
                         - ogEmdTemperatureTable
+                        - ogEmdHumidityTable
```
murrant, line 16: Don't need this it is only used by one sensor, having it here will only make things slower.

murrant: Actually, what are any of these doing here?

murrant: yeah, please remove the whole pre-cache entry.

murrant: ogEmdTemperatureTable isn't needed either...

murrant (changes requested):
Looks good, just need to remove the pre-cache entry and generate test data.  Check the docs linked by the bot.

murrant:
The snmprec file is still missing

PR #13233, opened 2021-09-13
Poller performance improvement

eskyuu:
Hi,

I've just started using librenms and noticed that the poller wrapper forks a new process for every device polled.  I also noticed that the CPU usage is not smooth (the poller gets more CPU bound at the end of the polling cycle as the devices respond quicker).

I was able to get a 2x improvement in the total polling speed after applying these patches and running the poller-wrapper 3 or 4 times.  It is important to do these 3 or 4 poller executions so the wrapper learns the new polling durations and correctly splits the work queues in half.  We are polling around 1,000 devices with a polling duration of between 1 and 140 seconds per device, so I appreciate that my results may differ from other people.

I have tested the code on our system, and it correctly handles the errors that I came across during testing.

#### Please note

#### Testers

### DIFF
--- a/poller-wrapper.py
+++ b/poller-wrapper.py
@@ -153,7 +153,7 @@ def printworker():
 """
 
 
-def poll_worker():
+def poll_worker(poll_queue):
     while True:
         device_id = poll_queue.get()
         # (c) 2015, GPLv3, Daniel Preussker <f0o@devilcode.org> <<<EOC5
@@ -309,7 +309,8 @@ if __name__ == "__main__":
     except (IndexError, ValueError):
         amount_of_workers = 16
 
-    devices_list = []
+    slow_devices_list = []
+    fast_devices_list = []
 
     """
         This query specificly orders the results depending on the last_polled_timetaken variable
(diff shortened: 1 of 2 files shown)


eskyuu:
I just noticed the bad formatting of the PHP script.  I will fix and re-submit.

paulierco:
Maybe you can have a look to this ? https://github.com/librenms/librenms/pull/13188

eskyuu:
I'll have a look - thanks.

I tracked down my skipped polls issue - I needed to call Config::reload() in between each poll.  Unfortunately this resulted in the poll time increasing back to the previous level.  My patch does appear to reduce CPU usage on our host by about 10%, which I can then compare against the pull request that you've sent after I have confirmed that my newest patch is 100% working.

deajan:
@eskyuu I have written #13188 and would be glad to have your feedback on my PR.
I've refactored most of the wrapped poller code, merging the poller, discovery and service-check codes into a single wrapper, but for other reasons than performance.
Maybe you could try to apply your patch on my fork which has yet to get upstreamed, and see whether it will increase the performance as a whole ?

PR #13248, opened 2021-09-17
Infer character encoding for ifAlias and sysLocation

murrant:
Tries to convert character encoding for non-UTF-8 encoded strings.
This will only work for snmp strings that are type-hinted, not ones forced to ASCII with -Oa or similar
Only works for your default charset or Windows-1251 or LATIN1.
You can set your character encoding in .env with CHARSET

**Please save us all and just use UTF-8**

Thanks to symfony/var-dumper for the hints.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Util/StringHelpers.php
+++ b/LibreNMS/Util/StringHelpers.php
@@ -117,7 +117,7 @@ class StringHelpers
             return (string) $converted;
         }
 
-        if ($charset !== 'CP850' && ($converted =  @iconv('Windows-1252', 'UTF-8', $string)) !== false) {
+        if ($charset !== 'CP850' && ($converted =  @iconv('CP850', 'UTF-8', $string)) !== false) {
             return (string) $converted;
         }
 
--- a/LibreNMS/Util/StringHelpers.php
+++ b/LibreNMS/Util/StringHelpers.php
@@ -103,7 +103,7 @@ class StringHelpers
      */
     public static function inferEncoding(?string $string): ?string
     {
-        if (empty($string) || ctype_print($string) || ! function_exists('iconv')) {
+        if (empty($string) || preg_match('//u', $string) || ! function_exists('iconv')) {
             return $string;
         }
 
--- a/LibreNMS/Util/StringHelpers.php
+++ b/LibreNMS/Util/StringHelpers.php
@@ -117,6 +117,12 @@ class StringHelpers
             return (string) $converted;
         }
 
-        return (string) iconv('CP850', 'UTF-8', $string);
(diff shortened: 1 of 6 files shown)


LibreNMS/Util/StringHelpers.php
```diff
@@ -93,4 +93,36 @@ public static function camelToTitle($string)
...
+            return (string) $converted;
+        }
+
+        if ($charset !== 'Windows-1252' && $converted = @iconv('Windows-1252', 'UTF-8', $string) !== false) {
+            return (string) $converted;
+        }
+
+        if ($charset !== 'CP850' && $converted =  @iconv('Windows-1252', 'UTF-8', $string) !== false) {
+            return (string) $converted;
+        }
```
Jellyfrog, line 122: Test fails due to $converted is set to the result of false test.

```suggestion
        if (($converted = @iconv($charset, 'UTF-8', $string)) !== false) {
            return (string) $converted;
        }

        if ($charset !== 'Windows-1252' && ($converted = @iconv('Windows-1252', 'UTF-8', $string)) !== false) {
            return (string) $converted;
        }

        if ($charset !== 'CP850' && ($converted =  @iconv('Windows-1252', 'UTF-8', $string)) !== false) {
            return (string) $converted;
        }
```

LibreNMS/Util/StringHelpers.php
```diff
@@ -93,4 +93,36 @@ public static function camelToTitle($string)
...
+
+        if ($charset !== 'Windows-1252' && ($converted = @iconv('Windows-1252', 'UTF-8', $string)) !== false) {
+            return (string) $converted;
+        }
+
+        if ($charset !== 'CP850' && ($converted =  @iconv('Windows-1252', 'UTF-8', $string)) !== false) {
+            return (string) $converted;
+        }
```
Jellyfrog, line 122: these 2 can be combined?

murrant: It is a typo

kimhaak:
Tested and it fixed problems for me.

kimhaak:
Seeing same issue on "sysname" - would that be considered a input like ifalias and location?

murrant:
@kimhaak yes, feel free to send a Pull Request for that :)

kimhaak:
> @kimhaak yes, feel free to send a Pull Request for that :)

ok, but i'll need a hint. Pushed the sysname string trough the helper function in includes/polling/core.inc.php - that helped when the poller ran. But when discovery runs, it changes from the correct string, to the wrong string, and then, back again.

Seems like discovery is asking for sysname from somewhere else (at one point at least) than includes/polling/core.inc.php



image

murrant:
https://github.com/librenms/librenms/blob/master/includes/discovery/core.inc.php#L12

PR #13257, opened 2021-09-21
Workaround for dashboard widgets showing over popups.

murrant:
z-index is being set on the popup, and I verified it is applied, but for some reason the widget with a z-index of 2 still shows over top.
I don't think this has any side-affects, but I don't know for sure.
Could not find another work-around.  Gridster is dead, so unlikely we will update the js and override this fix.

FYI, change is remove z-index from this class:
```css
  .gs-w {
    z-index: 2;
    position: absolute;
  }
```

Before:
Screenshot_20210921_093154

After:
Screenshot_20210921_095129

#### Please note

#### Testers


paulierco:
For me still doesn't work. Do I need some cleanup ?

murrant:
@paulierco I did not update the js file hash, so you need to force your browser to refresh the asset (CTRL-F5)

paulierco:
Thanks. I can confirm it works.

Issue #13253, opened 2021-09-20
Outages page does not use the user configurable time zone


ottorei:
### The problem

Outages pages currently show all outages as UTC time instead of user configured time zone. Should use the same time zone as other pages.

### Output of ./validate.php

```txt
====================================
Component | Version
--------- | -------
LibreNMS  | 21.1.0-733-g00b24d9f1
DB Schema | 2021_08_26_093522_config_value_to_medium_text (217)
PHP       | 7.4.3
Python    | 3.8.10
MySQL     | 10.3.31-MariaDB-0ubuntu0.20.04.1
RRDTool   | 1.7.2
SNMP      | NET-SNMP 5.8
====================================

[OK]    Composer Version: 2.1.8
[OK]    Dependencies up-to-date.
[OK]    Database connection successful
[OK]    Database schema correct
[INFO]  Detected Dispatcher Service
[WARN]  Your install is over 24 hours out of date, last update: Mon, 13 Sep 2021 21:43:27 +0000
        [FIX]:
        Make sure your daily.sh cron is running and run ./daily.sh by hand to see if there are any errors.
[WARN]  Your local git branch is not master, this will prevent automatic updates.
        [FIX]:
        You can switch back to master with git checkout master
```

ottorei:
This has been fixed in #13274

PR #13274, opened 2021-09-25
Use local timezone for outages pages

ottorei:
When viewing outages from Web UI, all the rows are converted to human readable timestrings using Carbon. However, the timezone is still UTC and not the correct local timezone of the PHP. Outages table uses Unix epoch timestamps in the database and not MySQL / MariaDB timestamps.

Tested the fix with Europe/Helsinki timezone and it seems to output correct values.

Fixes issue #13253 

#### Please note

#### Testers


SourceDoctor (approved):
LGTM

murrant:
Hmm, should we be sending utc to the browser and convert to the browser's timezone in javascript?

PR #13307, opened 2021-10-01
Remove text about report sql issues

Jellyfrog:
This is a bit controversial maybe, but the origin of that text was during the mess of `.sql` migrations.
Many people blindly just post this text on the forum because it's telling them too, and it's not really of any help.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Validations/Database.php
+++ b/LibreNMS/Validations/Database.php
@@ -328,8 +328,8 @@ class Database extends BaseValidation
         if (empty($schema_update)) {
             $validator->ok('Database schema correct');
         } else {
-            $result = ValidationResult::fail('We have detected that your database schema may be wrong, please report the following to us on Discord (https://t.libren.ms/discord) or the community site (https://t.libren.ms/5gscd):')
-                ->setFix('Run the following SQL statements to fix.')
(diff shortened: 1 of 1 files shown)


LibreNMS/Validations/Database.php
```diff
@@ -328,8 +328,8 @@ private function checkSchema(Validator $validator)
         if (empty($schema_update)) {
             $validator->ok('Database schema correct');
         } else {
-            $result = ValidationResult::fail('We have detected that your database schema may be wrong, please report the following to us on Discord (https://t.libren.ms/discord) or the community site (https://t.libren.ms/5gscd):')
-                ->setFix('Run the following SQL statements to fix.')
```
murrant, line 332: How about add on most distros you can access mysql with `mysql -u root librenms`?

Jellyfrog: I was thinking of updating it later so it's copy-paste `lnms tinker --execute` commands

Jellyfrog: that way the database connection/selection is already included.
We could just add a button from the webui, but feels a bit... hmm.. prone to breakage

PR #13308, opened 2021-10-01
Use PHPStan level 6

Jellyfrog:
Instead of using level 6 for new files only, lets use level 6 for everything and include a massive baseline :)
In the long run we want to reach level 6 anyway, so let's start now.

*Please note this will cause some friction since everything will need to be type-hinted*

#### Please note

#### Testers


murrant:
How hard is it to update the baseline?

Jellyfrog:
./vendor/bin/phpstan analyze --generate-baseline=phpstan-baseline.neon

murrant:
Fine with this, but not sure about when to merge, probably after the next release would be best.

Jellyfrog:
Since it probably will block a lot of prs, we can wait. Need to rebase before merge tho

PR #13334, opened 2021-10-05
Add a auto fix button for database validation problems 

Jellyfrog:

#### Please note

#### Testers


murrant:
Thinking about adding a prompt on the CLI too?

app/Http/Controllers/ValidateController.php
```diff
@@ -0,0 +1,30 @@
...
+class ValidateController extends Controller
+{
+    public function index(): View
+    {
+        $validator = new Validator();
+        $validator->validate();
+
+        $short_size = 10;
+
+        return view('validate.index', compact('validator', 'short_size'));
```
murrant, line 19: I've been favoring inline multi-line array  for $data of late.  At least when things are short.

One thing that has bothered my about the validation page is that is is not ajax. :D  Not bothered if you don't implement that in this page.

murrant:
@Jellyfrog ok, after pondering a bit, it feels like the fix should belong to the validation result, not the validator.  Each result could have a different fix. So, when $validator->fail() is called, it should be set on the validation result.

Ideally convert it to a fluent interface:
```php
$validator->fail('It broke')
    ->fixText('upgrade the thing')
    ->fix(function () { fixIt(); });
```

or more likely:
```php
$validator->fail('It broke')
    ->fixText('upgrade the thing')
    ->fix([$this, 'fixIt']);
```

Jellyfrog:
> @Jellyfrog ok, after pondering a bit, it feels like the fix should belong to the validation result, not the validator. Each result could have a different fix. So, when $validator->fail() is called, it should be set on the validation result.
> 
> Ideally convert it to a fluent interface:
> 
> ```
> $validator->fail('It broke')
>     ->withFixText('upgrade the thing')
>     ->withFix(function () { fixIt(); });
> ```
> 
> or more likely:
> 
> ```
> $validator->fail('It broke')
>     ->withFixText('upgrade the thing')
>     ->withFix([$this, 'fixIt']);
> ```

I'm probably missing something, but I don't see how this would work.
If I do like your example and save `[$this, 'fixIt']` on validatorResult...
in blade we need to  POST(ajax in perfect world) to a endpoint that will run the code, what parameters should i call that endpoint with then?

Currently it simply POST to `fix` endpoint with with class name, and that simply runs the `fix` function in that class if it exists.
I feel a bit stupid but I dont see a good way to pass along the validatorResult to it.

murrant:
You are right, it is two different requests, I was thinking about the CLI use case.

However, maybe validation results should have names or ids so you can re-run the validations for a category, and generate the same one to fix.

PR #13337, opened 2021-10-05
Implement OS/Poweralert.php

arjitc:

#### Please note

- [X]  If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

when discovery.php is run every 6 hours(?), the values for things like `version` get set to NULL (https://github.com/librenms/librenms/blob/master/LibreNMS/Modules/OS.php#L40), but since there isn't a OS file for Poweralert based devices, the value remains as NULL until the poller is run,

once the poller runs for the device, the actual `version` value gets setup again, and this creates a eventlog entry when the `version` is updated from NULL to $version

With this file being present `version` is made available right during the discovery process, so that it does not remain set as NULL until the poller is run (and also prevents the multiple eventlog entries.)

example of the multiple eventlog entries in the past,
image


arjitc:
> This should be implemented in yaml instead

I pretty much cloned what OS file for APC was doing (ie, https://github.com/librenms/librenms/blob/master/LibreNMS/OS/Apc.php)

I do see a YAML file for poweralert which has some values in it, I am looking further but I had some questions too which I've asked over at Discord :)

Jellyfrog:
If you remove the other one you break other peoples systems, just add both and you're done

arjitc:
> @arjitc if you add both oids to yaml, it will check which one returns data and use that (in order listed).
> 
> Also, ones not using yaml are likely because they were implemented before yaml existed.

I updated a device over to the newer PADM20 firmware for the LX series (https://www.tripplite.com/products/management-firmware-matrix)

It seems like the they merged everything into the first string for this, but LX15 devices continue to use (.1.3.6.1.2.1.33.1.1.4.0)

```
iso.3.6.1.2.1.1.1.0 = STRING: "Ubuntu 18.04 Linux 4.4.31 flash: 3.34G ram: 247204kB processor: armv7l software: PowerAlert 20.0.1 (Build 118)"
```

I'll update the yaml and send some more  updates for it

PR #13340, opened 2021-10-06
Improve Poweralert based devices support

arjitc:
- Add graphs / data for Input Voltage
- Add graphs / data for Total/Output Current
- Add 2nd version string for LX 15 / PADM15 version based devices

image

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/discovery/poweralert.yaml
+++ b/includes/definitions/discovery/poweralert.yaml
@@ -37,7 +37,7 @@ modules:
                     descr: "Current"
                 -
                     oid: 'tlpAtsOutputCurrent'
-                    num_oid: '.1.3.6.1.4.1.850.1.1.3.4.3.2.1.1.5.1.1'
+                    num_oid: '.1.3.6.1.4.1.850.1.1.3.4.3.2.1.1.5.1.{{ $index }}'
                     descr: "Current"
                     divisor: 100
         humidity:


includes/definitions/discovery/poweralert.yaml
```diff
@@ -4,9 +4,41 @@ modules:
         features: .1.3.6.1.4.1.850.10.2.2.1.9.2
         hardware: .1.3.6.1.2.1.33.1.1.2.0
         serial: .1.3.6.1.4.1.850.100.1.1.2.0
+        version: .1.3.6.1.2.1.33.1.1.4.0
```
Jellyfrog, line 7: Think you need an array when you hade multiple. Look at other yamls for examples

murrant:
You can add test data by following [the test documentation](https://docs.librenms.org/Developing/os/Test-Units/).
Short version:
`./scripts/collect-snmp-data.php -h <hostname>`
`./scripts/save-test-data.php -o <os> -v <unique variant name>`

drommc:
@arjitc @murrant  I'm using PDU devices with PADM 20 (https://www.tripplite.com/5-5-8kw-single-phase-monitored-pdu-lx-platform-interface-208-240v-outlets-20-c13-4-c19-l6-30p-0u-taa~pdumnv30hvlx) and the sensor_index tlpPduCircuitTotalCurrent.1.1 does not exist, it needs to be tlpPduCircuitTotalCurrent.1. How is it that the discovery/update process grabs/generates that $index of 1.1 instead of just 1, I'm reviewing the code but I'm not sure how the $index in `includes/definitions/discovery/poweralert.yaml` becomes 1.1 instead of just 1 which is what I need to correctly read Current/Voltage for this PDU device running Power Alert Device Manager 20 with an LX interface. Tested with 4 different devices.

drommc:
Actually I ran some tests and opened this pull request which seems to solve my issue: https://github.com/librenms/librenms/pull/13473 although to be honest I'm not entirely sure how that index is calculated/found but it does solve my particular issue. Please review kindly. Thank you!

PR #13363, opened 2021-10-14
Dispatcher: Fix poller groups reverting when setting via the Web UI.

murrant:
Don't update poller groups when updating stats.

Allows updates to the poller groups via the webui, but also means that changes to the local config.php won't be reflected if the groups have been set in the webui.  Also, the webui might show the wrong groups if it is controlled by config.php only.

Chose this way as it is backwards compatible.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/service.py
+++ b/LibreNMS/service.py
@@ -794,13 +794,12 @@ class Service:
         try:
             # Report on the poller instance as a whole
             self._db.query(
-                "INSERT INTO poller_cluster(node_id, poller_name, poller_version, poller_groups, last_report, master) "
-                'values("{0}", "{1}", "{2}", "{3}", NOW(), {4}) '
-                'ON DUPLICATE KEY UPDATE poller_version="{2}", last_report=NOW(), master={4}; '.format(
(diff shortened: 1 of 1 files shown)


haydenseitz:
I think this has caused a side effect that breaks non-default group pollers. If a dispatcher has a local config.php file that sets the poller group to something other than 0, the [`parse_group` function](https://github.com/librenms/librenms/blob/f8fba91b83433cab6cdc92fff89d36f45ded9fa6/LibreNMS/service.py#L315) runs and sets it properly. But somewhere along the way, the dispatcher adds a record back to the `poller_cluster` table. Now that `poller_groups` field isn't being set and defaults to an empty string, the next time the dispatcher starts, the `parse_groups` function reads the empty string from the poller_cluster table, throws an error, and sets the polling group back to 0. 

This isn't an issue if you set the poller group per dispatcher in the UI (updating the `poller_cluster` table), but I am using a docker setup where dispatcher are dynamic. I'm not sure what a good fix here would be - it seems like it's intended to have the poller_cluster settings override any local config.php settings?

PR #13350, opened 2021-10-10
newdevice: Added support for Teltonika RUTX series routers

hjcday:
Added support for the Teltonika RUTX series routers using the Teltonika RUTX MIB

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [x] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #13375, opened 2021-10-18
Add support for Teltonika RUT360

martinberg:
Due to changes in the TELTONIKA-MIB the current MIB in Librenms doesn't support newer router models.
This also means that it will break compatibility with older RUT devices when replacing the MIB.

It almost seems like #13350 uses the same OIDs but in a different MIB. However I don't have a RUTX device to test this.

Test data will be added later.

#### Please note

#### Testers


martinberg:
Oh boy this is a bit of a mess by Teltonika.
The MIB from the RUTX routers is named TELTONIKA-MIB and supports GPS lat/long and number of satellites connected.
The RUT360 routers also use a MIB named TELTONIKA-MIB but this version lacks the OIDs for GPS data due to the RUT360 not having GPS antennas.
In other words, they seem to have different MIBs for the RUT360 and the RUTX routers. 

This won't break things as of now due to the only OIDs that differ are for GPS and some digital io ports between the different versions. Hopefully they don't change the OIDs for other things in the future.

This PR is tested with a RUT360 and a RUT950. Feel free to merge if this seems good enough.

hjcday:
Yep the have multiple MIB versions, this is why I added RUTX to the one specific to the RUTX range.

martinberg:
> Yep the have multiple MIB versions, this is why I added RUTX to the one specific to the RUTX range.

Yes I understand that now. Didn't think the newer models would have different MIBs aswell.
Hopefully you are okay with me changing some of your contributions of the description fields to easier be able to search in InfluxDB? ie removing Cell ID and Connection State in the description.

A question to whoever knows, is there a way to show a DisplayString as a sensor? In this case, mOperator and CellID stored and displayed in some way

murrant:
There is no "string" sensor.  You could simply set it in the description or group.

murrant:
Because you modified some of the existing sensors (which is fine), you will need to update the existing test data to match.

martinberg:
I'm having trouble getting this to pass testing due to weird behavior with state_name
When running `./scripts/save-test-data.php -o rutos-rutx -m sensors -v rut360` the `state_name` for `SIM Pin State` becomes `Pinstate` even though I have the following in `rutos-rutx.yaml`.

```
                    oid: mPinState
                    num_oid: ".1.3.6.1.4.1.48690.2.2.1.10.{{ $index }}"
                    descr: SIM Pin State
                    group: Mobile
                    state_name: pinState
                    states:
                        - { descr: PIN REQUIRED, graph: 1, value: 0, generic: 2 }
                        - { descr: NOT READY, graph: 1, value: 1, generic: 1 }
                        - { descr: READY, graph: 1, value: 2, generic: 0 }
                        - { descr: unknown, graph: 1, value: 3, generic: 3 }
```

`simState` behaves like expected and has the correct name in the test JSON.

Is there some kind of caching in place or am I doing something wrong?
All files in the branch are up to date and the problem exists as of now

murrant:
Looks like a name collision with another state.  I would suggest using `MIB::stateName`  to prevent collisions.

martinberg:
> Looks like a name collision with another state. I would suggest using `MIB::stateName` to prevent collisions.

Thanks!
I believe the collision was with original rutos which has `Pinstate`.

Still failing tests though, is there something more that has to be changed or is failing tests expected when modifying old test data?

murrant:
If changes affect existing test data, you can regenerate the test data (like creating new).  Make sure to inspect the changes so you don't introduce bad data.

martinberg:
> If changes affect existing test data, you can regenerate the test data (like creating new). Make sure to inspect the changes so you don't introduce bad data.

Well I didn't think of that...Now fixed

murrant (approved):

PR #13399, opened 2021-10-21
Luminato device doesn't support alternative uptimes

eskyuu:
We have some Luminato devices on our network, and I needed to disable the HR uptime SNMP polling because the devices are not returning valid data.

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/luminato.yaml
+++ b/includes/definitions/luminato.yaml
@@ -4,6 +4,8 @@ type: network
 icon: teleste
 group: luminato
 mib_dir: teleste
+bad_snmpEngineTime: true
+bad_hrSystemUptime: true
 over:
     - { graph: device_bits, text: 'Device Traffic' }
 discovery:


Jellyfrog:
You should notify the vendor also!

eskyuu:
@Npeca75 I don't actually know which version.

When looking at this issue, I actually thought that it may be better to make the core module ignore any non-numeric values from uptime counters instead of hard-coding certain devices as being good/bad.

murrant:
@eskyuu sounds reasonable to me, could get a random "good" number though that is wrong.
Would love to see some proposed code for what you think might be a good fix.

Npeca75:
@eskyuu 

> @Npeca75 I don't actually know which version.
> 

iso.3.6.1.2.1.1.1.0 = STRING: "Teleste Luminato 5.2.34"
iso.3.6.1.2.1.1.1.0 = STRING: "Teleste Luminato 4X4 20.8.26"
iso.3.6.1.2.1.1.1.0 = STRING: "Teleste Luminato 10.4.14"

could you please check on which version you have issue ?

eskyuu:
@Npeca75 
.1.3.6.1.2.1.1.1.0 = STRING: Teleste Luminato 10.4.10

PR #13400, opened 2021-10-21
Poller optimisations

eskyuu:
Update the poller and wrapper to improve the efficiency of the poller process. On our system the average CPU usage reduced from 70% to 60% as a result of these changes, and also resulted in the same number of threads being able to poll the devices in less time.

The efficiency is gained by:

Running a persistent copy of the poller for each worker during each polling run. This removes the overhead of compiling PHP for each device polled (even with opcache enabled)
Splitting the workload into 2 queues, and processing the devices that poll fastest from fastest to slowest, while still polling the devices that poll slowest in from slowest to fastest.
Splitting the worker threads so there are more threads assigned to the "slow" devices.
Calculating the number of threads required to completed the poll in less than the $STEP time, and using the command-line value as the maximum
The net result of 2 and 3 above is that the instant CPU load from the "fast" devices is reduced because there are less threads processing these devices. As the time taken to poll each "slow" devices speeds up, the time taken to poll each "fast" device slows down. This results in a more even flow of poll results into the database and time series database.

The net result of 4 above is that the polling load is spread evenly, and we don't try and run at 100% for a short time, then sit idle for the remainder of the polling time.

To answer the questions from PR13398:
 - There is no problem if the poller is being re-worked.  I think that if these changes can be merged immediately, we can see immediate improvements.  I would hope that any future improvements to the poller would try to achieve the same results (spread the CPU load over the whole polling interval, and reduce the number of times the poller process is forked).  
 - The code still polls the slowest devices first, so it still makes sure the slowest device gets polled in time.  It just adds another set of workers to poll the fastest devices in parallel to make better use of the CPU, and avoid a big spike in CPU use at the end of the polling run when all workers are polling "fast" devices.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/wrapper.py
+++ b/LibreNMS/wrapper.py
@@ -229,6 +229,7 @@ def poll_worker(
     log_dir,  # Type: str
     wrapper_type,  # Type: str
     debug,  # Type: bool
+    persistent,  # Type: bool
 ):
     """
     This function will fork off single instances of the php process, record
@@ -237,7 +238,7 @@ def poll_worker(
 
     global ERRORS
 
-    if not DISTRIBUTED_POLLING:
+    if persistent and not DISTRIBUTED_POLLING:
         executable = os.path.join(
             os.path.dirname(os.path.dirname(os.path.realpath(__file__))),
             wrappers[wrapper_type]["executable"],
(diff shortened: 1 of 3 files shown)


eskyuu:
@murrant Thanks for the feedback.  I'm working on other things at the moment, but I will work on a re-write that keeps the default behaviour unless the new one is enabled.  I am thinking that one option for persistent poller processes, and another for adaptive thread tuning so each can be enabled/disabled as needed.

I also want to add a "lock file" option to make sure one one poller process works at any one time (our production server ran out of memory due to runaway poller processes a few nights ago).  I can achieve this easily by using the fcntl python module, but from what I can see this module does not work on Windows.  Does the librenms code need to be cross-platform compatible with Windows?

murrant:
@eskyuu we don't use lock files, they are incompatible with distributed polling.   Luckily we already have Cache::lock().

Also, adding locks directly in the PHP polling code is one of the goals of #13414, but not directly implemented in that PR.

eskyuu:
> @eskyuu we don't use lock files, they are incompatible with distributed polling. Luckily we already have Cache::lock().
> 
> Also, adding locks directly in the PHP polling code is one of the goals of #13414, but not directly implemented in that PR.

@murrant Are you talking about per-device lock files, or per-

> @eskyuu we don't use lock files, they are incompatible with distributed polling. Luckily we already have Cache::lock().
> 
> Also, adding locks directly in the PHP polling code is one of the goals of #13414, but not directly implemented in that PR.

@murrant Are you referring to per-device locking, or making sure that only one poller-wrapper.py process is running at a time?  From what I can see, the current locking is all based around making sure a single device is only being polled once.  What I am thinking about is making sure only one poller-wrapper.py process is running at a time.  The goal here is to avoid out of memory conditions if something happens to cause the polling to take longer than normal.

I wanted to use flock/lockf/fcntl to handle the file lock, and these require the fcntl python module.  I had a quick look at cross platform compatibility, and from what I can see this python module is not compatible with Windows, which is why I've asked whether there are any restrictions on the cross-platform support for LibreNMS, specifically running on Windows.

I figure I can write the code to show you what I am thinking regarding the file lock, and you can give feedback on whether I've got the wrong idea with regards to the locking compatibility.

murrant:
the python-wrapper needs to overlap sometimes, because things happen.  If you are concerned, put a timeout on the process.

eskyuu:
> 
> 
> the python-wrapper needs to overlap sometimes, because things happen. If you are concerned, put a timeout on the process.

My plan takes into account that the poller sometimes needs to overlap.  The poller on our system used to overlap all the time prior to making the changes in this pull request.  It no longer overlaps under normal conditions, however if the server ever gets put under an unusually large load, I would like to only have one poller process run at a time instead of having 5+ copies of the poller running and the server running out of memory.  Our server ran into this exact scenario a couple of weeks ago, which is why I would like to solve the issue.

My plan is to make this an optional behaviour, with the default behaviour being off.

What I would like to know is whether the LibreNMS code is cross platform compatible with Windows, and whether a patch that did not work on Windows would be rejected?

murrant:
Generally, it only runs on Windows within WSL.

murrant:
@eskyuu why aren't you using the dispatcher service?

eskyuu:
> @eskyuu why aren't you using the dispatcher service?

I like the idea of moving to use the dispatcher service after you pointed it out to me earlier.  I just need to have the time to test swapping over, and I'm in the process of building our dev environment (not just LibreNMS, but all our systems).  It will just take time before I am ready to look into this further.

Having said this, the dispatcher service still forks one poller.php process every time a device is polled, so I would need to re-implement my changes in this system if I want to see the benefits.  I can achieve most of what I want (reduce CPU, spread the CPU load evenly and reduce total polling time) within the wrapper process, so the change to the dispatcher is a lower priority for me right now.

murrant:
FYI, dispatcher already spreads the load.  So you won't see any benefit with changes on that front.
And because the dispatcher workers are long lived (they don't close after one polling interval like the wrapper) you may run into memory leaks and other issues.
It will also reduce total polling time, because it is much better at scheduling, making a high thread count work better than the wrapper.

eskyuu:
I have just updated the code to make all optimisations options that can be enabled, but are turned off by default.  Please let me know what you think of this iteration of the patches?

Jellyfrog:
You have done some bad merge and included other people's commits in your branch

eskyuu:
Thanks for letting me know.  I'm pretty new to branching in git :)

I think I've fixed the issue, so the request should be good now.

murrant:
Sorry for the delay on your PR, but you happen to send them when I was reworking some poller things :)  It will be a bit still.

eskyuu:
> Sorry for the delay on your PR, but you happen to send them when I was reworking some poller things :) It will be a bit still.

Not a problem.  I'm also happy if this doesn't end up getting merged if your changes address the underlying poller performance improvements.

PR #13408, opened 2021-10-22
WIP - Add the ability to use LibreNMS device groups for Oxidized integration

ottorei:
This patch adds an optional device grouping mode for Oxidized where device groups can be retrieved straing from existing LibreNMS device groups (dynamic or static). Using this mode disables all other configured mappings except os-mapping since the device group itself should have all the required devices filtered correctly.

TODO:
- [x] 1. Functionality to actually use device groups for list_oxidized
- [ ] 2. An option to include a device group in the WebUI
- [ ] 3. Logic to include each invidual device to the list only once if there are overlapping device groups
- [ ] 4. If the device is in multiple groups, user can choose which group to prefer
- [ ] 5. Docs
- [x] 6. WebUI translations

#### Please note

#### Testers


murrant:
@ottorei os mapping is ALWAYS configured by default, because our os names are not all the same as Oxidize's OS model names.

ottorei:
> 
> 
> @ottorei os mapping is ALWAYS configured by default, because our os names are not all the same as Oxidize's OS model names.

Just noticed that while testing this. The mappings should work now.

jepke:
Just to put my voice out, id really like to see this feature. To use it for login credentials got tons of regex lines to group devices now but already have them grouped in device groups.

murrant:
Still seems like you could get non-deterministic behavior when a device is in multiple groups.

ottorei:
> Still seems like you could get non-deterministic behavior when a device is in multiple groups.
> 
> Also, why remove other maps when you could just replace the "group" map?

1. What kind of situation did you find here? It does not really matter which group the device is placed as long as the credentials / auth method can be applied for the device in Oxidized.
2. I am not sure I understand this point but I originally tried to use the existing code with some hacking for the mappings - testing showed excessive slowness when generating the list of the devices (tested with ~8000 devices). When the other mappings are bypassed, the device listing is returned almost instantly when requested. Also, when using the device groups in NMS itself, I believe no other possible mappings should be used. The groups should be defined with the query builder.

murrant:
@ottorei FYI, you can use groups to assign credentials in oxidized.. so if the group is not right, auth will fail.

You can map other fields too, can't think of one off the top of my head, but think something like setting one to use telnet instead of ssh.

ottorei:
> @ottorei FYI, you can use groups to assign credentials in oxidized.. so if the group is not right, auth will fail.
> 
> You can map other fields too, can't think of one off the top of my head, but think something like setting one to use telnet instead of ssh.

Was just thinking about this PR today. What about just passing all device groups to the Oxidized as they are defined in NMS? The user should be able to choose between this new option and the current one. I believe the current group-mapping is obsolote when the device group mapping is used (because the user can define dynamic / static groups as wanted). Oxidized would only handle the ones where there is configuration options present so it should not be a problem to pass additional groups?

bonzo81:
@ottorei I have started using oxidized recently, and can see the benefits in using the existing logic of the device groups to then pass into oxidized. Thanks for working on this.
My own thought was to simply add the option to apply a toggle to a device groups e.g. 'Use as Oxidixed group'. Then within the device settings, an option to select the oxidized group based on the device group(s) that have this toggle set, and that the device is a member of. A device could obviously be in multiple groups, but the user can only select one. If no group is selected then default group is used. That would ensure the device is only in one group when sending to oxidized.
Perhaps a toggle on the device group isn't even necessary. It could just be the option to use any device group as an oxidized group. Either way the user will need to specify a group per device or use default.
I might be missing something here, but happy to help and test in anyway way I can.

murrant:
@ottorei Been thinking too and I think a config array (where the order matters) will do just fine.

But to make management easier for the user, in the device groups management have a field that shows if the device is enabled for oxidized (in_array of the oxidized groups).  Clicking that will bring up a dialog to enable/disable and re-order the groups.

For the api call it should simply be:
1. run legacy maps
2. set default group if it is set
3. iterate oxidized enabled groups and set only the first one the device is a member (no set if not in any of the enabled groups)

murrant:
@ottorei I think I started to work on this, but got stuck on something and haven't had time.  It may take some time before I get some time to look at it again.

laf:
Closing for now, if you have any updates then let us know and we can re-open them.

PR #13411, opened 2021-10-23
Fix syslog widget priority filtering

murrant:
syslog stores priority as a string
filter strings, not numbers

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Enum/CheckStatus.php
+++ b/LibreNMS/Enum/CheckStatus.php
@@ -1,0 +1,34 @@
+<?php
+/**
+ * Check.php
+ *
+ * Nagios/monitoring style check status
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * @link       https://www.librenms.org
+ *
+ * @copyright  2021 Tony Murray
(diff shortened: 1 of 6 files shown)


Talkabout:
Hi @murrant ,

for me the priority filtering is not working any more. In your commit the filter column is still "level" which stores a number, not a string. Only the "priority" column might contain a string, but only if the template part for "syslogpriority" for rsyslog (which I am using) is changed to "syslogpriority-text". From my point of view filtering for numbers is much faster than strings. What is your opinion on this?

Talkabout:
@murrant , Can be easily fixed by doing this:

image

Is this a valid solution for you?

murrant:
@Talkabout AHA, my sample data only contains strings.   We are breaking 1 of 2 groups of people everytime we fix this :D

The only proper fix is to search for the string or the number.

murrant:
Speaking of @Talkabout, Do you know why there is priority and level?  Do they contain the same values? Is one supposed to be numeric?

Talkabout:
Hi @murrant ,

I think it depends on the configuration of the log facility you are using. The rsyslog template I am using is storing numbers for both, priority and level, but this can be easily changed. I tried that also already but the strings that are getting stored are different from the strings that you are using inside your level array. So it not only seems to be a problem whether string/number, but also what kind of string. That is why I think chosing numbers is the best way to go, but I have suggested a solution in the other thread that might allow to handle both cases.

murrant:
So, what are the levels as stored by your rsyslogd?

Talkabout:
rsyslog stores "error", "warning", "notice", "critical" if I use "syslogpriority-text" instead of "syslogpriority" as described here:

https://docs.librenms.org/Extensions/Syslog/#rsyslog

murrant:
There are 8 levels :D

Talkabout:
Hm,

looking here

https://wiki.archlinux.org/title/rsyslog

there is a table, which gives similar values to what you have used. But this means that we need to replace all the syslog templates to use string severity, correct?

murrant:
Probably, I just checked the templates and the version 7 and 8 ones switched from text to number.  The legacy one uses text.

Talkabout:
good catch, I am using librenms since a few months now, wasn't aware of this :)

murrant:
Let's keep discussion on the bug.

PR #13456, opened 2021-11-01
Added support for Himoinsa gensets status state sensors

TheGreatDoc:
Im going to start saying this is, by far, the worst state sensor i've never worked on.

There are 3 kind of control units:
- CEM7 -> Engine control unit
- CEC7 -> Switchboard control unit
- CEA7 -> A combination of CEM7 and CEC7 in a single control unit.

So, depending if you have a combination of CEM7 + CEC7 (wich is my case) or a single CEA7, you have to retrieve values from one oid or other.

Now the OIDs. We have 2 oids. status.0 and statusConm.0.
status.0 belongs to CEM7/CEA7 and statusConm.0 to CEC7. **Values for both are bitmap**

But not every bit is an independent sensor.

For example, the status.0 (remember, it belongs to CEM7/CEA7 control units) returns this (R G Al Bt B T Mn Au P A):

Only CEA7
R = Mains switch closed
G = Genset switch closed
Both, CEM7/CEA7
Al = Active Alarm
Bt = Transfer Pump
B = Blocked mode
T = Test mode
Mn = Manual mode
Au = Auto mode
P = Engine Stopped
A = Engine Running

R and G are a single state sensor, as only one can be active and indicates what switch is closed.
Al is single one.
Bt is single one.
B, T, Mn and Au are the Mode state.
P and A are the Engine state.

Some state sensors comes in both oids, like the Mode one and I retrieved it from the status oid as if you have a CEA7, no statusConm oid is present.

In the custom poller, im trying to be as efficient as possible but the php is called for every sensor discovered, so redundant calls are made.

I think thats all. I'll try to support more sensors in the future, but for now I think this is a good start.

PS:
As english is not my native language, feel free to change words if you think they fit better.

#### Please note

#### Testers


includes/discovery/sensors/state/himoinsa-gensets.inc.php
```diff
@@ -0,0 +1,371 @@
...
+A = Motor Running
+*/
+
+// Check if CEC7 info is available and retrieve data
+$statusConm = SnmpQuery::get('HIMOINSAv14-MIB::statusConm.0')->value();
+// Check if CEA7/CEM7 info is available and retrieve data
+$status = SnmpQuery::get('HIMOINSAv14-MIB::status.0')->value();
+
+if (is_numeric($status)) {
+    define('A', 0b0000000001);
```
murrant, line 70: Don't use define() :)

murrant (changes requested):
You need to be using bitwise operators https://www.php.net/manual/en/language.operators.bitwise.php

Otherwise this is going to be way too confusing for anyone to try to figure out what it is doing.

Some psuedo code to show how the bitmasks would work

```php
$status = SnmpQuery::get(['HIMOINSAv14-MIB::status.0', 'HIMOINSAv14-MIB::statusConm.0'])->values();

// map of bits
if (isset($status['HIMOINSAv14-MIB::status.0'])) {
    $map = [
        'Mains commutator closed' => 256,
        'Gens commutator closed' => 512,
        'Active Alarm' => 128,
        'Transfer Pump' => 64,
        'Blocked mode' => 32,
        'Test mode' => 16,
        'Manual mode' => 8,
        'Auto mode' => 4,
        'Motor Stopped' => 2,
        'Motor Running' => 1,
    ];

    foreach ($map as $descr => $mask) {
        if ($status['HIMOINSAv14-MIB::status.0'] & $mask) {
            // discover sensor with $descr
        }
    }
}

if (isset($status['HIMOINSAv14-MIB::statusConm.0'])) {
    $map = [
        'Mains commutator closed' => 32,
        'Gen commutator closed' => 64,
    ];
    if ($status['HIMOINSAv14-MIB::statusConm.0'] & $mask) {
        // discover sensor with $descr
    }
}
```

murrant:
Oh, here is another "fun" way

```php
if (isset($status['HIMOINSAv14-MIB::status.0'])) {
    $map = [
        null,
        'Motor Running',
        'Motor Stopped',
        'Auto mode',
        'Manual mode',
        'Test mode',
        'Blocked mode',
        'Transfer Pump',
        'Active Alarm',
        'Mains commutator closed',
        'Gens commutator closed',
    ];

    // scan through each bit, of the value the position relates to the index of the $map
    $scan = 1;
    while ($status['HIMOINSAv14-MIB::status.0'] >= $scan) {
        if ($status['HIMOINSAv14-MIB::status.0'] & $scan) {
            // discover sensor with descr $map[$scan]
        }
        $scan <<= 1;
    }
}
```

TheGreatDoc:
@murrant Its already confusing to me so I understand what you mean. And Im already using bitwise operators ;)

> ```
> $status = SnmpQuery::get(['HIMOINSAv14-MIB::status.0', 'HIMOINSAv14-MIB::statusConm.0'])->values();
> 
> // map of bits
> if (isset($status['HIMOINSAv14-MIB::status.0'])) {
>     $map = [
>         'Mains commutator closed' => 256,
>         'Gens commutator closed' => 512,
>         'Active Alarm' => 128,
>         'Transfer Pump' => 64,
>         'Blocked mode' => 32,
>         'Test mode' => 16,
>         'Manual mode' => 8,
>         'Auto mode' => 4,
>         'Motor Stopped' => 2,
>         'Motor Running' => 1,
>     ];
> 
>     foreach ($map as $descr => $mask) {
>         if ($status['HIMOINSAv14-MIB::status.0'] & $mask) {
>             // discover sensor with $descr
>         }
>     }
> }
> 
> if (isset($status['HIMOINSAv14-MIB::statusConm.0'])) {
>     $map = [
>         'Mains commutator closed' => 32,
>         'Gen commutator closed' => 64,
>     ];
>     if ($status['HIMOINSAv14-MIB::statusConm.0'] & $mask) {
>         // discover sensor with $descr
>     }
> }
> ```

But wouldn't that code discover a different sensor for each bit?

murrant:
@TheGreatDoc yes, but I didn't write that part, because I figured you could figure out what to do from there to bin them into the correct sensors.

TheGreatDoc:
@murrant Im doing my best, but found a weird thing.
This:
```$status = SnmpQuery::get(['HIMOINSAv14-MIB::status.0', 'HIMOINSAv14-MIB::statusConm.0'])->values();```
return this:
```array(2) {
  ["HIMOINSAv14-MIB::status.0"]=>
  string(1) "6"
  ["HIMOINSAv14-MIB::statusConm[0]"]=>
  string(2) "66"
}
```
Note how the statusConm var name is saved.

TheGreatDoc:
Also, how does this code look?

murrant:
This means `statusConm` is part of a table.  Don't worry, the code can parse it just fine.  It is a bit unexpected in your use case though, isn't it.

includes/discovery/sensors/state/himoinsa-gensets.inc.php
```diff
@@ -0,0 +1,368 @@
...
+    ];
+
+    $statusMotor = 0;
+    $statusMode = 0;
+    $statusAlarm = 0;
+    $statusComm = 0;
+    $statusTransferPump = 0;
+    $statusCommAlarm = 0;
+
+    foreach ($map as $descr => $mask) {
```
murrant, line 87: Still confusing, why are we looping, why don't we just check each mask once and set the variable?

murrant: Like this?

```php
$status = $status['HIMOINSAv14-MIB::status.0'] ?? 0;
$statusMotor = $status & 512 | $status & 256;  // assuming these values are mutually exclisive, it will result in a value of 256, 512, or 0;  (768 if both can be set)
```

```

Comment (TheGreatDoc) [reply includes/discovery/sensors/state/himoinsa-gensets.inc.php]: Its a feature. If not, the code is going to be so fast and clean that some servers could implode and generate a micro black hole, destroying the whole solar system.

File: includes/discovery/sensors/state/himoinsa-gensets.inc.php [hunk 3]
```diff
@@ -0,0 +1,331 @@
...
+$status = SnmpQuery::get(['HIMOINSAv14-MIB::status.0', 'HIMOINSAv14-MIB::statusConm.0'])->values();
+
+if (isset($status['HIMOINSAv14-MIB::status.0'])) {
+    $statusMotor = ($status['HIMOINSAv14-MIB::status.0'] & 1) | ($status['HIMOINSAv14-MIB::status.0'] & 2);
+    $statusMode =
+        ($status['HIMOINSAv14-MIB::status.0'] & 4) |
+        ($status['HIMOINSAv14-MIB::status.0'] & 8) |
+        ($status['HIMOINSAv14-MIB::status.0'] & 16) |
+        ($status['HIMOINSAv14-MIB::status.0'] & 32);
+    $statusAlarm = ($status['HIMOINSAv14-MIB::status.0'] & 128) ?? 0;
```
murrant, line 72: ?? 0 does nothing here (and below), the only two possible values are 128 and 0.

murrant:
Looking good now!  Just need to update polling.  I would suggest sharing the same parsing code for discovery and polling.  The best place to put that code is the OS class file.

TheGreatDoc:
@murrant Im not sure how to do that. Do you mind pointing me an example?

includes/discovery/sensors/state/himoinsa-gensets.inc.php
TheGreatDoc: But doesnt that means that if `$status['HIMOINSAv14-MIB::status.0'] & 128` doesnt match then set $statusAlarm to 0?

TheGreatDoc:
Also, few translation quesitons:

Motor or Engine or...?
Commutator or Switch or...?

Any other word that need to be changed?

murrant:
Create LibreNMS/OS/HimoinsaGensets.php extending OS.  (check others for an example).

Create a new (or two) method on it and paste in the code to parse the sensor.  Then just call it in both discovery and polling.

TheGreatDoc:
I did that before asking you and couldn't find an example that fits this situation.

There is no (or I cant find) custom discovery/polling that makes uses of methods.

murrant:
You can't use the yaml method, it doesn't give you enough context, unless you extend it to pass more than the value.

I was just suggesting you extract the code from the current php discovery so you don't have to duplicate it.

TheGreatDoc:
Yay @murrant ! I think I found a problem using a method to parse data.

Custom discovery will run one time per device, which is fine.

Custom polling will run one time per sensor, so if there are 6 state sensors, `includes/polling/sensors/state/himoinsa-gensets.inc.php` is going to be run 6 times (which means its going to load the class and full parse all values 6 times).

Isnt that a bit inefficient?

murrant:
The amount of time it takes to parse the sensors is not very significant.
BUT, you should pass info about which sensor you are polling and only parse that one.  Call it multiple times, but only return data for one sensor each time.

TheGreatDoc:
@murrant What about now?

murrant (approved):
:fireworks:

PR #13460, opened 2021-11-02
Polling cleanup, fix PHP warnings

murrant:
Capture all warnings generated by test data and fix them.
Mostly by converting os polling files to yaml.
Go ahead and convert most of the os polling files to yaml while I wait.

a lot of isset and null coalescing ??

#### Please note

#### Testers


app/Console/Commands/FindWarnings.php
```diff
@@ -0,0 +1,96 @@
+<?php
```
Jellyfrog, line 1: should we keep this? it is pretty after all :D

Jellyfrog (approved):
null coalescing is not pretty, but what to do :)

app/Console/Commands/FindWarnings.php
murrant: I was on the fence.  I'll keep it for now and delete later, maybe. :)

tests/snmpsim/samsungprinter_clx.snmprec
```diff
@@ -1 +1 @@
-1.3.6.1.2.1.1.1.0|4|Samsung CLX
+1.3.6.1.2.1.1.1.0|4|Samsung CLX-3170 Series;V1.00.01.64 Sep-27-2010;Engine 1.77.81;NIC V4.01.20(CLX-3170) 02-05-2010;S/N AAAAAAA000AAAA000
```
Jellyfrog, line 1: where is all this new data coming from? same with all other snmprec changes

tests/data/datacom.json
```diff
@@ -23,7 +23,7 @@
                 {
                     "sysName": "",
                     "sysObjectID": ".1.3.6.1.4.1.3709",
-                    "sysDescr": null,
+                    "sysDescr": "",
```
Jellyfrog, line 26: should we make empty strings null?

tests/snmpsim/samsungprinter_clx.snmprec
murrant: These are from the stats site.  It logs sysObjectID and sysDescr

tests/data/datacom.json
murrant: Probably should be null.  LibreNMS seems inconsistent with sys* variables and null/""

glance-:
This will break newer areca cards, by just looking at 1.3.6.1.4.1.18928.1.1.1.1.0 and not 1.3.6.1.4.1.18928.1.2.1.1.0.

murrant:
@glance- that is unrelated to this change.  If you want to add additional detections, please send a pull request :) `1.3.6.1.4.1.18928.1.2.1.1.0` was never used for hardware name before.

glance-:
Yes, it was used:
https://github.com/librenms/librenms/blob/666638eeaabd0abb2f3d46d9cef6134cf55b9671/includes/polling/os/areca.inc.php#L5

The whole areca module is missing some vital functionality, like monitoring disks, raids and volumes but that have never bin there, and whenever I get time for it I'll try to make something up.

murrant:
Sending a fix, next time create an issue instead of commenting on a merged PR ;)

PR #13473, opened 2021-11-03
Support for PDUMNV30HVLX with PADM 20

drommc:
For some reason this pull request: https://github.com/librenms/librenms/pull/13340 does not work/fix issues with PADM 20 and PDUs like the above with the LX cards. By adding the additional sensor data in definitions > discovery the PDU data is correctly pulled by LibreNMS.

#### Please note

#### Testers


Jellyfrog (dismissed):
Looks resonoble to me but snmp oids and indexes aren't my thing. @murrant ?

includes/definitions/discovery/poweralert.yaml
```diff
@@ -35,6 +43,13 @@ modules:
                     index: 'tlpPduCircuitTotalCurrent.{{ $index }}'
                     divisor: 100
                     descr: "Current"
+                -
```
murrant, line 46: This is a duplicate of the existing tlpPduCircuitTotalCurrent sensor.  I suggest you delete the previous one, which must have had the wrong index.

includes/definitions/discovery/poweralert.yaml
```diff
@@ -19,6 +19,14 @@ modules:
                     multiplier: 1
                     low_limit:  110
                     high_limit:  240
+                -
```
murrant, line 22: Delete the duplicate sensor.

includes/definitions/discovery/poweralert.yaml
drommc: This was just recently merged into master: https://github.com/librenms/librenms/pull/13340/files#diff-46bf9d59eb5cac5439a3ab8e7b552e949981e045e3cc30ab60d14c571be1545eR30 

If I deleted the previous sensor wouldn't that affect @arjitc PR? I'm not sure why the index worked on his tests so recently, it would be interesting to know what device he was using to test.

drommc:
Removed duplicate sensor and ran tests again. Please review. Thank you.

includes/definitions/discovery/poweralert.yaml
arjitc: Hi @murrant  / @drommc

I'm not too sure which model we use exactly, but they were recently upgraded to the same LXCARD 20 version and tlpPduCircuitTotalCurrent should still work (basically tlpPDUCircuitTotalCurrent.1.1 or tlpPDUCircuitTotalCurrent.1.2) etc

I got the OIDs from https://assets.tripplite.com/technical-bulletins/supported-snmp-oids-technical-bulletin-en.pdf

Jellyfrog (approved):

PR #13479, opened 2021-11-05
[bug] Fix & extend MAC OUI table updates

PipoCanaja:
The source (macaddress.io) is now hiding the vendor names in free version. So we'll directly fetch the source in wireshark git readonly repository. 

We'll also now parse all prefixes, not only the /24 prefixes like before, but also the /28 and /36. 

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Util/Rewrite.php
+++ b/LibreNMS/Util/Rewrite.php
@@ -156,10 +156,11 @@ class Rewrite
     public static function readableOUI($mac)
     {
         $cached = Cache::get('OUIDB-' . (substr($mac, 0, 6)), '');
-        if ($cached == "IEEE Registration Authority") {
+        if ($cached == 'IEEE Registration Authority') {
             // Then we may have a shorter prefix, so let's try them one ater the other, ordered by probability
             return Cache::get('OUIDB-' . (substr($mac, 0, 7)), Cache::get('OUIDB-' . (substr($mac, 0, 9))));
         }
+
         return $cached;
     }
 
--- a/LibreNMS/Util/Rewrite.php
+++ b/LibreNMS/Util/Rewrite.php
@@ -155,9 +155,12 @@ class Rewrite
      */
     public static function readableOUI($mac)
     {
-        $key = 'OUIDB-' . (substr($mac, 0, 6));
-
-        return Cache::get($key, '');
+        $cached = Cache::get('OUIDB-' . (substr($mac, 0, 6)), '');
+        if ($cached == "IEEE Registration Authority") {
+            // Then we may have a shorter prefix, so let's try them one ater the other, ordered by probability
+            return Cache::get('OUIDB-' . (substr($mac, 0, 7)), Cache::get('OUIDB-' . (substr($mac, 0, 9))));
+        }
+        return $cached;
     }
 
     /**
--- a/LibreNMS/Util/Rewrite.php
+++ b/LibreNMS/Util/Rewrite.php
@@ -158,7 +158,7 @@ class Rewrite
         $cached = Cache::get('OUIDB-' . (substr($mac, 0, 6)), '');
         if ($cached == 'IEEE Registration Authority') {
             // Then we may have a shorter prefix, so let's try them one ater the other, ordered by probability
-            return Cache::get('OUIDB-' . (substr($mac, 0, 9)), Cache::get('OUIDB-' . (substr($mac, 0, 7))));
+            return Cache::get('OUIDB-' . substr($mac, 0, 9)) ?: Cache::get('OUIDB-' . substr($mac, 0, 7));
         }
 
         return $cached;
--- a/LibreNMS/Util/Rewrite.php
+++ b/LibreNMS/Util/Rewrite.php
@@ -158,7 +158,7 @@ class Rewrite
         $cached = Cache::get('OUIDB-' . (substr($mac, 0, 6)), '');
         if ($cached == 'IEEE Registration Authority') {
             // Then we may have a shorter prefix, so let's try them one ater the other, ordered by probability
-            return Cache::get('OUIDB-' . (substr($mac, 0, 7)), Cache::get('OUIDB-' . (substr($mac, 0, 9))));
+            return Cache::get('OUIDB-' . (substr($mac, 0, 9)), Cache::get('OUIDB-' . (substr($mac, 0, 7))));
         }
 
         return $cached;
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1432,7 +1432,8 @@ function cache_mac_oui()
     if ($lock->get()) {
         echo 'Caching Mac OUI' . PHP_EOL;
         try {
-            $mac_oui_url = 'https://raw.githubusercontent.com/wireshark/wireshark/master/manuf';
+            $mac_oui_url = 'https://gitlab.com/wireshark/wireshark/-/raw/master/manuf';
+            //$mac_oui_url_mirror = 'https://raw.githubusercontent.com/wireshark/wireshark/master/manuf';
             echo '  -> Downloading ...' . PHP_EOL;
             $get = Requests::get($mac_oui_url, [], ['proxy' => get_proxy()]);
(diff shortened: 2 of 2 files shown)


PipoCanaja:
And by the way, I also added support for OUI-28 and OUI-36 (smaller ranges than OUI). Not so many are allocated, so the probability to hit one is very low, but it does not cost much to add them (only 2 Cache::get request for them instead of one for any standard OUI).

LibreNMS/Util/Rewrite.php
```diff
@@ -155,9 +155,13 @@ public static function readableMac($mac)
      */
     public static function readableOUI($mac)
     {
-        $key = 'OUIDB-' . (substr($mac, 0, 6));
+        $cached = Cache::get('OUIDB-' . (substr($mac, 0, 6)), '');
+        if ($cached == 'IEEE Registration Authority') {
+            // Then we may have a shorter prefix, so let's try them one ater the other, ordered by probability
+            return Cache::get('OUIDB-' . (substr($mac, 0, 7)), Cache::get('OUIDB-' . (substr($mac, 0, 9))));
```
murrant, line 161: This indicates to me that the cache key is wrong.

includes/functions.php
```diff
@@ -1432,17 +1432,29 @@ function cache_mac_oui()
...
+            $csv_data = $get->body;
+            foreach (explode("\n", $csv_data) as $csv_line) {
+                unset($oui);
+                $entry = str_getcsv($csv_line, "\t");
+                if (is_array($entry) && count($entry) >= 3 && strlen($entry[0]) == 8) {
+                    // We have a standard OUI xx:xx:xx
+                    $oui = strtolower(str_replace(':', '', $entry[0]));
+                } elseif (is_array($entry) && count($entry) >= 3 && strlen($entry[0]) == 20) {
+                    // We have a smaller range (xx:xx:xx:X or xx:xx:xx:xx:X)
+                    if (substr($entry[0], -2) == '28') {
```
murrant, line 1448: what about caching an array of mac instead of individual ones in the case of a mac

LibreNMS/Util/Rewrite.php
murrant: I would try the longer one first, prevents collision

murrant: Also, this fetches both, determines if the /28 one is empty, then returns the /36 one.

includes/functions.php
```diff
@@ -1432,17 +1432,29 @@ function cache_mac_oui()
     if ($lock->get()) {
         echo 'Caching Mac OUI' . PHP_EOL;
         try {
-            $mac_oui_url = 'https://macaddress.io/database/macaddress.io-db.json';
+            $mac_oui_url = 'https://raw.githubusercontent.com/wireshark/wireshark/master/manuf';
```
Jellyfrog, line 1435: https://gitlab.com/wireshark/wireshark/-/raw/master/manuf is the offical

includes/functions.php
```diff
@@ -1432,17 +1432,29 @@ function cache_mac_oui()
...
+                    $oui = strtolower(str_replace(':', '', $entry[0]));
+                } elseif (is_array($entry) && count($entry) >= 3 && strlen($entry[0]) == 20) {
+                    // We have a smaller range (xx:xx:xx:X or xx:xx:xx:xx:X)
+                    if (substr($entry[0], -2) == '28') {
+                        $oui = strtolower(str_replace(':', '', substr($entry[0], 0, 10)));
+                    } elseif (substr($entry[0], -2) == '36') {
+                        $oui = strtolower(str_replace(':', '', substr($entry[0], 0, 13)));
+                    }
+                }
+                if ($oui) {
```
Jellyfrog, line 1454: $oui might not be defined

includes/functions.php
```diff
@@ -1432,17 +1432,29 @@ function cache_mac_oui()
...
+                $entry = str_getcsv($csv_line, "\t");
+                if (is_array($entry) && count($entry) >= 3 && strlen($entry[0]) == 8) {
+                    // We have a standard OUI xx:xx:xx
+                    $oui = strtolower(str_replace(':', '', $entry[0]));
+                } elseif (is_array($entry) && count($entry) >= 3 && strlen($entry[0]) == 20) {
+                    // We have a smaller range (xx:xx:xx:X or xx:xx:xx:xx:X)
+                    if (substr($entry[0], -2) == '28') {
+                        $oui = strtolower(str_replace(':', '', substr($entry[0], 0, 10)));
+                    } elseif (substr($entry[0], -2) == '36') {
+                        $oui = strtolower(str_replace(':', '', substr($entry[0], 0, 13)));
```
Jellyfrog, line 1451: you run `strtolower(str_replace(':', ''` 3 times, it can be 1 in the end

LibreNMS/Util/Rewrite.php
PipoCanaja: > I would try the longer one first, prevents collision

In fact, the OUI is always defined. If OUI is 'IEEE Registration Authority' then it means a small range exists. It can be either all /28 or all /36. Never /36 in /28. And statistically, it is /28 most of the time, only a very limited number is /36. That's why I ordered it that way.

includes/functions.php
PipoCanaja: I did not think that I could be able to cache an array. Will experiment this.

includes/functions.php
PipoCanaja: yep. Github is a readonly mirror as far as I can tell. Will change it.

includes/functions.php
PipoCanaja: right. Will do.

LibreNMS/Util/Rewrite.php
PipoCanaja: > This indicates to me that the cache key is wrong.

Could you be a little bit more precise ? I don't get that one.

PipoCanaja: > Also, this fetches both, determines if the /28 one is empty, then returns the /36 one.

As I have no way to know which one I search in advance, I can only fetch one and the other. Just trying to fetch the most probable one first.

includes/functions.php
PipoCanaja: done

includes/functions.php
PipoCanaja: fixed

includes/functions.php
PipoCanaja: Example : 
```
librenms@librenms:~$ cat manuf.txt | grep /36 | grep 70:B3:D5 | wc -l 
4079
```
So the array would have 4079 lines in case of the `70:B3:D5` prefix. I definitly think that caching each value is more efficient than storing 1 array of 4k elements and looping thru it each time we need a value inside this prefix.

LibreNMS/Util/Rewrite.php
```diff
@@ -155,9 +155,13 @@ public static function readableMac($mac)
      */
     public static function readableOUI($mac)
     {
-        $key = 'OUIDB-' . (substr($mac, 0, 6));
+        $cached = Cache::get('OUIDB-' . (substr($mac, 0, 6)), '');
+        if ($cached == 'IEEE Registration Authority') {
+            // Then we may have a shorter prefix, so let's try them one ater the other, ordered by probability
+            return Cache::get('OUIDB-' . (substr($mac, 0, 9)), Cache::get('OUIDB-' . (substr($mac, 0, 7))));
```
murrant, line 161:
```php
return Cache::get('OUIDB-' . substr($mac, 0, 9)) ?: Cache::get('OUIDB-' . substr($mac, 0, 7));
```

This will prevent cache::get from being called twice 100% of the time.

murrant, line 161:
```suggestion
            return Cache::get('OUIDB-' . substr($mac, 0, 9)) ?: Cache::get('OUIDB-' . substr($mac, 0, 7));
```

PipoCanaja: right ! I don't know why it took me so long to get it :D

murrant (approved):

PR #13414, opened 2021-10-23
Poller command rewrite

murrant:
`lnms device:poll` replaces ./poller.php

Still testing things out so, not deleting ./poller.php yet (eventually we will redirect it to the command)
Introduces Actions
Tweaks output to use Log::info as the normal output string
Directly calls all class based modules, loads legacy modules through an wrapper
Adds PollingDevice and DevicePolled events.

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [x] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [x] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #13483, opened 2021-11-07
Win Remote Management - agentless polling

thford89:
Adding WinRM distributed polling using PyWinRM and add base functionality of collected data to UI. 

image

image

image

image

image

image

image

Add alerting template.
image

#### Please note

#### Testers


app/Models/WinRMService.php
```diff
@@ -0,0 +1,18 @@
+<?php
```
Jellyfrog, line 1: This file looks weird and unused.

Jellyfrog:
Cool stuff!
Please look into the failing tests when you got time.
(Basically run your python files through python black and it will autofomat them for you)

app/Models/WinRMService.php
thford89: You are correct. It looks like the file got saved on accident. Thank you for pointing that out. I will work through the failures as i can.

thford89: Got all checks passing. Apologize for so many commits, its been about 12 years since I touched PHP and my first go with Python. Nice being back to my roots tho. 

In the future is there a way to run the checks locally? 

./lnms dev:check did help with the initial issues but seems to show passed even when github checks fail.

murrant:
@thford89 Very cool, but a few design items.
1. WinRM shouldn't be a separate polling process
2. Certain data should use existing storage such as processes
3. Credentials should be stored encrypted.  Currently LibreNMS has the capability to do this, but no code to implement it.

Also, be aware, I have a pending poller re-write that might make all this easier. #13414

thford89:
@murrant Thank you for the information, I am closing this pull request for now.

I will move the processes to the current tables and see if there are other options for the remaining data to reuse tables. 

I will also look into the two other design issues and see what I can get done. I am very limited on time so it may be a while before getting completed.

murrant:
@thford89 thanks, I started working on credential storage, but didn't get too far.

thford89:
@murrant What's your idea for credential storage? Separate table using encryption provided by laravel? If I can get some time I can start trying to work on something. I may not be the best PHP dev but definitely would like to get back into it. I have been in .net for way too long.

murrant:
Correct, this is what I have so far. https://github.com/librenms/librenms/compare/master...murrant:credential-storage

Main hurdle is I am using snmp to prove out the concept, which is far reaching.  If you want to take the base model and migrations and build it out for winrm, that would work.

PR #13502, opened 2021-11-12
Adjust low and high default temperature guesses 

TheMysteriousX:
TIA-942 states a datacentre should have a minumum temperature of 18 degrees. This works as a good value for both chassis cooling sensors, and for environmental as it's around the point where condensation can start to form in humid environments.

Maximum temperature is a little trickier, as it varies depending on context. For an environmental sensor in a data centre, 27 degrees is the bad number. In a wiring cupboard though, it could be 40 degrees or so.

A quick sample of Intel CPU specifications list the maximum temperature above 70 for core temperature, up to 100. I can't find any hard numbers but it looks like Turbo enabled processors start throttling around 80 degrees. Checking my own instance, around 3% of the temperature sensors are exceed that threshold; all under heavy load.

The real solution is for thresholds to be provided by the vendor, but one of the most common sensor types (lm_sensors) does not do this. My feeling is the current behaviour doesn't really help anyone.

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -359,7 +359,7 @@ function sensor_low_limit($class, $current)
     // matching an empty case executes code until a break is reached
     switch ($class) {
         case 'temperature':
-            $limit = $current - 10;
+            $limit = Config::get('sensors.temperature_low_guess', 18);
             break;
         case 'voltage':
             $limit = $current * 0.85;
@@ -397,7 +397,7 @@ function sensor_limit($class, $current)
     // matching an empty case executes code until a break is reached
     switch ($class) {
         case 'temperature':
-            $limit = $current + 20;
+            $limit = Config::get('sensors.temperature_high_guess', 70);
             break;
         case 'voltage':
             $limit = $current * 1.15;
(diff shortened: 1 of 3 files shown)


murrant:
Hard values are worse than the offset values in many cases.

We could make better guesses if we had more context.  Processor temperatures and ambient/environmental temperatures are on completely different scales.

For example, a lot of our equipment is in unheated outdoor enclosures.  It can get down to -30C around here in the winter. Right now (and it isn't very cold):
image

murrant:
Perhaps you want to set min/max in the lm_sensors discovery itself, since we think those are all computer chip sensors.

TheMysteriousX:
Is that possible to do it statically? I'll take a look.

Can then set the default values to null, so it can be set by the user if they want this behaviour. I think in your instance the current behaviour doesn't really work either as you see swings of >30 degrees.

murrant:
> Is that possible to do it statically? I'll take a look.
> 
> Could then set the default values to null, so it can be set by the user if they want this behaviour. I think in your instance the current behaviour doesn't really work either as you see swings of >30 degrees.

Yes, mostly it doesn't work :)  but without the swings it would.

PR #13548, opened 2021-11-20
Mark OID not increasing as invalid

murrant:

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Data/Source/SnmpResponse.php
+++ b/LibreNMS/Data/Source/SnmpResponse.php
@@ -68,7 +68,7 @@ class SnmpResponse
     {
         $this->errorMessage = '';
         // not checking exitCode because I think it may lead to false negatives
-        $invalid = preg_match('/(Timeout: No Response from .*|Unknown user name|Authentication failure)/', $this->stderr, $errors)
+        $invalid = preg_match('/(Timeout: No Response from .*|Unknown user name|Authentication failure|Error: OID not increasing: .*)/', $this->stderr, $errors)
             || empty($this->raw)
             || preg_match('/(No Such Instance|No Such Object|No more variables left).*/', $this->raw, $errors);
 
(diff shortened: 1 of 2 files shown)


tests/Unit/SnmpResponseTest.php
```diff
@@ -222,5 +222,10 @@ public function testErrorHandling(): void
         $response = new SnmpResponse('', "snmpget: Authentication failure (incorrect password, community or key) (Sub-id not found: (top) -> sysDescr)\n", 1);
         $this->assertFalse($response->isValid());
         $this->assertEquals('Authentication failure', $response->getErrorMessage());
+
+        // Authentication failure
```
PipoCanaja, line 226: Copy - Paste :)

murrant: Yep ;)

LibreNMS/Data/Source/SnmpResponse.php
```diff
@@ -68,7 +68,7 @@ public function isValid(): bool
     {
         $this->errorMessage = '';
         // not checking exitCode because I think it may lead to false negatives
-        $invalid = preg_match('/(Timeout: No Response from .*|Unknown user name|Authentication failure)/', $this->stderr, $errors)
+        $invalid = preg_match('/(Timeout: No Response from .*|Unknown user name|Authentication failure|Error: OID not increasing: .*)/', $this->stderr, $errors)
             || empty($this->raw)
```
Jellyfrog, line 72: shouldnt we check this first? faster than a large regex

murrant: @Jellyfrog had it first, but notice, "raw" is stdout, the first regex checks stderr.  This results in the wrong error message.  The program flow ends up looking like: if the output is empty check stderr regex, otherwise, check stderr regex.  You could check if stderr is empty first, but I suspect that is similar in performance as a regex against an empty string.

Jellyfrog (approved):

PR #13553, opened 2021-11-22
Fix discovery and pooling Racom Ray2

Martin22:

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 13553`

Ray2's discovery and pooling haven't been working for almost a year. Old data is no longer displayed, so there is nothing to lose.


Jellyfrog:
So they fixed this in newer firmware?

Jellyfrog:
Still will drop support for older firmware(?) versions in that case, which is why I ask.
But newer versions won't work without it, so it's a bit 50/50

Martin22:
> Still will drop support for older firmware(?) versions in that case, which is why I ask. But newer versions won't work without it, so it's a bit 50/50

I haven't tested the older firmware, we don't use old firmware, we don't want to have buggy versions on the network. I'll try it and possibly edit the code.

Martin22:
The current version of LibreNMS does not detect Racom Ray2 with older firmware. The change will not affect older firmware versions as it does not work anyway.

Please approve and I will try to fix the detection of older firmware.

Jellyfrog (approved):

PR #13554, opened 2021-11-22
Fix security vuls

enferas:
Fix XSS vulnerabilities in /include/html directory.

#### Please note

#### Testers

### DIFF
--- a/includes/html/forms/rediscover-device.inc.php
+++ b/includes/html/forms/rediscover-device.inc.php
@@ -65,4 +65,4 @@ $output = [
 ];
 
 header('Content-type: application/json');
-echo json_encode($output, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
+echo htmlspecialchars(json_encode($output, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE));
--- a/includes/html/forms/rediscover-device.inc.php
+++ b/includes/html/forms/rediscover-device.inc.php
@@ -48,10 +48,10 @@ if (isset($_POST['device_id'])) {
 
         if (! empty($update) || $update == '0') {
             $status = 'ok';
-            $message = 'Devices of group ' . $_POST['device_group_id'] . ' will be rediscovered';
+            $message = 'Devices of group ' . htmlspecialchars($_POST['device_group_id']) . ' will be rediscovered';
         } else {
             $status = 'error';
-            $message = 'Error rediscovering devices of group ' . $_POST['device_group_id'];
+            $message = 'Error rediscovering devices of group ' . htmlspecialchars($_POST['device_group_id']);
         }
     }
 } else {
@@ -65,4 +65,4 @@ $output = [
 ];
 
 header('Content-type: application/json');
-echo htmlspecialchars(json_encode($output, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE));
+echo json_encode($output, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
--- a/includes/html/pages/alert-log.inc.php
+++ b/includes/html/pages/alert-log.inc.php
@@ -18,5 +18,5 @@ $device_id = '';
 $vars['fromdevice'] = false;
 require_once 'includes/html/modal/alert_details.php';
 require_once 'includes/html/common/alert-log.inc.php';
-echo implode('', $common_output);
+echo htmlspecialchars(implode('', $common_output));
 unset($device_id);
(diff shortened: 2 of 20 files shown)


includes/html/forms/rediscover-device.inc.php
```diff
@@ -65,4 +65,4 @@
 ];
 
 header('Content-type: application/json');
-echo json_encode($output, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
+echo htmlspecialchars(json_encode($output, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE));
```
Jellyfrog, line 68: Can json_encode really be exploited for this?

murrant: htmlspecialchars() will break json encoding in some situations right?

FYI, you care move the flags `JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE`, they are inlined from an old function and were only used to make the json "look pretty"...

includes/html/pages/alert-log.inc.php
```diff
@@ -18,5 +18,5 @@
 $vars['fromdevice'] = false;
 require_once 'includes/html/modal/alert_details.php';
 require_once 'includes/html/common/alert-log.inc.php';
-echo implode('', $common_output);
+echo htmlspecialchars(implode('', $common_output));
```
murrant, line 21: I'm pretty sure $common_output is html.  You will need to make any fixes in `includes/html/common/alert-log.inc.php` instead.

murrant:
Thanks for the fixes!  Some however escape strings that contain html or json.  You'll need a different strategy for those.

includes/html/pages/alert-log.inc.php
enferas: Thank you for your response.

I can see that $common_output has the variable $selected_min_severity in [here](https://github.com/librenms/librenms/blob/87951d28fdd3951495f59e8cf2cec4d025973209/includes/html/common/alert-log.inc.php#L122).
Which has a source [here](https://github.com/librenms/librenms/blob/87951d28fdd3951495f59e8cf2cec4d025973209/includes/html/common/alert-log.inc.php#L72)
What do you think ?

murrant: Fix it at the source ;)

murrant: Funny below it is cast to int (preventing xss)

murrant: Don't forget the other $common_output ones.

includes/html/forms/rediscover-device.inc.php
enferas: Depending on this answer in stackoverflow [here](https://stackoverflow.com/questions/12062146/is-json-encode-sufficient-xss-protection).
json_encode is not enough to protect from XSS. But maybe the suggestion answer can be better than htmlspecialchars.

enferas: This code is giving an XSS vulnerability in my local computer.
```php
<?php
$output = ["A"=>"<img src='#' onerror=alert(1)>","B"=>"Safe"];
echo json_encode($output);
```

Jellyfrog: Could just add htmlspecialchars around $message on line 64

includes/html/pages/alert-log.inc.php
Jellyfrog: > Funny below it is cast to int (preventing xss)

This seems like the (more) correct fix

enferas: If I understand well the cast for int is for the function array_search. but the variable $selected_min_severity will keep the string value. Because there is a concat string .= in the second assign.

```php
$selected_min_severity = '<option value="' . $_POST['min_severity'] . '" selected="selected">';
$selected_min_severity .= array_search((int) $_POST['min_severity'], $alert_severities) . '</option>';
```

enferas:
> Thanks for the fixes! Some however escape strings that contain html or json. You'll need a different strategy for those.

The changes are done to sanitize at the level of the sources instead of the sinks.

murrant (approved):
Looks good now @enferas 

Thanks :)

Jellyfrog:
Thanks for not only reporting these, but also stepping up to fixing them!

enferas:
> Thanks for not only reporting these, but also stepping up to fixing them!

It is my pleasure. Thank you for your fast responses and merging my pull request.

enferas:
3 CVEs are assigned.

 CVE-2021-44277
Librenms 21.11.0 is affected by is affected by a Cross Site Scripting (XSS) vulnerability in includes/html/common/alert-log.inc.php.

CVE-2021-44278
Librenms 21.11.0 is affected by a path manipulation vulnerability in includes/html/pages/device/showconfig.inc.php.

CVE-2021-44279
Librenms 21.11.0 is affected by is affected by a Cross Site Scripting (XSS) vulnerability in includes/html/forms/poller-groups.inc.php.

PR #13565, opened 2021-11-24
WIP: Fix Cisco WLC AP-polling, part 2

ottorei:
1. Change the compositekey to uniquely identify each AP by WLC's device_id + ap_mac + ap_radionumber
2. Fix compare statement

WIP: Still testing this. This fix is to address an issue where each polling adds the same APs again to the access_points table. 

Edit: The logic is still not correct.

#### Please note

#### Testers


LibreNMS/OS/Ciscowlc.php
```diff
@@ -120,15 +121,13 @@ public function pollOS(): void
...
-                $ap = $db_ap->fill($ap->getAttributes());
+            if (! $db_aps->get($ap->getCompositeKey())) {
+                // Previously seen AP was not present during the poll, mark it as deleted
+                $ap->deleted = 1;
             }
-
             $ap->save(); // persist ap
         }
 
-        $db_aps->each->delete(); // delete those not removed
```
murrant, line 131: Instead of these changes, just save valid ap ids and then delete all not in that list.

app/Models/AccessPoint.php
```diff
@@ -26,6 +27,6 @@ class AccessPoint extends DeviceRelatedModel implements Keyable
 
     public function getCompositeKey()
     {
-        return "{$this->name}_{$this->radio_number}";
+        return "{$this->device_id}_{$this->mac_addr}_{$this->radio_number}";
```
murrant, line 30: It doesn't seem like you would need both mac_addr and radio_number

ottorei: Are you sure? Each physical AP has an unique MAC but can have multiple radios configured (== a/b/g/n...) so the same MAC has multiple records. Each radio can also have different data such as client counts. To uniquely identify each record, I think both of them are needed. However, using the MAC is sufficient for uniquely identifying each physical device.

LibreNMS/OS/Ciscowlc.php
ottorei: Thought of this approach initially but there is a catch - the WLC will return only the APs that are present during the poll. If we would like to see which previously seen APs are missing (== offline) we would need to somehow mark them instead of removing them. This is why I am using the softdelete in the new wireless module.

ottorei: I think I will try to finish the new module instead of fixing the remaining issues in the old one.

PR #13567, opened 2021-11-25
API 'list_parents_of_host' - mostly doco, but one small code enhancement

claude191:
There are three commits as you may choose to only take some of them:

1. [CODE CHANGE] Add extra 'list_devices' API query type of 'device_id'.  This makes the API call somewhat like 'get_devices', except that it returns a different set of attributes for the device.

2. [DOC CHANGE] Fix a minor typo in the documentation of 'mkdocs serve'.

3. [DOC CHANGE] Document an API call 'list_parents_of_host' that is really just showing a way of using 'list_devices' to achieve the required result.

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -306,13 +306,16 @@ function list_devices(Illuminate\Http\Request $request)
     if ($type == 'all' || empty($type)) {
         $sql = '1';
     } elseif ($type == 'device_id') {
-        $sql = "`d`.`device_id` = $query";
+        $sql = "`d`.`device_id` = ?";
+        $param[] = $query;
     } elseif ($type == 'active') {
         $sql = "`d`.`ignore`='0' AND `d`.`disabled`='0'";
     } elseif ($type == 'location') {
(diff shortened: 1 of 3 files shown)


includes/html/api_functions.inc.php
```diff
@@ -305,6 +305,8 @@ function list_devices(Illuminate\Http\Request $request)
 
     if ($type == 'all' || empty($type)) {
         $sql = '1';
+    } elseif ($type == 'device_id') {
+        $sql = "`d`.`device_id` = $query";
```
Jellyfrog, line 309: This introduces a sql vulnerability

Jellyfrog (changes requested):

includes/html/api_functions.inc.php
claude191: Thanks.  I foolishly copied the "hostname" code.  I've committed a fix for the two existing vulnerabilities and the one I introduced,

Jellyfrog (approved):

Jellyfrog:
Thanks!

PR #11719, opened 2020-05-27
Updated Support of Smartax GPON OLT

jozefrebjak:
Updated support of SMARTAX GPON OLT

this PR contains also @Negatifff https://github.com/librenms/librenms/pull/11718

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [ ] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)
- [ ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #13579, opened 2021-11-29
SmartAX supports IF-MIB, skip custom polling and fix polling GPON ports

carranzamarioagustin:
Together with Shev and murrant in the LibreNMS Discord general chat, we fixed SmartAX GPON port monitoring by removing custom polling of IN and OUT octets. 

After this change, Huawei SmartAX OLTs are able to poll GPON interfaces normally. Tested in production environment with various SmartAX's OLTs. 

#### Please note

#### Testers


Jellyfrog (dismissed):
You didn't remove the code handling it

carranzamarioagustin:
Hmm, could you please help me in figuring out why the tests failed?

Also, what do you mean by "removing the code handling it"? By removing the first two lines, I got GPON port monitoring back on my production environment. Should I remove the following 6 lines as well?

carranzamarioagustin:
A version check could work nicely, except that Huawei SmartAX OS versions are a bit weird. In my organization, we have 4 OLTs, and both are two different versions. Two are "MA5683V800R017C10B056" and the other are "MA5800V100R018C00B070".

I sadly do not have the data of on which OS versions IF-MIB works, and on which it doesnt.

In any case, I think we should skip all lines instead of just the first two, so we can bring multicast, broadcasts and unicast.

murrant:
@carranzamarioagustin well, you can at least check if these work too:
```
IF-MIB::ifInUcastPkts
IF-MIB::ifOutUcastPkts
IF-MIB::ifHCInUcastPkts
IF-MIB::ifHCOutUcastPkts
IF-MIB::ifInBroadcastPkts
IF-MIB::ifOutBroadcastPkts
IF-MIB::ifHCInBroadcastPkts
IF-MIB::ifHCOutBroadcastPkts
IF-MIB::ifInMulticastPkts
IF-MIB::ifOutMulticastPkts
IF-MIB::ifHCInMulticastPkts
IF-MIB::ifHCOutMulticastPkts
```

These are the ones you say already work (for comparison)
```
IF-MIB::ifInOctets
IF-MIB::ifOutOctets
IF-MIB::ifHCInOctets
IF-MIB::ifHCOutOctets
```

The HC oids might not be supported, but as long as one of the two sets works, it is good.

carranzamarioagustin:
I just did a complete "snmpwalk" on my OLTs with different OS, and I got data for all IF-MIBs, including the HC OIDs. I can share the results of the snmpwalk with you if you'd like to see them.

Also, if I query the OIDs specified in the file (for example, hwGponOltEthernetStatisticReceivedBytes, which is '.1.3.6.1.4.1.2011.6.128.1.1.4.21.1.15') I receive data per interface.

Maybe theres something wrong with the file, because I remember that Shev just deleted the file to get GPON polling back.

murrant:
@carranzamarioagustin The file in question overrides the default polling and uses these other OIDs instead.

Well, just had an interesting read through https://github.com/librenms/librenms/pull/11719

Looks like R18 fixed IF-MIB.   If you think it is likely most people are on R18 or newer of the OLT, I suggest just delete the file.  Otherwise, try to check the version and skip this file if R18 or newer.

This might work:
```php
if (! preg_match('/R0[0-1][1-7]/', $device['version']) {
    return;
}
```

carranzamarioagustin:
I've tried on my production environment adding the change you suggested, but I didnt get any GPON polling on by R17 nor my R18 OLTs.

I would suggest just deleting the file. That way I got monitoring back.

carranzamarioagustin:
I can share the debug capture of polling with the change, and afterwards deleting the file if you would like to see, maybe you can see something I dont.

murrant:
Ok, +1 for deleting.  We'll see if anyone complains :)

murrant:
Also, I probably messed up the code in my example.  Writing code from memory without an IDE to double check it is bound for mistakes :)

murrant (approved):
LGTM

PR #13583, opened 2021-11-30
Global search: search device display

murrant:
Add display field to search (also port_desc_descr, portName, and bgpPeerDescr)
Rewrite backend
update typeahead bundle
update devices and ports indexes
reduce some port field sizes so we can index them
Colors -> Color

#### Please note

#### Testers


app/Http/Controllers/Ajax/DeviceSearchController.php
```diff
@@ -0,0 +1,108 @@
...
+                    ->orWhere('serial', 'LIKE', $like_search)
+                    ->orWhere('notes', 'LIKE', $like_search);
+
+                if (\LibreNMS\Util\IPv4::isValid($search, false)) {
+                    $baseQuery->leftJoin('ports', 'ports.device_id', '=', 'devices.device_id')
+                        ->leftJoin('ipv4_addresses', 'ipv4_addresses.port_id', 'ports.port_id');
+
+                    $query->orWhere('ipv4_address', '=', $search)
+                        ->orWhere('overwrite_ip', '=', $search)
+                        ->orWhere('ip', '=', inet_pton($search));
```
Jellyfrog, line 66: This is now exact matching instead of wildcard, but the `isValid()` check above prevents it anyway;
```
>>> \LibreNMS\Util\IPv4::isValid("10.10", false);
=> false
```
I think that will be annoying

murrant: @Jellyfrog copied from the existing code.... The joins are painful.  Additionally, the ip field is binary, so it won't match partials without a lot of extra logic.

I was thinking of moving the interface IP stuff to the port search, but decided that would be a separate PR.

Jellyfrog: Hmm

$query_filter .= ' OR `V4`.`ipv4_address` LIKE ?
                                        OR `D`.`overwrite_ip` LIKE ?
                                        OR `D`.`ip` = ? ';
                 $query_args_list = array_merge($query_args_list, ["%$search%", "%$search%", 

Well it's a mix since it's wrapped in a validip check. 

Think we should change to some regex that just checks if it's something that looks like a ip. Ie numbers and dots :)

murrant: Sounds good, but I'd prefer a separate PR for that :)

html/ajax_search.php
```diff
@@ -1,362 +0,0 @@
...
-
-        if (! Auth::user()->hasGlobalRead()) {
-            $device_ids = Permissions::devicesForUser()->toArray() ?: [0];
-            $perms_sql = '`D`.`device_id` IN ' . dbGenPlaceholders(count($device_ids)) . ' AND ';
-        } else {
-            $device_ids = [];
-            $perms_sql = '';
-        }
-
-        if ($_REQUEST['type'] == 'group') {
```
Jellyfrog, line 33: So group and the other friends are not used anywhere?

murrant: I could not find any other references to _search.php or ajax_search.php, So, yeah, no idea why that code was there.

Jellyfrog (approved):

html/ajax_search.php
Jellyfrog: cant find any either

PR #13591, opened 2021-12-01
ipv6 discovery switch to new DB syntax

### LINKED ISSUE #13481: IPv6 cross connecting ?
### The problem

Hi
when i looks at PORTS, every port which have ipv6 address is cross connected with device which is in totally different v6 subnet

for ex
fd00:1:1::101/64 is connected with fd00:3:254::206/64
i suspected first on LLDP but no, two device is 10 hops away

if i try do delete one by one addresses from DB / ipv6_addressess, the ports get back to normal, they are NOT cros connected anymore, which is OK

on IPv4, everything is fine

looks like IPv6 does not account prefix length ???

lnms1
lnms2
lnms3
lnms4

### Output of ./validate.php

```txt
$ ./validate.php
Component | Version
--------- | -------
LibreNMS  | 21.10.0-95-g1ff48018f
DB Schema | 2021_10_20_224207_increase_length_of_attrib_type_column (224)
PHP       | 7.4.23
Python    | 3.6.9
MySQL     | 5.7.35-0ubuntu0.18.04.1
RRDTool   | 1.7.0
SNMP      | NET-SNMP 5.7.3

[OK]    Composer Version: 2.1.11
```

_No response_

_No response_

Npeca75:
Resolve #13481 

Problem was with DB ipv6_networks and DB ipv6_addresses
both have mismatch with NULL/empty and LNMS put IPv6 addresses without "network_id" in same space

furthermore, addresses without "network_id" does not show in "Overview -> IPv6 Adresses"

#### Please note

#### Testers

### DIFF
--- a/app/Models/Ipv6Address.php
+++ b/app/Models/Ipv6Address.php
@@ -29,4 +29,13 @@ class Ipv6Address extends PortRelatedModel
 {
     public $timestamps = false;
     protected $primaryKey = 'ipv6_address_id';
+    protected $fillable = [
+        'ipv6_address',
+        'ipv6_compressed',
+        'ipv6_prefixlen',
+        'ipv6_origin',
+        'ipv6_network_id',
+        'port_id',
+        'context_name',
+    ];
 }
--- a/app/Models/Ipv6Address.php
+++ b/app/Models/Ipv6Address.php
@@ -17,10 +17,14 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  *
+ * Original Code:
  * @link       https://www.librenms.org
  *
  * @copyright  2018 Tony Murray
  * @author     Tony Murray <murraytony@gmail.com>
+ *
+ * Modified by:
+ * @author Peca Nesovanovic <peca.nesovanovic@sattrakt.com>
  */
 
 namespace App\Models;
--- a/app/Models/Ipv6Address.php
+++ b/app/Models/Ipv6Address.php
@@ -17,15 +17,11 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  *
- * Original Code:
- *
  * @link       https://www.librenms.org
  *
  * @copyright  2018 Tony Murray
  * @author     Tony Murray <murraytony@gmail.com>
- *
- * Modified by:
- * @author Peca Nesovanovic <peca.nesovanovic@sattrakt.com>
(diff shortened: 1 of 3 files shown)


Npeca75:
testetd with

fresh tables:
TRUNCATE TABLE ipv6_addresses;
TRUNCATE TABLE ipv6_networks;

live tables:
UPDATE ipv6_networks SET context_name = NULL;
UPDATE ipv6_addresses SET context_name = NULL;

live tables:
UPDATE ipv6_addresses SET ipv6_network_id=rand() * 3333, context_name=RAND();
UPDATE ipv6_networks SET context_name=RAND();

app/Models/Ipv6Address.php
```diff
@@ -17,10 +17,15 @@
...
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  *
+ * Original Code:
+ *
  * @link       https://www.librenms.org
  *
  * @copyright  2018 Tony Murray
  * @author     Tony Murray <murraytony@gmail.com>
+ *
+ * Modified by:
```
murrant, line 27: You don't have to mark original code and modified by.  (we have git history for that)
Preferred way is just to add @author.  You can also add @copyright if you feel it is appropriate.

includes/discovery/functions.inc.php
```diff
@@ -661,55 +664,61 @@ function discover_process_ipv6(&$valid, $ifIndex, $ipv6_address, $ipv6_prefixlen
...
-            } else {
-                $ipv6_network_id = dbFetchCell('SELECT `ipv6_network_id` FROM `ipv6_networks` WHERE `ipv6_network` = ? AND `context_name` = ?', [$ipv6_network, $context_name]);
+        $ipv6netDB = Ipv6Network::firstOrNew([
+            'ipv6_network' => $ipv6_network,
+        ], [
+            'context_name' => $context_name,
+        ]);
+
+        //check if DB data match and check for DB NULL
+        if ($ipv6netDB->context_name != $context_name || $ipv6netDB->context_name == null) {
```
murrant, line 682: Models support an "isDirty()" method, this will check if the model has changed fields so you can print out the updated message.

murrant: My preferred method is updateOrCreate and use a listener to see when the devices is updated or created.

app/Models/Ipv6Address.php
Npeca75: hope now it is OK

includes/discovery/functions.inc.php
Npeca75: modified as you suggested
learned something new today, yes it is much simpler this way

murrant (approved):
Nice Peca :)

PR #13593, opened 2021-12-02
set fixed_coordinates via api

VirTechSystems:
Set the fixed coordinates value via api call

#### Please note

#### Testers


murrant:
perhaps this should either default to true for the api call, or just be true when the api is called instead of adding an additional variable.

VirTechSystems:
I could see arguments both for adding the value and making the default true.  Not making it default would allow it to be created with the understanding that the coordinates will come from the device at a later time.

Using the default as true or adding the api value both work for our use case.

murrant:
Either is fine with me, please update the code and fix the style issues.  Thanks.

murrant (approved):

PR #13600, opened 2021-12-06
Add Sla jitter packet loss percent graph

bonzo81:

Added a graph to show the percent of packet loss for the jitter slas.
Requires polling of new OID for NumPackets to calculate percentage. This exists within the existing RTTMON-MIB used for sla polling.

image

#### Please note

#### Testers


murrant:
Hey, you got it!  Nice.  Only one problem, you can't modify existing RRD DS like that.  It will require the user to delete their RRD files, otherwise the data will be put into the wrong DS.

There is a new experimental feature to migrate data from old RRD, but you need to change the rrd name in order to do that. #13480

The other option would be to create a new RRD just for the new data.

bonzo81:
Thanks @murrant. OK I'll have a look at the experimental feature and see if I can work out how to use it. 
To create a new rrd file; would that be, create a new case block within the cisco.php pollslas function rather than editing the exiting jitter one as I did? (I haven't completely got my head round all the parts yet!)

bonzo81 (commented):
I've created a new rrd just for the packet loss percent graph so no longer adding to the original jitter sla rrd file. Not sure if this is right way to do it, but it was the only way I could make it work without user having to delete existing rrd file.

LibreNMS/OS/Shared/Cisco.php
```diff
@@ -475,6 +473,20 @@ public function pollSlas($slas)
...
+                        'PacketLossSD' => $data[$sla_nr]['rttMonLatestJitterOperPacketLossSD'],
+                        'PacketLossDS' => $data[$sla_nr]['rttMonLatestJitterOperPacketLossDS'],
+                    ];
+                    $rrd_name = ['sla', $sla_nr, 'loss-percent'];
+                    $rrd_def = RrdDefinition::make()
+                        ->addDataset('NumPackets', 'GAUGE', 0)
+                        ->addDataset('PacketLossSD', 'GAUGE', 0)
+                        ->addDataset('PacketLossDS', 'GAUGE', 0);
+                    $tags = compact('rrd_name', 'rrd_def', 'sla_nr', 'rtt_type');
+                    data_update($device, 'sla', $tags, $lossPercent);
```
murrant, line 488: need to change the data name or the tags otherwise this will have the same file name as the previous rrd file.

bonzo81: I changed to $rrd_name on line 482 `$rrd_name = ['sla', $sla_nr, 'loss-percent'];` This appeared to pass through to the file name.

includes/html/graphs/device/sla_jitter-loss-percent.inc.php
```diff
@@ -0,0 +1,40 @@
...
+ * Copyright (c) 2016 Aaron Daniels <aaron@daniels.id.au>
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details.
+ */
+
+$sla = dbFetchRow('SELECT `sla_nr` FROM `slas` WHERE `sla_id` = ?', [$vars['id']]);
```
murrant, line 14:
```php
$sla_nr = Sla::where('id', $vars['id'])->value('sla_nr');

if ($sla_nr) {
$sla['sla_nr'] -> $sla_nr
```

murrant:
@bonzo81 So, we can also store just the total packets in the new RRD and pull data from both RRD to generate the graph, probably better than duplicating the data.

includes/html/graphs/device/sla_jitter-loss-percent.inc.php
bonzo81: Ah I wondered about the dBFetchRow check failing. Afraid I'm not sure how to implement this. Assume some further logic required instead of `...`?
Will `use App\Models\Sla` also be required?

bonzo81:
I didn't think about duplicating the data. Good shout. I've changed new rrd to just include just NumPackets. Altered the graph file accordingly.

murrant:
@bonzo81 Great! I think this looks correct now just needs style fixes and maybe a little testing and should be good to merge.

bonzo81:
Hi @murrant, what else can I do to help this progress? Do I need up add test data? Never done that but will if needed. Also does the dbFetchRow need changed? I see that is still failing checks.

includes/html/graphs/device/sla_jitter-loss-percent.inc.php
murrant: Yes, the use would be required.  You should be able to just paste my code in.

murrant: The ... is your existing code. and change any references of $sla['sla_nr'] to  $sla_nr

laf:
If you can update the `dbFetchRows` usage as per murrants message and it should be good

includes/html/graphs/device/sla_jitter-loss-percent.inc.php
```diff
@@ -10,13 +10,14 @@
  * option) any later version.  Please see LICENSE.txt at the top level of
  * the source code distribution for details.
  */
+use App\Models\Sla;
 
-$sla = dbFetchRow('SELECT `sla_nr` FROM `slas` WHERE `sla_id` = ?', [$vars['id']]);
+$sla_nr = Sla::where('sla_id', $vars['id'])->value('sla_nr');
```
laf, line 16: Just missing the `if ($sla_nr) { .... }` now to make sure we are acting on results from mysql

laf (changes requested):

includes/html/graphs/device/sla_jitter-loss-percent.inc.php
bonzo81: thanks for the pointers @laf. I couldn't see the reason for it at first but think I understand now. I've added it in.

bonzo81: styleci doesn't like the closing brackets. I don't know why...

murrant:
Usually, the best option is to just download the diff from styleci and apply it.
You were missing a new line at the end of the file.

bonzo81:
> Usually, the best option is to just download the diff from styleci and apply it. You were missing a new line at the end of the file.

Thanks @murrant I'll remember that for my next PR! 😜

murrant (approved):
Nice work :)

PR #13603, opened 2021-12-07
Added support for Fiberstore S5800 switch

lennarttd:
Added support for Fiberstore S5800 switch, including CPU, memory, temperature values, fanspeed and PSU states.
As Fiberstore sells multiple switches it might be necessary to create multiple SWITCH-MIBs with different enterprise values and check for compatibility with other switch models.

#### Please note

#### Testers


lennarttd:
Test data has been added.

murrant:
Why did you delete the existing detections?

lennarttd:
I have updated the YAML so the existing detections are combined with the new detections.

lennarttd:
I believe the tests are failing as I adapted the MIB and set the enterpise line to match the MIB of the S5800.
The older versions of this switch have a different enterprise ID.

Is there a possibility to merge the two into one single MIB? Besides the enterprise line the MIBs should be identical.

The new model uses `	::= { enterprises 52642 1 }`
and the older model uses `	::= { enterprises 27975 }`

murrant:
@lennarttd I would highly suggest not doing that unless you know it works 100%.  Also, keep in mind that sysObjectID does not specify which oids the device will respond on.   Some bad SNMP implementations change the MIB base OID, which I hope they haven't done.  If they have changed the Base OID, unfortunately, it is basically a different OS and we probably need a seperate OS for it.

Jellyfrog:

PR #13609, opened 2021-12-10
Improve getFullDn() to retrieve RDN

a084ed22:
The previous instance of getFullDn() naively composed the RDN that is then used in the user bind attempt by prefixing and suffixing the attributes defined in config.php.

It may be desirable to allow login with an arbitrary attribute (i.e. mail address), or, as it is the case for me, when the LDAP server does not store the username as the primary key for the entity, and therefore a typical login based on uid would fail.

I believe this change should not impact existing installations as it would still compose the same DN. My testing on this has however been limited. 

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Authentication/LdapAuthorizer.php
+++ b/LibreNMS/Authentication/LdapAuthorizer.php
@@ -286,7 +286,12 @@ class LdapAuthorizer extends AuthorizerBase
      */
     protected function getFullDn($username)
     {
-        return Config::get('auth_ldap_prefix', '') . $username . Config::get('auth_ldap_suffix', '');
+        $connection = $this->getLdapConnection();
+        $filter = '(' . Config::get('auth_ldap_prefix') . $this->userloginname . ')';
+        $base_dn = preg_replace('/,ou=[^,]+,/', '', Config::get('auth_ldap_suffix'));
+        $search = ldap_search($connection, $base_dn, $filter);
+        $firstuser = ldap_first_entry($connection, $search);
+        return ldap_get_dn($connection, $firstuser);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


murrant:
I'm pretty sure fullDn was written that way to avoid LDAP queries to generate it.  Did you check how many additional queries this generates?

murrant:
There is a chance this will break some installs.  I'm tempted to mark the ldap auth module as deprecated and suggest people use https://ldaprecord.com/docs/laravel/v2/

murrant:
It was written with memberOf before and that didn't work with some user's setups as not all ldap servers implement memberof.

murrant:
Hmm, I was just glancing at the Ldap code and saw almost a duplicate of what you have written here, related to the setting `auth_ldap_wildcard_ou`  Does that do what you want?  If it does, perhaps we just need a documentation update.

a084ed22:
Unfortunately that is not sufficient. Even with `auth_ldap_wildcard_ou`, the code uses the naive uid=... DN when attempting to bind the logging in user to validate its credentials, and in my case it fails due to the LDAP being keyed on cn.

murrant:
So, set auth_ldap_wildcard_ou and set auth_ldap_prefix to ''?

laf:
@a084ed22 Does currants suggestion work for you?

a084ed22:
Apologies for the late reply, I've not had the chance to run tests.

enabling auth_ldap_wildcard_ou and setting auth_ldap_prefix to "" is not sufficient for logins to pass, it'll report invalid credentials from the web interface due to AuthenticationException being thrown (as tested with the auth_test.php script).

a084ed22:
Unfortunately, this appears to have broken again, beyond what my skills with php allow me to fix.

My approach seems to also not be compatible with the new changes of the php-ldap module for php >=8.x

As a result, we've migrated away from LDAP and to AD-based authentication for librenms.

I still believe it is erroneous, or at least naive, to construct the dn by composing it as a string, whereas it would be more correct to search for a specific user-unique attribute and derive the dn with ldap_get_dn()

PR #13617, opened 2021-12-13
Lenovo think station Support

Fehler12:
I've created initial support for Lenovo ThinkStation servers with XClarityController IMPI. Supported functions are basic sensors (temp, fan, voltages, general system status). Not (yet) supported are hard drive status, power supply status.

#### Please note

- [no] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers


Fehler12:
Got a hint for states in my other pull request. Will include this in these models, too.

murrant:
@Fehler12 what was the issue that you closed your PR?

Fehler12:
Hi, I just closed the request because of your hints of stats and of the failing unit tests. I wanted to check this before you invest some time. However I found no way of fixing the unit tests and have the detection working. This status is working with my systems, but the unit tests are failing. Any hints?

laf:
In your discovery yaml file:

`state_name:` should be removed unless it's for state sensor.
`skip_values: 'N/A'` can't be used, only numerical values can be put here. N/A should be skipped when discovery runs as it's not numeric.

Fehler12:
Hi, I've updated the yaml file. For some reason the detection fails if I add the $index value in the oid tag as requested in the unit tests. Do you have any idea on how to correct that? I've tried several ways, no one worked for me.

mibs/lenovoxcc/lnvgy_fw_xcc_tei3c0n-4.83_lenovoproducts-anyos_noarch.mib
```diff
@@ -0,0 +1,99 @@
+LENOVO-PRODUCTS-MIB DEFINITIONS ::= BEGIN
```
PipoCanaja, line 1: Please rename this file as 
`LENOVO-PRODUCTS-MIB`
(without any extension). General naming rule is to name the file exactly as the MIB is defined inside the file
`LENOVO-PRODUCTS-MIB DEFINITIONS ::= BEGIN`

mibs/lenovoxcc/lnvgy_fw_xcc_tei3c0n-4.83_lenovosmi-anyos_noarch.mib
```diff
@@ -0,0 +1,72 @@
+LENOVO-SMI-MIB DEFINITIONS ::= BEGIN
```
PipoCanaja, line 1: Same here : 
`LENOVO-SMI-MIB`

and same for both 2 remaning MIB files

includes/definitions/discovery/lenovoXClarityIMPI.yaml
```diff
@@ -0,0 +1,66 @@
...
+        hardware: LENOVO-XCC-MIB::machineLevelProductName.0
+        serial: LENOVO-XCC-MIB::machineLevelSerialNumber.0
+        version: LENOVO-XCC-MIB::xccVpdVersionString.1    
+    sensors: 
+        temperature: 
+            data: 
+                - 
+                    oid: temperature
+                    value: tempReading
+                    #num_oid: '.1.3.6.1.4.1.19046.11.1.1.1.2.'
```
PipoCanaja, line 13: you need to define `num_oid` for the tests to pass. 
Probably something like : 
`num_oid: '.1.3.6.1.4.1.19046.11.1.1.1.2.{{ $index }}' `

PipoCanaja: (if you look at the debug data displayed during discovery (`./discovery.php -v -d -h <deviceID> -m sensors`), you should have hints on which num_oid was computed automatically out of the MIB file.

PipoCanaja (dismissed):

mibs/lenovoxcc/lnvgy_fw_xcc_tei3c0n-4.83_lenovosmi-anyos_noarch.mib
Fehler12: Done!

includes/definitions/discovery/lenovoXClarityIMPI.yaml
Fehler12: Done! Thank you for your explanation!

laf (dismissed):
If tests pass, LGTM

includes/definitions/lenovoXClarityIMPI.yaml
```diff
@@ -0,0 +1,21 @@
+os: lenovoXClarityIMPI
```
murrant, line 1: I don't think introducing camel case OS names into our existing kebab case os names is a good idea.

Please rename the OS to `lenovo-xclarity` or `xclarity`  Thanks.

murrant (changes requested):
Please rename OS to keep consistency.

Fehler12:
Renamed the OS! Would you please recheck?

mibs/lenovo/LENOVO-PRODUCTS-MIB
```diff
@@ -10,29 +7,29 @@ IMPORTS
         FROM LENOVO-SMI-MIB;
 
 lenovoProductsMIB MODULE-IDENTITY
-    LAST-UPDATED    "201705010000Z"
```
PipoCanaja, line 13: Are you sure you want to replace current mib file with an older one ? 
Renaming it properly is a good think, reverting to an older file usually not.

Fehler12: Ah, I did not check that, because for my test system (from 2019) the 2016er version was the most recent one and I assumed that the other file was older. Fixed!

murrant (approved):

murrant:
Thanks for getting this there @Fehler12

PR #13621, opened 2021-12-14
Allow device renaming even when the RRD files are on a remote rrdcached

Frazew:
Trying to rename a device on LibreNMS while using rrdcached on another host / distinct filesystem fails because LibreNMS cannot find the RRD folder to rename (`Rrd::dirFromHost` returns a file path that doesn't exist locally, which makes the call to `rename` fail).

As there is no way for LibreNMS to move a folder on a remote host, we can at least allow the device to be renamed in the database and warn the user that they should move the RRD folder themselves.

We detect that we're using a remote rrdcached by checking that the folder **does not exist** by path but exists by calling `Rrd::checkRrdExists`.

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -262,17 +262,22 @@ function getImageName($device, $use_database = true, $dir = 'images/os/')
 function renamehost($id, $new, $source = 'console')
 {
     $host = gethostbyid($id);
+    $host_rrd_dir = Rrd::dirFromHost($host);
+    $is_remote_rrd = false;
 
-    if (! is_dir(Rrd::dirFromHost($new)) && rename(Rrd::dirFromHost($host), Rrd::dirFromHost($new)) === true) {
+    if (! is_dir($host_rrd_dir) && Rrd::checkRrdExists($host_rrd_dir)) { // We're most likely using a remote rrdcached
+        $is_remote_rrd = true;
+    }
+    if ($is_remote_rrd || (! is_dir(Rrd::dirFromHost($new)) && rename(Rrd::dirFromHost($host), Rrd::dirFromHost($new)) === true)) {
         dbUpdate(['hostname' => $new, 'ip' => null], 'devices', 'device_id=?', [$id]);
         log_event("Hostname changed -> $new ($source)", $id, 'system', 3);
 
-        return '';
+        return ['message' => '', 'manual_rrd_rename' => $is_remote_rrd];
     }
 
     log_event("Renaming of $host failed", $id, 'system', 5);
 
-    return "Renaming of $host failed\n";
+    return ['message' => "Renaming of $host failed\n", 'manual_rrd_rename' => false];
 }
 
 function device_discovery_trigger($id)
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1939,8 +1939,13 @@ function rename_device(Illuminate\Http\Request $request)
     } elseif ($new_device) {
         return api_error(500, 'Device failed to rename, new hostname already exists');
     } else {
-        if (renamehost($device_id, $new_hostname, 'api') == '') {
-            return api_success_noresult(200, 'Device has been renamed');
+        $result = renamehost($device_id, $new_hostname, 'api');
+        if ($result['message'] == '') {
+            if ($result['manual_rrd_rename']) {
+                return api_success_noresult(200, 'Device has been renamed but the RRD folder will have to be renamed manually, are you running a remote rrdcached?');
+            } else {
+                return api_success_noresult(200, 'Device has been renamed');
+            }
         } else {
             return api_error(500, 'Device failed to be renamed');
         }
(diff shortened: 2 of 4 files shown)


includes/functions.php
```diff
@@ -262,17 +262,22 @@ function getImageName($device, $use_database = true, $dir = 'images/os/')
...
-    if (! is_dir(Rrd::dirFromHost($new)) && rename(Rrd::dirFromHost($host), Rrd::dirFromHost($new)) === true) {
+    if (! is_dir($host_rrd_dir) && Rrd::checkRrdExists($host_rrd_dir)) { // We're most likely using a remote rrdcached
+        $is_remote_rrd = true;
+    }
+    if ($is_remote_rrd || (! is_dir(Rrd::dirFromHost($new)) && rename(Rrd::dirFromHost($host), Rrd::dirFromHost($new)) === true)) {
         dbUpdate(['hostname' => $new, 'ip' => null], 'devices', 'device_id=?', [$id]);
         log_event("Hostname changed -> $new ($source)", $id, 'system', 3);
 
-        return '';
+        return ['message' => '', 'manual_rrd_rename' => $is_remote_rrd];
```
murrant, line 275: I suggest throwing an exception instead of returning an array.

Frazew: You're right, I just pushed new changes to change that.

includes/functions.php
```diff
@@ -262,11 +263,20 @@ function getImageName($device, $use_database = true, $dir = 'images/os/')
 function renamehost($id, $new, $source = 'console')
 {
     $host = gethostbyid($id);
+    $host_rrd_dir = Rrd::dirFromHost($host);
+    $is_remote_rrd = false;
```
murrant, line 267: All the logic here is very sketchy.  I can't seem to untangle it.

It looks like if $is_remote_rrd is true, it ALWAYS throws an exception.  If that is the case, why use a variable instead of just throwing the exception.

includes/functions.php
```diff
@@ -262,11 +263,20 @@ function getImageName($device, $use_database = true, $dir = 'images/os/')
...
-    if (! is_dir(Rrd::dirFromHost($new)) && rename(Rrd::dirFromHost($host), Rrd::dirFromHost($new)) === true) {
+    if (! is_dir($host_rrd_dir) && Rrd::checkRrdExists($host_rrd_dir)) { // We're most likely using a remote rrdcached
+        $is_remote_rrd = true;
+    }
+    if ($is_remote_rrd || (! is_dir(Rrd::dirFromHost($new)) && rename(Rrd::dirFromHost($host), Rrd::dirFromHost($new)) === true)) {
         dbUpdate(['hostname' => $new, 'ip' => null], 'devices', 'device_id=?', [$id]);
         log_event("Hostname changed -> $new ($source)", $id, 'system', 3);
 
+        if ($is_remote_rrd) {
+            throw new RemoteRrdRenameException('Host renamed but the associated RRD folder must be renamed manually');
```
murrant, line 277: Wouldn't changing the hostname without renaming the files cause data loss? (and new files to be created next poll)

murrant:
@Frazew Hey, I just saw your work to make TUNE work with rrdcached.  Nice work!

includes/functions.php
Frazew: I agree that it's not very clear... The logic was:

- If we can't the folder locally but Rrd says it exists, we're running with a remote rrdcached so we set `$is_remote_rrd` to true
- If we're running with a remote rrdcached, then the condition to perform the actual renaming will definitely fail because of the call to `rename`
- We therefore change the condition to pass if `$is_remote_rrd` is true, but we still want a way to warn the user that their RRD folder will have to be renamed manually to prevent data loss

To be clear: we need to throw the exception **after** the renaming in the database has succeeded, so we need the variable for the condition to pass. Hopefully this makes things clearer!

includes/functions.php
Frazew: Yes indeed, however currently the only way to rename a device when running with a remote rrdcached is to perform the SQL query manually **and** rename the folder on the rrdcached server.

The idea here is to remove the overly complicated step of manual SQL action, the folder renaming cannot be avoided but that's expected.

Frazew:
> @Frazew Hey, I just saw your work to make TUNE work with rrdcached. Nice work!

Thanks! When this change makes it into a rrdcached release, I'll come back here with a PR to update the LibreNMS docs accordingly!

Bierchermuesli:
> @Frazew Hey, I just saw your work to make TUNE work with rrdcached. Nice work!

huh, may I ask where? looking for a solution on this issue...

murrant:
> > @Frazew Hey, I just saw your work to make TUNE work with rrdcached. Nice work!
> 
> huh, may I ask where? looking for a solution on this issue...

https://github.com/oetiker/rrdtool-1.x/pull/1139

murrant:
This needs a rework.  Feel free to re-open.  Thanks!

Frazew:
Oh wow sorry I had completely forgotten about this PR. I'll reopen when I get some time to rework it, thanks!

PR #13626, opened 2021-12-17
Added support for user timezones and user browser timezone by default (incomplete)

eskyuu:
We have users who are in a different timezone to the NMS server, and wanted to allow users to select a timezone to view the graphs in.  This patch achieves this by adding a new user preference, and stores the timezone value in the session, then passes the TZ environment variable to rrdgraph so it will draw the graphs in the requested timezone.

This also modifies the default behaviour by using the browser timezone by default.  I realise that this may need some thought/discussion, and this default behaviour can be changed by reverting 2 changes in resources/views/auth/login.blade.php and resources/views/user/preferences.blade.php if desired (or even made into a config option on whether to use the system or browser timezone by default).

#### Please note

#### Testers


eskyuu:
I'm getting a PHP type error, and need help to see if there is a better place to assign a once-off session variable from a user preference at login.

murrant:
Shouldn't the default be the browser timezone?

That type mismatch is expected, easiest solution would be to update the phpstan baseline, but you could do something with the types on the methods to fix it too.

eskyuu:
The default is the browser timezone, which I thought would be the most logical choice. It's just a change from the existing behaviour of the front end.For the type mismatch, is there a preferred solution? Can you give an example code snippet of what you would expect the best solution to look like? I'm happy to test it on my Dev server and update the code as needed.On 17 Dec 2021 21:21, Tony Murray ***@***.***> wrote:
Shouldn't the default be the browser timezone?
That type mismatch is expected, easiest solution would be to update the phpstan baseline, but you could do something with the types on the methods to fix it too.

routes/web.php
```diff
@@ -198,6 +198,17 @@
     Route::permanentRedirect('demo', '/');
 });
 
+// routes that don't need authentication
+Route::group([], function () {
```
murrant, line 202: Don't really need the groups here, you can just set it on one line (or one group).

resources/views/auth/login.blade.php
```diff
@@ -9,3 +9,12 @@
     </div>
 </div>
 @endsection
+
+@section('scripts')
```
murrant, line 13: I'm not sure this is the right place for this.  It will only set it if the user visits the login page.  What about when the user remembers their login, then is in a different timezone?

This also requires the route to be unauthenticated.

Take a look at the set_resolution route, it is very similar to what you want to do.  (although it does need to be removed at some point :D)

routes/web.php
eskyuu: I've just simplified this config.

resources/views/auth/login.blade.php
eskyuu: I implemented it this way because there is no trigger for the timezone like there is for the screen resolution.  This means that to make this dynamic I would need to post the current browser timezone on every page load just in case it's changed, and then check to see if the user has hard-coded their timezone.  I don't think the additional load is worthwhile, and would expect people to want to see consistent timezones instead of it changing.  

I'm happy to hear if you have any further thoughts on eliminating the additional requests, or if you still think it's worth making it more dynamic.

I have had a second thought on this feature - should I be aiming to record a device timezone, and having graphs shown in the timezone local to the device?  This could become a global option to enable/disable showing graphs in the device timezone.

eskyuu:
I don't believe the php tests are failing based on my commit.  The only thing I can see is that on the 1st of December there was a commit (d74bb379710bb7356833dc1fe4ca28e4d883fb4c) that added the ports_vlans config to vrp_ac6605-26.json.  

Can you please confirm that I need to look at my code to fix the test errors?

murrant:
No, there was an upstream issue.

resources/views/auth/login.blade.php
murrant: IMO, everything should default to the browser timezone by default allowing the user to override.

You could definitely write the javascript in such a way that a post is only made when the browser timezone differs from what is given by the server (either the default server time or whatever is stored in the user session).

murrant: I think this is a very worthwhile thing to solve.  But it is not simple.  Mostly, because time in LibreNMS is a mess :)

eskyuu: How does the latest commit look?

eskyuu:
Can I please get some feedback on this request?

Jellyfrog:
This is definitely something we could like to fix/have consistency on.  
However this is a much deeper problem that I feel we sadly can't fix this way.  
I think the real fix lies within first making sure all dates are stored as UTC in the database, then writing some helper functions (use or a package such as; https://github.com/jamesmills/laravel-timezone) to display/convert.

eskyuu:
I don't really understand how your suggestion is different to my implementation, other than suggesting a different library (I used the existing Carbon library for the timezone conversions in PHP).  As I've demonstrated above, I'm happy to re-work if there is any detailed feedback, but find the latest feedback confusing.  It sound like you want to fix the problem, but want to use a different library and change the database timezone, instead of using the Carbon library and just converting to/from whatever timezone the database data is in.

For now, I'll maintain this as a manual patch on our systems, because we need this now, and the patch in this merge request works for us.  I'm interested to hear of any examples of where this patch does not handle timezones correctly, so I can either fix using the existing methods, or understand the deeper issues with timezones in the system and resolve the issues with this patch.

If you're looking at implementing an alternate solution, the issues I had to fix were:
1 - Making sure all displayed date filters are shown in the session timezone
2 - Making sure all dates in database queries are converted to the database timezone
3 - Making sure all database results are converted back to the sessiontimezone
4 - Getting rrdtool to generate the graphs in the session timezone (the easiest part of the patch - rrdtool does this natively)

murrant:
This honestly feels extremely incomplete, but I don't think it would conflict with converting stored dates to utc.

eskyuu:
Thanks for the feedback.  I don't know the code well enough to have an idea of how many places dates are displayed.  I can see how this could be a problem, and understand why this could be a block for merging my code.  Please also note that I'm happy to work on this ongoing one page at a time until you're happy that it is complete enough to merge.

Feel free to close this request without merging if you don't think this is the way forward for this request.

murrant (dismissed):

ciroiriarte:
This would be very interesting. We currently work with a team spread in different timezones, for the time being we work with UTC.

I wonder if this should/must include not only KPI graphs but things like syslog and events too.

eskyuu:
I noticed some activity/interest in this merge request, so I've rebased to master.  I'm not sure if the checks are failing due to my changes or something in master.

I agree with the comment about needing to provide consistent timezones in all parts of LibreNMS.  Please let me know of any places where times are not being adjusted, and I'll sort them out.

murrant:
Thanks, I was going to merge it this month, but when I re-ran tests, some failed :(

murrant (approved):

dagbdagb:
@eskyuu Thank you for this contribution! Can you suggest what I am doing wrong, because I do not see any effect of setting a user preference for the timezone.

Dumping the timezone variable appears to give me 'null'...

eskyuu:
We have just moved our environment, and this feature stopped working.  The fix was to run the following commands on the SQL server to load the timezone data and then update the global variable for MySQL to know the correct system timezone:
mysql_tzinfo_to_sql /usr/share/zoneinfo | mysql -u _user_ -p_password_ -h _host_ mysql
mysql -u _user_ -p_password_ -h _host_ -e "set @@global.time_zone=@@global.system_time_zone;"

I'm wondering if the verify script should be updated to check for this, or if the documentation should just be updated.  I could add a check around the user preference function to hide the option if the MySQL server's @@global.time_zone variable is set to "SYSTEM".

dagbdagb:
As far as I can see, events are logged with UTC. Which sounds right? I am probably missing an important point here, how does fixing/changing the timezone for mysqld impact the display of timestamps on the client side?

Tried summarizing (some) sources of (timestamp) confusion below

- the host running mysqld got a timezone
- mysqld may have its own timezone
- the host running the librenms poller may have its own timezone
- the host running the librenms front-end may have its own timezone
- the php environment on the front-end server may have its own timezone
- the client browser may have its own timezone, which comes from client system or a local user preference.
- the librenms user may set a preferrred timezone in the librenms config, which applies to display of times in librenms only

I have no illusions about having caught them all.

IMO:
Events in LibreNMS should be logged as UTC.
Timestamps in received syslog messages are whatever they are, and should not be trusted. And should be logged as UTC
Timestamps in SNMP traps  should not be trusted and logged as received with a UTC timestamp.

I am positive the LibreNMS developers have thought about this numerous times and in various degrees of desperation. 

Could we talk someone into adding the current state about timezones and timestamps to the Documentation, as well as suggest areas in need of improvement and the direction of said improvements?

murrant:
researching changing database access to be set to UTC...  LibreNMS timezone support is an utter mess :)

eskyuu:
> researching changing database access to be set to UTC... LibreNMS timezone support is an utter mess :)

If you're going to do this, I would think that the PHP and system timezones for all LibreNMS servers should also be set to UTC.

I have found the Unix epoch to be the nicest to work with.  Any time you want to store a time you just convert it to epoch, which is the same regardless of timezone.  Conversions back to a human readable time can be done to any timezone you like.  This is probably why I approached the issue as just needing to convert from the SQL date to the user's timezone.

@murrant  - As an immediate fix, do you think that a timezone check needs adding in the validation script for the MySQL server's timezone?  Or do you think the timezone adjusting feature should be disabled if the MySQL server does not have the timezone data loaded (and the documentation should be updated to tell people how to enable it)?

murrant:
@eskyuu the timezone can be set at the beginning of the SQL connection (ignoring the server setting).
As long as we convert to from UTC when storing (which some things in Laravel do for us automatically, it won't matter what timzone is set in PHP.

Going forward, it probably would be good to have a validation for UTC in the DB, but I don't see how we can fix all the unattended installs.

PR #13630, opened 2021-12-21
DELL iDrac improvements

Fehler12:
I've added status comments to the Dell iDrac yaml files. 
Now the sensors and their status are interpreded correctly:
image

However the overall system status seems not to be touched. If anyone can give a hint on how to mark the system as Warning or Critical (e.g. in the avalability map) that would be very helpful! See: 
image

Might also touch #13606 

#### Please note

- [systems already there] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers


includes/definitions/drac.yaml
```diff
@@ -2,11 +2,11 @@ os: drac
 text: 'Dell DRAC'
 icon: dell
 type: appliance
-mib_dir: dell
+mib_dir: drac
```
Jellyfrog, line 5: Mibs are ordered by vendor so dell is the correct value

Fehler12: in that case we need to delete the MIBs in the folder drac and copy the two files to the dell folder. Does this make sense as DELL has a lot of different MIBs. I don't know if they may interfere?

Fehler12: I just pushed an update with the correct folders

murrant:
If you want the overall status to reflect state sensor status, that would be a separate PR.  I have a feeling a lot of people might not like that because they really like the color green. :D

includes/definitions/discovery/drac.yaml
```diff
@@ -20,6 +20,8 @@ modules:
                         - { value: 4, descr: nonCritical, graph: 1, generic: 1 }
                         - { value: 5, descr: critical, graph: 1, generic: 2 }
                         - { value: 6, descr: nonRecoverable, graph: 1, generic: 2 }
+                    high_limit: 5
```
murrant, line 23: high_limit, warn_limit and other limit values are not valid for state sensors.   Generic values should be set instead.

Fehler12:
> If you want the overall status to reflect state sensor status, that would be a separate PR. I have a feeling a lot of people might not like that because they really like the color green. :D

Yep, but I defeinitly want to know if e.g. a hard drive is failing. And if a hard drive or fan is broken, for me the monitoring must not be green. Can you give me a hint on how to implement that? I have no idea yet?

mpikzink:
It should be sufficient when you create a corresponding alert rule, shouldn't it?

Fehler12:
No, not really. For some reason those (predefined) rules don't work: 
image
despite the OIDs seem to be correct.

However if I define e.g. the mentioned limits in the device tab and use the generic rule sensor over/under limit it works. 
And I find it much more intuitive if critical values, which are defined in the yaml file for these device types, become monitored automatically. Therefore I added the attributes high an low limits in order to archieve automatic correct values.
Unfortunately the generic stats seem to be only in use for making the background of the value green/orange/red or grey. I did not found any alert rule which triggers for e.g. state generic value 2:
                        - { value: 5, descr: critical, graph: 1, **generic: 2** }
                        - { value: 6, descr: nonRecoverable, graph: 1, **generic: 2** }

In my opinion this is a not very comfortable situation for servers at the moment. Thats what I would like to change.

mpikzink:
I think the predefined sensor values are wrong. The generic sensor values should be queried, at least that's how I do it...

mpikzink:
Screenshot_2021-12-25_21-08-34
This is how I handle all sensor states.

includes/definitions/discovery/drac.yaml
Fehler12: But they work :). And if I define warning values manually if have to enter these values as well.

image
There is no way that I define that a specific predefined value should trigger a warning event.

Maybe there is a general logic problem with the static values?

Fehler12:
> Screenshot_2021-12-25_21-08-34 This is how I handle all sensor states.

This is working, just tested this. However the predefined rules do not match (state sensor warning should also match, this is a predefined rule). Maybe because the overall system is green?

image

All over all not very intuitive! Any ideas on how to make this more intuitive? E.g. make the predefined rules work?

PipoCanaja:
Warning is "generic=1". And Critical is "generic=2". So it is expected to have only Critical and not Warning if your sensor generic value is 2. What behaviour would you expect to see in that case ?

includes/definitions/discovery/lenovoXClarityIMPI.yaml
```diff
@@ -0,0 +1,68 @@
+mib: LENOVO-XCC-MIB:LENOVO-XCC-ALERT-MIB:LENOVO-SMI-MIB:LENOVO-PRODUCTS-MIB
```
PipoCanaja, line 1: Could you create a dedicated PR for this lenovoXClarityIMPI discovery file ?

PipoCanaja (changes requested):

includes/definitions/discovery/lenovoXClarityIMPI.yaml
PipoCanaja: Oups, seems you already did. So please just remove it from this PR :)

Fehler12:
> Warning is "generic=1". And Critical is "generic=2". So it is expected to have only Critical and not Warning if your sensor generic value is 2. What behaviour would you expect to see in that case ?

My problem is that the predefined alert rules don't work because the rule also checks if the overall device is in a critical state which is not true for state sensors.

I would like to make a suggestion: would it be possible to add a new parameter "critical_for_system_status"? If this value is set to true the sensor would influence the overall system status. For examle I would set this parameter for the overall system status for all server systems. If this status becomes warning, the device would also be yellow in the availablity dashboard and the predefined alert rule would trigger:
image
If so I would withdraw this change request and try to implement this. Could anyone give a hint where to implement this?

PipoCanaja:
Predefined alert rules are only a hint. But you can create your alert rule that will trigger exactly when you want it, without any need to change anything in the current meaning of the colors.

PR #13642, opened 2021-12-31
Added the option to show or hide unreachable devices in the alert widget

eskyuu:
I have added an option on the alerts widget for unreachable hosts (hosts with parent(s) set where all parents have a device status of "0".  It can be used to stop the alert widget from having too much noise from a single site with multiple devices being offline.

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Widgets/AlertsController.php
+++ b/app/Http/Controllers/Widgets/AlertsController.php
@@ -43,6 +43,7 @@ class AlertsController extends WidgetController
         'sort' => 1,
         'hidenavigation' => 0,
         'uncollapse_key_count' => 1,
+        'unreachable' => null,
     ];
 
     public function getView(Request $request)
--- a/includes/html/table/alerts.inc.php
+++ b/includes/html/table/alerts.inc.php
@@ -45,7 +45,7 @@ if (is_numeric($vars['fired'])) {
 }
 
 if (is_numeric($vars['unreachable'])) {
-    # Sub-select to flag if at least one parent is set, and all parents are offline
+    // Sub-select to flag if at least one parent is set, and all parents are offline
     $where .= ' AND (SELECT IF(COUNT(`dr`.`parent_device_id`) > 0 AND COUNT(`dr`.`parent_device_id`)=count(`d`.`device_id`),1,0) FROM `device_relationships` `dr` LEFT JOIN `devices` `d` ON `dr`.`parent_device_id`=`d`.`device_id` AND `d`.`status`=0 WHERE `dr`.`child_device_id`=`devices`.`device_id`)=' . $vars['unreachable'];
 }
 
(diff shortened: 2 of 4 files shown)


ottorei:
Might be a bit confusing to describe them as unreachable. A device could be unreachable if it is down itself for any reason. Have you considered wording this as something like "Hide alerts when depended devices are down"?

While this might clearly be useful option in some cases, the alerts tab would still list all invidual devices. The widget would probably still list invidual devices for a while if the child was polled before the parent. They would probably also cause any configured alerts / transports to trigger. A more generic approach could probably be to enhance the poller to check the connectivity to one or more parents and changing it's status before polling the child device (maybe several possible options for this like 1) polling order 2) check one or more parents with fastping).

Anyway, I think this PR enhances the functionality of the widget itself just fine and is OK - just pointing out that a more general approach should be implemented at some point to the poller.

eskyuu:
The term unreachable comes from the nagios/icinga monitoring systems 
where devices are down if they are offline but the devices they depend 
on are online, and unreachable if they are offline and the devices they 
depend on are also offline.  I'm happy to change the wording if you 
still think it's clearer, but I'm aware of the limited space in the 
textbox.  I would imagine that an update to the documenation could clear 
up any confusion if you're happy with my explanation of unreachable, as 
well as the use in other monitoring systems.

Yes, this is the same as the "Show acknowledged" option, where the 
alerts tab acts independently of the widget.  I don't think this is an 
issue.

The questions around the polling and alerting are good, but as you've 
noted this PR is not attempting to address any of these, and just builds 
on top of the dependencies used by the alerter.

After reading tbe above do you still think I should change the wording 
on the option?

ottorei:
Well, the wording is fine if it is consistent and usage of the feature is documented.

murrant:
I think unreachable is a bit misleading too.  Perhaps "Hide devices without a parent".

eskyuu:
I've just updated the alert settings page to give a clear description

murrant (dismissed):

eskyuu:
This failed to merge on the final checks because some of the test data has changed.  I've rebased, so can you please re-approve.

laf:
I think I have to agree that this should just be a checkbox to turn on 'Hide alerts for devices with parents that are down' so you only have two options:

option 1) All devices are shown in the alerts
option 2) Only parent devices are show if they are down, or devices that have no parent that is down (it actually would be good to highlight this is a parent device and how many child devices are also down).

eskyuu:
> I think I have to agree that this should just be a checkbox to turn on 'Hide alerts for devices with parents that are down' so you only have two options:
> 
> option 1) All devices are shown in the alerts option 2) Only parent devices are show if they are down, or devices that have no parent that is down (it actually would be good to highlight this is a parent device and how many child devices are also down).

My current use case for this option is to have 2 widgets on the dashboard - one to show all the parents that are offline, and another to show all the children.  The point of having both is to quickly see all the urgent alerts on one widget, while still being able to see the effect.  This means that the tickbox option would not suit my current needs, so I do not agree with the tickbox suggestion.

I agree that adding another column to show how many devices are affected would be very useful.  I can have a look at adding another column with this additional info.

eskyuu:
Having said this, I'm not sure when I will get a chance to look at adding the extra column.  Please let me know how important this additional feature is?

eskyuu:
It looks like this has been merged elsewhere

PR #13648, opened 2022-01-03
fix: Add consistent output of name and app_id to Poller for all Applications (fixes #13641)

### LINKED ISSUE #13641: os-update: Title missing in poller output
### The problem

FYI, noticed the other day that os-updates does not show the title in the poller output. Not major, but also makes checking the output a bit problematic. Still trying to find why this is the issue (i.e. missing), but raising it here so others are aware. FYI, the output from applications, from poller.php,
```
#### Load poller module applications ####
 apache: UUUUUUU.............
: .
SMART: .........................
 ups-nut: ......................
```

### Output of ./validate.php

```txt
Not understanding the daily.sh warning - ran daily to update (it ran fine), still seeing that note. But that aside,

Component | Version
--------- | -------
LibreNMS  | 21.12.0-49-g0543bbefa
DB Schema | 2021_11_29_165436_improve_ports_search_index (229)
PHP       | 8.0.8
Python    | 3.9.7
MySQL     | 10.5.13-MariaDB-0ubuntu0.21.10.1-log
RRDTool   | 1.7.2
SNMP      | 5.9

[OK]    Composer Version: 2.2.2
[WARN]  Your install is over 24 hours out of date, last update: Tue, 28 Dec 2021 02:36:04 +0000
        [FIX]:
        Make sure your daily.sh cron is running and run ./daily.sh by hand to see if there are any errors.
        [FIX]:
        Modified Files:
         LibreNMS/Util/Url.php
         html/.htaccess
```
```


_No response_

_No response_

westerterp:
This PR fixes not only #13641, but standardizes the way the output of each Application polling starts.
Each Application will now first output the Application `name` and `app_id` before any application-specific output starts.

Every Application has been checked. There was quite some inconsistency.
- If no `echo` statement was found, it was added
- Some Applications only printed the name, but not the `app_id`, the `app_id`'s have been added to the `echo`
- The `echo` now always uses the `$name` variable and not hardcoded values
- The use of `d_echo` debug function was removed.

#### Please note

#### Testers

### DIFF
--- a/includes/polling/applications.inc.php
+++ b/includes/polling/applications.inc.php
@@ -4,12 +4,12 @@ $app_rows = dbFetchRows('SELECT * FROM `applications` WHERE `device_id`  = ?', [
 
 if (count($app_rows) > 0) {
     foreach ($app_rows as $app) {
-        echo "Application: ". $app['app_type']. ", app_id=" . $app['app_id'];
+        echo "Application: ". $app['app_type']. " , app_id: " . $app['app_id'];
         $app_include = \LibreNMS\Config::get('install_dir') . '/includes/polling/applications/' . $app['app_type'] . '.inc.php';
         if (is_file($app_include)) {
             include $app_include;
         } else {
-            echo ' ERROR: ' . $app_include . ' include file missing!';
+            echo 'ERROR: ' . $app_include . ' include file missing!';
         }
     }
     echo "\n";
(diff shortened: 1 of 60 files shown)


murrant:
I like the idea, but wouldn't it be better to output this in applications.inc.php so we don't have to modify every application?

westerterp:
Thanks for the suggestion!
That would make a lot more sense indeed.
Although if we do that, we won't have access to the Application `name`, but this was solved by using the `app_type` from the database

laf (approved):
LGTM

PR #13653, opened 2022-01-05
Fix horizontal scrollbar in device list always visible

efelon:
#### Please note

- [ ] ~If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).~

#### Testers

#### Description

Using the class `row` for each icon line [introduced](https://github.com/librenms/librenms/commit/94ee737f3d158a57efec280139852817eb929244) an overflow coming from the "margins" which always shows the scroll bar:
image

I propose to remove the class `row` completely to:
* fix the horizontal scroll bar
* compact the rows which allows for more devices to be listed in the same space
(the icons themself will get even closer, but are too small anyway to be suited for good handling without a mouse pointer)

image

### DIFF
--- a/resources/views/device/actions.blade.php
+++ b/resources/views/device/actions.blade.php
@@ -1,6 +1,6 @@
 <div class="container-fluid" style="padding-left: 0; padding-right: 0;">
     @foreach($actions as $row)
-        <div class="row" style="padding-bottom: 5px">
+        <div>
             @foreach($row as $action)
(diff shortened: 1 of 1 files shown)


efelon:
I noticed that in the dark mode the header is not centered.
image

It seems that the class `.device-table-header-actions` is not used for some reason. I couldn't figure that out.

murrant (approved):

Issue #13647, opened 2022-01-03
Ubiquiti AirOS Temperature Scale


JKJameson:
### The problem

Temperature is reported as 385.88 M°C when expected temp is about 30 to 40c

![Untitled](https://user-images.githubusercontent.com/275770/147987365-e49f1a39-b432-4eff-b560-cbe0a3004df2.png)

### Output of ./validate.php

```txt
====================================
Component | Version
--------- | -------
LibreNMS  | 21.12.0-4-g4b64d33d1
DB Schema | 2021_25_01_0129_isis_adjacencies_nullable (229)
PHP       | 7.4.25
Python    | 3.9.2
MySQL     | 10.5.12-MariaDB-0+deb11u1
RRDTool   | 1.7.2
SNMP      | 5.9
====================================

[OK]    Composer Version: 2.2.3
[OK]    Dependencies up-to-date.
[OK]    Database connection successful
[OK]    Database schema correct
```

aztec102:
Hello!
Apparently, Ubiquiti used to send via OID Interger (0 ... 255). Now, on the new AirOS 8, they began to give values more, which must be divided by 10.

JKJameson:
AirOS 6.3.6 does not seem to poll temperatures. Nothing in the GUI, also nothing in the poller debug log. This is fine - it doesn't throw an erroneous alert.


AirOS 8.7.5
`Checking (snmp) temperature Temperature... 
335.54432 C
RRD[last 10.2.0.103/sensor-temperature-airos-0.rrd  --daemon unix:/run/rrdcached.sock]  
RRD[update 10.2.0.103/sensor-temperature-airos-0.rrd N:335.54432 --daemon unix:/run/rrdcached.sock]  
`

Adding a divisor of 10000000 does the trick for me. As the temperature isn't detected for AirOS 6 and presumably earlier, I am assuming this is fine, but have no earlier devices to test.

https://github.com/librenms/librenms/pull/13655

murrant:
So, looks like 8.7.4 and 8.7.5 have the bad values... Probably the fix is to add two temperature sensor entries and skip them for os versions where it doesn't apply.   Very cool stuff Ubiquiti.

JKJameson:
It's very unfortunate that I started using LibreNMS with AirOS firmware from 8.7.4 or I would have picked up that the specific update broke this.

https://community.ui.com/releases/airMAX-AC-8-7-4/6b38ecfe-8486-40c0-9be5-5fb168c7a010

[...]

PR #13655, opened 2022-01-05
Fix Temperature Sensor for AirOS 8.7.4+

JKJameson:
https://github.com/librenms/librenms/issues/13647

#### Please note

#### Testers


Jellyfrog:
This doesn't look correct according to the test data we got

JKJameson:
I can't seem to get the tests to pass. Could someone share some insight as to why this is failing? This whole process is a bit convoluted for a simple number change which fixes things on my side.

Jellyfrog:
> I can't seem to get the tests to pass. Could someone share some insight as to why this is failing? This whole process is a bit convoluted for a simple number change which fixes things on my side.

Because according to our test data your fix does not seem valid.
You're trying to divide 66 with 10000000.
Maybe it something that changed in newer firmwares? I'd recommend you find the latest MIB and check if there is some clue there

JKJameson:
LibreNMS was recording the temperature as 450 Million degrees Celsius before. Here's the graph with my patch applied.  Good luck trying to get any updated MIB from Ubiquity themselves. If I can contribute anything from the devices please let me know.

Hopefully I'm not the only person monitoring AirOS devices and someone else can chime in with their findings too. Clearly whatever test data you have is no longer valid.

image

image

murrant:
Stupid stuff.  Almost all my airos devices show normal temperatures, except 5.  I think it is airos 8.7,  setting this for all airos would probably break more than it fixes.

murrant:
After looking closely, only 8.7.4 has this issue.  (We don't have any at 8.7.5)   Can you confirm if it reports these bad values on 8.7.5?

bile0026:
I can confirm 8.7.5 reports bad values. Just updated our radios a couple weeks ago and got slammed with temp alerts. Had to shut off temp alerting for now.

murrant (approved):

bile0026:
Looks like ubiquiti noticed their oops and possibly fixed it in 8.7.7. Going to find a time to upgrade one of my test radios and see. So this might either need logic for the affected versions only. 

image

bile0026:
> What a joke. My next network will be Mikrotik powered. A big thanks to @murrant for the effort regardless.

Generally I'm a fan of their Airmax gear but UniFi and edge switch/route is less than desirable. They have been slipping though. MikroTik all the way for route/switch and I'm starting to like their wireless more and more. They might not have the latest Wi-Fi 700 but it's pretty darn solid. 

I also echo the thanks to @murrant

murrant:
HAHA, I suspected that would happen.  I can easily update the code to only do 8.7.4 to 8.7.6

bile0026:
Confirmed, airos 8.7.7 fixes the snmp issue. LibreNMS version 22.1.0.

PR #13666, opened 2022-01-11
added oids.no_bulk os setting

Npeca75:
in LNMS there is only 2 possibilty

use bulkwalk 
or drop bulkwalk and use plain walk for complete OS

this PR try to address gap between these two option
it would be ideal if we could specific only certain MIBs for plain walk and rest is going on full speed bulkwalk

#### Please note

#### Testers


Npeca75:
lnms1

murrant:
I feel like the nobs are not at the right level for this.  What if I have a device that cannot bulk walk IF-MIB::ifTable? Does that mean ALL devices are now set to normal walk IF-MIB if this is set.

This seems like it should be a combination of the OS and the OID (also an example of out of order walks here). Like:
includes/definitions/bados.yaml
```yaml
...
snmp:
    no_bulk:
        - IF-MIB::ifTable
    out_of_order_oids:
        - IP-MIB::ipNetToPhysicalPhysAddress
```

Or something like that.

Npeca75:
> I feel like the nobs are not at the right level for this. What if I have a device that cannot bulk walk IF-MIB::ifTable? Does that mean ALL devices are now set to normal walk IF-MIB if this is set.

hi @murrant 
yes, i understood your oppinion. Let me explain from my point of view. With full respect of your work for community
i see you force YAML. there is already snmp_bulk: false for independent sensor group

1. there is lot of old code / OS witch does not have  YAML and will never had one
2. it is unreal to expect that all code could be ported to YAML, like sensors which are not ending with indexes
3. i newer see "standard" MIB which will timeout on any device. They are vendor specific. Like TPLINK-DDM, Radlan-PHY, PoE etc which are problematic

so, my idea was, there is for ex problematic MARVELL-POE-MIB
it will be problem on any device, not only on Eltex. These devices are old with weak CPUs
or TPLINK-DDM, same story, no matter how fresh is device, it will slow down (almost break) with DDM bulkwalk
this way, if we set globaly 10-15 problematic MIBs, old code will benefit from this approach
to repeat myself, they are vendor specific MIBs, with vendor specific (unique) names

and your observation, not for a nobs ... well, snmp timeouts, mibs, monitoring, etc is out of scope for nobs :)

but, you are a core developer, so if you reject it this way, OK

i have no other solution, because setting "force_walk" for specific OS is only doable in YAML which lead us back, not every OS have YAML. 

extending old "gen_snmp_walk" function with new "forcewalk" param ? maybe
extending your SnmpQuery, for ex ->forcewalk(1) ? doable
after all, these two "extending" will help with PHP code, and YAML is already covered

i don't know, maybe you are right
this was an attempt for global solution for problematic MIBs in visual way, and yes, for advanced users who press DEBUG button and see that XXX-MIB will timeout and add this mib from GUI to "blacklist"

waiting for your answer
BR Peca

Npeca75:
@murrant 

as you suggested
per OS

tested with config.php
$config['os']['eltex-mes23xx']['mibs_forcewalk'] = 'MARVELL-POE-MIB,RADLAN-PHY-MIB';

Npeca75:
Tested with YAML

-----
mibs_forcewalk: IPv6-MIB,Q-bridge-mib
-----

SNMP: Q-bridge-mib forced to snmpwalk  
SNMP['/usr/bin/snmpwalk' '-M' '

SNMP: IPV6-MIB forced to snmpwalk  
SNMP['/usr/bin/snmpwalk'

Jellyfrog (changes requested):
The variable naming is confusing for me, such as `$forcewalk` means "don't use bulk version".
Then we have `$flwalk`, `$wkeys` and `$mnames` which are not very obvious what they represent.

includes/snmp.inc.php
```diff
@@ -130,7 +130,19 @@ function gen_snmpget_cmd($device, $oids, $options = null, $mib = null, $mibdir =
...
+    if (isset($cnfsnmp['no_bulk'])) {
+        foreach ($cnfsnmp['no_bulk'] as $wkey => $mnames) {
+            if (Str::contains(strtoupper($mib), strtoupper($mnames))) {
+                $flwalk = true;
+                d_echo("\n" . 'SNMP: ' . $mib . ' forced to snmpwalk');
+            }
+        }
+    }
+
+    if ($device['snmpver'] == 'v1' || (isset($device['os']) && Config::getOsSetting($device['os'], 'snmp_bulk', true) == false) || $flwalk) {
```
Jellyfrog, line 145: `|| $flwalk` seems to be different from the logic from NetSnmpQuery

Jellyfrog:
Please also add docs and modify a device to use it (so we can see it works)

Npeca75:
> Please also add docs and modify a device to use it (so we can see it works)
ok @Jellyfrog 

tested with:
/opt/librenms/includes/definitions/eltex-mes23xx.yaml

added
snmp:
    no_bulk:
        - RADLAN-PHY-MIB
        - RADLAN-Physicaldescription-MIB
        
-------------------
result from snmp.inc.php:
Pre-cache eltex-mes23xx: 
rlPhyTestGetResult 
SNMP: RADLAN-PHY-MIB forced to snmpwalk  

SNMP['/usr/bin/snmpwalk' '-v2c' '-c' 'COMMUNITY' '-Osqn' '-m' 'RADLAN-PHY-MIB'

------------------
result from SnmpQuery:
SNMP: RADLAN-Physicaldescription-MIB forced to snmpwalk  

SNMP['/usr/bin/snmpwalk' '-v2c' '-c' 'COMMUNITY' '-OteQUs' '-Ih' '-m' 'RADLAN-Physicaldescription-MIB:RADLAN-HWENVIROMENT:RADLAN-PHY-MIB:RADLAN-rndMng:ELTEX-MES-HWENVIROMENT-MIB' '-M' '/opt/librenms/mibs:/opt/librenms/mibs/radlan:/opt/librenms/mibs/eltexmes23xx' 'udp:HOSTNAME:2004' 'rlEnvMonFanStatusTable']

includes/snmp.inc.php
Npeca75: var names and SnmpQuery logic modified
hope now is OK

Npeca75:
> Please also add docs 
could you please point me in which DOC file should i put these changes ?
thank you

Npeca75:
tested on Luminato 4x4 with TELESTE-COMMON-MIB
device will benefit from this setting because older Luminatos stuck sometime with detecting state sensors
lnms1

murrant:
@Npeca75 I do not agree that MIB is the correct thing to key off of, it should be OID.  (Same argument as LibreNMS only supports bulk or no_bulk for the entire device)

Only affect walks using the format "MIB::oid".  This is the preferred format for LibreNMS going forward.  If a call does not use that format, it should be updated to.

I thought about the config location more and it should be `snmp.oids.no_bulk` (and `os.<os>.oids.no_bulk`)

murrant:
@Npeca75 Sorry, Probably a failure of communication on my side, plus sometimes I don't get a full view of the code when reviewing.  I assure you, you are doing a good job.  I'm just looking at the code with long-term maintainability in mind.

murrant:
> tested on Luminato 4x4 with TELESTE-COMMON-MIB device will benefit from this setting because older Luminatos stuck sometime with detecting state sensors
> 
> lnms1

FYI, this forced UCD-SNMP-MIB::laLoadInt to use snmpwalk.  I don't think that was intended.

Npeca75:
> @Npeca75 Sorry, Probably a failure of communication on my side, plus sometimes I don't get a full view of the code when reviewing. I assure you, you are doing a good job. I'm just looking at the code with long-term maintainability in mind.

tnx @murrant 
nice code

will wait for this PR and 
https://github.com/librenms/librenms/pull/13671
to be accepted, then will send new eltex YAMLs with "slow down" MIBs
thank you again for helping about this case

PR #13668, opened 2022-01-11
New OS: Adva XG 304

h-barnhart:
This PR introduces the Adva XG300 series to LibreNMS, specifically the XG304 model variant. Includes basic device identification and information, port data, optical device health graphs, and device temperature sensors. Haven't been able to capture any meaningful traps yet, I will probably do that later.

The Adva XG300 series is based on an acquisition of MRV Networking, so the OS is completely different from the other Adva devices currently in LibreNMS. Under the hood its Linux with a Cisco IOS-like CLI on top. MIBs are bit of MRV, generic, and Adva. I think I've only added the MIBs required for discovery and polling of the device as required for the ports and sensors in this PR.

#### Please note

#### Testers


LibreNMS/OS/AdvaXg300.php
```diff
@@ -0,0 +1,46 @@
...
+namespace LibreNMS\OS;
+
+use App\Models\Device;
+use LibreNMS\Interfaces\Discovery\OSDiscovery;
+use LibreNMS\OS;
+
+class AdvaXg300 extends OS implements OSDiscovery
+{
+    public function discoverOS(Device $device): void
+    {
```
Jellyfrog, line 35: Would you mind implementing this in yaml instead? You can use regex to get the fields

PipoCanaja (dismissed):
LGTM

laf:
This might have changed since I was previously involved in LIbreNMS but OS' are usually not device range specific, do the Adva XG 300 range use a firmware/os that doesn't exist on any other XG ranges?

h-barnhart:
> This might have changed since I was previously involved in LIbreNMS but OS' are usually not device range specific, do the Adva XG 300 range use a firmware/os that doesn't exist on any other XG ranges?

Yes, the XG300 series is different. In the current version of LibreNMS the XG300 series is discovered as a generic device, it is neither identified as Adva or MRV.

The XG100 and XG200 series were developed by Adva and run their in house OS which has a usable web UI and an "interesting" CLI. The XG300 series comes from an acquisition of MRV Networks and runs their OS. The CLI is IOS like and there is no a web UI.

includes/definitions/discovery/adva_xg300.yaml
```diff
@@ -0,0 +1,96 @@
+mib: OA-SFP-MIB:DEV-CFG-MIB
+modules:
+    os:
+        sysDescr_regex: '/(?<hardware>FSP150-XG3\d+\s\w+\s\w+-\w) Ver\. (?<version>[\d+_]+\d \(\d+\)$)/'
```
Jellyfrog, line 4: Maybe nitpicking but I'd suggest using something a bit more ... forgiving:
```suggestion
        sysDescr_regex: '/(?<hardware>.+) Ver\. (?<version>[^)]+\))/'
```

h-barnhart: Ok, I could've clicked a button and fixed that. Oh well, I put it in.

PR #13673, opened 2022-01-12
Add supervisord application

hvanoch:
This PR adds stats for [supervisord](http://supervisord.org/). It will show the total number of processes per status. It also adds the up-time and state (number used by supervisor) per process. This up-time can be useful to detect if a process is stuck if we would expect to only run for a certain amount of time.

This allows to set alerting on certain supervisor statuses.

PR for the agent: https://github.com/librenms/librenms-agent/pull/392

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Applications.md
+++ b/doc/Extensions/Applications.md
@@ -137,6 +137,7 @@ by following the steps under the `SNMP Extend` heading.
 1. [Seafile](#seafile) - SNMP extend
 1. [SMART](#smart) - SNMP extend
 1. [Squid](#squid) - SNMP proxy
+1. [Supervisord](#supervisord) - SNMP extend
 1. [TinyDNS/djbdns](#tinydns-aka-djbdns) - Agent
 1. [Unbound](#unbound) - SNMP extend, Agent
 1. [UPS-nut](#ups-nut) - SNMP extend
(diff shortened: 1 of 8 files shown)


SourceDoctor:
nice work
but i don't see in docs any dependencies for Application listed which are needed.
i think about xmlrpc and supervisor module

hvanoch:
Well, supervisor is actually a python module. If I take a look at the supervisor executable it's just running the supervisor module.  So I think it is safe to consider that dependency to be already available on that system. 
```
#!/usr/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'supervisor==3.4.0','console_scripts','supervisord'
__requires__ = 'supervisor==3.4.0'
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.exit(
        load_entry_point('supervisor==3.4.0', 'console_scripts', 'supervisord')()
    )
 
```
For the xmlrpc dependency, this is a default python module.
https://docs.python.org/2.7/library/index.html

I noticed I only added support for python2. So I updated the agent to make it also compatible with python3.

Jellyfrog (approved):

Jellyfrog:
Congratulations on your first pull request! Tests and all on first try! :partying_face:

PR #13679, opened 2022-01-16
Check for empty SLA tags so we skip bad entries

laf:

Fixes these errors due to bad tag data from Cisco SLAs:

```
Error discovering slas module for x.x.x.x. SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'tag' cannot be null (SQL: insert into `slas` (`deleted`, `sla_nr`, `owner`, `tag`, `rtt_type`, `rtt`, `status`, `opstatus`, `device_id`) values (0, 13, , ?, echo, 5, 1, 0, 119))
```

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/OS/Shared/Cisco.php
+++ b/LibreNMS/OS/Shared/Cisco.php
@@ -319,14 +319,14 @@ class Cisco extends OS implements OSDiscovery, SlaDiscovery, ProcessorDiscovery,
 
         foreach ($sla_data as $index => $sla_config) {
             $sla_tag = $this->getSlaTag($sla_config);
-            if (empty($sla_config['rttMonCtrlAdminRttType']) || empty($sla_tag)) {
+            if (empty($sla_config['rttMonCtrlAdminRttType'])) {
                 continue; // skip garbage entries
             }
 
             $slas->push(new Sla([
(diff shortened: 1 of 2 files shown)


murrant:
Is skipping what we want or do we want to write an empty string.  (I don't know what data the tags field holds)

ottorei:
> Is skipping what we want or do we want to write an empty string. (I don't know what data the tags field holds)

Its varchar(255) in the DB. The values in that field are mostly IP-addresses (* tested with some random Cisco gear).

murrant:
> > Is skipping what we want or do we want to write an empty string. (I don't know what data the tags field holds)
> 
> Its varchar(255) in the DB. The values in that field are mostly IP-addresses (* tested with some random Cisco gear).

So, should we discard the ones with empty tags, or keep them?

bonzo81:
Hope I can chip in. We use IP SLA quite a bit. We tend to use the tag like a description. I think it's just free text. Cisco guide explains it as 

> Creates a user-specified identifier for an IP SLAs operation.

We have some configured without the tag so I would think it best to keep those too, as they still produce SLA data.

murrant:
Thanks @bonzo81, if that is the case I suggest we store empty strings.

PR #13689, opened 2022-01-19
Replace Requests with HTTP Client

Jellyfrog:

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Util/GitHub.php
+++ b/LibreNMS/Util/GitHub.php
@@ -27,8 +27,7 @@
 namespace LibreNMS\Util;
 
 use Exception;
-use Requests;
-use Requests_Response;
+use Illuminate\Support\Facades\Http;
 
 class GitHub
 {
@@ -112,9 +111,9 @@ class GitHub
      */
     public function getRelease($tag)
     {
-        $release = Requests::get($this->github . "/releases/tags/$tag", $this->getHeaders());
+        $release = Http::withHeaders($this->getHeaders())->get($this->github . "/releases/tags/$tag");
 
-        return json_decode($release->body, true);
+        return $release->json();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


LibreNMS/Util/GitHub.php
```diff
@@ -391,14 +389,14 @@ public function createChangelog($write = true)
             $this->getPullRequest();
         }
 
-        if (! isset($previous_release['published_at'])) {
+        if (! isset($previous_release['created_at'])) {
```
Jellyfrog, line 392: @murrant I think this was a bug.

>The created_at attribute is the date of the commit used for the release, and not the date when the release was drafted or published.

We want to compare from the tag date rather than the "github release" date

Jimmyord:
Work a bit on these (Oxidized only) : https://github.com/Jimmyord/librenms/commit/b0dc4da58f2e2858e46ee792c1a354826841b551

Add some changes to Proxy to keep avoiding sending localhost call to proxy.

Jellyfrog:
> Got Oxidized part working fine but find some error in Proxy module (out from our changes) :
> 
> Regex Improvement : function shouldBeUsed match a regex //:(localhost|127\.|::1) This assume user write URL starting with http... but in http call url is like http:// (and not //:). So regex will never match. This regex could be better (?:^|://)(localhost|127\.|::1) (Patch + Handling my own lazyness to not mention http://)
> 
> Naming Improvement : ($target_url && ! self::shouldBeUsed($target_url)) should be ($target_url && self::shouldBeUsed($target_url)).
> 
> The name of the function is a bit messy : shouldBeUsed return True when regex match but when match occured (localhost call triggered), we should avoid using Proxy.
> 
> Maybe using AvoidProxy for function name could help to clarify this (don't forget to delete ! in if condition).

https://github.com/librenms/librenms/pull/13702

murrant (approved):
LGTM

murrant:
@Jellyfrog Needs a quick lock file update.

murrant (approved):

PR #13706, opened 2022-01-25
Remove non-working Dell specific alert rules from the collection

Fehler12:
Removed non-working Dell alert rules and removed constraints fo state_sensor_warning in order to make this rule working

#### Please note

- [no changes] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.
- [no changes here] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers

### DIFF
--- a/misc/alert_rules.json
+++ b/misc/alert_rules.json
@@ -85,7 +85,7 @@
     "severity": "critical"
   },
   {
- "rule": "macros.state_sensor_critical && sensors.sensor_alert = 1"
+    "rule": "macros.state_sensor_critical && sensors.sensor_alert = 1",
     "name": "State Sensor Critical",
     "default": true
   },
--- a/misc/alert_rules.json
+++ b/misc/alert_rules.json
@@ -85,12 +85,12 @@
     "severity": "critical"
   },
   {
-    "rule": "macros.state_sensor_critical && sensors.sensor_alert = 1",
+    "rule": "macros.state_sensor_critical",
     "name": "State Sensor Critical",
     "default": true
   },
   {
-    "rule": "macros.state_sensor_warning && sensors.sensor_alert = 1",
+    "rule": "macros.state_sensor_warning",
     "name": "State Sensor Warning"
   },
   {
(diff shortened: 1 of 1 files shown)


misc/alert_rules.json
```diff
@@ -85,12 +85,12 @@
     "severity": "critical"
   },
   {
-    "rule": "macros.state_sensor_critical && sensors.sensor_alert = 1",
```
ottorei, line 88: Did this cause some kind of issues? This should be working fine if the sensor has the alert-flag set. This option allows ignoring sensors per device.

misc/alert_rules.json
```diff
@@ -313,30 +313,6 @@
...
-    "name": "Dell Server Disk Array State Failed/Degraded"
-  },
-  {
-    "rule": "sensors.sensor_current ~ \"[5|6]\" && sensors.sensor_oid = \".1.3.6.1.4.1.674.10892.1.600.12.1.5\"",
-    "name": "Dell Server PSU State Critical/NonRecvoverable"
-  },
-  {
-    "rule": "sensors.sensor_current ~ \"[2|6]\" && sensors.sensor_oid = \".1.3.6.1.4.1.674.10893.1.20.140.1.1.4\"",
-    "name": "Dell Server Virtual Disk Failed/Degraded"
-  },
```
ottorei, line 339: I agree that these should probably be removed or converted to use the more general state sensor rule. The whole Dell-specific code should probably be upgraded.

misc/alert_rules.json
PipoCanaja: Hello
If you remove this, then you have no more option to disable alerting on this specific sensor, and you break the alerting logic. So could you please revert this change ?
Thanx

PipoCanaja (dismissed):

Fehler12:
Hi, from my experience the alert-flag does not work for state sensors. If I add ` && sensors.sensor_alert = 1` to the rule, it won't be triggered, even if the state sensor is in a warning or critical state. Can someone please check this? 
And can you give me an explanation where to troubleshoot this?

laf:
The `sensor_alert` column is actively populated, toggling a health sensor alerts to be off updates that column as well so it's valid in alert rules.

murrant (approved):
Removed unrelated changes

PR #13728, opened 2022-01-28
Oxidized reload Nodes when add device

Jimmyord:
Fix automatic Oxidized reload when adding device. Function got called when deleting device only.

#### Please note

#### Testers

### DIFF
--- a/app/Observers/DeviceObserver.php
+++ b/app/Observers/DeviceObserver.php
@@ -16,6 +16,7 @@ class DeviceObserver
     public function created(Device $device)
     {
         Log::event("Device $device->hostname has been created", $device, 'system', 3);
+        // Reload Oxidized if needed
         oxidized_reload_nodes();
     }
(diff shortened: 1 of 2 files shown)


murrant (changes requested):
There is a much better place for this.

https://github.com/librenms/librenms/blob/master/app/Observers/DeviceObserver.php#L18

Jimmyord:
Should we move the same function call from include/functions.php to DeviceObserver (deleting function)?

murrant:
@Jimmyord Great! Exactly what is needed.  But I realized that not all places that call the code include the functions.inc.php file that has oxidized_reload_nodes().  This means we need to move it from there to a class that can be dynamically included.

That was not a small task and snowballed a bit so I made a new PR here #13730 

Thanks! without your PR that other one would not have been written.  Look forward to your next PR :)

PR #13748, opened 2022-02-03
[widget] Top Devices

Npeca75:
calculate bitrate only from ports which are UP

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Widgets/TopDevicesController.php
+++ b/app/Http/Controllers/Widgets/TopDevicesController.php
@@ -196,6 +196,7 @@ class TopDevicesController extends WidgetController
             }, function ($query) {
                 return $query->has('device');
             })
+            ->where('ifOperStatus','up')
             ->orderByRaw('SUM(ifInOctets_rate + ifOutOctets_rate) ' . $sort)
             ->limit($settings['device_count']);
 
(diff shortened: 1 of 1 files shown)


Npeca75:
hi @PipoCanaja 
i stumbled in this bug when i shut down one of high bitrate port
so, device was still on top of list, because TopDevices widget does not consider which ports are up or down, simply SUM all ports.
Expected behavior is, if device was on top, and you shut down a port, widget need to reflect these changes

Npeca75:
@PipoCanaja 

to be more specific, ports which are DOWN are left "as is" in DB with last data, so they are never zeroed out

PipoCanaja:
> hi @PipoCanaja i stumbled in this bug when i shut down one of high bitrate port so, device was still on top of list, because TopDevices widget does not consider which ports are up or down, simply SUM all ports. Expected behavior is, if device was on top, and you shut down a port, widget need to reflect these changes

Do you use "per port polling" by any chance ?

Npeca75:
> "per port polling"

no, nothing like that :)

ottorei:
@PipoCanaja After looking at the code it seems that the widget just sums ingress and egress bits from all available interfaces. If there was a port with massive amounts of data transferred and that port went down for any reason, looking at the sum itself would skew the result if other ports would not compensate, right? The device would still be listed as the top device even when it would not pass much traffic at all. I think the calculation should only sum the ports that were up during the last poll or were up during the last n-minutes. This is for the real time view.

Npeca75:
hi @ottorei 

we discussed this case with @murrant on discord
there is 2 solution, one to constant fill port counters with zero, for ports which are down (very ugly solution)
or
as @murrant  suggested, to skip counting "down" ports in widget
so i made this modification

ottorei (approved):

PR #13750, opened 2022-02-04
Zynos Discovery and Polling Enhancements

Jimmy-Cl:

Enhancements for Zynos OS -
- Human Readable port numbers
- Added Dbm levels to ports
- Added Stacking state sensor

image

image

image

#### Please note

#### Testers


PipoCanaja:
Hi @Jimmy-Cl 
Could you check the existing test JSON data and re-generate them to take into account your improvements ? This is needed so the regression tests succeed.

`tests/data/zynos.json`:
https://github.com/librenms/librenms/runs/5145911327?check_suite_focus=true

Jimmy-Cl:
@PipoCanaja I have updated the zynos.json and zynos.snmprec files so the tests should pass now.

includes/definitions/discovery/zynos.yaml
```diff
@@ -26,6 +26,61 @@ modules:
                 oid: sysMgmtCPU5SecUsage
                 num_oid: '.1.3.6.1.4.1.890.1.15.3.2.7.{{ $index }}'
     sensors:
+        pre-cache:
+            data:
+                -
+                    oid:
+                        - ifDescr
```
PipoCanaja, line 33: ifDescr is not necessary here, unused.

includes/definitions/discovery/zynos.yaml
```diff
@@ -26,6 +26,61 @@ modules:
...
+                        - zyStackingSlotInfoStackingStatus
```
PipoCanaja, line 34: Almost sure this one is not necessary, as it is part of zyxelStackingSlotInfoTable so it will be already loaded. 
```suggestion
                        - zyStackingSlotInfoStackingStatus
```

tests/snmpsim/zynos.snmprec
```diff
@@ -64,66 +58,30 @@
...
-1.0.8802.1.1.2.1.3.7.1.3.422|4|214
-1.0.8802.1.1.2.1.3.7.1.3.423|4|215
-1.0.8802.1.1.2.1.3.7.1.3.424|4|216
-1.0.8802.1.1.2.1.3.7.1.3.425|4|217
-1.0.8802.1.1.2.1.3.7.1.3.426|4|218
-1.0.8802.1.1.2.1.3.7.1.3.427|4|219
-1.0.8802.1.1.2.1.3.7.1.3.428|4|220
-1.0.8802.1.1.2.1.3.7.1.3.429|4|221
-1.0.8802.1.1.2.1.3.7.1.3.430|4|222
+1.0.8802.1.1.2.1.4.1.1.4.120.128.1408|2|4
```
PipoCanaja: You are not supposed to change an existing snmprec file. Only create new ones. And (re) calculate the JSON files. 
SNMPREC is a capture of the device, and JSON is the calculated as a "reference" on how LibreNMS sees the device.
And regression test will be compared again at every change of the source code  to detect any issue.

PipoCanaja (changes requested):

includes/definitions/discovery/zynos.yaml
Jimmy-Cl: Well spotted, I originally used ifDescr but must have forgotten to remove it from the pre-cache

includes/definitions/discovery/zynos.yaml
Jimmy-Cl: Once again forgot to remove the pre-cache after I stopped using zyStackingSlotInfoStackingStatus with different tables.

tests/snmpsim/zynos.snmprec
Jimmy-Cl: I have reverted my zynos.snmprec changes. Can you just confirm that what I did to the JSON files was correct, as I ran a new test and renamed it to be zynos.json not zynos_xgs4600.json? It would not let me run a test without a variant. If this is wrong and you have another script you want me to run just let me know.

laf:
It's showing as the `zynos.json` being modified. If you take the contents of this file and save it as `zynos.json` that will restore it, then commit the file and push here.

Jimmy-Cl:
I have restored zynos.json as you wanted however the tests will most likley still fail as the changes I have made effect the ifDescr, ifName and ifAlias of zynos switch ports. When your tests run they use the existing zynos.json not the zynos_xgs4600.json file that I generated so they show an error which is, OS zynos: Discovered ports data does not match that found in tests/data/zynos.json. Can we simply ignore this error or is there a way of me regenerating the existing zynos.json file to have my new port fDescr, ifName and ifAlias.

includes/discovery/ports/zynos.inc.php
```diff
@@ -0,0 +1,60 @@
...
+//loop through ports
+foreach ($port_stats as $index => $port) {
+    //Logically convert swp ports to human readable
+    //only convert swp ports
+    if (substr($port['ifName'], 0, 3) === 'swp') {
+        $portNum = preg_replace('/[a-z]+/', '', $port['ifName']);
+        if ($portNum < 50) {
+            $portNum++;
+            //Leading 0 for single digits
+            if ($portNum < 10) {
```
murrant, line 20: instead of doing that, you can just do this:
```php
$ifDescr = '1/' . zeropad($portNum);
```

includes/discovery/ports/zynos.inc.php
```diff
@@ -0,0 +1,60 @@
...
+                $ifDescr = '4/' . $portNum;
+            }
+        }
+    } else {
+        //Set port number to match current
+        $ifDescr = $port['ifName'];
+    }
+    //set the libre ifDescr value to new port name
+    $port_stats[$index]['ifDescr'] = $ifDescr;
+    $port_stats[$index]['ifName'] = $ifDescr;
```
murrant, line 59: It might be worthwhile to only override ifName.  if ifDescr is empty, you could set it to the original ifName value.
ifDescr is the port as described by the os
ifName should be the name of the port as used in cli commands
ifAlias is the human configured description of the port

murrant: There is also an os definition flag to set which field to prefer for displaying the port name.

Jimmy-Cl: We tested the OS definition flag but the reason that we changed both ifDescr and ifName was because some areas of libre would still show as swp numbers, The cli uses the 1/1 format and so does the writing on the switch so thats another reson why we wanted to change ifDescr. I will test it again to see which areas use ifDescr and if it is still necessary.

includes/discovery/ports/zynos.inc.php
Jimmy-Cl: Never used zeropad but it looks interesting, I will test it and get back to you.

Jimmy-Cl: Zeropad works really well, Thanks for pointing this out :)

includes/discovery/ports/zynos.inc.php
Jimmy-Cl: After testing it would be best if we can change both ifDescr and ifName so we can completely remove the swp numbers. If you guys don't like us changing ifDescr let me know and I will remove it. Please can someone rerun the workflows now I have added the zeropad config so it can test the pull request.

murrant:
We can't merge it without succeeding tests.
If you changed the expected result(data) then you will need to update the LibreNMS test data to reflect that new data.
[./scripts/save-test-data.php ](https://docs.librenms.org/Developing/os/Test-Units/#2-save-test-data)

Jimmy-Cl:
@murrant I have updated my pull request so that it doesn't change ifDescr like you requested. I have added ifname: true to the zynos.yaml file so that ifName overrides the ifDescr for the name of the port. I have also included new test data which confirms that ifDescr stays as the swp01 format and confirmed on our dev instance that the ports on the front end change to be the new ifName like we want. Let me know if you spot any issues.

Jimmy-Cl:
I have moved my Re-Generated Zynos.json to this pull request please can someone re-run the tests below.

Jimmy-Cl:
Closing this pull request and changing my script so it can pass the tests

PR #13764, opened 2022-02-10
Implement OAuth and SAML2 support

Jellyfrog:
Thanks @laf for help with the docs

#### Please note

#### Testers


Jellyfrog:
Have to finalize the docs

laf:
I've popped a few more updates in to the docs @Jellyfrog (personally I think we merge as is and adapt the docs based on feedback from users).

It made me realise that we should have a visible thing on the page to show what version of LibreNMS is needed for the feature to work. I've added that in as text for now but a nice button would be better. One for later again. But if this PR doesn't make the 22.2.0 release then we'll just need to update the docs, I actually think we should release this just after 22.2.0 lands so we can get some beta testers using it in anger.

database/migrations/2022_02_03_164059_increase_auth_id_length.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->string('auth_id')->nullable()->change();
```
murrant, line 17: Note to look out for: This could allow different auth_ids because they are no longer being cast from string to int...  but we will only know once this is in the wild.

murrant (approved):

Jellyfrog:
@murrant @laf I have some other changes that allows a user to register their current account with a socialite provider.
However since it will replace the auth_id  and auth_type it breaks the login via mysql since that uses `thisAuth()`

https://github.com/librenms/librenms/blob/d8f96a46b05646442549254b39bcee597191ca55/LibreNMS/Authentication/MysqlAuthorizer.php#L21

https://github.com/librenms/librenms/blob/d8f96a46b05646442549254b39bcee597191ca55/app/Models/User.php#L151-L161

one way is to add new fields for oauth

a second way would be to add a new legacyauthprovider, that just inherits MysqlAuthorizer and overrides the login command to omit the `thisAuth()`

third way, omit the `thisAuth()` from MysqlAuthorizer

laf:
@Jellyfrog I'm of the opinion we drop the `thisAuth` call for MySQL. No other auth modules will create a password for users and even if they do, it means we can always allow a fall back. I'll leave you to decide if you want to do that now (if @murrant  agrees I guess) or merge as is and let's get some feedback.

murrant:
My take:
External auth should not have the password field set.
Password empty = block local login
Password set = allow local login

Most of the LegacyAuth stuff should be refactored out sometime.

murrant (approved):
LGTM

PR #13783, opened 2022-02-14
Hide community and make it XSS safer

PipoCanaja:
Community is now hidden by default, until field receives focus. 
And community is unchanged if value is "********" when saved.

And value is now protected to avoid HTML / JS injections.
Capture d’écran 2022-02-14 à 08 54 55

Capture d’écran 2022-02-14 à 08 55 01

#### Please note

#### Testers

### DIFF
--- a/includes/html/pages/device/edit/snmp.inc.php
+++ b/includes/html/pages/device/edit/snmp.inc.php
@@ -33,7 +33,7 @@ if ($_POST['editing']) {
                 $update['retries'] = ['NULL'];
             }
 
-            if ($snmpver != 'v3') {
+            if ($snmpver != 'v3' && $_POST['community'] != '********') {
                 $community = $_POST['community'];
                 $update['community'] = $community;
             }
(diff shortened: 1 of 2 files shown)


PipoCanaja:
The community beeing hidden is only an "over the shoulder" security (the value is indeed in the HTML source anyway). 
It also helps to protect from injection by using the "onfocus" event to load the data with `decodeURIComponent`.

includes/html/pages/device/edit/snmp.inc.php
```diff
@@ -318,7 +318,7 @@
     <div class='form-group'>
     <label for='community' class='col-sm-2 control-label'>SNMP Community</label>
     <div class='col-sm-4'>
-    <input id='community' class='form-control' name='community' value='" . htmlspecialchars($device['community']) . "'/>
+    <input id='community' class='form-control' name='community' value='********' onfocus='this.value=(this.value==\"********\" ? decodeURIComponent(\"" . rawurlencode($device['community']) . "\") : this.value);'/>
```
Jellyfrog, line 321: Change the type instead, using alpine it would be like
`<input type="password" value="hello" x-data="{ type: 'password' }" :type="type" @focus="type='text'">`

laf: Doesn't seem to work and still causes an XSS:

`<input type='password' value='" . $device['community'] . "' x-data='{ type: \'password\' }' :type='type' @focus='type=\'text\'' >`

Tried adding the onfocus back in but all that does is keep the output masked as `******`

murrant:
Not sure this PR had anything to do with xss...

PipoCanaja:
> Not sure this PR had anything to do with xss...

I reproduced it in the addhost page itself. But an XSS that impacts only the attacker is not an XSS. And because the community "input" field is filled with unescaped data, it is theoretically possible to put something bad here. Now it is safely displayed in the "input" field and also hidden from over-the-shoulder colleagues :D

laf:
I reproduced it in the Edit device page as well which means it would impact
all Admin or device enabled users.

PR #13788, opened 2022-02-15
Add LDAP bind and userlist filter support to ldap-authorization

pfromme25:
The ldap-authorization module lacks a lot of the options included in the normal ldap auth module, therefore this PR adds the userlist filter and bind user functionality to ldap-authorization.

#### Please note

#### Testers

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1301,11 +1301,6 @@ parameters:
 			path: LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
 
 		-
-			message: "#^Method LibreNMS\\\\Authentication\\\\LdapAuthorizationAuthorizer\\:\\:getFullDn\\(\\) has parameter \\$username with no type specified\\.$#"
-			count: 1
-			path: LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
-
-		-
 			message: "#^Parameter \\#1 \\$link_identifier of function ldap_error expects resource, false given\\.$#"
 			count: 1
 			path: LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1301,6 +1301,11 @@ parameters:
 			path: LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
 
 		-
+			message: "#^Method LibreNMS\\\\Authentication\\\\LdapAuthorizationAuthorizer\\:\\:getFullDn\\(\\) has parameter \\$username with no type specified\\.$#"
+			count: 1
+			path: LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
+
+		-
 			message: "#^Parameter \\#1 \\$link_identifier of function ldap_error expects resource, false given\\.$#"
 			count: 1
 			path: LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
--- a/LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
+++ b/LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
@@ -167,7 +167,9 @@ class LdapAuthorizationAuthorizer extends AuthorizerBase
         $userlist = [];
 
         $filter = '(' . Config::get('auth_ldap_prefix') . '*)';
-
+        if (Config::get('auth_ldap_userlist_filter') != null) {
+            $filter = '(' . Config::get('auth_ldap_userlist_filter') . ')';
+        }
         $search = ldap_search($this->ldap_connection, trim(Config::get('auth_ldap_suffix'), ','), $filter);
         $entries = ldap_get_entries($this->ldap_connection, $search);
 
--- a/LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
+++ b/LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
@@ -233,7 +233,7 @@ class LdapAuthorizationAuthorizer extends AuthorizerBase
      *
      * @return string
      */
-    protected function getFullDn($username)
+    protected function getFullDn(string $username)
     {
         return Config::get('auth_ldap_prefix', '') . $username . Config::get('auth_ldap_suffix', '');
     }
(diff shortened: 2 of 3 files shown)


phpstan-baseline.neon
```diff
@@ -1300,6 +1300,11 @@ parameters:
 			count: 1
 			path: LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
 
+		-
```
Jellyfrog, line 1303: Sorry we only support removing from this file, please type hint the parameter

pfromme25: Thanks for the feedback. I've added another commit to incorporate your suggestion.

laf:
@pfromme25 Thank you for contributing 👍 Do the docs https://docs.librenms.org/Extensions/Authentication/#http-authentication-ldap-authorization need updating to reflect these changes as well?

laf:
> If you want, I can make sure which options are missing in this module and update the docs accordingly, as to not lead someone astray when configuring this module.

That would be super amazing if you could?

pfromme25:
Ok, I checked the LdapAuthorizer module and I think I found one option actually missing from the basic ldap module docs as well.
```@ -192,6 +192,7 @@ $config['auth_ldap_prefix'] = 'uid=';               // prepended to usernames
 $config['auth_ldap_group']  = 'cn=groupname,ou=groups,dc=example,dc=com'; // generic group with level 0
 $config['auth_ldap_groupmemberattr'] = 'memberUid'; // attribute to use to see if a user is a member of a group
 $config['auth_ldap_uid_attribute'] = 'uidnumber';   // attribute for unique id
+$config['auth_ldap_timeout'] = 5;                   // time to wait before giving up (or trying the next server)
 $config['auth_ldap_debug'] = false;                 // enable for verbose debug messages
 $config['auth_ldap_userdn'] = true;                 // Uses a users full DN as the value of the member attribute in a group instead of member: username. (it’s member: uid=username,ou=groups,dc=domain,dc=com)
 $config['auth_ldap_userlist_filter'] = 'service=informatique'; // Replace 'service=informatique' by your ldap filter to limit the number of responses if you have an ldap directory with thousand of users
```

In terms of the ldap_authorization module, I have two suggestions, depending on whether the missing or available options should be listed:

Missing options variant:
This one would add the missing options as a list, making the overall documentation smaller, though with the possibility of future additional options in the ldap module not being added to this list by mistake.
````@@ -338,7 +339,7 @@ will use LDAP to determine and assign the userlevel of a user. The
 userlevel will be calculated by using LDAP group membership
 information as the ___ldap___ module does.
 
-The configuration is the same as for the ___ldap___ module with one extra option: auth_ldap_cache_ttl.
+The configuration is similar to the ___ldap___ module with one extra option: auth_ldap_cache_ttl.
 This option allows to control how long user information (user_exists, userid, userlevel) are cached within the PHP Session.
 The default value is 300 seconds.
 To disabled this caching (highly discourage) set this option to 0.
@@ -347,6 +348,15 @@ To disabled this caching (highly discourage) set this option to 0.
 $config['auth_ldap_cache_ttl'] = 300;
 ```
 
+Most of the options from the ___ldap___ module carry over, though the following are absent:
+```
+auth_ldap_wildcard_ou
+auth_ldap_userdn
+auth_ldap_uid_attribute
+auth_ldap_debug
+auth_ldap_timeout
+```
````

Available options variant:
This one would basically copy most of the documentation of the ldap module, resulting in a lot of redundant information, though making it less error prone in the future.

``````
@@ -338,15 +339,48 @@ will use LDAP to determine and assign the userlevel of a user. The
 userlevel will be calculated by using LDAP group membership
 information as the ___ldap___ module does.
 
-The configuration is the same as for the ___ldap___ module with one extra option: auth_ldap_cache_ttl.
+The configuration is similar to the ___ldap___ module with one extra option: auth_ldap_cache_ttl.
 This option allows to control how long user information (user_exists, userid, userlevel) are cached within the PHP Session.
 The default value is 300 seconds.
 To disabled this caching (highly discourage) set this option to 0.
 
+### Standard config
+
 ```php
+$config['auth_mechanism'] = 'ldap';
+$config['auth_ldap_server'] = 'ldap.example.com';               // Set server(s), space separated. Prefix with ldaps:// for ssl
+$config['auth_ldap_suffix'] = ',ou=People,dc=example,dc=com';   // appended to usernames
+$config['auth_ldap_groupbase'] = 'ou=groups,dc=example,dc=com'; // all groups must be inside this
+$config['auth_ldap_groups']['admin']['level'] = 10;             // set admin group to admin level
+$config['auth_ldap_groups']['pfy']['level'] = 5;                // set pfy group to global read only level
+$config['auth_ldap_groups']['support']['level'] = 1;            // set support group as a normal user
+```
+
+### Additional options (usually not needed)
+
+```php
+$config['auth_ldap_version'] = 3; # v2 or v3
+$config['auth_ldap_port'] = 389;                    // 389 or 636 for ssl
+$config['auth_ldap_starttls'] = True;               // Enable TLS on port 389
+$config['auth_ldap_prefix'] = 'uid=';               // prepended to usernames
+$config['auth_ldap_group']  = 'cn=groupname,ou=groups,dc=example,dc=com'; // generic group with level 0
+$config['auth_ldap_groupmemberattr'] = 'memberUid'; // attribute to use to see if a user is a member of a group
+$config['auth_ldap_userlist_filter'] = 'service=informatique'; // Replace 'service=informatique' by your ldap filter to limit the number of responses if you have an ldap directory with thousand of users
 $config['auth_ldap_cache_ttl'] = 300;
 ```
 
+### LDAP bind user (optional)
+
+If your ldap server does not allow anonymous bind, it is highly
+suggested to create a bind user, otherwise "remember me", alerting
+users, and the API will not work.
+
+```php
+$config['auth_ldap_binduser'] = 'ldapbind'; // will use auth_ldap_prefix and auth_ldap_suffix
+#$config['auth_ldap_binddn'] = 'CN=John.Smith,CN=Users,DC=MyDomain,DC=com'; // overrides binduser
+$config['auth_ldap_bindpassword'] = 'password';
+```
+
 ## View/embedded graphs without being logged into LibreNMS
 
 ```php
``````

pfromme25:
> Probably go with option b for the docs.
> 
> As mentioned, we have no way of testing your patch won't break other users.

Thanks for the feedback. I've added two commits to add the documentation part.

Regarding breakage for users. I've tested this PR on our test system for the last few weeks on releases 22.1.0 and 22.2.2 (both with and without bind) without issues. As the code is encased in an if statement, it should only run if the bind part of the options is set, which did not do anything before this PR in the ldap-authorization module.

murrant (approved):

PR #13791, opened 2022-02-15
Updated Nokia MIB files

rsys-dev:
Since we Updated our 7250 to 21.7R1, MPLS could not be fetched anymore. Replacing the MIB files with the new Versions solves this problem.

#### Please note

#### Testers


PipoCanaja:
@rsys-dev 
Could you please add test files for 7750 devices ? I don't see any, and that would also test those new mibs.

Jellyfrog:
I'm pretty sure this will break the current tests, but it won't show here due to our tests filtering them :)

Jellyfrog (changes requested):
Blocked by https://github.com/librenms/librenms/pull/13795

tests/data/timos_hc.json
```diff
@@ -8273,48 +8273,6 @@
     "wireless": {
         "discovery": {
             "wireless_sensors": [
-                {
```
PipoCanaja, line 8276: @rsys-dev It seems that this change of MIB currently breaks some wireless sensors. Could you have a look there ? 
That does not explain why tests and save-test-data don't agree on how to regenerate the tests, but that must be fixed anyway.

PipoCanaja (changes requested):

Jellyfrog:
(Trigger new tests)

tests/data/timos_hc.json
rsys-dev: Will check this

rsys-dev: Sorry for the delay, had not much time the last two weeks, will have a look at it this week

peelman:
These were updated in #14734.

PR #13796, opened 2022-02-16
[comware] dropped dbfetch from discovery/sensors

Npeca75:
dbfetch cleanup from discovery/sensors

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/sensors/temperature/comware.inc.php
+++ b/includes/discovery/sensors/temperature/comware.inc.php
@@ -8,11 +8,18 @@
  * Free Software Foundation, either version 3 of the License, or (at your
  * option) any later version.  Please see LICENSE.txt at the top level of
  * the source code distribution for details.
- */
+ *
+ * @author     Peca Nesovanovic <peca.nesovanovic@sattrakt.com>
+*/
+use App\Models\EntPhysical;
 
 echo 'Comware ';
 
-$entphydata = dbFetchRows("SELECT `entPhysicalIndex`, `entPhysicalClass`, `entPhysicalName` FROM `entPhysical` WHERE `device_id` = ? AND `entPhysicalClass` REGEXP 'module|sensor' ORDER BY `entPhysicalIndex`", [$device['device_id']]);
+$entphydata = EntPhysical::where('device_id', $device['device_id'])
+    ->where('entPhysicalClass', 'regexp', 'module|sensor')
+    ->orderBy('entPhysicalIndex')
+    ->get(['entPhysicalIndex','entPhysicalClass', 'entPhysicalName'])
+    ->toArray();
 
 if (! empty($entphydata)) {
     $tempdata = snmpwalk_cache_multi_oid($device, 'hh3cEntityExtTemperature', [], 'HH3C-ENTITY-EXT-MIB');
@@ -55,7 +62,6 @@ $divisor_alarm = 1000;
 foreach ($pre_cache['comware_oids'] as $index => $entry) {
     if (is_numeric($entry['hh3cTransceiverTemperature']) && $entry['hh3cTransceiverTemperature'] != 2147483647 && isset($entry['hh3cTransceiverDiagnostic'])) {
         $oid = '.1.3.6.1.4.1.25506.2.70.1.1.1.15.' . $index;
-        $dbquery = dbFetchRows("SELECT `ifDescr` FROM `ports` WHERE `ifIndex`= ? AND `device_id` = ? AND `ifAdminStatus` = 'up'", [$index, $device['device_id']]);
         $limit_low = $entry['hh3cTransceiverTempLoAlarm'] / $divisor_alarm;
         $warn_limit_low = $entry['hh3cTransceiverTempLoWarn'] / $divisor_alarm;
         $limit = $entry['hh3cTransceiverTempHiAlarm'] / $divisor_alarm;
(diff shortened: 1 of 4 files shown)


includes/discovery/sensors/temperature/comware.inc.php
```diff
@@ -8,11 +8,18 @@
...
- */
+ *
+ * @author     Peca Nesovanovic <peca.nesovanovic@sattrakt.com>
+*/
+use App\Models\EntPhysical;
 
 echo 'Comware ';
 
-$entphydata = dbFetchRows("SELECT `entPhysicalIndex`, `entPhysicalClass`, `entPhysicalName` FROM `entPhysical` WHERE `device_id` = ? AND `entPhysicalClass` REGEXP 'module|sensor' ORDER BY `entPhysicalIndex`", [$device['device_id']]);
+$entphydata = EntPhysical::where('device_id', $device['device_id'])
```
murrant, line 18: This is only a partial fix because $device needs to go away too.

`DeviceCache::getPrimary()->entityPhysical()` ... (then where entPhysicalClass etc)

murrant: A not for those who don't know ->entityPhysical will return a collection of the relationship items and ->entityPhysical() will return a query for the relationship items.

includes/discovery/sensors/temperature/comware.inc.php
```diff
@@ -8,11 +8,16 @@
...
- */
-
+ *
+ * @author     Peca Nesovanovic <peca.nesovanovic@sattrakt.com>
+*/
 echo 'Comware ';
 
-$entphydata = dbFetchRows("SELECT `entPhysicalIndex`, `entPhysicalClass`, `entPhysicalName` FROM `entPhysical` WHERE `device_id` = ? AND `entPhysicalClass` REGEXP 'module|sensor' ORDER BY `entPhysicalIndex`", [$device['device_id']]);
+$entphydata = DeviceCache::getPrimary()->entityPhysical()
+    ->where('entPhysicalClass', 'regexp', 'module|sensor')
```
Jellyfrog, line 17: This can probably have a tiny performance win by
```suggestion
    ->whereIn('entPhysicalClass', ['module', 'sensor'])
```
but untested.

includes/discovery/sensors/temperature/comware.inc.php
```diff
@@ -55,16 +60,16 @@
...
         $oid = '.1.3.6.1.4.1.25506.2.70.1.1.1.15.' . $index;
-        $dbquery = dbFetchRows("SELECT `ifDescr` FROM `ports` WHERE `ifIndex`= ? AND `device_id` = ? AND `ifAdminStatus` = 'up'", [$index, $device['device_id']]);
         $limit_low = $entry['hh3cTransceiverTempLoAlarm'] / $divisor_alarm;
         $warn_limit_low = $entry['hh3cTransceiverTempLoWarn'] / $divisor_alarm;
         $limit = $entry['hh3cTransceiverTempHiAlarm'] / $divisor_alarm;
         $warn_limit = $entry['hh3cTransceiverTempHiWarn'] / $divisor_alarm;
         $current = $entry['hh3cTransceiverTemperature'];
         $entPhysicalIndex = $index;
         $entPhysicalIndex_measured = 'ports';
-        foreach ($dbquery as $dbindex => $dbresult) {
```
Jellyfrog, line 66: was this foreach useless?

includes/discovery/sensors/voltage/comware.inc.php
```diff
@@ -18,16 +19,16 @@
...
         $oid = '.1.3.6.1.4.1.25506.2.70.1.1.1.16.' . $index;
-        $dbquery = dbFetchRows("SELECT `ifDescr` FROM `ports` WHERE `ifIndex`= ? AND `device_id` = ? AND `ifAdminStatus` = 'up'", [$index, $device['device_id']]);
         $limit_low = $entry['hh3cTransceiverVccLoAlarm'] / $divisor_alarm;
         $warn_limit_low = $entry['hh3cTransceiverVccLoWarn'] / $divisor_alarm;
         $limit = $entry['hh3cTransceiverVccHiAlarm'] / $divisor_alarm;
         $warn_limit = $entry['hh3cTransceiverVccHiWarn'] / $divisor_alarm;
         $current = $entry['hh3cTransceiverVoltage'] / $divisor;
         $entPhysicalIndex = $index;
         $entPhysicalIndex_measured = 'ports';
-        foreach ($dbquery as $dbindex => $dbresult) {
```
Jellyfrog, line 29: same question as above

includes/discovery/sensors/temperature/comware.inc.php
Npeca75: hi @Jellyfrog 
that was one big question, but test data is passing
from my point of view, ifIndex is unique for each port and could not be two port indexed with same value
so this query (original)
$dbquery = dbFetchRows("SELECT `ifDescr` FROM `ports` WHERE `ifIndex`= ? AND `device_id` = ? AND `ifAdminStatus` = 'up'", [$index, $device['device_id']
simply return array in array, that was the reason for "foreach" to extract array

includes/discovery/sensors/temperature/comware.inc.php
Npeca75: tried that at first run, before commit, but then changed to regexp, to be same as original :)
yes, it is working boh way

includes/discovery/sensors/voltage/comware.inc.php
Npeca75: same answer as above :)

Jellyfrog (approved):

PR #13804, opened 2022-02-18
 Added peer context/vrf to the devices bgp peers page, added getting of vrf from vrf table instead of only vrflites

Senetus:

Added peer context/vrf to the devices bgp peers page

image

#### Please note

#### Testers

### DIFF
--- a/doc/Support/Configuration.md
+++ b/doc/Support/Configuration.md
@@ -421,21 +421,22 @@ lnms config:set overview_show_sysDescr true
 Enable or disable the sysDescr output for a device.
 
 ```bash
-lnms config:set force_ip_to_sysname false
+lnms config:set device_default_display '{{ $hostname }}'
 ```
 
-When using IP addresses as a hostname you can instead represent the
-devices on the WebUI by its SNMP sysName resulting in an easier to
-read overview of your network. This would apply on networks where you
-don't have DNS records for most of your devices.
+This is a simple template to control the display of device names by default.
+You can override this setting per-device.
 
-```bash
-lnms config:set force_hostname_to_sysname false
-```
+You may enter any free-form text including one or more of the following template replacements:
 
-When using a dynamic DNS hostname or one that does not resolve, this
-option would allow you to make use of the SNMP sysName instead as the
-preferred reference to the device.
+| Template                    | Replacement                                                          |
+|-----------------------------|----------------------------------------------------------------------|
+| `{{ $hostname }}`           | The hostname or IP of the device that was set when added  *default   |
+| `{{ $sysName_fallback }}`   | The hostname or sysName if hostname is an IP                         |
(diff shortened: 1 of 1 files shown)


Senetus:
Thanks @murrant for the vrf code!

murrant:
The one question I have is should this always include the default context?

Senetus:
> The one question I have is should this always include the default context?

At least when concerning aristas, yes I would say.

Since there the deafult is the default vrf too and that can't be changed. So all the context polling should be in addition of the default.

Senetus:
Well, bgp peers of course should only be polled with their own context, but discovery should loop though all. Eg. the routes module should first run on the default, and then on any additional context.

PipoCanaja:
Hum, matching between peers and context does not return correct results on other devices (Huawei VRP currently in my tests). Will also test Cisco and see how it goes

PipoCanaja:
@Senetus Contexts, VRF, VRF Lite, all these elements are mixed in LibreNMS but they are not used the same way by all vendors. 
Setting a context for a Device may change the way bgp-discovery and OSPF discovery is done. Which is not what you expect. In fact the context is like a kind of "view" for some mibs. 

What would probably make sense is fetch in which VRF the peer is connected, and store it during BGP discovery in the DB to display it later on.

Senetus:
That's corrext. In arista world a vrf is a namespace and thus also a context. Routes and bgp peers need to be polled/discovered within this context.

PipoCanaja:
> That's correct. In arista world a vrf is a namespace and thus also a context. Routes and bgp peers need to be polled/discovered within this context.

Then you have to enable this context setting only for an Arista device, and keep the new column hidden by default. (which means rewrite this table so that columns can be displayed/hidden).

PipoCanaja:
In fact, context is used for context polling, and context polling may be used for VRFs or other polled data, but this is completely vendor-specific. For most of the devices, this column is not used at all. 
Displaying it unconditionnally is probably not a good idea. And changing [app/Models/Device.php](https://github.com/librenms/librenms/pull/13804/files?show-viewed-files=true&file-filters%5B%5D=#diff-56d0c99da95b2d1d1621c1b694577893589ba6560b344eab93142ccad8c49ff6) is probably not a good idea as it breaks a lot of regression tests. 

So if you can make a proper use of this column for Arista devices, you are very welcome to do so, and adapt the test data as well.
But you'll have to avoid any side effect on other devices.

Senetus:
Yeah, would make sense to display it only if there is something to display. Regarding the broken tests, I adjusted with help of murrant the getvrf's function, since it was only getting vrf lites instead of also data from the vrfs table, like the name would suggest. He did mention probably merging those tables together, so that wouldn't fix the tests most likely.  I don't really know what to do about that.

PR #13810, opened 2022-02-24
Added --ping-only to snmp-scan.py

IVI053:
Added possibility to add found devices as ping-only devices with `snmp-scan.py`. Up to now it was not possible if the device reponds to SNMP. Also changed the arguments for `snmp-scan.py` to be consistent. It now has `--ping-only` and `--ping-fallback` the later does the same as `-P` and `--ping` before. Marked these two as "deprecated" as it they are not explicit anymore but kept them working for compatibility reasons.

I did not update the test data as it was not fully clear if it is needed and how to do it. If so, please leave a comment what needs to be done - I'll update this PR accordingly.

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Auto-Discovery.md
+++ b/doc/Extensions/Auto-Discovery.md
@@ -161,24 +161,26 @@ within your LibreNMS installation directory.
 Here the script's help-page for reference:
 
 ```text
-usage: snmp-scan.py [-h] [-r NETWORK] [-t THREADS] [-l] [-v]
+usage: snmp-scan.py [-h] [-t THREADS] [-g GROUP] [-l] [-v] [--ping-fallback] [--ping-only] [-P] [network ...]
 
 Scan network for snmp hosts and add them to LibreNMS.
 
-optional arguments:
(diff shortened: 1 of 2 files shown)


murrant:
Seems a little odd to use "snmp" scan as a icmp scan :)

murrant (approved):

IVI053:
A little, I thought so as well. But I think it's better to use only one script as it is such a little change.

I did not change anything on the code part lint complains about. If I understand the output right it complains about a string being explicitly formated as unicode. I see no real problem in this and don't know why it was put there. Changing it could break a case we don't know. I would asume it was put there for a reason and it's there for a long time already.
What do you think?

murrant:
According to git history it was a change to appease python black linting.  So I guess it changed it's mind :)

Go ahead and remove the u.  Tests need to pass before merging one way or another.

https://github.com/librenms/librenms/blame/e665642a71a0219139d33a9f33169ab4e6562b73/snmp-scan.py#L281

murrant (approved):

PR #13813, opened 2022-02-24
Fiberstore Network Management Unit

jonathanjdavis:

#### Please note

#### Testers


includes/definitions/fs-nmu.yaml
```diff
@@ -1,10 +1,110 @@
...
+    applications: false #default false
+    services: false #default true
+    slas: false #default false
+    stp: false #default true
+    ntp: false #default true
+    loadbalancers: false #default false
+    mef: false #default false
+    wireless: false #default true
+    fdb-table: false #default true
+    mpls: false
```
Jellyfrog, line 110: Please remove these changes, if one module is causing troubles, you can disable it tho.

includes/definitions/discovery/fs-nmu.yaml
```diff
@@ -1,5 +1,42 @@
+mib: OAP-NMU
 modules:
     os:
-        hardware: OAP-C1-OEO::vDeviceType.0
-        serial: OAP-C1-OEO::vSerialNumber.0
```
Jellyfrog, line 4: If you look at the failing tests you see that this causes the information to be missing, you can add multiple oids instead. Any reason you removed it?

jonathanjdavis: The OEO is a module in an FS NMU device. The vDeviceType and vSerialNumber will  based on which OEO is installed or replaced in position C1, if an OEO Module is even installed at position C1. In place of the OEO Module, there might be an EDFA Module, or a MUX | DEMUX Module (which has nothing to SNMP), or another FS Module installed at position C1.

For the FS-NMU devices hardware, serial, and version, the above was removed and replaced with the OAP-NMU::serialNumber.0, OAP-NMU::softwareVersion.0 and OAP-NMU::hardwareVersion.0 to report the details of the FS NMU device rather than a module installed within it.

includes/definitions/fs-nmu.yaml
jonathanjdavis: done

includes/definitions/discovery/fs-nmu.yaml
jonathanjdavis: modified fs-nmu test data

tests/snmpsim/fs-nmu.snmprec
```diff
@@ -4,13 +4,9 @@
 1.3.6.1.2.1.1.4.0|4|<private>
 1.3.6.1.2.1.1.5.0|4|<private>
 1.3.6.1.2.1.1.6.0|4|<private>
-1.3.6.1.4.1.40989.10.16.1.2.2.0|4|OEO
```
PipoCanaja, line 7: Hi @jonathanjdavis 
Please do not change any existing snmprec file. Add your new snmprec file for sure, and compute existing and new JSON files as well.
Thanx

PipoCanaja (changes requested):

nistorj:
I'd love to see this integrated soon, we have the exact same issue on our end.

includes/definitions/fs-nmu.yaml
```diff
@@ -1,10 +1,19 @@
...
+    - { graph: device_dbm, text: 'dbm' }
 discovery:
     -
         sysObjectID:
             - .1.3.6.1.2.1.1.2
+            - .1.3.6.1.4.1.40989
+poller_modules:
+    sensors: true
+discovery_modules:
+    sensors: true
```
Jellyfrog, line 19: isnt this default on? please remove in that case

Jellyfrog:
Please only add MIBs that you actually use/need

Jellyfrog:
Please see failing tests below

includes/definitions/fs-nmu.yaml
jonathanjdavis: removed.

jonathanjdavis: Unfortunately due to the way FS has written their mibs, OAP-C1-OEO through OAP-C16-OEO are mibs for each of the 16 slots in the device chassis. 1.3.6.1.4.1.40989.10.16.1.2.1, 1.3.6.1.4.1.40989.10.16.2.2.1, etc. If we were to add another OEO module, or move one, to slot 9, it would need the OAP-C9-OEO mib.

Same with the OAP-C1-EDFA through OAP-C16-EDFA mibs.

Other FS customers will also have their modules populated in different slots to ours, and we believe all 16 mibs for both OEOs and EDFAs are required to cover all combinations.

tests/snmpsim/fs-nmu.snmprec
jonathanjdavis: Hi @PipoCanaja , the existing fs-nmu os was updated because in the hardware, serial, and version the previous contributor had these details pulled from a module that might not be installed in that location on other extract same fs chassis. OEOs are modules that can be installed in 1 of 16 slots, with different mibs (1-16) depending on the slot the OEO(EDFA, or other modules) are installed in. The modules hardware, serial, and version were replaced with the chassis hardware serial and version, and I had to update the test data accordingly.

I'll revert and add our testing data without deleting existing.

jonathanjdavis:
updated style fixes

laf:
@jonathanjdavis Hopefully we've not just trod on one anothers toes :) I've helped fix the styling issues as well so let's see how the tests run now.

murrant (approved):

murrant:
@jonathanjdavis test data did not match... not sure why.  Please resubmit.

murrant:
Additionally, that was a LOT of individual snmpget request, you might consider changing it to send more than one oid per request.

PR #13830, opened 2022-03-06
Fixed displaying hostname in create bill when port is passed

laf:
Fixes: https://community.librenms.org/t/cant-create-bill-from-newly-added-device-ports/18286

#### Please note

#### Testers

### DIFF
--- a/includes/html/modal/new_bill.inc.php
+++ b/includes/html/modal/new_bill.inc.php
@@ -23,6 +23,7 @@ if (Auth::user()->hasGlobalAdmin()) {
         $bill_data['bill_ref'] = $port['port_descr_circuit'];
         $bill_data['bill_notes'] = $port['port_descr_speed'];
         $port_device_id = $port['device_id'];
+        $device['hostname'] = $port['hostname'];
     } ?>
 
  <div class="modal fade bs-example-modal-sm" id="create-bill" tabindex="-1" role="dialog" aria-labelledby="Create" aria-hidden="true">


jozefrebjak:
@laf this patch fixed creating new Bill from Port view. It will find interfaces for specified device, but it don't work for creating bill in `Ports->Bills->Crate Bill`. There is no interfaces for the same device as we tried to add via Port view. 

There is also different problem.

This is from Port view creating bill. After patch it works as expected.

Snímka obrazovky 2022-03-07 o 9 42 50

If we just try to specify in dropdown menu different device and then choose previous device interfaces are gone and we can't choose.

Snímka obrazovky 2022-03-07 o 9 45 09

If we try to create bill from `Ports->Bills->Crate Bill` it the same as on the second image.

laf:
I'm sorry but I don't understand what your issue is. This patch fixes only the issue around creating a bill from a specific port. Unless it's introducing a bug then this fix is doing what it's intended.

jozefrebjak:
Ok, so it's better to fix only partial of a problem ? Why to not fix problem with choosing port in bill in one PR ? Is better to make  multiple PR's to fix one problem ?   

If yes, then ok. I'm just tried to describe a whole problem. Maybe on wrong place, sorry.

laf:
I don't understand what your other issue is.

However I only have limited time to work on code for LibreNMS and in this instance, that's all I could do.

PR #13837, opened 2022-03-10
Fix an issue with the APC current discovery

eskyuu:
I discovered that the APC "current" sensor discovery is broken for our PDUs.  This branch provides a fix which I expect will work on all devices, but I also realise there may be other versions out there that could be affected by this change.

The underlying issue for me was that the original code used indexes from one SNMP table on another, and the indexes don't match on my devices (one table has 3 entries, one has 2 and the last has 1).  The solution is to use data from the table with all 3 entries, and then work back to the records with 2 or 1 as appropriate.

#### Please note

#### Testers


PipoCanaja:
Hi @eskyuu 

Those PDU/UPS/etc from APC are a nightmare to support as they all do "almost" the same things, but not exactly ... 

First thing you should do : add test-data to this PR, so that when next PR will come and try to fix something else, it won't break your device :) 

Could you fix the Style test ?
https://github.styleci.io/analyses/2QwvJB?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Thanx

eskyuu:
Test data has been added.  Please let me know if I didn't get it right.

It didn't work at first because the SNMP community that was in use has an @ character in it, and the collect-snmp-data.php script was duplicating everything including and after the @ in the community string. (e.g. if our community was 1234@5678, the script was using 1234@5678@5678 as the community).

Jellyfrog (approved):

PR #13854, opened 2022-03-15
WIP - Sensors - Adding count_rate sensortype (with RRD COUNTER instead of GAUGE)

PipoCanaja:
The goal of this PR is to add a new sensor_class, currently named "Count Rate". Current "Count" type, as all other sensors, is using "GAUGE" RRD type. So a counter that increase is shown exactly the same, increasing. In case you want to see the variation, you are currently left with no option in LibreNMS, except creating a completely new piece of code to do this polling. 

Here, the new sensor_class creates a RRD type COUNTER instead of GAUGE, so the graph variation is computed directly by RRD. 
And a very limited amount of changes is done to display the overview value as (current - prev) / (interval). 

Before: Only an absolute value, no idea of the rate.
image

After: You can say the proxy is not very loaded currently. 
image

#### Please note

#### Testers

### DIFF
--- a/app/Models/Sensor.php
+++ b/app/Models/Sensor.php
@@ -19,6 +19,7 @@ class Sensor extends DeviceRelatedModel
         'chromatic_dispersion' => 'indent',
         'cooling'              => 'thermometer-full',
         'count'                => 'hashtag',
+        'count_rate'           => 'hashtag',
         'current'              => 'bolt fa-flip-horizontal',
         'dbm'                  => 'sun-o',
         'delay'                => 'clock-o',
--- a/doc/Developing/os/Health-Information.md
+++ b/doc/Developing/os/Health-Information.md
@@ -6,38 +6,39 @@ information for your new device.
 Currently we have support for the following health metrics along with
 the values we expect to see the data in:
 
-| Class                           | Measurement                 |
-| ------------------------------- | --------------------------- |
-| airflow                         | cfm                         |
-| ber                             | ratio                       |
-| charge                          | %                           |
-| chromatic_dispersion            | ps/nm                       |
-| cooling                         | W                           |
-| count                           | #                           |
-| current                         | A                           |
-| dbm                             | dBm                         |
-| delay                           | s                           |
-| eer                             | eer                         |
-| fanspeed                        | rpm                         |
-| frequency                       | Hz                          |
-| humidity                        | %                           |
-| load                            | %                           |
-| loss                            | %                           |
-| power                           | W                           |
-| power_consumed                  | kWh                         |
-| power_factor                    | ratio                       |
-| pressure                        | kPa                         |
-| quality_factor                  | dB                          |
-| runtime                         | Min                         |
-| signal                          | dBm                         |
-| snr                             | SNR                         |
-| state                           | #                           |
-| temperature                     | C                           |
-| tv_signal                       | dBmV                        |
-| bitrate                         | bps                         |
-| voltage                         | V                           |
-| waterflow                       | l/m                         |
-| percent                         | %                           |
+| Class                       | Measurement     | Comment |
+| --------------------------- | --------------- | ------- |
+| airflow                     | cfm             | |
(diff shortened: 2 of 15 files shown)


laf:
What about being able to specify an RRD type in the sensors discovery that is then recorded in the DB to avoid the custom code?

PipoCanaja:
> What about being able to specify an RRD type in the sensors discovery that is then recorded in the DB to avoid the custom code?

Good point. I thought about it, but decided to go the simpler variant:
- more data in mysql for every sensor -> space wasted (ok, not a lot if we use ENUM to store it)
- only "count" sensor seems to be a candidate for COUNTER/DERIVE type, for the moment. 
- code would end up being more complex. 
- you could end up with very unexpected sensor units if one uses COUNTER/DERIVE on something that is not really compatible. 
- if we touch the DB, we probably should as well add a "sensor_rate" column, which would allow a more precise rate calculation (at the time you can calculate sensor rate, you have current timestamp, previous timestamp, as well as the difference between the 2 sensor values.)

Let see what other say :)

murrant:
Relational database is terrible for this kind of data.  See IF-MIB stuff for details.
I'm always hesitant for these types of sensors, because they are a little too generic and over time become catch-alls.

PipoCanaja:
> Relational database is terrible for this kind of data. See IF-MIB stuff for details. I'm always hesitant for these types of sensors, because they are a little too generic and over time become catch-alls.

Agreed, @murrant . But on the other end, I have a few devices that provide only a counter for "things" that are not packets or bits in an interface. Hits in a firewall, a proxy, and displaying the DERIVE value makes a lot more sense than absolute. What would be the choices here ? Except creating a dedicated page with dedicated graphs ? It is possible but adds more maintenance work in LNMS than creating a sensor that would be YAML compatible, I think.

jerji:
I am trying to implement Juniper Firewall counters in https://github.com/librenms/librenms/pull/13866 and this is probably the only good way to do this. I need a packet and byte count per firewall rule and they need to be dynamically generated.

PipoCanaja:
> Relational database is terrible for this kind of data. See IF-MIB stuff for details. I'm always hesitant for these types of sensors, because they are a little too generic and over time become catch-alls.

I can rename this "count_rate" with "hits_rate" so it is less generic. We could also define that any sensor name with "_rate" suffix is handled as "DERIVE" instead of "GAUGE". 

What would you think ?

murrant:
> I can rename this "count_rate" with "hits_rate" so it is less generic. We could also define that any sensor name with "_rate" suffix is handled as "DERIVE" instead of "GAUGE".
> 
> What would you think ?

Sounds reasonable, care to give it a try?

murrant:
It might be better to add field to the sensor definition, instead of having an alternate name.

jerji:
It seems that the health page for "Count Rate" is broken
The overview page just fails to load the graph ( it gives me the missing image glyph) and the Count Rate page just shows me the Overview page.

jerji:
@PipoCanaja Do you plan on working on this again?

PipoCanaja:
> @PipoCanaja Do you plan on working on this again?
Hi @jerji 
Yes, I'll fix this PR. Moving to a new place limited drastically my spare time so I cannot give any time window for this work, but I definitely need it done as well :) 
Bye

murrant:
Just pulled this up and had a thought:  Why is called count_rate and not just rate?

PipoCanaja:
> Just pulled this up and had a thought: Why is called count_rate and not just rate?

This is a good point. No name is really covering 100% of the cases here. My 1st idea was to add "_rate" sensor to any existing sensor type, and that would convert it to use DERIVE RRD type. 

But I'll go with your 1st suggestion : add the type of RRD as a discover_sensor / YAML option, which means any sensor will have the capability to use any RRD type if necessary, and without any change in the sensortype. Default value will remain the same as currently, so no change for any existing RRD.

PipoCanaja:
Replaced by #14208

PR #13870, opened 2022-03-23
Migrate addhost.php to lnms device:add

murrant:
Have snmp-scan.py call lnms device:add (make exit codes line up so this works)
Fix issue with ping only devices trying to detect os via snmp
Reorder options in device:add help and improve formatting
Add short options 1,2,3 for snmp version
Update docs to remove references to addhost.php
Fix a bit of code that was in functional code

#### Please note

#### Testers

### DIFF
--- a/addhost.php
+++ b/addhost.php
@@ -95,9 +95,9 @@ if (! empty($argv[1])) {
                 }
             }
         } elseif ($seclevel === 'anp' or $seclevel === 'authNoPriv') {
-            $device->authlevel  = 'authNoPriv';
+            $device->authlevel = 'authNoPriv';
             $v3args = array_slice($argv, 4);
-            $device->authname= array_shift($v3args);
+            $device->authname = array_shift($v3args);
             $device->authpass = array_shift($v3args);
 
             while ($arg = array_shift($v3args)) {
--- a/addhost.php
+++ b/addhost.php
@@ -9,6 +9,8 @@
  * @copyright  (C) 2006 - 2012 Adam Armstrong
  */
 
+use App\Actions\Device\ValidateDeviceAndCreate;
+use App\Models\Device;
 use LibreNMS\Config;
 use LibreNMS\Enum\PortAssociationMode;
 use LibreNMS\Exceptions\HostUnreachableException;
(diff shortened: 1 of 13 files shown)


addhost.php
```diff
@@ -9,6 +9,8 @@
  * @copyright  (C) 2006 - 2012 Adam Armstrong
```
Jellyfrog, line 9: Imo we should just delete this file if it's features are ported to device:add now

murrant: I don't think we can... it has been around too long.

at best we can make it s stub that only prints, please use `lnms device:add`

Jellyfrog: We could symlink to artisan, and then in artisan check what script called it, and route to device:add command.

murrant: The input for addhost.php is garbage... we can't really handle it nicely no matter what we do.  I prefer this change for now, which seeks to have as few side effects as possible, but still enable legacy code removal.  (after it is removed in other places)

We can definitely do more changes later.

Jellyfrog: I meant so it just prints the help of device:add, but nvm its a bit excessive.
I'm all for replacing it with `echo "Please use lnms device:add instead";`, the project doesn't have the manpower to maintain old backward compatibilities

laf:
Is the output supposed to replace things like `:version:` with the version used?

`  SNMP :version: No reply with community arista_eos`

addhost.php
```diff
@@ -18,16 +20,14 @@
...
+$device = new Device;
+
 if (isset($options['g']) && $options['g'] >= 0) {
     $cmd = array_shift($argv);
     array_shift($argv);
     array_shift($argv);
     array_unshift($argv, $cmd);
-    $poller_group = $options['g'];
-} elseif (Config::get('distributed_poller') === true) {
-    $poller_group = Config::get('default_poller_group');
```
laf, line 28: Is this not needed before because it's set in `app/Console/Commands/DeviceAdd.php`?

laf (commented):
Could we change:

```
Arguments:
  device spec                                        Hostname or IP to add
```

To
```
Arguments:
  hostname/ip                                        Hostname or IP to add
```

As `device spec` just seems unintuitive.

Other than that, I've tested many different ways of using `./addhost.php` and nothing seems broken.

Great work as always @murrant 👍

addhost.php
murrant: Correct, but in ValidateDeviceAndCreate it sets up default values.  So there is no need to set them before hand.

murrant:
> Is the output supposed to replace things like `:version:` with the version used?
> 
> ` SNMP :version: No reply with community arista_eos`

Fixed

murrant:
Merging since this contains a bugfix

PR #13871, opened 2022-03-23
Change PlaySMS to send alert message instead of just title

optimdevel:
fixed playsms php transport, that it actually works

fixed message body, so that it is appropriate for SMSes.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Playsms.php
+++ b/LibreNMS/Alert/Transport/Playsms.php
@@ -43,7 +43,10 @@ class Playsms extends Transport
 
     public static function contactPlaysms($obj, $opts)
     {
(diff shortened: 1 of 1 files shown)


murrant (changes requested):
As I understand, only sending the title was due to length limitations.  Are you sure this is desired?

Additionally, you no longer send `$obj['title']` effectively breaking alerts for everyone that currently is using PlaySMS.

optimdevel:
Only title means that, title of the rule is sent which tells me nothing and is always the same.
It works for services, should work also fine for others.
Lenght of SMSes does not seem to be a problem for gammu or playsms.

murrant:
I am pretty sure the title has access to all the variables the message body does so you can put whatever you want in there.

murrant:
No response, closing for now.  Comment to reopen or resubmit.

PR #13877, opened 2022-03-27
Fix usw flex switch hardware detection

Fehler12:
I've updated the regex for unifi switches because it seems that there are at least three ways of syntax:

USW GigaBit Ethernet Switch, firmware 5.76.7.13442
US-8, 5.76.7.13442, Linux 3.6.5
USW-Flex 5.76.7.13442

the first two lines are from the same switch model (US-8) and seem only to differ from hardware revision. I hope that this software version regex is universal now.

Added test data for all three cases and removed false hardware snmprec (it is not an usw8, but an us8)

#### Please note

#### Testers


Jellyfrog (approved):

Jellyfrog:
Thanks!

PR #13894, opened 2022-04-04
implement support for usernames comming from reverse proxies

HolgerHees:
I use the http-auth authentication method, because my authentication is done front apache using openidc. For that reason I want to provide the username of my already authenticated user. For that reason I added support for a specific custom header which is used to extract this username from there.

The custom header "HTTP_X_REMOTE_USER" is only valid for that authentication method. If we want an additional layer of security, we could also add a config switch to control this behavior.

My setup looks now like

$config['auth_mechanism'] = "http-auth";

and the username is used from this customheader. The header itself is set by ma front apache which handles this authentication.

### DIFF
--- a/LibreNMS/Authentication/HttpAuthAuthorizer.php
+++ b/LibreNMS/Authentication/HttpAuthAuthorizer.php
@@ -63,4 +63,13 @@ class HttpAuthAuthorizer extends MysqlAuthorizer
 
         return -1;
     }
+
+    public function getExternalUsername()
+    {
+        if (isset($_SERVER['HTTP_X_REMOTE_USER'])) {
+            return $_SERVER['HTTP_X_REMOTE_USER'];
+        }
+
+        return parent::getExternalUsername();
+    }
 }


Jellyfrog:
What is wrong with REMOTE_USER/PHP_AUTH_USER that AuthorizerBase already implements?

HolgerHees:
My user is stored in a different variable.

Main Problem is that openidc is storing the email in REMOTE_USER. After that I use a rewrite map to convert this email in a username. This username is store in a different ENV var. So in my case REMOTE_USER does not contain the correct user and there is no way to modify this.

murrant:
@HolgerHees neat project.

I'm dubious of accepting a random header for auth things because I'm not sure of the implications.
It might be reasonable to add a setting for http-auth to set the variables to fetch username.

HolgerHees:
I changed my pull request, so the header name is configurable now. I think it is also secure, because it is disable by default. And if you enable it, you have to take care that this field is always set by your front apache server. If you always set this header field, there is no way to manipulate, because it is overwritten by your webservers configuration all the time.

Jellyfrog:
shouldnt the fix be in the base controller instead? that it inherits from anyway

murrant:
Looks good, just need to add it to config_defintions.json

HolgerHees:
Implementation is move to AuthorisationBase class and definition is added to config_defintions.json

misc/config_definitions.json
```diff
@@ -3502,6 +3502,12 @@
                 "custom8": "Custom 8"
             }
         },
+        "http_auth_header": {
+            "group": "auth",
+            "section": "general",
+            "order": 4,
+            "type": "text"
```
murrant, line 3509: Can you set the default to REMOTE_USER and streamline getExternalUsername?
We also need to add the text for this option in `resources/lang/en/settings.php`

LibreNMS/Authentication/AuthorizerBase.php
```diff
@@ -78,6 +78,11 @@ public function getExternalUsername()
             return $_SERVER['REMOTE_USER'];
```
murrant, line 78:
```php
return $_SERVER[Config::has('http_auth_header', 'REMOTE_USER')] ?? $_SERVER['PHP_AUTH_USER'];
```

HolgerHees: Default value handling and i18n string is added

LibreNMS/Authentication/AuthorizerBase.php
```diff
@@ -74,10 +74,13 @@ public function authIsExternal()
 
     public function getExternalUsername()
     {
-        if (isset($_SERVER['REMOTE_USER'])) {
-            return $_SERVER['REMOTE_USER'];
-        } elseif (isset($_SERVER['PHP_AUTH_USER'])) {
+        if (isset($_SERVER['PHP_AUTH_USER'])) {
```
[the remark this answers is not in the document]
HolgerHees: is fixed too... Was not aware of this double question mark behavior, because of my outdated php skills ;-)

added also a missing import/use

murrant (approved):

PR #13902, opened 2022-04-09
[mikrotik] fixed RouterOS ipv4/ipv6 routes

Npeca75:
ipv4 routes was duplicated on all ROS versions, 
ipv6 routes was broken since 7.1.5

ipv4 tested on 6.47.10, 6.49.5, 7.1.5, 7.2
ipv6 tested on 7.1.5, 7.2 , CHR, ARM, MIPSBE

#### Please note

#### Testers


Npeca75:
lnms1

duplicated ipv4 routes (ccitt.0 / zeroDotZero)

Npeca75:
lnms2
working ipv6 routes

includes/discovery/route.inc.php
```diff
@@ -64,7 +65,7 @@
 //if the device does not support IP-FORWARD-MIB, we can still discover the ipv4 (only)
 //routes using RFC1213 but no way to limit the amount of routes here !!
 
-if (! isset($ipForwardNb['0']['inetCidrRouteNumber'])) {
```
Jellyfrog, line 67: Not a fan of this change, if it supports RFC1213, can't we use this and remove the custom discovery instead?

Npeca75: hi @Jellyfrog 
custom discovery is only for IPv6 !!!  not IPv4
and this line
if (! isset($ipForwardNb['0']['inetCidrRouteNumber']) && $device['os'] != 'routeros') 
avoid duplicate IPv4 discovery

Npeca75: > Not a fan of this change, if it supports RFC1213, can't we use this and remove the custom discovery instead?

if you be kind to look at picture attached above, snapshot from database, you will see that there is DUPLICATE ipv4 route discovery, with ccitt.0 and dotZeroDot atributes
so, ROS is discovered TWICE in old route.inc.php code
that is what i try to avoid with != routeros to skip duplicates

Jellyfrog: Yes I understand, but where does the duplicates come from then? feels like a more generic fix would be better

Npeca75: > Yes I understand, but where does the duplicates come from then? feels like a more generic fix would be better

Hi @Jellyfrog 
it is discovered twice because MT dispose both
RFC1213 && IP-FORWARD-MIB

on first, it is discovered as ccitt.0, on second, it is discovered as zeroDotZero

here is the output with DISABLED RFC1213 as in PR

SNMP['/usr/bin/snmpbulkwalk' '-v2c' '-c' 'COMMUNITY' '-OQUsetX' '-m' 'IP-FORWARD-MIB' '-M' '/opt/librenms/mibs:/opt/librenms/mibs/mikrotik' 'udp6:HOSTNAME:255::1]:161' '.1.3.6.1.2.1.4.24.4.1']

Exitcode: 1  

Error: OID not increasing: ipCidrRouteDest[169.254.106.0][255.255.255.0][0][169.254.255.206]
 >= ipCidrRouteDest[169.254.106.0][255.255.255.0][0][169.254.255.206]  

ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *
ipCidrRouteDest[*][*][0][*] = *  

Error: OID not increasing: ipCidrRouteDest[169.254.106.0][255.255.255.0][0][169.254.255.206]
 >= ipCidrRouteDest[169.254.106.0][255.255.255.0][0][169.254.255.206]

murrant:
@Npeca75 I merged the change for setting unordered OIDs on the OS #13923 Can you update this PR to use that?

Npeca75:
> @Npeca75 I merged the change for setting unordered OIDs on the OS #13923 Can you update this PR to use that?

hi @murrant 

working as expected

SNMP['/usr/bin/snmpbulkwalk' '-M' '/opt/librenms/mibs:/opt/librenms/mibs/mikrotik' '-v2c' '-c' 'COMMUNITY' '-OQXUte' '-Cc' 'udp:HOSTNAME:161' 'IP-FORWARD-MIB::ipCidrRouteEntry'] 

SNMP['/usr/bin/snmpbulkwalk' '-M' '/opt/librenms/mibs:/opt/librenms/mibs/mikrotik' '-v2c' '-c' 'COMMUNITY' '-OQXUte' '-Cc' 'udp:HOSTNAME:161' 'IPV6-MIB::ipv6RouteTable']

includes/discovery/route.inc.php
murrant: zeroDotZero and ccitt.0 are both 0.0, just with different mibs.

perhaps we can tell it to return numeric data instead to prevent duplicates.

includes/definitions/routeros.yaml
```diff
@@ -13,3 +13,8 @@ discovery:
     -
         sysObjectID:
             - .1.3.6.1.4.1.14988.1
+oids:
+    unordered:
+        - IP-FORWARD-MIB::ipCidrRouteEntry
```
murrant, line 18: It is better to use OIDs called Table instead of Entry.  Can you please change this and the related walk to `IP-FORWARD-MIB::ipCidrRouteTable`

Npeca75: > It is better to use OIDs called Table instead of Entry. Can you please change this and the related walk to `IP-FORWARD-MIB::ipCidrRouteTable`

done
yes, you are right, need to get used to it table/entry

includes/discovery/route.inc.php
Npeca75: > zeroDotZero and ccitt.0 are both 0.0, just with different mibs.
> 
> perhaps we can tell it to return numeric data instead to prevent duplicates.

i don't feel brave enough to make huge change in route.inc.php
as you could see, i choose pCidrRouteTable to "modernize" with SnmpQuery because of low impact, only few simdata was changed. Don't have enough fantasy to predict what will happen if i mess with rest of route.inc.php

murrant: @Npeca75 you do pretty well ;)  I think it is just these if os checks tend to bite us later.

murrant (approved):

PR #13906, opened 2022-04-10
Install: Pre-Install settings

murrant:
Create checkboxes to various relevant settings at the end of the install wizard.

Clear stats but remove stats right away instead of next daily.sh run. (needs confirmation now?)

Depends on #14444 

image
image
image

#### Please note

#### Testers


config/honeybadger.php
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * When reporting an exception, we'll automatically include relevant environment variables.
+     * See the Environment Whitelist (https://docs.honeybadger.io/lib/php/reference/configuration.html#environment-whitelist) for details.
+     * Use this section to filter or include env variables.
+     */
+    'environment' => [
+        /**
+         * List of environment variables that should be filtered out when sending a report to Honeybadger.
+         */
+        'filter' => [
```
Jellyfrog, line 30: Probably need to populate this with some database variables

config/honeybadger.php
```diff
@@ -0,0 +1,126 @@
...
+            // "credit_card_number",
+        ],
+    ],
+
+    /**
+     * The current version of your application. Use this to easily tie errors to specific releases or commits.
+     * If you'd like to automatically use the Git commit hash as the version, set this to:
+     *   'version' => trim(exec('git log --pretty="%h" -n1 HEAD')).
+     */
+    'version' => env('APP_VERSION'),
```
Jellyfrog, line 60: This should be Version::local();

Jellyfrog:
Does honeybager support this use case?

laf:
We definitely need docs on how and why we use honeybadger.io if this was to make it in. Not being transparent about that is what can really cause friction in a project.

From my perspective I'd like to know those answers as well and how we even use it when we're passing our API Key around. That could just open the account up to abuse. Who will have access to the data from the Lnms team? We need to call out that this is a US company and data will reside there as that may be a problem for European companies rolling out Lnms (I know we can just say well don't enable it).

murrant:
I actually am switching it to bugsnag.  it seems to have a lot better privacy protections.  And it should probably be clear that we can't prevent data leakage.

app/Providers/AppServiceProvider.php
```diff
@@ -55,6 +56,28 @@ public function boot(MeasurementManager $measure)
         $this->app->booted('\LibreNMS\Config::load');
         $this->app->booted('\App\Http\Controllers\Auth\SocialiteController::registerEventListeners');
 
+        // set up BugSnag
+        $this->app->booted(function () {
+            if (Config::get('reporting.error')) {
+                /** @var \Bugsnag\Client $bugsnag */
+                $reporting_key = '1df714c933472bdf666fc50b2a6a5d98';
```
Jellyfrog, line 63: why set this here and not in config/bugsnag.php ?

murrant: I didn't see an on/off value for bugsnag.  If we set this in .env reporting will be enabled before we bring up the database and check if `reporting.error` is enabled.

murrant:
Yeah, this is waiting on #13927 to clean up boot a bit... then it is more clear when we can do things.  Not very helpful to have the config loaded AFTER all ServiceProviders run

PR #13921, opened 2022-04-14
Port Validation Page to Laravel

murrant:
No longer run validation on page load, but after page is loaded.
That way the user isn't wondering why it is taking so long.

#### Please note

#### Testers


Jellyfrog:
Does it take the comments here into consideration?
https://github.com/librenms/librenms/pull/13334

murrant:
No, I did not change the structure at all.  Only change is that it makes a call after the page loads then populates the webui instead of waiting. 

I noticed after clicking go to the validation page when done installing, it was easy to think it wasn't doing anything because it just sat there.

PR #13926, opened 2022-04-15
Added: auth_ldap_skip_group_check when ldap_compare is not supported

### LINKED ISSUE #13916: Google LDAP does not support ldap_compare
### The problem

Google (GSuite) LDAP does not support "Compare", check https://support.google.com/a/answer/9167101?hl=en&ref_topic=9048334 

When trying to authenticate() a user an array of all groups is built so the authentication logic will always return "Protocol Error" even though the client is already authenticated.

I propose a configuration option (auth_ldap_skip_group_check) that would skip the ldap_compare portion of authenticate() so that an Exception is not returned.

Here is the diff:

```
diff --git a/LibreNMS/Authentication/LdapAuthorizer.php b/LibreNMS/Authentication/LdapAuthorizer.php
index 1373cf2ba..0d1b4fd89 100644
--- a/LibreNMS/Authentication/LdapAuthorizer.php
+++ b/LibreNMS/Authentication/LdapAuthorizer.php
@@ -29,6 +29,10 @@ class LdapAuthorizer extends AuthorizerBase
                 if ((Config::has('auth_ldap_binduser') || Config::has('auth_ldap_binddn')) && Config::has('auth_ldap_bindpassword')) {
                     $this->bind();
                 }
+                if (Config::get('auth_ldap_skip_group_check') === true) {
+                   // skip group check since google ldap does not support ldap_compare
+                   return true;
+                }
                 $ldap_groups = $this->getGroupList();
                 if (empty($ldap_groups)) {
                     // no groups, don't check membership
```

### Output of ./validate.php

```txt
Component | Version
--------- | -------
LibreNMS  | 22.3.0-37-g5d0670cdf
DB Schema | 2022_04_08_085504_isis_adjacencies_table_add_index (236)
PHP       | 8.0.17
Python    | 3.9.10
MySQL     | 8.0.26-google
RRDTool   | 1.7.2
SNMP      | 5.9.1

[OK]    Composer Version: 2.3.4
[INFO]  Detected Python Wrapper
[OK]    Connection to memcached is ok
	[FIX]: 
	Modified Files:
	 LibreNMS/Authentication/LdapAuthorizer.php
```

_No response_

_No response_

micko:
I've added an option to skip ldap_compare when the server (in this case Google (GSuite) LDAP) is not supporting the action.

You can view https://support.google.com/a/answer/9167101?hl=en&ref_topic=9048334 for reference.

fixes #13916

#### Please note

#### Testers


murrant:
You need to add all settings to config_definitions.json

murrant:
Sorry for forgetting this the first time, but also in `resources/lang/en/settings.php`

murrant:
@micko I had to merge upstream because your branch was too old.  https://docs.github.com/en/get-started/using-git/getting-changes-from-a-remote-repository

murrant (approved):

murrant (changes requested):
Oh, I just checked the surrounding code.  If you have no groups, there will be no group checks.

murrant:
So, I looked through the code and the default behavior is to perform no group check if auth_ldap_group and auth_ldap_groups are empty.  Also, the defaults for those config settings is empty.

So the default behavior is no group check.

It might be a good idea to add some documentation.

micko:
> So, I looked through the code and the default behavior is to perform no group check if auth_ldap_group and auth_ldap_groups are empty. Also, the defaults for those config settings is empty.
> 
> So the default behavior is no group check.
> 
> It might be a good idea to add some documentation.

As I said, "auth_ldap_groups" is needed to properly assign user permissions based on ldap data.

murrant:
@micko you cannot assign groups with your setting enabled or am I missing something?

murrant:
I see, and looking around a bit more I see we have `auth_ad_require_groupmembership`  can you make this setting name match that for consistency?

joelsdc:
I tested this PR, and although authentication does work after applying it, essentially we are bypassing the group functionality, so how do you differentiate regular users from admin users?

joelsdc:
Actually, not true, after adding to `config.php`:

```
$config['auth_ldap_groups']['librenms-admins']['level'] = 10;
```

It does find me as an admin:

```
Authentication Method: ldap
Password:
Authenticate user joel:
AUTH SUCCESS

User (63454735):
  username => joel
  realname => joel
  user_id => 63454735
  email => joel@example.com
  level => 0
Groups: ; cn=librenms-admins,ou=Groups,dc=example,dc=com
```

joelsdc:
Ok, so I was correct, when bypassing the **compare** section, membership is not verified (`ldap_compare()`) and therefore the permissions assigned is `0`.

joelsdc:
And after more testing, with the patch submitted by @micko and the following configuration, I can confirm authentication works with correct group assignment:

```
$config['auth_ldap_server'] = '127.0.0.1';
$config['auth_ldap_suffix'] = ',ou=Users,dc=example,dc=com';
$config['auth_ldap_groupbase'] = 'ou=Groups,dc=example,dc=com';
$config['auth_ldap_version'] = 3;
$config['auth_ldap_port'] = 1636;
$config['auth_ldap_userdn'] = true;
$config['auth_ldap_groupmembertype'] = 'username';
$config['auth_ldap_attr.uid'] = 'uid';
$config['auth_ldap_groups']['librenms-admins']['level'] = 10;
$config['auth_ldap_groups']['librenms-engineers']['level'] = 5;
$config['auth_ldap_group'] = 'cn=librenms-users,ou=Groups,dc=example,dc=com';
$config['auth_ldap_groupmemberattr'] = 'member';
```

NOTE: Our `stunnel4` is listening on port 1636. 

Adjust the group section to match your config.

micko:
I am running into another issue with the following scenario:

```
dn: uid=username1,ou=Project1,ou=Engineering,ou=Users,dc=domain,dc=com
dn: uid=username2,ou=Engineering,ou=Users,dc=domain,dc=com
```

where as userExists() will try `$search = ldap_search($connection, trim(Config::get('auth_ldap_suffix'), ','), $filter);` will try to match based on: `$config['auth_ldap_suffix'] = ',ou=Engineering,ou=Users,dc=domain,dc=com';` for username1 and it will fail.

Config I am testing with:

```
$config['auth_mechanism'] = 'ldap';
$config['auth_ldap_skip_group_check'] = true;
$config['auth_ldap_version'] = 3;
$config['auth_ldap_server'] = 'ldap://127.0.0.1';
$config['auth_ldap_port'] = 389;
$config['auth_ldap_prefix'] = "uid=";
$config['auth_ldap_suffix'] = ',ou=Engineering,ou=Users,dc=domain,dc=com';
$config['auth_ldap_groupbase'] = 'ou=Groups,dc=domain,dc=com';
$config['auth_ldap_group'] = 'cn=librenms-users,ou=Groups,dc=domain,dc=com';
$config['auth_ldap_groupmemberattr'] = 'member';
#$config['auth_ldap_uid_attribute'] = 'uidNumber';
$config['auth_ldap_userdn'] = true;
$config['auth_ldap_groups'] = ['librenms-users' => ['level' => 10]];
```

murrant (approved):

PR #13930, opened 2022-04-17
Automatic fixes for validation failures

murrant:
Only database and rrd converted in this PR.

#### Please note

#### Testers


LibreNMS/Validations/BaseValidation.php
```diff
@@ -25,12 +25,30 @@
...
+    {
+        if ($this->directory) {
+            foreach (glob(__DIR__ . "/$this->directory/*.php") as $file) {
+                $base = basename($file, '.php');
+                $class = __NAMESPACE__ . "\\$this->directory\\$base";
+                $validation = new $class;
+                if ($validation instanceof Validation && $validation->enabled()) {
+                    $validator->result($validation->validate(), $this->name);
+                }
+            }
```
Jellyfrog, line 49: I dont like this pattern, but weirdly enough there isn't a really good way.
Could use this and filter on `__NAMESPACE__` https://github.com/facade/ignition/blob/main/src/Support/ComposerClassMap.php#L29=

murrant: Also, I see a bug in that code :D

Issue #13938, opened 2022-04-22
check_ping broken for IPv6 addresses


steffann:
### The problem

Since last night my `check_ping` service check is broken.

My configuration:
- Check Type: ping
- Remote Host: 2a07:3503:1033:64::808:808
- Parameters: -6 -w 20,20% -c 100,40%

It used to work fine, but it broke "13 hours 3 minutes 51 seconds" ago, which is roughly the time my cron job ran daily.sh, so I suspect something changed in the code.

I now get the error message: "check_ping: Invalid hostname/address - 2a073503103364808808"

Has something in the code changed to strip away the colons?

### Output of ./validate.php

```txt
====================================
Component | Version
--------- | -------
LibreNMS  | 22.4.0-4-g4aaf6f290
DB Schema | 2022_04_08_085504_isis_adjacencies_table_add_index (236)
PHP       | 7.4.20
Python    | 3.8.10
MySQL     | 10.3.34-MariaDB-0ubuntu0.20.04.1
RRDTool   | 1.7.2
SNMP      | 5.8
====================================

[OK]    Composer Version: 2.3.5
[OK]    Dependencies up-to-date.
[OK]    Database connection successful
[OK]    Database schema correct
```

### What was the last working version of LibreNMS?

22.3.0

steffann:
I think I found the culprit: https://github.com/librenms/librenms/blob/4aaf6f290506e57c139be7987dabadc401e433b4/includes/services.inc.php#L126

PR #13939, opened 2022-04-22
Patch 2

steffann:
This is a proposed fix for #13938. Since LibreNMS 22.4.0 the services commands are cleaned by passing the parameters through `Clean::fileName()`. This causes problems for commands where the hostname is an IPv6 address, as that removes all colons from the address.

My proposed solution is to add `Clean::hostName()` that first tries to parse the provided parameter as an IPv4 or IPv6 address. If it parses as an address then it is converted back to a canonically formatted IP address and returned. Otherwise the parameter is passed to `Clean::fileName()` so that every input that is not an IP address is handled the same as before.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Util/Clean.php
+++ b/LibreNMS/Util/Clean.php
@@ -42,6 +42,24 @@ class Clean
     {
         return preg_replace('/[^a-zA-Z0-9\-._]/', '', $file);
     }
+    
+    /**
+     * Sanitize host name by removing all invalid characters.
+     * Behaves the same as Clean::fileName() except that it also allows valid IP addresses.
+     *
+     * @param  string  $file
+     * @return string|string[]|null
+     */
+    public static function hostName($string)
+    {
+        // If the string parses as a valid IP address it's ok
+        if (inet_pton($string) !== false) {
+            return $string;
+        }
+        
+        // Otherwise delegate to fileName
+        return self::fileName($string);
+    }
 
     /**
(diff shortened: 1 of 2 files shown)


LibreNMS/Util/Clean.php
```diff
@@ -43,6 +43,26 @@ public static function fileName($file)
...
+     * Sanitize host name by removing all invalid characters.
+     * Behaves the same as Clean::fileName() except that it also allows valid IP addresses.
+     *
+     * @param  string  $file
+     * @return string|string[]|null
+     */
+    public static function hostName($string)
+    {
+        // If the string parses as a valid IP address it's ok
+        $address = inet_pton($string);
```
murrant, line 56: Use `\LibreNMS\Util\IP::isValid()` instead.  In fact, maybe inline that because I plan on deleting that code anyway to fix the vulnerabilities for real.

steffann: I could replace the `inet_pton`/`inet_ntop` pair with a call to `filter_var($string, FILTER_VALIDATE_IP)`, but that would only validate, not reformat/clean the IP address notation. Either implementation would be safe, but the current implementation cleans up the notation, which would match the name of the class better I think :)

steffann:
That works fine too :)

murrant:
> That works fine too :)

Thanks for finding the issue and proposing a fix!  :)   I want to get this merged asap so I can release 22.4.1

bakerds:
Thank you both for this!

PR #13965, opened 2022-05-04
Fix cisco dbm sensors port link on discovery (NCS55A2)

ollb:
dBm sensors were not linked to port on Cisco NCS55A2 platform (iosxr)
causing false alarms triggered by optical modules in administratively shut down ports.

#### Please note

#### Testers


murrant:
In your test data there are some entPhysicalIndex_measured that are numbers, is this intended?

ollb:
Your question seems very relevant.

First of all, I think we should open another issue to fix the problem you raised since it seems not directly related to the change I've submited in this PR. Indeed, those integers are directly taken from `entSensorMeasuredEntity` in CISCO-ENTITY-SENSOR-MIB since 2011 (046833b67ee592eafa4b64758986a574b2e96baf). 

What librenms say : 

> entPhysicalIndex_measured (optional): If the sensor belongs to a physical entity then you can specify the entity type here.

What cisco mib say about entSensorMeasuredEntity :  

>         "This object identifies the physical entity for which the
>         sensor is taking measurements.  For example, for a sensor
>         measuring the voltage output of a power-supply, this object
>         would be the entPhysicalIndex of that power-supply; for a sensor
>         measuring the temperature inside one chassis of a multi-chassis
>         system, this object would be the enPhysicalIndex of that
>         chassis.
> 
>         This object has a value of zero when the physical entity
>         for which the sensor is taking measurements can not be
>         represented by any one row in the entPhysicalTable, or that
>         there is no such physical entity."

In the test data only cisco ios is affected (tests/data/ios_ir809.json, tests/data/iosxe_c9200l.json, tests/data/iosxe_c9400.json, tests/data/iosxe_ir1101.json, tests/data/iosxr_asr9001.json, tests/data/iosxr_asr9010.json, tests/data/iosxr_asr9901.json, tests/data/iosxr_ncs55a2.json).

I don't understand yet why those numerical value have been put in `entPhysicalIndex_measured`. I think we should determine if it is an error or a feature.

ollb:
> In your test data there are some entPhysicalIndex_measured that are numbers, is this intended?

@murrant I think the fact there are numbers in entPhysicalIndex_measured  is independent from this PR (it is the case since 2011 (https://github.com/librenms/librenms/commit/046833b67ee592eafa4b64758986a574b2e96baf) ).

Is there something else I can do ?

murrant:
Wow, looks like it was originally added in that PR, but wasn't used anywhere.  Some time later, it was co-opted for other ids.

murrant (dismissed):

ollb:
It seems some cleanup has been made in test data (especially #14963)

I've merged master branch and update test data for iosxr ncs55a2.

The checks are now ok.

PR #13967, opened 2022-05-05
Use display name when sending alerts to Sensu

TheMysteriousX:
Avoids IP addresses being registered as Sensu entities to allow alerts from other systems to be correctly correlated.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Sensu.php
+++ b/LibreNMS/Alert/Transport/Sensu.php
@@ -131,7 +131,7 @@ class Sensu extends Transport
                     'namespace' => $opts['namespace'],
                 ],
                 'system' => [
-                    'hostname' => $obj['display'],
+                    'hostname' => $obj['hostname'],
                     'os' => $obj['os'],
                 ],
             ],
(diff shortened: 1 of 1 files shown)


TheMysteriousX:
The behaviour is configurable, and the value changed here is non-default behaviour - default is $obj['hostname']. Sensu is fine with long/duplicate entity names.

ottorei (commented):
Is this actually the same name that is always present in the GUI? If so, then this is probably a good choice.

TheMysteriousX:
Yes, this is passed through to the GUI.

The default behaviour of the module is to just send the hostname or IP in all circumstances - I'm not proposing changing that as it's what most users would expect. This change just updates the 'shortname' naming scheme, which works poorly with IP's. The true hostname/IP is included in another field, so is still available in the event if needed.

Users can also choose to send the sysName instead, or the entity naming was written to be configurable so they could propose their own scheme.

murrant (approved):

murrant:
I guess I was thrown off by how wrong it was before :D

PR #13971, opened 2022-05-06
Workaround issues with lldp information from GS108Tv1

glance-:
The GS108Tv1 doesn't present the same information over snmp as they do
over lldp. This tries to detect such a remote device and figure out
which remote port we actually should map it to.

#### Please note

#### Testers


ottorei:
While this workaround probably works fine for the device in question, I think the LLDP-module should probably be converted into the newer Laravel-style module at some point and these kinds of OS-specific overrides should be handled in the OS classes. An example of this approach is the ISIS-module.

ottorei (approved):

PR #13978, opened 2022-05-11
Add Eaton SC200 OS model

slimey99uk:
Add OS Model for Eaton SC200 controller (as used in rectifier shelves)

#### Please note

#### Testers


includes/definitions/eaton-sc200.yaml
```diff
@@ -0,0 +1,33 @@
...
+    vlans: false
+    ucd-diskio: false
+    stp: false
+    ntp: false
+    fdb-table: false
+
+poller_modules:
+    mempools: false
+    cisco-vrf-lite: false
+    ospf: false
```
Jellyfrog, line 33: Please remove these, only add a explict disable if they are causing issues

Jellyfrog:
Does this need to be its own OS? or could it be merged with Eaton?

includes/definitions/eaton-sc200.yaml
slimey99uk: > Please remove these, only add a explict disable if they are causing issues

I specifically disabled these because the CPU is woefully underpowered on these controllers, so was aiming to minimise the number of SNMP requests sent to the box. If that's the wrong way to do it, I'm happy to remove this from the model.

slimey99uk: Okay, I've pushed another commit with those disables removed

PR #13980, opened 2022-05-13
Fix polling and discovery of FortiGate cluster sensors

mkuurstra:
As discussed [here](https://community.librenms.org/t/name-based-polling-not-working/18567)

All these checks where not working because of an error in the discovery. I also moved 1 sensor from `state` to `count` as it was counting cluster members, so that sensor type seemed more appropriate.

With these changes the customised state polling for Fortigate wasn't needed anymore so I deleted that file as well.

#### Please note

#### Testers


murrant:
Looks good.

As for the deleted test file, does the new 1500d test file replace it?  what is the difference?

mkuurstra:
I might’ve misinterpreted the 3th checkpoint asking to update test data when making discovery/polling changes.

If I remember correctly the tests threw an error. My interpretation was that the error was caused because by fixing the created test I added sync cluster status for every member in the cluster, which wasn’t already in the captured test data.

murrant:
You are probably correct, but no way for me to tell that without comparing the changes from the previous test data :)

Would be great if you can re-add the snmprec file and generate the json file from it from scratch.

mkuurstra:
Added the test data like you requested. I also noticed that I was the one who added the 1500d data 3 years ago and wanted to update and obscure some of the content.

PR #13987, opened 2022-05-19
fix: inconsistency in ldap starttls config parameter

nox-x:
Fixes an inconsistency in the value of config parameter `auth_ldap_starttls`.

`lnms` only allows to set `auth_ldap_starttls` to `required` (see. [misc/config_definitions.json](https://github.com/librenms/librenms/blob/c3cf527301c6594321caf7715549fc66f7cd660a/misc/config_definitions.json#L597)) but both [LibreNMS/Authentication/LdapAuthorizationAuthorizer.php](https://github.com/librenms/librenms/blob/c3cf527301c6594321caf7715549fc66f7cd660a/LibreNMS/Authentication/LdapAuthorizationAuthorizer.php) and [LibreNMS/Authentication/LdapAuthorizer.php](https://github.com/librenms/librenms/blob/c3cf527301c6594321caf7715549fc66f7cd660a/LibreNMS/Authentication/LdapAuthorizer.php) only utilize `auth_ldap_starttls` with value `require` (instead of `required`)

This PR now testing  `auth_ldap_starttls` for `required` in both scripts (`Ldap{Authorization}Authorizer.php`), enabling the user to successful authenticate via WebUI if `auth_ldap_starttls` is set to `required`.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
+++ b/LibreNMS/Authentication/LdapAuthorizationAuthorizer.php
@@ -53,9 +53,9 @@ class LdapAuthorizationAuthorizer extends AuthorizerBase
             ldap_set_option($this->ldap_connection, LDAP_OPT_PROTOCOL_VERSION, Config::get('auth_ldap_version'));
         }
 
-        if (Config::get('auth_ldap_starttls') && (Config::get('auth_ldap_starttls') == 'optional' || Config::get('auth_ldap_starttls') == 'require')) {
+        if (Config::get('auth_ldap_starttls') && (Config::get('auth_ldap_starttls') == 'optional' || Config::get('auth_ldap_starttls') == 'required')) {
             $tls = ldap_start_tls($this->ldap_connection);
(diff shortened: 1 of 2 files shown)


Jellyfrog (approved):

Jellyfrog:
Nice catch 🥳

joachimtingvold:
FYI, this change breaks all setups that has `$config['auth_ldap_starttls'] = 'require';` in the configuration file. You get the error "Confidentiality required" upon login (even for local non-LDAP users) if you have LDAP servers that only accepts StartTLS connections.

image

```
librenms@librenms1:~$ /opt/librenms/scripts/auth_test.php -u foobar
Authentication Method: ldap
Password:
Authenticate user foobar:
Error: LibreNMS\Exceptions\AuthenticationException thrown!
Confidentiality required
```

The fix is easy enough (change `require` to `required`). I guess I don't expect a fix to be "backwards compatible" with old inconsitency, but, yeah :D

murrant:
I think this was fixed the wrong way.  You should change the config variable to require.

Changing the existing check is almost guaranteed to break existing setups.

PR #13991, opened 2022-05-24
Fix 500 error in validations when capabilities are not supported or set

gs-kamnas:
This change addresses a 500 error that may occur when validate.php is invoked via the UI if the fping executable lacks the appropriate capabilities, such as in a container lacking support for capabilities.

I believe this behavior occurs as a result of the matches array returned by preg_match being an empty array (not null) if the regex is valid but fails to match the provided string. Therefore, I have updated the match/no-match check to instead use the return value of preg_match which returns 1 (per the PHP API docs) if the regex matches the provided string.

#### Please note

- [NA] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.
- [NA] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers

### DIFF
--- a/LibreNMS/Validations/Programs.php
+++ b/LibreNMS/Validations/Programs.php
@@ -105,9 +105,9 @@ class Programs extends BaseValidation
 
         if ($getcap = $this->findExecutable('getcap')) {
             $getcap_out = shell_exec("$getcap $cmd");
-            $found = preg_match("#^$cmd = (.*)$#", $getcap_out, $matches);
+            preg_match("#^$cmd = (.*)$#", $getcap_out, $matches);
 
-            if (! $found || ! Str::contains($matches[1], 'cap_net_raw+ep')) {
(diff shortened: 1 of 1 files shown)


LibreNMS/Validations/Programs.php
```diff
@@ -105,9 +105,9 @@ private function failFping($validator, $cmd, $output)
 
         if ($getcap = $this->findExecutable('getcap')) {
             $getcap_out = shell_exec("$getcap $cmd");
-            preg_match("#^$cmd = (.*)$#", $getcap_out, $matches);
+            $found = preg_match("#^$cmd = (.*)$#", $getcap_out, $matches);
 
-            if (is_null($matches) || ! Str::contains($matches[1], 'cap_net_raw+ep')) {
```
murrant, line 110: probably, I would have just changed is_null() to empty(), this works though

murrant (approved):

PR #13992, opened 2022-05-24
Add support for Cisco Nexus 3550 series (formerly Exalink Fusion) devices

gs-kamnas:
This change adds support for Cisco Nexus 3550  (formerly Exalink Fusion) devices - https://www.cisco.com/c/en/us/products/switches/nexus-3550-series/index.html

#### Please note

- [NA] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.

#### Testers


includes/definitions/ciscon3550.yaml
```diff
@@ -0,0 +1,26 @@
+os: ciscon3550
```
murrant, line 1: Shouldn't the OS be `exalink-fusion`, I checked and the firmware files bear that name.

includes/polling/ports/ciscon3550.inc.php
```diff
@@ -0,0 +1,63 @@
+<?php
+/**
+ * ciscon3550.inc.php
```
murrant, line 3: I think this file is getting called twice by the poller because of this on line 320 of ports.inc.php
```php
if (file_exists($os_file)) {
    require $os_file;
}
```

includes/definitions/ciscon3550.yaml
gs-kamnas: Both would be technically correct in my opinion as Cisco has purchased Exablaze and rebranded their products under the Cisco Nexus branding (https://www.cisco.com/c/en/us/about/corporate-strategy-office/acquisitions/exablaze.html#~tab-products).

That being said, I'm personally in favor of keeping Cisco in the os name if possible; would cisco-exalink-fusion or similar be acceptable or does that seem too verbose?

gs-kamnas:
Valid point, will refactor later today to rename the OS name (only) back to exalink-fusion.

gs-kamnas:
@murrant Renamed as requested, apologies for the delay in getting back.

includes/definitions/ciscon3550.yaml
gs-kamnas: Renamed the OS as advised; valid point.

murrant (approved):

PR #13993, opened 2022-05-25
feat(apps/mysql): add error-state to non-responsive mysql-servers

bennet-esyoil:
This PR updates the `app_state` field for mysql-Apps that returned an error during polling.
This can be caused by an offline server or invalid user configuration.

I also added an alert template to show how to filter these incidents.

#### Please note

#### Testers

### DIFF
--- a/includes/polling/applications/mysql.inc.php
+++ b/includes/polling/applications/mysql.inc.php
@@ -100,7 +100,7 @@ $mapping = [
 
 $data = explode("\n", $mysql);
 if (sizeof($data) > 0 && strpos($data[0], 'ERROR') === 0) {
-    $mysql= 'ERROR';
+    $mysql = 'ERROR';
 }
 
 $map = [];
--- a/includes/polling/applications/mysql.inc.php
+++ b/includes/polling/applications/mysql.inc.php
@@ -99,6 +99,7 @@ $mapping = [
 ];
 
 $data = explode("\n", $mysql);
+if (sizeof($data) > 0 && strpos($data[0], "ERROR") === 0) $mysql= "ERROR";
 
 $map = [];
 foreach ($data as $str) {
--- a/includes/polling/applications/mysql.inc.php
+++ b/includes/polling/applications/mysql.inc.php
@@ -99,7 +99,9 @@ $mapping = [
 ];
 
 $data = explode("\n", $mysql);
-if (sizeof($data) > 0 && strpos($data[0], "ERROR") === 0) $mysql= "ERROR";
+if (sizeof($data) > 0 && strpos($data[0], "ERROR") === 0) {
+    $mysql= "ERROR";
+}
 
 $map = [];
 foreach ($data as $str) {
--- a/includes/polling/applications/mysql.inc.php
+++ b/includes/polling/applications/mysql.inc.php
@@ -99,8 +99,8 @@ $mapping = [
 ];
 
 $data = explode("\n", $mysql);
-if (sizeof($data) > 0 && strpos($data[0], "ERROR") === 0) {
-    $mysql= "ERROR";
+if (sizeof($data) > 0 && strpos($data[0], 'ERROR') === 0) {
+    $mysql= 'ERROR';
 }
 
 $map = [];
--- a/includes/polling/applications/mysql.inc.php
+++ b/includes/polling/applications/mysql.inc.php
@@ -99,9 +99,6 @@ $mapping = [
 ];
 
 $data = explode("\n", $mysql);
-if (sizeof($data) > 0 && strpos($data[0], 'ERROR') === 0) {
-    $mysql = 'ERROR';
-}
 
 $map = [];
 foreach ($data as $str) {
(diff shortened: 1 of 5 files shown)


ottorei (dismissed):

includes/polling/applications/mysql.inc.php
```diff
@@ -99,6 +99,9 @@
 ];
 
 $data = explode("\n", $mysql);
+if (sizeof($data) > 0 && strpos($data[0], 'ERROR') === 0) {
```
murrant, line 102: Probably the correct place to update is https://github.com/librenms/librenms/blob/master/includes/polling/functions.inc.php#L447

and change it to check if the string starts with the error message and then mark it as ERROR.  That way we don't delete the additional context after the ERROR message.

The other option would be to set $state directly to the update_application() call.

misc/alert_rules.json
```diff
@@ -527,5 +527,10 @@
     "rule": "applications.app_type = \"suricata\" && application_metrics.metric = \".total_error_delta\" && application_metrics.value >= \"2\"",
     "name": "Suricata Packet Error >= 2%",
     "severity": "critical"
+  },
+  {
+    "rule": "applications.app_type = \"mysql\" && applications.app_state != \"OK\"",
```
murrant, line 532: What about other applications?  Also, is "ERROR" guaranteed to be MySQL server not responding?  Or could it be another issue?

includes/polling/applications/mysql.inc.php
bennet-esyoil: From how I understand that function there is no real way to save an error-message in the `application` table. Also the `app_state` would only hold 32 chars and the `app_state` 8... So that'd not fit with a full error-message.

We could modify the log_event function that is later called to include the error-message from the response, but i'm not sure what implication that would have for all the other apps really.

If I move it to the $status parameter of the update_application function it would not trigger the alerts (I think) since there is no comparison here https://github.com/librenms/librenms/blob/master/includes/polling/functions.inc.php#L474 that references $status. Should I add that so it also triggers when I update the $status?

misc/alert_rules.json
bennet-esyoil: This could in theory be extended to other application, but my use-case just includes mysql so I made it for that. I can also revert the changes here since it maybe really isn't wort it.

Also ERROR: should be pretended to any error, since that's coming from the php snmpd-extend script. Tested it with server-offline, wrong port, wrong IP and wrong user-credentials. I don't expect there to come anything else.

includes/polling/applications/mysql.inc.php
murrant: I'm saying inside update_applicatuon(), status is detected.  You need to improve the detection to see if it starts with one of the listed strings set the status to the exact string.  The full message should be in the response field.

Hope that is clear enough, I'm on my phone.

includes/polling/functions.inc.php
```diff
@@ -459,12 +459,18 @@ function update_application($app, $response, $metrics = [], $status = '')
...
+         * also set app_status to the raw response
+         */
+        if (Str::startsWith($response, [
+            'ERROR',
+            'LEGACY',
+            'UNSUPPORTED',
+            'Traceback (most recent call last):'
         ])) {
+            $data['app_status'] = $response;
             $data['app_state'] = 'ERROR';
```
bennet-esyoil, line 473: This now would result in an Error-State even if the application returns an `UNSUPPORTED` or `LEGACY`. Is this something we can live with or should that also be set correctly?

murrant: Sorry to bring this up, but it seems as though you have bumped into some deep issues in the application code.

I just looked through the code a bit and currently it seems to confuse app_state and app_status a lot.

app_state should be a fixed set of values 'OK', 'ERROR', 'LEGACY', 'UNSUPPORTED', and 'UNKNOWN' (maybe more) to make alerting easy.

app_status seems mostly to be the output of the script, but not 100%

The update_application() function does not line up with that, confusing things.

```php
 * @param  array  $app  app from the db, including app_id
 * @param  string  $response  This should be the return state of Application polling
 * @param  array  $metrics  an array of additional metrics to store in the database for alerting
 * @param  string  $status  This is the current value for alerting
 */
function update_application($app, $response, $metrics = [], $status = '')
```

It probably should be this:

```php
 * @param  array  $app  app from the db, including app_id (Note: we should probably be using models instead of arrays)
 * @param  string  $app_status  This should be the output from the application script. Shown when an ERROR state is set.
 * @param  array  $metrics  a key value array of additional metrics to store in the database for alerting.  Must manually set _prev fields if you want them.
 * @param  string  $state The state for alerting. If not set, will attempt to detect from $response, or set to UNKNOWN
 */
function update_application($app, $app_status, $metrics = [], $state = 'UNKNOWN')
```

bennet-esyoil: Yes! That's what I was thinking... Should I rebuild that and make a sepearate PR based on this one that updates the behaviour for all other applications?

murrant (approved):

PR #14007, opened 2022-06-01
Procurve hardware description cleanup

murrant:
Include part number and other details where available
Covers 100% of known sysDescr
Adds Aruba Instant On parsing
adds string replacements for OS fields

Could use hardware: ENTITY-MIB::entPhysicalDescr.1. It is not supported on all devices and strings contain a lot of fluff and some are just "HP", I leave that for another PR

#### Please note

#### Testers


tests/data/procurve.json
```diff
@@ -549,7 +549,7 @@
                     "sysDescr": "HP Stack, revision KA.15.03.3004, ROM KA.15.05 (/sw/code/build/tam(KA_15_03))",
                     "sysContact": "<private>",
                     "version": "KA.15.03.3004",
-                    "hardware": "",
+                    "hardware": "Stack",
```
Jellyfrog, line 552: This seems wrong

tests/data/procurve_131.json
```diff
@@ -3,12 +3,12 @@
...
             "devices": [
                 {
-                    "sysName": "",
+                    "sysName": null,
                     "sysObjectID": ".1.3.6.1.4.1.11.2.3.7.11.131",
                     "sysDescr": "HP J9625A 2620-24-PoEP Switch, revision RA.16.02.0012, ROM RA.15.13 (/ws/swbuildm/rel_spokane_qaoff/code/build/xform(swbuildm_rel_spokane_qaoff_rel_spokane))",
                     "sysContact": null,
                     "version": "RA.16.02.0012",
-                    "hardware": "2620-24-PoEP",
+                    "hardware": "J9625A 2620-24-PoEP",
```
Jellyfrog, line 11: I'm thinking we might want to filter out the JxxxxA part, pattern seems to be `J[A-Z0-9]{4}A` or maybe `J[A-Z0-9][0-9]{3}A`

tests/data/procurve.json
murrant: I thought it was more right than "".  This is also something that could be fixed by `ENTITY-MIB::entPhysicalDescr.1` maybe.

tests/data/procurve_131.json
murrant: I specifically included the part number.  I find it more useful than the model name.

mpikzink: I agree to @murrant. Would it make sense to put this information in the (actually empty) features tag?

murrant: Not straight forward to do. You have this format: 516733-B21, some end with B, or C, sometimes it is after the model and sometimes before, and sometimes there is no model and only the part number.

mpikzink: That's a point...

murrant: @mpikzink What I mean is, you can't split out the part number without doubling or more the number of regexes required.

Jellyfrog (approved):

PR #14037, opened 2022-06-13
Fix use of incorrect variable to retrieve current sensor value

dennypage:
The current Axis sensor discovery uses the wrong variable to retrieve the current sensor value. This results in attempting to perform an array access into a string:

TypeError: Cannot access offset of type string on string in /opt/librenms/includes/discovery/sensors/temperature/axiscam.inc.php:15

This patch discovery to use the correct variable.

#### Please note

#### Testers


murrant (approved):
The original code here is very silly.
Thanks for the fix!

dennypage:
You're quite welcome.

PR #14039, opened 2022-06-13
Use Laravel schema migration command directly

Jellyfrog:
It's time to get rid of the old migration files :)

#### Please note

#### Testers


app/Console/Commands/MigrateCommand.php
```diff
@@ -0,0 +1,68 @@
...
+
+use App\Console\LnmsCommand;
+use Illuminate\Contracts\Cache\LockTimeoutException;
+use Illuminate\Support\Facades\Cache;
+use Log;
+use Schema;
+
+class MigrateCommand extends LnmsCommand
+{
+    protected $signature = 'migrate:librenms {--force : Force running even if already running}';
```
Jellyfrog, line 30: Didn't manage to override the real command, so gave it a "work in progress" name.

Jellyfrog: @murrant can probably solve it. But come thinking about it, wondering if this just complicates stuff for no reason?

murrant:
What is the goal here?  Couldn't you just remove references to the old update.php script?

app/Console/Commands/MigrateCommand.php
murrant: Wouldn't schema:migrate be more appropriate.

"THE INTERNET" indicates setting $name to the command name will work, but I have no idea.

Jellyfrog:
I think its good to get rid of old unused files in general.
But not sure if the custom migration command adds that much?

app/Console/Commands/MigrateCommand.php
Jellyfrog: overriding the command "name" itself works, but not extending it or trying to call the original one.
Think its due to how its created;

https://github.com/laravel/framework/blob/ccefc764871fe263dd20f8f7108c18180f78241b/src/Illuminate/Database/MigrationServiceProvider.php#L117-L122

murrant:
@Jellyfrog I'm a bit dubious about removing the old migrations just yet.  They are still needed for all those stupidly old installs.

murrant:
Sounds good.

What do you think about overriding the regular migrate command?

Jellyfrog:
> Sounds good.
> 
> What do you think about overriding the regular migrate command?

Didnt find a way to call the underlying command then

app/Console/Commands/MigrateCommand.php
```diff
@@ -0,0 +1,77 @@
...
+        try {
+            if ($this->option('force-run')) {
+                optional($lock)->forceRelease();
+            }
+            optional($lock)->block(10);
+
+            // These are for the real MigrateCommand, but we hide them from the user
+            $this->addOption('--database', null, InputOption::VALUE_OPTIONAL);
+            $this->addOption('--force', null, InputOption::VALUE_OPTIONAL, '', true);
+            $this->addOption('--pretend');
```
murrant, line 61: Might want to allow --pretend.   It helps people fix migrations that fail.

Jellyfrog:
Latest laravel got support for running commands one at the time. Should rely on that instead.

murrant (approved):

PR #14043, opened 2022-06-14
Add new executable finder

Jellyfrog:
This also fixes the problem where validate doesn't complain about missing binaries

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Config.php
+++ b/LibreNMS/Config.php
@@ -32,9 +32,12 @@ use Illuminate\Database\QueryException;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use LibreNMS\Data\Store\Rrd;
+use LibreNMS\DB\Eloquent;
 use LibreNMS\Util\Debug;
+use LibreNMS\Util\Notifications;
 use LibreNMS\Util\Version;
 use Log;
+use Symfony\Component\Process\Process;
 
 class Config
 {
@@ -138,6 +141,60 @@ class Config
         }
 
         return Arr::get(self::$config, $key, $default);
+    }
+
+    public static function getExecutable(string $key): string
+    {
+        $binary = self::get($key, $key);
+
+        if (Str::startsWith($binary, '/')) {
+            return $binary;
+        }
+
+        $process = new Process(['whereis', '-b', $binary]);
+        $process->run();
+        if ($process->getExitCode() === 0) {
+            $output = trim($process->getOutput());
+            $list = substr($output, strpos($output, ':') + 1);
+            $possible_targets = explode(' ', $list);
+        } else {
+            $process = new Process(['which', $binary]);
+            $process->run();
+            $possible_targets = [trim($process->getOutput())];
+        }
+
+        foreach ($possible_targets as $target) {
+            if (is_executable($target)) {
+                self::persist($key, $target);
+
+                return $target;
+            }
+        }
+
+        if (Eloquent::isConnected()) {
+            Notifications::create(
+                'Missing executable',
+                "The executable: $binary, cannot be found in your \$PATH, please make sure you install it!",
+                'Config',
+                1
+            );
+        }
+
+        return $binary;
+    }
+
+    public static function getAllExecutables(): array
+    {
+        $definitions = self::getDefinitions();
+        $executables = [];
+
+        foreach ($definitions as $key => $definition) {
+            if ($definition['type'] == 'executable') {
+                $executables[$key] = $definition['default'];
+            }
+        }
+
+        return $executables;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


LibreNMS/Config.php
```diff
@@ -137,6 +137,41 @@ public static function get($key, $default = null)
         return Arr::get(self::$config, $key, $default);
     }
 
+    public static function getExecutable(string $key): string
+    {
+        $binary = self::get($key, $key);
+
+        if (Str::startsWith($binary, '/')) {
```
murrant, line 144: maybe to prevent whereis calls every time in a bad situation, just check self::has($key)

LibreNMS/Config.php
```diff
@@ -137,6 +137,41 @@ public static function get($key, $default = null)
...
+            return $binary;
+        }
+
+        $output = `whereis -b $binary`;
```
murrant, line 148: What happens when whereis doesn't exist?

murrant: Is bash guaranteed @mpikzink ?

I know the daily.sh script requires bash, but I"m not sure about any other scripts.

Jellyfrog: `whereis` doesn't only rely on `PATH` to find binaries, I guess thats why we use it.
But if `PATH` is not set, `whereis` itself can't be found :) cat and mouse

LibreNMS/Config.php
Jellyfrog: Either we try to find a binary once, and save the result forever, then the validator can complain, but user needs to update the config manually.
Or we test it each time, which causes it to auto-solve if the user installs the binary...

Jellyfrog:
Here is what's left to solve/decide:

### Executable not found
1. Test once and save the result forever, no matter the result. Validator will complain, but user needs to update the config manually.
2. Test it each time until found, then save, which causes it to auto-solve if the user installs the executable.

Both methods should(?) also send a notification (they will be de-duped, so only one per binary)

### Method to find executable

1. `whereis -b`, it looks in PATH and also at the systems default paths. portability? (`-b` flag doesn't exist on macos)
2. `which`, more portable? only looks in PATH
4. `find`, veeery slow
5. Only static paths (user need to configure each manually)

Jellyfrog:
Test fails due to chain: `composer install` -> `php artisan package:discover --ansi` ->  `App\Console\Commands\DeviceAdd->__construct()` ...
```
#0 librenms/LibreNMS/SNMPCapabilities.php(99): LibreNMS\Config::getExecutable('...')
#1 librenms/LibreNMS/SNMPCapabilities.php(45): LibreNMS\SNMPCapabilities::detectCapabilities()
#2 librenms/LibreNMS/SNMPCapabilities.php(72): LibreNMS\SNMPCapabilities::supportsSHA2()
#3 librenms/LibreNMS/SNMPCapabilities.php(62): LibreNMS\SNMPCapabilities::authAlgorithms()
#4 librenms/app/Console/Commands/DeviceAdd.php(46): LibreNMS\SNMPCapabilities::supportedAuthAlgorithms()
#5 [internal function]: App\Console\Commands\DeviceAdd->__construct()
```
Not sure if we can defer it or if we should remove it. seems very wasteful it needs to spawn a process each time artisan is called.

murrant:
I believe the reason whereis is used, is because of portability.  Some OS didn't have which.

murrant:
One solution would be to find the binary during install or daily.sh and not application boot up :D

LibreNMS/Config.php
murrant: `whereis whereis` lol

LibreNMS/Config.php
```diff
@@ -137,6 +138,67 @@ public static function get($key, $default = null)
...
+        foreach ($possible_targets as $target) {
+            if (is_executable($target)) {
+                self::persist($key, $target);
+
+                return $target;
+            }
+        }
+
+        if (Eloquent::isConnected()) {
+            $notification = new Notification();
```
murrant, line 170: Port this to the new \LibreNMS\Util\Notifications helper?

It would be nice to remove the notification if the issue is resolved.

Jellyfrog: Sure but maybe for next release then

laf:
Is this still relevant and being worked on or should we close it down @Jellyfrog?

LibreNMS/Config.php
mpikzink: Since @laf asked about relevance, I remembered the discussion and read up on it a bit because it interested me. With a POSIX shell, you can achieve similar behaviour with ‘command -v bash’. I have also seen this in some Debian ‘.deb’ packages. With POSIX compatibility, we don't cover all cases either, but it will cover most cases.

murrant:
#

PR #14048, opened 2022-06-17
update scripts/ifAlias to work on FreeBSD as well as Linux

VVelox:
Allows interface descriptions set in FreeBSD to be read.

```
[root@vixen42]0|/usr/local/etc/namedb/master>ifconfig em3
em3: flags=8802<BROADCAST,SIMPLEX,MULTICAST> metric 0 mtu 1500
        description: this is a test
        options=81049b<RXCSUM,TXCSUM,VLAN_MTU,VLAN_HWTAGGING,VLAN_HWCSUM,LRO,VLAN_HWFILTER>
        ether 00:15:17:bf:9e:46
        media: Ethernet autoselect
        status: no carrier
        nd6 options=29<PERFORMNUD,IFDISABLED,AUTO_LINKLOCAL>
[root@vixen42]0|/usr/local/etc/namedb/master>
[kitsune@vixen42]0|/home/kitsune/github/librenms/scripts (fbsd-ifdesc $)>./ifAlias -n 4
.1.3.6.1.2.1.31.1.1.1.18.5
string
this is a test
```

Will pull the description config from there if set. :)

Also pointed it at "/etc/network/interfaces" if Ubuntu. Does not currently support netplan.

#### Please note

#### Testers


Jellyfrog:
Does ifconfig behave the same on bsd and Linux? 
In that case I would rather check if the binary exists instead of checking OS, then it works on other flavors of bsd also

VVelox:
> Does ifconfig behave the same on bsd and Linux? In that case I would rather check if the binary exists instead of checking OS, then it works on other flavors of bsd also

Ifconfig on other unixes and Linux are two totally different thing. It is utterly unmaintained on Linux and increasingly not present in installs, having been replaced by ip where as it has under gone lots of development on other OSes.

On top of that using uname will make it easier to add support for other systems going forward.

That said checking the man files, description is supported on NetBSD and OpenBSD as well. Updating it now to only call uname once and also check if it is a BSD.

VVelox:
Updated it to only call uname once and now operates in a generic manner on BSDs as looking at the docs it works the same on the other two as well. :)

scripts/ifAlias
```diff
@@ -18,8 +18,17 @@ BASE='.1.3.6.1.2.1.31.1.1.1.18'
...
+COUNT="$(echo "${UNAME}" | grep -ci 'bsd$')"
+if [ "${COUNT}" -eq 1 ]; then
+    UNAME="BSD"
+fi
+
 # cache ip link output
-IP_LINK="$(ip link)"
+if [ "${UNAME}" = 'Linux' ]; then
+    IP_LINK="$(ip link)"
+fi
```
mpikzink, line 31: Do we have to cache this? We could thus do without a global variable.

scripts/ifAlias
```diff
@@ -18,8 +18,17 @@ BASE='.1.3.6.1.2.1.31.1.1.1.18'
 GET_TYPE="$1"
 GET_OID="$2"
 
+UNAME="$(/usr/bin/uname)"
+
+COUNT="$(echo "${UNAME}" | grep -ci 'bsd$')"
+if [ "${COUNT}" -eq 1 ]; then
```
mpikzink, line 24: `if [ "$(echo "${UNAME}" | grep -ci 'bsd$')"  -eq 1 ] ; then`  We could thus do without a global `COUNT` variable. It's reassigned as local var in `alias_from_ip_link`

scripts/ifAlias
```diff
@@ -51,20 +60,36 @@ interface_id()
...
-                    then
-                        echo -n "$N"
-                        return 0
-                    fi
-                    L="$N"
-                done
+                if [ "${UNAME}" = 'Linux' ]; then
+                    for N in $(echo "$IP_LINK" | grep -oE "^[0-9]+:" | cut -d':' -f 1)
+                    do
+                        if [ "$L" = "$ID" ] || [ ! "$ID" ]
```
mpikzink, line 76: ` if [ "$L" = "$ID" ] || [ -z  "$ID" ]`

scripts/ifAlias
```diff
@@ -51,20 +60,36 @@ interface_id()
...
+                        then
+                            printf '%s' "$N"
+                            return 0
+                        fi
+                        L="$N"
+                    done
+                else
+                    for N in $(ifconfig | grep -E '^[A-Za-z]+' | cut -d: -f1 | cat -n -b | sed 's/^  *//' | sed 's/[\t\ ].*//'); do
+                        if [ "$L" = "$ID" ] || [ ! "$ID" ]
```
mpikzink, line 85: `if [ "$L" = "$ID" ] || [ -z "$ID" ]`

scripts/ifAlias
```diff
@@ -73,42 +98,70 @@ interface_id()
...
-    local CONFIG_FILE=
+    CONFIG_FILE=
 
     if [ -x "$DISTRO_BIN" ]
     then
-        case "$("$DISTRO_BIN" | cut -d " " -f 1)" in
+        if [ "$(/usr/bin/uname)" = 'Linux' ]; then
+            DISTRO_VAR="$($DISTRO_BIN | cut -d " " -f 1)"
+        else
+            DISTRO_VAR="$(/usr/bin/uname)"
```
mpikzink, line 117: `DISTRO_VAR="$UNAME"`

scripts/ifAlias
```diff
@@ -73,42 +98,70 @@ interface_id()
...
 alias_from_interfaces_config_file()
 {
-    local IFACE="$1"
-    local CONFIG_FILE=
+    CONFIG_FILE=
 
     if [ -x "$DISTRO_BIN" ]
     then
-        case "$("$DISTRO_BIN" | cut -d " " -f 1)" in
+        if [ "$(/usr/bin/uname)" = 'Linux' ]; then
```
mpikzink, line 114: `if [ "$UNAME" = 'Linux' ]; then"`

scripts/ifAlias
```diff
@@ -73,42 +98,70 @@ interface_id()
...
-    local IFACE="$1"
-    ip link show "$IFACE" | grep -e "^[[:space:]]*alias[[:space:]]" | sed -e 's/^[[:space:]]*alias //'
+    case "${UNAME}" in
+        Linux)
+            ip link show "$1" | grep -e "^[[:space:]]*alias[[:space:]]" | sed -e 's/^[[:space:]]*alias //'
+            ;;
+        BSD)
+            COUNT=$(/sbin/ifconfig "$1" | grep 'description:' | head -n 1 | cut -d: -f 2- | wc -l)
+            if [ "${COUNT}" -eq 1 ]; then
+                /sbin/ifconfig "$1" | grep 'description:' | head -n 1 | cut -d: -f 2- | sed 's/^ //'
```
mpikzink, line 158:
```
LINK="$(/sbin/ifconfig "$1" | grep 'description:' | head -n 1 | cut -d: -f 2- | sed 's/^ //')"
if [ -n "$LINK" ]; then
echo "$LINK"
```

scripts/ifAlias
```diff
@@ -73,42 +98,70 @@ interface_id()
...
+        else
+            DISTRO_VAR="$(/usr/bin/uname)"
+        fi
+
+        case "${DISTRO_VAR}" in
             Debian)
                 CONFIG_FILE="/etc/network/interfaces"
             ;;
+            Ubuntu)
+                CONFIG_FILE="/etc/network/interfaces"
```
mpikzink, line 125: We need to handle interfaces.d/* as well

mpikzink (commented):
Nice work! I like it much. I haven't noticed about ifAliases in Linux. Have we a description in our doc how to set them with ip or in interfaces syntax?

scripts/ifAlias
```diff
@@ -73,42 +98,70 @@ interface_id()
...
+                CONFIG_FILE="/etc/rc.conf"
             ;;
         esac
     fi
     if [ "$CONFIG_FILE" ]
     then
-         echo "$(grep -i "^# $IFACE:" $CONFIG_FILE | sed "s/^# $IFACE: //i")"
+         # echo squashes possible multi line replies to a single line
+         TO_ECHO_AND_MAKE_LING_HAPPY="$(grep -i "^# $1:" $CONFIG_FILE | sed "s/^# $1: //i")"
+         echo "${TO_ECHO_AND_MAKE_LING_HAPPY}"
```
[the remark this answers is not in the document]
VVelox: Depends on if we want to squash potential multi line returns down to a single line for config parsing. Went with the current operation which is to squash mutli line replies to a single line.

And aye. Docs should be updated. Grabbing that now.

scripts/ifAlias
VVelox: Results in fewer external calls.

Really should do the same with ifconfig, but that can be done some other time.

VVelox:
Okay. Should all be good now and every one happy. :)

VVelox:
> Does ifconfig behave the same on bsd and Linux? In that case I would rather check if the binary exists instead of checking OS, then it works on other flavors of bsd also

To follow up on my earlier reply, just check Linux docs out of curiosity as it is has been forever since I've really used it on there(need ip for nearly everything I do with Linux) and yup, very much does not support the description field.

murrant:
I have no idea what this script does, but yes, ifconfig should **not** be used on Linux.

scripts/ifAlias
```diff
@@ -73,42 +97,77 @@ interface_id()
...
+    if [ "$CONFIG_FILE" ]; then
+         # echo squashes possible multi line replies to a single line
+         FOUND_LINES="$(grep -i "^# $1:" $CONFIG_FILE | sed "s/^# $1: //i")"
+         if [ "$(echo "${FOUND_LINES}" | wc -l)" -ge 1 ]; then
+             echo "${FOUND_LINES}"
+             return 0
+         fi
+    fi
+    if [ "${UNAME}" = "Linux" ] && [ -d '/etc/network/interfaces.d' ]; then
+         if [ "$(/bin/ls /etc/network/interfaces.d/ | wc -l)" -ge 1 ]; then
```
mpikzink, line 148: Better use find instead of ls

VVelox: Both work, pointless change made though. That said the next line did have an issue with grep.

VVelox:
@mpikzink Anything else, I know I will likely regret asking as we have very different ideas on what needs changed? =^.^= :)

murrant:
@VVelox I still can't figure out the purpose of scripts/ifAlias.  Care to explain what it is used for?

VVelox:
@murrant Allows for interface descriptions akin to like on Cisco/Juniper devices. https://docs.librenms.org/Extensions/Interface-Description-Parsing/#setup

So this allows you to add something like 'Transit: ' etc for a interface and it will be picked up.

Jellyfrog:
This script should probably move to the agent repo

VVelox:
> This script should probably move to the agent repo

hmm, true! can make it so! :)

PR #14050, opened 2022-06-20
Merge request for branch tud-zih/22.4; adding/changing features x5

mwobst:
This pull requests contains some new features (or so) as well some adjustments

- For APC UPS devices, the battery operation time is determined and stored (difference between today and last change date in minutes, will be displayed in years and minutes)
- Messages sent via matrix or email may (or rather will) be formatted in html now. Especially in Matrix chatrooms it’s nice to have a red colored “Alert” etc.
- Wrapper script (specifically poller + discovery; only these are covered) did not have an option for passing specific modules (like the php scripts). Now they do.
- CMCIII devices would deploy the used unit for a temperature as a (bad encoded) °C (or °F) and where thus treated as (utterly useless) 'state' sensors. Now temperature parsing works properly
- GUI: the buttons for web/ssh/telnet did only use the hostname, which might be a problem at some environments(read: at ours … thus the change motivation …). Now – also with the help of additional config parameters in config.php – you may use the overwrite ip ('Assigned IP') or even the system name.

Please note:
- I deliberately ignored squashing(?) the commits, as I currently do not know, what the resulting message would be and several of them aren’t super short. So, as it is your repository after all, I will leave the remaining message content up to you.
- Though I changed the behaviour of the ssh/telnet/web buttons, the look wasn’t changed at all (not counting the mouseover text change at device table listing page). Therefore I won’t attach a screenshot.

#### Please note

#### Testers


Jellyfrog:
Sorry you need to split this into separate pull requests

PR #14055, opened 2022-06-21
Add module support for wrapper script calls

mwobst:
The scripts poller.php and discovery.php offer a module
option (-m), which may be used to specify specific modules
for polling/discovery, possibly for special (and thus faster) testing
or for example rediscovering the fdb table (on all hosts).

Until now, this was not possible with the python wrapper scripts.
Now they support a '-m' option, where comma separated module names
may be passed. This will currently only work with poller and discovery, though.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/wrapper.py
+++ b/LibreNMS/wrapper.py
@@ -513,7 +513,7 @@ def wrapper(
                 "log_dir": log_dir,
                 "wrapper_type": wrapper_type,
                 "debug": _debug,
-                "modules": kwargs.get('modules', ""),
+                "modules": kwargs.get("modules", ""),
             },
         )
         worker.setDaemon(True)
--- a/LibreNMS/wrapper.py
+++ b/LibreNMS/wrapper.py
@@ -513,7 +513,7 @@ def wrapper(
                 "log_dir": log_dir,
                 "wrapper_type": wrapper_type,
                 "debug": _debug,
-                "modules": kwargs.get('modules', "")
+                "modules": kwargs.get('modules', ""),
             },
         )
         worker.setDaemon(True)
@@ -683,7 +683,7 @@ if __name__ == "__main__":
     if wrapper_type in ["discovery", "poller"]:
         modules_validated = modules
     else:
-        modules_validated = "" # ignore module parameter
+        modules_validated = ""  # ignore module parameter
 
     wrapper(
         wrapper_type,
@@ -691,5 +691,5 @@ if __name__ == "__main__":
         config,
         log_dir,
         _debug=debug,
-        modules=modules_validated
+        modules=modules_validated,
     )
--- a/LibreNMS/wrapper.py
+++ b/LibreNMS/wrapper.py
@@ -229,7 +229,7 @@ def poll_worker(
     log_dir,  # Type: str
     wrapper_type,  # Type: str
     debug,  # Type: bool
-    modules = '',  # Type: string
+    modules="",  # Type: string
 ):
     """
     This function will fork off single instances of the php process, record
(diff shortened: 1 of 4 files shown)


murrant:
Not 100% sure of the use case here, wouldn't adding `./discover.php -h all -m fdb-tables` to the cron work?

mwobst:
> Not 100% sure of the use case here, wouldn't adding `./discover.php -h all -m fdb-tables` to the cron work?

But that would only start a single thread, which is way too slow when asking around 2000 devices, isn’t it? Please correct me, if I’m wrong, but we want to start a python script for spawning multiple php processes at once, so that a single discovery run for e. g. the fdb-tables module finishes within a reasonable time.

mwobst:
First an explanation:
I believe that the PHP scripts (both polling and discovery) work sequentially when using "-h all" … am I wrong about it? My motivation is parallelized processing of the devices.

Currently, we run the discovery for almost 2000 devices, with modules = arp-table, fdb-table + discovery-protocols, using 20 threads. It still takes around 18 minutes to finish. So a sequential attempt would take 20*18 minutes … okay, probably less, but still a lot of time. (The usual full discovery for newly added devices is still running, of course, in case you’re wondering)

> Depending on how frequently you schedule the fdb discovery, it could end up with multiple running at the same time and should result in all being completed within that frequency depending on how consistent the time to run is.

Well, we also have a systemd unit (instead of using crontab) starting the "big" discovery, so overlapping is prevented that way. (Sadly, there doesn’t seem to be a distributed version of discovery like it is with polling.)

> Say you set fdb discovery to 5 minute intervals and it takes 30 minutes to complete all. This would result in 6 discovery processes running all the time and all devices being completed every 5 minutes.

Yeah, with crontab this would be an issue, but alas, not with systemd (using timers).

> I think the wrapper has a mechanism to prevent overlap, so if the process isn't completed within the interval, it is canceled. That is why I would suggest the php script directly for a rolling poll.

A "rolling poll"? :D

> One difficulty with either approach is overloading the device CPUs. Multiple processes fetching SNMP can bring a device CPU down pretty easily. The easiest to trigger side effect of that is polling backing up and causing gaps in graphs.

With this you’re (sadly) right on track. Two of our pollers having long runtimes every hour (the current systemd unit interval), although often only two or one, and almost always the same ones (we use distributed polling) – there are some devices which are quite … challenged. 

The thing, though, is that we need a regularly refreshed fdb table, and polling doesn’t seem to do this – at least [this page](https://docs.librenms.org/Support/Poller%20Support/#poller-config) doesn’t seem to list an appropiate option.

murrant (dismissed):

murrant:
Systemd won't allow overlapping services via timer, but cron will.

Run 1: 1, 2, 3, 4, 5, 6, 7, 8
Run 2:          1, 2, 3, 4, 5, 6, 7, 8
Run 3:                   1, 2, 3, 4, 5, 6 ,7, 8
Run 4:                            1, 2, 3, 4, 5, 6, 7, 8
Run 5:                                     1, 2, 3, 4, 5, 6, 7, 8

So if you used cron, you would effectively have 4 parallel discoveries running.  Because discovery doesn't update rrd data (thereby having no deadline), this is absolutely no problem.

The following cron entry would discover those three modules every 5 minutes.  The process taking longer than 5 minutes is inconsequential for your particular use case.

*/5  *    * * *   librenms    /opt/librenms/discovery.php -h all -m arp-table,fdb-table,discovery-protocols >> /dev/null 2>&1

A real fix for what you want may be more complex per-module scheduling.  But that is currently a long way off.

PipoCanaja:
Hi @mwobst  and @murrant 
Shouldn't we add a little bit of documentation with this very good feature ?

electrocret (approved):

PR #14057, opened 2022-06-21
WebUI: Improved device links

mwobst:
Both on the device main page and also the table view for all devices, there are buttons/links for accessing the device via ssh/https/telnet. Until now only the hostname was used, but with this feature, usage of the assigend ip address (or the systen name) is possible. Note for device page: The 3 links are located behind the 3 dot panel at the almost-top-right of the device page or at the right of the table view.

Now you can use the config setting device_external_link.use_assigned_ip to ensure,
that your links a generated with the overwrite IP (this one must be set, of course):

    $config['device_external_link']['use_assigned_ip'] = true;

The default is not true.

You may also enable the usage of the system name (sysName),
if this one contains your FQN for whatever reason:

    $config['device_external_link']['use_sysName'] = true;

The default is not true.

Otherwise, the current behaviour is kept – using the hostname for building the links.

Along the way, the title value has also been adjusted slightly (affects
only table view).

Please note: As the changes do not have any visible effect (save for the element title at table view), I refrain from uploading a screenshot.

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/DeviceController.php
+++ b/app/Http/Controllers/DeviceController.php
@@ -150,7 +150,7 @@ class DeviceController extends Controller
         return $graphs;
     }
 
-    private function buildDeviceLinkTargetValue($device)
+    private function buildDeviceLinkTargetValue(Device $device): string
     {
         if (Config::get('device_external_link.use_assigned_ip') && $device->overwrite_ip) {
             $target = $device->overwrite_ip;
--- a/app/Http/Controllers/DeviceController.php
+++ b/app/Http/Controllers/DeviceController.php
@@ -150,7 +150,6 @@ class DeviceController extends Controller
         return $graphs;
     }
 
-
     private function buildDeviceLinkTargetValue($device)
     {
         if (Config::get('device_external_link.use_assigned_ip') && $device->overwrite_ip) {
@@ -160,6 +159,7 @@ class DeviceController extends Controller
         } else { // fallback to hostname
             $target = trim($device->hostname);
         }
+
         return $target;
     }
 
--- a/app/Http/Controllers/DeviceController.php
+++ b/app/Http/Controllers/DeviceController.php
@@ -150,19 +150,6 @@ class DeviceController extends Controller
         return $graphs;
     }
 
-    private function buildDeviceLinkTargetValue(Device $device): string
-    {
-        if (Config::get('device_external_link.use_assigned_ip') && $device->overwrite_ip) {
-            $target = $device->overwrite_ip;
-        } elseif (Config::get('device_external_link.use_sysName') && strlen(trim($device->sysName))) {
-            $target = trim($device->sysName);
-        } else { // fallback to hostname
-            $target = trim($device->hostname);
-        }
-
-        return $target;
-    }
-
     private function deviceLinkMenu(Device $device)
     {
         $device_links = [];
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/DeviceController.php
```diff
@@ -173,19 +173,42 @@ private function deviceLinkMenu(Device $device)
...
+
         // Web
+        $https_target = $device->hostname;
+        if (Config::get('device_external_link.use_assigned_ip') && $device->overwrite_ip) {
+            $https_target = $device->overwrite_ip;
+        } elseif (Config::get('device_external_link.use_sysName') && strlen(trim($device->sysName))) {
+            $https_target = trim($device->sysName);
+        } else { // fallback to hostname
+            $https_target = $hostname;
+        }
```
Jellyfrog, line 187: this code is repeated 6 times below

mwobst: I included a method for creatung the “url target” value (referring to that block) – at both files. That made the code amount smaller.

app/Http/Controllers/DeviceController.php
```diff
@@ -150,6 +150,19 @@ private function buildDeviceGraphArrays($device)
         return $graphs;
     }
 
+    private function buildDeviceLinkTargetValue(Device $device): string
```
PipoCanaja, line 153: Hi
This private function is repeated twice. Probably better to write it once and use it twice instead.

PipoCanaja (changes requested):

app/Http/Controllers/DeviceController.php
mwobst: Yes, I wasn’t exactly happy with that, too, but that method is used across two different classes, both having only Controller as parent class. I don’t think that this class is a fitting place for that rather specific method, isn’t it? Do you have some suggestion where to place it? Sth. like a helper module/class/file?

PipoCanaja: There is indeed such helpers : You can find them in `/opt/librenms/LibreNMS/Util`. And one particular class would be the perfect candidate : 
`/opt/librenms/LibreNMS/Util/Url.php`.

The function itself could then be : 
`public static function directDeviceUrl($device): string

And this function could be called from your code to build the HTTPs, telnet etc etc.  The class is already loaded in the 2 files (`use LibreNMS\Util\Url;`) so you can use any function of this class directly with something like :
`$telnet = 'telnet://' . Url::directDeviceUrl($device);`

mwobst: Thank you very much for your advice; apparently I was not able to detect that "Url" class import. Now, after being pointed out by you, it seems pretty obvious …
Nevertheless: I added a "Part" to the method name, because it is does not include the protocol, so Url alone might be misleading.

mwobst:
The failed tests are strange – the file `tests/data/apc_managementmoduleaosv6.8.8-appv6.8.8.json` does not even exist in this branch (origin would be the pull request #14058, the one about the apc ups battery operation time). Could it be, that this test will now fail on every pull request …? Or have I just met some very curious bug?

It might pose a problem, though:
`tests/data/apc_managementmoduleaosv6.8.8-appv6.8.8.json` contains the date for the last battery change, since the the battery is running. The difference until today is calculated as minutes and saved; that value is present within `snmpsim/apc_managementmoduleaosv6.8.8-appv6.8.8.snmprec`. It seems, though, that the tests calculate the difference anew, which will rise each day, of course …

murrant:
@mwobst not sure I understand this PR.  Hostname/IP is no resolvable?

If you don't have resolvable hostnames.  I would add devices by IP and set the display name.

mwobst:
@murrant Erm … I’m a little stumped, that I never noticed that feature … and yes, that makes the PR practically superfluous. Thanks for the advice – which is helpful and leaves me a little stumped about why I never noticed that feature. The PR is now superfluous … so I will just close it.

Issue #12915, opened 2021-05-25
NS-BSD Plugin (Stormshield) dosn't work since 3.7.19 LTSB


ilymare:
Hi,

Since the LTSB version of the stromshield OS in 3.7.19 (and also in last version in 3.7.20) the stormshield firewall are seen as a generic device.

The objectID in the MIB as change from ".1.3.6.1.4.1.8072.3.2.8" to ".1.3.6.1.4.1.11256.2.0" in the file "librenms/includes/definitions/ns-bsd.yaml".

After running the script ./discovery.php -h <Stromshield_device>, everything is back to normal.

Regards,

Q.

PipoCanaja:
> Hi,
> 
> Since the LTSB version of the stromshield OS in 3.7.19 (and also in last version in 3.7.20) the stormshield firewall are seen as a generic device.
> 
> The objectID in the MIB as change from ".1.3.6.1.4.1.8072.3.2.8" to ".1.3.6.1.4.1.11256.2.0" in the file "librenms/includes/definitions/ns-bsd.yaml".
> 
> After running the script ./discovery.php -h <Stromshield_device>, everything is back to normal.
> 
> Regards,
> 
> Q.

Hello @ilymare 
As it seems you found out the solution, could you please open a pull request directly with the added objectID (please keep the original as well so you don't break older devices) ? You would also need to provide test data (to allow regression tests to take place). 
Thanx in advance

ilymare:
Hi,

I already had this issue last time i post on the git but every pull request that I make are incomplete because I cannot extract data form my compagny devices. 

I post the info here so if someone needs it or/and can complete a pull request for the community.

Sorry for the inconvenience.

Ilymare.

PipoCanaja:
Please at least do the PR for the YAML file.

PR #14060, opened 2022-06-22
ns-bsd updated for SNS LTSB 3.7.19

Mar974:
Since the LTSB version of the stromshield OS in 3.7.19 (and also in last version in 3.7.20) the stormshield firewall are seen as a generic device.

The objectID in the MIB as change from ".1.3.6.1.4.1.8072.3.2.8" to ".1.3.6.1.4.1.11256.2.0" 

see https://github.com/librenms/librenms/issues/12915

Tested on SNS 3.11.10

#### Please note

#### Testers


Jellyfrog:
Now it won't work for older firmwares, you need to add both of them.
Could you also post the sysdesc of the device?

Mar974:
OK i added a table so the old and the new one will match.
The sysDescr contain the serial number, exemple NS-BSD SN510XXXXXXXX amd64 so we need to keep the regex

murrant:
Need to add a little test data.
I do think the multi-line array that was already there was a little more readable, just need to add another line `- .1.3.6.1.4.1.11256.2.0`

Mar974:
Ok i Changed the array to multiligne and collected some test data, where should i upload the test data ?

murrant (dismissed):

murrant (approved):

PR #14071, opened 2022-06-26
Rewrite service checks

murrant:
Rewrite services to improve security
Adds a services polling module, but still polls through the separate check-services.php or lnms services:poll scripts by default.
Attempts to parse help text into valid options.
Default values for options such as -H.

TODO:

- [ ] Improve help parsing and tests
- [x] Test polling
- [ ] Test all the webui
- [x] Convert all old check scripts.

#### Please note

#### Testers


moschroe:
As suggested on discord, here my proposed extensions for service plugins:

1. It ought to be possible to keep the original name of perf metrics for the graph and use the canonicalised form only for interfacing with RRD, allowing spaces and punctuation in the legend.
2. Also, there should be an option available to pass SNMP credentials (+ protocol, sec level, etc.) via envvar to the service script. This would allow using existing SNMP facilities (extend scripts, forwarded ports, firewall rules) while not increasing exposed surface.

I guess the second point would be more straightforward to implement. I might be able to contribute there, let me know whether that would be desirable or I should open a separate PR (probably waiting on this one then)

murrant:
@moschroe Thanks for the input.  If you would like to help out with this, you can send a pull request directly to my branch. https://github.com/murrant/librenms/tree/rewrite_services

Also, creating services is quite broken right now in this PR.

murrant:
Needs rework due to not including arguments, only options.

murrant:
I think if I had to restart this, I would create a list of arguments/options for each check with the option for the user to add custom ones in-case the interface changes.

PR #14073, opened 2022-06-27
Improved sorting options for Availability Map

mwobst:
As we do not have resolvable hostnames (yet), we want to enter the IP address as hostname while using the actual hostname as display name. The Availability Map does support displaying the Display Name (now), but it does not support sorting by it, yet.

Update 2022-07-11:

The dropdown now presents two options:

- Label: Sort by the selected value of the drop down 'Label'
- Status: Sort by status, then by selected value of drop down 'Label'

As the field 'label' (The display name) may contain template functions etc., sorting is not done by SQL means; instead custom sorting is done within the controller.

The sorting menu would look like this: Picture

#### Please note

#### Testers

### DIFF
--- a/resources/lang/de.json
+++ b/resources/lang/de.json
@@ -363,7 +363,7 @@
     "Groups": "Gruppen",
     "Performance": "Performance",
     "History": "Historie",
-    "Device Label": "Anzeigewert"
+    "Display Name": "Anzeigename",
     "Status, Hostname": "Status, Hostname",
     "Status, Display Name": "Status, Anzeigename"
 }


murrant:
The thing about sorting by the display name is, that it is a template.  This means it cannot be sorted in SQL.
You would have to load them, parse all the templates, then sort them.

Also, if you have translation changes, you should submit them in a separate PR :)

mwobst:
> The thing about sorting by the display name is, that it is a template. This means it cannot be sorted in SQL. You would have to load them, parse all the templates, then sort them.

I am confused … I’m using the column 'display' of the table 'devices'. It is perfectly sortable and can also be filled by using the 'Display Name' textbox when editing a device – it is even usable by API.

> 
> It seems like it should always be sorted by the selected display text. Sort should just be display text and status. (secondary will be the display text)

What about the hostname field? If all your hosts are resolvable by name, you will want to sort by hostname rather than 'display'. Or maybe some don’t use resolvable names, use IP addresses in that field and rely on sorting them. In our case we’re just starting to use the display name and sort by hostname until now.

> 
> Also, if you have translation changes, you should submit them in a separate PR :)

Ok.

murrant:
@mwobst This is what is stored in the database for the display field `{{ $hostname }}`  also, by default it is empty to fall back to the default setting: `device_display_default`  In most cases, the field display will be empty.

mwobst:
> @mwobst This is typically what is stored in the database for the display field: `{{ $hostname }}`. Also, by default it is empty to fall back to the default setting: `device_display_default` In most cases, the field display will be empty. Yes, people can place explicit text directly in the display field, but that will not be common.

Well, that makes it a little bit difficult. Suggestion: The sorting options for Display Name will contain "Static Display Name" (or "Custom Display Name" or "Fixed Display Name" … or maybe even "Display Name Overwrite"?) instead of only "Display Name". It should also be possible to add a tool tip ("Sorting by the […] Display Name will sort by the given value, but will not consider template values like {{ hostname }}"), shouldn’t it.

It should also possible to add another column when sorting by Display Name – hostname would probably be the best idea – "order by [status, ]display, hostname" in SQL.

While I’m at it: The tool tip for the Device Name contains "Keep short." which is … vague. "At most 128 signs" is probably more informative.

murrant:
The display name tip is probably vague because it affects various layouts, not because there is a specific length limit of 512 or something.

As I said before we want to present as few options to the user as possible.  When you see something labeled by a specific field it should be sorted by that label.  So we really only need two sorting options:  Sort by label, sort by status.

One solution is to sort the data after it has been loaded from the database (in php). Because this data isn't paginated, there should be no downside to doing so.

mwobst:
Had to remove the conflict merge, as it pulled the whole master into the branch – which I don’t want right now as unknown/unlogged errors with map occurred.

Concerning the conflict: It removes two lines that were only "required" for the expanded dropdown menu within this PR. As I condensed the menu, those two lines are now obsolete and can safely be removed.

resources/views/widgets/settings/availability-map.blade.php
```diff
@@ -20,7 +20,7 @@
             <option value="1" @if($color_only_select == 1) selected @endif>{{ __('empty') }}</option>
             <option value="4" @if($color_only_select == 4) selected @endunless>{{ __('Display Name') }}</option>
             <option value="2" @if($color_only_select == 2) selected @endunless>{{ __('Hostname') }}</option>
-            <option value="3" @if($color_only_select == 3) selected @endunless>{{ __('Sysname') }}</option>
+            <option value="3" @if($color_only_select == 3) selected @endunless>{{ __('System Name') }}</option>
```
Jellyfrog, line 23: if you change a name, please update all lang files

mwobst: I can’t find any entry for Sysname, so there should be no need for updating any key.

mwobst: a) I should be more specific: There is no trace of "Sysname" in any lang file (afaik), so a name change should be no problem.

b) Are that tests really stuck …? It is like this for a month now.

SourceDoctor:
This PR breaks ordering in Availability Map.
Devices are no more ordered by it's Hostname by default.

I having Label = Device Status, Devices are no more sorted by device name.

Thoughts to solve this issue, and enhance ordering options:

"Ordering" should also have all meaningful options which "label" has and "by label" Option should be removed.
With this Device Ordering can be specified, directly. "Display Name" should be default.

Ordering by label, without a default sort on top like "hostname" makes no sense if label is set to something like status, which is default.

murrant:
@SourceDoctor sort by status

murrant:
@SourceDoctor Probably need to add new options.  I think status is the only one that doesn't make sense.

SourceDoctor:
@murrant 

sort by status does not change anything.

before status was label an order was alphabetic by display name
That's what we need.

So my i think order options should have  same options as label.
With this you can label it as needed, and independent from it sort them as needed

murrant:
I don't think the showing by one form of "hostname" and sorting by another makes sense...

mwobst:
@SourceDoctor I interpret this as "Our availability map shows tons of "up" and "down", but it is important to sort by device name (=hostname?)".

Originally I wanted separate dropdowns too, but @murrant was right in my opinion: Sorting and displaying should usually go hand in hand – your version strikes me as odd … but maybe there is a reason I cannot fathom yet (?)

SourceDoctor:
@mwobst 
that is the case.
alphabetic sorting of Device by hostname is essential on big infratstructures. Independent of it's label.

So in my opinion, order options should be independend from label option, to solve/recover this issue.

murrant:
#14329 I added "display name" as the default sort.  Nothing else makes sense imo.

SourceDoctor:
Thats a partial Screenshot of our availability map to visualize the issue:

Bildschirmfoto vom 2022-09-09 17-12-33

murrant:
@SourceDoctor I already fixed that use case, see #14329

SourceDoctor:
> @SourceDoctor I already fixed that use case, see #14329

ah nice. Thanks i will review it :)

murrant:
All it needed was when status is set, fallback to display name.

murrant:
Additionally, the code should be a lot simpler (not as simple as it could be) to read now.

PR #13954, opened 2022-04-27
add Sneck support, making it very trivial to use Nagios style checks in a NRPE like manner

VVelox:
Sneck is basically NRPE over SNMP.

This means one can easily throw what ever checks on the VM/server/etc when provisioning and poll it via SNMP and get everything there with out having to know what has been setup to poll it.

It tracks totals for each server.

Any found alerts are reported on the application page and stay there until told to clear it.

Screenshots...

[graphs, no alert](https://i.imgur.com/pgwLzbr.png)
[graphs, alert](https://i.imgur.com/FmhQs6Y.png)
[alert via example](https://i.imgur.com/izbuALg.png)

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [X] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [X] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.
- [X] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #14087, opened 2022-07-03
add the ability for storing app data to prevent spamming of the event log via via component usage

VVelox:
Adds a new column to the applications table and two new functions for storing/fetching the data.

ZFS, Suricata, Chronyd, Portactivity, and Fail2ban will no longer spam the event log.  Logged entries for them will now be meaningful, documented added/removed instances/ports/pools or the like.

#### Please note

#### Testers


VVelox:
Updated Suricata as well while I am here to use this.

VVelox:
Tests working and add some documentation on writing app stuff.

VVelox:
Woot! Everything updated and all tests passed.

includes/polling/applications/chronyd.inc.php
```diff
@@ -95,43 +101,50 @@
...
+        }
+    }
+    if (! $source_found) {
+        $removed_sources[] = $source_check;
+    }
+}
 
-    $component->setComponentPrefs($device_id, $chrony_cpnt);
+// if we have any source changes, log it
+if (isset($added_sources[0]) or isset($removed_sources[0])) {
```
bennet-esyoil, line 139:
```suggestion
if (count($added_sources) > 0 || count($removed_sources) > 0) {
```
[sizeof](https://www.php.net/manual/en/function.count.php)

includes/polling/applications/chronyd.inc.php
```diff
@@ -95,43 +101,50 @@
...
+    $log_message = 'Chronyd Source Change:';
+    if (isset($added_sources[0])) {
+        $log_message = $log_message . ' Added' . json_encode($added_sources);
+    }
+    if (isset($removed_sources[0])) {
+        $log_message = $log_message . ' Removed' . json_encode($removed_sources);
+    }
```
bennet-esyoil, line 146: When one is added and one is removed this would only catch the removed one, maybe something like this:
```
$log_message = 'Chronyd Source Change: ';
$log_message .= count($added_sources) > 0 ? ' Added ' . json_encode($added_sources) : '';
$log_message .= count($removed_sources) > 0 ? ' Removed ' . json_encode($added_sources) : '';

```

also: spaces between data and strings

VVelox: Used, thanks!

includes/polling/applications/chronyd.inc.php
VVelox: I think this is just really a preference, unless one is actually faster or the like as they both do the exact same thing, given the first item in the array will be set if there is something there.

VVelox: Converted! Thanks!

VVelox:
https://github.com/librenms/librenms/pull/13954 now uses this as well.

includes/functions.php
```diff
@@ -1488,3 +1491,51 @@ function describe_bgp_error_code($code, $subcode)
...
+
+/**
+ * Gets stored app data JSON and returns it.
+ * If none is present or it is bad, a blank array is returned.
+ *
+ * @params int $app_id
+ *
+ * @return array
+ */
+function get_app_data($app_id)
```
murrant, line 1503: Question why a two whole functions instead of just using
`Application::find($app_id)->data`

That will return a decoded array (and also encode the array when saved)  if  you set `protected $casts = ['data' => 'array'];` on the Application model.

This also has the added benefit of caching the data if you pass around the loaded Application object.

VVelox: Three whole functions, I am a bit lost, I seem to be missing the third.

As to why, it seemed it would be easier than calling something like this each time.

```
$app = Application::where(['app_id' => $app_id])->first();
$parsed_json = json_decode($app->data, true);
if (json_last_error() !== JSON_ERROR_NONE) {
    //whatever
}
```

That said if you know how to get `app/Models/Application.php` to play nice with a helper function or the like for this, that would also be cool and I am totally open to any input on actually getting that to work.

VVelox: So a fresh set of eyes on it and got a save helper function working...

```
   public function save_data($data = [])
   {
      $this->fill(['data'=>json_encode($data)]);
      $this->save();
   }
```

Aye. Going to be updating this and ditching the functions in `include/functions.php`.

VVelox: Woot! All replaced with helper functions in the model now every where. Docs updated as well.

doc/Developing/Application-Notes.md
```diff
@@ -0,0 +1,90 @@
...
+
+The data will be fetched, decoded as JSON and returned as a array. The
+required variables are as below via the helper function `get_data`.
+
+An example from `includes/html/pages/device/apps/zfs.inc.php`...
+
+```php
+use App\Models\Application;
+
+$app=Application::where(['app_id' => $app_id])->first();
```
murrant, line 88: `Application::find($app_id])` is a lot nicer way to express this.

doc/Developing/Application-Notes.md
```diff
@@ -0,0 +1,90 @@
...
+$app_data=$app->get_data();
```
murrant, line 89: `$app_data = $app->data;`  is also a simpler way to fetch the data.

doc/Developing/Application-Notes.md
```diff
@@ -0,0 +1,90 @@
...
+table using the helper function `save_data`.
+
+- Array :: The data to convert to JSON for storage.
+
+Example from `includes/polling/applications/zfs.inc.php`...
+
+```php
+use App\Models\Application;
+
+$app=Application::where(['app_id' => $app_id])->first();
```
murrant, line 73: $app = Application::find($app_id);

doc/Developing/Application-Notes.md
```diff
@@ -0,0 +1,90 @@
...
+
+$app->save_data($app_data);
```
murrant, line 75:
```php
$app->data = $app_data;
$app->save();
```

doc/Developing/Application-Notes.md
VVelox: Not sure about that, given parsing it into JSON is then needed.

VVelox: Just tested this and yup, just returns a string.

doc/Developing/Application-Notes.md
VVelox: Awesome, thanks!

doc/Developing/Application-Notes.md
VVelox: $app->data is a JSON string, so unless the data is just a string, likely not that useful.

VVelox: That said, documented working with it as both JSON and non-JSON now incase some one has a use case for something else.

doc/Developing/Application-Notes.md
VVelox: That said, documented working with it as both JSON and non-JSON now incase some one has a use case for something else.

doc/Developing/Application-Notes.md
```diff
@@ -0,0 +1,84 @@
+# Notes On Application Development
+
+## LibreNMS JSON SNMP Extends
+
+The polling funtion `json_app_get` makes it easy to poll complex data
```
murrant, line 5: funtion -> function

doc/Developing/Application-Notes.md
```diff
@@ -0,0 +1,84 @@
...
+using SNMP extends and JSON.
+
+The following exceptions are provided by it.
+
+It takes three paramters, in order in the list below.
```
murrant, line 10: paramters -> parameters

doc/Developing/Application-Notes.md
```diff
@@ -0,0 +1,84 @@
...
+use App\Models\Application;
+
+// find the app in question via it's ID integer
+$app = Application::find($app_id);
+
+// set the varaible data to $foo
+$app->data = $foo;
+
+// save the change
+$app->save
```
murrant, line 74: This needs to be `$app->save();`

doc/Developing/Application-Notes.md
```diff
@@ -0,0 +1,84 @@
...
+model.
+
+```
+use App\Models\Application;
+
+// find the app in question via it's ID integer
+$app = Application::find($app_id);
+
+// set the varaible data to $foo
+$app->data = $foo;
```
murrant, line 71: Should probably show an array structure as an example.
```php
$app->data = [
    'stat_a' => 432,
    'stat_b' => 123.43,
];
```

VVelox:
Awesome, thanks!

includes/polling/applications.inc.php
```diff
@@ -1,13 +1,22 @@
...
+use App\Models\Application;
+
 $app_rows = dbFetchRows('SELECT * FROM `applications` WHERE `device_id`  = ?', [$device['device_id']]);
 
 if (count($app_rows) > 0) {
     foreach ($app_rows as $app) {
         echo 'Application: ' . $app['app_type'] . ', app_id=' . $app['app_id'];
         $app_include = \LibreNMS\Config::get('install_dir') . '/includes/polling/applications/' . $app['app_type'] . '.inc.php';
         if (is_file($app_include)) {
+            $app_model = Application::find($app['app_id']);
```
murrant, line 12: Remember that Application::find() is actually a database query.  So you should fetch that outside the loop.  And try to only call it once in general.

includes/polling/applications.inc.php
```diff
@@ -1,13 +1,22 @@
 <?php
 
+use App\Models\Application;
+
 $app_rows = dbFetchRows('SELECT * FROM `applications` WHERE `device_id`  = ?', [$device['device_id']]);
```
murrant, line 5: This query should be 
`$app_rows = \DeviceCache::primary()->applications;`

or more succinctly:
```php
\DeviceCache::primary()->applications->each(function ($app) {
        echo 'Application: ' . $app->app_type. ', app_id=' . $app->app_id;
        $app_include = base_path('includes/polling/applications/' . \LibreNMS\Util\Clean::fileName($app->app_type) . '.inc.php');
        if (is_file($app_include)) {
            include $app_include;
        } else {
            echo 'ERROR: ' . $app_include . ' include file missing!';
        }
});
```

Models implement array access so array style access will work (although it is not preferred)

murrant: With this change, you never need to call `Application::find($app['app_id'])`

Just access $app->data directly

murrant:
These locations should be updated to fetch the model:
`includes/html/graphs/application/auth.inc.php`
`includes/html/pages/device/apps.inc.php`

Then you can remove the duplicate query further down in the code.

VVelox:
Can definitely help with that, another time.

This one is already massive enough that that is going to involve re-writing even more.

murrant (approved):

PR #14102, opened 2022-07-10
[gui] enable permanent vertical scroll

Npeca75:
on some pages when mini graph is shown under mouse pointer and graph is long enough to trigger vertical scroll, then whole page is shifted in left direction because of vscroll on right edge.
This is enough to mouse pointer leave the object, and mini graph & vscroll will disappear
Then mouse pointer will be again on object, and mini graph & vscroll will be shown again
This loop could be deadly for chrome browser, causing hard lockup of web browser
so enabling permanently vertical scroll solve this sitation

#### Please note

#### Testers

### DIFF
--- a/html/css/blue.css
+++ b/html/css/blue.css
@@ -1,3 +1,7 @@
+body {
+    overflow-y: scroll;
+}
+
 .navbar-default {
   background-color: #d6e7ff;
   border-color: #0052cc;
--- a/html/css/blue.css
+++ b/html/css/blue.css
@@ -1,7 +1,3 @@
-body {
-    overflow-y: scroll;
-}
-
 .navbar-default {
   background-color: #d6e7ff;
   border-color: #0052cc;
--- a/html/css/dark.css
+++ b/html/css/dark.css
@@ -1,3 +1,7 @@
+  body {
+    overflow-y: scroll;
+  }
+
   .fa-nav-icons {
     color: #bfc0c0;
   }
--- a/html/css/dark.css
+++ b/html/css/dark.css
@@ -1,7 +1,3 @@
-  body {
-    overflow-y: scroll;
-  }
-
   .fa-nav-icons {
     color: #bfc0c0;
   }
--- a/html/css/light.css
+++ b/html/css/light.css
@@ -1,3 +1,7 @@
+body {
+    overflow-y: scroll;
+}
+
 .fa-nav-icons {
     color: #383637;
 }
--- a/html/css/light.css
+++ b/html/css/light.css
@@ -1,7 +1,3 @@
-body {
-    overflow-y: scroll;
-}
-
 .fa-nav-icons {
     color: #383637;
 }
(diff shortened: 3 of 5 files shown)


Npeca75:
lnms1

murrant:
I'm dubious about this fix.

Is this with the css popup or the old overlib popup?  If it is overlib, we could probably fix it by migrating to css.

Npeca75:
> I'm dubious about this fix.
> 
> Is this with the css popup or the old overlib popup? If it is overlib, we could probably fix it by migrating to css.

https://user-images.githubusercontent.com/59750439/178447897-004eff8b-0d32-4947-b512-6bc2ac3024c7.mp4

Ok @murrant 
maybe the video will explain better

these problems are gone with vscroll enabled

murrant:
I am aware of the problem.  You can actually get it stuck in a loop if conditions are just right.

What I mean is I think this will add scroll bars in a lot of unintended places.

Check the popups on /devices  these are the new ones.

murrant:
@Npeca75 if you can find a more localized fix or somehow prove this isn't breaking other things, I'm all for it.

Otherwise, I think we just need to get things ported away from overlib.

Npeca75:
> @Npeca75 if you can find a more localized fix or somehow prove this isn't breaking other things, I'm all for it.
> 
> Otherwise, I think we just need to get things ported away from overlib.

Hi @murrant 

believe or not, i am using LNMS every day :) 
but just to be shure , started this day with opening every single window in program
overviews, all OK 
vlan + whereisvlan plugins OK
devices , all OK
services, all OK
ports, all OK
health, all OK
wireless, all OK
apps, all OK
alerts, all OK
my setting / global settings / poller / api, all OK
widgets which i use, worldmap, top devices, top interfaces, alerts, all OK
on individual device page, again, clicked on all menu points

tested with Chrome & Firefox

sorry but i don't know what other prove i could offer :(

murrant:
Testing a lot of pages does make me feel better about it.

However, this is in the wrong place, you should put it in the librenms.css file, since this isn't really theme related.

Npeca75:
hi @murrant 
> However, this is in the wrong place, you should put it in the librenms.css file, since this isn't really theme related.

Yes, i agree. When i started to look for solution, i was focused only "mono" thene which i use, and prom there... copy->paste :)

>put it in the librenms.css file

sorry, i analyzed page source and librenms.css is not loaded. maybe i am missing something ?

`169.254.102.252/css/bootstrap.min.css" rel="stylesheet">`
`169.254.102.252/css/styles.css?ver=20210421" rel="stylesheet">`
`169.254.102.252/css/blue.css?ver=632417643" rel="stylesheet" >`

bootstrap.min.css is vendor file, it is untouchable...
but the "styles.css" is common for all pages :)

tested again with 4 themes, Chrome & Firefox ...  i don't see any problem for now

murrant (approved):
I probably mentioned the wrong file :(  Thanks!

PR #14105, opened 2022-07-13
Fix Cisco polling BGP peers in non-default VRF

### LINKED ISSUE #14104: Polling BGP peers fails for Cisco devices with peers in VRF, but no peers in default VRF
### The problem

If a Cisco device has VRFs defined and BGP peers configured in those VRFs, but no BGP sessions/peers in the default VRF, a SNMP (v3) walk without context will not find any cbgpPeer2RemoteAs.
The code at the top of `includes/polling/bgp-peers.inc.php` populating `$peer_data_check` will return an empty Array:
```
$peer_data_check = snmpwalk_cache_oid($device, 'cbgpPeer2RemoteAs', [], 'CISCO-BGP4-MIB');
```
This causes the code further below in the same file to skip reading the BGP peer data for the peers in the VRFs.

We've observed the issue on at least these devices and software:
- Cisco Nexus 9000 series, running NX-OS 9.3.8 and 9.3.9
- Cisco ASR 9901 running IOS-XR 7.3.2

... but the issue will be present on all Cisco devices with BGP peers in VRFs, and no BGP peers in the default VRF.

### Output of ./validate.php

```txt
/opt/librenms/validate.php
Component | Version
--------- | -------
LibreNMS  | 22.5.0-15-gfaaca7e6f
DB Schema | 2022_05_25_090027_drop_widgets_table (241)
PHP       | 8.1.8
Python    | 3.9.2
MySQL     | 10.5.15-MariaDB-0+deb11u1
RRDTool   | 1.7.2
SNMP      | 5.9

[OK]    Composer Version: 2.3.5
        [FIX]: 
        Modified Files:
         includes/polling/bgp-peers.inc.php
```

_No response_

_No response_

ajsiersema:
(comments and commented out code left on purpose)

If a Cisco device has VRFs defined and BGP peers configured in those VRFs, but no BGP sessions/peers in the default VRF, a SNMP (v3) walk without context will not find any cbgpPeer2RemoteAs in the code at the top of `includes/polling/bgp-peers.inc.php`. The line for os_group cisco populating `$peer_data_check` will return an empty Array:
```
$peer_data_check = snmpwalk_cache_oid($device, 'cbgpPeer2RemoteAs', [], 'CISCO-BGP4-MIB');
```
This causes the code further below in the same file to skip reading the BGP peer data for the peers in the VRFs.

We've observed the issue on at least these devices and software:
- Cisco Nexus 9000 series, running NX-OS 9.3.8 and 9.3.9
- Cisco ASR 9901 running IOS-XR 7.3.2

... but the issue will be present on all Cisco devices with BGP peers in VRFs, and no BGP peers in the default VRF.

NB: although the code changes the polling, it doesn't add or change the SNMP queries or parsing thereof. Changing to tests/data/pfsense_frr-bgp.json shouldn't be necessary.

fixes  #14104

#### Please note

#### Testers

### DIFF
--- a/includes/polling/bgp-peers.inc.php
+++ b/includes/polling/bgp-peers.inc.php
@@ -48,7 +48,13 @@ if (\LibreNMS\Config::get('enable_bgp')) {
                 echo "Checking BGP peer $peer_ip ";
 
                 // --- Collect BGP data ---
-                if (count($peer_data_check) > 0) {
+                // If a Cisco device has BGP peers in VRF(s),
+                // but no BGP peers in the default VRF,
+                // a SNMP (v3) walk without context will not find any
+                // cbgpPeer2RemoteAs, resulting in empty $peer_data_check.
+                // Without the or clause, we won't see the VRF BGP peers.
+                // ($peer_data_check isn't used in the Cisco code path,)
+                if (count($peer_data_check) > 0 || ($device['os_group'] == 'cisco' && count(DeviceCache::getPrimary()->getVrfContexts()) > 1)) {
                     if ($generic) {
                         echo "\nfallback to default mib";
 


ajsiersema:
I've fixed the polling with the code in this PR, but would like some advice if this is acceptable to proceed with as a pull request.
Although the $peer_data_check variable is not used by the Cisco code path, it feels a bit hacky. Other solutions probably touch more code lines too.

ajsiersema:
not quite sure how to fix the bgp tests

murrant:
@ajsiersema if you scroll up a bit, you will see a stactrace and and error `SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect integer value: '' for column `  which means you are trying to set an empty string to a datetime column.

includes/polling/bgp-peers.inc.php
```diff
@@ -26,6 +26,28 @@
             $peer_data_check = snmpwalk_cache_multi_oid($device, 'hwBgpPeerEntry', [], 'HUAWEI-BGP-VPN-MIB', 'huawei');
         } elseif ($device['os_group'] == 'cisco') {
             $peer_data_check = snmpwalk_cache_oid($device, 'cbgpPeer2RemoteAs', [], 'CISCO-BGP4-MIB');
```
murrant, line 28: SnmpQuery::context()->walk() handles context a bit better.  And is open for improvements.  Take a look at using that instead of snmpwalk_cache_oid()

ajsiersema: Hey Tony, thanks for the help so far.
The issue we face is that we're missing info for BGP peers in VRF's for Cisco devices. This call tries to fetch BGP peers in the default VRF, which will rreturn empty. The `if (count($peer_data_check) > 0) {` on line 73 will cause the subsequent code to be skipped, although the Cisco code path doesn't even use $peer_data_check.
To fix this for (only) CIsco devices, we're forcing $peer_data_check to be non-empty. To pass the automated tests I do however need to execute this first ("useless" snmpwalk.

I.e. our current issue isn't snmpwalk_cache_oid vs context()->walk() and more importantly, I'm not quite sure why this code was added in the first place. I tried to figure this out from the git history. It seems to have been added for a specific set of devices. But not sure why it is now intertwined in the code and checked for all devices, even if the code path for those devices doesn't need it. Or in this case even worse: if it causes the polling to fail for those devices.
As such I'm a bit reluctant to tweak the polling code for functionality and code paths I can't test and for which I'm not 100% sure what it's for.

murrant: bgp code is kind of a nightmare and inter-woven.  If you require the walk for tests to pass, then the data is stored in the db, so it must not be useless.

ajsiersema: Well useless... it's doesn't seem to be checked by the Cisco code path. The automated tests however do require the data. 
I'm trying to fix the BGP peers in VRF issue while stickting to the mantra "don't break the poller" :-) 
Forcing $peer_data_check to not empty (with dummy data) isn't pretty, but changes the least amount of code to satisfy the check at line 78.

murrant: But at least it is obvious what  is going on.

If the bgp code ever does get re-written it will be like all other new modules where per-os code lives with the OS.

ajsiersema: added or clause for cisco devices, used the group for a better link to the code at the top

includes/polling/bgp-peers.inc.php
```diff
@@ -48,7 +48,13 @@
...
 
                 // --- Collect BGP data ---
-                if (count($peer_data_check) > 0) {
+                // If a Cisco device has BGP peers in VRF(s),
+                // but no BGP peers in the default VRF,
+                // a SNMP (v3) walk without context will not find any
+                // cbgpPeer2RemoteAs, resulting in empty $peer_data_check.
+                // Without the or clause, we won't see the VRF BGP peers.
+                // ($peer_data_check isn't used in the Cisco code path,)
+                if (count($peer_data_check) > 0 || device['os_group'] == 'cisco') {
```
murrant, line 57:
```suggestion
                if (count($peer_data_check) > 0 || ($device['os_group'] == 'cisco' && count(DeviceCache::getPrimary()->getVrfContexts()) > 1)) {
```

Perhaps count the vrfs too

murrant (approved):

PR #14109, opened 2022-07-14
Test PHP 8.1

murrant:
Would love to remove the PHP7 versions, as they will both be no longer supported upstream soon, but we need to update our minimum version first.

#### Please note

#### Testers

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,12 +26,17 @@ jobs:
       matrix:
         include:
           -
+            php-version: 8.1
+            name: Unit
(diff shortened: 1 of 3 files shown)


Jellyfrog:
Hmm it really shouldn't create so many connections, I except... 2 ish, default max connections is 151 so something seems off here. And it fails at the first device test

murrant:
@Jellyfrog yeah, the too many connections is really throwing me off.  Has to be something not obvious going on here.

murrant:
Red

Jellyfrog (approved):

PR #14114, opened 2022-07-15
Allow Custom OIDs to interpret hex output

peelman:
When a custom OID output is in hexadecimal form, the existing implementation will chop and trim, and usually return the first two characters.  Using the existing `isHexString` and `snmp_hexstring` methods we can check for, and convert this data to accommodate manufacturers who are not playing completely by the rules.

It could be argued that the `is_numeric` check can be broken out, and both string and numeric results could be in, with only numerics getting the `enableGraph`treatment.  I can submit an additional commit for that if there's a desire.

#### Please note

#### Testers

### DIFF
--- a/includes/polling/customoid.inc.php
+++ b/includes/polling/customoid.inc.php
@@ -18,6 +18,9 @@ foreach (dbFetchRows('SELECT * FROM `customoids` WHERE `customoid_passed` = 1 AN
     if (is_numeric($rawdata)) {
         $os->enableGraph('customoid');
         $oid_value = $rawdata;
+    } elseif (isHexString($rawdata) && is_numeric(string_to_float(snmp_hexstring($rawdata)))) {
+        $os->enableGraph('customoid');
+       $oid_value = string_to_float(snmp_hexstring($rawdata));
     } elseif (
         $customoid['customoid_unit'] &&
         str_i_contains($rawdata, $customoid['customoid_unit']) &&


murrant:
Isn't this unneeded if you use MIBs?

peelman:
i don't think so?

in this case i'm monitoring SolidOptics EDFAMUXes, and the MIB is so poorly done i didn't even want to send a PR through for it and add them as a device/vendor. i have some feedback sent off to them, but in the interest of monitoring the three values i care about (amp temp, input power, and output power), it was easiest to just add custom OIDs. except they return numeric values in hex strings instead of integers, and the previous code would just chop the first two characters and return that as the value. 

since they are generic devices i couldn't even generate proper test data (at least not without some guidance from the peanut gallery).

peelman:
here is an example of the output from the customoid poller:

```
array (
  'customoid_id' => 4,
  'device_id' => 394,
  'customoid_descr' => 'Post-Amp Power In',
  'customoid_deleted' => 0,
  'customoid_current' => -21.7,
  'customoid_prev' => -21.7,
  'customoid_oid' => '.1.3.6.1.4.1.51628.1.3.3.0',
  'customoid_datatype' => '',
  'customoid_unit' => 'dBm',
  'customoid_divisor' => 1,
  'customoid_multiplier' => 1,
  'customoid_limit' => NULL,
  'customoid_limit_warn' => NULL,
  'customoid_limit_low' => NULL,
  'customoid_limit_low_warn' => NULL,
  'customoid_alert' => 0,
  'customoid_passed' => 1,
  'lastupdate' => '2022-07-16 08:55:56',
  'user_func' => NULL,
)  

SNMP['/usr/bin/snmpget' '-v2c' '-c' 'COMMUNITY' '-Oqv' '-M' '/opt/librenms/mibs' '-r' '15' 'udp:HOSTNAME:161' '.1.3.6.1.4.1.51628.1.3.3.0']

"2D 32 31 2E 37 30 00 "  

  

Custom OID Post-Amp Power In: -21.7 dBm

SQL[UPDATE `customoids` set `customoid_current`=?,`lastupdate`=NOW(),`customoid_prev`=? WHERE `customoid_id` = ? [-21.7,-21.7,4] 2.51ms] 
```

this is with my patch in place. if you run `2D 32 31 2E 37 30 00` through the existing pipe, you'll see that it will just return the integer of 2D and be done.  in the case of temp, it would return 25.00 as `32 35 2E 30 30 00` and report the temp as 32°

murrant:
MIBs should convert the hex for you with a display hint.  If they aren't using MIBs, it would be idiotic to return raw hex.

murrant:
Amazing, the device is returning the data at the wrong OID.
According to the MIB, postPowerGainValue should be `.1.3.6.1.4.1.51628.1.3.1.2.0` instead of `.1.3.6.1.4.1.51628.1.3.2.0` Same for all other table entries.  Also, they don't use DisplayString for the display type instead of OCTET STRING.

Can you control the snmp flags?  Just add `-Oa` to convert any OCTET STRING values to ascii.

peelman:
I mean, I can do flags in testing from a command line; but i can't specify them to be used on the Custom OID config in libre...

murrant:
Can we just add as a regular sensor instead?

After changing the oid address in snmpsim and changing OCTET STRING to DisplayString in the mib:
$ snmpwalk -v2c -M +./ -m +SO-MUX-MIB -c output 127.1.3.1:1062 .1.3.6.1.4.1.51628
SO-MUX-MIB::dcmDistance.0 = INTEGER: 140
SO-MUX-MIB::preIndex.0 = INTEGER: 1
SO-MUX-MIB::prePowerGainValue.0 = STRING: No EDFA available.
SO-MUX-MIB::prePowerIn.0 = STRING: No EDFA available.
SO-MUX-MIB::prePowerOut.0 = STRING: No EDFA available.
SO-MUX-MIB::preTemperature.0 = STRING: No EDFA available.
SO-MUX-MIB::postIndex.0 = INTEGER: 2
SO-MUX-MIB::postPowerGainValue.0 = STRING: 21.00.
SO-MUX-MIB::postPowerIn.0 = STRING: -21.70.
SO-MUX-MIB::postPowerOut.0 = STRING: 0.40.
SO-MUX-MIB::postPowerOut.0 = No more variables left in this MIB View (It is past the end of the MIB tree)

murrant:
@peelman hmm, looking at this, should it return more than one set of values?

PR #14115, opened 2022-07-18
Device - Basic hatteras dslam integration

Bierchermuesli:
This is a basic integration of Hatteras HN4000 DLSAM. (Now days known as Overture / ADVA. HN4000 is a discontinued product, therefore keeping the name old fashioned). 

This PR is mainly for GUI beauty and propper OS detection to work with oxidized Backup. More sensor/discovery might follow. SVGs are vectorized by my own.

179507799-bb36cf4c-8bbe-4280-b580-9a7d00103276

#### Please note

#### Testers


Bierchermuesli:
snmprec and json testfiles added!

includes/definitions/hatteras.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - sysDescr:
+            - HN4000-I
+            - HN4000e
+bad_if:
+  - notPresent
+  - serConsole
+good_if: 
+  - ras #all interfaces has 'Hatteras' in it...
+discovery_modules:
+    ntp: false
```
PipoCanaja, line 20: Hi @Bierchermuesli 

Any reason to force modules being activated and deactivated ? If they don't fail, better to leave default and remove completely the section. If any of them is harming the device, then only disable these. 

Thanx

includes/definitions/hatteras.yaml
```diff
@@ -0,0 +1,58 @@
+os: hatteras 
+text: 'Hatteras Overdue DSLAM'
+type: dslam
```
PipoCanaja, line 3: `type: dslam` does not exist. 
`type: network` is what should be deployed for a network equipment.

html/images/logos/hatteras.svg
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="200"
```
PipoCanaja, line 5: html/images/logos/hatteras.svg: SVG files must not contain length or width attributes . This is verified automatically and makes the test fail. Could you please remove those from all the SVG files ? 
Thanx

PipoCanaja (changes requested):

includes/definitions/hatteras.yaml
Bierchermuesli: fixed

includes/definitions/hatteras.yaml
Bierchermuesli: Those modules really don't work with this device, so I hope for performance better to disable. wrong approach?

html/images/logos/hatteras.svg
Bierchermuesli: fixed

includes/definitions/hatteras.yaml
```diff
@@ -0,0 +1,59 @@
...
+discovery:
+    - sysDescr:
+            - HN4000-I
+            - HN4000e
+bad_if:
+  - notPresent
+  - serConsole
+good_if:
+  - ras #all interfaces has 'Hatteras' in it...
+discovery_modules:
```
Jellyfrog, line 19: We prefer to not set these at all if not for good reason, like if the device crashes or so

includes/definitions/hatteras.yaml
murrant: You gain about 2ms in polling time by disabling these and there could be some random user that wants a specific module enabled or someone adds support for another one, and this will prevent that.

We prefer to have them enabled to allow the users full control.  If there is a module that is slow polling when there is no data, then that is a bug.

murrant:
I'm going to remove the modules so we can merge this.
You are more than welcome to disable them on your local install:
`lnms config:set os.hatteras.discovery_modules.wireless false`

murrant (approved):

PR #14117, opened 2022-07-18
Ping perf

00gh:
Correcting generated legend value for MIN in ping_perf. 

The generated graphs shows wrong legend value for "Min".
It is displaying the calculated AVERAGE value instead of th
calculated MIN.

Changed AVARAGE into MIN.


murrant (approved):
Odd.  I also note that this RRD only records the RTT for a single ping where as LibreNMS pings 3 times by default and has min/max/avg for all the pings sent.  These values are stored in the DB and graphed by nonstandard means...

PR #14141, opened 2022-07-26
View Ports in Portgroups

electrocret:

Add view of PortGroups similar to Devicegroups.

#### Please note

#### Testers


resources/views/layouts/menu.blade.php
```diff
@@ -334,9 +334,17 @@ class="fa fa-rocket fa-fw fa-lg"
...
 
                             <li role="presentation" class="divider"></li>
-                            <li><a href="{{ url('port-groups') }}"><i class="fa fa-th fa-fw fa-lg"
-                                                                      aria-hidden="true"></i> {{ __('Manage Groups') }}
-                            </a></li>
+                            <li class="dropdown-submenu">
+							<a href="{{ url('port-groups') }}"><i class="fa fa-th fa-fw fa-lg" aria-hidden="true"></i> {{ __('Port Groups') }}</a>
+							@if($port_groups->isNotEmpty())
+								<ul class="dropdown-menu scrollable-menu">
+								<li><a href="{{ url('port-groups') }}"><i class="fa fa-th fa-fw fa-lg" aria-hidden="true"></i> {{ __('Manage Groups') }} </a></li>
```
murrant, line 341: Manage Groups does not show now if there are no port group.

includes/html/pages/ports.inc.php
```diff
@@ -404,6 +404,13 @@
                     }
                 }
                 break;
+                case 'group':
```
murrant, line 407: This only filters the graph view, and your current code does not work.  Try:
```php
                $where .= ' AND port_id IN (';
                $where .= DB::table('port_group_port')->where('port_group_id', $vars['group'])->pluck('port_id')->implode(',');
                $where .= ')';
```

murrant: You could also join the table and filter that way, but the query isn't set up for that so it would take a little deeper changes.

To filter the ports table, you'll need to edit app/Http/Controllers/Table/PortsController.php

murrant: Adding this to the filterFields function
```php
'group' => function ($query, $group) {
    /** @var \Illuminate\Database\Eloquent\Builder $query */
    return $query->whereHas('groups', function ($query) use ($group) {
        /** @var \Illuminate\Database\Eloquent\Builder $query */
        return $query->where('id', $group);
    });
 ```
 and this to ncludes/html/pages/ports/list.inc.php
 ```
 group: '<?php echo $vars['group']; ?>',
 ```

File: includes/html/pages/ports.inc.php [hunk 3]
```diff
@@ -404,6 +404,11 @@
                     }
                 }
                 break;
+            case 'group':
+                $where .= ' AND port_id IN (';
+                $where .= DB::table('port_group_port')->where('port_group_id', $vars['group'])->pluck('port_id')->implode(',');
```
Jellyfrog, line 409: Normally you do a subquery instead of two queries

murrant: Yeah, would you cheat and just add ->toSql() on the end of the current Fluent query? :D

murrant (dismissed):
Nice work @electrocret

murrant (approved):

PR #14145, opened 2022-07-26
Correct logic for recurring alert rules that span UTC days

gs-kamnas:
This change addresses a bug where the behavior of recurring alert rules that span days (after conversion to UTC) in the polling/alerting processes differs from the status (set or current) reported via the UI.

Specifically in the existing / suspected buggy implementation, the logic for the UI here: https://github.com/librenms/librenms/blob/master/app/Models/AlertSchedule.php#L165

...is not logically/functionally equivalent to the database query at https://github.com/librenms/librenms/blob/master/app/Models/AlertSchedule.php#L192 used when actually dispatching alerts.

Therefore, this change updates the DB query logic to be logically equivalent to what is shown in the UI, under the only assumption that the timezone configured on the poller running the alerting task and the timezone on the UI server are identical.

#### Please note

#### Testers

### DIFF
--- a/app/Models/AlertSchedule.php
+++ b/app/Models/AlertSchedule.php
@@ -173,11 +173,12 @@ class AlertSchedule extends Model
     {
         return $query->where(function ($query) {
             $now = CarbonImmutable::now('UTC');
+            $localDayNum = CarbonImmutable::now()->format('N');
             $query->where('start', '<=', $now)
                 ->where('end', '>=', $now)
-                ->where(function ($query) use ($now) {
+                ->where(function ($query) use ($now, $localDayNum) {
                     $query->where('recurring', 0) // Non recurring simply between start and end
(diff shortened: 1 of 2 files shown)


murrant:
Nice, @gs-kamnas 

Any chance you could add some cases in `tests/Feature/TestScheduledMaintenance.php`?

Actually, it does look like there are two commented out there, maybe just uncommenting them is enough.

gs-kamnas:
Will look into testing this better indeed.

However, I believe the current test cases are currently only executing the UI status logic (not the actual DB query); which is how this slipped through initially.

Therefore, I believe this will require more work than just uncommenting those 2 cases.

murrant:
@gs-kamnas I uncommented the two test cases and they pass with your changes, so great :)

murrant (approved):
Fixes a known bug, thanks!

PR #14146, opened 2022-07-26
Fix 500 error in validation when UI containers and poller containers have differing node_ids

gs-kamnas:
This change fixes a 500 error during validation that is observed in containerized deployments of LibreNMS where the node_id field of UI/API server containers is unique and not equal to the poller containers.

Expected behavior in this case would be for the message "Dispatcher service is enabled on your cluster, but not in use on this node" to be printed.

#### Please note

#### Testers


murrant:
Unfortunately, this validation looks just wrong.  Your current fix makes it even more wrong.

It should check if PollerCluster has any nodes.
Check if this node is a poller.
Check if any poller nodes have not reported in recently and list them.

murrant (approved):

PR #14148, opened 2022-07-27
New Poller validations

murrant:
Seperated poller and distributed poller validations to make poller validations generally available
One auto fixer added
Translatable strings

#### Please note

#### Testers


LibreNMS/Validations/Poller/CheckLocking.php
```diff
@@ -0,0 +1,55 @@
...
+
+class CheckLocking implements \LibreNMS\Interfaces\Validation
+{
+    /**
+     * @inheritDoc
+     */
+    public function validate(): ValidationResult
+    {
+        try {
+            $lock = \Cache::lock('dist_test_validation');
```
Jellyfrog, line 38: Just to be sure it disappears?
```suggestion
            $lock = \Cache::lock('dist_test_validation', 5);
```

LibreNMS/Validations/Poller/CheckPythonWrapper.php
```diff
@@ -0,0 +1,93 @@
...
+     * @inheritDoc
+     */
+    public function validate(): ValidationResult
+    {
+        if (Poller::exists() || $this->wrapperCronEnabled()) {
+            return $this->checkPythonWrapper();
+        }
+
+        if ($this->could_not_check_cron) {
+            return ValidationResult::warn(trans('validation.validations.poller.CheckPythonWrapper.cron_unread'));
```
Jellyfrog, line 46: Hm i think we should add another level. Notice or something. Because it's not really a warning if we can't read those folders, quite the opposite I'd say, it's a good secuity measure :)

LibreNMS/Validations/Poller/CheckLocking.php
murrant: Good idea.

LibreNMS/Validations/Poller/CheckPythonWrapper.php
murrant: ok is basically the same as info/notice.

murrant: Turns out info already exists :D

murrant: it always shows on my system, because I don't have a cron installed.

murrant:
Merging these so we can see what sort of false positives pop up in the wild.

PR #14152, opened 2022-07-27
[apps] Docker only show current containers

Npeca75:
RRDs associated with APPs are created when new APP is discovered, but newer deleted until RRD purge period. This is normally few months/year 
one of problematic APPs is Docker
Problem is that LibreNMS will stack these inactive RRDs on graph, so after few "lets try some containers" it is getting messy. Dozen of inactive RRDs are shown in cumulative graph

This PR add filter ability to filter out unused RRDs matching against active APPs in ApplicationMetrics table

#### Please note

#### Testers


Npeca75:
lnms1
lnms2

Before

Npeca75:
lnms3
lnms4

... after

LibreNMS/Data/Store/Rrd.php
```diff
@@ -511,7 +515,20 @@ public function getRrdApplicationArrays($device, $app_id, $app_name, $category =
             }
         }
 
-        return $entries;
+        //apply filter
+        if ($filter == 'metrics' && $entries) {
+           foreach ($entries as $rrdName) {
+                $cnt = ApplicationMetrics::where('app_id', $app_id)
+                    ->where('metric', 'like', $rrdName . '%')
+                    ->count();
```
Jellyfrog, line 523:
```suggestion
                    ->exists();
```

phpstan-baseline.neon
```diff
@@ -8290,6 +8290,11 @@ parameters:
 			count: 1
 			path: app/Models/Application.php
 
+		-
+			message: "#^PHPDoc type array of property App\\\\Models\\\\ApplicationMetrics\\:\\:\\$primaryKey is not covariant with PHPDoc type string of overridden property Illuminate\\\\Database\\\\Eloquent\\\\Model\\:\\:\\$primaryKey\\.$#"
```
Jellyfrog, line 8294: Not allowed to add stuff to this file

LibreNMS/Data/Store/Rrd.php
Npeca75: done, changed as suggested

phpstan-baseline.neon
Npeca75: hi @Jellyfrog 
if you look at table "application_metrics" you will see that there is composite primary key
`UNIQUE INDEX `application_metrics_app_id_metric_unique` (`app_id`, `metric`) USING BTREE`

only existing  App/Models with two primaryKeys is UserPref.php
and it is listed in "phpstan-baseline.neon" exactly from this reason
`    protected $primaryKey = ['user_id', 'pref'];`

without this, error is thrown on check
so i make same as UserPref.php

if you have any other suggestion ? how to solve this other way ?

Npeca75: error is about $primaryKey is array in my case

`PHPDoc type array of property App\Models\ApplicationMetrics::$primaryKey is not covariant with PHPDoc type string of overridden property Illuminate\Database\Eloquent\Model::$primaryKey. `

murrant (changes requested):
NACK from me.

There is a new data field on the applications table.  Just store the list of current application items there and use that to generate the graphs.

Npeca75:
hi @murrant 
> There is a new data field on the applications table. 

ok, first time dealing with this field
done as in ZFS
hope it is OK now

murrant (approved):
Much better! Thanks @Npeca75

murrant (approved):

murrant (approved):

PR #14154, opened 2022-07-28
Cisco SB/CBS environment sensors

Fehler12:
Added detection for environment sensors (FAN and PSU) for cisco small business devices

Pduring development I have encountered an error with the newer CBS models, which are also covered by this: the status for POE-usage and locked stats are only polled for the first ten interfaces: https://community.librenms.org/t/cbs250-snmp-does-not-fetch-all-values/19235

I'm pretty sure that this has nothing todo with this changes but I'm not 100% sure because I cannot find any reason.

#### Please note

#### Testers


Fehler12:
PS: I did not add any test data because currently the generated test data produce false test entries because of the bug mentioned above.

murrant:
There still needs to be test data.  Add it and if any error show up, we can help you resolve them.

Fehler12:
I've added test data but with false results for the POE detection.

tests/data/ciscosb_cbs250-24p-4x.json
```diff
@@ -0,0 +1 @@
+null
```
murrant, line 1: ???

PR #14169, opened 2022-08-01
base64 gzip compression support for json_app_get

VVelox:
This adds compressed return handling for json_app_get.

#### Please note

#### Testers


includes/polling/functions.inc.php
```diff
@@ -626,17 +630,43 @@ function json_app_get($device, $extend, $min_version = 1)
...
+                $i = strpos($output, "\0", $i) + 1;
+            }
+            if ($flg & 16) {
+                $i = strpos($output, "\0", $i) + 1;
+            }
+            if ($flg & 2) {
+                $i = $i + 2;
+            }
+        }
+        $output = gzinflate(substr($output, $i, -8));
```
Jellyfrog, line 661: not sure what all of this does, but why not do something like;

```php
if (
  (($output= base64_decode($output, true)) !== false) 
  &&
  (($output= gzdecode($output)) !== false)
) {
 // unpacked gzip+base64 data
}


```

VVelox: It yanks the header off so gzinflate works on the string in question. http://www.faqs.org/rfcs/rfc1952.html

Just tried gzdecode and that worked with you needing to yank the header, so switched to that. Thanks!

doc/Extensions/Applications.md
```diff
@@ -51,6 +51,84 @@ like that for proxmox:
...
+items.
+
+This does require changing out the extend works, requires using it
+from a cached state as netsnmp does not play nice with commands
+involving a pipe.
+
+To do this just set up a crontab entry such as below.
+
+```
+*/5 * * * * /etc/snmp/extends/portactivity smtps,http,imap,imaps,postgresql,https,ldap,ldaps,nfsd,syslog-conn,ssh,matrix,gitea | /usr/local/bin/lnms_return_optimizer > /var/cache/portactivity.json
```
murrant, line 76: Because of the way native snmpd caching works, I'm not a fan of this approach.

doc/Extensions/Applications.md
```diff
@@ -51,6 +51,84 @@ like that for proxmox:
...
+```
+
+And then snmpd.conf as below.
+
+```
+extend portactivity /bin/cat /var/cache/portactivity.json
```
murrant, line 82: I suspect prepending it to the command would be better.

```
extend portactivity /usr/local/bin/lnms_return_optimizer /etc/snmp/extends/portactivity smtps,http,imap,imaps,postgresql,https,ldap,ldaps,nfsd,syslog-conn,ssh,matrix,gitea
```

Then have your optimizer run the command (passing arguments) and then compress the output.

Also, please avoid using lnms and instead use librenms everywhere (except when referencing the actual lnms script)

murrant (changes requested):

doc/Extensions/Applications.md
VVelox: Awesome! Sounds good and I like that idea. :)

includes/polling/functions.inc.php
```diff
@@ -626,17 +630,26 @@ function json_app_get($device, $extend, $min_version = 1)
...
+    $orig_output = $output;
+
     // make sure we actually get something back
     if (empty($output)) {
         throw new JsonAppPollingFailedException('Empty return from snmp_get.', -2);
     }
 
+    // checks for base64 decoding and converts it to non-base64 so it can gunzip
+    if (preg_match('/^[A-Za-z0-9\/\+\n]+\=*\n*$/', $output)) {
+        $output = base64_decode($output);
```
Jellyfrog, line 643: should probably add the strict flag and check the result

includes/polling/functions.inc.php
```diff
@@ -626,17 +630,26 @@ function json_app_get($device, $extend, $min_version = 1)
...
+        $output = gzdecode($output);
```
Jellyfrog, line 644: maybe check if output is false

doc/Extensions/Applications.md
VVelox: Renamed in https://github.com/librenms/librenms-agent/pull/419 .

Still need to get the docs updated as well as re-write it a bit to work with the command appended instead of just pipped through.

includes/polling/functions.inc.php
VVelox: No need for strict. The regexp takes care of that.

But aye, possible it was improperly generated and is malformed outside of improper characters.

doc/Extensions/Applications.md
VVelox: Done!

includes/polling/functions.inc.php
```diff
@@ -626,17 +630,26 @@ function json_app_get($device, $extend, $min_version = 1)
...
+    }
+
     //  turn the JSON into a array
     $parsed_json = json_decode(stripslashes($output), true);
 
     // improper JSON or something else was returned. Populate the variable with an error.
     if (json_last_error() !== JSON_ERROR_NONE) {
-        throw new JsonAppParsingFailedException('Invalid JSON', $output, -3);
+        throw new JsonAppParsingFailedException('Invalid JSON', $orig_output, -3);
```
murrant, line 652: Why orig_output?  base64 encoded string won't be very helpful.

VVelox:
```
    // improper JSON or something else was returned. Populate the variable with an error.
    if (json_last_error() !== JSON_ERROR_NONE) {
        throw new JsonAppParsingFailedException('Invalid JSON', $orig_output, -3);
    }
```

Because decoding it may be a mistake given the expect operation of this function supporting returning the raw return upon a JSON decoding failure for fall back non-JSON processing via the poller.

That said not aware of thing currently that could vaguely be mistaken as looking like BASE64 for when it comes to any of the pollers that have older possible non-JSON data handling. This is more for being future friendly.

Also lets say there is a extend that gets created that produces some really simple new line seperated data, this could easily be mistaken for BASE64. Then some time down the line it gets converted to use JSON and then this function, on the off chance it properly decades the base64 and then amazingly ungzips it, it would then fail at this point and return corrupted data.

That said will want to do return that as well when it comes to implementing the checks that @Jellyfrog brought up. Will need to add in a two new exceptions as well for that. :)

includes/polling/functions.inc.php
VVelox: @Jellyfrog Added!

murrant (approved):
LGTM

murrant (approved):

murrant (approved):

PR #14179, opened 2022-08-04
Fix unnecessary Updates of DB when values are equal

opalivan:
With some devices (Packetlight), we have at each discovery an update of the limits. But the update uses the exact same value. It ends up filling the eventlogs for nothing.

While checking function.inc.php, it appears that the auto-cast of string/int/float does not end up right. This PR forces (string) cast to ensure proper string comparison. 

Here is a capture of the eventlog (with some extra data added during the debug phase).

image

#### Please note

#### Testers


PipoCanaja (approved):
LGTM. It is during discovery, so even if comparing strings is a little slower, no big deal ...

PR #14185, opened 2022-08-05
Fix for number in sensor string

Schouwenburg:
In the discovery_process function there is a regex to find the first number in a sensor value result if the result is a string. That same function was missing in the poll_sensor function.

#### Please note

#### Testers


includes/polling/functions.inc.php
```diff
@@ -80,6 +80,13 @@ function poll_sensor($device, $class)
...
 
+            if (! is_numeric($sensor_value)) {
+                preg_match('/-?\d*\.?\d+/', $sensor_value, $temp_response);
+                if (! empty($temp_response[0])) {
+                    $sensor_value = $temp_response[0];
+                }
+            }
+
             if ($class == 'temperature') {
                 preg_match('/[\d\.\-]+/', $sensor_value, $temp_response);
```
murrant, line 91: This duplicates your above code (with a worse regex) and can e removed.

includes/polling/functions.inc.php
```diff
@@ -80,6 +80,13 @@ function poll_sensor($device, $class)
                 require 'includes/polling/sensors/' . $class . '/' . $device['os_group'] . '.inc.php';
             }
 
+            if (! is_numeric($sensor_value)) {
+                preg_match('/-?\d*\.?\d+/', $sensor_value, $temp_response);
```
Jellyfrog, line 84: This regex is wrong. It would match `-.1` for example. 
```suggestion
                preg_match('/-?\d+(?:\.\d+)?/', $sensor_value, $temp_response);
```
Think this covers the cases

Schouwenburg: Hi, I think -.1 is correct as a result of the regex and no change is needed. Also, this is a 1:1 copy of the same code in the discovery logic.
<?php
$value = -.1;
$type = gettype($value);
print $type;
$value++;
print "<br>";
print $value;
>>>
double
0.9

includes/polling/functions.inc.php
Schouwenburg: Duplicate code now removed.

murrant:
@Schouwenburg test data says you fixed some devices that were broken :)

Schouwenburg:
> @Schouwenburg test data says you fixed some devices that were broken :)

Yes I did see. So the question is, what now? Do I need to create new test data for the ones that fail?

murrant:
> Do I need to create new test data for the ones that fail?

Yes

murrant (approved):

PR #14190, opened 2022-08-09
Error reporting

Jellyfrog:

#### Please note

#### Testers


config/flare.php
```diff
@@ -0,0 +1,70 @@
...
+    | These options determine which information will be transmitted to Flare.
+    |
+    */
+
+    'reporting' => [
+        'anonymize_ips' => true,
+        'collect_git_information' => true,
+        'report_queries' => true,
+        'maximum_number_of_collected_queries' => 200,
+        'report_query_bindings' => true,
```
murrant, line 40: You sure about query bindings?

config/flare.php
```diff
@@ -0,0 +1,70 @@
...
+        'report_view_data' => true,
```
murrant, line 41: And view data?

app/Providers/AppServiceProvider.php
```diff
@@ -55,6 +62,94 @@ public function boot()
...
+            // Check if repo is modified
+            if (! Git::unchanged()) {
+                return false;
+            }
+
+            return true;
+        });
+
+        Flare::determineVersionUsing(function () {
+            return \LibreNMS\Util\Version::VERSION;
```
murrant, line 104: Might be good to get the more specific version from Version::get()->local() (I think)

config/flare.php
Jellyfrog: Think we should start with it and see if it contains too much sensitive info

config/flare.php
Jellyfrog: Ditto

Jellyfrog:
It doesn't report deprecation errors, but we don't currently log that anyway.

Jellyfrog:
Seems normal exceptions (unauthorized etc) is thrown when using app debug, needs investigation...

app/Providers/AppServiceProvider.php
```diff
@@ -55,6 +62,94 @@ public function boot()
...
+                return false;
+            }
+
+            // Check if git installation
+            if (! Git::repoPresent()) {
+                return false;
+            }
+
+            // Repo url must be offical one
+            if (! Str::contains(Git::remoteUrl(), ['git@github.com:librenms/librenms.git', 'https://github.com/librenms/librenms.git'])) {
```
Jellyfrog, line 91: This check is wrong since it only checks origin branch, and not *current* branch.

Jellyfrog:
if `APP_DEBUG=true` and you install composer with `--no-dev` it shows the symfony exception handler instead of ignition.
Due to `filp/whoops` being a dev dep.

murrant:
1. Why not remove filp/whoops?
2. It is non-trivial to remove the full path from the reports.  Stack trace will contain full path (and many other things).  The best place to do this would be upstream.  I think it is fairly harmless.

resources/views/layouts/error.blade.php
```diff
@@ -76,8 +76,8 @@
 
     <p>{{ __("If you need additional help, you can find how to get help at") }} <a target="_blank" href="https://docs.librenms.org/Support">https://docs.librenms.org/Support</a>.</p>
 
-    @if(config('flare.key'))
-    <p>{{ __("Please include this error-ID when reporting problems:") }} <b>{{ \Facade\Ignition\Facades\Flare::sentReports()->latestUuid() }}</b></p>
+    @if(! empty(Flare::sentReports()->all()))
+    <p>{{ __("Please include this Error-ID when reporting problems:") }} <a href="{{ Flare::sentReports()->latestUrl() }}">{{ Flare::sentReports()->latestUuid() }}</a></p>
```
Jellyfrog, line 80: Shouldn't be a link. That just takes the user to a login prompt on flare they don't have access to anyway.

murrant: Ah, makes sense

murrant (approved):

murrant (approved):

crcro:
i thinks this commit breaks the app:

image

murrant:
@crcro please elaborate

crcro:
@murrant this commit adds the line "return new \App\Facades\LogManager($app)" which then generates the above error when trying to access librenms ui

murrant:
@crcro it didn't add that line.  It has always been there, only changed it from bind to a singleton so that we don't get a new LogManager created every time.  (upstream was already singleton)

PR #14191, opened 2022-08-09
Fix number sensor string

Schouwenburg:
Hi, after fixing the number from string extraction I would also like to propose to prefer a float extraction above a int extraction. I ran into a use case where this really makes sense.

#### Please note

#### Testers


murrant:
You are going to have troubles.  This is the quantastor string it is parsing `32 C (89.60 F)`  (LibreNMS wants temperatures in C)

Schouwenburg:
> You are going to have troubles. This is the quantastor string it is parsing `32 C (89.60 F)` (LibreNMS wants temperatures in C)

So i could skip this if sensor_class is temperature. What do you think, should I also skip this for sensor_class state?

tests/data/asa.json
```diff
@@ -3131,126 +3131,6 @@
     "sensors": {
         "discovery": {
             "sensors": [
-                {
```
PipoCanaja, line 3134: We see multiple sensors deleted. It probably shows that something is going wrong with the new code  preventing existing sensors to survive the change. You should end up with no sensor deleted if the new code is safe.

tests/data/bats.json
```diff
@@ -202,7 +202,7 @@
                     "group": "Positioner",
                     "sensor_divisor": 1,
                     "sensor_multiplier": 1,
-                    "sensor_current": 17,
```
PipoCanaja, line 205: Here, the value looks very different, it must be checked if this is correct or not

tests/data/vrp_ar169sfp.json
```diff
@@ -3509,102 +3509,6 @@
                     "sensor_prev": null,
                     "user_func": null,
                     "state_name": null
-                },
```
PipoCanaja, line 3512: Same here. This device I have it running and would rather have a problem if temperature is broken :)

PipoCanaja (changes requested):
Hi @Schouwenburg 
Seems that the code is not completely accurate currently. I added a few comments but most of the JSON files with deleted code should not exist. Modified values are acceptable, but deleted sensors means loss of data.

Schouwenburg:
> Hi @Schouwenburg Seems that the code is not completely accurate currently. I added a few comments but most of the JSON files with deleted code should not exist. Modified values are acceptable (if they can be explained), but deleted sensors means loss of data.

Agree, I'm not yet sure where it goes wrong. I will do some debugging, should all be logical :-)

tests/data/bats.json
murrant: 344.4 is incorrect, looking at the data (it should be 17).

Jellyfrog:

PR #14192, opened 2022-08-09
Allow specifying the background colour in graph images

washcroft:
As per title, otherwise the background of the graph image is transparent which may not be desirable.

(extreme example - yellow background)

image

bg is still graph background, new bbg parameter controls the back-background

#### Please note

#### Testers

### DIFF
--- a/includes/html/graphs/common.inc.php
+++ b/includes/html/graphs/common.inc.php
@@ -94,11 +94,14 @@ if ($_GET['bg']) {
     $rrd_options .= ' -c CANVAS#' . Clean::alphaDash($_GET['bg']) . ' ';
 }
 
+if ($_GET['bbg']) {
+    $rrd_options .= ' -c BACK#' . Clean::alphaDash($_GET['bbg']) . ' ';
+}
+
 if ($_GET['font']) {
     $rrd_options .= ' -c FONT#' . Clean::alphaDash($_GET['font']) . ' ';
 }
 
-// $rrd_options .= " -c BACK#FFFFFF";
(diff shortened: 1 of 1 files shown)


murrant:
What is the use case for this, since the default is transparent?  You can just make whatever is behind it yellow :D

washcroft:
> What is the use case for this, since the default is transparent? You can just make whatever is behind it yellow :D

Not always possible... in this case the graphs are being displayed in an app that changes it's background to light/dark depending on time of day. The graphs look much better / more readable if they have their own consistent solid background, irrespective of what they're being displayed on.

murrant (approved):

LukaszC86:
It seems that this solution doesn't work with @signedGraphTag in the template - the attached graphs to the mail messages are always transparent.

Tested with:
@signedGraphTag([
    'id' => $value['port_id'],
    'type' => 'port_bits',
    'from' => time() - 43200,
    'to' => time(),
    'width' => 1000, 
    'height' => 355,
    'bbg' => 'FFFFFF',
])

washcroft:
Does the signed URL that gets included in the email include the bbg parameter?

LukaszC86:
Dear Warren,
I pasted exactly the part of the template which I use to attach the graph to the mail messages - it has "bbg" parameter. I tried it also together with "bg" parameter, but there's no difference. 
image

washcroft:
I can see, but what actually gets put in the email, the fully constructed image URL, does it have the bbg parameter in the URL?

PR #14210, opened 2022-08-16
Added support for UIO monitoring on APC NMUs

Jimmy-Cl:

Added support for temperature and contact switch monitoring for APC NMUs.

image

This is used with UPS Network Management Card 2 & 3 which are installed in a standalone unit without a UPS. When they are used without a UPS they report UIO data on different OIDs. 

#### Please note

#### Testers


murrant:
@Jimmy-Cl Thanks for the contribution.  There are a couple of issues with your PHP.  But can you change these to yaml instead?
Let me know if you need any help.

Jimmy-Cl:
@murrant Cheers for the feedback. I did try to use YAML originally however I found it wouldn't handle the IF statements. Just for some context on why I'm am using the below, 
`$device_type = snmp_get($device, '.1.3.6.1.2.1.1.2.0', '-OqvnU', 'PowerNet-MIB');` 
`&& $device_type == '.1.3.6.1.4.1.318'`
One of our other UPS's an APC SMX750i will show the above temperature and state sensors twice without the IF statement as they are already monitored with different OIDs, I added the Device OID check to make sure it can only be used for the standalone environmental monitoring units. I will try again with YAML if you are still not happy or let me know what your issues were with the PHP and I will try it improve that.

murrant:
@Jimmy-Cl You would do that like this:

pre-fetch it because the value is not already fetch
```yaml
        pre-cache:
            data:
                -
                    oid: SNMPv2-MIB::sysObjectID.0
                    snmp_flags: '-On'
```

then in your sensors:
```yaml
                    skip_values:
                        -
                            oid: SNMPv2-MIB::sysObjectID.0
                            op: '!='
                            value: '.1.3.6.1.4.1.318'
```

(this may not be exactly correct as I didn't test it)

murrant:
Actually, It is even easier than that (you don't need the pre-fetch):
                    skip_values:
                        -
                            device: sysObjectID
                            op: '!='
                            value: '.1.3.6.1.4.1.318'

PR #14211, opened 2022-08-16
Update ports test data to match current data

murrant:

#### Please note

#### Testers


Jellyfrog:
Hmm, why does it work on master?

murrant:
@Jellyfrog because it uses lazy comparisons so null == 0 is true.  But when updating the data, it causes a lot of noise.  This is an effort to reduce that noise.

Jellyfrog:
That makes sense. Does it make sense to do strict compares?

murrant:
@Jellyfrog maybe

PipoCanaja:
Is there something we can do with these backslash ?
I searched (without success) where this would come from. Dirty solution : check the snmprec file and remove them, so they don't keep breaking everything later...

murrant:
That is the easy way out, which might be worth it.  I think it has something to do with mibs.

PR #14213, opened 2022-08-18
autodiscovery fix

Npeca75:
Two fix:

1. function cleanPort expect array, but simple variable is passed

[2022-08-18T05:48:54.991362+02:00] production.ERROR: Error discovering discovery-protocols module for 169.254.101.12. TypeError: Cannot access offset of type string on string in /opt/librenms/includes/rewrites.php:45
Stack trace:
#0 /opt/librenms/includes/discovery/functions.inc.php(88): cleanPort()
#1 /opt/librenms/includes/discovery/discovery-protocols.inc.php(233): discover_new_device()
#2 /opt/librenms/includes/discovery/functions.inc.php(158): include('...')
#3 /opt/librenms/discovery.php(118): discover_device()
#4 {main}

2. autodiscovery put new device in default poller group, which is not correct. 
if we have poller/discovery for group 99 and this group is unreachable from main LNMS, then polling will fail until newly discovered device is not changed manualy to poller_group 99
so it is best to inherit poller_group from device which discovered NEW device

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -83,14 +83,7 @@ function discover_new_device($hostname, $device = [], $method = '', $interface =
         device_by_id_cache($remote_device_id, 1);
 
         if ($remote_device_id && is_array($device) && ! empty($method)) {
-            $extra_log = '';
-            // cleanPort expect array
-            $ifArray = ['ifName' => $interface, 'ifDescr' => $interface, 'ifAlias' => $interface, 'device_id' => $device['device_id']];
-            $int = cleanPort($ifArray);
-            if (is_array($int)) {
-                $extra_log = ' (port ' . $int['label'] . ') ';
-            }
-
+            $extra_log = isset($interface) ? ' (port ' . htmlentities($interface) . ') ' : '';
             log_event('Device ' . $remote_device['hostname'] . " ($ip) $extra_log autodiscovered through $method on " . $device['hostname'], $remote_device_id, 'discovery', 1);
         } else {
             log_event("$method discovery of " . $remote_device['hostname'] . " ($ip) failed - Check ping and SNMP access", $device['device_id'], 'discovery', 5);
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -83,11 +83,7 @@ function discover_new_device($hostname, $device = [], $method = '', $interface =
         device_by_id_cache($remote_device_id, 1);
 
         if ($remote_device_id && is_array($device) && ! empty($method)) {
-            $extra_log = '';
-            if (is_array($interface)) {
-                $int = cleanPort($interface);
-                $extra_log = ' (port ' . $int['label'] . ') ';
-            }
+            $extra_log = is_array($interface) ? ' (port ' . cleanPort($interface)['label'] . ') ' : '';
 
             log_event('Device ' . $remote_device['hostname'] . " ($ip) $extra_log autodiscovered through $method on " . $device['hostname'], $remote_device_id, 'discovery', 1);
         } else {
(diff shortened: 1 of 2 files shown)


includes/discovery/functions.inc.php
```diff
@@ -76,15 +76,17 @@ function discover_new_device($hostname, $device = [], $method = '', $interface =
...
         echo '+[' . $remote_device['hostname'] . '(' . $remote_device['device_id'] . ')]';
         discover_device($remote_device);
         device_by_id_cache($remote_device_id, 1);
 
         if ($remote_device_id && is_array($device) && ! empty($method)) {
             $extra_log = '';
-            $int = cleanPort($interface);
+            // cleanPort expect array
+            $ifArray = ['ifName' => $interface, 'ifDescr' => $interface, 'ifAlias' => $interface, 'device_id' => $device['device_id']];
+            $int = cleanPort($ifArray);
```
murrant, line 89: I think the correct fix is to move `$int = cleanPort($interface);` down one line inside the `is_array($int)` block

murrant: (and change $int to $interface in that check)

Npeca75: > I think the correct fix is to move `$int = cleanPort($interface);` down one line inside the `is_array($int)` block

Hi @murrant 
i like to ask you to take careful look at discovery-protocols.inc.php file
you will see that every time function is called similar to this:
```$remote_device_id = discover_new_device($lldp['lldpRemSysName'], $device, 'LLDP', $interface);```

so, it always pass plain variable

if i put ```is_array($interface)``` check in discover_new_device function, the EXTRA_LOG will bi never triggered, because we always send flat $interface variable like 'gigabitethernet 1/0/11'

my intention was to convert plain $interface to array so we could get this extra_log line in eventlog

but if you insist ...

**Comment by Npeca75** [comment]
```
2022-08-18 09:32:17 | discovery | soada-wr07 | Device soada-wr07 (169.254.102.107) (port 1/0/23) autodiscovered through LLDP on 169.254.102.7
```

File: includes/discovery/functions.inc.php [hunk 2]
```diff
@@ -76,19 +76,14 @@ function discover_new_device($hostname, $device = [], $method = '', $interface =
...
         device_by_id_cache($remote_device_id, 1);
 
         if ($remote_device_id && is_array($device) && ! empty($method)) {
-            $extra_log = '';
-            $int = cleanPort($interface);
-            if (is_array($int)) {
-                $extra_log = ' (port ' . $int['label'] . ') ';
-            }
-
+            $extra_log = isset($interface) ? ' (port ' . htmlentities($interface) . ') ' : '';
```
murrant, line 86: FYI, $interface is never a string.  EXCEPT when it is not set because of the braindead function defintion here. `$interface = ''`

htmlentities($interface) will fail because it is an array (when port is actually given to the function)  otherwise, it will just add ` (port )` to the log because it will always be an empty string.

murrant: cleanPort() will set $interface['label'] which is not a database field.  (Which is why it is called)

murrant: FYI, I found out the above by looking at all the calls to discover_new_device()

includes/discovery/functions.inc.php
Npeca75: > (and change $int to $interface in that check)

Ok, done that way as you suggested

PipoCanaja:
LGTM but not tested currently

includes/discovery/functions.inc.php
murrant: @Npeca75 if you scroll up a bit in the code, you see `$interface = get_port_by_id($local_port_id);` get_port_by_id() returns an array, not a string.

murrant (approved):

PR #14219, opened 2022-08-19
Add Cisco Flash storage support

fbouynot:
Add discovery and polling for Cisco Flash storage.

I am not a developer and I could have do really bad code, please tell me if I have to rewrite things or change everything or whatever.
Also  at the moment, there is no tests data, I'm not sure what to provide. This change impact Cisco IOS, IOS-XE, and NXOS (probably IOS-XR but I have none). Do I have to provide one dataset for each device type, and variant for each OS version, or one dataset with variant for each device type?

Whatsoever, this PR works in dev + prod for me, with multiple OS versions of IOS, IOS-XE, and NXOS.

#### Please note

- [Y] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [N/A] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.
- [N] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers


fbouynot:
Also, I don't know if ternary if respect the code standard, if not I will change them.

includes/polling/storage/CISCO-FLASH-MIB.inc.php
```diff
@@ -0,0 +1,35 @@
+<?php
```
PipoCanaja, line 1: Could you please rename both files to "cisco.inc.php" instead ?

includes/discovery/storage/CISCO-FLASH-MIB.inc.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @package    LibreNMS
+ * @link       https://www.librenms.org
+ * @copyright  2022 Félix Bouynot
+ * @author     Félix Bouynot <felix.bouynot@setenforce.one>
+ */
+
+use LibreNMS\Util\Number;
+
+$ciscoFlashPartitionName = snmpwalk_cache_oid($device, 'ciscoFlashPartitionName', null, 'CISCO-FLASH-MIB');
```
PipoCanaja, line 28: Could you please add a condition here to check if we are indeed running the discovery against a Cisco Device (You can check for the group "cisco", should be set for all cisco devices).

includes/polling/storage/CISCO-FLASH-MIB.inc.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @package    LibreNMS
+ * @link       https://www.librenms.org
+ * @copyright  2022 Félix Bouynot
+ * @author     Félix Bouynot <felix.bouynot@setenforce.one>
+ */
+
+use LibreNMS\Util\Number;
+
+$oids = array('ciscoFlashPartitionSize.' . $storage['storage_index'], 'ciscoFlashPartitionFreeSpace.' . $storage['storage_index'], 'ciscoFlashPartitionSizeExtended.' . $storage['storage_index'], 'ciscoFlashPartitionFreeSpaceExtended.' . $storage['storage_index']);
```
PipoCanaja, line 28: Here, it is mandatory to check:
- Are we running polling of a Cisco device ?
- If yes, did we discoverany flash storage ?
- If still yes, then let's poll the device to get values.

PipoCanaja (changes requested):
Hello @fbouynot 
That seems like a good start. To ensure performances, we need to add a few checks. You only want to discover Cisco Flash against Cisco devices. Polling should only occur on a device with discovered storage so that one should be safe. 
When that will be done, the next step will be to upload test data, and you'll be on tracks. 
Thanx

includes/discovery/storage/CISCO-FLASH-MIB.inc.php
```diff
@@ -0,0 +1,40 @@
...
+foreach ($ciscoFlashPartitionName as $index => $name) {
+    $name = array_shift($name);
+    $oids = array('ciscoFlashPartitionSize.' . $index, 'ciscoFlashPartitionFreeSpace.' . $index, 'ciscoFlashPartitionSizeExtended.' . $index, 'ciscoFlashPartitionFreeSpaceExtended.' . $index);
+    $entry = array_shift(snmp_get_multi($device, $oids, '-OQUs', 'CISCO-FLASH-MIB'));
+    $storage_size = (Number::cast($entry['ciscoFlashPartitionSize']) === 4294967295 ? $entry['ciscoFlashPartitionSizeExtended'] : $entry['ciscoFlashPartitionSize']);
+    $storage_free = (Number::cast($entry['ciscoFlashPartitionFreeSpace']) === 4294967295 ? $entry['ciscoFlashPartitionFreeSpaceExtended'] : $entry['ciscoFlashPartitionFreeSpace']);
+    $storage_used = $storage_size - $storage_free;
+    $storage_units = 1;
+    discover_storage($valid_storage, $device, $index, 'flash', 'CISCO-FLASH-MIB', $name, $storage_size, $storage_units, $storage_used);
```
PipoCanaja, line 37: You'll have to rename to "cisco-flash" instead of "CISCO-FLASH-MIB"

includes/discovery/storage/CISCO-FLASH-MIB.inc.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
PipoCanaja, line 1: Could you please rename both files to "cisco-flash.inc.php" instead ?

includes/polling/storage/CISCO-FLASH-MIB.inc.php
fbouynot: Hi, I did originally named them like this but polling was not working because of https://github.com/librenms/librenms/blob/master/includes/polling/storage.inc.php I think, which is expecting the mib name. I will try again.

fbouynot:
> Hello @fbouynot That seems like a good start. To ensure performances, we need to add a few checks. You only want to discover Cisco Flash against Cisco devices. Polling should only occur on a device with discovered storage so that one should be safe. When that will be done, the next step will be to upload test data, and you'll be on tracks. Thanx

Hi, thanks a lot for your feedback, I'll do this on Monday I think.

includes/polling/storage/CISCO-FLASH-MIB.inc.php
murrant: This old code is a little wonky.  When you call `discover_storage()` the "mib" name is just the file name it will use for polling.  It actually has no other meaning.

includes/discovery/storage/cisco-flash.inc.php
```diff
@@ -0,0 +1,41 @@
...
+    $ciscoFlashPartitionName = snmpwalk_cache_oid($device, 'ciscoFlashPartitionName', null, 'CISCO-FLASH-MIB');
+    foreach ($ciscoFlashPartitionName as $index => $name) {
+        $name = array_shift($name);
+        $oids = array('ciscoFlashPartitionSize.' . $index, 'ciscoFlashPartitionFreeSpace.' . $index, 'ciscoFlashPartitionSizeExtended.' . $index, 'ciscoFlashPartitionFreeSpaceExtended.' . $index);
+        $entry = array_shift(snmp_get_multi($device, $oids, '-OQUs', 'CISCO-FLASH-MIB'));
+        $storage_size = (Number::cast($entry['ciscoFlashPartitionSize']) === 4294967295 ? $entry['ciscoFlashPartitionSizeExtended'] : $entry['ciscoFlashPartitionSize']);
+        $storage_free = (Number::cast($entry['ciscoFlashPartitionFreeSpace']) === 4294967295 ? $entry['ciscoFlashPartitionFreeSpaceExtended'] : $entry['ciscoFlashPartitionFreeSpace']);
+        $storage_used = $storage_size - $storage_free;
+        $storage_units = 1;
+        discover_storage($valid_storage, $device, $index, 'flash', 'CISCO-FLASH-MIB', $name, $storage_size, $storage_units, $storage_used);
```
murrant, line 38:
```suggestion
        discover_storage($valid_storage, $device, $index, 'flash', 'cisco-flash', $name, $storage_size, $storage_units, $storage_used);
```

includes/discovery/storage/CISCO-FLASH-MIB.inc.php
fbouynot: Thanks, it's renamed.

includes/discovery/storage/CISCO-FLASH-MIB.inc.php
fbouynot: Thanks, it's renamed.

includes/discovery/storage/CISCO-FLASH-MIB.inc.php
fbouynot: Thanks, there is now a check.

PipoCanaja (approved):
LGTM

PipoCanaja:
Bonjour @fbouynot 
It looks good. Thanx for your contribution !
Bonne journée !

PR #14220, opened 2022-08-19
Fix legacy os extends

murrant:
Now compatible with oids that could have been used by older software:
extend .1.3.6.1.4.1.2021.7890.2 hardware
extend .1.3.6.1.4.1.2021.7890.3 manufacturer
extend .1.3.6.1.4.1.2021.7890.3 vendor
extend .1.3.6.1.4.1.2021.7890.4 serial

Correct the oids for VERY old exec options
Update to SnmpQuery
Could not find evidence of legacy version extends
Dell service tag is already discovered, don't fetch twice.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/OS/Shared/Unix.php
+++ b/LibreNMS/OS/Shared/Unix.php
@@ -117,8 +117,7 @@ class Unix extends \LibreNMS\OS implements MempoolsDiscovery
             $device->hardware = $hardware;
         }
 
-        $serial = SnmpQuery::mibDir('dell')->get([
-            'MIB-Dell-10892::chassisServiceTagName.1', // Dell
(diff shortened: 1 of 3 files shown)


isarandi:
After the Aug 21 autoupdate, my use case broke, probably related to changes to the "manufacturer" OID handling.

To have a customizable "Platform" column in the devices list I set a custom manufacturer in `/etc/snmp/snmpd.conf` of each machine, with

```
extend .1.3.6.1.4.1.2021.7890.3 manufacturer somescript.sh
```

Previously, the information output by `somscript.sh` showed up in the "Platform" column of the LibreNMS web UI device list. But now it does not show up, only "Generic x86 64-bit" is shown.

Do you know where the problem could be?

Issue #13611, opened 2021-12-10
Dashboard tiles aren't locked on touch devices


tim427:
### The problem

Like stated in this Pull-Request: https://github.com/librenms/librenms/pull/2042

```
By default lock all tiles
Only allow moving/resizing if in Edit mode
```

This behaviour is true for normal browsers, controlled by a pointing device. When approaching the dashboard with a mobile device, with Edit mode off, I'm still able to move tiles.

This results in accidentally moving tiles on the dashboard, while scrolling on a mobile device.

### Output of ./validate.php

```txt
====================================
Component | Version
--------- | -------
LibreNMS  | 21.11.0-72-g9e65901
DB Schema | 2021_11_29_165436_improve_ports_search_index (229)
PHP       | 7.4.26
Python    | 3.6.8
MySQL     | 5.7.36
RRDTool   | 1.4.8
SNMP      | 5.7.2
====================================

[OK]    Composer Version: 2.1.14
[OK]    Dependencies up-to-date.
[OK]    Database connection successful
[OK]    Database schema correct
```

### Anything in the logs that might be useful for us?

_No response_ #

murrant:
Was able to reproduce this.  It is a bug in the upstream js library, which is no longer in development.  Last release, 2016.

tim427:
I found something; when the browser/window is slightly resized; Gridster becomes active again and therefore "losing" the disabled state.

I'm currently debugging why/when this happens.

PR #14222, opened 2022-08-22
Disable Gridster after window resize when Gridster is/should be disabled

tim427:
This will fix issue #13611

#### Please note

#### Testers

### DIFF
--- a/resources/views/layouts/menu.blade.php
+++ b/resources/views/layouts/menu.blade.php
@@ -760,7 +760,7 @@
         }, 100);
     });
 
-@if($browser_push)
+    @if($browser_push)
         if (localStorage.getItem('notifications') !== 'disabled') {
             Notification.requestPermission().then(function (permission) {
                 if (permission === "denied") {
--- a/resources/views/layouts/menu.blade.php
+++ b/resources/views/layouts/menu.blade.php
@@ -751,7 +751,16 @@
         });
     }
 
-    @if($browser_push)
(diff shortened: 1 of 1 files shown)


murrant:
This seems to be a 100ms loop that only disables gridster, never enables it... should it ever be enabled?

tim427:
> This seems to be a 100ms loop that only disables gridster, never enables it... should it ever be enabled?

It only triggers when the window is resized (which is problem with gridster); after resize of the browser window, gridster is unwantedly reactivated again

With this addition the preferred state will be checked (`gridster_state`) and disabled again when this state is prefered.

By adding a `console.log('gridster disabled');` before `gridster.disable();` and `gridster.disable_resize();`, you could see/verify this is only executed once, after a window resize event.

I've tested all scenarios;
- Desktop browser; "Edit Dashboard"-enabled -> stays moveable/resizeable after window resize
- Desktop browser; "Edit Dashboard"-disabled -> stays non-moveable/resizeable after window resize (which is a fix)
- Mobile browser; "Edit Dashboard"-enabled -> stays moveable/resizeable (which is a fix)
- Mobile browser; "Edit Dashboard"-disabled -> stays non-moveable/resizeable (which is a fix)

murrant:
Is there any event we can listen for instead of a busy loop?

murrant:
Nevermind, I totally mis-read the code.  I thought it was calling itself in the setTimeout()

murrant (approved):
Nice work!

PR #14248, opened 2022-08-25
fix issue with new firmware requiring new MIB. Iss#10449

MajesticFalcon:
The new Vertiv Geist Watchdog devices come with firmware version 3.4.0 which requires a new MIB "VERTIV-BB-MIB". I don't have the ability to run snmpsim in my environment, however, since I installed these modules and needed support, I figured I would send the changes upstream. I noticed there was a stale issue regarding the sensor support and this PR happens to address their concerns.

Additionally, I added two dry contact sensors. This is what I am using, but it also gives the user a starting point for adding or changing according to their environment. The documentation around this part isn't very clear, I think this will help future users of geist.

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/discovery/geist-watchdog.yaml
+++ b/includes/definitions/discovery/geist-watchdog.yaml
@@ -1,9 +1,36 @@
-mib: GEIST-V4-MIB:GEIST-MIB-V3
+mib: VERTIV-BB-MIB
 modules:
     os:
         version: GEIST-MIB-V3::productVersion.0
         serial: GEIST-MIB-V3::climateSerial.1
     sensors:
+      state:
+         data:
+             -
+                 oid: InternalTable
+                 value: internalIO1
+                 num_oid: '.1.3.6.1.4.1.21239.5.1.2.1.8.1'
+                 descr: 'Door Position'
+                 group: Switch Sensors
+                 index: 'doorSwitchSensor.{{ $index }}'
+                 state_name: doorSwitchSensor
+                 states:
+                     - { value: -1, generic: 3, graph: 0, descr: 'null' }
+                     - { value:  1, generic: 2, graph: 0, descr: Open }
+                     - { value:  0, generic: 0, graph: 0, descr: Closed }
+             -
+                 oid: InternalTable
+                 value: internalIO2
+                 num_oid: '.1.3.6.1.4.1.21239.5.1.2.1.9.1'
+                 descr: 'Flood Sensor'
+                 group: Switch Sensors
+                 index: 'floodSwitchSensor.{{ $index }}'
+                 state_name: floodSwitchSensor
+                 states:
+                     - { value: -1, generic: 3, graph: 0, descr: 'null' }
+                     - { value:  0, generic: 2, graph: 0, descr: 'Wet' }
+                     - { value:  1, generic: 0, graph: 0, descr: 'Dry' }
+
         temperature:
             options:
                 divisor: 10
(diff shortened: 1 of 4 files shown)


PipoCanaja:
Hi @MajesticFalcon 
Unfortunately, changing the file as you currently does most probably kills the support for devices with the old MIBs. A better option would be: 
- to find out if you have old or new firmware and discover accordingly. 
- add support for both the old and new firmware side to side (if possible) meaning you need to support all sensors of all MIB variants

MajesticFalcon:
> Hi @MajesticFalcon Unfortunately, changing the file as you currently does most probably kills the support for devices with the old MIBs. A better option would be:
> 
> * to find out if you have old or new firmware and discover accordingly.
> * add support for both the old and new firmware side to side (if possible) meaning you need to support all sensors of all MIB variants

So I should discover the os based on the existing parameters + firmware version. If the firmware version is greater than or equal to 3.4 (the latest), then use a different OS. Maybe the os is called geist-watchdog3.4?

PipoCanaja:
> > Hi @MajesticFalcon Unfortunately, changing the file as you currently does most probably kills the support for devices with the old MIBs. A better option would be:
> > 
> > * to find out if you have old or new firmware and discover accordingly.
> > * add support for both the old and new firmware side to side (if possible) meaning you need to support all sensors of all MIB variants
> 
> So I should discover the os based on the existing parameters + firmware version. If the firmware version is greater than or equal to 3.4 (the latest), then use a different OS. Maybe the os is called geist-watchdog3.4?

That's option one indeed. If you can make it with that one, this is probably the cleanest of the 2 variants :) To make it work, you need to exclude from the existing OS any firmware > 3.4, and in the new one accept only if firmware >= 3.4. Cause the OS selecting code should have only one solution.

murrant:
It is probably appropriate to define both sets of sensors and let discovery figure out which ones are valid.

MajesticFalcon:
Since these changes have not been pulled into master, what is the best way to get the daily_update script to skip trying to update these files?

I get the following:
WARN: Your local git contains modified files, this could prevent automatic updates.

includes/definitions/discovery/geist-watchdog.yaml
includes/definitions/geist-watchdog.yaml
includes/discovery/sensors/temperature/geist-watchdog.inc.php
includes/polling/sensors/pre-cache/geist-watchdog.inc.php
includes/polling/sensors/temperature/geist-watchdog.inc.php

PipoCanaja:
> Since these changes have not been pulled into master, what is the best way to get the daily_update script to skip trying to update these files?
> 
> I get the following: WARN: Your local git contains modified files, this could prevent automatic updates.
> 
> includes/definitions/discovery/geist-watchdog.yaml includes/definitions/geist-watchdog.yaml includes/discovery/sensors/temperature/geist-watchdog.inc.php includes/polling/sensors/pre-cache/geist-watchdog.inc.php includes/polling/sensors/temperature/geist-watchdog.inc.php

Unfortunately, not. Either you submit a pull request that does not break the OS for everybody else, or you break you own install which means you may suffer issues with later updates on the same files.

murrant:
The daily script will never overwrite your local changes.  You can safely ignore the warning until this gets merged.

murrant:
@MajesticFalcon I was just going to evaluate your changes, but `VERTIV-BB-MIB` is missing from this PR.

MajesticFalcon:
> This is why we encourage you and help you getting it "mergeable" :)

I just don't quite have the time to devote to wrapping this up at this time.

MajesticFalcon:
> @MajesticFalcon I was just going to evaluate your changes, but `VERTIV-BB-MIB` is missing from this PR.

[mib (1) 2.zip](https://github.com/librenms/librenms/files/9686332/mib.1.2.zip)

Here is the MIB file

PR #14256, opened 2022-08-29
Added support for Ubiquiti UFiber OLT

Martin22:

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 14256`


Martin22:
@Jellyfrog 
Why was this PR not included in the new version? I don't understand that some tests don't pass, but it's a bit strange that others where there is only a different version of php pass, I see it more of a bug elsewhere than in the PR in question.

includes/discovery/ports/edgeosolt.inc.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+unset($port_stats);
+
+$sfp_stats = snmpwalk_group($device, 'ubntSfps', 'UBNT-UFIBER-MIB', 1, $sfp_stats);
+
+$index = 0;
+for ($index = 0; $index < count($sfp_stats); $index++) {
```
murrant, line 8: Can you change this to use foreach instead?

```php
foreach ($sfp_stats as $index => $stats) {
    $port_stats[$index + 1]['ifDescr'] = $stats['ubntSfpName'];
...
```

murrant (commented):
Looks much better, nice work.

Can you please use ubntSfpTable instead of ubntSfpEntry (for all instances of Entry)

Entry is structural, not intended for direct use.

includes/discovery/ports/edgeosolt.inc.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+unset($port_stats);
+
+$sfp_stats = snmpwalk_group($device, 'ubntSfps', 'UBNT-UFIBER-MIB', 1, $sfp_stats);
```
murrant, line 5:
```suggestion
$sfp_stats = snmpwalk_group($device, 'ubntSfps', 'UBNT-UFIBER-MIB', 1);
```

includes/polling/ports/os/edgeosolt.inc.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+unset($port_stats);
+
+$sfp_stats = snmpwalk_group($device, 'ubntSfps', 'UBNT-UFIBER-MIB', 1, $sfp_stats);
```
murrant, line 5:
```suggestion
$sfp_stats = snmpwalk_group($device, 'ubntSfps', 'UBNT-UFIBER-MIB');
```

Martin22: The detection didn't work before, but I found out that I need to use a different oid.  I reverted foreach and added an offset, without which even if I set unset to port_stats, some ports got the wrong id.

includes/discovery/ports/edgeosolt.inc.php
Martin22: > 

When using table instead of entry, nothing is detected.

Martin22:
> Looks much better, nice work.
> 
> Can you please use ubntSfpTable instead of ubntSfpEntry (for all instances of Entry)
> 
> Entry is structural, not intended for direct use.

When using table instead of entry, nothing is detected.

Martin22:
> Looks much better, nice work.
> 
> Can you please use ubntSfpTable instead of ubntSfpEntry (for all instances of Entry)
> 
> Entry is structural, not intended for direct use.

Changed. Thank you for your advice.

Jellyfrog (approved):

warren-lcf:
@Martin22 great job! is it possible to get the ONU's out as their own devices with associated RX/TX/Connection States? Being listed as sensors makes it hard to see RX/TX together for the same ONU

PR #14259, opened 2022-08-29
Add support to Datacom (Teracom Telematica S/A) device families

lhwolfarth:
- Enhancements:
datacom (DmSwitch)
- Basic support:
datacom-dm1200e
datacom-dm1205e
datacom-dm2500
datacom-dmos

#### Please note

#### Testers


lhwolfarth:
> Please add test data so we can ensure your change is not broken in the future. Read the docs to find out how: https://docs.librenms.org/Developing/os/Test-Units


https://github.com/librenms/librenms/pull/14259/commits/9147fd93dde4fa157d3610093c85d2a6edb2ca2b

murrant:
Why have you created so many OS?  Wouldn't it be better to fill in the hardware field instead?

includes/definitions/discovery/datacom-dm1205e.yaml
```diff
@@ -0,0 +1,11 @@
+mib: SNMPv2-MIB
+modules:
+    os:
+     hardware: 
+             - SNMPv2-MIB::sysDescr.0
```
murrant, line 5: https://docs.librenms.org/Developing/os/Initial-Detection/#os-discovery
Use sysDescr_regex instead, what you have done here is call snmp to the device again, when you already have sysDescr

lhwolfarth:
> Why have you created so many OS? Wouldn't it be better to fill in the hardware field instead?

Not shure if there is other way

lhwolfarth:
> Why have you created so many OS? Wouldn't it be better to fill in the hardware field instead?

Not sure if there is other way to implement because each .yaml in fact refer to different OS despite they are from the same company. Each OS have it's own snmp tree structure so please confirm if in this case could be feasible use the hardware field anyway.

murrant:
They don't look that different from this standpoint, but I don't have time to look in depth.  And often when vendors use Broadcom silicon, they leave the fastpath mibs, but implement theirs as well.  (And sometimes later remove the broadcom mibs).

Sorry, but I need to make sure, as fixing os level mistakes is a pain.

includes/definitions/discovery/datacom-dm1205e.yaml
lhwolfarth: Just a question. How can I use the information fetched by the sysDescr in hardware field without call the snmp again? Sometimes I don't have the hardware description in another OID then I use the sysDescr.

PR #14261, opened 2022-08-30
Add instance id to error reports

Jellyfrog:
Also adds a basic support for "internal" settings that users can't override with config:set

#### Please note

#### Testers

### DIFF
--- a/app/Providers/ErrorReportingProvider.php
+++ b/app/Providers/ErrorReportingProvider.php
@@ -27,6 +27,7 @@ namespace App\Providers;
 
 use App\Logging\Reporting\Middleware\CleanContext;
 use App\Logging\Reporting\Middleware\SetGroups;
+use App\Logging\Reporting\Middleware\SetInstanceId;
 use ErrorException;
 use Facade\FlareClient\Report;
 use Facade\Ignition\Facades\Flare;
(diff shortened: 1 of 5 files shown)


Jellyfrog:
Flare still doesnt allow grouping by "user", waiting to see if its just a hiccup or a bug

app/Providers/ErrorReportingProvider.php
```diff
@@ -127,6 +127,17 @@ public function isReportingEnabled(): bool
         return true;
     }
 
+    public static function getInstanceId(): string
+    {
+        $uuid = Config::get('_reporting.error_uuid');
```
murrant, line 132: I don't think I like storing this in the config table, it is a lot more likely for the user to expose it.

An appropriate spot might be the callback table.

```php
$uuid = Callback::get('reporting-uuid')

if (empty($uuid)) {
    $uuid = Str::uuid();
    Callback::set('reporting-uuid', $uuid);
}
```

Jellyfrog: I think we should get rid of callback table. It's just a config value anyway, and it doesn't matter if a user actually changes it. They can as easily change callback table :)

app/Models/User.php
```diff
@@ -42,7 +43,10 @@ class User extends Authenticatable
 
     public function toFlare(): array
     {
-        return $this->only(['level', 'auth_type', 'enabled']);
+        $user = $this->only(['level', 'auth_type', 'enabled']);
+        $user['id'] = ErrorReportingProvider::getInstanceId();
```
Jellyfrog, line 47: This is this wrong place to inject it I've realized. Doesn't work with poller since we got no user instance then. Should be Middleware

app/Providers/ErrorReportingProvider.php
murrant: Changing it isn't the problem, exposing it is.

murrant (changes requested):
I think I want this in the callback table.

Feel free to rename the table if you want, like "instance_variables" or something.

PR #14262, opened 2022-08-31
Don't send zeros to Graphite for missing metrics

### LINKED ISSUE #14258: LibreNMS submits zeros to Graphite for missing metrics
### The problem

LibreNMS sends 0 to Graphite for missing datapoints, rather than omitting them. This causes issues as the validity of the data is no longer retained. Zero is a valid measurement and shouldn't be sent unless LibreNMS has actually collected a zero from the monitored device, so missing data can continue to be handled correctly in downstream data processing.

I believe this is a result of #7166 / #7176.

```
// Send zero for fields without values
if (empty($v)) {
   $v = 0;
}
```

Should just `continue` here instead to skip the metric.

### Output of ./validate.php

```txt
Component | Version
--------- | -------
LibreNMS  | 1.69-1-gbc02ab3
DB Schema | 2020_07_27_00522_alter_devices_snmp_algo_columns (188)
PHP       | 7.2.31
Python    | 3.6.8
MySQL     | 5.5.65-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
OpenSSL   | 
```

_No response_

_No response_

ktims:
Fix #14258 - Graphite logic checking for empty metrics sends them as zeros instead of skipping them

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Data/Store/Graphite.php
+++ b/LibreNMS/Data/Store/Graphite.php
@@ -113,9 +113,9 @@ class Graphite extends BaseDatastore
         }
 
         foreach ($fields as $k => $v) {
-            // Send zero for fields without values
-            if (empty($v)) {
-                $v = 0;
+            // Skip fields without values
+            if (is_null($v)) {
+                continue;
             }
             $metric = implode('.', array_filter([$this->prefix, $hostname, $measurement, $ms_name, $k]));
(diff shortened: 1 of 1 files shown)


murrant (approved):

murrant:
Thanks!

PR #14269, opened 2022-09-01
Add @signedGraphTag() and @signedGraphUrl() blade directives

murrant:
Takes either an array of graph variables or a url to a graph
Uses a signed url that is accessible without user login, embeds signature in url to authenticate access
See Laravel Signed Url for more details.
Adds Laravel route to graphs (does not change links to use it yet)
This relies on APP_KEY to do the signing, changing APP_KEY will invalidate old signed urls.

Are the directive names correct or should they be something like `@secureGraph` or `@externalGraph`?

- [x] documentation

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Util/Url.php
+++ b/LibreNMS/Util/Url.php
@@ -32,6 +32,7 @@ use Carbon\CarbonImmutable;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Str;
 use LibreNMS\Config;
+use Request;
 use Symfony\Component\HttpFoundation\ParameterBag;
 
 class Url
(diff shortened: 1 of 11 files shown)


Jellyfrog:
Nice, I since long wanted to use signed routes for this. 
Is there any further authorization of logged in users? Ie if the user have access to the device in question, or doesn't it work like that today?

murrant:
@Jellyfrog yes, that already exists.  The trick is, if $auth = true, then those user-access checks are skipped.

murrant:
Using a signed url will bypass user authentication. Basically, sharing a signed url gives anyone the with it access to the given graph.  If you need to secure access to graphs against people that previously might have had urls, you can rotate APP_KEY.

One way is to never give out relative date graph links, that way it only gives access to a specific time frame.

PR #14278, opened 2022-09-03
Fix a few Db* to Eloquent requests

PipoCanaja:
- VRP BGP code - move to eloquent and fix "nullable" columns issues
- DellOS10 - move to eloquent for a few requests and adapt Model
- Firebrick - move to eloquent for a few request
- Timos - move to eloquent for a few request

#### Please note

#### Testers


includes/discovery/bgp-peers/dell-os10.inc.php
```diff
@@ -40,7 +42,7 @@
         unset($bgpPeersCache);
 
         foreach ($bgpPeers as $vrfInstance => $peer) {
-            $vrfId = dbFetchCell('SELECT vrf_id from `vrfs` WHERE vrf_oid = ?', [$vrfInstance]);
+            $vrfId = Vrf::select('vrf_id')->where('device_id', '=', $device['device_id'])->where('vrf_oid', '=', $vrfInstance)->first();
```
murrant, line 45: There is a shortcut for this `->where('device_id', '=', $device['device_id'])`

`DeviceCache::getPrimary()->vrfs()->firstWhere('vrf_oid', $vrfInstance);`

(also = is the default operation, so you can omit it.

includes/discovery/bgp-peers/dell-os10.inc.php
```diff
@@ -52,7 +54,7 @@
                 // Setting it here avoids the code that resets it to null if not found in BGP4-MIB.
                 $bgpLocalAs = $value['os10bgp4V2PeerLocalAs'];
 
-                if (dbFetchCell('SELECT count(*) FROM `bgpPeers` WHERE device_id = ? AND bgpPeerIdentifier = ? AND vrf_id = ?', [$device['device_id'], $address, $vrfId]) == 0) {
+                if (BgpPeer::where('device_id', '=', $device['device_id'])->where('bgpPeerIdentifier', '=', $address)->where('vrf_id', '=', $vrfId)->count() < 1) {
```
murrant, line 57: `! DeviceCache::getPrimary()->bgppeers()->where('bgpPeerIdentifier', '=', $address)->where('vrf_id', '=', $vrfId)->exists()`

murrant: Better yet:

```php
BgpPeer::updateOrCreate([
    'device_id' => $device['device_id'],
    'bgpPeerIdentifier' => $address,
    'vrf_id' => $vrfId,
], $row);

```

murrant: If it won't create because the other fields don't have values, it should be handled by default values or nullable.

includes/discovery/bgp-peers/dell-os10.inc.php
```diff
@@ -97,8 +99,8 @@
         }
 
         // clean up peers
-        if (dbFetchCell('SELECT count(*) FROM `vrfs` WHERE `device_id` = ?', [$device['device_id']]) == 0) {
```
murrant, line 100: For cleanup, what I usually do is store ids of valid rows, then delete everything that doesn't have those ids.
The below code seems like a very long way to do that.

murrant:
In modern modules (of which the bgppeers is not), we try to take syncing away from individual OS code, that way there is less to do for them.  Looks like a lot of duplication of code there.  You probably don't want to convert bgppeers to a modern module, so it might make sense to make as few changes as possible.

murrant:
Using a signed url will bypass user authentication.  Basically, sharing a signed url gives anyone the with it access to the given graph.

includes/discovery/bgp-peers/dell-os10.inc.php
PipoCanaja: Thx for tip. Will apply it.

PipoCanaja:
> In modern modules (of which the bgppeers is not), we try to take syncing to the db away from individual OS code, that way there is less to do for them. Looks like a lot of duplication of code there. You probably don't want to convert bgppeers to a modern module, so it might make sense to make as few changes as possible.

Right. The idea was to keep current structure of the code, mainly written via "duplicate and adapt". I played my part of this game :) 
And because I am not (yet, who knows) confident with my "eloquent" skills, I sticked with simple ones for now.

includes/discovery/bgp-peers/dell-os10.inc.php
PipoCanaja: It slightly changes the behaviour for the '+' or '.' beeing typed ... But it can be achieved with an 'else' statement.

murrant: Yes indeed.  There is also a ModuleModelObserver which you can call like this ModuleModelObserver::observe(BgpPeer::class);  which will install an event listener to output +, -, and .

tests/module_tables.yaml
```diff
@@ -36,7 +36,7 @@ ports:
         order_by: ports.ifIndex, ports.ifDescr, ports.ifName
 cipsec-tunnels:
     ipsec_tunnels:
-        excluded_fields: [tunnel_id, device_id]
+        excluded_fields: [local_addr, tunnel_id, device_id]
```
PipoCanaja, line 39: local_addr is either "private" (so not really possible to do any evaluation on it) or the IP address of the snmpsimd (which can change, so no evaluation must be done). Now it is excluded.

PipoCanaja:
Hi @librenms/reviewers 
This one is ready for review. It will be continued with timos and a few others, but one step at a time :)

SantiagoSilvaZ:
This PR is ready for version 22.10.0, what's left to merge?

PipoCanaja:
> This PR is ready for version 22.10.0? what's left to merge?

Hi @PipoCanaja 
Unfortunately, the test system is broken somehow : test are running OK locally but the result in the auto-test in github is different. I cannot understand what's going on, so until somebody finds it, we are stuck here :(

SantiagoSilvaZ (approved):

murrant (approved):

PR #14285, opened 2022-09-04
Remove unused modules that users may have set

murrant:
Block wifi and mib modules as they no longer exist.
Remove file that should not exist.

#### Please note

#### Testers


LibreNMS/Config.php
```diff
@@ -445,6 +445,10 @@ private static function processConfig()
         self::deprecatedVariable('poller_modules.cisco-sla', 'poller_modules.slas');
         self::deprecatedVariable('oxidized.group', 'oxidized.maps.group');
 
+        // remove unused settings
+        self::forget('poller_modules.wifi');
+        self::forget('poller_modules.mib');
```
Jellyfrog, line 450: Can't we do this in a migration instead? So wasteful here :(

murrant: No, because with these it is very likely to be set in config.php.

The other option is an is file check, but that is bad because the module still "runs" if we do the check where it is included.

PR #14296, opened 2022-09-05
Dont set mempool percent to null

Jellyfrog:
`normalizePercent()` can return null, `setMempoolPercAttribute()` applies `round()` and throws deprecation warning 

#### Please note

#### Testers

### DIFF
--- a/app/Models/Mempool.php
+++ b/app/Models/Mempool.php
@@ -63,7 +63,6 @@ class Mempool extends DeviceRelatedModel implements Keyable
         $this->mempool_used = $used * $this->mempool_precision;
         $this->mempool_free = $free * $this->mempool_precision;
         $percent = $this->normalizePercent($percent); // don't assign to model or it loses precision
-        $this->mempool_perc = $percent;
 
         if (! $this->mempool_total) {
             if (! $percent && $percent !== 0.0) {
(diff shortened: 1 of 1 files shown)


Jellyfrog:
> Hmm, I remember some very specific semantics around null here. I fixed this further up (in the model) I think.

The only difference is in the scenario:
if the input `$percent` is null, and `mempool_total` is not set,
then the models `mempool_perc` will not be set, compared to old behaviour where it sets it = round(null) = 0.0

murrant:
We can't always calculate mempool_perc.  Sometimes we only have used amount and not total.

murrant:
That should fix it.

murrant (approved):

Jellyfrog:
No, because the error is caused by line 67, that sends null to :
```php
    public function setMempoolPercAttribute($percent)
    {
        $this->attributes['mempool_perc'] = round($percent);
    }
```

murrant:
> No, because the error is caused by line 67, that sends null to :
> 
> ```
>     public function setMempoolPercAttribute($percent)
>     {
>         $this->attributes['mempool_perc'] = round($percent);
>     }
> ```

That is not the current code, this is:
    public function setMempoolPercAttribute($percent)
    {
        $this->attributes['mempool_perc'] = is_numeric($percent) ? round($percent) : null;
    }

Jellyfrog:
hmm im not sure that commit was the best, schema doesnt allow null

murrant:
I noticed that, feel free to revert.

PR #14300, opened 2022-09-05
Routeros - Fix displaying distance in charts

Martin22:

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 14300`


LibreNMS/OS/Routeros.php
```diff
@@ -378,7 +378,7 @@ public function discoverWirelessDistance()
                 '60G: Sta > ' . $entry['mtxrWl60GStaRemote'],
                 $entry['mtxrWl60GStaDistance'],
                 1,
-                1000
+                100000
```
murrant, line 381: Interesting, considering the MIB says this is defined in meters not centimeters :D

LibreNMS/OS/Routeros.php
```diff
@@ -378,7 +378,7 @@ public function discoverWirelessDistance()
                 '60G: Sta > ' . $entry['mtxrWl60GStaRemote'],
                 $entry['mtxrWl60GStaDistance'],
```
murrant, line 379: This should be divided by the divisor I think.

LibreNMS/OS/Routeros.php
Martin22: The graph shows kilometres and not metres, so the MIB is saying it wrong. The connection is not really 18 km.

Before changes:
image

After changes:
image

murrant: Yes, the graph data is scaled to km for consistency.

Speaking of consistency, do all devices/firmwares report this value the same?

Martin22: > Yes, the graph data is scaled to km for consistency.

I understand, that's not a problem.

> 
> Speaking of consistency, do all devices/firmwares report this value the same?

Tested on the first firmware with 60G support, the middle and the latest, the same value is returned everywhere. Also tested Cube 60G, WAP 60G, RBLHG 60G (no longer sold) and the same values are returned everywhere.

murrant: Thanks for confirming.   Mikrotik isn't the best at snmp (but also not the worst)

Martin22:
@murrant 
Why was this PR not included in the new version? I don't understand that some tests don't pass, but it's a bit strange that others where there is only a different version of php pass, I see it more of a bug elsewhere than in the PR in question.

Jellyfrog (approved):

PR #14302, opened 2022-09-05
Ray2 - Oid repair at ber

Martin22:

#### Please note

#### Testers

### DIFF
--- a/tests/data/ray_ray.json
+++ b/tests/data/ray_ray.json
@@ -712,7 +712,7 @@
                     "sensor_deleted": 0,
                     "sensor_class": "ber",
                     "poller_type": "snmp",
-                    "sensor_oid": ".1.3.6.1.4.1.33555.1.3.2.11..0",
+                    "sensor_oid": ".1.3.6.1.4.1.33555.1.3.2.11.0",
                     "sensor_index": "0",
                     "sensor_type": "ray",
                     "sensor_descr": "BER",


includes/definitions/discovery/ray.yaml
```diff
@@ -37,7 +37,7 @@ modules:
             data:
                 -
                     oid: ber
-                    num_oid: '.1.3.6.1.4.1.33555.4.3.2.11.{{ $index }}'
+                    num_oid: '.1.3.6.1.4.1.33555.1.3.2.11..{{ $index }}'
```
PipoCanaja, line 40: You cannot have 2 dots in an OID. So this line must be wrong. Could you fix it  ?

PipoCanaja (changes requested):

includes/definitions/discovery/ray.yaml
```diff
@@ -37,7 +37,7 @@ modules:
             data:
                 -
                     oid: ber
-                    num_oid: '.1.3.6.1.4.1.33555.4.3.2.11.{{ $index }}'
+                    num_oid: '.1.3.6.1.4.1.33555.1.3.2.11.{{ $index }}'
```
PipoCanaja, line 40: Looks better for the OID. But looking at the MIB, it seems that a divisor is necessary here:
`1.3.6.1.4.1.33555.1.3.2.11: Bit Error Rate (BER) multiplied by 10^9. `

PipoCanaja:
Whenever the divisor is corrected, you just need to run again the "save-test-data.php" script and you'll be good to go.

includes/definitions/discovery/ray.yaml
Martin22: I don't think so. As far as I have seen graphs of other radio links, there was no divider used and the graphs showed the same values if there was a ber.

Martin22: The value is not divisible and is reported correctly.
https://www.techtarget.com/whatis/definition/bit-error-rate-BER

PipoCanaja: All right. Then you just need to generate again the JSON file and you should be good to go.

PipoCanaja (approved):
LGTM

PR #14317, opened 2022-09-07
 UHP VSAT modems

loopodoopo:
Created support for UHP VSAT modems. 

image
image
image
image

#### Please note

#### Testers


tests/data/uhp.json
```diff
@@ -0,0 +1 @@
+null
```
PipoCanaja, line 1: Hi @loopodoopo 
The file was not generated correctly. Please re-do it.
Thx

PipoCanaja (approved):

tests/data/uhp.json
loopodoopo: This works fine: php ./scripts/collect-snmp-data.php -h 172.19.121.17 -v ""

but php ./scripts/collect-snmp-data.php -h 46 -v "" does not produce anything?

includes/definitions/uhp.yaml
```diff
@@ -0,0 +1,11 @@
...
+text: 'UHP Networks VSAT Terminal'
+type: wireless
+icon: uhp
+discovery:
+    -
+        sysObjectID: 
+            - .1.3.6.1.4.1.8000
+mib_dir: uhp
+poller_modules:
+    netstats: false
```
PipoCanaja, line 11: Hi @loopodoopo 
Looks very good now. Just a question : Why did you disable netstats ? Does it fail with this device in any way ? If it is only because you don't need it, please leave it enabled in the YAML and disable it in your device config.

loopodoopo: These VSAT devices take a pretty long time to pol because of VSAT is slow... The UHP modems don't support netstats (tested), so disabling it saves 70 seconds in polling which is never available. I prefer to keep it disabled :)

includes/definitions/discovery/uhp.yaml
```diff
@@ -0,0 +1,718 @@
+mib: UHP-MIB
+modules:
+  os:
+    hardware: ENTITY-MIB::entPhysicalDescr.1 
+    version: swVersion
+    serial: serialNumber
```
Jellyfrog, line 6: According to test data these don't work, or are null, maybe worth investigating a bit

loopodoopo: Good catch, thank you. Fixed this.

PipoCanaja: Good @loopodoopo 
Now, the build fails because the test data is still broken compared to the code you fixed: 
You can check here: 
https://github.com/librenms/librenms/runs/8244891935?check_suite_focus=true

So just update the test data, and you'll be good to go.

includes/definitions/discovery/uhp.yaml
```diff
@@ -1,9 +1,9 @@
 mib: UHP-MIB
 modules:
   os:
-    hardware: ENTITY-MIB::entPhysicalDescr.1 
-    version: swVersion
-    serial: serialNumber   
+    hardware: RFC1213-MIB::sysDescr.0
```
murrant: sysDescr is already fetched, use sysDescr_regex to paste the hardware out of it.

loopodoopo: Should be good now.

PR #14318, opened 2022-09-07
PHPStan: Enable more checks

Jellyfrog:

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Elasticsearch.php
+++ b/LibreNMS/Alert/Transport/Elasticsearch.php
@@ -39,7 +39,7 @@ class Elasticsearch extends Transport
     {
         $es_host = '127.0.0.1';
         $es_port = 9200;
-        $index = strftime('librenms-%Y.%m.%d');
+        $index = date("\l\i\b\\r\\e\\n\m\s\-Y.m.d");
         $type = 'alert';
         $severity = $obj['severity'];
 
@@ -61,7 +61,7 @@ class Elasticsearch extends Transport
         }
 
         if (! empty($opts['es_index'])) {
-            $index = strftime($opts['es_index']);
+            $index = date($opts['es_index']);
         }
 
         $host = $es_host . ':' . $es_port . '/' . $index . '/' . $type;
@@ -222,7 +222,7 @@ class Elasticsearch extends Transport
                 [
                     'title' => 'Index Pattern',
                     'name' => 'es-pattern',
-                    'descr' => 'Elasticsearch Index Pattern',
+                    'descr' => 'Elasticsearch Index Pattern | Default: \l\i\b\\r\\e\\n\m\s\-Y.m.d | Format: https://www.php.net/manual/en/function.date.php',
                     'type' => 'text',
                 ],
                 [
--- a/LibreNMS/Proc.php
+++ b/LibreNMS/Proc.php
@@ -260,6 +260,7 @@ class Proc
      */
     public function isRunning()
     {
+        /* @phpstan-ignore-next-line */
         if (! is_resource($this->_process)) {
             return false;
         }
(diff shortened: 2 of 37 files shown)


Jellyfrog:
@murrant what do you say about this error ; https://github.com/librenms/librenms/pull/14318/files#file-librenms-device-processor-php-L145

```
        // if no processors found, check OS discovery (which will fall back to HR and UCD if not implemented
        if (empty($processors) && $os instanceof ProcessorDiscovery) {
```
> Instanceof between LibreNMS\OS and LibreNMS\Interfaces\Discovery\ProcessorDiscovery will always evaluate to true.

Since OS itself implements ProcessorDiscovery, it doesn't really get overridden by a custom os?

murrant:
@Jellyfrog all OS inherit the OS class. Some may implement custom processor polling, overriding that.

The error is correct, the if check can be removed.

LibreNMS/Proc.php
```diff
@@ -260,9 +260,6 @@ public function getStatus()
      */
     public function isRunning()
     {
-        if (! is_resource($this->_process)) {
```
Jellyfrog, line 263: It has this check in __construct which I guess is why phpstan complains:
```
        if (! is_resource($this->_process)) {
            throw new Exception("Command failed: $cmd");
        }
```

LibreNMS/Validations/Poller/CheckRedis.php
```diff
@@ -62,6 +62,7 @@ public function enabled(): bool
     private function redisIsAvailable(): bool
     {
         try {
+            /* @phpstan-ignore-next-line */
             Redis::ping();
```
Jellyfrog, line 66: > Call to an undefined static method Illuminate\Support\Facades\Redis::ping().
These needs to be stubbed in larastan, think its fixed in laravel9

app/Models/UserPref.php
```diff
@@ -99,9 +99,6 @@ protected function setKeysForSaveQuery($query)
     {
         /** @var array */
         $keys = $this->getKeyName();
-        if (! is_array($keys)) {
```
Jellyfrog, line 102: > Call to function is_array() with array will always evaluate to true.

The key is an array, not sure what it's suppose to check?

`protected $primaryKey = ['user_id', 'pref'];`

app/Providers/ErrorReportingProvider.php
```diff
@@ -46,6 +46,7 @@ class ErrorReportingProvider extends \Facade\Ignition\IgnitionServiceProvider
 
     public function boot(): void
     {
+        /* @phpstan-ignore-next-line */
         if (! method_exists(\Facade\FlareClient\Flare::class, 'filterReportsUsing')) {
```
Jellyfrog, line 50: >  Call to function method_exists() with 'Facade\\FlareClient\\Flare' and 'filterReportsUsing' will always evaluate to true.

Lets just ignore this, its for people who didn't update.

LibreNMS/Alert/Transport/Elasticsearch.php
```diff
@@ -61,7 +61,7 @@ public function contactElasticsearch($obj, $opts)
         }
 
         if (! empty($opts['es_index'])) {
-            $index = strftime($opts['es_index']);
+            $index = date($opts['es_index']);
```
Jellyfrog, line 64: This change might break elastic transport

murrant: Maybe strtotime() is what was supposed to be here?

Jellyfrog: No its to build a index name for elasticsearch

Jellyfrog: The default string is the change above it

murrant (approved):

Issue #10919, opened 2019-12-09
Cannot update device location attribute via API


cacheman:
Hi…

I am having trouble updating the ‘location’ field of a device using the update_device_field API call. This used to work, but it has been a while since I added new devices, and so I haven’t needed to update a device’s location in a while either.

<pre><code>-bash-4.2$ ./validate.php
====================================
Component | Version
--------- | -------
LibreNMS  | 1.58.1-21-gb555ede
DB Schema | 2019_04_22_220000_update_route_table (147)
PHP       | 7.2.13
MySQL     | 5.5.56-MariaDB
RRDTool   | 1.4.8
SNMP      | NET-SNMP 5.7.2
====================================
[OK]    Composer Version: 1.9.1
[OK]    Dependencies up-to-date.
[OK]    Database connection successful
[OK]    Database schema correct</code></pre>

When I run the following curl command to update the location field, it does say it worked, but it never updates the field:

<pre><code>-bash-4.2$ curl -X PATCH -d '{"field": ["location"], "data": ["Timbuktu, [16.7666, 3.0026]"]}' -H 'X-Auth-Token: 6da3sometoken6dasometoken6dasome' http://mylibrenms/api/v0/devices/some_network_switch
{
    "status": "ok",
    "message": "Device fields have been updated"
}</code></pre>

…and I get the following in the librenms.log file:

<pre><code>[2019-11-04 14:53:39] production.ERROR: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'location' in 'field list' (SQL: UPDATE `devices` set `location`=Timbuktu, [16.7666, 3.0026] WHERE `device_id`=228) (SQL: UPDATE `devices` set `location`=Timbuktu, [16.7666, 3.0026] WHERE `device_id`=228)#0 /opt/librenms/includes/html/api_functions.inc.php(1652): dbUpdate(Array, 'devices', '`device_id`=?', Array)
#1 [internal function]: update_device(Object(Illuminate\Http\Request), 'some_network_switch')
#2 /opt/librenms/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(32): call_user_func_array('update_device', Array)
#3 /opt/librenms/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(78): Illuminate\Container\BoundMethod::Illuminate\Container\{closure}()
[...]

PR #14325, opened 2022-09-08
API update_device, make location field work (as does location_id)

bp0:
The field 'location' can't be changed using API update_device_field, even when used with 'override_sysLocation'. It should not return success, with the "Device fields have been updated" message that is untrue.

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1922,7 +1922,7 @@ function update_device(Illuminate\Http\Request $request)
     // use hostname as device_id if it's all digits
     $device_id = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
     $data = json_decode($request->getContent(), true);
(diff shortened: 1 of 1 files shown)


murrant:
Or, you could make it work...

bp0:
It is possible to do, although it is not clear. First, the location must be added with add_location, then update_device_field can be used to set location_id on device. Only setting location field directly is not allowed, so this patch makes sense. It is fine if that's how they want it done, but the update_device_field should not return success when it can't actually set the field.

murrant:
Probably need to keep it here for compatibility.
Just:
$location = Location::firstOrCreate(['location' => $data['location']);
$data['location_id'] = $location->location_id;

murrant (approved):

bp0:
Right on. Thanks for fixing it

bp0:
This commit fixed https://github.com/librenms/librenms/issues/10919

PR #14330, opened 2022-09-09
Fix up graph percentile

murrant:
Show settings in webui
Line disabled by default
Use PERCENTNAN to discard missed polling intervals.
Remove "real_95th" or whatever, no idea what that is supposed to be.  All calculations are "real" now (aka use MAX instead of AVG)
Remove "Highest" from generic data graph, very pointless.
Remove a couple unused base graphs
Remove from multi_bits graph, it was too busted to fix there. (Feel free to send PR)
Never show percentile lines for previous intervals

#### Please note

#### Testers


murrant:
Uncertain about this PR.  It might go from broken to broken.  Maybe it would be clearer to mark this as an estimate.  The data consolidation breaks the 95th % calculation I think. The billing module is the way to get the real 95th.

Thoughts?

PR #14333, opened 2022-09-09
VLANs: Ironware workaround for Q-BRIDGE-MIB

avatar4d:
@murrant 

As originally noted in https://github.com/librenms/librenms/pull/14326:

The VLANS tab was populated with incorrect data for my Brocade ICX6450-48P. Specifically, the information coming from Q-BRIDGE-MIB::dot1qVlanCurrentEgressPorts is wrong. It appears to need the data in Q-BRIDGE-MIB::dot1qVlanStaticEgressPorts to be correct. 

This PR modifies q-bridge-mib.inc.php in order to test for ironware version 08.0.30uT313 (which is the last version supported on this device) when polling for vlan data. 

This change fixed the VLANs tab for my switch, but I am not sure if this is an acceptable approach and/or if someone else has a device running this version (or others) that can confirm the issue/fix. The only other thing I can think of would be to compare dot1qVlanCurrentEgressPorts with dot1qVlanStaticEgressPorts and if they differ prefer the latter (possibly again only for version 08.0.30uT313 of the firmware). 

Side note: I'm also having trouble creating testing data in the official docker container. I can create the snmprec in step one, but can't get step two to complete. 

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/vlans/q-bridge-mib.inc.php
+++ b/includes/discovery/vlans/q-bridge-mib.inc.php
@@ -36,7 +36,7 @@ if ($vlanversion == 'version1' || $vlanversion == '2') {
     // fetch vlan data
     $vlans = SnmpQuery::walk('Q-BRIDGE-MIB::dot1qVlanCurrentUntaggedPorts')->table(2);
     $vlans = SnmpQuery::walk('Q-BRIDGE-MIB::dot1qVlanCurrentEgressPorts')->table(2, $vlans);
-    // Q-BRIDGE-MIB::dot1qVlanCurrentX returns incorrect data for ICX64XX with v08.0.30uT313 
+    // dot1qVlanCurrentX returns incorrect data on ICX64XX with v08.0.30uT313
     if (empty($vlans) || $device['version'] == '08.0.30uT313') {
         // fall back to static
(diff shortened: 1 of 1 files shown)


avatar4d:
This change works to provide the correct data when running Discover from the capture page of the device and discover.php from the CLI. For some reason it reverts to the incorrect data when running rediscover in the device settings. It also reverts on its own periodically, I assume due to an automatic background process being run… I’m just not sure what or how it could possibly get bad data again when this change works when running discover the other two ways.

murrant:
Working inside the docker container can be tricky. You may wish to set up a linux environment for development.

Using the version as detection is the least desirable option.  Preferably, we look carefully at the shape of the data from the first walk.  To decide to walk the other.  Otherwise, walk both and combine them.

avatar4d:
Update: Placed a new ICX7150-12p into operation with the latest firmware that Ruckus recommends and it still returns the wrong vlan info. I don't see how the current code could be right for any ICX switch, unless maybe it was right at some point on a much older release of fastiron.

murrant:
I'm not sure this is still applicable, closing.  Please re-open if approriate.

PR #14335, opened 2022-09-10
API: update_device tweaks + fix location as only update field

bp0:
Changes the update_device function to simplify it, and allow setting only location. API allows either a single field or an array of fields. Location could not be set unless it was sent as array. Combine the two paths into one by converting the single field case to an array and then doing the regular work on an array, with a small check on the result to send the old single field output message for compatibility.

#### Please note

#### Testers


includes/html/api_functions.inc.php
```diff
@@ -1922,44 +1922,51 @@ function update_device(Illuminate\Http\Request $request)
...
-                return api_error(500, 'Device fields failed to be updated');
+            if ($field == 'location') {
+                $field = 'location_id';
+                $field_data = \App\Models\Location::firstOrCreate(['location' => $field_data])->id;
             }
+
+            $update[$field] = $field_data;
+        }
+        if (dbUpdate($update, 'devices', '`device_id`=?', [$device_id]) >= 0) {
+            return (count($data['field']) == 1)
```
murrant, line 1962: Should be `isset($data['field'][0])` instead of count.  count will throw an error in several situations.

murrant (approved):

murrant (approved):

murrant:
I think this or at least some of this was fixed by another PR.  If that is not the case, please rebase and reopen this PR.  Thanks :)

PR #14343, opened 2022-09-12
Ignore suppressed errors

Jellyfrog:
Prevents @-prefixed code from generating errors

#### Please note

#### Testers


Jellyfrog:
Another solution would be to start maintaining a list of errors to ignore

app/Providers/ErrorReportingProvider.php
```diff
@@ -154,7 +154,7 @@ public function isReportingEnabled(): bool
     public function handleError($level, $message, $file = '', $line = 0, $context = []): bool
     {
         // report errors if they are allowed
-        if ($this->errorReportingLevel & $level) {
+        if (error_reporting() & $level && $this->errorReportingLevel & $level) {
```
murrant, line 157: If we want to use this, we will have to look about changing: https://github.com/librenms/librenms/blob/master/LibreNMS/Util/Debug.php#L121

Jellyfrog: Hhmmmmm I think this is the wrong fix for it actually.

murrant:
Seems like this was close to correct.

I think returning false early might be better.

And figuring out if we can remove the error_reporting setting in the debug class might be needed too.

PR #14350, opened 2022-09-14
New SNMP Trap transport

ciroiriarte:
New snmptrap transport. Usually required to report events to northbound platforms. Compliant with [RFC2578](https://www.rfc-editor.org/rfc/rfc2578)

Enterprise Number is not random :)

59372
  Tigo Technology Center
    TTC Cloud Delivery Team

[Reference](https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers)

Hope to not mess the pull request this time.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Snmptrap.php
+++ b/LibreNMS/Alert/Transport/Snmptrap.php
@@ -25,6 +25,8 @@
 namespace LibreNMS\Alert\Transport;
 
 use LibreNMS\Alert\Transport;
+use LibreNMS\Config;
+
 
 class Snmptrap extends Transport
 {
@@ -45,7 +47,7 @@ class Snmptrap extends Transport
         $trapdefinition = $this->config['snmptrap-definition'];
         $pdu = $this->config['snmptrap-pdu'];
         $community = $this->config['snmptrap-community'];
-        $binary = $this->config['snmptrap-path'];
+	$binary = implode([Config::get('snmptrap')]);
         $mibdir = $this->config['mib-dir'];
 
         /**
--- a/LibreNMS/Alert/Transport/Snmptrap.php
+++ b/LibreNMS/Alert/Transport/Snmptrap.php
@@ -22,153 +22,151 @@
  ** @license GPL
  **/
 
-
 namespace LibreNMS\Alert\Transport;
 
 use LibreNMS\Alert\Transport;
 
 class Snmptrap extends Transport
 {
-	/**
-	 * @param  array  $obj
-	 * @param  array  $opts
-	 */
-	public function deliverAlert($obj, $opts)
-	{
-		$host = $this->config['snmptrap-destination-host'];
-		if (! empty($this->config['snmptrap-destination-port']) ) {
-			$port = $this->config['snmptrap-destination-port'];
-		} else {
-			/* Default SNMP trap port */
-			$port = '162';
-		}
-		$transport = $this->config['snmptrap-transport'];
-		$trapdefinition = $this->config['snmptrap-definition'];
-		$pdu = $this->config['snmptrap-pdu'];
-		$community = $this->config['snmptrap-community'];
-		$binary = $this->config['snmptrap-path'];
-		$mibdir = $this->config['mib-dir'];
+    /**
+     * @param  array  $obj
+     * @param  array  $opts
+     */
+    public function deliverAlert($obj, $opts)
+    {
+        $host = $this->config['snmptrap-destination-host'];
+        if (! empty($this->config['snmptrap-destination-port'])) {
+            $port = $this->config['snmptrap-destination-port'];
+        } else {
+            /* Default SNMP trap port */
+            $port = '162';
+        }
+        $transport = $this->config['snmptrap-transport'];
+        $trapdefinition = $this->config['snmptrap-definition'];
+        $pdu = $this->config['snmptrap-pdu'];
+        $community = $this->config['snmptrap-community'];
+        $binary = $this->config['snmptrap-path'];
+        $mibdir = $this->config['mib-dir'];
 
+        /**
+         * @param  array  $obj
+         */
+        return $this->contactSnmptrap($binary, $mibdir, $transport, $host, $port, $community, $trapdefinition, $pdu, $obj);
+    }
 
-		/**
-		 * @param array $obj
-		 */
-		return $this->contactSnmptrap($binary,$mibdir,$transport,$host,$port,$community,$trapdefinition,$pdu,$obj);
-	}
+    /**
+     * Returns if the call was successful
+     *
+     * @return bool
+     */
+    private function contactSnmptrap(string $binary, string $mibdir, string $transport, string $host, string $port, string $community, string $trapdefinition, string $pdu, array $obj)
+    {
+        $binary_opts = '';
+        switch ($pdu) {
+            case 'TRAPv2':
+                 $binary_opts = '-v 2c';
+                break;
+            case 'INFORM':
+                $binary_opts = '-v 2c -Ci';
+                break;
+            default:
+                echo 'This should not happen!!!';
+                break;
+        }
 
-	/**
-	 * Returns if the call was successful
-	 * @return bool
-	 */
-	private function contactSnmptrap(string $binary,string $mibdir,string $transport,string $host,string $port,string $community,string $trapdefinition,string $pdu,array $obj)
-	{
-		$binary_opts = "";
-		switch ($pdu) {
-			case 'TRAPv2':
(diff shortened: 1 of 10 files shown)


ciroiriarte:
Sample taken with Wireshark:

image

In this case all OIDs were sent, with the template conditionals we can send only OID that can be populated.

ciroiriarte:
It's not clear to me what is styleci complaining about.

Jellyfrog:
Tabs or windows linebreaks I guess

ciroiriarte:
> Tabs or windows linebreaks I guess

Oh god....

image

LibreNMS/Alert/Transport/Snmptrap.php
```diff
@@ -0,0 +1,173 @@
...
+            $port = $this->config['snmptrap-destination-port'];
+        } else {
+            /* Default SNMP trap port */
+            $port = '162';
+        }
+        $transport = $this->config['snmptrap-transport'];
+        $trapdefinition = $this->config['snmptrap-definition'];
+        $pdu = $this->config['snmptrap-pdu'];
+        $community = $this->config['snmptrap-community'];
+        $binary = implode([Config::get('snmptrap')]);
```
murrant, line 49: Why wrap in an array then implode?  Are you trying to cast to string?

LibreNMS/Alert/Transport/Snmptrap.php
```diff
@@ -0,0 +1,173 @@
...
+                echo 'This should not happen!!!';
+                break;
+        }
+
+        $msgsingle = preg_replace('~\R~', ' ', $obj['msg']);
+
+        putenv('SNMP_PERSISTENT_FILE=/tmp/snmpapp.conf.$USER');
+
+        //exec('/usr/bin/echo'
+        exec($binary
```
murrant, line 83: This allows cli injection.
Try using Symfony Process class (other examples can be found in source) and send an array to it.
(it also allows you to specify env vars)

LibreNMS/Alert/Transport/Snmptrap.php
```diff
@@ -0,0 +1,173 @@
...
+        return [
+            'config' => [
+                [
+                    'title' => 'Destination host',
+                    'name' => 'snmptrap-destination-host',
+                    'descr' => 'Hostname or IP of the host that will receive the trap.',
+                    'type' => 'text',
+                ],
+                [
+                    'title' => 'Destination port',
```
murrant, line 111: I think most of these settings should have a default value. (Either here or in the backend when they are accessed)

murrant (changes requested):

LibreNMS/Alert/Transport/Snmptrap.php
ciroiriarte: It's already considered for port, it should be defined in a different way?. You see MIB::event as something that should have a default?, something else?

LibreNMS/Alert/Transport/Snmptrap.php
ciroiriarte: Looking at Signal.php, is something like this what you expect?:

`    public function contactSignal($obj, $opts)
    {   
        exec($opts['path']
           . ' --dbus-system send'
           . $opts['recipient-type']
           . $opts['recipient']
           . ' -m ' . escapeshellarg($obj['title']));

        return true;
    }
`

LibreNMS/Alert/Transport/Snmptrap.php
ciroiriarte: yup, any advice is welcomed :)

LibreNMS/Alert/Transport/Snmptrap.php
murrant: In the webui, like this:

https://github.com/librenms/librenms/blob/master/LibreNMS/Alert/Transport/Mail.php#L67

LibreNMS/Alert/Transport/Snmptrap.php
murrant: No, something like this:
```php
use Symfony\Component\Process\Process;
 ...
$process = new Process(
    [Config::get('snmptrap'), '--dbus-system', 'send', $opts['recipient-type'], $opts['recipient'], '-m', $obj['title']],
    null,
    ['SNMP_PERSISTENT_FILE'=>'/tmp/snmpapp.conf.$USER']
);
$process->run();

return $process->isSuccessful();
```

LibreNMS/Alert/Transport/Snmptrap.php
murrant: `$binary = (string) Config::get('snmptrap');`


Additionally, it will be cast below in the cli call, so explicit cast is not needed.

LibreNMS/Alert/Transport/Snmptrap.php
ciroiriarte: testing with 'default', but I don't see that default presented in the UI. Also, should I remove the 'required' portion in the validation section?

murrant: @ciroiriarte yeah, I suspect default was only coded for checkboxes... just need to add it for input too.  If you feel up to it, you can add that, otherwise, it might take me a bit to get to it.  Changes will mostly be here: https://github.com/librenms/librenms/blob/master/includes/html/modal/edit_alert_transport.inc.php

ciroiriarte: Took a quick glance. To be honest I wouldn't know where to start.

LibreNMS/Alert/Transport/Snmptrap.php
```diff
@@ -0,0 +1,173 @@
...
+                $binary_opts = '-v 2c -Ci';
+                break;
+            default:
+                echo 'This should not happen!!!';
+                break;
+        }
+
+        $msgsingle = preg_replace('~\R~', ' ', $obj['msg']);
+
+        putenv('SNMP_PERSISTENT_FILE=/tmp/snmpapp.conf.$USER');
```
Jellyfrog, line 80: process might not be able to access /tmp/, use https://www.php.net/manual/en/function.sys-get-temp-dir.php

or `tmpfile()` if you dont need to persist itbetween requests

murrant:
https://www.iana.org/assignments/enterprise-numbers/?q=60652

murrant:
@ciroiriarte any plan to update this to use LibreNMS enterprise id?

ciroiriarte:
> @ciroiriarte any plan to update this to use LibreNMS enterprise id?

hello!, not really, unless it's a requirement to accept it.

The ID I used is valid for my organization.

I still didn't have time to:
- include execution safeguards as required by the project.
- extend MIB for specific traps, like a "port" trap (beyond the generic one)
- provide template examples for the non generic traps.

If changing the ID is required, it shouldn't be a problem, we would probably also exclude from the MIB file the structure for the other tool. Any thoughts on how would you like to structure the tree once excluding the other tool and replacing parent enterprise?

murrant:
Yes, I think using the LibreNMS enterprise id is a requirement.

ciroiriarte:
Any word on the tree structure you expect for the organization tree in
relation to this usecase?.

For the TTC implementation, I used a branch dedicated for alerting because
the same enterprise org is used for some LDAP schemas and also to avoid
hindery any future usecase (the enterprise ID assignment is expected to be
multipurpose, as my LDAP/SNMP example).

Also take into account that the way it was implemented allows the user to:
- use any available MIB file
- manipulate the mapping of parameters through a template.

This is usually desired because:
- The northbound system might require a specific formatting (hardcoded)
- Although you've built a beautiful and functional trap that would allow
the northbound to do amazing things like using comments from LibreNMS
acknowledgments to manipulate an incident ticket, that system might be
limited to a very basic format: the party in the receiving end might not
have the expertise to parse a new format or might be charged insane amounts
from third parties to do it.
- flexibility is king :)

Also, providing the templates with "full functionality" in the
documentation should abstract the blessed users from all those nuances.

Regards,
CI.-

murrant:
It can still allow custom MIBs.  Just the default would be one owned by LibreNMS.  One that outside systems can expect and implement support for.

I am not very knowledgeable on traps, but I would think a path like LibreNMS.Alerting.Trap or something would be ok. (Feel free to completely throw that out the window as I'm clueless about this)

laf:
@ciroiriarte Any plans to come back to this PR?

ciroiriarte:
Hello, would love to, most probably won't happen before Christmas since I probably will need to rebase it.

laf:
No updates provided for a while, if any are then we can re-open this pull request.

PR #14353, opened 2022-09-14
Load device selection dynamically on Ports page

KayckMatias:
This pull request is to add the realtime search function for device filter on the ports page, the same function is present on other pages but not on this one.

After add:
2e9133bd-66ca-4791-bc1e-11a85ec901fb

2893360d-fe9e-4bca-be9b-af7efcab6f7f

#### Please note

#### Testers

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -217,7 +217,7 @@ LibreNMS contributors:
 - Jason Cheng <sanyu3u@gmail.com> (jasoncheng7115)
 - Daniel Baeza <doctoruve@gmail.com> (TheGreatDoc)
 - Thom Cleary <me@thomcat.rocks> (thomcatdotrocks)
-- Kayck Matias <kayckmatias@gmail.com> [KayckMatias](https://github.com/KayckMatias/)
+- Kayck Matias <kayckmatias04@gmail.com> [KayckMatias](https://github.com/KayckMatias/)
 
 Observium was written by:
 - Adam Armstrong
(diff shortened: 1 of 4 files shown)


ottorei (dismissed):

murrant (changes requested):
Nice improvement, but:
1. This leaves the old query(s) to fill the select box
2. This does not use the init_select2 helper
3. This does not use the newer select2 backend

KayckMatias:
> Nice improvement, but:
> 
> 1. This leaves the old query(s) to fill the select box
> 2. This does not use the init_select2 helper
> 3. This does not use the newer select2 backend (which is automatically used with init_select2)

So I should use init select2 to do this filtering and remove the old query, right?

KayckMatias:
> Nice improvement, but:
> 
> 1. This leaves the old query(s) to fill the select box
> 2. This does not use the init_select2 helper
> 3. This does not use the newer select2 backend (which is automatically used with init_select2)

I can use only select2 without ajax data? Since librenms itself already makes the query before loading the page? Or only init_select2?

KayckMatias:
can use only select2 without ajax data? Since librenms itself already makes the query before loading the page? Then I could remove the old query and use select2() ajax or only init_select2?

KayckMatias:
> Nice improvement, but:
> 
> 1. This leaves the old query(s) to fill the select box
> 2. This does not use the init_select2 helper
> 3. This does not use the newer select2 backend (which is automatically used with init_select2)

Updated to init_select2 and removed old query

murrant:
Nice work! I want this finished for the release today, so putting the last bit of polish on it.

Need to use "DeviceCache" to load devices (soon to be renamed to DeviceRepository)
Fix xss from injecting $vars['device_id'] into the page
Handle invalid device_id
Load select2 later, so target id exists
Load select2 on graphs page too

murrant (approved):

PR #14361, opened 2022-09-16
Update devices filter in alert-logs to init_select2

KayckMatias:
Added init_select2 to devices filter in alert-logs page
image

#### Please note

#### Testers

### DIFF
--- a/includes/html/common/alert-log.inc.php
+++ b/includes/html/common/alert-log.inc.php
@@ -54,13 +54,9 @@ $common_output[] = '<div class="panel panel-default panel-condensed">
                 </div>
             ';
 
-if (isset($_POST['device_id'])) {
-    $selected_device = '<option value="' . (int) $_POST['device_id'] . '" selected="selected">';
-    $selected_device .= htmlentities($_POST['hostname']) . '</option>';
-} else {
-    $selected_device = $device_id;
-    $_POST['device_id'] = $device_id;
-}
+$device = DeviceCache::get((int) $vars['device_id']);
+$device_selected = json_encode($device->exists() ? ['id' => $device->device_id, 'text' => $device->displayName()] : '');
+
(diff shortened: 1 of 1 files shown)


includes/html/common/alert-log.inc.php
```diff
@@ -54,13 +54,9 @@
...
 
-if (isset($_POST['device_id'])) {
-    $selected_device = '<option value="' . (int) $_POST['device_id'] . '" selected="selected">';
-    $selected_device .= htmlentities($_POST['hostname']) . '</option>';
-} else {
-    $selected_device = $device_id;
-    $_POST['device_id'] = $device_id;
-}
+$device = DeviceCache::get((int) $vars['device_id']);
+$device_selected = json_encode($device->exists() ? ['id' => $device->device_id, 'text' => $device->displayName()] : '');
```
murrant, line 58: `$device->exists()` -> `$device->exists`

KayckMatias: Ok, I will update, I put the same function you put when updating my pull, can you fix it there too? https://github.com/librenms/librenms/blob/fb489f61edb29ed391fb5e62b8a3e98d33a686f5/includes/html/pages/ports/list.inc.php#L20

There is also "$device->exists()"

murrant: ->exists() is for db queries.   If we have a loaded model, you should use ->exists.

murrant: @KayckMatias I already fixed it in the previous PR, that is why I know it was wrong here :D

murrant: Final update, I checked and I swear I made that change, but it was not in master.  Sending that small fix. Thanks.

murrant (approved):

PR #14370, opened 2022-09-19
Fix oxidized web requests unclosed connections or responding fast enough

fbouynot:

#### Please note

#### Testers

### DIFF
--- a/app/ApiClients/Oxidized.php
+++ b/app/ApiClients/Oxidized.php
@@ -36,9 +36,9 @@ class Oxidized extends BaseApi
 
     public function __construct()
     {
+        parent::__construct(90);
         $this->base_uri = Config::get('oxidized.url');
         $this->enabled = Config::get('oxidized.enabled') === true && $this->base_uri;
-        $this->timeout = 90;
     }
 
     /**
--- a/app/ApiClients/Oxidized.php
+++ b/app/ApiClients/Oxidized.php
@@ -36,7 +36,7 @@ class Oxidized extends BaseApi
 
     public function __construct()
     {
-        $this->timeout = 3;
+        $this->timeout = 90;
         $this->base_uri = Config::get('oxidized.url');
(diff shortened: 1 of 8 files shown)


fbouynot:
> We are using file_get_contents() for that? sigh. I think this will have to wait until after the 22.9.0 release.

What should we use? Curl? I could rewrite with curl if it's better / cleaner.

murrant:
If you want to rewrite, use the Http client. https://laravel.com/docs/9.x/requests#main-content

fbouynot:
Will require https://github.com/librenms/librenms/pull/14371 to be merged first to avoid timeout on config loading

app/ApiClients/Oxidized.php
```diff
@@ -36,6 +36,7 @@ class Oxidized extends BaseApi
 
     public function __construct()
     {
+        parent::__construct(90);
```
Jellyfrog, line 39: Think it's better we do as the option below, set the timeout value to what you want here, and set 3s in base

fbouynot: Hi, I'm not sure to understand if you agree or request a change, and if so, what change? (set the timeout instead of calling the parent contructor?)

fbouynot: @Jellyfrog is this what you asked for?

Jellyfrog: Yes. And remove the other constructs ofc :)

murrant:
@fbouynot Looks good overall, thanks!  Only minor thing is there are some unneeded string casts in there.

fbouynot:
> @fbouynot Looks good overall, thanks! Only minor thing is there are some unneeded string casts in there.

@Jellyfrog Thanks! It's fixed, I was confused with typing not working and casted instead.

Jellyfrog (approved):

murrant:
Nice!

PR #14344, opened 2022-09-13
Fix reference to Rule ID

ciroiriarte:
Documentation makes reference to $alert->rule which is actually empty. $alert->rule_id on the other hand is populated.

Unsure if this came from a change, or it never worked. Documenting here so people don't chase ghost variables when building templates.

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [x] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [x] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.
- [x] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #14374, opened 2022-09-20
Populate the alert rule field with the builder json

murrant:
That way it isn't empty. (which is unexpected)

#14344 

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/RunAlerts.php
+++ b/LibreNMS/Alert/RunAlerts.php
@@ -188,7 +188,7 @@ class RunAlerts
         $obj['uid'] = $alert['id'];
         $obj['alert_id'] = $alert['alert_id'];
         $obj['severity'] = $alert['severity'];
-        $obj['rule'] = $alert['rule'];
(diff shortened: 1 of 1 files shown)


LibreNMS/Alert/RunAlerts.php
```diff
@@ -188,7 +188,7 @@ public function describeAlert($alert)
         $obj['uid'] = $alert['id'];
         $obj['alert_id'] = $alert['alert_id'];
         $obj['severity'] = $alert['severity'];
-        $obj['rule'] = $alert['rule'];
+        $obj['rule'] = $alert['rule'] ?: json_encode($alert['builder']);
```
PipoCanaja, line 191: do we keep $obj['builder'] = $alert['builder']; line 187 then ?

murrant: @PipoCanaja yes, that is the array of the builder rule... not json.  Other option is to remove the "rule" field completely, but since we haven't removed legacy rules (especially from the collection).  It doesn't make much sense yet.

Issue #14246, opened 2022-08-25
ALERT RULES | "Match devices, groups and locations list" is not honoring "locations "


teixemf:
### The problem

Alert Rules are not matching against "locations". 
If a Location is chosen from the list, the Alert Rule ignores it and matches against any device.
This behavior is not observed if a Group or specific Devices are chosen on the list.

Setup:
1) Create an Alert Rule
2) Setup "Match devices, groups and locations list" to match a Location.

Expected behavior:
It is expected the alert rule to match only devices which are part of that specifc Location.

Result:
The alert rule matches against all devices.

### Output of ./validate.php

```txt
====================================
Component | Version
--------- | -------
LibreNMS  | 22.8.0
DB Schema | 2022_07_03_1947_add_app_data (244)
PHP       | 8.0.22
Python    | 3.10.5
Database  | MariaDB 10.5.13-MariaDB-1:10.5.13+maria~focal
RRDTool   | 1.7.2
SNMP      | 5.9.3
====================================

[OK]    Installed from the official Docker image; no Composer required
[OK]    Database connection successful
[OK]    Database Schema is current
[OK]    SQL Server meets minimum requirements
[OK]    lower_case_table_names is enabled
[OK]    MySQL engine is optimal
[OK]    
[OK]    Database schema correct
[OK]    MySQl and PHP time match
[OK]    Active pollers found
[OK]    Dispatcher Service is enabled
[OK]    Locks are functional
[OK]    No python wrapper pollers found
[OK]    Redis is functional
[WARN]  IPv6 is disabled on your server, you will not be able to add IPv6 devices.
[OK]    rrd_dir is writable
[OK]    rrdtool version ok
[WARN]  Updates are managed through the official Docker image
```

murrant:
Need more info such as generated sql and sql results and expected results.

electrocret:
I can confirm the issue exists.
Test Alert Created:
![image](https://user-images.githubusercontent.com/6815956/186721465-06effaf0-f9a7-4ead-a800-a560c2dc87c5.png)

Alert on device not at specified location:
![image](https://user-images.githubusercontent.com/6815956/186721786-d38dd7ed-2b03-4ff7-bf87-a222bcf5d008.png)

teixemf:
> Need more info such as generated sql and sql results and expected results.

Sorry but I am still trying to figure how LibreNMS works under the hood to provide proper feedback.
I can I help exactly?  Does the output of "alert.php -d" helps?

[...]

Issue #12921, opened 2021-05-26
Alert rules: locations match doesn't work


andrzejmaczka:
Adding location in match field doesn't take any effect.

andrzejmaczka:
Found missing part at poller.php output:

SQL[SELECT DISTINCT a.* FROM alert_rules a
        LEFT JOIN alert_device_map d ON a.id=d.rule_id AND (a.invert_map = 0 OR a.invert_map = 1 AND d.device_id = ?)
        LEFT JOIN alert_group_map g ON a.id=g.rule_id AND (a.invert_map = 0 OR a.invert_map = 1 AND g.group_id IN (SELECT DISTINCT device_group_id FROM device_group_device WHERE device_id = ?))
        LEFT JOIN alert_location_map l ON a.id=l.rule_id AND (a.invert_map = 0 OR a.invert_map = 1 AND l.location_id IN (SELECT DISTINCT location_id FROM devices WHERE device_id = ?))
        LEFT JOIN device_group_device dg ON g.group_id=dg.device_group_id AND dg.device_id = ?
        **WHERE a.disabled = 0 AND (
            (d.device_id IS NULL AND g.group_id IS NULL)
            OR (a.invert_map = 0 AND (d.device_id=? OR dg.device_id=?))
            OR (a.invert_map = 1  AND (d.device_id != ? OR d.device_id IS NULL) AND (dg.device_id != ? OR dg.device_id IS NULL))
        )**

should be :

WHERE a.disabled = 0 AND (
            (d.device_id IS NULL AND g.group_id IS NULL AND l.location_id IS NULL)
            OR (a.invert_map = 0 AND (d.device_id = ? OR dg.device_id = ? OR l.location_id = ?))
            OR (a.invert_map = 1 AND (d.device_id != ? OR d.device_id IS NULL ) AND ( dg.device_id != ? OR dg.device_id IS NULL ) AND (l.location_id != ? l.location_id IS NULL))
            )

what do you think?

lflinders:
Also seeing this same behavior.
Set an alerting rule and added locations in order to filter here:
![image](https://user-images.githubusercontent.com/70960998/120815193-8171c980-c547-11eb-947e-8107cb8b609d.png)
The rule started alerting based on the other variables for all locations.

murrant:
If you have a fix, please send a pull request for others :)  Thanks.

PR #14380, opened 2022-09-22
Alert map location fix

KayckMatias:
Fixed location filter in alert rules

This PR fixes issues #14246 and #12921

#### Please note

#### Testers


murrant (approved):
Looks good, thanks.  I started to convert the query to Fluent/Eloquent, but it was to complex without some more analysis.

murrant:
Generally it is better to have separate PRs for things, both look good.

ottorei:
I think this broke the filtering on the device page itself (device -> alerts -> alert history). The Ajax-query which loads the table to the page does not have any device_id associated on the device page but the query works when done from the alerts -> alert history page.

I managed to get my installation working with this dirty and quick "fix" (includes/html/common/alert-log.inc.php, L151): 
`device_id: \'' . htmlspecialchars($_POST['device_id'] ?? $device_id) . '\'`

murrant:
Send a fix @ottorei ?

ottorei:
> Send a fix @ottorei ?

Sure. Not the most elegant way to fix it but at least on my setup it seems to be working. #14460

KayckMatias:
In case what I sent doesn't work correctly? But after your PR worked again and with the location fix?

PR #14388, opened 2022-09-25
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,8 +4,15 @@ on:
   pull_request:
     branches: [master, main]
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   super-linter:
+    permissions:
+      contents: read # to fetch code (actions/checkout)
+      checks: write # to mark the status of each individual linter run
+
     name: Lint Code Base
     runs-on: ubuntu-latest
 
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -16,6 +16,9 @@ on:
       - 'doc/**'
       - 'mkdocs.yml'
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   publish:
     runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -4,8 +4,15 @@ on:
...
     branches: [master, main]
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   super-linter:
+    permissions:
+      contents: read # to fetch code (actions/checkout)
+      checks: write # to mark the status of each individual linter run
```
Jellyfrog, line 14: Phpstan also needs write permissions

sashashura: I have added `checks: write` according [to the comment](https://github.com/github/super-linter/blob/d47cecf31ccb100f515b9c729be24ef3f7c501b8/README.md?plain=1#L124) in super-linter action. I think phpstan is not an action and doesn't know how to call github api to make use of the permission, although GitHub should show the overall check for the job as a whole - failed or succeed. Please, let me know if you still think it is better to grant `checks: write` to the phpstan job.

Jellyfrog (approved):

PR #14408, opened 2022-09-29
Openmanage enterprise modular

jgelinas:
Add support for Dell OpenMange Enterprise Modular currently used in the MX7000 blade chassis.

#### Please note

#### Testers

OMEM-Generic
OMEM-Supported


Jellyfrog:
missing json part of the test data

mpikzink:
Nice work! Are you sure it's the right OID for idle Power? On your screenshot it's more than Max Power wich seems to be wrong.

jgelinas:
I believe it ends up being the power still available to the system, but is the documented oid in the most recent docs I can find.

https://www.dell.com/support/manuals/en-us/idrac9-lifecycle-controller-v5.x-series/om_10.2_snmp_idrac/power-information?guid=guid-d82beef6-c305-4b6a-9e5d-3bb246d2eeef&lang=en-us

mpikzink:
Ah THX for the Link. As I understand, dmmPowerWattsPeakUsage is just for the chassis and dmmPowerIdlePower also includes the the powered on Servers. Can we add something to the descr. That the sensor value is more clear for the rest?

jgelinas:
Is there a way to do mouse over extra descriptions for sensors? That would be neat. I think I was trying to keep similar naming to the drac cmc stuff that already exists for the M1000e chassis, but perhaps some better names could be used. From what I can tell it is:

dmmPowerIdlePower: Idle power still available
dmmPowerWattsPeakUsage: Most power it has drawn
dmmPowerWattsMinUsage: Least power it has drawn
dmmPowerWattsReading: Current power draw

mpikzink:
Interesting Idea! Actually I think we do not have a field which we could use to show for a mouse over extra descr. But LNMS is a community driven project. If you make a PR it will come to the project. I would like it!
If you also think there are better names, change it! You know the system and for shore you will do it better then everyone else. DELL namings are always confusing. :-)
In my opinion it should be more clear what is the difference between min and idle. And the first power sensor Chasis1 I also do not understand without context. 

I for myself would do some little changes but this is just my opinion and you should decide...
I would remove all "chasis 1" strings from description and move it to the $group variable. I also would remove the types (temp,power,...) From the description field it's already defined via $type. I think will cleanup the view a bit.

Then I would rename the Power Descr. I like this
Chasis 1 => current
Chasis 1 idle => Chasis + running slot's (min)
Chasis 1 Max Power => max

As last step I would think about the use of Power min and max. I would put the two values to the Chasis1 (current) sensor as Min and Max warn limits and not as seperate sensors.

PipoCanaja:
Hi @jgelinas 
I don't know what's wrong with the snmpsim setup you have. So I did run it on one of my test VMs.

@Jellyfrog Any idea of what's going wrong and how to fix it ?

Jellyfrog:
No idea sorry

Jellyfrog:
> I changed Max/Min Power to Peak/Min Usage, which hopefully makes more sense, and is more inline with the older M1000e chassis graphs. I prefer to keep names as close as possible to the older chassis so people can reuse any pattern matching for alerts or dashboards or whatever.
> 
> Those OIDs aren't useful as limits as they aren't static values they are something that is a point in time measurement and will change. Looks like idle power is a bit weird on the older chassis too, perhaps I will open a ticket with dell to fix those.
> 
> Anyways, I have no luck with creating json files and I'm not sure what magic @PipoCanaja did in my other ticket to create that, but would appreciate any assistance in making one for this too.
> 
> -bash-4.2$ ./scripts/save-test-data.php -o dell-ome-m -v dell-ome-m Starting snmpsim listening on 127.1.6.1:1161... Scanning "/opt/librenms/.snmpsim/variation" directory for variation modules... Directory "/opt/librenms/.snmpsim/variation" does not exist Scanning "/usr/snmpsim/variation" directory for variation modules... Directory "/usr/snmpsim/variation" does not exist Scanning "/usr/share/snmpsim/variation" directory for variation modules... Directory "/usr/share/snmpsim/variation" does not exist Scanning "/opt/librenms/.local/lib/python3.6/site-packages/snmpsim/variation" directory for variation modules... Directory "/opt/librenms/.local/lib/python3.6/site-packages/snmpsim/variation" does not exist --- SNMP Engine configuration SNMPv3 EngineID: 0x80001234567890abcdef1234567890abcdef1234567890abcde --- Data directories configuration SNMPv3 Context Engine ID: 0x80001234567890abcdef1234567890abcdef1234567890abcde Scanning "/opt/librenms/tests/snmpsim/" directory for *.dump, *.MVC, *.sapwalk, *.snmpwalk, *.snmprec data files... Configuring /opt/librenms/tests/snmpsim/arubaos-cx.snmprec controller SNMPv1/2c community name: arubaos-cx Traceback (most recent call last): File "/usr/local/bin/snmpsimd.py", line 1247, in snmpContext File "/usr/local/bin/snmpsimd.py", line 900, in configureManagedObjects agentName = contextName = md5(univ.OctetString(communityName).asOctets()).hexdigest() ValueError: unknown reasons
> 
> Failed to start Snmpsim. Scroll up for error.
> 
> chassis-compare

if you installed snmpd from a package manager, remove it and install via pip

jgelinas:
I grabbed the VM image and got snmpsim to produce the json from that.

Jellyfrog:
@jgelinas  you need to fix the broken tests. Try deleting the json file and regenerate it

jgelinas:
@Jellyfrog I've regenerated the snmprec and json files, but still get failures.

jgelinas:
I was thinking about this a bit, could this be failing because the enclosure has failover interfaces? The json looks like it might be doubled up on the ports with some having data and some not, presumably because it has never had the need to failover?

murrant:
Something in @jgelinas dev environment did not match the test environment.

murrant (approved):

PR #14423, opened 2022-10-04
Add Sensor Enum class

Jellyfrog:

#### Please note

#### Testers


Jellyfrog:
Untested for now

Jellyfrog:
Maybe should split the enum to `SensorClass` and `SensorUnit`

PipoCanaja:
> Maybe should split the enum to `SensorClass` and `SensorUnit`

Other question : Should we make the Unit "overridable" ? Like default value is the current one, but it becomes possible from code to change it ? 
Count sensorType would largely benefit from this, because this sensor has no unit by default ... Other sensors might as well use this new feature, whenever the unit to describe them is not universal. Would also allow to translates units (for instance, RPM could be translated). Or even to change scale (runtime, in minutes currently, would sometimes be more interesting in other units like seconds or days depending on the sensors), etc

murrant:
Feels a little the wrong way to me.
I think the enum values should be the name sensor type (aka class).  Then it should define an array and maybe static method to get the unit display names for the type.  (reminder that you can't use enum for array keys right now)

murrant:
The unit can be overridden in the graph definition @PipoCanaja

laf:
Is this still relevant @Jellyfrog ?

Jellyfrog:
I think it is. Sensors are defined all over the place today. Totalt forgot about it tho.

Jellyfrog:
Superseeded by https://github.com/librenms/librenms/pull/17571

PR #14424, opened 2022-10-04
Syslog sink over http for logstash (or any json source)

Bierchermuesli:
This is an attempt of another way how LibreNMS can receive syslog messages from Logstash or any json source. The difference and benefits to the [already known Logstash variant ](https://docs.librenms.org/Extensions/Syslog/#logstash) are the follow:

- better / easier integration in existing (complex) Logstash/Elastic setups
- no libreNMS components on Logstash server necessary
- Logstash and LibreNMS hosts can be different network zones
- Logstash can buffer and submit multiple syslogs in one HTTP POST
- http/json as lightweight protocol for transport over WAN
- reverse proxy/load sharing as option
- easy scale out options on LibreNMS side
- firewall / WAF friendly
- http auth, SSL etc (available on logstash side)
- other json sources will work as well
- ....? 

Sample schema:
```
  ┌──────┐
  │Device├─►┌───────────────────┐                ┌──────────────┐
  └──────┘  │Logstash Cluster   ├┬──────────────►│ElasticSearch ├┐
            │  RabbitMQ         ││               │ Cluster      ││
 ┌──────┬──►│    Filtering etc  ││ ───────┐      └┬─────────────┼│
 │Device│   └┬──────────────────┼│        │       └──────────────┘
 └──────┘    └───────────────────┘        ▼
                                      ~~~WAN~~~
                                          │
                                        ┌─┼─┐
                                        │┼┼┼│ LB / Firewall
                                        └─┼─┘
                                          │
                                          ▼
                         ┌────────────────────┐    ┌────────────────────┐
                         │LibreNMS Sink       ├┬──►│LibreNMS Master     │
                         │/api/v0/syslogsink/ ││   │ MariaDB            │
                         └┬───────────────────┼│   └────────────────────┘
                          └────────────────────┘
```
Minimal [Logstash](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-http.html) config
```
     output {
     ....
     		#feed it to LibreNMS
     		http {
     						http_method => "post"
     						url => "http://localhost/api/v0/syslogsink/
     						format => "json_batch"
     						retry_failed => false
                                                headers => ["X-Auth-Token","xxxxxxx-LibreNMSApiToken]


     		}
}
```
Test Data:
```
curl -L -X POST 'http://localhost:8000/api/v0/syslogsink/' -H 'X-Auth-Token: xxx' --data-raw '[   
    {
        "msg": "kernel: minimum Message",
        "host": "mydevice.fqdn.com"
    },
    {
        "msg": "Line protocol on Interface GigabitEthernet1/0/41, changed state to up",
        "facility": 23,
        "priority": "189",
        "program": "LINEPROTO-5-UPDOWN",
        "host": "172.29.10.24",
        "@timestamp": "2022-12-01T20:14:28.257Z",
        "severity": 5,
        "level": "ERROR"
    },
    {
        "msg": "kernel: a unknown host",
        "host": "unknown.fqdn.com"
    }
]'
```

Further thoughts:
- [x] ~For performance and simplify I decide to leave this 'module' standalone instead of the common LibreNMS API~. Done
- [ ]  A proper Documentation in https://docs.librenms.org/Extensions/Syslog/ and API
- Idea for further improvement:
  - deadman detection
  - statistics / graphs
  - stats / json output
  - [x] ~auth as as default (or everything disabled as default)~ API integraded
  - [x] ~no plain html / document output if GET is used~
  - [ ] pimp the `process_syslog()` function to validate the host exist
  - [ ] API route below `/logs/*` ? alltrought we should use can:admin

-> Thanks for any advice or inputs how i can  contribute

``````

#### Please note

#### Testers

### CODE REVIEW

**Comment by murrant** [comment]
Seems decent, but why ingest the logs into LibreNMS?

Elastic search is a superior form of storage for logs.  I would much rather see something similar to how Graylog works.

**Comment by Bierchermuesli** [comment]
1. centralized Alarm triggers 
   - often some Events are not handled by trap or visible trough polling. 
   - e.g. Authentication failure, Maxprefix, Severy-x in n-time,  etc.. - already available in the Alert rule collection
2. centralizing of information
   - side by side/coloration with Outages, Event Log, Syslog, Traps etc..
   - good for a quick view for simple NOC staff
 
Indeed Kibana/Grafana/Graylog can do it better. IMHO libreNMS do it all right for a simple view and we have everything in one tool.

In my case I  use external link for Kibana for long term or aggregation searches. I would set a short libreNMS log retention. Also with (my) ELK stack there is no alert handling - graylog would be an option but centraliced alert rule handling is a plus.

**Comment by murrant** [comment]
Sorry for the slow review. Reasoning seems sound.

This should be part of the API.  Current code allows unauthenticated database injection.
We do not allow any new legacy pages to be added.  All need be defined in the routes files.

Let me know if you need any further help/guidance.  Again, thanks for your submission.

**Review by murrant — CHANGES_REQUESTED** [review]
Move to API

**Comment by Bierchermuesli** [comment]
Thanks for the reply. I can understand the API thing but this adds some extra overhead.

Can you point me some API examples or similar existing use cases where I can try to adapt?

**Comment by Bierchermuesli** [comment]
Soo... I did some home work and I hopefully integrated this feature well into the libreNMS API. Comment above is updated with curl test data. let me know what you think .

**Comment by murrant** [comment]
Looks more reasonable.  Could use a refinement pass.

The main thing missing is documentation :)

File: includes/html/api_functions.inc.php [hunk 1]
```diff
@@ -2911,6 +2911,26 @@ function edit_service_for_host(Illuminate\Http\Request $request)
     return api_error(500, "Failed to update the service with id $service_id");
 }
 
+/**
+ * recieve syslog messages via json https://github.com/librenms/librenms/pull/14424
+ */
+function put_syslogsink(Illuminate\Http\Request $request)
+{
+    $json = json_decode($request->getContent(), true);
```
murrant, line 2919:
```suggestion
    $json = $request->json();
```

murrant: then below, you only need the loop:
```php
        foreach ($logs as $entry) {
            process_syslog($entry, 1);
        }
```

includes/html/api_functions.inc.php
```diff
@@ -2911,6 +2911,21 @@ function edit_service_for_host(Illuminate\Http\Request $request)
...
+function put_syslogsink(Illuminate\Http\Request $request)
+{
+    $json = $request->json();
+    $logs = array_is_list($json) ? $json : [$json];
+
+    foreach ($logs as $entry) {
+        process_syslog($entry, 1);
+    }
+
+    return api_success_noresult(200, 'Syslog recived: ' . count($json));
```
murrant, line 2926:
```suggestion
    return api_success_noresult(200, 'Syslog received: ' . count($logs));
```

Bierchermuesli:
@murrant  documentation is done. I tried to re-org the syslog page a bit. let me know what you think or what it is missing.

murrant (approved):
LGTM

PR #14433, opened 2022-10-10
Revert "Update radius auth to accept permissions attribute"

Martin22:
Reverts librenms/librenms#14382

This commit is very bad. Filter-ID is a string, not a number. It broke our login.
Please cancel and rework.


Jellyfrog (approved):

kimhaak:
Sorry if it broke..

Could you elaborate on how and why this broke? I tested it, and it does indeed work.

If you are sending filter id in your reply's with other values, then i could see how it may break. If you do not send filter-id - it behaves as normal, before this patch.

PR #14436, opened 2022-10-10
Check that device exists on update_device API-route

ottorei:
Checks that the device exists before making update queries on device fields using the API. Currently the API responds with "ok" status even when the device does not exist at all. The issue probably occurs because the update is done using `dbUpdate`-function and only checking that the return value is `>= 0`. 

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1907,7 +1907,7 @@ function update_device(Illuminate\Http\Request $request)
     $device_id = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
 
     // Check that the device exists before updating fields
-    if (Device::where('device_id', $device_id)->doesntExist()) {
+    if (! DeviceCache::get($device_id)->exists) {
         return api_error(404, 'Device does not exist');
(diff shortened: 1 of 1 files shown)


includes/html/api_functions.inc.php
```diff
@@ -1905,6 +1905,12 @@ function update_device(Illuminate\Http\Request $request)
     $hostname = $request->route('hostname');
     // use hostname as device_id if it's all digits
     $device_id = ctype_digit($hostname) ? $hostname : getidbyname($hostname);
+
+    // Check that the device exists before updating fields
+    if (Device::where('device_id', $device_id)->doesntExist()) {
```
murrant, line 1910: Use DeviceCache, because the device should already be loaded by the above code.

```suggestion
    if (! DeviceCache::get($device_id)->exists) {
```

ottorei: Thanks for the suggestion. Just tested this quickly and it seems to be working.

murrant:
I wonder if this will break some API interactions?

PR #14462, opened 2022-10-16
User bgp peer remote addr instead of identifier where possible

zappiehost:
Historically, the BGP "Peer IP" section was only showing the BGP "identifier" which the peer can set to any Ipv4 they wish (regardless if its a v4 or v6 session) Now some SNMP implementation dont return `peerRemoteAddr` properly, so what this PR does is simply try to use `bgpRemoteAddr` and if that isnt set in DB then it will fall back to `bgpPeerIdentifier`

Also small adjustment of variable names to clarify their purpose

#### Please note

#### Testers


murrant:
I would still like to see bgpPeerIdentifier somewhere.

PR #14466, opened 2022-10-17
Add BIRD2 BGP Peers application support

zappiehost:
BIRD BGP Daemon does not have any snmp support nor its own sub agents. There are some (very very very old) agentx subagents out there that are semi broken and dont fully work with BIRD, whats worse is that they are all BGP4-MIB only meaning you can only see IPv4 peers (and not IPv6 peers) 

This PR, skips the issues with vendor specific MIBs (and bypassing totally the limitation of IPv4 only peers like in BGP4-MIB) and simply queries BIRD directly via SNMPExtened (like any other LibreNMS application) and parsers that BGP information into the LibreNMS DB.

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Applications.md
+++ b/doc/Extensions/Applications.md
@@ -310,6 +310,30 @@ chmod +x /usr/lib/check_mk_agent/local/bind
 
 3. Set the variable 'agent' to '1' in the config.
 
+## BIRD2
+
+The BIRD Internet Routing Daemon (BGP) 
+
+Due to the lack of SNMP support in the BIRD daemon, this application extracts all configured BGP protocols and parses it into LibreNMS.
+This application supports both IPv4 and IPv6 Peer processing.
+
+### SNMP Extend
+1. Edit your snmpd.conf file (usually /etc/snmp/snmpd.conf) and add:
+```
+extend bird2 '/usr/bin/sudo /usr/sbin/birdc -r show protocols all'
+```
+
+2.  Edit your sudo users (usually `visudo`) and add at the bottom:
+
+```
+Debian-snmp ALL=(ALL) NOPASSWD: /usr/sbin/birdc
+```
+_If your snmp daemon is running on a user that isnt `Debian-snmp` make sure that user has the correct permission to execute `birdc`_
+3. Restart snmpd on your host
+
+The application should be auto-discovered as described at the top of the page. If it is not, please follow the steps set out under `SNMP Extend` heading top of page.
+
+
 ## Certificate
 
 A small python3 script that checks age and remaining validity of certificates
--- a/includes/polling/applications/bird2.inc.php
+++ b/includes/polling/applications/bird2.inc.php
@@ -1,0 +1,198 @@
+<?php
+
+use App\Models\BgpPeer;
+use App\Models\Device;
+use Carbon\Carbon;
+
+$name = 'bird2';
+
+if (!\LibreNMS\Config::get('enable_bgp')) {
+    echo PHP_EOL . $name . ': BGP is not enabled in config' . PHP_EOL;
+    return;
+}
+
+$birdOutput = snmp_get($device, 'nsExtendOutputFull.' . string_to_oid($name), '-Oqv', 'NET-SNMP-EXTEND-MIB');
+
+// make sure we actually get something back
+if (empty($birdOutput)) {
+    echo PHP_EOL . $name . ': has empty output' . PHP_EOL;
+    return;
+}
+
+// ========
+// Process the actual BIRD2 output
+$protocolsData = [];
+
+// Remove headers
+$birdOutput = trim(explode('Name       Proto      Table      State  Since         Info', $birdOutput, 2)[1]);
+$protocolSegments = explode("\n\n", $birdOutput);
+
+// Remove the first title
+unset($protocolSegments[0]);
+
+foreach ($protocolSegments as $protocolSegment) {
+    // Deal with the title first
+    $protocolSegmentParts = explode("\n", $protocolSegment, 2);
+    $titleParts = preg_split("/\s+/", $protocolSegmentParts[0], 5);
+
+    // make sure we only look at BGP protocols
+    if ($titleParts[1] !== 'BGP') {
+        continue;
+    }
+
+    $protocolData = [
+        'name' => $titleParts[0],
+        'type' => $titleParts[1],
+        'table' => $titleParts[2],
+        'protocol_state' => $titleParts[3],
+        'since' => preg_split("/\s+/", $titleParts[4], 3)[0] . ' ' . preg_split("/\s+/", $titleParts[4], 3)[1],
+    ];
+
+    // Deal with the rest of the body
+    $protocolBodys = preg_split("/^\s{2}([A-Z])/m", $protocolSegmentParts[1]);
+
+    // Loop through all BGP protocols
+    foreach ($protocolBodys as $protocolBody) {
+        // Deal with the BGP block
+        if (strpos($protocolBody, 'GP') === 0) {
+            foreach (explode("\n", 'B' . $protocolBody) as $protocolBodyLine) {
+                if (strpos($protocolBodyLine, ':') !== false) {
+                    $lineParts = explode(':', $protocolBodyLine, 2);
+                    $protocolData[str_replace(' ', '_', strtolower(trim($lineParts[0])))] = trim($lineParts[1]);
+                }
+            }
+
+            // Fix up the error string
+            if (isset($protocolData['last_error'])) {
+                // Trim the received
+                $protocolData['last_error'] = trim(str_ireplace('Received:', '', $protocolData['last_error']));
+            }
+        }
+
+        // Process the Ip channel (v4/v6)
+        $IpVersion = 4;
+        if (filter_var($protocolData['neighbor_address'], FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
+            $IpVersion = 6;
+        }
+
+        if (strpos($protocolBody, 'hannel ipv' . $IpVersion) === 0) {
+
+            foreach (explode("\n", 'C' . $protocolBody) as $protocolBodyLine) {
+                if (strpos($protocolBodyLine, ':') !== false) {
+                    $lineParts = explode(':', $protocolBodyLine, 2);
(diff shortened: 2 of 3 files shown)


PipoCanaja:
Hi @zappiehost 
I understand your usecase.
That triggers an issue: the APP you wrote is basically updating BGP data, instead of leaving it to the BGP existing module. But, BGP existing module is not supposed to need an agent/extend to poll the data. Tricky one :) 
- What happens if you enable BGP module while this Application is active ? 
- @librenms/reviewers : what do you think would be the best solution here ?

zappiehost:
> Hi @zappiehost I understand your usecase. That triggers an issue: the APP you wrote is basically updating BGP data, instead of leaving it to the BGP existing module. But, BGP existing module is not supposed to need an agent/extend to poll the data. Tricky one :)

So the issue (as you mentioned) is the any/all BGP processing is done with the expectations of SNMP Mibs, and the issue is that BIRD is installed on systems that are Linux/Unix which leads us to having only the "generic" (BGP4-MIB) queried from the `bgp.inc.php` polling libreNMS script. (since for example `CISCO-BGP4-MIB` is for cisco devices while `ARISTA-BGP4V2-MIB` is for Astria routers, so linux with BIRD gets only  `BGP4-MIB`)

As mentioned, `BGP4-MIB` has no IPv6 support which means we rule that our (even if we were to use some AgentX subagent for BIRD no the linux server (which can work. but again, IPv4 peers only) So this rules out even the option of trying to write a sub agent on the BIRD linux machine using say `CISCO-BGP4-MIB`...

So this leads me to getting the most independent and isolated with dealing with BIRD. via an application, this reduces the risks of adding/removing/adjusting code in the current BGP processing classes to cater for a non SNMP datasource (which BIRD doesnt nor will have according to their team) 

The draw backs I see with this PR is

1) More code to maintain being outside the BGP processing class, however the plus is it only affect one functions to BIRD so less risk there
2) LibreNMS "Applications" aren't meant to be like this, they are "meant" to have their own graphs and charts and stats while this is just a means to processes "SNMP Extension" data 

If there a better place (not an application) get the data processed via SNMP extend please let me know

murrant:
bird doesn't expose v6 data at all via snmp?

zappiehost:
> bird doesn't expose v6 data at all via snmp?

Bird (or even Bird2) doesnt have any concept of snmp what so ever, some other people have on the internet have created external AgentX scripts to manually parse bird output and populate it in a BGP4-MIB (these are half broken scripts and unmaintained) 

Now the issue is the BGP4-MIB has no concept of IPv6 (you will see the same issue with other BGP daemons that support BGP4-MIB only, like FRR: https://github.com/FRRouting/frr/issues/2350 and https://github.com/FRRouting/frr/issues/9020 )

murrant:
Is there any way to use BGP4V2-MIB?

murrant:
Read the links, what a nightmare.  That explains a lot.

zappiehost:
> Is there any way to use BGP4V2-MIB?

So this has a 'gotcha' also... *if* bird has say a custom AgentX subagent which implemented BGP4V2-MIB the issue will be that the way LibreNMS polls/discovers BGP devices is based on their OS and BIRD is going to be installed on a linux device rather than say "cisco" which means LibreNMS will try default MIB of `BGP4-MIB` so a fair little bit of a addition of code will also need to be added to the bgp discovery and polling to try BOTH `BGP4V2-MIB` and `BGP4-MIB` on all generic devices 

Example seen here: https://github.com/librenms/librenms/blob/e4fdbbd82d652cfb197fafb3552de7f59afde9d7/includes/polling/bgp-peers.inc.php#L58-L59 

I cant think of a "clean" way of solving this problem for any software BGP daemon (BIRD/Quagga/FRR) and to make matters worse non of these software BGP daemons have any snmp support that IPv6 able (or even SNMP at all) 

@murrant do you have any ideas in mind from 'big code restructure' to some 'small hack' that would essentially enable users have better BGP peering states for their software BGP routers?

murrant:
I was looking though LibreNMS' mibs and a lot of vendors seemed to use BGP4V2-MIB as a base for their mibs.  
Brocade for example is almost a verbatim copy of an older version of the draft just under the foundry oid.  
H3C just plucked out a couple of tables and prefixed them under their vendor oid.
Arista did the same, except they actually used the latest version of the mib and implement all of it.
Dell has two different copies of the rfc mib, with different prefixes (likely a relic of acquiring force10)
Juniper did the same prefix thing and added routing instance id.

Honestly a shame the RFC failed, because it looks like it was just used anyway for the most part with little tweaks/additionas/omissions.

murrant:
> @murrant do you have any ideas in mind from 'big code restructure' to some 'small hack' that would essentially enable users have better BGP peering states for their software BGP routers?

Yes, we have a new style of polling module in LiberNMS.  It makes it very easy to implement both standard mibs supported by many OS and vendor specific MIBs.  You can check the Mpls and Stp modules for an examples.

murrant:
They probably need to do what the vendors did and put it under their own enterprise oid.
Unless the IETF assigns a common root for the draft MIB, there isn't much you can do.

fcqpl:
Hi,
Great work!
No problems with bird2. I need only change line 153 from:
```
$bgpPeer->bgpPeerIdentifier = $protocol['neighbor_id'] ?: '0.0.0.0';
```
to:
```
$bgpPeer->bgpPeerIdentifier = $protocol['neighbor_address'] ?: '0.0.0.0';
```
Looks like bgpPeerIdentifier is "Peer address" in LibreNMS Routing table.

streamthing (commented):
Hmm.. i can't add commit here. Please change bgpPeerIdentifier line and merge it to LibreNMS.

doc/Extensions/Applications.md
```diff
@@ -310,6 +310,30 @@ chmod +x /usr/lib/check_mk_agent/local/bind
...
+### SNMP Extend
+1. Edit your snmpd.conf file (usually /etc/snmp/snmpd.conf) and add:
+```
+extend bird2 '/usr/bin/sudo /usr/sbin/birdc -r show protocols all'
+```
+
+2.  Edit your sudo users (usually `visudo`) and add at the bottom:
+
+```
+Debian-snmp ALL=(ALL) NOPASSWD: /usr/sbin/birdc
```
murrant, line 329: Might want to be a little more restrictive and specify arguments.

includes/polling/applications/bird2.inc.php
```diff
@@ -0,0 +1,192 @@
...
+                ];
+            }
+        }
+    }
+
+    $protocolsData[] = $protocolData;
+}
+
+// ---
+$deviceObj = Device::findOrFail($device['device_id']);
```
murrant, line 122:
```suggestion
$deviceObj = DeviceCache::getPrimary();
```
Object already exists, avoid fetching it from the database.

includes/polling/applications/bird2.inc.php
```diff
@@ -0,0 +1,192 @@
...
+    $bgpPeer = BgpPeer::firstOrNew([
+        'device_id' => $device['device_id'],
+        'bgpPeerRemoteAs' => $protocol['neighbor_as'],
+        'bgpLocalAddr' => $protocol['source_address'] ?: '0.0.0.0',
+        'bgpPeerRemoteAddr' => $protocol['neighbor_address'],
+    ]);
+
+    $bgpPeer->device_id = $device['device_id'];
+    $bgpPeer->astext = get_astext($protocol['neighbor_as']);
+    $bgpPeer->bgpPeerIdentifier = $protocol['neighbor_id'] ?: '0.0.0.0';
```
murrant, line 153: Commentor indicated there is an issue with neighbor_id, does it exist?

murrant (changes requested):

doc/Extensions/Applications.md
zappiehost: @murrant birdc (interactive shell to print out status) needs to be used here, not sure how to further add safety restrictions, what do you think?

includes/polling/applications/bird2.inc.php
zappiehost: I think there is some confusions between `neighbour address` and `neighbour id`

- neighbour address = This is the actual address used to establish the BGP session (most people care about this)
-  neighbour id = this is the `router-id` which is set as an arbitrary ipv4 address, from my understabnding, no one uses or cares about this, its a simple `id` in the form of an ipv4 address (even for an ipv6 session this router-id is used as a ipv4 address. 

Here I tried to remain consistent with the naming of the variables, `bgpPeerIdentifier` as the `ID` while `bgpPeerRemoteAddr` is the actual address to establish the BGP session. 

This is another reason why its less confusing to display the `remote address` and NOT `id` here: https://github.com/librenms/librenms/blob/master/includes/html/pages/routing/bgp.inc.php#L283C15-L283C15

Where it uses `peer_ident` which is **NOT** the peer's actual IPv4/IPv6 address but the `router-id` 


tl;dr with the current commit I tried to stay as consistant as possible with the current way librenms displays the peer address (even though I believe that its done wrong in `includes/html/pages/routing/bgp.inc.php`

zappiehost: @murrant On second thought I figured it will be actually to fix the bug on the display of the LibreNMS - see the additional commit here: https://github.com/librenms/librenms/pull/14466/commits/2d3210eee99812cc1e2187f198c4e0311cf794bd

murrant:
@zappiehost just need to apply the style fixes and this LGTM

zappiehost:
> @zappiehost just need to apply the style fixes and this LGTM

@murrant All done and StyleCI passes

murrant (approved):

netravnen:
Fyi, SNMP support has been put on the BIRD Feature Roadmap

https://gitlab.nic.cz/labs/bird/-/blob/master/doc/roadmap.md#L5-L6

> ### SNMP AgentX plugin for BIRD status export
> Allow for easier status monitoring.

https://gitlab.nic.cz/labs/bird/-/blob/master/doc/roadmap.md#L27-L29

> ### Consolidate protocol statistics
> Consolidate protocol statistics, make them useful for SNMP plugin and implement 'show XX stats' command.

https://gitlab.nic.cz/search?search=snmp&nav_source=navbar&project_id=6&group_id=2&search_code=true&repository_ref=master

netravnen:
> great news, it should be supported via our standard routing modules at that point instead of this ad-hoc method.

My own expectation is SNMP development will be targeted for with (or after) Bird 3 has been released.

PR #14476, opened 2022-10-18
Type API methods and properties

fbouynot:
Type API methods and properties

#### Please note

#### Testers


app/ApiClients/BaseApi.php
```diff
@@ -30,10 +30,9 @@
 
 class BaseApi
 {
-    protected $base_uri;
-    /** @var int */
-    protected $timeout = 3;
-    private $client;
+    protected ?string $base_uri;
```
Jellyfrog, line 33: this should not be nullable

app/ApiClients/BingApi.php
```diff
@@ -34,7 +34,7 @@ class BingApi extends BaseApi implements Geocoder
 {
     use GeocodingHelper;
 
-    protected $base_uri = 'http://dev.virtualearth.net';
+    protected ?string $base_uri = 'http://dev.virtualearth.net';
```
Jellyfrog, line 37: this should not be nullable

app/ApiClients/BingApi.php
```diff
@@ -34,7 +34,7 @@ class BingApi extends BaseApi implements Geocoder
...
     protected $geocoding_uri = '/REST/v1/Locations';
```
Jellyfrog, line 38: While at it, could you typehint all of the extra properties?

app/ApiClients/BaseApi.php
fbouynot: OK, I think I had error with it non-nullable, so I'll deal with them

app/ApiClients/BingApi.php
fbouynot: Sure, I was trying to achieve something simple first. Now it works, I'll do the same with the extra properties.

app/ApiClients/BaseApi.php
fbouynot: I don't understand how it can be related: https://github.com/librenms/librenms/actions/runs/3273454425/jobs/5385675803

Jellyfrog: Because oxidized.php line 37 sets it to null

Jellyfrog:
> What do I do with the getconfig error @Jellyfrog ? Do I add an exception in the neon file, or should I use something else, or should I use baseapi instead of guzzle?

Baseapi would be good, it probably happens because you typehinted guzzle for it, so it could check deeper into it.

app/ApiClients/GraylogApi.php
```diff
@@ -158,7 +146,7 @@ function ($address) {
         return $addresses->filter()->unique();
     }
 
-    public function isConfigured()
+    public function isConfigured(): bool
     {
         return isset($this->client->getConfig()['base_uri']);
```
murrant, line 151:
```suggestion
        return (bool) Config::get('graylog.server');
```

Prevent future bug

murrant:
Odd usage, we don't even need to call getConfig, I suggested a fix.

app/ApiClients/GraylogApi.php
```diff
@@ -120,7 +120,7 @@ public function buildSimpleQuery(?string $search = null, ?Device $device = null)
         return implode(' && ', $query);
     }
 
-    public function getAddresses(Device $device): array
+    public function getAddresses(Device $device): Illuminate\Support\Collection
```
murrant: This is missing the leading \ like `\Illuminate\Support\Collection`  You can only omit the leading slash when the file is in the global namespace.

murrant (approved):

PR #14376, opened 2022-09-21
Adding Devices field in get_alert_rules and list_alert_rules API request

geg347:
Hello dear LibreNMS Community,

Here I added the ability to retrieve the Devices field in get_alert_rules and list_alert_rules API request either if it is a device, group or location.
Before that nothing was returned.
If one of these 3 variables are set they will be shown as a array of integer.
If there is no such variable nothing will be returned.

I know that the SQL request is quite complex and can maybe add latency on simple query.
I also added some PHP code to reset the variable if empty and to convert the string returned in SQL to an array of integer.

Do not hesitate to challenge me if you see some possible improvements.

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [ ] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [ ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.
- [ ] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #14481, opened 2022-10-20
Adding device's field in get_alert_rule and list-alert-rules API function (new)

geg347:
Hello dear LibreNMS community,

Sorry for posting twice but here is the same PR as #14376 (I created it on master by mistake)
I am trying to solve the pipeline error by rebasing from upstream master and not creating a PR on my fork master.

Previous text : 

Here I added the ability to retrieve the Devices field in get_alert_rules and list_alert_rules API request either if it is a device, group or location.
Before that nothing was returned.
If one of these 3 variables are set they will be shown as a array of integer.
If there is no such variable nothing will be returned.

I know that the SQL request is quite complex and can maybe add latency on simple query.
I also added some PHP code to reset the variable if empty and to convert the string returned in SQL to an array of integer.

I split the SQL in several lines. Tell me if you really want to use eloquent, but I will need help on this.

Have a nice day!

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1161,11 +1161,40 @@ function list_alert_rules(Illuminate\Http\Request $request)
     $sql = '';
     $param = [];
     if ($id > 0) {
-        $sql = 'WHERE id=?';
+        $sql = 'WHERE `a`.`id`=?';
         $param = [$id];
     }
 
-    $rules = dbFetchRows("SELECT * FROM `alert_rules` $sql", $param);
+    $rules = dbFetchRows("SELECT `a`.*,
+        GROUP_CONCAT(DISTINCT `d`.`device_id`) AS `devices`,
+        GROUP_CONCAT(DISTINCT `g`.`group_id`) AS `groups`,
+        GROUP_CONCAT(DISTINCT `l`.`location_id`) AS `locations`
+        FROM `alert_rules`
+        AS `a` LEFT JOIN `alert_device_map`
+        AS `d` ON `a`.`id`=`d`.`rule_id` LEFT JOIN `alert_group_map`
+        AS `g` ON `a`.`id`=`g`.`rule_id` LEFT JOIN `alert_location_map`
+        AS `l` ON `a`.`id`=`l`.`rule_id` $sql
+        GROUP BY `a`.`id`", $param);
+
+    $i = 0;
+    foreach ($rules as $rule) {
+        if ($rule['devices'] == null) {
+            unset($rules[$i]['devices']);
+        } else {
+            $rules[$i]['devices'] = array_map('intval', explode(',', $rules[$i]['devices']));
+        }
+        if ($rule['groups'] == null) {
+            unset($rules[$i]['groups']);
+        } else {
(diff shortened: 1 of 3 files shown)


includes/html/api_functions.inc.php
```diff
@@ -1161,11 +1161,40 @@ function list_alert_rules(Illuminate\Http\Request $request)
     $sql = '';
     $param = [];
     if ($id > 0) {
-        $sql = 'WHERE id=?';
+        $sql = 'WHERE `a`.`id`=?';
         $param = [$id];
     }
 
-    $rules = dbFetchRows("SELECT * FROM `alert_rules` $sql", $param);
```
murrant, line 1168: Eloquent goes something like this (but for all relationships), but you need #14500
```php
$rules = AlertRule::with(['devices' => function ($query) {
    return $query->select('devices.device_id');
}])->get();
```

geg347: OK thanks for your answer. I tried this and in the API I have some strange results : 
For a request with 2 devices and a group I get
`"devices": [
				{
					"device_id": 1,
					"pivot": {
						"rule_id": 25,
						"device_id": 1
					}
				},
				{
					"device_id": 2,
					"pivot": {
						"rule_id": 25,
						"device_id": 2
					}
				}
			]`
It misses the groups. I assume that it is normal as we did not put the group in the Eloquent query.

But worse, instead of only returning the device ids, it returns these pivot thing..

A working answer from my old SQL way is : 
`"devices": [
				1,
				2
			],
			"groups": [
				6
			]`

And if there is all devices selected it return an empty array as devices. It may break something if we reinject this Json code into a PUT request.

murrant: Sorry, I'll give you more complete code.  I only gave you a bit to get you started.

murrant: To make this as clean as possible, I had to use a json resource:
`php artisan make:resource AlertRule`
```php
    public function toArray($request)
    {
        $rule = parent::toArray($request);
        $rule['devices'] = $this->devices->pluck('device_id')->all();
        $rule['groups'] = $this->groups->pluck('id')->all();
        $rule['locations'] = $this->locations->pluck('id')->all();

        return $rule;
    }
```

Then, the api method becomes just:
```php
function list_alert_rules(Illuminate\Http\Request $request)
{
    $id = $request->route('id');

    $rules = \App\Http\Resources\AlertRule::collection(
        \App\Models\AlertRule::when($id, fn($query, $id) => $query->where('id', $id))
        ->with(['devices:device_id', 'groups:id', 'locations:id'])->get()
    );

    return api_success($rules->toArray($request), 'rules');
}
```

geg347:
Thanks a lot it is working like a charm.
I tried to update alerts with the output from list_alert_rules through the edit_rule API request and everything is working. If a devices or groups or locations exist, they are kept. If everything is empty it stays with the All_devices.

I added the resource's file but for #14500 It is already merged after this PR (I hope it will be kept back)

geg347:
Ah I guess I need to add the changes from #14500 to pass the pipeline... Or rebase to master.

PR #14492, opened 2022-10-21
Graphing cleanup

murrant:
Started to fix why some settings weren't applied in embedded email graphs and ended up with a massive cleanup.

After this is merged and confirmed to not break anything, I will seek and destroy the following from the graphing code:

- $_GET
- global variables
- includes/html/graphs/common.inc.php

#### Please note

#### Testers


LibreNMS/Data/Graphing/GraphParameters.php
```diff
@@ -0,0 +1,265 @@
...
+
+        return $options;
+    }
+
+    public function __toString(): string
+    {
+        return implode(' ', $this->toRrdOptions());
+    }
+
+    public function parseAtTime(string|int $time): int
```
Jellyfrog, line 196: This function/logic should probably live in a date/time class?

murrant: @Jellyfrog sounds good to me

Jellyfrog:
Nice!
It failed as I suspected due to test being done on php7. Think it's rather safe to simply disable it for now

fcqpl:
@murrant looks like this commit broke charts on dashboard.
How to disable title bar from this charts?
Also font on timeline is smaller than before this commit and unreadable.
image

murrant:
@fcqpl update your install

murrant:
As for the font size, hmmm maybe an off by one.

fcqpl:
> @fcqpl update your install

Thanks, this two commit fixed our issue.

https://github.com/librenms/librenms/commit/55d81fde1eee8654f3225acb3cbebd227d7f6513
https://github.com/librenms/librenms/commit/e864ed7795d80fe46c62bafbbb52f9b0e129d7a6

PR #14522, opened 2022-10-25
More Replacement Capabilities for sensor index computation

PipoCanaja:
Allow more replacement capabilities for YAML sensor uindex computation. Goal is to use any precached OID as uniq string to create our index, instead of the SNMP index (in case this one is not stable).

UseCase : 
A proxy is exposing a table of categories, including categoryName, categoryNumberBlockedPages, but the table expands as soon as a new category receives hits, and the index is not stable in time.  This PR will allow to use categoryName is the index, resulting in a stable LibreNMS sensor index, and a more readable RRD file name.

MNI yaml OS definition was wrong and this patch made it visible. I added a fix (probably not the most aesthetic one but at least all sensors are now visible). 

#### Please note

#### Testers


includes/discovery/functions.inc.php
```diff
@@ -968,7 +968,15 @@ function discovery_process(&$valid, $os, $sensor_class, $pre_cache)
...
-                    $uindex = str_replace('{{ $index }}', $index, isset($data['index']) ? $data['index'] : $index);
+                    $uindex = $index;
+                    if (isset($data['index'])) {
+                        if (Str::contains($data['index'], '{{')) {
+                            $uindex = trim(YamlDiscovery::replaceValues('index', $index, null, $data, $pre_cache));
+                        } else {
+                            $uindex = str_replace('{{ $index }}', $index, $data['index']);
+                        }
+                    }
+
```
murrant, line 979:
```suggestion
                    $uindex = isset($data['index']) 
                        ? YamlDiscovery::replaceValues('index', $index, null, $data, $pre_cache) 
                        : $index;

```

replaceValues will cover the {{ $index }} too.

murrant: And also, the str_replace would never replace anything anyway because of the Str::contains check.

PipoCanaja: We cannot use replaceValues alone because replace value will also match an OID without `{{ $` and ` }}`. Which breaks a lot of indexes. We must avoid to call replaceValues in that case. 

The cases are : 
- uindex defaults to $index.
- if we have an index defined in the YAML
  - if it is safe to use replaceValues on it, then we do it
  - if not we use this YAML defined index without any replacement

so the code could be simplified to : 
```
                    $uindex = $index;
                    if (isset($data['index'])) {
                        if (Str::contains($data['index'], '{{')) {
                            $uindex = trim(YamlDiscovery::replaceValues('index', $index, null, $data, $pre_cache));
                        } else {
                            $uindex = $data['index'];
                        }
                    }
```

murrant: I forgot that does literal swapping, LGTM

murrant (approved):
LGTM

PR #14526, opened 2022-10-26
Remove Log::event

Jellyfrog:
Use the Eventlog class directly instead

#### Please note

#### Testers


Jellyfrog:
Blocked by https://github.com/librenms/librenms/pull/14546

tests/OSModulesTest.php
```diff
@@ -165,13 +165,6 @@ public function dumpedDataProvider()
 
     private function stubClasses(): void
     {
-        $this->app->bind('log', function ($app) {
-            $mock = \Mockery::mock('\App\Facades\LogManager[event]', [$app]);
-            $mock->shouldReceive('event');
-
-            return $mock;
-        });
-
```
Jellyfrog, line 174: @murrant what's left here is this, I suppose we could stub it out in the same way as astext?

murrant: You can't mock static methods.

app/Models/Eventlog.php
```diff
@@ -38,17 +38,33 @@ class Eventlog extends DeviceRelatedModel
...
+     * Log events to the event table
+     *
+     * @param  string  $text
+     * @param  Device|int|null  $device
+     * @param  string  $type
+     * @param  int  $severity
+     * @param  int|string|null  $reference
      */
-    public static function log($text, $device = null, $type = null, $severity = Alert::INFO, $reference = null)
+    public static function _log($text, $device = null, $type = null, $severity = Alert::INFO, $reference = null): void
```
murrant, line 67: This should be non-static

app/Models/Eventlog.php
```diff
@@ -59,6 +75,10 @@ public static function log($text, $device = null, $type = null, $severity = Aler
             'username'  => (class_exists('\Auth') && Auth::check()) ? Auth::user()->username : '',
         ]);
 
+        if (is_numeric($device)) {
```
murrant, line 78: Hmm, I think I purposely required a Device model.  I let the event_log() function convert int to Device before calling Log::event()

Jellyfrog: log_event probably does

murrant (approved):

PR #14530, opened 2022-10-27
Change axiscam to type camera from

sorano:
Change axiscam to type camera from appliance.

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/axiscam.yaml
+++ b/includes/definitions/axiscam.yaml
@@ -1,6 +1,6 @@
 os: axiscam
 text: 'AXIS Network Camera'
-type: appliance
+type: camera
 icon: axis
 mib_dir: axis
 discovery:


PipoCanaja:
Hi @sorano 
Camera is not an official type currently. This must be discussed before anything else

Jellyfrog:
Yes we should device on device categories. Any suggestions and/or lists or standards we could follow?

murrant:
@PipoCanaja what makes a type an official type?  Are they defined somewhere?

sorano:
Well,
I was actually not aware that a camera type existed until I added some ipolis cameras and they defaulted to camera type. 
https://github.com/librenms/librenms/blob/e864ed7795d80fe46c62bafbbb52f9b0e129d7a6/includes/definitions/ipolis.yaml#L3

Vivotek also has the same:
https://github.com/librenms/librenms/blob/e864ed7795d80fe46c62bafbbb52f9b0e129d7a6/includes/definitions/vivotek.yaml#L3

Then I manually tried to change my Axis cameras to the same but Camera is not a visible type in the drop-down. Maybe the drop-down is what defines "official type" ?
image

murrant:
I think you are correct.  You can set any type in is yaml, but users can only set predefined types...  Which seems kind of silly.

PR #14531, opened 2022-10-27
Update Radius auth to accept permissions attributes

kimhaak:
This is a new approach to PR #14382

This enables the user to send a radius attribute called "Filter-ID" with a string that RadiusAuthorizer.php interprets to a valid userlevel, or ignores.

As this requires "Filter-ID" to be a predetermined string, it will not break login for users that have "Filter-ID" in use for other purposes, as the authenticator will ignore it and revert to default value defined in config.php.

Updated docs, and also gave a hint for using Radius Huntgroups in Freeradius if someone already uses the attribute, but also want to use the new functionality this PR gives.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Authentication/RadiusAuthorizer.php
+++ b/LibreNMS/Authentication/RadiusAuthorizer.php
@@ -33,7 +33,46 @@ class RadiusAuthorizer extends MysqlAuthorizer
 
         $password = $credentials['password'] ?? null;
         if ($this->radius->accessRequest($credentials['username'], $password) === true) {
-            $this->addUser($credentials['username'], $password, Config::get('radius.default_level', 1));
+            if ($this->userExists($credentials['username'])) {
+
+
+
+                //attribute 11 is "Filter-Id"
+                //Always set password change to 0 - password resides in AAA, not LibreNMS
+                //If attribute 11 is sent in reply after accept - update user
+                //If user exists - update, not add.
+                //If new user - add user with attribute value if present, or use default from config.
+                if ($this->radius->getAttribute(11)) {
+                    switch($this->radius->getAttribute(11)){
+                        case "lnms_admin":
+                            $attribute = 10;
+                            break;
+                        case "lnms_normal":
+                            $attribute = 1;
+                            break;
+                        case "lnms_globalRead":
+                            $attribute = 5;
+                            break;
+                        case "lnms_demo":
+                            $attribute = 11;
+                            break;
+                    }
+                }
+
+                if ($attribute) {
+                    $this->updateUser($this->getUserid($credentials['username']), $credentials['username'], $attribute, 0, '');
+                } else {
+                    $this->updateUser($this->getUserid($credentials['username']), $credentials['username'], Config::get('radius.default_level', 1), 0, '');
+                }
+            }
+
+            if (! $this->userExists($credentials['username'])) {
+                if ($attribute) {
+                    $this->addUser($credentials['username'], $password, $attribute, '', $credentials['username'], 0, '');
+                } else {
+                    $this->addUser($credentials['username'], $password, Config::get('radius.default_level', 1), '', $credentials['username'], 0, '');
+                }
+            }
 
             return true;
         }
--- a/doc/Extensions/Authentication.md
+++ b/doc/Extensions/Authentication.md
@@ -267,9 +267,24 @@ setsebool -P httpd_can_connect_ldap 1
 ## Radius Authentication
 
 Please note that a mysql user is created for each user the logs in
-successfully. User level 1 is assigned to those accounts so you will
-then need to assign the relevant permissions unless you set
-`$config['radius']['userlevel']` to be something other than 1.
+uccessfully. User level 1 is assigned by default to those accounts 
+unless radius sends a reply attribute with the correct userlevel. 
+
+You can change the default userlevel by setting
+`$config['radius']['userlevel']` to something other than 1.
+
+The attribute `Filter-ID` is a standard Radius-Reply-Attribute (string) that
+can be assigned a value which translates into a userlevel in LibreNMS. 
+
+The strings to send in `Filter-ID` reply attribute is *one* of the following:
+
+`lnms_normal` - Sets the value `1`, which is the normal user level.
+`lnms_admin` - Sets the value `5`, which is the administrator level.
+`lnms_globalRead` - Sets the value `10`, which is the global read level.
+`lnms_admin` - Sets the value `11`, which is the demo level.
+
+LibreNMS will ignore any other strings sent in `Filter-ID` and revert to default userlevel that is set in `config.php`.
+
 
 ```php
 $config['radius']['hostname']      = 'localhost';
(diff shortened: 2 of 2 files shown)


kimhaak:
@ottorei - Everything works as expected,style is ok, but the linter complains about my code.

Do i really need to declare $attribute first? It will only be created, if needed.

And, the` if ($attribute){}` statement would only validate to true, if $attribute exists....

I am not a very experienced coder, just enough to get by :P 

Tips are very welcome :P

doc/Extensions/Authentication.md
```diff
@@ -267,9 +267,24 @@ setsebool -P httpd_can_connect_ldap 1
 ## Radius Authentication
 
 Please note that a mysql user is created for each user the logs in
-successfully. User level 1 is assigned to those accounts so you will
-then need to assign the relevant permissions unless you set
-`$config['radius']['userlevel']` to be something other than 1.
+uccessfully. User level 1 is assigned by default to those accounts
```
PipoCanaja, line 270: s missing in `successfully`

kimhaak: Fixed :)

ottorei:
> @ottorei - Everything works as expected,style is ok, but the linter complains about my code.
> 
> Do i really need to declare $attribute first? It will only be created, if needed.
> 
> And, the` if ($attribute){}` statement would only validate to true, if $attribute exists....
> 
> I am not a very experienced coder, just enough to get by :P
> 
> Tips are very welcome :P

This might be because for certain inputs the attribute will never be defined at all. One way to fix this would be to use a default value on the switch statement or use a key-value lookup mapping using regular arrays https://www.php.net/manual/en/language.types.array.php. If using arrays, you might still want to have a default value defined if the key does not exist in the array. You could use something like `$value = array[key] ?? 1` to lookup the array but return 1 if the key does not exist.

kimhaak:
> 

Hm, i did not use the default in the switch statement, as i do not want a default action, as this logically would mean to override a value that is unknown, hence making the same problem as before.

I might set a "unknown" value for defautl and test for that in the if's further down.

ottorei:
> > 
> 
> Hm, i did not use the default in the switch statement, as i do not want a default action, as this logically would mean to override a value that is unknown, hence making the same problem as before.
> 
> I might set a "unknown" value for defautl and test for that in the if's further down.

Or try to get a known value using switch or array and default that to null. Then check if the value is not null before using it.

kimhaak:
> > > 
> > 
> > 
> > Hm, i did not use the default in the switch statement, as i do not want a default action, as this logically would mean to override a value that is unknown, hence making the same problem as before.
> > I might set a "unknown" value for defautl and test for that in the if's further down.
> 
> Or try to get a known value using switch or array and default that to null. Then check if the value is not null before using it.

I did not see your comment before i commited a new version. I fixed a misplacement of the whole attrib mapping block, it needs to go first, after auth has gone "true". Then i just declared $attribute as '' so it will be false, if mapping never sets an attribute value.

kimhaak:
> > > 
> > 
> > 
> > Hm, i did not use the default in the switch statement, as i do not want a default action, as this logically would mean to override a value that is unknown, hence making the same problem as before.
> > I might set a "unknown" value for defautl and test for that in the if's further down.
> 
> Or try to get a known value using switch or array and default that to null. Then check if the value is not null before using it.

Looks like we are good to go, test-wise.

I have also tested all variants of logins, feel free to do the same :)

kimhaak:
@Martin22 - Your problem with the previus update should be fixed in this version.

Martin22:
> @Martin22 - Your problem with the previus update should be fixed in this version.

Thank you. Looks okay.

ottorei:
@murrant What do you think about the predefined hardcoded role names? If in some near future the role based access policy is implemented, these would probably change?

murrant:
@ottorei yes, I agree.  One problem is they don't line up with the planned built-in role names (normal, global-read, admin)

Maybe have a consistent prefix like librenms_role_admin librenms_role_normal.  This would allow custom roles like librenms_role_csr.

murrant:
Correct it would need an update, but we don't want is to require users to change their configs or to handle two different sets of configs. :)  Thanks!

LibreNMS/Authentication/RadiusAuthorizer.php
```diff
@@ -33,7 +33,47 @@ public function authenticate($credentials)
...
+                //attribute 11 is "Filter-Id"
+            //Always set password change to 0 - password resides in AAA, not LibreNMS
+            //If attribute 11 is sent in reply after accept - update user
+            //If user exists - update, not add.
+            //If new user - add user with attribute value if present, or use default from config.
+
+            $attribute = '';
+            if ($this->radius->getAttribute(11)) {
+                switch ($this->radius->getAttribute(11)) {
+                        case 'lnms_admin':
```
murrant, line 46: Rename these attributes to something like this, you could even use a match statement.
I renamed $attribute to $level  to make it easier to understand what the variable is and included the default level to save the if ($attribute) statements below.

```php
$level = match($this->radius->getAttribute(11)) {
    'librenms_role_admin' => 10,
    'librenms_role_normal' => 1,
    'librenms_role_global-read' => 5,
    default => Config::get('radius.default_level', 1)
};

if ($this->userExists($credentials['username'])) {
    // existing user, update the level only
    $user = \App\Models\User::find($this->getUserid($credentials['username']));
    $user->level = $level
    $user->save();
} else {
    $this->addUser($credentials['username'], $password, $level, '', $credentials['username'], 0, '');
}
```

murrant: A couple of edits there after thinking about the code a bit.

murrant: @kimhaak yes, there is no way to make the code support custom roles until LibreNMS does.

I just want the user configuration to be compatible so we only need to change the code to support RBAC.

murrant: @kimhaak yes, I use the return that way for readability.  else is hard to read (maybe not one, but the more you have the harder it gets to follow).

https://levelup.gitconnected.com/how-you-can-avoid-using-else-in-your-code-871197a1adbc

LibreNMS/Authentication/RadiusAuthorizer.php
```diff
@@ -33,7 +33,47 @@ public function authenticate($credentials)
...
+                }
+            }
+
+            if (! $this->userExists($credentials['username'])) {
+                if ($attribute) {
+                    $this->addUser($credentials['username'], $password, $attribute, '', $credentials['username'], 0, '');
+                } else {
+                    $this->addUser($credentials['username'], $password, Config::get('radius.default_level', 1), '', $credentials['username'], 0, '');
+                }
+            }
```
murrant, line 76:
```suggestion
            // attribute 11 is "Filter-Id", apply and enforce user role (level) if set

            $filter_id_attribute = $this->radius->getAttribute(11);
            $level = match($filter_id_attribute) {
                'librenms_role_admin' => 10,
                'librenms_role_normal' => 1,
                'librenms_role_global-read' => 5,
                default => Config::get('radius.default_level', 1)
            };

            // if Filter-Id was given and the user exists, update the level
            if ($filter_id_attribute && $this->userExists($credentials['username'])) {
                            $user = \App\Models\User::find($this->getUserid($credentials['username']));
                            $user->level = $level;
                $user->save();
                return true;
            }

            $this->addUser($credentials['username'], $password, $level, '', $credentials['username'], false);
```

LibreNMS/Authentication/RadiusAuthorizer.php
murrant: Fixed up my hand written code so you an commit it and test below.

PR #14545, opened 2022-10-31
XSS sanitize pwrstatd application script inputs

bnerickson:
Sanitizing and validating input for the pwrstatd application's script's data.  We expect integers for all data except the serial number which is a string.

A bad actor could, in theory, construct a pwrstatd application that returned a serial number with "<script>..." in its name to arbitrarily execute on the pwrstatd application page.  filter_var has been added to negate this possibility.

#### Please note

#### Testers

### DIFF
--- a/includes/polling/applications/pwrstatd.inc.php
+++ b/includes/polling/applications/pwrstatd.inc.php
@@ -31,15 +31,22 @@ $rrd_def = RrdDefinition::make()
 
 $metrics = [];
 foreach ($pwrstatd_data as $data) {
-    $sn = $data['sn'];
-    $mruntime = $data['mruntime'];
-    $pcapacity = $data['pcapacity'];
-    $pload = $data['pload'];
-    $voutput = $data['voutput'];
-    $vrating = $data['vrating'];
-    $vutility = $data['vutility'];
-    $wload = $data['wload'];
-    $wrating = $data['wrating'];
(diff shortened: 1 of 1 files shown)


includes/polling/applications/pwrstatd.inc.php
```diff
@@ -31,15 +31,22 @@
...
-    $wrating = $data['wrating'];
+    $sn = is_string($data['sn']) ? filter_var($data['sn'], FILTER_SANITIZE_STRING) : null;
+
+    if (is_null($data['sn'])) {
+        echo PHP_EOL . $name . ':' . ' Invalid or no psu serial number found.' . PHP_EOL;
+
+        continue;
+    }
+
+    $mruntime = is_int($data['mruntime']) ? $data['mruntime'] : null;
```
murrant, line 42: I don't think any of these is_int() calls are necessary (and it might not even work).  All these are stored in rrd (numeric only) or database fields that are numeric.

bnerickson: $data is the json data pulled from the script output before its ingested into an rrd file, and, after monitoring overnight, these changes do work.  I agree, however, that they aren't necessary.  My rationale is to filter the data received at the edge rather than let rrd ingestion figure it out.  For instance, I'm not sure what rrd does if it gets passed a string number such as "123", but it should be an invalid input IMO; so I set it to null.

murrant (approved):

PR #14564, opened 2022-11-02
Fix PTOPO mib in module discovery-protocols

### LINKED ISSUE #14474: "Error discovering discovery-protocols module - Juniper - after update to PHP8.1"
### The problem

After needed update PHP to ver 8.1 discovery protocol cannot find any Junipers. Seems to be that code is not fully functional with new PHP. Did I saw that right? Any clues?

### Output of ./validate.php

```txt
===========================================
Component | Version
--------- | -------
LibreNMS  | 22.9.0-58-gae3925b09 (2022-10-16T18:15:17+02:00)
DB Schema | 2022_09_03_091314_update_ports_adsl_table_with_defaults (246)
PHP       | 8.1.11
Python    | 3.6.8
Database  | MariaDB 10.3.28-MariaDB
RRDTool   | 1.7.0
SNMP      | 5.8
===========================================
```

_No response_

```txt
production.ERROR: Error discovering discovery-protocols module for x.x.x.x. TypeError: trim(): Argument #1 ($string) must be of type string, array given in /opt/librenms/includes/discovery/discovery-protocols.inc.php:320
[2022-10-17T19:03:04.981623+02:00] production.ERROR: trim(): Argument #1 ($string) must be of type string, array given {"exception":"[object] (TypeError(code: 0): trim(): Argument #1 ($string) must be of type string, array given at /opt/librenms/includes/discovery/discovery-protocols.inc.php:320)"}
```

PipoCanaja:
Changing snmpwalk mode to get only a simple array. Each row of this array should contain a hash as expected by the MIB. 

fixes #14474

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 14564`

### DIFF
--- a/includes/discovery/discovery-protocols.inc.php
+++ b/includes/discovery/discovery-protocols.inc.php
@@ -314,7 +314,7 @@ if (($device['os'] == 'routeros')) {
                     $remote_device_id = discover_new_device($lldp['lldpRemSysName'], $device, 'LLDP', $interface);
 
                     if (! $remote_device_id && Config::get('discovery_by_ip', false)) {
-                        $ptopo_array = snmpwalk_group($device, 'ptopoConnEntry', 'PTOPO-MIB');
+                        $ptopo_array = snmpwalk_cache_multi_oid($device, 'ptopoConnEntry', [], 'PTOPO-MIB');
                         d_echo($ptopo_array);
                         foreach ($ptopo_array as $ptopo) {
                             if (strcmp(trim($ptopo['ptopoConnRemoteChassis']), trim($lldp['lldpRemChassisId'])) == 0) {


murrant:
So, this is like doing `SnmpQuery::walk('PTOPO-MIB::ptopoConnTable')->values()`
instead of `SnmpQuery::walk('PTOPO-MIB::ptopoConnTable')->table()`?

PipoCanaja:
> So, this is like doing `SnmpQuery::walk('PTOPO-MIB::ptopoConnTable')->values()` instead of `SnmpQuery::walk('PTOPO-MIB::ptopoConnTable')->table()`?

Looks like it. But as I have no way to test the code, I am not really ready to make big changes here. I am not even convinced this code should be kept inside the LLDP loop. Seems that PTOPO MIB could be used in parallel of CDP, LLDP, and other XDP Mibs. Just in case some funny device only populates PTOPO and not LLDP ...

PipoCanaja:
Let's wait first from a feedback from @albertsulva and we'll see what we do after.

murrant (approved):

PR #14569, opened 2022-11-03
[WIP] My first eloquent commit: AlertUtil

fbouynot:

#### Please note

#### Testers


LibreNMS/Alert/AlertUtil.php
```diff
@@ -36,14 +37,11 @@ class AlertUtil
...
-     * @param  int  $alert_id
      * @return mixed|null
      */
-    private static function getRuleId($alert_id)
+    private static function getRuleId(int $alert_id)
     {
-        $query = 'SELECT `rule_id` FROM `alerts` WHERE `id`=?';
-
-        return dbFetchCell($query, [$alert_id]);
+        return Alert::select('rule_id')->where('id', $alert_id)->get();
```
Jellyfrog, line 44: have you tested this? I dont think its correct, since it returns a Collection, and not a value

Jellyfrog:
```
>>> Alert::select('rule_id')->where('id', 2)->get();
=> Illuminate\Database\Eloquent\Collection {#5198
     all: [
       App\Models\Alert {#5202
         rule_id: 2,
       },
     ],
   }

>>> Alert::where('id', 2)->value('rule_id');
=> 2

>>>
```

fbouynot: No I didn't know how to, now I know Tinker, I'll test. Thanks!

LibreNMS/Alert/AlertUtil.php
```diff
@@ -36,14 +37,11 @@ class AlertUtil
     /**
      * Get the rule_id for a specific alert
      *
-     * @param  int  $alert_id
      * @return mixed|null
      */
-    private static function getRuleId($alert_id)
+    private static function getRuleId(int $alert_id)
```
murrant, line 42: because the field is an integer, we know this will return an integer. (or null if not found)

0 and null are the same for the code, so setting the return type to int and casting to int, would be just fine here.

LibreNMS/Alert/AlertUtil.php
```diff
@@ -62,14 +59,10 @@ public static function getAlertTransports($alert_id)
 
     /**
      * Returns the default transports
-     *
-     * @return array
      */
-    public static function getDefaultAlertTransports()
+    public static function getDefaultAlertTransports(): array
```
murrant, line 63: You can probably return a Collection here, because it implements ArrayAccess.

The only thing that it causes an issue with is when is_array() is called.

fbouynot: I used Tinked to understand what would return the old function and do the same, and I thought it was an array: 

>>> DB::select('SELECT transport_id, transport_type, transport_name FROM alert_transports WHERE is_default=true', (array) []);
=> [
     {#4880
       +"transport_id": 1,
       +"transport_type": "browserpush",
       +"transport_name": "Browser notification",
     },
   ]

>>> AlertTransport::where('is_default', true)->select('transport_id', 'transport_type', 'transport_name')->get()->toArray()
=> [
     [
       "transport_id" => 1,
       "transport_type" => "browserpush",
       "transport_name" => "Browser notification",
     ],
   ]

Is it better to use a collection?

murrant: Try this in tinker:

```php
$transport = AlertTransport::first()
$transport->transport_name
$trasnport['transport_name']
$trasnport['transport_name'] = 'new name'

$transport
```

murrant:
Closing for now, Feel free to reopen or try again.

PR #14575, opened 2022-11-04
Clarify Dispatcher-Service.md for systemd

fuzzbawl:
There are two systemd service files suggested for use, only one should be used. My changes clarify the usage and adjust the librenms-watchdog.service to install as librenms.service as that appears to be the intention from #12188

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Dispatcher-Service.md
+++ b/doc/Extensions/Dispatcher-Service.md
@@ -169,22 +169,25 @@ Once the LibreNMS service is installed, the cron scripts used by LibreNMS to sta
 A systemd unit file is provided - the sysv and upstart init scripts
 could also be used with a little modification.
 
-### systemd
+### systemd service
 
 A systemd unit file can be found in `misc/librenms.service`. To
 install run `cp /opt/librenms/misc/librenms.service
 /etc/systemd/system/librenms.service && systemctl enable --now
 librenms.service`
 
-### systemd-watchdog
+### systemd service with watchdog
+
+This service file is an alternative to the above service file. It uses the systemd WatchdogSec= option to restart the service if it does not receive a keep-alive from the running process.
 
 A systemd unit file can be found in `misc/librenms-watchdog.service`. To
 install run `cp /opt/librenms/misc/librenms-watchdog.service
-/etc/systemd/system/librenms-watchdog.service && systemctl enable --now
-librenms-watchdog.service`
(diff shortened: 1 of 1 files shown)


doc/Extensions/Dispatcher-Service.md
```diff
@@ -169,22 +169,25 @@ Once the LibreNMS service is installed, the cron scripts used by LibreNMS to sta
...
-### systemd-watchdog
+### systemd service with watchdog
+
+This service file is an alternative to the above service file. It uses the systemd WatchdogSec= option to restart the service if it does not receive a keep-alive from the running process.
 
 A systemd unit file can be found in `misc/librenms-watchdog.service`. To
 install run `cp /opt/librenms/misc/librenms-watchdog.service
-/etc/systemd/system/librenms-watchdog.service && systemctl enable --now
-librenms-watchdog.service`
+/etc/systemd/system/librenms.service && systemctl enable --now
```
murrant, line 185: This seems incorrect, it should still be starting the librenms-watchdog unit right?

fuzzbawl: I did a diff between librenms.service and librenms-watchdog.service. The only difference was the watchdog line in systemd. From the discussion in #12188 it appears the intent was to copy one or the other service file as librenms.service, depending on if the end user wanted normal launch or launch with watchdog timer. In my experiences with systemd files, you want the service file name to be the same or similar to the application being launched. In my opinion as an end user, librenms-watchdog.service would be just the watchdog and not the librenms poller. This is actually what made me think that both service files needed to be used and ended up launching the poller twice. My CPU cried a little.

I'm happy to adjust the PR to launch this as librenms-watchdog.service if you prefer it that way.

murrant: I see, I didn't notice the && I thought it was another line.

Honestly, I think the above text is enough.  It is a little confusing to be renaming the file when you copy it.

VVelox: The diff makes is painful to look at as it does definitely look wrong, but I agree, since it is the only LibreNMS instance, no need to name it something else other than `librenms`.

VVelox: Looks good to me.

VVelox (approved):

murrant (approved):
Thanks!

PR #14600, opened 2022-11-08
feat: Add sortorder parameter to list_logs

talkstraightuk:
Add sortorder parameter to list_logs endpoint which accepts ASC or DESC to replace the original hardcoded sortorder.

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -2601,9 +2601,11 @@ function list_logs(Illuminate\Http\Request $request, Router $router)
         $param[] = $to;
     }
 
+    $sort_order = $request->get('sortorder') === 'DESC' ? 'DESC' : 'ASC';
+
     $count_query = $count_query . $query;
     $count = dbFetchCell($count_query, $param);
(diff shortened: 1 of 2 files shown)


includes/html/api_functions.inc.php
```diff
@@ -2601,9 +2601,11 @@ function list_logs(Illuminate\Http\Request $request, Router $router)
         $param[] = $to;
     }
 
+    $sort_order = $request->get('sortorder') === 'DESC' ? 'DESC' : 'ASC';
+
     $count_query = $count_query . $query;
     $count = dbFetchCell($count_query, $param);
-    $full_query = $full_query . $query . " ORDER BY $timestamp ASC LIMIT $start,$limit";
+    $full_query = $full_query . $query . " ORDER BY $timestamp $sort_order LIMIT $start,$limit";
```
Jellyfrog, line 2608: this should be inserted via $param

talkstraightuk: PDO doesn't allow binding outside of the prepared statement as far as I'm aware.  Which is why I assume you're interpolating timestamp, start and limit.

Jellyfrog (approved):

PR #14609, opened 2022-11-10
Updated check_dns arguments

RipJitter:

Normally for a service check LibreNMS will pass -H $HOSTNAME at the beginning of the service check, but for some reason it's passing -s $HOSTNAME at the end for the check_dns plugin. The -s should be for the DNS server, and -H should be the host to resolve.

#### Please note

#### Testers

### DIFF
--- a/includes/services/check_dns.inc.php
+++ b/includes/services/check_dns.inc.php
@@ -14,4 +14,4 @@ if ($service['service_ip']) {
     $resolver = $service['hostname'];
 }
 
-$check_cmd = \LibreNMS\Config::get('nagios_plugins') . '/check_dns -H ' . $nsquery . ' -s ' . $resolver;
+$check_cmd = \LibreNMS\Config::get('nagios_plugins') . '/check_dns -H ' . $resolver . ' $nsquery';


Jellyfrog:
You should rename the variables instead

murrant:
This check is intended to check if a DNS server is functioning.

RipJitter:
I am aware of what this plugin does.

Maybe my point is getting lost. With the current implementation, how do I specify the DNS server I want to resolve an address from?
$check_cmd = \LibreNMS\Config::get('nagios_plugins') . '/check_dns -H ' . $nsquery . ' -s ' . $resolver;
I cannot, as the -s switch is hardcoded to the host to resolve, instead of the server. The arguments are not correct. The -s switch should not be forced.

There are two Nagios plugins. DNS & DIG.

DNS:  This plugin uses the nslookup program to obtain the IP address for the given host/domain query.
An optional DNS server to use may be specified.
If no DNS server is specified, the default server(s) specified in /etc/resolv.conf will be used.

DIG: This plugin tests the DNS service on the specified host using dig

For DNS the server is always specified to be the host. It should not be.

murrant:
You set the target ip of the service check.  that sets the dns server.

murrant:
image

electrocret:
According to the nagios documentation, the command that this PR generates wouldn't work.

https://www.monitoring-plugins.org/doc/man/check_dns.html

murrant:
`check_dns -H mydomain.com -s 8.8.8.8` wouldn't work?

murrant:
Are you trying to look up the dns entry of the host the check is registered on?

Wouldn't you use this check to see if a DNS server you operate is functioning?  That is why the default host it looks up is "localhost"

PR #14613, opened 2022-11-10
RFC1628 - Force cast on some values before doing arithmetic

fufroma:
Some devices (like Delta InsightPower G3 Mini SNMP Card) return STRING instead of numerical value.

A few lines after, when doing a division on it, everything go bad ( / $divisor).

I added force casting, no more crash and the values are OK.

#### Please note

#### Testers


murrant:
There is a helper funtion for that. Number::cast I think.

fufroma:
> There is a helper function for that. Number::cast I think.

Thank you, I found `cast_number()` in `includes/helpers.php`, look like the wrapper I should have used.

PipoCanaja:
Hi @fufroma 
Could you by chance provide test data (from any device you like) that now behaves correctly thanx to this PR ? The idea is to ensure that if this code is changed later and this cast is removed, tests will catch it (and the code will then be adapted to avoid any regression)
Thanx

murrant:
I would prefer that you use the actual function call cast_number() is just a compatibility function:
```php
function cast_number($number)
{
    return \LibreNMS\Util\Number::cast($number);
}
```

fufroma:
@murrant OK, I'll fix it :). It may be useful to add a _deprecated_ tag on the "old" function.

murrant:
@fufroma good suggestion, but I'll probably just remove it because it is an internal function.

However, in your test data I don't see anything that would require this change.  Are you sure it is included in the snmprec?  What is the string output that you are getting from the device?  (technically all data returned from snmp is php string)

fufroma:
I don't have access to the equipment until monday.

If my memory is right, it's the NULL data that require to be cast as 0.

Example: 
`1.3.6.1.2.1.33.1.3.3.1.2.2|5|`

fufroma:
I checked: the crash are on the NULL lines.

Example:
```
$ ./discovery.php -h 269 -m sensors -ddd

[...]

Stack trace:
#0 /opt/librenms/includes/discovery/functions.inc.php(1011): include()
#1 /opt/librenms/includes/discovery/sensors.inc.php(84): sensors()
#2 /opt/librenms/includes/discovery/functions.inc.php(153): include('...')
#3 /opt/librenms/discovery.php(106): discover_device()
#4 {main}  
SQL[insert into `eventlog` (`reference`, `type`, `datetime`, `severity`, `message`, `username`, `device_id`) values (?, ?, ?, ?, ?, ?, ?) [null,"discovery","2022-11-14 07:49:17",5,"Error discovering sensors module. Check log file for more details.","",269] 0.56ms] 
  
Unsupported operand types: string / int {"exception":"[object] (TypeError(code: 0): Unsupported operand types: string / int at /opt/librenms/includes/discovery/sensors/voltage/rfc1628.inc.php:58)"} 
```

OID for this crash : .1.3.6.1.2.1.33.1.4.4.1.2.2

snmprec for this OID : 1.3.6.1.2.1.33.1.4.4.1.2.2|5|

```
$ snmpget -v 2c -c REDACTED  HOST .1.3.6.1.2.1.33.1.4.4.1.2.2
iso.3.6.1.2.1.33.1.4.4.1.2.2 = NULL
```

@murrant does it look OK to you?

fufroma:
I there something else I can provide?

murrant:
@fufroma sorry, no, I still have a question.  If the response is NULL, that means the sensor should not be discovered right?

fufroma:
This choice of implementation is entirely yours, I don't know if you rather have an empty sensor line for an existing sensor, or no sensor.

In my case, I chose to have a _minimal invasive surgery_ for the Librenms code as I'm not an expert :wink:

murrant:
I will probably merge this, so it is less broken for those devices.  But, yeah, we should not discover those.

fufroma:
Hello there,

Any news regarding this patch?

electrocret (approved):

murrant (approved):

PR #11022, opened 2020-01-10
Support of "disable alerting" in availability map and device summary widget

louis-oui:
image

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [X] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #11011, opened 2020-01-08
Clarify disable, ignore and disable all alerts in device edit section

louis-oui:
Clarify disable, ignore and disable all alerts in device edit section

Reason for this PR: PR https://github.com/librenms/librenms/pull/10809/files introduces wording change. But this change is same than https://github.com/librenms/librenms/pull/8900/files that was rejected with laf's comment: https://github.com/librenms/librenms/pull/8900#issuecomment-405308064 
>IMHO I think the clarification of this belongs in the docs really. Disable doesn't just disable polling so it's a bit misleading if we say it does. Same for ignore.

Removed section "alert settings" and moved their settings "device settings" section.
Changed "Disable Polling" to "Disable polling and alerting":
Changed "Ignore Alerts:" to "Ignore alert tag:"
Changed "Disable all alerting for this host" to "Disable alerting".

Current implementation:
image
image

With the PR:
image

DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [x] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #14619, opened 2022-11-10
added disable_notify and location_id

pertruccio:
in addition to PR [#14596](https://github.com/librenms/librenms/pull/14596) the fields disable_notify and location_id where missing and therefore not editable via api.

#### Please note

#### Testers


murrant:
I don't think those should be editable.  
Set the location via string.
Disable_notify is a defunct field and should be removed from the database.

murrant:
WTF?  https://github.com/librenms/librenms/pull/11022 https://github.com/librenms/librenms/pull/11011

What is the difference between ignore and notify_disable? What a terrible change...

pertruccio:
> I don't think those should be editable.
> Set the location via string.
> Disable_notify is a defunct field and should be removed from the database.

Setting location via String has the identical Problem: 
- HTTP Response 200 but no change

I didn't know that disable_notify was deprecated, the ignore tag will result in changes for alert rules.
The documentation has no detail information disable_notify and the Button is still in the UI.

murrant:
Apparently, disable_notify was introduced not that long ago, but I don't understand why.

pertruccio:
> Apparently, disable_notify was introduced not that long ago, but I don't understand why.

In my opinion there are 2 options to disable alarms per device:
- enable ignore_alert_tag
- enable disable_notify

The ignore_alert_tag will only work if you have alert rules that honor this option.

The disable_notify will work without preparing alert rules.

We used disable_notify to disable alarm events for devices that are marked as not active in our source of truth. We had set this option via api through our automation framework. Since we did not use alert rules which honor ignore_alert_tag we had no problems.

To get some progress on our problem...what do you think we should do in our situation? 
Neither location, nor location_id, nor disable_notify, nor ignore fields are changeable via api at the moment.
I started the pull request to get back a working process. Which fields should we add in the PR to get back functionality.

murrant:
@pertruccio sure, but why would you want the first option if the second option exists?

murrant (approved):

murrant:
Sorry to hijack your PR, but I was utterly mind-blown that two things that do the same thing exist, only one is worse.

murrant:
What's worse, is unfucking that is not simple.

PR #14625, opened 2022-11-11
Adding wireguard application support

bnerickson:
This PR adds application support for monitoring Wireguard clients' traffic and "minutes since last handshake" on a per-interface basis.  The corresponding script PR is located here: https://github.com/librenms/librenms-agent/pull/432

Below is a screenshot of the changes to the UI:

wireguard

#### Please note

#### Testers


Npeca75:
but why python ???

think about OpenWRT devices
they are limited in flash size
LibreNMS have already support for OpenWRT, so it would be shame to left out WG from OpenWRT because interpreter could not fit in limited flash

is there any chance to rewrite this in ASH ?
yes, ASH is default in OpenWRT, and it will run nicely on full Linux bash environment

bnerickson:
> but why python ???
> 
> think about OpenWRT devices they are limited in flash size LibreNMS have already support for OpenWRT, so it would be shame to left out WG from OpenWRT because interpreter could not fit in limited flash
> 
> is there any chance to rewrite this in ASH ? yes, ASH is default in OpenWRT, and it will run nicely on full Linux bash environment

No, this will not be rewritten in ASH.  Anyone interested in writing a version that can run on OpenWRT is welcome to do so using my code as a baseline, however.

Jellyfrog (approved):

arrmo:
FYI, getting an error with the script (in Ubuntu) - pretty sure it's because `wg` needs to be run as root, but `snmpd` runs as `Debian-snmp`. I have tried adding wg, python, and the script to visudo ... but no joy. Any pointers?

Thanks!

arrmo:
OK, even `visudo` isn't cutting it 😞. Seems you have to be root to access the wg interfaces, not just be able to run sudo. Hmm.

Any pointers appreciated!

arrmo:
Hi,

OK, two issues I found 😄. Let me try to explain,
1) On Linux (Ubuntu), snmpd runs as Debian-snmp ... can't access / run wg. BUT - if you add this user to `visudo`, and allow `bash` to be run, then you can update the row in snmpd.conf to be, `extend wireguard /usr/bin/bash -c "sudo -i /etc/snmp/wireguard.py"`
2) The current script fails to run if any parameters are set for the interface, like (for example), `FwMark` ... the reason is, `wg show all dump` outputs this info as well, which doesn't work with the current logic. I made a small mod (captured below), then it runs correctly. Perhaps update the code?

```
        if len(line.strip().split()) == 9:
            for intf, intf_data in output_parser(line, interface_clients_dict).items():
                if intf not in output_data["data"]:
                    output_data["data"][intf] = {}
                for client, client_data in intf_data.items():
                    output_data["data"][intf][client] = client_data
```
Thanks!

PR #14633, opened 2022-11-15
Added Moxa switching ifAlias/portName mapping

rhinoau:
The Moxa human-edited port descriptions/names are not accessible in IF-MIB.

A default port with a custom 'name' added is as below:
```
interface ethernet 1/5
 no shutdown
 name pr1_G1_name
 speed-duplex Auto
 media cable-mode auto
 no flowcontrol
 no gmrp
 mcast-filter forward-unknown
 switchport access vlan  1
 rate-limit port-disable ingress rate none
 no ptp
!
```

Event log showing some names being ingested with correct indexes:
image

I'm not excited about having to do it this way, but I need the custom port descriptions from the device.

Each Moxa EDS model has a unique MIB/sysObjectID, but is seems like stable mapping within them to the portTable at {sysObjectID}.1.9.1

I have tested across 2 models I have on my desk and they both work as intended, but with my limited PHP foo I can't find a nicer way to do it and leaving as WIP so I can learn what better ways we might find.

I'm ashamed of my OID hackery to obtain index, and had some troubles getting clean accessible info from the snmpwalk_cache_xxx variants that I could walk without tripping over.

Hoping someone may also have some other models to test against. 

#### Please note

#### Testers


rhinoau:
Open to better ways, but removing WIP to progress.

murrant:
I mean, it is garbage the the workaround is needed and that it is this obtuse, but I don't think there is an alternative.

murrant (approved):

PR #14639, opened 2022-11-16
Change ESXi port naming to be stable

TheMysteriousX:
Currently every time discovery runs against ESXi, the port names change generating a lot of log spam.

This is because ESXi uses this as the ifDescr:
```Device vmnic1000202 at 01:00.0 nmlx4_en```

The poller parses this, and extracts the interface name, whereas discovery does not, causing the interface names to flip every 6 hours.
https://github.com/librenms/librenms/blob/22.10.0/includes/polling/ports.inc.php#L578-L580

This change gives the discovery process the same behaviour, so the naming is now stable.

#### Please note

#### Testers

### CODE REVIEW

**Comment by TheMysteriousX** [comment]
I'll add test data etc. once someone's confirmed this is the best/correct way to fix the issue.

**Comment by murrant** [comment]
I believe the correct way to fix this is to set ifname:true in the os yaml.

**Comment by murrant** [comment]
Modifying ifDescr doesn't seem like a very good idea.

**Comment by TheMysteriousX** [comment]
Set ifname in the ESXi definition, but might have to restore the workaround anyway as vCenter looks to rely on it too - and ifname doesn't look to be a sane value.

**Comment by murrant** [comment]
Oh boy, that original code dates back to pre-fork.  I'm going to have to do some digging as it makes 0 sense to me.

**Comment by murrant** [comment]
Based on what I found, you should revert the yaml change and just delete the code in question instead.

Also, it looks like some interfaces are bit by the dumb default bad_if values.  Perhaps you need to override that at the os level with this change.  Maybe I'll look at changing the default in 22.12...

**Comment by TheMysteriousX** [comment]
I think the bad_if values are set fine for ESXi - what gets picked up are the physical ports, and the virtual ports/virtual management interfaces are discarded, e.g.:

```
.ignored by ifDescr: Traditional Virtual VMware switch: iSCSI 0 (matched: irtual)
Xignored by ifDescr: Traditional Virtual VMware switch: iSCSI 1 (matched: irtual)
Xignored by ifDescr: Distributed Virtual VMware switch: DvsPortset-0 (matched: irtual)
Xignored by ifDescr: Virtual interface: vmk0 on port 67108870 DVS 50 3c 31 41 3f 84 80 dc-09 b6 16 f9 fe 0f bf 9f (matched: irtual)
Xignored by ifDescr: Virtual interface: vmk1 on vswitch iSCSI 0 portgroup: iSCSI 0 (matched: irtual)
Xignored by ifDescr: Virtual interface: vmk2 on vswitch iSCSI 1 portgroup: iSCSI 1 (matched: irtual)
```

For vCenter, unfortuantely there is no ifname - only an ifdescr like:
```
Device eth0 at 03:00.0 VMware - VMXNET3 Ethernet Controller
```

So unless ifdescr can be parsed in YAML the ESXi yaml file needs ifname set, and the vCenter one needs the PHP interface name parsing to avoid log spam when discovery runs.

**Comment by murrant** [comment]
I don't understand, the test data says `"ifName": "vmnic4",`  which means it does have an ifName value.

File: includes/discovery/ports.inc.php [hunk 1]
```diff
@@ -94,6 +94,10 @@
     if (is_port_valid($snmp_data, $device)) {
         port_fill_missing_and_trim($snmp_data, $device);
 
+        if ($device['os'] == 'vmware-vcsa' && preg_match('/Device ([a-z0-9]+) at .*/', $snmp_data['ifDescr'], $matches)) {
+            $snmp_data['ifDescr'] = $matches[1];
```
murrant, line 98: Shouldn't this be ifName?

TheMysteriousX: Sure, can be, just needs an extra change.

murrant (approved):
Looks good matches my expectations.  The only odd bit is ifAlias, but likely that is not user editable here, so it is fine.

PR #14647, opened 2022-11-18
Ensure irc fifo-file is shared between dispatcher and core containers

Olen:

When running as docker containers, we need to share the ircbot.alert-file between the core container and the dispather container to send alerts using the irc-transport.

Here, the fifo-file is moved to the shared /data-folder if it exists and is owned by the user "librenms" which is true for the Docker images.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Irc.php
+++ b/LibreNMS/Alert/Transport/Irc.php
@@ -41,7 +41,7 @@ class Irc extends Transport
         if (file_exists($container_dir) and posix_getpwuid(fileowner($container_dir))['name'] == 'librenms') {
             $f = $container_dir . '/.ircbot.alert';
         } else {
-            $f = $this->config['install_dir'] . '/.ircbot.alert';
+            $f = Config::get('install_dir') . '/.ircbot.alert';
         }
         if (file_exists($f) && filetype($f) == 'fifo') {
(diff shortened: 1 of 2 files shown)


murrant:
it uses a fifo, hmm, there has to be a better way.

Olen:
> it uses a fifo, hmm, there has to be a better way.

A better way than using a fifo altogether to communicate with the bot?  
Or a better way to share the fifo between containers?
Or a better way to detect if we are using the containerized install to place the fifo in shared folder?

murrant:
A better way than using a fifo.

Olen:
> A better way than using a fifo.

You may be right*, although a fifo is a pretty simple and reliable way to share data between processes.

But just to be clear. The fifo has been the way to communicate with the ircbot since day one. All this PR do is moving the fifo file to inside a shared volume, so it is reachable from more than one container.

* Other options might be using sql to store and read the alerts, although that would lead to more load on the database (but probably not too much of a problem).
Another option could be to create a new api between the bot and the alerting system, but that would require a larger rewrite, and possibly some new endpoints.
All of this takes time and requires more tests etc. While this PR is just ensuring that the bot can send alerts even if it runs in a different container than the dispatcher.

PR #14660, opened 2022-11-21
Fix discovery ignores custom ipmi port

VoipTelCH:

Fix that allow a custom ipmi port set in the device edit section to be used by librenms

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/sensors/ipmi.inc.php
+++ b/includes/discovery/sensors/ipmi.inc.php
@@ -6,7 +6,7 @@ use LibreNMS\Config;
 // IPMI - We can discover this on poll!
 if ($ipmi['host'] = get_dev_attrib($device, 'ipmi_hostname')) {
     echo 'IPMI : ';
-    $ipmi['port'] = filter_var($device['attribs']['ipmi_port'], FILTER_VALIDATE_INT) ? $device['attribs']['ipmi_port'] : '623';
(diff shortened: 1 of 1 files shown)


includes/discovery/sensors/ipmi.inc.php
```diff
@@ -6,17 +6,17 @@
...
 if ($ipmi['host'] = get_dev_attrib($device, 'ipmi_hostname')) {
     echo 'IPMI : ';
-
+    $ipmi['port'] = get_dev_attrib($device, 'ipmi_port');
     $ipmi['user'] = get_dev_attrib($device, 'ipmi_username');
     $ipmi['password'] = get_dev_attrib($device, 'ipmi_password');
     $ipmi['kg_key'] = get_dev_attrib($device, 'ipmi_kg_key');
 
     $cmd = [Config::get('ipmitool', 'ipmitool')];
     if (Config::get('own_hostname') != $device['hostname'] || $ipmi['host'] != 'localhost') {
```
murrant, line 15: This will break of ipmi_port is not set.  is it always set or only sometimes?

VoipTelCH: Yes, you are right, didnt notice. We also found out that the form to set the IPMI parameters seems to be ignoring the default port (623) and it writes 0 if not manually specified

VoipTelCH: We found a solution to that, we took it from includes/polling/pimi.inc.php, I'll update the pull request

murrant (approved):

PR #14665, opened 2022-11-22
Device - Add support for "Smartpack R"

mrwold:
Add support for "Smartpack R"

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/enexus.yaml
+++ b/includes/definitions/enexus.yaml
@@ -16,8 +16,8 @@ discovery:
     -
         snmpget:
             oid: .1.3.6.1.4.1.12148.10.2.6.0
-            op: regex
-            value: '/Micropack System|Smart[Pp]ack S|Smart[Pp]ack R|SmartPack2 Touch|SmartPack2 Syst.|SPS-FPS|Flatpack S|Micropack 1U/'
+            op: '!='
+            value: false
 poller_modules:
     entity-physical: false
     hr-mib: false
--- a/includes/definitions/enexus.yaml
+++ b/includes/definitions/enexus.yaml
@@ -17,7 +17,7 @@ discovery:
         snmpget:
             oid: .1.3.6.1.4.1.12148.10.2.6.0
             op: regex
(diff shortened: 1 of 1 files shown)


murrant:
Can we maybe just include the string "smartpack", or does that conflict with other Eltek devices that use other OS?

includes/definitions/enexus.yaml
```diff
@@ -17,7 +17,7 @@ discovery:
         snmpget:
             oid: .1.3.6.1.4.1.12148.10.2.6.0
             op: regex
-            value: '/Micropack System|Smart[Pp]ack S|SmartPack2 Touch|SmartPack2 Syst.|SPS-FPS|Flatpack S|Micropack 1U/'
+            value: '/Micropack System|Smart[Pp]ack S|R|SmartPack2 Touch|SmartPack2 Syst.|SPS-FPS|Flatpack S|Micropack 1U/'
```
murrant, line 20: Also, this regex is wrong, includes anything with an `R` in the snmpget

murrant:
I have a better idea, we could change this to: 
```yaml
discovery:
    -
        sysObjectID:
            - .1.3.6.1.4.1.12148.10
    -
        snmpget:
            oid: .1.3.6.1.4.1.12148.10.2.6.0
            op: '!='
            value: false
```

That way if it responds to that oid with any value, it will work.  Is it true that some of these devices don't have SNMPv2-MIB::sysObjectID.0?

mrwold:
Regarding "SNMPv2-MIB::sysObjectID.0" I do not know, we currently only use Eltek SmartPack R. 
Unfortunately I have a issue generating test-data due to security policy, so I have attached screenshot as proof
image

peelman:
So I applied this, as we have a solitary SmartpackR in our fleet, along with a multitude of SmartpackS systems.  The problem, as I see it, is that .1.3.6.1.4.1.12148.10.2.6.0 is user-configurable.  This isn't the default, but it does complicate trying to use that as something to populate meaningful info into Libre with.

image

`.1.3.6.1.4.1.12148.10.13.8` would be a much more meaningful place to look for OIDs about the control unit.  That's where the current discovery YAML pulls the software version info from.

PR #14669, opened 2022-11-22
Handle multiple string index

murrant:

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/functions.inc.php
+++ b/includes/discovery/functions.inc.php
@@ -925,7 +925,7 @@ function discovery_process(&$valid, $os, $sensor_class, $pre_cache)
                     // strlen($index) as first number, and each letter converted to a number, separated by dots
                     if (preg_match_all('/"([^"]+)"/', $index, $index_string_match)) {
                         $oid = str_replace('{{ $index_string }}', array_reduce($index_string_match[1], function ($result, $string) {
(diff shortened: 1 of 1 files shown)


includes/discovery/functions.inc.php
```diff
@@ -923,7 +923,11 @@ function discovery_process(&$valid, $os, $sensor_class, $pre_cache)
                     $oid = str_replace('{{ $index }}', $index, $data['num_oid']);
                     // if index is a string, we need to convert it to OID
                     // strlen($index) as first number, and each letter converted to a number, separated by dots
```
Jellyfrog, line 925: It would be good with some explanation of this here, it's not really straight forward :)

murrant: For starters, it doesn't work :)

The junos one for example is a mix of strings and enums, neither of which is accessible otherwise, so it needs both the string and oid representations of the value.

`KPN.KPN-RPM.currentTest`  is the index without -b `3.75.80.78.7.75.80.78.45.82.80.77.1` with -b.

The main problem is the last one is an enum, not a string.

laf:
Is this still relevant @murrant ?

murrant:
Yes, but it breaks a lot of stuff by flipping the default.  It is kind of a painful change, I would be willing to re-visit it.

PR #14529, opened 2022-10-27
McAfee Web Proxy with COUNTER rrd_type on sensors

PipoCanaja:
First YAML file using `rrd_type: COUNTER`
Also fixed the sensor_index cause the one used before was not stable in time (resulting in mix between graphs) (that depended on #14522) 

Before (on prod): 
Capture d’écran 2022-10-27 à 10 08 16

After (on test with this PR): 
Capture d’écran 2022-10-27 à 10 07 26

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [ ] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [ ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.
- [ ] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #14670, opened 2022-11-23
Use 'counter' RRD type for some TrueNAS sensors

TheMysteriousX:
Like #14529, TrueNAS presents some counters that are best graphed discretely rather than continuously.
Also fix a typo, and give ARC P / ARC C better values (derived from arcstats:c / arcstats:p).

#### Please note

#### Testers


TheMysteriousX:
This works, but only for new hosts - the type of the RRD appears to be baked in at creation time.

PipoCanaja:
> This works, but only for new hosts - the type of the RRD appears to be baked in at creation time.

Hi
Yes. If you want to generate new graphs, the best is to change the index definition:
`index: zfsArcHits.{{ $index }}`
to
`index: zfsArcHits_Count.{{ $index }}`

That will force the creation of a new RRD. And the old one will be autocleaned after a while (depends on the config of the user).

TheMysteriousX:
Thanks, I'll take a look at that - is '_Count' a magic value? My memory of how the sensor indexes work is a little fuzzy.

PipoCanaja:
> Thanks, I'll take a look at that - is '_Count' a magic value? My memory of how the sensor indexes work is a little fuzzy.

Nop, not magic at all :) The sensor index in YAML is a pure arbitrary value that must be unique between sensors ... And the RRD file contains it in its name :) 
Unicity is achieved in 2 way : 
- The first part is common to all sensors extracted from the same table
- The 2nd part is the index which is by definition unique in the table 

So here, the only magic is to change the 1st part to something else. 

I could have said : 
`index: zfsArcHits.{{ $index }}`
to
`index: dJwuDIsg.{{ $index }}`

Result would have been exactly the same :)

murrant:
Seems like a good change, but I think _rate would be a better suffix.  (because you are taking a counter data type and storing it as a rate)

TheMysteriousX:
Switched over to _rate as suggested.

ottorei (approved):

PR #14676, opened 2022-11-25
Stop net-snmp from interpreting the octet-string from rttMonEchoAdmin…

not-known:
Most SNMP agent libraries do interpret the hexadecimal result from snmp as ascii value because it's an SNMP octet-string. In most cases this is good. In a specific case however it's not. The SNMP agent should not interpret the octet-string from IPSLA field rttMonEchoAdminTargetAddress but should leave it in hex format. After digging into de MIB file I found out that rttMonEchoAdminTargetAddress can be multiple types of value. DNS name, mac address, IP address, etc. Before you know how to interpret the address, the system needs to evaluate SNMP field rttMonEchoAdminProtocol. This field tells you what type of SLA check it was and so you know how to read the address.
Librenms has the logic build in, but the underlying agent interpreted the rttMonEchoAdminTargetAddress already before it reached the Librenms. The application didn't know what to do with the ascii string and ignored it resulting in a corrupted output for the IPSLA.

#### Please note

#### Testers


not-known:
> Please add test data so we can ensure your change is not broken in the future. Read the docs to find out how: https://docs.librenms.org/Developing/os/Test-Units

I'm not sure how to add test data for this change. In the existing tests I cannot find any reference to the existing code so my guess is this wasn't explicitly tested at all at the moment?

murrant:
Makes sense considering your change didn't cause a test failure.  Which means you should definitely add some new test data ;)

Jellyfrog (approved):

PR #14677, opened 2022-11-26
Adding linux_iw application support, new cdef graph template, and updating shared generic multi line exact graph

bnerickson:
This commit has three parts:

1. Update the generic_multi_line_exact_numbers.inc.php shared graph to include three new features:

1a. A new $print_format variable that allows the user to specify how the values are output in rrdtool (represented in the RRDTOOL GPRINT options).  Instead of the (still default) "%8.0lf%s" option, a user can now specify "%8.0lf" as the $print_format which would print the exact number instead of its larger magnitude SI unit.

(without $print_format):
print_format_not_set

(with $print_format set to %8.0lf):
print_format_set

1b. New $upper_limit and $lower_limit optional variables to specify the upper and lower bounds of the graphs.  This is useful for certain measurements (such as wireless noise and signal strength in dBm) where the values cannot be lower than -120 dBm.  In such a use-case, the lower limit can be set to -120 at all times and provide a nicer view of the graph.  I find it useful for setting the lower bound of the y-axis to 0 so the graphs don't autoscale with a constant and nearly invisible line at the base of the graph.  Example:

(autoscale - "DTIM Interval" is barely visible):
auto_scale

(new lower-limit set to 0 - "DTIM Interval" is clearly visible):
lower_limit_0

1c. The ability to set the "multiplier" or "divider" for each RRD specified in the rrd_list.  This will enable users to apply these modifications on a case-by-case basis.  For instance, a single graph page (say includes/html/graphs/application/app_time.inc.php) could include all time values that were collected in seconds, milliseconds, or nanoseconds.  The user can then choose to represent all of the values in milliseconds on the graph by applying an individual multiplier (1000) to the values represented in seconds, and a divider (1000) to all values represented by nanoseconds.  Here's what this looks like in code:

```
...
$unit_text = 'ms';

$rrdArray = [
    'connected_time' => ['descr' => 'Connected Time', 'multiplier' => 1000], // represented in seconds originally
    'inactive_time' => ['descr' => 'Inactive Time'], // represented in milliseconds
    'rx_duration' => ['descr' => 'RX Duration', 'divider' => 1000], // represented in nanoseconds originally
];

if (Rrd::checkRrdExists($rrd_filename)) {
    foreach ($rrdArray as $rrdVar => $rrdValues) {
        $rrd_list[] = [
            'colour'     => $rrdValues['colour'] ?? null,
            'descr'      => $rrdValues['descr'],
            'divider'    => $rrdValues['divider'] ?? null,
            'ds'         => $rrdVar,
            'filename'   => $rrd_filename,
            'multiplier' => $rrdValues['multiplier'] ?? null,
        ];
    }
} else {
    d_echo('RRD ' . $rrd_filename . ' not found');
}

require 'includes/html/graphs/generic_multi_line_exact_numbers.inc.php';
```

2. A new shared graph (generic_multi_line_exact_numbers_allow_maths.inc.php) has been created.  It is a direct copy of generic_multi_line_exact_numbers.inc.php, however, it allows the user to define custom CDEFs for values that are not stored in RRDs.  Instead the CDEFs are calculated based off of existing, user-specified RRDs using the "cdef_rpn" field in the rrd_list.  For example, say you define the following in (includes/html/graphs/application/app_frequencies.inc.php):

```
...
$rrdArray = [
    'center1' => ['descr' => 'Channel Center 1'],
    'center2' => ['descr' => 'Channel Center 2'],
    'channel' => ['descr' => 'Channel Floor'],
    'width' => ['descr' => 'Channel Width'],
    'channel_ceiling' => ['descr' => 'Channel Ceiling', 'cdef_rpn' => ['val1' => 'channel', 'val2' => 'width', 'oper' => '+']],
];

if (Rrd::checkRrdExists($rrd_filename)) {
    foreach ($rrdArray as $rrdVar => $rrdValues) {
        $rrd_list[] = [
            'cdef_rpn'   => $rrdValues['cdef_rpn'] ?? null,
            'colour'     => $rrdValues['colour'] ?? null,
            'descr'      => $rrdValues['descr'],
            'divider'    => $rrdValues['divider'] ?? null,
            'ds'         => $rrdVar,
            'filename'   => $rrd_filename,
            'multiplier' => $rrdValues['multiplier'] ?? null,
        ];
    }
} else {
    d_echo('RRD ' . $rrd_filename . ' not found');
}

require 'includes/html/graphs/generic_multi_line_exact_numbers_allow_maths.inc.php';
```

The "center1", "center2", "channel", and "width" values will be extracted directly from RRD file(s).  However, "channel_ceiling" will, thanks to the changes, create a line on the graph that is represented by the value of "channel" PLUS the value of "width".  Here's an example graph of what that looks like:
print_format_set

3. Adding wireless_ap application support.  The script polls "iw" statistics on a Linux device running as a wireless AP and the support added in this PR displays those statistics.  Sample graphs can be viewed in the following PR which adds the application script to librenms-agent: https://github.com/librenms/librenms-agent/pull/442

#### Please note

#### Testers


murrant:
I like all the changes you have displayed here, but they should just be baked in to the given graphs.

bnerickson:
> I like all the changes you have displayed here, but they should just be baked in to the given graphs.
> 
> If these are required for other changes, they should be included in this PR. A change without any usages is pointless.

Sure, I can add that to the PR.  Just didn't want to slam reviewers with too much code to review.

bnerickson:
@murrant , went ahead and merged my wireless_ap application changes with the original graphs changes and updated the description.

VVelox:
This looks good to me.

That said the only thing that really stands out to me is client_lifetime. This is something I would personally move into the LibreNMS config as well as adding a optional setting of allowing it accept settings for that from the remote system.

murrant:
What do you think about also populating the access_points database table?  Maybe also populating graphs there too.

bnerickson:
@VVelox Changed the name of the application to "linux_iw".  The latest version also sets the "client_lifetime" (now referred to as "linux_iw_cap_lifetime") global default to "0" (never expire).  Users can change this in LibreNMS config.php as the global setting.  In-turn, that value can be overwritten by specifying the same value in the script JSON file.  This is all specified in the doc.  I also fixed a lifetime calculation bug I noticed while testing this out.  Lastly, I updated the graph title logic to display "AP" (if the host we are polling is a client) or "Client" (if the host we are polling is an AP) so the script can be used in either case.

@murrant I'll investigate this week whether updating the access_points database yields any tangible benefits.

bnerickson:
Hey @murrant , I looked into the access_points table and it doesn't appear to fit this use-case.  The access_points table is geared towards wireless controllers and their managed APs.  Whereas this application is simply polling a Linux server running as a wireless station or AP.

I think that a better approach might be adding "implements Wireless*" to the LibreNMS/OS/Shared/Unix.php Unix class ala...
```
class Unix extends \LibreNMS\OS implements
    MempoolsDiscovery,
    WirelessClientsDiscovery,
    WirelessCcqDiscovery,
    WirelessCcqPolling,
    WirelessFrequencyDiscovery,
    WirelessFrequencyPolling,
    WirelessPowerDiscovery,
    WirelessUtilizationDiscovery
...
```
...and adding the discovery functions to said shared class.  I think I would also need to modify the linux_iw.py script to output OIDs in a more standardized format as opposed to a single OID with JSON for the WirelessSensor object.  Before I work on this, however, would you confirm whether this is or is not the right approach?

PR #14694, opened 2022-12-02
Adding extra variables recovery and invert + adding invert_map direct…

geg347:
…ly in DB for edit alert_rule API request

Hello dear LibreNMS community,

After this PR https://github.com/librenms/librenms/pull/14481, we tried to modify alert rules with the API and found some weird behaviour.
We can't set the extra group fields in the add_edit alert rules API request. Default fields are restored.
We have to break each field of this extra group to put them separately in the PUT request.
The 2 fields recovery and invert were not evaluated so first change I added them.
Another fields was not handled, invert_map but this one is not under extra group, it has his proper fields directly in the DB.
I added it the same way as the field disabled is handled.
I also quickly updated the documentation.

Feel free to challenge me if changes are needed.

Thanks and have a great day!
geg347

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1250,10 +1250,17 @@ function add_edit_rule(Illuminate\Http\Request $request)
         $disabled = 0;
     }
 
+    $invert_map = $data['invert_map'];
+    if ($invert_map != '0' && $invert_map != '1') {
+        $invert_map = 0;
+    }
+
     $count = $data['count'];
     $mute = $data['mute'];
     $delay = $data['delay'];
     $interval = $data['interval'];
+    $recovery = $data['recovery'];
+    $invert = $data['invert'];
     $override_query = $data['override_query'];
     $adv_query = $data['adv_query'];
     $delay_sec = convert_delay($delay);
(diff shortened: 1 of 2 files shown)


includes/html/api_functions.inc.php
```diff
@@ -1250,10 +1250,17 @@ function add_edit_rule(Illuminate\Http\Request $request)
         $disabled = 0;
     }
 
+    $invert_map = $data['invert_map'];
```
murrant, line 1253: Is this just a much longer way to write:
```php
$invert_map = (int) ($data['invert_map'] ?? 0);
```

Note: a lot of this code is missing isset or nullable checks.  Probably causes errors when fields are missing in the request.

murrant: or maybe like this `($data['invert_map'] ?? 0) ? 1 : 0`

murrant: or maybe like this: `$invert_map = (int) ! empty($data['invert_map');`

geg347: Hello,
You may be right I don't know all the syntax in PHP. I just copied the existing code for the $disabled variable.
I don't know which syntax is the best in term of performance, though.
Do you want me to update those if conditions for $disabled and $invert_map?

I tested both variables with your new syntax and it is working.

geg347: Your second proposition is causing the API to crash and it is not easy to troubleshoot for me.

murrant:
#

PR #14695, opened 2022-12-05
remove round from timeFormat function

Uzoma8Chi:
If we have 16 hours time elapsed should be just 16h not 1d 16h. Using round there is a mistake. 16 hours is 0.67 day. If we round 0.67 we get 1 which is wrong.

This solution probably needs some testing because I am not sure how precise is PHP numerically but all my tests went through.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/RunAlerts.php
+++ b/LibreNMS/Alert/RunAlerts.php
@@ -209,11 +209,11 @@ class RunAlerts
     public function timeFormat($secs)
     {
         $bit = [
-            'y' => round($secs / 31556926) % 12,
-            'w' => round($secs / 604800) % 52,
-            'd' => round($secs / 86400) % 7,
-            'h' => round($secs / 3600) % 24,
-            'm' => round($secs / 60) % 60,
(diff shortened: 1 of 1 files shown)


murrant:
This code is also wrong.  You should use floor() instead of round() ;)

murrant:
Actually, the modulo should do that, so maybe it is correct.  Perhaps adding some tests for this function might help...

murrant:
Ok, I see why the round was added, yes, it should be floor:
"PHP Deprecated:  Implicit conversion from float 1.9038333333333335 to int loses precision"

Uzoma8Chi:
Also a possible solution if some modifications will be done.
`
function timeFormat($seconds){
    $dtF = new \DateTime('@0');
    $dtT = new \DateTime("@$seconds");
    return $dtF->diff($dtT)->format('%yy %ad %hh %im %ss');
}
`

murrant:
@Uzoma8Chi I think the reason that isn't used is we don't show 0s.  If you can cleanly make that work, fine by me otherwise, the only fix needed is to change round to floor.

PR #14697, opened 2022-12-05
Device Overview for Network Devices

Bierchermuesli:
This is a attempt / disussion for better device overview in case of a network device.
processor, memory etc is good for servers. For network device other metrics like latency, avaibility, is more important. Also a port list is not very useful if the deivce has more than 10 interfaces. 

As an example before:
image

After:
image

What do you think? Other metrics? other ways to acomplish my idea?

#### Please note

#### Testers


murrant:
I would rather have user configurable device dashboards similar to the main dashboard.  (with sane default dashboards)

Bierchermuesli:
I agree on the user perspective, custom Dashboard per device or OS type?
Although it sounds a bit overkill for me (loading time etc?). Could predefined dashboards maintained by the OS maintainer be a lightweight alternative?
I can look into this but not sure if I can contribute on the frontend stuff.

includes/html/pages/device/overview/ports.inc.php
```diff
@@ -50,11 +50,12 @@
...
-    foreach (dbFetchRows("SELECT * FROM `ports` WHERE device_id = ? AND `deleted` != '1' AND `disabled` = 0 ORDER BY ifName", [$device['device_id']]) as $data) {
-        $data = cleanPort($data);
-        $data = array_merge($data, $device);
-        echo "$ifsep" . generate_port_link($data, makeshortif(strtolower($data['label'])));
-        $ifsep = ', ';
+    if ($device['type'] == 'network') {
+        //a semicolon list for >10 Ports does not makes sense, so we save queries/rendering and make a link for network devices
+        echo '<a href="' . url('device') . '/' . DeviceCache::getPrimary()->device_id . '/ports/' . '">Go to Port list</a>';
+    } else {
+        foreach (dbFetchRows("SELECT * FROM `ports` WHERE device_id = ? AND `deleted` != '1' AND `disabled` = 0 ORDER BY ifName", [$device['device_id']]) as $data) {
```
Jellyfrog, line 57: This doesn't do much

Bierchermuesli: Oops, local patch failure. Fixed now

murrant:
I think removing some of the panels will result in some negative responses from users (I'm not saying we should only add things).  I'm a little hesitant about this PR.

As for editable dashboards, I think type -> os -> device would be appropriate override tiers.  The last one user editable.

ottorei:
> I think removing some of the panels will result in some negative responses from users (I'm not saying we should only add things). I'm a little hesitant about this PR.
> 
> As for editable dashboards, I think type -> os -> device would be appropriate override tiers. The last one user editable.

I think this should be a configurable option which the users can configure as they wish by device type / os and use the existing behavior as a default. In my case I am actually very interested in line card / dataplane loads of the device rather than the latency and not interested in the availability graphs at all since I use the API to calculate thousands of lines of outages and availabilities for arbitrary time intervals. But I get it that some users might want to customize the overview page and I think the correct way would be to implement a configuration option for this.

As for the traffic graph it is true that with some devices there are thousands of interfaces and plotting or printing them all to the graph is not that useful - the graph should probably only contain physical interfaces by default.

murrant:
Just a little whitespace fix needed.  I think we can merge this for now as a short term gain.

PipoCanaja:
Hi @Bierchermuesli 
I agree with above discussion. I need to see my proc and mem details on the network device overview page, at least until vendors will provide firmware with no memory leaks and no processes eating all the CPU ... If you have Cisco, Huawei, Zyxel, Fortinet, then you probably also want to keep these displayed, as we have regular occurrences of these leaks ... 
So I would not approve any PR that statically removes these from device-overview. 
Of course, an approach to make this configurable (either via config or even better via GUI) would probably benefit some specific cases like yours.

Bierchermuesli:
this week I can fully agree with you (i'm looking at you cisco...) - alltrought the timerange should be longer for the memory leak case.  anyway... other topic. 

should I scrap it or atleast keep the "if more than 10 ports dont make a list" function?

PipoCanaja:
I personally have switches with 150 interfaces and it is perfectly usable (quicker than opening the port page to search for a precise interface). So I wouldn't go for any arbitrary change that is not user configurable.

PR #14709, opened 2022-12-07
Add download button to device config page

gazpachoking:
Adds a button to the device configuration page to download the viewed config version from oxidized.

On modern browsers this will just open the raw config in a new tab, since the 'download' attribute on the link doesn't apply for security reasons when linking to another domain. This could be fixed by adding an endpoint to librenms to proxy the raw config, but it didn't seem worth it since you can just right click the link and do a save as, or save from the resulting tab.

This also fixes a small bug where the version number passed in oxidized url parameters was always equal to the maximum version number. Seems oxidized ignored this argument though, so it doesn't matter much.

Feature request here:
https://community.librenms.org/t/oxidized-configuration-download-button/11780

image

#### Please note

#### Testers

### DIFF
--- a/includes/html/pages/device/showconfig.inc.php
+++ b/includes/html/pages/device/showconfig.inc.php
@@ -206,9 +206,9 @@ if (Auth::user()->hasGlobalAdmin()) {
                 $download_uri = Config::get('oxidized.url') . $uri;
             }
         } else {  // just fetch the only version
-		    $uri = '/node/fetch/' . (! empty($node_info['group']) ? $node_info['group'] . '/' : '') . $oxidized_hostname;
+            $uri = '/node/fetch/' . (! empty($node_info['group']) ? $node_info['group'] . '/' : '') . $oxidized_hostname;
             $text = (new \App\ApiClients\Oxidized())->getContent($uri);
-			$download_uri = Config::get('oxidized.url') . $uri;
+            $download_uri = Config::get('oxidized.url') . $uri;
         }
 
         if (is_array($node_info) || $config_total > 1) {
--- a/includes/html/pages/device/showconfig.inc.php
+++ b/includes/html/pages/device/showconfig.inc.php
@@ -201,10 +201,14 @@ if (Auth::user()->hasGlobalAdmin()) {
                 $text = (new \App\ApiClients\Oxidized())->getContent($uri); // fetch diff
             } else {
                 // fetch current_version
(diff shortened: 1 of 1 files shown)


gazpachoking:
> I like the idea. However, I have a problem with it linking directly to the Oxidized URI.
> 
> For security reasons, we have our Oxidized segregated from the rest of our network, and the only way to access it is through a secure jumphost or through the pin hole, allowing API requests only from Libre. Imagine how big of a problem it would be if your Oxidized server got compromised? Copy of all the configs and access to the equipment.
> 
> The download request should be proxied through Libre.

Yeah, that would be nicer for other reasons as well, including making the download link actually download the file, and letting us provide a filename hint. I'm not super familiar with the codebase though, do you know where that proxying should occur?

electrocret:
The API does a proxying like function already, you could likely copy it.
https://github.com/librenms/librenms/blob/fbd4bc126973b394a17a7efcdf04d306411f1fd0/includes/html/api_functions.inc.php#L1434

gazpachoking:
Just found that as well. Upgrading that endpoint to allow requesting a specific versions seems like a good plan. I'm not sure how I call that api endpoint from the webui though.. I'll keep digging.

Another note, what is securing your oxidized instance that this method would open up? If I add an endpoint that proxies the oxidized config, how would that be, (or rather, how would I ensure that it is,) more secure?

ottorei:
> The API does a proxying like function already, you could likely use it for reference.
> 
> https://github.com/librenms/librenms/blob/fbd4bc126973b394a17a7efcdf04d306411f1fd0/includes/html/api_functions.inc.php#L1434

I agree that the proper way to achieve this download feature is to load the specified configuration file through Oxidized API using NMS itself just like the existing configuration fetching does. With direct access the default configuration would not even authenticate users and this would possibly pose a security issue. Also, using NMS server to fetch the configs, there would be no need for any routing between the http client and the Oxidized server unless one explicitly wishes to do so.

ottorei (changes requested):
Please check the existing Oxidized API access code and try to convert this feature to use connectivity of the NMS server to the Oxidixed instead of direct access from web clients.

PipoCanaja:
> Another note, what is securing your oxidized instance that this method would open up? If I add an endpoint that proxies the oxidized config, how would that be, (or rather, how would I ensure that it is,) more secure?

In fact, it is as secure as LibreNMS is. But as Oxidized GUI is minimalist, a few of us prefer to keep it locked on 127.0.0.1 and only access it thru LibreNMS. So any direct link to it would not even work in that case. 
Using the proxy function is indeed the only way to ensure that the download button will work for all, and that the security will remain as "good" as it currently is.

PipoCanaja:
Hi @gazpachoking 
Are you still working on this PR ?
Thanx

murrant:

PR #14725, opened 2022-12-14
add exclude pattern to snmp-scan.py

slalomsk8er:
Added argument to skip host names matching a regex pattern.

#### Please note

#### Testers

### DIFF
--- a/snmp-scan.py
+++ b/snmp-scan.py
@@ -243,7 +243,7 @@ Example: 192.168.0.1/32 will be treated as a single host address""",
         "-e",
         "--exclude",
         help="Exclude pattern to check hostnames against",
-        dest="pattern",
+        dest="pattern"
     )
 
     args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


murrant:
This use case seems extremely niche, I don't think it makes sense to include in upstream.

slalomsk8er:
You don't have testing and production servers, separated by a naming convention?

------- Original Message -------

murrant:
They are in different IP spaces.

slalomsk8er:
I'm not so lucky but at least development separate.

murrant:
Again, thanks a bunch for your contribution.  I hope to see more in the future.

I'm going to close this for now as I don't see it to be generally useful.  Feel free to carry the patch locally.  Also, I suggest not running snmp-scan.py after the initial network scan.  We use netbox for inventory management etc and synchronize that to LibreNMS.

PR #14755, opened 2022-12-22
Add support for PHP 8.2

Aerowinder:
PHP 8.2 was not accepted for package fgrosse/phpasn1. This causes an error in the auto-update script on Debian 11, and would probably prevent a new installation.

#### Please note

- [X ] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [X ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.
- [X ] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers


Jellyfrog:
This looks like a manual edit of the composer.lock file, which is not allowed, and I'm surprised the tests even passed. 
But maybe I'm wrong?
Else you need to https://getcomposer.org/doc/01-basic-usage.md#updating-dependencies-to-their-latest-versions

murrant:
#14759 should fix this without breaking things (your PR would have allowed it to run, but with webpush broken)

PR #14767, opened 2023-01-02
Add more docker stats

hvanoch:
This pull requests add extra stats for docker.
This includes
- container uptime
- container size
- container filesystem 
- Totals per status

#### Please note

#### Testers

### DIFF
--- a/includes/polling/applications/docker.inc.php
+++ b/includes/polling/applications/docker.inc.php
@@ -3,31 +3,16 @@
 use LibreNMS\Exceptions\JsonAppException;
 use LibreNMS\Exceptions\JsonAppMissingKeysException;
 use LibreNMS\RRD\RrdDefinition;
+use LibreNMS\Util\Number;
 
 $name = 'docker';
+$version = 1;
 $output = 'OK';
 
-function convertToBytes(string $from): ?int
-{
-    $units = ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB'];
-    $number = substr($from, 0, -3);
-    $suffix = substr($from, -3);
-
-    //B or no suffix
-    if (is_numeric(substr($suffix, 0, 1))) {
-        return (int) $from;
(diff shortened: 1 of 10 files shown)


hvanoch:
Tests are failing. I believe the reasoning for the exception is that I have test for both v1 and v2 which probably include the same source file. not sure on how to get rid of that one.

Jellyfrog:
I think its because its includes docker.inc.php twice,

but this functions should be generic instead, and it probably exists one that already can do the work in some helper class
```
In docker.inc.php line 11:
                                                                               
  Cannot redeclare convertToBytes() (previously declared in /home/runner/work  
  /librenms/librenms/includes/polling/applications/docker.inc.php:11)      
```

hvanoch:
I moved the function to a separate helper class. Tests are also fixed so this MR is ready for review again.

Jellyfrog:
Would you mind moving your new class into a function in one of the existing helper classes? Number.php would be a good match I suppose. Also add a comment for the function what it does

hvanoch:
I have moved the conversion function to Number.php class as requested.

hvanoch:
@Jellyfrog are you able to give it an other review?
Thanks

includes/polling/applications/docker.inc.php
```diff
@@ -37,33 +22,73 @@ function convertToBytes(string $from): ?int
     return;
 }
 
+$version = intval($version);
+
+if ($version == 1) {
+    $output = 'LEGACY';
```
Jellyfrog, line 28: @murrant is this how we usually do this?

hvanoch: I took the dhcp stats as a reference: https://github.com/librenms/librenms/blob/master/includes/polling/applications/dhcp-stats.inc.php#L29

Jellyfrog (approved):

PR #14779, opened 2023-01-10
API get_location

VirTechSystems:
Adds API endpoint to check if location exists and return the data if it does.  Can be done via name or id.

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -2837,6 +2837,7 @@ function edit_location(Illuminate\Http\Request $request)
 
     return api_error(500, 'Failed to update location');
 }
+
 function get_location(Illuminate\Http\Request $request)
 {
     $location = $request->route('location_id_or_name');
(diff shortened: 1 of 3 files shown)


includes/html/api_functions.inc.php
```diff
@@ -2838,6 +2838,21 @@ function edit_location(Illuminate\Http\Request $request)
...
 }
 
+function get_location(Illuminate\Http\Request $request)
+{
+    $location = $request->route('location_id_or_name');
+    if (empty($location)) {
+        return api_error(400, 'No location has been provided to get');
+    }
+    $location_id = ctype_digit($location) ? $location : get_location_id_by_name($location);
+    $data = dbFetchRows('SELECT `locations`.* FROM `locations` WHERE `locations`.`id` = ?', [$location_id]);
```
murrant, line 2848: You should not use dbFetchRows, it is deprecated.
```php
$data = \App\Models\Location::find($location_id);
```

You could also prevent the get_location_id_by_name() name above with a more complex query.  (that function calls the database)

electrocret (approved):

murrant:
Odd, seems like a glaring omission.

PR #14793, opened 2023-01-19
Fix entity-state polling/discovery

fdomain:
entity-state is sometimes not updated in the database according the real state on the device.
I found out that array_diff doesn't work as expected with associatiave arrays and doesn't return a diff where we should have one, so that's why we use array_diff_assoc instead.

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/entity-state.inc.php
+++ b/includes/discovery/entity-state.inc.php
@@ -56,7 +56,7 @@ if (! empty($entPhysical)) {
 
             if (isset($db_states[$id])) { // update the db
                 $db_state = $db_states[$id];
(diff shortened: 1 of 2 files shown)


Jellyfrog (approved):

Jellyfrog:
Makes sense

fdomain:
Hello @murrant
I noticed that this commit has been revert, do you have more information on that ?

murrant:
@fdomain we don't know.  Several users reported the issue, that is why we reverted.  I suggest you profile it a bit to see, likely it is a very large array or something.

PR #14795, opened 2023-01-19
Changed discovery OID in ets.yaml

da-me:
The original OID matches all devices from the manufacturer Gude, which makes it rather dificult to add different products which belong to a different product category. Since the corresponding MIB file GUDEADS-ETS-MIB only references .1.3.6.1.4.1.28507.4 it should be safe to use only this specific OID for device discovery.

Merging this change would then enable supporting other Gude devices like some PDUs using OIDs like .1.3.6.1.4.1.28507.81,  .1.3.6.1.4.1.28507.56 or others.

#### Please note

#### Testers


da-me:
Yes, absolutely. 

Support for the Gude PDU EPC8221 is almost finished and looks good on my local development instance, and EPC8031 will be added as soon as I get the hardware delivered.

I just thought it might be good to do this in multiple parts: this PR to lay the groundwork, and another PR for actually adding the devices.

Jellyfrog (approved):

PR #14799, opened 2023-01-20
Allows GEO lookup with detailed internal Location information wrapped inside parentheses.

LoveSkylark:
This change allows Location information to contain data that is not relevant to GEO lookups without breaking them.

Anything inside parentheses will be ignored ruing GEO lookup so if you are able to add internal information to the Location string:
```
1100 Congress Ave, Austin, TX 78701 (blue cabinet on the left)
```
This will look up "1100 Congress Ave, Austin, TX 78701" but you will still have the detail information that this is in the blue cabinet

If approved I suggest adding Location parsing  to the documentation

#### Please note

#### Testers


LoveSkylark:
don know how to get past this fail, anyone here willing to help me fix this?

LoveSkylark:
Should I bee adding this to phpstan-beaseline.neon:?
			message: "#^Property App\\\\Models\\\\Location\\:\\:\\$location_ignore_regex has no type specified\\.$#"
			count: 1
			path: app/Models/Location.php

rhinoau:
I don't actually know for sure but maybe baseline as you showed and also in here?

https://github.com/librenms/librenms/blob/44578b2935eb274f1d1ea0de47d0de4b28940057/misc/discovery_schema.json#L106-L108

LoveSkylark:
> I'm concerned that parenthesis is already used for something, but I don't have the brainpower to parse the other regex at this time.

This is only used when the messages is sent to do a  GEO lookup.

Text inside parentheses during a geo lookup typically represents additional, optional information about the location.

Examples:

-   Paris-Orly Airport (ORY), Orly, France
-   Keflavik International Airport (KEF), Reykjanesbaer

but it does not yield any additional search result:

-   looking up (ORY) & ORY yields nothing.
-   looking up (KEF) & KEB both yields the same results.

and the is no difference in looking up:

-   Paris-Orly Airport (ORY) or Paris-Orly Airport

But I suppose a different parser can be used if this is your concern?

I do however believe this function is necessary for the Location system to be of use for any large site.

LoveSkylark:
Right now if you want to use SNMP locations like:

- Sydney Opera House, 1st floor
- Sydney Opera House, 2nd floor
- Sydney Opera House, 3rd floor
- Sydney Opera House, server room

You would have to put in ever single lat/lng manually, or reduce the SNMP info down to "Sydney Opera House" only and write the location under notes or something.

With this change you can have:

- Sydney Opera House (1st floor)
- Sydney Opera House (2nd floor)
- Sydney Opera House (3rd floor)
- Sydney Opera House (server room)

and lat/lng is automatically populated.

app/Models/Location.php
```diff
@@ -127,7 +128,10 @@ protected function fetchCoordinates()
         try {
             /** @var \LibreNMS\Interfaces\Geocoder $api */
             $api = app(\LibreNMS\Interfaces\Geocoder::class);
-            $this->fill($api->getCoordinates($this->location));
+
+            // Removes Location info inside () when looking up lat/lng
+            $this->fill($api->getCoordinates(preg_replace($this->location_ignore_regex, '', $this->location)));
+            // $this->fill($api->getCoordinates($this->location));
```
murrant, line 134: You'll need to remove this commented code, thanks :)

LoveSkylark: removed

murrant (approved):

PR #14803, opened 2023-01-23
Update XDP string to exclude modern Cisco lightweight APs from discovery

tristanbob:
Existing string was not changed, as older APs might use that format.
To verify device string, view: librenms\tests\snmpsim\aos6.snmprec (line 233)

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Auto-Discovery.md
+++ b/doc/Extensions/Auto-Discovery.md
@@ -130,6 +130,7 @@ These devices are excluded by default:
 
 ```php
 $config['autodiscovery']['xdp_exclude']['sysdesc_regexp'][] = '/-K9W8-/'; // Cisco Lightweight Access Point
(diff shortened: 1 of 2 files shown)


tristanbob:
To give some context: I discovered this issue when I noticed my large c9400 devices had empty FDB tables in LibreNMS.

I was able to manually collect these by only running the FDB discovery module.  
However, a full discovery would hang trying to connect to all the lightweight APs. 
I think LibreNMS was giving up before the FDB module could start.

After making this change, our c9400 now have fully populated FDB tables.

tristanbob:
> Why not just add `i` at the end of the regex?

That should work to deal with the capitalization change (but I haven't tested it yet).

We also need to deal with the change from the last character being a dash "-" to a space " ".  
The simplest issue might be to remove that last character.

I'm open to either of these ideas, I just need to know which method the project recommends.

tristanbob:
I tested this some more, and the regex is case-insensitive by default.
In order to match old and new APs, the only change required is to remove the trailing "-".

Jellyfrog (approved):

PR #14805, opened 2023-01-25
Implement authentication for Redis/Sentinel

TheMysteriousX:
As far as I can tell, the redis-sentinel support only works if it's unauthenticated. An attempt to pass through the password is made, but the way it's passed through means that redis-py just drops it on the floor.

This change adds support for basic (requirepass) authentication, as well as the new ACL's introduced in Redis 6.

It may work on redis-py 3.5.3, but that is now no longer receiving security fixes.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/__init__.py
+++ b/LibreNMS/__init__.py
@@ -542,7 +542,7 @@ class RedisUniqueQueue(object):
                 for k, v in redis_kwargs.items()
                 if k in ["decode_responses", "username", "password", "db", "socket_timeout"]
             }
-            self._redis = Sentinel(sentinels, sentinel_kwargs, **kwargs).master_for(sentinel_service)
+            self._redis = Sentinel(sentinels, sentinel_kwargs=sentinel_kwargs, **kwargs).master_for(sentinel_service)
         else:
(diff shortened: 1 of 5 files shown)


TheMysteriousX:
Works for me with 3 sentinel instances/3 redis instances with ACL's.

This should be a noop for anyone running standalone redis, or sentinel without authentication.

It is definitely a noop for anyone using Sentinel with authentication or Redis with ACL's as they don't work currently.

It would be handy to have someone with a standalone or unauth'd sentinel setup to verify this works for them.

ottorei:
> Works for me with 3 sentinel instances/3 redis instances with ACL's.
> 
> This should be a noop for anyone running standalone redis, or sentinel without authentication.
> 
> It is definitely a noop for anyone using Sentinel with authentication or Redis with ACL's as they don't work currently.
> 
> It would be handy to have someone with a standalone or unauth'd sentinel setup to verify this works for them.

I have a distributed setup with unauthenticated Redis. Could probably test this later this week. In my case the authentication was never required since the internal connectivity between NMS- and DB-servers are using dedicated connections.

murrant (approved):

PR #14825, opened 2023-02-02
Fix fping 4.0+ prefers IPv6

TheMysteriousX:
Additionally fping6 is no longer a thing - this breaks IPv4 hosts entirely in a dual stack environment

It's not possible to reconcile these options, so we have to have a toggle until distro's with fping 3.x fall out of support.

Open to other ways to solve this - I couldn't think of anything other than invoking fping to check its version, which doubles the process overhead.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Data/Source/Fping.php
+++ b/LibreNMS/Data/Source/Fping.php
@@ -46,11 +46,13 @@ class Fping
         $interval = max($interval, 20);
 
         $fping = Config::get('fping');
+        $fping6 = Config::get('fping6');
         $fping_tos = Config::get('fping_options.tos', 0);
-        $cmd = [$fping];
+
         if ($address_family == 'ipv6') {
-            $fping6 = Config::get('fping6');
             $cmd = is_executable($fping6) ? [$fping6] : [$fping, '-6'];
+        } else {
+            $cmd = is_executable($fping6) ? [$fping] : [$fping, '-4'];
         }
 
         // build the command


TheMysteriousX:
This works for me - though as I've just decomissioned my pollers running EL7, I can't test fping 3.x

TheMysteriousX:
This is actually only a partial fix. app/Jobs/PingCheck.php uses a completely different code path, and is not easily fixable.

Whereas the poller/discovery scripts took into account the transport, just using the wrong CLI arguments, PingCheck completely ignores the device configuration as it just feeds a stream of hostnames to fping - so it's not possible to influence whether it should use v4 or v6.

End result is for devices that are configured for v4 polling but have v6 DNS set up and don't respond to ipv6 ping (e.g. Dell iDrac) will fall off monitoring.

TheMysteriousX:
This works as a fix for the poller - I've just disabled fast pings on our instance for now.

murrant:
Interesting, I just checked the device:add code path where we saw it not working on (CentOS) because it required fping6.  It was working because it wasn't detecting it as IPv6, so it was not sending the -6 option.

electrocret (approved):

electrocret:
@TheMysteriousX , can you correct the merge conflicts? Thanks

TheMysteriousX:
Will do.

murrant:
@TheMysteriousX I think this can be done without a setting.  Just check if fping6 has a binary.

TheMysteriousX:
Removed the config check, still appears to be working OK for me

murrant (approved):
LGTM

murrant:
Probably should have a way to control this manually (including snmp) at some point:
1. OS Default (6 -> 4)
2. IPv4 Only
3. IPv6 Only

PR #14830, opened 2023-02-04
Fix poller service logging configuration

TheMysteriousX:
A previous code cleanup moved these parameters into an 'if' block, but didn't account for the order they needed to be in to function.

This change re-orders the parameters so they work as intended again.

In addition, there were 3 ways logging could be set, with the hierarchy between them being something like

* Legacy Config
* Global Config
* CLI args
* Default

With per poller config being defined, but unused as it was parsed after the log level was set.

This change should also implement the following behaviour, which is opinionated, but I feel reasonable.

* CLI args set a minimum threshold - i.e. if you set -v, attempting to set the log filter to WARN, ERROR or CRIT via config will be ignored. If you don't want verbose output, don't set the flag.

Log level configuration will be parsed in this order:
* CLI Args
* Per-poller
* Global (new)
* Global (legacy)
* Default

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/queuemanager.py
+++ b/LibreNMS/queuemanager.py
@@ -493,7 +493,12 @@ class PollerQueueManager(QueueManager):
         if self.lock(device_id, timeout=self.config.poller.frequency):
             logger.info("Polling device {}".format(device_id))
 
-            exit_code, output = LibreNMS.call_script("poller.php", ("-h", device_id))
+            poller_args = ("-h", device_id)
+
+            if self.config.debug:
+                poller_args = ("-d", *poller_args)
+
+            exit_code, output = LibreNMS.call_script("poller.php", poller_args)
             if exit_code == 0:
                 self.unlock(device_id)
             else:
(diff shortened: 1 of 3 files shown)


ottorei (commented):
Code looks good to me. Did you test the possible config combinations with and without debugging?

ottorei (approved):

ottorei:
There seems to be an issue if the log level was not set. Just tested this on my system and the service is no longer starting.

`Feb 14 17:11:05 nms-poller-01 librenms-service.py[3471948]: MainThread(INFO):Initial log level was set to INFO
Feb 14 17:11:05 nms-poller-01 librenms-service.py[3471948]: MainThread(INFO):Configuring LibreNMS service
Feb 14 17:11:05 nms-poller-01 librenms-service.py[3471948]: MainThread(INFO):Attempting to load .env from '/opt/librenms/.env'
Feb 14 17:11:05 nms-poller-01 librenms-service.py[3471948]: 'ServiceConfig' object has no attribute 'poller_log_level'
Feb 14 17:11:06 nms-poller-01 systemd[1]: librenms.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Feb 14 17:11:06 nms-poller-01 systemd[1]: librenms.service: Failed with result 'exit-code'.
`

ottorei (changes requested):
I was having trouble starting multiple pollers with these changes in place. It seems that the `poller_log_level` can be accessed before it is defined.

TheMysteriousX:
Ah, that's my bad - that line works as expected if this was ansible :)

I'll fix it.

PipoCanaja:
Hi @TheMysteriousX 
As this PR seems to be stalled, we'll close it for now. Whenever you have time to continue it, feel free to re-open.
Thanx

Issue #13781, opened 2022-02-13
Paginate Pseudowires


WillIrvine:
### The problem

When viewing the global pseudowires list, all try and load at once.
When you have a few thousand of these the request times out before finishing

This issue is to request that these are paginated like the "Ports" or "Devices" list

### Output of ./validate.php

```txt
====================================
Component | Version
--------- | -------
LibreNMS  | 22.1.0
DB Schema | 2021_11_29_165436_improve_ports_search_index (229)
PHP       | 7.4.27
Python    | 3.6.8
MySQL     | 10.3.28-MariaDB
RRDTool   | 1.7.0
SNMP      | 5.8
====================================

[OK]    Composer Version: 2.2.6
[OK]    Dependencies up-to-date.
[OK]    Database connection successful
[OK]    Database schema correct
[INFO]  Detected Dispatcher Service
```

### Anything in the logs that might be useful for us?

```txt
2022/02/08 09:58:25 [error] 1274#0: *455409 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Maximum execution time of 30 seconds exceeded in /opt/librenms/vendor/laravel/framework/src/Illuminate/Support/Facades/DB.php on line 48" while reading response header from upstream, client: 10.1.5.5, server: librenms.local, request: "GET /pseudowires HTTP/1.1", upstream: "fastcgi://unix:/run/php-fpm-librenms.sock:", host: "librenms.local", referrer: "https://librenms.local/"
```

PR #14882, opened 2023-03-07
Pseudowires refresh

Bierchermuesli:
I did a small refresh on the pseudowire page:

- cut the sql queries in half if the remote endpoint is not known
- proper tables and out put if endpoint is not known
- pseudowire description, class, type and MTU is also visible
- visible port state
- less html table, common CSS styles

Im aware this page needs a general refactoring #13781, I took this as a mid step for how to present the data and check if we have the correct data at all (I assume there is a missing cleanup task for deleted hosts, PR will follow)

Before:
image

After:
223430454-7e500662-3563-4df3-924f-7a18c9eded90

Question:
- is someone into the minigraph view? (in my opinion this does not make much sense)
- does somebody know why there is no cisco-pw polling module?

#### Please note

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply 14882`

### DIFF
--- a/includes/html/pages/device/pseudowires.inc.php
+++ b/includes/html/pages/device/pseudowires.inc.php
@@ -49,7 +49,7 @@ foreach (dbFetchRows('SELECT * FROM pseudowires AS P, ports AS I WHERE P.port_id
         continue;
     }
 
-    # if the remote device is valid, resolve their pw & port details
+    // if the remote device is valid, resolve their pw & port details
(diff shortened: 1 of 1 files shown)


murrant:
Additionally, this is an old template file.  Ideally, it would be converted to blade and eloquent.  You could save a lot of queries by using eloquent to pre-fetch relationships.

murrant (approved):

PR #14883, opened 2023-03-08
Jira transport: use template title

not-known:
Use the title from the Alert Template for the Jira Alert Transport if it's not empty. Else use the default title as it was before the change.
This change gives the flexibility to customize the title of the created Jira issue. This way you can use custom variables in the title like sysLocation, sysName, etc.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Jira.php
+++ b/LibreNMS/Alert/Transport/Jira.php
@@ -52,7 +52,7 @@ class Jira extends Transport
         $password = $opts['password'];
         $prjkey = $opts['prjkey'];
         $issuetype = $opts['issuetype'];
-        $details = (!empty($obj['title'])) ? $obj['title'] : 'Librenms alert for: ' . $obj['hostname'];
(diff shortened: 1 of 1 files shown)


LibreNMS/Alert/Transport/Jira.php
```diff
@@ -52,7 +52,7 @@ public function contactJira($obj, $opts)
         $password = $opts['password'];
         $prjkey = $opts['prjkey'];
         $issuetype = $opts['issuetype'];
-        $details = 'Librenms alert for: ' . $obj['hostname'];
+        $details = ($obj['title'] != '') ? $obj['title'] : 'Librenms alert for: ' . $obj['hostname'];
```
ottorei, line 55: I think using `empty` might a better choice to test the string against.
See https://www.php.net/manual/en/function.empty.php

not-known:
Updated the statement!

ottorei (dismissed):

murrant (approved):

PR #14893, opened 2023-03-10
Add Cisco FTD Support for security appliances (x3)

mwobst:
Models: 4110, 4120, 4125 (each with K9 postfix)

This request is similar to https://github.com/librenms/librenms/pull/14872/

#### Please note

=> although nothing of this applies …


Jellyfrog (approved):

murrant (commented):
Seems like this is about due for a regex instead of explicit numbers.  LGTM

lukeofthetauri:
@murrant I agree- as I just added a device a couple weeks ago, I just worry about Cisco doing Cisco things, and throwing something silly in with this perhaps, since I don't see a rhyme or reason to the current numbering scheme.

PR #14914, opened 2023-03-20
Extend Avocent Support

electrocret:

#### Please note

#### Testers


includes/definitions/discovery/avocent.yaml
```diff
@@ -1,6 +1,11 @@
-mib: PM-MIB
+mib: PM-MIB:ACS-MIB
 modules:
     sensors:
+        pre-cache:
+            data:
+                -
+                    oid:
+                        - acsSerialPortTableName
```
murrant, line 8: I suggest caching ACS-MIB::acsSerialPortTableName instead.  Reduces issues with mib search.  (You'll have to update the usage below too.

electrocret: Done

electrocret:
Updated existing tests for ACS6048

murrant:
Question.  If you deleted the previous test case, how do we know you didn't break it or what changed?

electrocret:
None of the existing YAML was changed, so no less data would be collected, and both tests were performed on an ACS6048, so existing data points should be there, just different values cause they've been configured/used differently by the user. Would we prefer to have multiple tests for the same device?

murrant:
No and yes. More variants = tests take more time for possibly no gain.
But it is extremely difficult if not very time consuming to determine if the new test data actually covers all the same cases as the existing data.

One way to do it would be to separate the code and tests data consolidation.  That way with no code changes, the new test data is more likely to be correct.

Minimizing changes in the snmprec can be another good way to reduce noise.  But has most of the same issues and could be just invalid data with the other data.

murrant:
FYI, this PR is likely fine, just taking a second to reflect on this.

electrocret:
@murrant Bump?

murrant (approved):

PR #14915, opened 2023-03-20
Add Quantum Scalar I6000 Tape Library

mpikzink:

#### Please note

#### Testers


mpikzink:
Can someone help me to figure out why the tests fails? Do I have to update something?

electrocret:
The step in the test units that are failing are the dev:check CI. From my understanding - this check puts the SNMPRec data provided into SNMPSim to simulate the device. It then has Libre discover the simulated device then compares what Libre found with control data that is captured in the JSON file. The fail output you're getting says there is a mismatch between what Libre is seeing from the simulation and the control data in the JSON.

@murrant or @Jellyfrog, can you confirm my understanding?

murrant:
Correct @electrocret (maybe the docs need extra clarification).

Let's look at a little snippet:
```
-            'sensor_current' => 21
-            'sensor_limit' => null
-            'sensor_limit_warn' => 60
-            'sensor_limit_low' => null
-            'sensor_limit_low_warn' => 10
+            'sensor_current' => 21.0
+            'sensor_limit' => 41.0
+            'sensor_limit_warn' => 60.0
+            'sensor_limit_low' => 11.0
+            'sensor_limit_low_warn' => 10.0
```

Unfortunately, there is noise here.  For the code and tests 21 == 21.0, but it still shows up in the error output making it harder to read.

So the only changes here are `sensor_limit` and `sensor_limit`, they are now 41 and 11 respectively, but the existing test data expects them to be undefined.

Your code added definitions for these values now, so this seems like the test data (json file) just needs to be updated to include the new values.

includes/definitions/quantum.yaml
```diff
@@ -0,0 +1,13 @@
+os: quantum
+type: server
+text: Quantum
+mib_dir: adic
+ifXmcbc: true
+ifname: true
+poller_modules:
```
murrant, line 7: Disabling poller modules is not what you want.  They should be skipped if there is no discovered data for the devices anyway.

electrocret:
> Correct @electrocret (maybe the docs need extra clarification).
> 
 Hint hint, nudge nudge lol. 
I'll look over the docs and update whenever I have time. :)

murrant:
@mpikzink FYI, in upstream several unused port fields have been deleted too.

murrant (approved):

PR #14928, opened 2023-03-29
Stacked graphs on zoom to speed

electrocret:
Zoom to speed doesn't scale the egress properly on inverse graphs, however if it were to scale it properly, the graphs would be small and unreadable. So best course of action is to stack when Zoomed to speed.

Edit:
While testing this out, I found some issues in the logic, probably from changes in PHP. I fixed them.
Edit2:
First Edit was fixed with #14967

#### Please note

#### Testers

### DIFF
--- a/includes/html/functions.inc.php
+++ b/includes/html/functions.inc.php
@@ -1010,9 +1010,9 @@ function get_oxidized_nodes_list()
  * @param  string  $transparency  value of desired transparency applied to rrdtool options (values 01 - 99)
  * @return array containing transparency and stacked setup
  */
-function generate_stacked_graphs(force_stack = false,$transparency = '88')
+function generate_stacked_graphs($force_stack = false ,$transparency = '88')
 {
-    if (Config::get('webui.graph_stacked') == true or force_stack == true) {
+    if (Config::get('webui.graph_stacked') == true or $force_stack == true) {
         return ['transparency' => $transparency, 'stacked' => '1'];
     } else {
         return ['transparency' => '', 'stacked' => '-1'];
--- a/includes/html/functions.inc.php
+++ b/includes/html/functions.inc.php
@@ -1010,7 +1010,7 @@ function get_oxidized_nodes_list()
  * @param  string  $transparency  value of desired transparency applied to rrdtool options (values 01 - 99)
  * @return array containing transparency and stacked setup
  */
-function generate_stacked_graphs($force_stack = false ,$transparency = '88')
+function generate_stacked_graphs($force_stack = false, $transparency = '88')
 {
     if (Config::get('webui.graph_stacked') == true or $force_stack == true) {
         return ['transparency' => $transparency, 'stacked' => '1'];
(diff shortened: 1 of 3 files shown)


electrocret:
Fixed Toggle text logic in [f7e18fb](https://github.com/librenms/librenms/pull/14928/commits/f7e18fbbb370f38c6945b0bc9c2d4a7fc8919dc0)

includes/html/pages/graphs.inc.php
electrocret, line 1: Fixed Toggle text logic

includes/html/graphs/generic_data.inc.php
electrocret, line 1: Fixed some of the toggle logic. Wouldn't toggle off if port_speed_zoom was defined at all.

includes/html/functions.inc.php
```diff
@@ -1010,9 +1010,9 @@ function get_oxidized_nodes_list()
  * @param  string  $transparency  value of desired transparency applied to rrdtool options (values 01 - 99)
  * @return array containing transparency and stacked setup
  */
-function generate_stacked_graphs($transparency = '88')
+function generate_stacked_graphs($force_stack = false, $transparency = '88')
 {
-    if (Config::get('webui.graph_stacked') == true) {
+    if (Config::get('webui.graph_stacked') == true or $force_stack == true) {
```
murrant, line 1015: use the `||`  instead of or (and take a break from python for a bit.

includes/html/graphs/generic_data.inc.php
```diff
@@ -18,7 +18,7 @@
 
 require 'includes/html/graphs/common.inc.php';
 
-$stacked = generate_stacked_graphs();
+$stacked = generate_stacked_graphs(! empty($port['ifSpeed']) && ($vars['port_speed_zoom'] ?? Config::get('graphs.port_speed_zoom')) == 1);
```
murrant, line 21: I don't think the `== 1`  actually does anything in these.  The only case where it does is if someone sets port_speed_zoom=0 on the url, in that case I don't think we should respect it.

It might be more obvious to cast to a bool

murrant: > '1' == true
= true

> '0' == true
= false

> '1' == 1
= true

> '0' == 1
= false

electrocret: When I added it, the toggle started working correctly. 🤷

I'll try replacing it with true.

murrant: Give me an example graph and I'll test it.

electrocret: Go to a random port in an instance. Click the zoom to speed link on, then click it back off. As long as the Zoom to speed toggle is in the URL, even when off (0), it will still be zoomed to speed. When the "==1" is added it only does zoom to speed when URL toggle is 1.

murrant: I dumped the value on a graph and when it is port_speed_zoom=0, for some reason the value in $vars['port_speed_zoom'] is "yes".

murrant: Fixed, now none of the == 1 are needed
#14967 14967

electrocret: Updated my PR. I've applied your url handling fix, and my zoom to speed fix, and it's working as expected!

murrant (approved):

PR #14947, opened 2023-04-07
set focus to Global Search on Ctrl-F

Bierchermuesli:
Simple set the focus the Global Search form in the upper right corner if the user press Ctrl-F.
As this is the main search field and many of my users hits the search function when entering LibreNMS - this should make sense. 
Other option would be a `autofocus` attribute on the field but this might clash with other forms. 

As java is not my native language any feedback or better ideas are welcome. 

Tested with
- Brave Browser 111.1.49.132 
- Mozilla Firefox 111.0.1
- Google Chrome 112.0.5615.49 

#### Please note

#### Testers

### DIFF
--- a/resources/views/layouts/menu.blade.php
+++ b/resources/views/layouts/menu.blade.php
@@ -760,4 +760,22 @@
             });
         }
     @endif
+
+    $(document).ready(function(){
+    // Function to focus Global Search on Ctrl-F 
+    window.addEventListener("keydown",function (e) {
+        if (e.keyCode === 114 || (e.ctrlKey && e.keyCode === 70)){
+            if($('#gsearch').is(":focus")) {
+                //allow normal Ctrl-F on a 2nd Hit
+                return true;
+            } else {
+                //set Focus on Global Search and ignore Browsers defaults
+                e.preventDefault();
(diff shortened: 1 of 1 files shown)


fbouynot:
I personnally use the browser search and would like to keep it. Does it stop me to use it?
If so, is it possible to choose another shortcut? Is this a huge gain to avoid a click, or is there some accessibility improvements ?

Bierchermuesli:
Valid point, I can understand. You can hit CTRL-F a second time and the normal browser search will work. This is common behavior like Discourse forums, git etc also uses

better use autofocus ? any other votes?

fbouynot:
OK my bad, I didn't know about the second CTRL+F, I'm fine with it then.

electrocret:
> I personnally use the browser search and would like to keep it. Does it stop me to use it?
> If so, is it possible to choose another shortcut? Is this a huge gain to avoid a click, or is there some accessibility improvements ?

You are not alone. I regularly use the browser search.

Hitting CTRL + F twice for browser search works.

electrocret (approved):

murrant:
Nice @Bierchermuesli I was actually thinking we should do this just the other day.

PR #14953, opened 2023-04-07
Correct Temp Detection for Cisco SB Switches

Fehler12:
Added Temp high and low warning (otherwise librenms will guess some values and added Fan speed sensor

#### Please note

#### Testers


includes/definitions/discovery/ciscosb.yaml
```diff
@@ -71,8 +71,29 @@ modules:
                     index: '{{ $index }}'
                     descr: 'FAN {{ $index }} Temperature'
                     user_func: fahrenheit_to_celsius
+                    high_limit: 50 # true for all switches
+                    low_limit: -5 # true for nearly all switches except CBS250-8T-D and CBS250-8PP-D, they would like 0°C
```
electrocret, line 75: The default setting should be inclusive for all models of this equipment.

electrocret (changes requested):

tests/snmpsim/ciscosb_cbs250-24p-4x.snmprec
```diff
@@ -2058,15 +2058,15 @@
 1.3.6.1.2.1.31.1.1.1.18.3|4|
 1.3.6.1.2.1.31.1.1.1.18.4|4|
 1.3.6.1.2.1.31.1.1.1.18.5|4|
-1.3.6.1.2.1.31.1.1.1.18.6|4x|447275636B657220536368756C62C3BC0a373220364620
+1.3.6.1.2.1.31.1.1.1.18.6|4x|Drucker
```
PipoCanaja, line 2061: Hi @Fehler12 
Did you manually change this or did you use the script ? I would say that this line expects HEX data and not string. So "Drucker" should be `447275636B6572`.

PipoCanaja (changes requested):

includes/definitions/discovery/ciscosb.yaml
Fehler12: thank you for your feedback, I've changed it.

tests/snmpsim/ciscosb_cbs250-24p-4x.snmprec
Fehler12: yes and no: For values with only normal characters the SNMPREC gives a cleartext value. For values with special characters (umlauts in my case) it seems that they get decoded in HEX-format. To make it easier for me I copied a value of another network port to this port because it seems to me that your test systems and my system handle umlauts differently and therefore the checks fail.

Fehler12:
I've observed another issue which I would like to cope with this change as well:
sometimes the switches don't report any valid value back for temp and fan-speed (in both cases read value = 0),
In both cases this means that an event is triggered because this value lays outside of the allowed range (fanspeed is 500 at minimum, for temperature the value gets first converted to °C and then -17.78° is lower than 0). I tried to except these false values with defining skip_values but is seems that if once the value in initiated the skip_values are getting ignored. Any idea on how I could fix this?

PipoCanaja:
> I've observed another issue which I would like to cope with this change as well: sometimes the switches don't report any valid value back for temp and fan-speed (in both cases read value = 0), In both cases this means that an event is triggered because this value lays outside of the allowed range (fanspeed is 500 at minimum, for temperature the value gets first converted to °C and then -17.78° is lower than 0). I tried to except these false values with defining skip_values but is seems that if once the value in initiated the skip_values are getting ignored. Any idea on how I could fix this?

`skip_value` only takes place during discovery, which is expected. If the switch replies right during discovery, and wrong during polling, then there is a problem that should be alerted, with temperature and/or fan (or there is a problem in the switch reachability, or in the switch SNMP reply). We can't filter those as they are not "valid" behaviours.

Fehler12:
This is the output in error case (temperature in F)

[root@localhost ~]# snmpget -v2c -c public IP.IP.IP.IP 1.3.6.1.4.1.9.6.1.101.83.5.1.1.2.1
SNMPv2-SMI::enterprises.9.6.1.101.83.5.1.1.2.1 = INTEGER: 0

PR #14959, opened 2023-04-12
Fixes #14957 missing IPv6 BGP peers

### LINKED ISSUE #14957: Polling IPv6 BGP peers fails for Cisco devices with IPv6 peers in VRF, but no IPv6 peers in default VRF
### The problem

( related to #14105 , but for IPv6 BGP peers)

If a Cisco device has VRFs defined and IPv6 BGP peers configured in those VRFs, but no IPv6 BGP sessions/peers in the default VRF,  a SNMP (v3) walk without context will not find any cbgpPeer2RemoteAs. The code at the top of `includes/polling/bgp-peers.inc.php` populating $peer_data_check will return an empty Array:
```php
$peer_data_check = snmpwalk_cache_oid($device, 'cbgpPeer2RemoteAs', [], 'CISCO-BGP4-MIB');
```

This causes the code near line 34 that checks for an empty $peer_data_check to set the device type/class to generic:
```php
        if (empty($peer_data_check)) {
             $peer_data_check = snmpwalk_cache_oid($device, 'bgpPeerRemoteAs', [], 'BGP4-MIB');
             $generic = true;
         }
```

IPv6 BGP peers will never be polled, because the network OS specific code is skipped due to the `if ($generic) {` at the top of the checks.  The `} elseif ($device['os_group'] == 'cisco') {` further below will never execute, and instead of using the CISCO-BGP4-MIB, the code will use the default BGP4-MIB, which won't find the IPv6 peers.

### Output of ./validate.php

```txt
===========================================
Component | Version
--------- | -------
LibreNMS  | 23.4.0 (2023-04-07T01:52:52+02:00)
DB Schema | 2023_03_14_130653_migrate_empty_user_funcs_to_null (249)
PHP       | 8.2.4
Python    | 3.9.14
Database  | MariaDB 10.5.16-MariaDB
RRDTool   | 1.7.2
SNMP      | 5.9.1
===========================================

[OK]    Composer Version: 2.5.4
[OK]    Database Schema is current
[OK]    SQL Server meets minimum requirements
[OK]    lower_case_table_names is enabled
[OK]    MySQL engine is optimal
[OK]    Database and column collations are correct
[OK]    MySQl and PHP time match
[OK]    Distributed Polling setting is enabled globally
[OK]    Connected to rrdcached
[OK]    Active pollers found
[OK]    Dispatcher Service is enabled
[OK]    Locks are functional
[OK]    No python wrapper pollers found
[OK]    Redis is functional
[OK]    rrdtool version ok
[OK]    Connected to rrdcached
        [FIX]: 
        Modified Files:
         includes/polling/bgp-peers.inc.php
```

_No response_

_No response_

ajsiersema:
Polling IPv6 peers in VRF(s) on Cisco devices with no IPv6 peers in the
default VRF failed, because the code path will use $generic instead of the
network OS specific if section.  This PR fixes that issue.

NB: although the code changes the polling, it doesn't add or change the SNMP
queries or parsing thereof.  Changing to tests/data/pfsense_frr-bgp.json
shouldn't be necessary.

Fixes #14957 ( related to #14105 , but for IPv6 BGP peers )

#### Please note

#### Testers

### DIFF
--- a/includes/polling/bgp-peers.inc.php
+++ b/includes/polling/bgp-peers.inc.php
@@ -39,7 +39,8 @@ if (! empty($peers)) {
     // if it has a switch configured SNMP context.
     // The issues occured on NX-OS (Nexus) and IOS-XR (ASR) devices.
     // Using os_group 'cisco' breaks the 3560g snmpsim tests.
-    $cisco_with_vrf = (($device['os'] == 'iosxr' || $device['os'] == 'nxos') && count(DeviceCache::getPrimary()->getVrfContexts()) > 0);
+    $vrf_contexts = DeviceCache::getPrimary()->getVrfContexts();
+    $cisco_with_vrf = (($device['os'] == 'iosxr' || $device['os'] == 'nxos') &&  (count($vrf_contexts) > 1 || ! empty($vrf_contexts[0])));
     if (empty($peer_data_check) && ! $cisco_with_vrf) {
         $peer_data_check = snmpwalk_cache_oid($device, 'bgpPeerRemoteAs', [], 'BGP4-MIB');
         $generic = true;
--- a/includes/polling/bgp-peers.inc.php
+++ b/includes/polling/bgp-peers.inc.php
@@ -30,7 +30,15 @@ if (! empty($peers)) {
     } else {
         $peer_data_check = snmpwalk_cache_oid($device, 'bgpPeerRemoteAs', [], 'BGP4-MIB');
     }
-    if (empty($peer_data_check)) {
+    // If a Cisco device has BGP peers in VRF(s), but no BGP peers in
+    // the default VRF: don't fall back to the default MIB, to avoid
+    // skipping IPv6 peers (CISCO-BGP4-MIB is required.)
+    // count(getVrfContexts) returns VRF's with a configured SNMP context
+    // e.g. snmp-server context context_name vrf vrf_name
+    // "> 0" because the default VRF is only included in the count,
+    // if it has a switch configured SNMP context.
+    $cisco_with_vrf = ($device['os_group'] == 'cisco' && count(DeviceCache::getPrimary()->getVrfContexts()) > 0);
+    if (empty($peer_data_check) && ! $cisco_with_vrf) {
         $peer_data_check = snmpwalk_cache_oid($device, 'bgpPeerRemoteAs', [], 'BGP4-MIB');
         $generic = true;
     }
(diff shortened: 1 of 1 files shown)


ajsiersema:
I'm experiencing a bit of a deja vu. These SQL errors appeared during #14105 as well, but I can't quite figure out how they were resolved back then...

test (8.1, Unit, 1, mariadb:10.5)
Incorrect integer value: '' for column `librenms_phpunit_78hunjuybybh`.`bgpPeers`.`bgpPeerInUpdateElapsedTime` at row 2 

test (8.1, Unit, 1, mysql:8.0)
Incorrect integer value: '' for column 'bgpPeerInUpdateElapsedTime' at row 2

murrant:
Likely you are sending '' for those, you need to return an integer or null.

ajsiersema:
I haven't changed any of the tests, not sure where this came for or where to start looking

murrant:
Look at the code you added first.

ajsiersema:
Tried with snmpsim, the 3560g test case doesn't work with the Cisco BGP MIB.
I can't confirm the issue is present for IOS-XE/classic, I can however confirm the issue is very much present for NX-OS and IOS-XR. And that this fixes the issue and doesn't break the poller for these OS's. As such, I've limited the exception to those specific Cisco OS's.

JohnnyBBravo:
All checks passed :-)

JohnnyBBravo:
Is anyone able to review / approve this PR?
I'm also affected by this and would really like to see this implemented :-)

murrant:
The fix doesn't seem correct to me.

For one thing, this always evaluates to true:
`count(DeviceCache::getPrimary()->getVrfContexts()) > 0`

ajsiersema:
> The fix doesn't seem correct to me.
> 
> this always evaluates to true: `count(DeviceCache::getPrimary()->getVrfContexts()) > 0`

If there are no VRF (contexts) surely the count will be 0 and thus the check false ?
See the comments in the code.
> 1 doesn't fix the issue for us, because the default VRF isn't counted if there is no SNMP context defined .
I.e. if there is one (1) VRF defined besides the default VRF, the BGP peers are only in that VRF, and there isn't any SNMP context configuration for the default VRF because it isn't needed ("snmp-server context context_name vrf vrf_name"), then the count will be 1. I'll gladly change it to >= 1 if you like that better.

> For one thing
Any other problems I can expand on ?

murrant:
Side note, the bgp module is a spaghetti nightmare mess.

ajsiersema:
> getVrfContexts always returns the default context if there are no contexts. It cannot return 0.
> So, you would need to differentiate between the default context and custom contexts.

I tried on a switch with only the default (and a management) VRF, getVrfContexts() returns:
  
```php
array(1) {
  [0]=>
  string(0) ""
}
```
how about instead of checking for >0 (of >= 1), checking for a non-empty [0] from getVrfContexts() ?
Something like:

```php
$vrf_contexts = DeviceCache::getPrimary()->getVrfContexts();
$cisco_with_vrf = (($device['os'] == 'iosxr' || $device['os'] == 'nxos') && (count($vrf_contexts) > 1 || ! empty($vrf_contexts[0]));
```

murrant:
That would work, but you don't need the count() anymore, it is redundant.

JohnnyBBravo:
Does someone have time to review this PR please?
Looking forward to this being merged :-)

JohnnyBBravo (approved):
Tested and works for me :-)

PipoCanaja (approved):
LGTM

PR #2974, opened 2016-02-09
allow to set SNMPv3 context name per device

ohrensessel:
Context name is used as parameter to snmpwalk/get/...
Helpful for scenarios where SNMPv3 contexts are used for proxying

This interferes with #2820: Schema change and changes in functions/snmp.inc.php are partly taken from there. Did not see a nice possibility to cherry-pick, sorry.

Let me know what I can improve.

PR #14973, opened 2023-04-17
Added feature to set SNMPv3 context name for device

ajsiersema:
This basically reopens on the abandoned https://github.com/librenms/librenms/pull/2974, allowing to set the SNMPv3 context name for a specific device.

BGP/OSPF polling already use the SNMPv3 context name for VRF support.
The BGP polling code passes the context name by overwriting $device['context_name']. which the current snmp.inc.php already uses, i.e. no changes needed in snmp.inc.php

#### Please note

#### Testers

### DIFF
--- a/database/migrations/2023_04_17_120044_device_add_snmpv3_context_name_field.php
+++ b/database/migrations/2023_04_17_120044_device_add_snmpv3_context_name_field.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class DeviceAddSnmpv3ContextNameField extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('devices', function (Blueprint $table) {
+            $table->string('context_name', 64)->nullable()->after('cryptoalgo');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('devices', function (Blueprint $table) {
+            $table->dropColumn(['context_name']);
+        });
+    }
+}
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -322,6 +322,7 @@ function deviceArray($host, $community, $snmpver, $port = 161, $transport = 'udp
         $device['authalgo'] = $v3['authalgo'];
         $device['cryptopass'] = $v3['cryptopass'];
         $device['cryptoalgo'] = $v3['cryptoalgo'];
+        $device['context_name'] = $v3['context_name'];
     }
 
     return $device;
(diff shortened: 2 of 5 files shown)


murrant:
Thanks, this is definitely needed for some devices. (Such as printers)

1. You forgot to update NetSnmpQuery
2. This should be configurable per-OS

ajsiersema:
> 1. You forgot to update NetSnmpQuery
Could you help me out how to fix this ? I see there is a `NetSnmpQuery->context("...")`, and FacadeAccessorSnmp returns the class, as well as something in the tests, but I don't quite see how to get a device or os context into this ?

> 2. This should be configurable per-OS
Would this be best modeled in e.g. `config['os']['eg_a_printer']['snmp_context'] = "myContext"` or should the `config` table be used?

> 3. It doesn't seem to handle putting it back to the device/os value when the module overrides it.
The OSPF and BGP modules in this case ? I concur that assumptions are dangerous, but my assumption was that devices that use modules with a SNMP context override in their modules don't actually need the explicit context configuration.

murrant:
> > 1. You forgot to update NetSnmpQuery
> >    Could you help me out how to fix this ? I see there is a `NetSnmpQuery->context("...")`, and FacadeAccessorSnmp returns the class, as well as something in the tests, but I don't quite see how to get a device or os context into this ?
> 

It would need to be added during the command building part. Two options:
1. Set $this->context in __construct()
2. Set the context in buildAuth()

I think 1 is better. 

When context is explicitly set and snmp is v2, we append it to the community.  That may not be desirable for your case, so maybe need to change some semantics there.

> > 2. This should be configurable per-OS
> >    Would this be best modeled in e.g. `config['os']['eg_a_printer']['snmp_context'] = "myContext"` or should the `config` table be used?
> 

They are the same.  os config structure is defined in os_schema.json

> > 3. It doesn't seem to handle putting it back to the device/os value when the module overrides it.
> >    The OSPF and BGP modules in this case ? I concur that assumptions are dangerous, but my assumption was that devices that use modules with a SNMP context override in their modules don't actually need the explicit context configuration.

I'm saying when you get to one of those modules. It will set the context to something else, then never set it back to the context you have defined.

ajsiersema:
Is there an active use case for appending `@$this->context` for v2c in NetSnmpQuery ? Granted, appending this way passes a different community which is common practice in reading data from different instances with v2c, but a **context** is a SNMPv3 concept. The current implementation seems to overload "context", which kind of muddles the waters a bit. But of course, I'm not aware of the design decisions that went into the current implementation.

If we need to keep the current implementation, it looks like additional code needs to be added to NetSnmpQuery to signal that the context is only to be used for v3 (feels kinda ugly).

ajsiersema:
I won't be able to work on this until the first week of September, but I'll definitely return to this one at that stage.
Please keep this PR open until then

polskikrol:
I am also looking for an update on this request. Jetdirect printers cannot be polled via snmp v3 without the specifying of the context. Works fine via snmpwalk when specifying it.

PipoCanaja:
Hi @ajsiersema 
As this PR seems to be stalled, we'll close it for now. Whenever you have time to continue it, feel free to re-open.
Thanx

PR #14987, opened 2023-04-20
Add a validation for checking if the scheduler is running

Jellyfrog:
This is just preparation for future use.

#### Please note

#### Testers


LibreNMS/Validations/Scheduler.php
```diff
@@ -0,0 +1,44 @@
...
+     * Validate this module.
+     * To return ValidationResults, call ok, warn, fail, or result methods on the $validator
+     *
+     * @param  Validator  $validator
+     */
+    public function validate(Validator $validator): void
+    {
+        if (! Cache::has('scheduler_working')) {
+            $validator->fail('Scheduler is not running!',
+                'cp /opt/librenms/librenms.nonroot.cron /etc/cron.d/librenms');
```
murrant, line 40: This fix won't be correct if they have disabled some of the cron for the dispatcher.

Maybe sed?

murrant: or a separate cron file.

Jellyfrog: > This fix won't be correct if they have disabled some of the cron for the dispatcher.
> 
> Maybe sed?

TLDR; no hacks with sed plz. The source file contains the row for this scheduler, we can't really do much about users editing files, there is enough of warnings in place for that already?

Since this could/should replace all/most of the cron file in the future, maybe we can do separate ones.
I'll fix.

murrant:
Honestly, we should switch them all to timers.  (but keep the cron file around for non-systemd systems)

murrant (approved):

PR #15009, opened 2023-04-27
Updated WebUI config options in docs to support direct linking to installs

laf:
Updated documentation into new format for quick access to peoples installs.

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Distributed-Poller.md
+++ b/doc/Extensions/Distributed-Poller.md
@@ -59,7 +59,9 @@ The following is a standard config, combined with a locking mechanism below:
 If you want to customise the hostname for the poller then you will need
 to set this in `config.php`:
 
-`lnms config:set distributed_poller_name 'custom'`
+```php
+$config['distributed_poller_name']           = php_uname('n');
+```
 
 ## Locking mechanisms
 Pick one of the following setups, do not use all of them at the same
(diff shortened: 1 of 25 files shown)


doc/Extensions/Distributed-Poller.md
```diff
@@ -41,23 +41,26 @@ These requirements are above the normal requirements for a full LibreNMS install
...
+!!! setting "poller/distributed"
+    ```bash
+    lnms config:set distributed_poller true
+    lnms config:set distributed_poller_group 0
+    ```
+
+If you want to customise the hostname for the poller then you will need
+to set this in `config.php`:
+
+`lnms config:set distributed_poller_name 'custom'`
```
murrant, line 62: This one actually needs to be set in config.php otherwise it applies to all nodes.

Jellyfrog: it should go in .env in that case?

murrant: yes, and there is an .env for the dispatcher, but not the python wrapper.

Maybe it is time to deprecate using the python wrapper for distributed polling?

murrant:
Looks good overall.  Two thoughts:
1. Did you test running every lnms config:set command?  If not I can.
2. I noticed we lost some variable documentation in comments.  I think it might be an good idea to implement `lnms config:help` to spit out the translatable help text from the webui.

doc/Extensions/Distributed-Poller.md
laf: Well spotted, I've reverted that change. Didn't change it elsewhere, must have just got over zealous with this one!

murrant (approved):

murrant (approved):

PR #15020, opened 2023-05-01
Add support for Okta Group claims to set user role

peejaychilds:

Allow the Socialite Okta support to ask for groups claim, and then use those okta groups to map to User level's

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Auth/SocialiteController.php
+++ b/app/Http/Controllers/Auth/SocialiteController.php
@@ -168,7 +168,7 @@ class SocialiteController extends Controller
             foreach ($scopes as $scope) {
                 foreach (Arr::wrap($attributes[$scope] ?? []) as $scope_data) {
                     $user->level = max($user->level,
-                        $this->getRoleAsLevel($groups[$scope_data]['role'] ?? 'none')
+                        $this->getRoleAsLevel($groups[$scope_data]['role'] ?? null)
                     );
                 }
             }
@@ -213,7 +213,7 @@ class SocialiteController extends Controller
         return ! empty($name) ? $name : '';
     }
 
-    private function getRoleAsLevel(string $role = 'none'): int
+    private function getRoleAsLevel(?string $role): int
     {
         return match ($role) {
             'admin' => 10,
--- a/app/Http/Controllers/Auth/SocialiteController.php
+++ b/app/Http/Controllers/Auth/SocialiteController.php
@@ -27,6 +27,7 @@ use App\Models\User;
 use Config;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Auth;
 use Laravel\Socialite\Contracts\User as SocialiteUser;
 use Laravel\Socialite\Facades\Socialite;
@@ -55,11 +56,9 @@ class SocialiteController extends Controller
         $driver = Socialite::driver($provider);
 
         // https://laravel.com/docs/10.x/socialite#access-scopes
-        if ($driver instanceof \Laravel\Socialite\Two\AbstractProvider &&
-            LibreNMSConfig::has('auth.socialite.scopes')) {
+        if ($driver instanceof \Laravel\Socialite\Two\AbstractProvider) {
             $scopes = LibreNMSConfig::get('auth.socialite.scopes');
-            if (is_array($scopes) &&
-                count($scopes) > 0) {
+            if (! empty($scopes) && is_array($scopes)) {
                 return $driver
                     ->scopes($scopes)
                     ->redirect();
@@ -119,6 +118,7 @@ class SocialiteController extends Controller
 
             Auth::login($user);
             $this->setLevelFromClaim($provider, $user);
+            $user->save();
 
             return redirect()->intended();
         } catch (AuthenticationException $e) {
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Auth/SocialiteController.php
```diff
@@ -52,6 +52,11 @@ public function redirect(Request $request, string $provider)
         // Re-store target url since it will be forgotten after the redirect
         $request->session()->put('url.intended', redirect()->intended()->getTargetUrl());
 
+        if( $provider == "okta" && LibreNMSConfig::get('auth.socialite.configs.okta.group_claim', false) ){
```
murrant, line 55: This feels out of place.
It is complaining below because the interface does not support "scopes".  So either check that it is the Okta driver, or that "scopes" exists.

I'm not sure if there is a generic way of doing this or not.

murrant (dismissed):

app/Http/Controllers/Auth/SocialiteController.php
```diff
@@ -123,9 +129,29 @@ private function register(string $provider): void
...
+        $user->level  =  intval( LibreNMSConfig::get('auth.socialite.defaultlevel', 0 ));
+        $this->setLevelFromGroupsClaim( $provider, $user );
+
         $user->save();
     }
 
+    private function setLevelFromGroupsClaim(string $provider, $user)
+    {
+        if( $provider == "okta" && LibreNMSConfig::get('auth.socialite.configs.okta.group_claim', false) ){
+            $user->level  =  intval( LibreNMSConfig::get('auth.socialite.defaultlevel', 0 ));
```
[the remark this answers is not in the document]
peejaychilds: ok added function

app/Http/Controllers/Auth/SocialiteController.php
```diff
@@ -123,9 +129,29 @@ private function register(string $provider): void
...
+            if( array_key_exists( 'groups', $this->socialite_user->user ) && is_array( $this->socialite_user->user['groups'] )){
+                $groups = $this->socialite_user->user['groups'] ;
+                foreach ( $groups as $groupname ){
+                    $newlevel = intval( LibreNMSConfig::get('auth.socialite.groups.' . $groupname . '.level' , 0 ));
```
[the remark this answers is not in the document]
peejaychilds: done

app/Http/Controllers/Auth/SocialiteController.php
peejaychilds: I don't believe there is a generic way, but I could be wrong.  checking 
```$driver instanceof \Laravel\Socialite\Two\AbstractProvider```
should mean that ->scopes() is valid ... but it could mean different things for different providers (probably) 

I would hope that if these are available for example from Git or Microsoft etc that the next person will deal with appropriately, and should some generic or repeated patterns appear then adjust as appropriate?

Jellyfrog (changes requested):
You need to remove the Okta specific code
And change the logic so you specify the scope name for groups instead of hardcoded "groups"

And last, need to make the group name mapping configurable

app/Http/Controllers/Auth/SocialiteController.php
murrant: I looked through providers the other day and there were very few with group support and the documentation of groups sucked.

Jellyfrog:
> @Jellyfrog shouldn't the "group" name line up with "role" names in LibreNMS? Or do want to require explicit mapping? Or should there be a "user group" concept in LibreNMS?

Well since we don't have groups in LibreNMS, I think this is the best way atm. We cant expect people to create a group in AD called "admin" for example :)

app/Http/Controllers/Auth/SocialiteController.php
peejaychilds: I looked at docs for GitHub and Azure AD as well.    It appears with Azure AD you can define 'roles' on the application and have these sent if the claim is 'role', and also a wait to have AD groups (potentially filtered) via 'groups'.

Since I don't have a Azure AD lying around I can't actually see what would be returned, so it might not work 'out-of-the-box' but it in theory shouldn't be too hard if the data is somewhere in the user structure.

murrant:
I think code design is good now.  Code style could use a clean up pass (I can tell you copied code from some older parts of LibreNMS).  If I submit a cleanup, would you test it after?

Jellyfrog:
I'll give it a go to try it out on some other provider.

peejaychilds:
I have put the refactored changes through our lower testing environments and it works as expected @murrant  🙏

peejaychilds:
I noticed the gui is borked for the Groups->Roles mapping.   

I assume perhaps have to make a new config type rather than use ldap-groups and then make some type of javascript helper function to help with the populating the select boxes?

We don't really ever look in the GUI but I guess it would be nice if it worked as expected.

Jellyfrog:
So I did some more testing and github doesn't work
And Google via saml2 doesnt work(but quite Close actually.)

This kinda means I'm against adding it in its current state, have to give it some thought.

peejaychilds:
> So I did some more testing and github doesn't work And Google via saml2 doesnt work(but quite Close actually.)
> 
> This kinda means I'm against adding it in its current state, have to give it some thought.

But it does work for Okta, which makes it pretty useful for shops with Okta with a fairly minor impact.   

I'm assuming getting all the other providers to work would involve some upstream work to get it all nicely abstracted to be generic

rgilijamse:
As I understand it, the problem is that claims/groups do not work for providers other than Okta. Would it otherwise be possible to separate just the 'Default Role' part of this PR and merge that? That should work for all providers, and is a useful feature on it's own.

WindoC:
Is there any plan to add the default_role feature on the next version?

Please add it is very useful for me. Thanks!

murrant:
@peejaychilds roles have been merged, can you rebase this?

PR #15022, opened 2023-05-02
Fix Firebrick mempool polling

cjsoftuk:
This makes a minor correction to my earlier PR which got merged.

As LibreNMS doesn't allow named OIDs in the mempools definition without it being indexed of some description, it appears we have to put the raw OID here.

I welcome corrections that allow named OIDs to work - but based on my reading of https://github.com/librenms/librenms/blob/master/LibreNMS/OS/Traits/YamlMempoolsDiscovery.php#L109, named OIDs without an index aren't supported.

#### Please note

#### Testers


cjsoftuk:
@electrocret - I thought there might be something not quite right with polling as discovery was fine and then polling was always showing 0% used.... This change fixes that issue, but not in the way I'd have liked to be doing it!

I don't like the fact that I can't use named OIDs, but it seems I have little choice based on the code.

I also need to fix the data for the tests....

cjsoftuk:
There is also something off by a factor of 1024 somewhere..... so give me an hour or two to fix that as well before merging please!

electrocret (approved):

PR #15028, opened 2023-05-09
fix for hrStorageIndex agility

peejaychilds:

It is possible that the HOST-RESOURCES-MIB::hrStorageTable has updated between discovery and polling. If we assume it hasn't we can end up assigning the values of the wrong mount point to another mount point. This can cause
issues with not only display, but also alarming.

We can compare the description expected with one from the hrStorageIndex entry and if they are not equal then we can filter the table to return an entry that does match.

We have found with JunOS EVO when anyone logs into the box it mounts
`tmpfs                   3.1G          0       3.1G        0%  /run/user/{uid}`
which causes the ordering of the HOST-RESOURCES-MIB::hrStorageTable to change which causes all sorts of issues ... (like Storage alarms because some partition just got muddled up with some other suppressed partition (ie a loopback mount) that has 0% free space)

#### Please note

#### Testers

### DIFF
--- a/includes/polling/storage/hrstorage.inc.php
+++ b/includes/polling/storage/hrstorage.inc.php
@@ -22,7 +22,7 @@ $entry = $storage_cache['hrstorage'][$storage['storage_index']];
  *
  */
 
-if( $entry['hrStorageDescr'] !== $descr ){
+if( $entry['hrStorageDescr'] !== $descr && is_array( $storage_cache['hrstorage'] ) ){
     d_echo( '🚨 We are after [' . $descr . '] but hrStorageIndex entry has [' . $entry['hrStorageDescr'] . '] ' );
     d_echo( 'Before: ' . var_export( $entry, true ));
     $entry = array_values( array_filter( $storage_cache['hrstorage'], function($entry) use ($descr) {
--- a/includes/polling/storage/hrstorage.inc.php
+++ b/includes/polling/storage/hrstorage.inc.php
@@ -15,7 +15,6 @@ if (($entry['hrStorageDescr'] ?? null) !== $storage['storage_descr']) {
     // couldn't find a match, check by storage_descr or fall back to data found by index in case descr changed
     $entry = collect($storage_cache['hrstorage'])->firstWhere('hrStorageDescr', $storage['storage_descr']) ?? $entry;
 }
-dump($entry);
 
 if (empty($entry)) {
     Log::error("%rCould not find storage data for {$storage['storage_descr']}%n", ['color' => true]);
--- a/includes/polling/storage/hrstorage.inc.php
+++ b/includes/polling/storage/hrstorage.inc.php
@@ -8,6 +8,31 @@ if (! isset($storage_cache['hrstorage'])) {
 
 $entry = $storage_cache['hrstorage'][$storage['storage_index']];
 
+/*
+ * It is possible that the HOST-RESOURCES-MIB::hrStorageTable has updated between
+ * discovery and polling.    If we assume it hasn't we can end up assigning the
+ * values of the wrong mount point to another mount point.   This can cause
+ * issues with not only display, but also alarming.
+ *
+ * We can compare the description expected with one from the hrStorageIndex entry
+ * and if they are not equal then we can filter the table to return an entry
+ * that does match.
+ *
+ * (example of how to debug --> ./poller.php -h 42 -r -f -m storage -d)
+ *
+ */
+
+if( $entry['hrStorageDescr'] !== $descr ){
+    d_echo( '🚨 We are after [' . $descr . '] but hrStorageIndex entry has [' . $entry['hrStorageDescr'] . '] ' );
+    d_echo( 'Before: ' . var_export( $entry, true ));
+    $entry = array_values( array_filter( $storage_cache['hrstorage'], function($entry) use ($descr) {
+        return $entry['hrStorageDescr'] === $descr;
+    }))[0];
+    d_echo( 'After: ' . var_export( $entry, true ) );
+    d_echo( '🚨 updated to [' . $entry['hrStorageDescr'] . ']' );
+}
+
+
 $storage['units'] = $entry['hrStorageAllocationUnits'];
 $entry['hrStorageUsed'] = fix_integer_value($entry['hrStorageUsed'] ?? 0);
 $entry['hrStorageSize'] = fix_integer_value($entry['hrStorageSize']);
--- a/includes/polling/storage/hrstorage.inc.php
+++ b/includes/polling/storage/hrstorage.inc.php
@@ -1,37 +1,33 @@
 <?php
 
+use Illuminate\Support\Facades\Log;
+
 // HOST-RESOURCES-MIB - Storage Objects
 if (! isset($storage_cache['hrstorage'])) {
     $storage_cache['hrstorage'] = snmpwalk_cache_oid($device, 'hrStorageEntry', null, 'HOST-RESOURCES-MIB:HOST-RESOURCES-TYPES');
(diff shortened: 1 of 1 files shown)


includes/polling/storage/hrstorage.inc.php
```diff
@@ -8,6 +8,31 @@
...
+ *
+ * We can compare the description expected with one from the hrStorageIndex entry
+ * and if they are not equal then we can filter the table to return an entry
+ * that does match.
+ *
+ * (example of how to debug --> ./poller.php -h 42 -r -f -m storage -d)
+ *
+ */
+
+if( $entry['hrStorageDescr'] !== $descr ){
```
murrant, line 25: If the data is already fetched, why not always updated it?

Jellyfrog:
Would it be possible to index on something else, like mont point instead? Guessing it breaks the design too much?

includes/polling/storage/hrstorage.inc.php
peejaychilds: I'm not sure.   

All the data is fetched so in theory a 'rediscovery' could be done each time to update the appropriate indexes etc in the database.

I assume it was done this way originally because
* it is lightweight and fast
* storage indexes on servers didn't change much

Which is probably a good set of assumptions, and this just adds a little check incase something has changed between discovery and polling (like temporary devices, user-login /tmpfs stuff, USBs, etc... ) that has caused the hrStorageIndex to change

peejaychilds:
> Would it be possible to index on something else, like mountpoint instead? Guessing it breaks the design too much?

HOST-RESOURCES-MIB::hrStorageTable doesn't have the mount point unfortunately.    

```
    29 => 
    array (
      'hrStorageIndex' => '29',
      'hrStorageType' => 'hrStorageFixedDisk',
      'hrStorageDescr' => 're0:/dev/sda6, mounted on /config',
      'hrStorageAllocationUnits' => '4096',
      'hrStorageSize' => '251878',
      'hrStorageUsed' => '375',
      'hrStorageAllocationFailures' => '0',
    ),
```

murrant:
Mounts changing hrStorageIndex is a really annoying bug in your snmpd implementation...

includes/polling/storage/hrstorage.inc.php
murrant: It was originally done because the MIB specifies that indexes must not change.

As far as the MIB is concerned a removable disk is a new disk if it has been removed and re-plugged in.

peejaychilds: I can't see that spelt out clearly in the MIB although it does state hrStorageIndex is "A unique value for each logical storage area contained by the host." so you could consider that it should persist.

Looking at net-snmp's snmpd on a recent Ubuntu it does not persist between restarts of snmpd (although interestingly net-snmp does persist ifIndexes in /var/lib/snmpd).
 
One behavioural difference with my tested net-snmp implementation is that users filesystems in /run/user/{uid} all appear inserted with hrStorageIndex's past existing index's rather than getting stuffed into the middle like JunOS EVO's snmpd implementation does.

We will obviously kick the vendor, however I think the patch has some value as a sanity check for the messy world we live in (and all those JunOS EVO versions out there that will never get upgraded etc...)

peejaychilds:
@murrant tested works ok

*Test with JunOS EVO box*

Pre-Test : Set device 42 into non-polling poller group? (use GROUP 5 TYO )
Test Command - ./poller.php -d -v -h 42 -m storage
Examine re0:/dev/sda1, mounted on /boot (id=1579)

*Control Test*
On one dispatcher perform the following
1. 'unpatch my patch' >> git checkout includes/polling/storage/hrstorage.inc.php
2. run ./poller.php -d -v -h 42 -m storage > hr-test1
3. log user into device
4. run ./poller.php -d -v -h 42 -m storage > hr-test2
5. log user out of device
6. run ./poller.php -d -v -h 42 -m storage > hr-test3

```
grep 1579] hr-te*
hr-test1:SQL[UPDATE `storage` set `storage_used`=?,`storage_free`=?,`storage_size`=?,`storage_units`=?,`storage_perc`=? WHERE `storage_id` = ? ["19242496","186197504","205440000",512,9,1579] 1.07ms]
hr-test2:SQL[UPDATE `storage` set `storage_used`=?,`storage_free`=?,`storage_size`=?,`storage_units`=?,`storage_perc`=? WHERE `storage_id` = ? ["39051264","0","39051264",8192,100,1579] 0.91ms]
hr-test3:SQL[UPDATE `storage` set `storage_used`=?,`storage_free`=?,`storage_size`=?,`storage_units`=?,`storage_perc`=? WHERE `storage_id` = ? ["19242496","186197504","205440000",512,9,1579] 1.03ms]
```

```
grep "/boot" hr-te* | grep :hrStorage
hr-test1:hrStorageDescr.25 = re0:/dev/sda1, mounted on /boot
hr-test2:hrStorageDescr.26 = re0:/dev/sda1, mounted on /boot
hr-test3:hrStorageDescr.25 = re0:/dev/sda1, mounted on /boot
```
Results -> Notice the index's change, and the storage_used, storage_free etc follow the index, so are set incorrectly in 'test2' for id=1597

*Patch Test*

On second dispatcher perform the following
1. 'unpatch' >> git checkout includes/polling/storage/hrstorage.inc.php
2. apply new patch via ./scripts/github-apply 15028
3. run ./poller.php -d -v -h 42 -m storage > hr-test1
4. log user into device
5. run ./poller.php -d -v -h 42 -m storage > hr-test2
6. log user out of device
7. run ./poller.php -d -v -h 42 -m storage > hr-test3

```
grep 1579] hr-te*
hr-test1:SQL[UPDATE `storage` set `storage_used`=?,`storage_free`=?,`storage_size`=?,`storage_units`=?,`storage_perc`=? WHERE `storage_id` = ? ["19242496","186197504","205440000",512,9,1579] 1.37ms]
hr-test2:SQL[UPDATE `storage` set `storage_used`=?,`storage_free`=?,`storage_size`=?,`storage_units`=?,`storage_perc`=? WHERE `storage_id` = ? ["19242496","186197504","205440000",512,9,1579] 0.46ms]
hr-test3:SQL[UPDATE `storage` set `storage_used`=?,`storage_free`=?,`storage_size`=?,`storage_units`=?,`storage_perc`=? WHERE `storage_id` = ? ["19242496","186197504","205440000",512,9,1579] 0.83ms]
```

```
grep "/boot" hr-te* | grep :hrStorage
hr-test1:hrStorageDescr.25 = re0:/dev/sda1, mounted on /boot
hr-test2:hrStorageDescr.26 = re0:/dev/sda1, mounted on /boot
hr-test3:hrStorageDescr.25 = re0:/dev/sda1, mounted on /boot
```

```
 grep quick hr*
hr-test2:Storage re0:/dev/sda5, mounted on /etc changed index 27 > 28, applying quickfix until discovery runs
hr-test2:Storage re0:/dev/sda6, mounted on /config changed index 28 > 29, applying quickfix until discovery runs
hr-test2:Storage re0:/dev/sda7, mounted on /var changed index 29 > 30, applying quickfix until discovery runs
hr-test2:Storage re0:/dev/sda2, mounted on /soft changed index 26 > 27, applying quickfix until discovery runs
hr-test2:Storage re0:/dev/loop0, mounted on /data/var/external changed index 9 > 10, applying quickfix until discovery runs
hr-test2:Storage re0:/dev/sda1, mounted on /boot changed index 25 > 26, applying quickfix until discovery runs
```

Results -> Notice the index's change HOWEVER the storage_used, storage_free etc follow the description, so are set correctly in 'test2' for id=1597

Post-Test set device back to poller group = 0

murrant (approved):

PR #15033, opened 2023-05-10
rutos-rutx: Map GPS lat and lon to os module

QuadPiece:
Teltonika routers have GPS hardware and it's discovered by LibreNMS, but for some reason only gets graphed.
This maps the coordinates to the OS module, which makes its position appear on the pretty map.

As the GPS coordinates are already there and discovered by LibreNMS, simply used erroneously, I do not believe updated test data is needed.

Web UI before:

before

Web UI after:

after

GPS coordinates were available before, but in the form of a graphed count:

firefox_AzrNaXsovj

This is still the case and this count has not been removed.

#### Please note

#### Testers

### DIFF
--- a/includes/definitions/discovery/rutos-rutx.yaml
+++ b/includes/definitions/discovery/rutos-rutx.yaml
@@ -4,6 +4,8 @@ modules:
         hardware: TELTONIKA-RUTX-MIB::routerName.0
         serial: TELTONIKA-RUTX-MIB::serial.0
         version: TELTONIKA-RUTX-MIB::fwVersion.0
+        lat: TELTONIKA-RUTX-MIB::latitude.0
+        long: TELTONIKA-RUTX-MIB::longtitude.0
     sensors:
         pre-cache:
             data:


murrant:
Great!

But because tests didn't fail, this means none of the test files include those two OIDs, can you add them to one of the snmprec and update the test data (or at least add example snmp data here).

hjcday:
How does this behave when locations change? I played with this a while ago and found there were some issues updating the lat/lon for the location associated with the device

QuadPiece:
@murrant I attempted to add test data for my device, but unfortunately failed. 

Turns out my device wasn't detected out of the box (RUT956 was added to the definitions file locally on our install). So I attempted to follow [the docs](https://docs.librenms.org/Developing/os/Test-Units/) and added my own device as a variant in `includes/definitions/rutos-rutx.yaml`, after which I ran `./scripts/collect-snmp-data.php` without issue and got an snmprec file.

But when running `./scripts/save-test-data.php -o rutos-rutx -v rut956`, it ran without reporting issues, however the json file it saved was blank: 

image

I most likely have done something wrong. But if it helps, this appears to be the GPS OIDs:

```
1.3.6.1.4.1.48690.3.1.0|4|59.912195 # Lat
1.3.6.1.4.1.48690.3.2.0|4|10.753871 # Lon
1.3.6.1.4.1.48690.3.5.0|4|10 # Number of GPS sattelites
```

This appears to already be present in an snmprec from a different device: https://github.com/librenms/librenms/blob/1f97b4491dd7bc0d45f82fa3b26eb5866382b30c/tests/snmpsim/rutos-rutx_rutx11.snmprec#L1331

Jellyfrog:
@murrant 
> But because tests didn't fail, this means none of the test files include those two OIDs

lat/long is not selected in the dump function for OS module :)

murrant:
Well shit :)   Ok lets merge this the snmp data is included.

murrant (approved):

PR #15045, opened 2023-05-17
adding human readable version of /api/v0/fdb/{mac}/detail

sjtarik:

#### Please note

#### Testers


ottorei:
In my opinion the idea for reducing API-calls and providing a single output with this kind of approach works for this particular use case but there are lots of other models and API-routes where this same problem would occur.

So the basic issue here is that each model has it's own API and if one wishes to have all the related data, multiple API-calls would be needed. I wonder if there is an elegant generic solution to produce single output with all or selected fields from any related models, basically a data API where queries could be made in a similar fashion than normal Eloquent queries that use relations. @Jellyfrog @murrant what do you think about this kind of data API idea?

tkaraoglu:
@Jellyfrog @murrant @ottorei folks, any chance we can land this? in our company we rely on this API. Other ways of pulling this data with existing APIs  times out (getting 500s) since mac and interface table is large, takes up large memory and existing APIs don't use any limit for the queries.

electrocret:
@sjtarik, can you add to the docs how to use this new API endpoint?

electrocret (approved):
LGTM

ottorei:
> @sjtarik, can you add to the docs how to use this new API endpoint?
> 
> @ottorei , I agree that the Libre API needs a major overhaul (v1 perhaps?).The current API is largely a carry over from pre-Libre. However this is what we currently have.

That's right. I don't see a problem adding additional "recursive" API-routes that resolve IDs and other data from different models into plain string representations. But if we are doing this, then we should probably do it in a unified way for all routes where this kind of behaviour is needed.

tkaraoglu:
@electrocret @ottorei thank you for your comments and reviews. Please let me know if I can contribute to larger effort for APIs in the future.

PR #15054, opened 2023-05-22
Add new option to discover rebooted devices

eskyuu:
I noticed that LibreNMS fails to correctly poll some resources after a device has been rebooted because some of the OIDs change.  This patch adds another option to the discovery php script to allow discovery of rebooted devices, which are detected by looking for devices that are online where the time since the last discovery was run is greater than the current uptime.

#### Please note

#### Testers

### DIFF
--- a/discovery.php
+++ b/discovery.php
@@ -37,6 +37,10 @@ if (isset($options['h'])) {
         $new_discovery_lock = Cache::lock('new-discovery', 300);
         $where = 'AND `last_discovered` IS NULL';
         $doing = 'new';
+    } elseif ($options['h'] == 'rebooted') {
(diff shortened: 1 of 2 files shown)


electrocret:
I believe there is already a discovery on reboot config option. It's disabled by default, so you may have missed it.

eskyuu:
Thanks for the pointer.  I found the code in the calculate uptime function that clears the last discovery time if this option is set and a reboot is detected, which would cause the device to be re-discovered as long as a "new" device.

electrocret:
I personally find the setting a double edged sword. When discovery happens, it can create alot of events in the event log. Which isn't ideal when you're going through the event log trying to figure out why this device rebooted.

PR #15059, opened 2023-05-24
Add support for PrimeKey Software Appliance

TheMysteriousX:
Only supports sensors. I've added support for all sensors except for some hardware specific HSM ones, as I do not have access to them to test.

For anyone using this:
* I don't know why internal database is persistently faulted
* I don't know why Utimaco HSM's are persistently faulted

Both are probably bugs in the appliance, or a check is failing that doesn't affect production usage. The sensors are reporting the data the appliance is returning accurately.

Also worth noting that this is *very* similar to the physical appliance OS definition - but due to the MIB name being different, the two are not interchangable.

(Back to draft as I haven't tested the tweaked detection rule on a real system)

#### Please note

#### Testers


includes/definitions/discovery/primekey-sw.yaml
```diff
@@ -0,0 +1,295 @@
+mib: PK-SOFTWARE-APPLIANCE-V2
+modules:
+    os:
+        version: PK-SOFTWARE-APPLIANCE-V2::pk-SAV2-systemVersion.0
+    sensors:
+        count:
```
murrant, line 6: This shouldn't be a count sensor, it should be storage.

TheMysteriousX: I'm not certain that it makes sense to implement this as a storage plugin, as this isn't measuring a filesystem.

I'm not entirely sure what it *is* measuring - I'm assuming there's some mechanism that limits the database to a fixed, possibly expandable file size, but it's not documented.

Happy either way though, it'd only be a few lines to convert it.

TheMysteriousX:
Forgot to take this out of draft, worksforme.

electrocret (approved):
LGTM

PR #15072, opened 2023-05-29
Add FDB table vendor search drop down

bonzo81:
The aim of this PR is to give the ability to search the FDB table by vendor name.

Some things to consider:
1. I could not see a way to use the existing OUIDB cache data to retrieve list of OUIs from vendor name. So new caching of data required.

2. In order to keep the existing OUI vendor lookup fast, the OUI data is stored again in cache in a separate array, with OUIs stored under a vendor name. This is duplicating data in cache and might be undesirable but as the current OUI lookup is used often in varies places in the app I didn't want to slow this down. 

3. As far as I understand this new cache data won't be created when running daily.sh until the 'refresh interval' expires i.e $mac_oui_refresh_int_min'. So from my experience applying this PR and running daily.sh doesn't immediately allow this feature to work. The 'refresh interval' needs to expire first. I might be missing something.

I'm not a developer so it's been a mission to get this working so please pick it apart. Happy to make amendments.

Example of it working in web GUI- partial search & multiple results:
vendor_search

#### Please note

#### Testers


murrant:
@bonzo81 thanks!  After your chats about this, I think it is appropriate for OUI to be stored in the database.  That way it can be queried.

bonzo81:
Okay thanks for that @murrant. Won't lie, working with DB is alien to me but I'll poke around, maybe ask ChatGPT some questions (only sort of kidding) . 

So if it's going in the DB we no longer need it in cache too?

At this point then, I'm thinking I need to create a database migration, and swap out the cache_mac_oui() function to one that stores the OUI data it into DB. I like the minimum refresh time in the current function as the data doesn't get updated that often so I might use that again. We don't need to rewrite the data every day. What do you think?

Will also need to rework my code and the existing readableoui().

That's enough for starters!

bonzo81:
This is my initial attempt at moving OUI data to DB.
* Created Migration for 'vendor_oui' table
* Updated daily functions to store data in DB - using same parsing logic as before.
* Used different cache locking names to allow new code to run without waiting for old timer to end.
* Updated all functions that previously pointed to Cache for OUI lookups to use DB.

Tested on my small dev box and working.

I see two failed test, not sure what I need to do to sort them.

murrant:
Looking good, you just need to run `lnms schema:dump` as described by the test output.

Maybe instead of the caching we use the new scheduling?
https://laravel.com/docs/10.x/scheduling

bonzo81:
Thanks @murrant. It would appear I made a mess of the migrations - I renamed my first attempt and made a mess! Figured out where I went wrong I have sorted that now. Ran 'lnms schema:dump' and tests look happy now.

Re: using scheduler. I guess that would make sense, and if that's the preferred way I'll take a look at it. I don't see anything using the scheduler that might give me a clue as to how to implement it, so I might need some pointers.

bonzo81:
@murrant I've got something working with the scheduler. I created a new command, added it to the schedule in kernel.php.  I can successfully run it manually as an artisan command it also runs via scheduler. If this sounds like the right approach, I'll commit the changes for review.

I was hoping to log the output of the scheduled task somewhere.
I've tried sending the output of the schedule entry to a file i.e. logs/scheduler.log but it's not working so far. Is there a better way to log the scheduler tasks output? I think it would be good to have this for troubleshooting later.

electrocret:
@bonzo81, are you still looking into getting this into the scheduler? 

I feel it'd be nice to get it into scheduler, but not absolutely necessary since scheduler's relatively new and nothing else has been converted over to it. Your work can be converted over whenever there's an effort to move existing tasks to it.

bonzo81:
Thanks @electrocret. I've got it working with scheduler. As there's nothing using the scheduler yet, I'm not sure if I've implemented it in the best way, but it's been running successfully for a while now. I could certainly commit those changes for review?

electrocret:
@bonzo81 , since the Scheduler piece is uncharted waters. I'll merge this PR as is with cron, and you can make a separate PR for changing it over to Scheduler. 

I don't want to hold this feature back due to a push to a possible debate on the best way to implement scheduler.

electrocret (approved):
Looks good!

tkaraoglu:
@electrocret @murrant looks like wireshark removed manuf file which mac vendor database based on. check [here](https://gitlab.com/wireshark/wireshark/-/raw/master/manuf)

mac processing code return success most of the cases so that's misleading. 

this will fix itself if wireshark foundation fixes this since a lot of people depends on that file (mac resolution websites, apps ...) however we might need to find an alternative.

bonzo81:
@tkaraoglu Good spot. 

Looks like the file has been reformatted and moved to https://gitlab.com/wireshark/wireshark/-/blob/master/epan/manuf-data.c

The function will need reworked to parse this new structure. If I get time I'll take a look, but not sure when.

tkaraoglu:
@bonzo81 I wonder if we can just migrate this [file](https://gitlab.com/wireshark/wireshark/-/blob/master/tools/make-manuf.py) into Libre repo and generate the manuf file in json format ourselves. Or directly dump this data to database. 

Other option can be to generate manuf file with each Libre version part of CI/CD.

bonzo81:
@tkaraoglu I'm in favour of any integration with libre but a little beyond my skills set at the moment.

I just discovered that you can download .csv direct from ieee. Not one file but one for each block size i.e. [MA-L](https://standards-oui.ieee.org/oui/oui.csv), [MA-M](https://standards-oui.ieee.org/oui28/mam.csv), and [MA-S](https://standards-oui.ieee.org/oui36/oui36.csv).  Easy to work with csv to get it into the database.

I'm not sure the maintainers will see this chat. Might be worth moving it to discord to get their thoughts on it.

tkaraoglu:
> @tkaraoglu I'm in favour of any integration with libre but a little beyond my skills set at the moment.
> 
Definitely. Let's remove the wireshark as middleman. 
> I just discovered that you can download .csv direct from ieee. Not one file but one for each block size i.e. [MA-L](https://standards-oui.ieee.org/oui/oui.csv), [MA-M](https://standards-oui.ieee.org/oui28/mam.csv), and [MA-S](https://standards-oui.ieee.org/oui36/oui36.csv). Easy to work with csv to get it into the database.

I have some spare time to work on this. Will ping them and cc you on discord.
> I'm not sure the maintainers will see this chat. Might be worth moving it to discord to get their thoughts on it.

sliddjur:
I can see this `Vendor` option in the dropdown, but searching is actually not working, it doesnt give me any results when I hit `Search`

image

bonzo81:
This was reworked a couple of times and the search stopped working. It has  now been fixed in PR #15234.

murrant:
Oopsie, added an SQL injection...

PR #15097, opened 2023-06-11
add generic stats graph munging support and make uptime optional between seconds and days

VVelox:
This adds in munging to the generic stats graph.

Also revert uptime back to being days again and make it able to optionally show in graphs as either days or seconds via a new config def.

#### Please note

#### Testers

### DIFF
--- a/includes/html/graphs/device/uptime.inc.php
+++ b/includes/html/graphs/device/uptime.inc.php
@@ -9,8 +9,13 @@ $ds = 'uptime';
 $colours = 'greens';
 $float_precision = 3;
 
-$munge=TRUE;
+$descr = 'Uptime';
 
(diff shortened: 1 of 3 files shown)


murrant (changes requested):
Remove the option and this looks good to me.

VVelox:
> Remove the option and this looks good to me.

Awesome. Done! :)

Any reason we don't want it selectable or just wanting to avoid CPU overhead for something that is not going to be commonly used?

murrant (approved):

murrant:
@VVelox added settings is the death of LibreNMS (maintain ability).

Basically, hidden options end up with unused code paths which is very bad for maintainability.

PR #15128, opened 2023-06-29
Fix MPLS Tab Display for Nokia devices using LDP and not RSVP

peelman:
If the device being monitored used did not have any LSP Paths defined, the MPLS tab was not available.  In our cases, many devices just use LDP and therefore have no LSP paths.  Now use the count of services to determine if the tab is shown, instead of the count of LSPs.

Also fixed some text casing issues (`Service Id` vs `Service ID`).

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/Device/Tabs/RoutingController.php
+++ b/app/Http/Controllers/Device/Tabs/RoutingController.php
@@ -44,7 +44,7 @@ class RoutingController implements DeviceTab
             'bgp' => $device->bgppeers()->count(),
             'vrf' => $device->vrfs()->count(),
             'cef' => $device->cefSwitching()->count(),
-            'mpls' => $device->mplsLsps()->count(),
+            'mpls' => $device->mplsServices()->count(),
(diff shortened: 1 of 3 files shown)


ottorei:
I have not reviewed the actual code in a depth but wouldn't this change hide the MPLS-tab for devices where there are no actual MPLS-services configured other than the static or protocol driven LSP?

peelman:
If there aren’t any services to tie things to, would this populate with useful data?

my goal was to show the service information if it exists (which is more useful than LSP paths that are roads to nowhere)

FadhliRais:
image

as for nokia switch, it work 100% for me,
because for switch, its not using LSP path, it using local VPLS (vpls based on MPLS too), so it have service, but not have lsps

now the tab mpls is shown.

ottorei:
Since this module works correcly with just Nokia equipment (Timos), I guess it's safe to merge this for now. The original MPLS module was made for Timos and using it with other operating systems would probably need some adjustments anyway.

ottorei (approved):
LGTM

FadhliRais (approved):

PR #15134, opened 2023-07-04
change rrd_prune find command to exlcude `.gitignore`

robje:
the daily job rrd_prune deletes all files older than $rrd_purge, including `.gitignore`
This leads to validation warnings as there are changes to files in git.
So with this change any `.gitignore` file is excluded

#### Please note

#### Testers

### DIFF
--- a/daily.php
+++ b/daily.php
@@ -66,7 +66,7 @@ if ($options['f'] === 'rrd_purge') {
         $rrd_dir = Config::get('rrd_dir');
 
         if (is_numeric($rrd_purge) && $rrd_purge > 0) {
-            $cmd = "find $rrd_dir -type f -mtime +$rrd_purge -print -exec rm -f {} +";
+            $cmd = "find $rrd_dir -name .gitignore -prune -o -type f -mtime +$rrd_purge -print -exec rm -f {} +";
             $purge = `$cmd`;
             if (! empty($purge)) {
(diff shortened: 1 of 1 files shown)


daily.php
```diff
@@ -66,7 +66,7 @@
         $rrd_dir = Config::get('rrd_dir');
 
         if (is_numeric($rrd_purge) && $rrd_purge > 0) {
-            $cmd = "find $rrd_dir -type f -mtime +$rrd_purge -print -exec rm -f {} +";
+            $cmd = "find $rrd_dir -name .gitignore -prune -o -type f -mtime +$rrd_purge -print -exec rm -f {} +";
```
laf, line 69: I think you need to add `\!` before `-name` i.e:

`\! -name`

laf (changes requested):

daily.php
robje: that is what `-prune` does, no `\! -name` required

from man page

       -prune True; if the file is a directory, do not descend into it.  If -depth is given, then -prune has no effect.  Because
              -delete implies -depth, you cannot usefully use -prune and -delete together.  For example, to skip  the  directory
              src/emacs  and all files and directories under it, and print the names of the other files found, do something like
              this:
                  find . -path ./src/emacs -prune -o -print

electrocret (approved):

PR #15144, opened 2023-07-13
Skip hostnames which are blank when generating the smokeping config

### LINKED ISSUE #12778: lnms smokeping:generate exits on error instead of skipping misconfigured host breaking smokeping.
I have a multi-user librenms install and someone had to stop while adding more hosts to it and left one misconfigured. (empty hostname in this case)
This broke the smokeping config generator script, instead of skipping a bad host (which i think is the intent when reading the code) it exits with an error because that's how lnms works.
I can't push a commit and I doubt I'll ever have to again so here's a quick fix patch file.
Original error was:
In SmokepingGenerateCommand.php line 348:
checkdnsrr(): Host cannot be empty

```
diff --git a/app/Console/Commands/SmokepingGenerateCommand.php b/app/Console/Commands/SmokepingGenerateCommand.php
index cc0809a03..fd57c9a13 100644
--- a/app/Console/Commands/SmokepingGenerateCommand.php
+++ b/app/Console/Commands/SmokepingGenerateCommand.php
@@ -344,7 +344,11 @@ class SmokepingGenerateCommand extends LnmsCommand
     private function deviceIsResolvable($hostname)
     {
         // First we check for IP literals, then for a dns entry, finally for a hosts entry due to a PHP/libc limitation
-        // We look for the hosts entry last (and separately) as this only works for v4 - v6 host entries won't be found
+       // We look for the hosts entry last (and separately) as this only works for v4 - v6 host entries won't be found
+       if (empty($hostname)) { 
+           //But lets not completely break config configuration over 1 misconfigured host and instead just skip it.
+           return false;
+       }
         if (filter_var($hostname, FILTER_VALIDATE_IP) || checkdnsrr($hostname, 'ANY') || is_array(gethostbynamel($hostname))) {
             return true;
         }
```

laf:

Fixes #12778

#### Please note

#### Testers

### DIFF
--- a/app/Console/Commands/SmokepingGenerateCommand.php
+++ b/app/Console/Commands/SmokepingGenerateCommand.php
@@ -338,6 +338,9 @@ class SmokepingGenerateCommand extends LnmsCommand
      */
     private function deviceIsResolvable($hostname)
     {
+        if (empty($hostname)) {
+            return false;
+        }
         // First we check for IP literals, then for a dns entry, finally for a hosts entry due to a PHP/libc limitation
         // We look for the hosts entry last (and separately) as this only works for v4 - v6 host entries won't be found
         if (filter_var($hostname, FILTER_VALIDATE_IP) || checkdnsrr($hostname, 'ANY') || is_array(gethostbynamel($hostname))) {


PipoCanaja (approved):
LGTM

PR #15147, opened 2023-07-16
API: Change display query of list_devices to pattern match

hirose003:

Changed display query of list_devices from exact match to pattern match.
Because hostname query is a pattern match.

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -352,8 +352,8 @@ function list_devices(Illuminate\Http\Request $request)
         $sql = '`d`.`type`=?';
         $param[] = $query;
     } elseif ($type == 'display') {
-        $sql = '`d`.`display`=?';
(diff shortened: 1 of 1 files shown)


ottorei (approved):
This makes the display selection work with the same logic as hostname selection so looks good to me.

murrant:
:/  I think we need to cache the generated display name in the database.  Code like this only works if people explicitly set display names, and not if they use a template.

ottorei:
> :/ I think we need to cache the generated display name in the database. Code like this only works if people explicitly set display names, and not if they use a template.

Oh, I forgot it was possible not to explicitly set display names..

electrocret:
Merging. This PR has no effect for those who don't manually set display names, and fixes it for those who do manually set. Putting display name in DB is a separate issue.

PR #15156, opened 2023-07-22
Added basic support for distributing billing

laf:
This adds support to allow distributed poller nodes to poll ports from devices assigned to it's poller group.

A new config option has been used so that it doesn't break current distributed setups. Anyone running it by the dispatcher service or running poll-billing.php across multiple pollers will maintain how it works as of today.

With the new config option enabled though, only ports assigned to devices in that pollers group will be polled.

Docs updated.

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Distributed-Poller.md
+++ b/doc/Extensions/Distributed-Poller.md
@@ -186,6 +186,15 @@ It's not necessary to run discovery services on all pollers. In fact, you should
 only run one discovery process per poller group.
 Designate a single poller to run discovery (or a separate server if required).
 
+If you run billing, you can do this in one of two ways:
+
+1. Run poll-billing.php and calculate-billing.php on a single poller which will 
+create billing information for all bills. Please note this poller must have 
+snmp access to all of your devices which have ports within a bill.
+1. The other option is to enable `$config['distributed_billing'] = true;` in 
+config.php. Then run poll-billing.php on a single poller per group. You can run 
+calculate-billing.php on any poller but only one poller overall.
+
 **Dispatcher service**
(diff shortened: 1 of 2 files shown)


electrocret:
The little @murrant in my head is telling me the dbfetch should be translated to Eloquent. My gut is telling me it's a small change, so it doesn't really matter. Beyond that, it LGTM.

murrant:
Did this get tested with the dispatcher service too?

laf:
>

laf:
> Did this get tested with the dispatcher service too? I suspect changes would be needed there too.

Dispatcher support is untouched as this change fails back to the current support. It only adds support for running poll-billing.php using cron across multiple poller groups.

I'm not versed enough to tackle making the dispatcher service billing section distributed aware.

murrant (approved):

murrant:
FYI, I'm thinking of removing the cron based distributed documentation, because it is very confusing to people and I think the service is very stable by now.

electrocret:
@murrant and @laf , I'm fairly certain this PR should also work for dispatch too. I've done some poking in the dispatch code, and it all appears to call poll-billing.php.

https://github.com/librenms/librenms/blob/7e0b41a05a6ff6a53b5d86ec607c595986f48a3b/routes/console.php#L105

https://github.com/librenms/librenms/blob/7e0b41a05a6ff6a53b5d86ec607c595986f48a3b/LibreNMS/queuemanager.py#L366

I was looking into this cause I run dispatch service and also need to leverage distributed ness for my bandwidth billing.

electrocret:
I can confirm. This works for dispatch too. I suggest restarting dispatch once you enable this.

PR #15166, opened 2023-07-26
Fix Oxidized OS Exclude when OS changed in Mapping

electrocret:
Currently if the OS is changed via Variable Mapping, the ignore_os setting is not used. This fixes that.

#### Please note

#### Testers

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1527,8 +1527,8 @@ function list_oxidized(Illuminate\Http\Request $request)
                 }
             }
         }
-        //Exclude groups from being sent to Oxidized
-        if (in_array($output['group'], Config::get('oxidized.ignore_groups'))) {
+        //Exclude groups or os from being sent to Oxidized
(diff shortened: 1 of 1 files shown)


electrocret:
Failed test doesn't appear to be caused by PR.

murrant:
This is probably correct. But is there any concern that we are muddying the definition of this config setting?
It is intended to skip os as LibreNMS names them, this makes them skip os as both LibreNMS and Oxidized names them, which could cause issues in some overlapping scenario.

electrocret:
I get the concern, however I highly doubt there would ever be a point where this is a problem for anyone.

I originally found this issue when dealing with ACI Nexus switches. They're Nexus switches that don't act like normal Nexus switches via cmd line. So I made a mapping for them to the Libre APIC model (ACI controller server). With the plan of eventually making an Oxidized model for it. In the interim I excluded the APIC model from Oxidized.

I came up with a workaround that doesn't require this PR. Mapping ACI Nexus switches to an excluded Oxidized device group until I make the Oxidized Model.

murrant:
I think I am going to close this one. "The OS must match the LibreNMS OS name" is in the config setting description.

PR #15199, opened 2023-08-04
Set mibs explicitly in SnmpQuery

murrant:
Always override system mibs, so we have consistency
New method to control mibs

#### Please note

#### Testers


Jellyfrog:
I think this is a good idea. I remember we did some stupid tests around this. Iirc we had to set some env variable also to really clean it? Like `MIBS`

murrant:
@Jellyfrog Yep, that informed this code :)
I found out if we want to replace some usages of the old methods, we will need to allow mib to be specified.  So took the opportunity to work around the MIBs set oddly in system issue.

PR #15205, opened 2023-08-06
LDAP: remove user_id cast to int

jacobw:
Unnecessary due to field being varchar.

#### Please note

#### Testers


murrant:
I would change the strict comparisons to non-strict.

jacobw:
This patch changes the following block, which has resulted in the foreach working correctly now.

```
$search = ldap_search($connection, trim(Config::get('auth_ldap_suffix'), ','), $filter);
$entries = ldap_get_entries($connection, $search);
foreach ($entries as $entry) {
  $user = $this->ldapToUser($entry);
  if ((int) $user['user_id'] !== (int) $user_id)
    continue;
  }

  return $user;
}
```

I've tested this both using auth_test.php and by logging in to the web site and both work correctly with this patch.

murrant (approved):

PR #14691, opened 2022-12-01
Polling poller debug

electrocret:
Add polling poller_name to alert debug. 

-I'm not sure if I referenced the correct variable to get the poller name.

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [ ] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [ ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.
- [ ] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #15209, opened 2023-08-09
Show which host marked a device as down.

TheMysteriousX:
One of my pollers can't reach a device, it'd be nice to know which one is causing the issue.

Screenshot 2023-08-09 at 03 31 21

#### Please note

#### Testers

### DIFF
--- a/app/Observers/DeviceObserver.php
+++ b/app/Observers/DeviceObserver.php
@@ -42,7 +42,7 @@ class DeviceObserver
         if ($device->isDirty(['status', 'status_reason'])) {
             $type = $device->status ? 'up' : 'down';
             $reason = $device->status ? $device->getOriginal('status_reason') : $device->status_reason;
-            $actor = Config::get('distributed_poller_name', gethostname());
+            $actor = \config('librenms.node_id', gethostname());
             Eventlog::log(sprintf('Device status changed to %s from %s check%s.', ucfirst($type), $reason, Config::get('distributed_poller') ? " by $actor" : ''), $device, $type);
         }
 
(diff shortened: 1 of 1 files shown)


electrocret:
I actually had this exact same issue a while back. My solution was to add the polling poller to the alert data's debug, then added it to my personal transport templates. 

I fear adding it to the alert log may confuse less technical team members in an organization.


https://github.com/librenms/librenms/pull/14691

TheMysteriousX:
I've added a filter to only display the poller name if distributed polling is in use.

I spotted the values in the alert JSON but the information isn't exposed in the UI. If someone spots something going in the UI, it shouldn't generally be necessary for them to go and look in another system or pull bits out of JSON blobs by hand to find information LibreNMS knows.

electrocret:
You bring up good points, however I feel it clutters Libre's interface for a fringe use case. I know I personally wouldn't want it on my distributed instance. I could see it confusing my support team causing them to look in the wrong direction.

A middle ground solution. Add an alertlog.device.statuschange setting to the config (spit balling the name). It'd hold a small Blade that would be used for this alert log entry, so people can customize it to their needs.
Do you have any thoughts @murrant ?



Example of using Blade:
https://github.com/librenms/librenms/blob/master/LibreNMS/Alert/Template.php#L78

murrant:
I guess I'm curious about the proper identifier would be, I would lean toward `\config('librenms.node_id')`
I actually don't set distributed_poller_name on any nodes and hostname is not meaningful (or maybe even unique)
However, node_id is often an auto-generated id and not very readable.

Filtering by the `distributed_poller` setting seems reasonable to me.

TheMysteriousX:
`node_id` is fine too - both are displayed in the UI and are overridable if desired. `node_id` does make a little more sense as it has to be configured in the env file per instance which works better for e.g. docker. 

> hostname is not meaningful (or maybe even unique)
That's just a failback, in case there's a code path outside of the poller (e.g. the API) that could trigger the event.

murrant (approved):

murrant (approved):

PR #15214, opened 2023-08-11
Make the service template controller re-evaluate dynamic membership when apply all button is pushed

eskyuu:
We use dynamic templates, and noticed that the device membership for the service template is not being re-evaluated. On our system we had a number of services that were still being applied even though a device no longer met the requirements for the dynamic template.  We also had quite a few new devices that did not have the template applied.

This pull request causes the dynamic template membership to be re-evaluated whenever the "apply service templates" button is pressed.  It is really only a partial fix because this should happen at some point during discovery.

Thinking about this a bit further, I think that having the template apply rules as well as the group membership is redundant, and possibly confusing.  It appears that the template is applied if a device is either a member of the device group OR matches the apply rules.  This is not obvious from the UI.  It appears that the service template does apply dynamically during discovery if it is linked through a device group.  Is it worth working to remove the apply rules for service templates and having dynamic templates work only through device group membership?

#### Please note

#### Testers

### DIFF
--- a/app/Http/Controllers/ServiceTemplateController.php
+++ b/app/Http/Controllers/ServiceTemplateController.php
@@ -317,6 +317,9 @@ class ServiceTemplateController extends Controller
      */
     public function apply(ServiceTemplate $template)
     {
+        if ($template->type == 'dynamic') {
+            $template->updateDevices();
+        }
         ServiceTemplateController::applyDevices($template);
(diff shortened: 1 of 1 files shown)


murrant (approved):

murrant:
I have never used or understood the service templates.  Feel free to make it work as expected.
This looks correct enough to me, so I'll merge it.

PR #15225, opened 2023-08-16
rework: Cisco SB Switch Temp and Fanspeed fix

Fehler12:
Sorry for the long waiting time. This is an update to #14953, which I cannot update anymore for some reason.

I've made some tests and found that the value 0 should be ignored anyway because some Cisco switches report a non-existing fan/temp-sensor with a speed/temp of 0 which causes several errors when adding e.g. a fanless or desktop switch. Therefore I would like to suggest to keep the skip_value for this case.

I've also seen no more fan flapping with the most recent switch firmware so maybe Cisco has fixed something internally.

#### Please note

#### Testers


Fehler12:
I have currently no idea why the checks fail. I've recreated the test-data for the cbs250 switch with the following comands:
./scripts/collect-snmp-data.php -h 338 -v cbs250-24p-4x
./scripts/save-test-data.php -o ciscosb -v cbs250-24p-4x

However if I upload these test files the checks fail. Any idea beside manually correcting it?

murrant:
@Fehler12 it is because you generated the test data on an old version of LibreNMS.

Fehler12:
Thx. For some reasaon I had to regenerate the test data several times, now it's working. Hoping for pull :).

murrant (changes requested):
don't delete those snmprec files please (especially without explanation)

Fehler12:
> don't delete those snmprec files please (especially without explanation)

Hi, sorry. I've removed those snmprec data because these files were not used for testing. In addition these files did not contain a full SNMP-walk of these devices and therefore I consider these files as obsolete.

murrant:
Those files are used to test OS Detection (but not any polling modules).

https://github.com/librenms/librenms/blob/master/tests/OSDiscoveryTest.php

Please restore them, thanks.

murrant (approved):

PR #15228, opened 2023-08-17
Add global NAC page and refactor the per-device one

PipoCanaja:
Extending existing NAC page to show all NAC entries. 

image

#### Please note

#### Testers


electrocret:
Can you encapsulate the Menu blade entry so that it only displays if there are NAC ports in Libre? 
Similar to Pseudowires.

PipoCanaja:
Of course, makes sense.

PipoCanaja:
Test done on 2 setups, one with NAC entries and one without.

app/Http/Controllers/Table/PortNacController.php
```diff
@@ -72,10 +73,16 @@ protected function sortFields($request)
      */
     public function baseQuery($request)
     {
-        return PortsNac::select('port_id', 'mac_address', 'ip_address', 'vlan', 'domain', 'host_mode', 'username', 'authz_by', 'timeout', 'time_elapsed', 'time_left', 'authc_status', 'authz_status', 'method')
-            ->where('device_id', $request->device_id)
-            ->hasAccess($request->user())
-            ->with('port');
+        if (is_numeric($request->device_id)) {
+            return PortsNac::select('device_id', 'port_id', 'mac_address', 'ip_address', 'vlan', 'domain', 'host_mode', 'username', 'authz_by', 'timeout', 'time_elapsed', 'time_left', 'authc_status', 'authz_status', 'method')
```
murrant, line 77: Convert this to one query to prevent divergence between the queries using ->when()

app/Http/Controllers/Table/PortNacController.php
```diff
@@ -88,6 +95,9 @@ public function formatItem($nac)
         $item['mac_oui'] = Rewrite::readableOUI($item['mac_address']);
         $item['mac_address'] = Rewrite::readableMac($item['mac_address']);
         $item['port'] = null; //free some unused data to be sent to the browser
+        if (isset($item['device_id'])) {
```
murrant, line 98: This is always true.

murrant:
Looks good overall, but I don't want to add any more legacy pages that need to be converted to Laravel.  Can you please convert it?

PipoCanaja:
> Looks good overall, but I don't want to add any more legacy pages that need to be converted to Laravel. Can you please convert it?

I expected that :) hoped to escape it, but I agree with you indeed. Will do. I am definitely not familiar enough with Laravel logic to do it easily, but it will eventually improve :)

app/Http/Controllers/Table/PortNacController.php
PipoCanaja: seems that you are right. don't know how I ended up convinced I needed to add it ...

PipoCanaja: done

app/Http/Controllers/Table/PortNacController.php
PipoCanaja: done

PipoCanaja:
> Looks good overall, but I don't want to add any more legacy pages that need to be converted to Laravel. Can you please convert it?

Could you help me start ? Like, at first, what is needed here ? Only a controller and a blade file, with a web route pointing to the controller ?

PipoCanaja:
At least I have a nac page displayed with Controller and View ... StyleCi is already complaining but it works !

app/Http/Controllers/Table/PortNacController.php
```diff
@@ -72,10 +73,14 @@ protected function sortFields($request)
...
-            ->hasAccess($request->user())
-            ->with('port');
+        $query = PortsNac::select('device_id', 'port_id', 'mac_address', 'ip_address', 'vlan', 'domain', 'host_mode', 'username', 'authz_by', 'timeout', 'time_elapsed', 'time_left', 'authc_status', 'authz_status', 'method')
+                ->hasAccess($request->user())
+                ->with('port');
+        $query->when(is_numeric($request->device_id), function ($q) use ($request) {
+            return $q->where('device_id', $request->device_id);
+        });
+
+        return $query;
```
murrant, line 83:
```suggestion
        return PortsNac::select('device_id', 'port_id', 'mac_address', 'ip_address', 'vlan', 'domain', 'host_mode', 'username', 'authz_by', 'timeout', 'time_elapsed', 'time_left', 'authc_status', 'authz_status', 'method')
            ->when($request->device_id, fn($q, $id) => $q->where('device_id', $id))
            ->hasAccess($request->user())
            ->with('port');
```

Just a little cleaner

resources/views/nac.blade.php
```diff
@@ -0,0 +1,132 @@
...
+    </style>
+@endsection
+
+@section('javascript')
+    <script>
+        $(document).ready(function () {
+            locations_grid = $("#nac-grid").bootgrid({
+                ajax: true,
+                rowCount: [25, 50, 100, -1],
+                url: "{{ url('/ajax/table/port-nac') }}",
```
murrant, line 54: Try to use the route() helper over url() when you can.
Add `->name('table.nac')` to the web.php route file and change this to `route('table.nac')`

routes/web.php
```diff
@@ -66,6 +66,8 @@
     Route::get('/', 'OverviewController@index')->name('home');
     Route::view('vminfo', 'vminfo');
 
+    Route::any('nac', 'NacController@index');
```
murrant, line 69:
```suggestion
    Route::get('nac', 'NacController@index');
```

Only need the get verb I think.

murrant (changes requested):
Looks pretty good overall, just some small nitpicks.

app/Http/Controllers/Table/PortNacController.php
```diff
@@ -88,6 +93,7 @@ public function formatItem($nac)
         $item['mac_oui'] = Rewrite::readableOUI($item['mac_address']);
         $item['mac_address'] = Rewrite::readableMac($item['mac_address']);
         $item['port'] = null; //free some unused data to be sent to the browser
+        $item['device_id'] = Url::deviceLink($nac->device);
```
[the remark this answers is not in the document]
PipoCanaja: Should be fixed by now.

resources/views/nac.blade.php
```diff
@@ -0,0 +1,132 @@
...
+    </div>
+</div>
+@endsection
+
+@section('css')
+    <style>
+    </style>
+@endsection
+
+@section('javascript')
```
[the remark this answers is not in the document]
PipoCanaja: Don't get that one at all :(

app/Http/Controllers/Table/PortNacController.php
PipoCanaja: on its way

resources/views/nac.blade.php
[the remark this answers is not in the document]
murrant: Well, the script includes are kind of a mess in the librenmsv1 template.
@section('javascript') is for loading scripts before the pages loads (in the header)
@section('scripts') is an old one then I realized we could use stack instead to include multiple so:
@push('scripts') is the preferred per-page script location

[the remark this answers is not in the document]
PipoCanaja: should be ok now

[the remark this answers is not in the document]
PipoCanaja: should be ok now

resources/views/nac.blade.php
PipoCanaja: should be ok as well

PipoCanaja:
Seems that it is more or less OK. Would make sense to refactor the "per-device" page now ... Should be easy.

murrant (approved):

PR #15235, opened 2023-08-20
Added empty string check

Fehler12:
I've added a check to make sure that in the device list in every case a name is shown:

before:
image
after:
image

I think the underlaying reason is that the printer reports an empty name back via SNMP. In this case the hostname is shown (as this field should never be empty).
With this improvement the entry is clickable by its IP again. 

#### Please note

#### Testers

### DIFF
--- a/app/Models/Device.php
+++ b/app/Models/Device.php
@@ -188,10 +188,10 @@ class Device extends BaseModel
         ]);
 
         // in case the displayName is a string without any content
-        if($return_value == '' or is_null($return_value)) {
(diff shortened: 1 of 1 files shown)


PipoCanaja:
Nice catch. This was on my todolist indeed ! Thanx !

PipoCanaja:
LGTM, tested

PipoCanaja (approved):

murrant:
@Fehler12 hmm, it already checks for an empty string, so you are checking it twice.  Perhaps just needed to add trim() to the previous check.

PR #15236, opened 2023-08-20
Long and short vendor name stored in OUIs

PipoCanaja:
- Store both short and long vendor name.
- default to long one
- short one is accessible with column "vendor_short" in TABLE vendor_ouis.
- trimed the variables before storage to avoid the whitespaces in vendor_short.

#### Please note

#### Testers

### DIFF
--- a/app/Console/Commands/MaintenanceFetchOuis.php
+++ b/app/Console/Commands/MaintenanceFetchOuis.php
@@ -104,7 +104,7 @@ class MaintenanceFetchOuis extends LnmsCommand
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


PipoCanaja:
@murrant : PR done to load both values. And it defaults to the long one (as it used to be in the initial implementation of OUIs. I did not create any mecanism to force upgrade. So the current rules will delay the effect of this PR.

murrant:
:/  I just reviewed the "short" names in the database.  They seem to just be the long names chopped off.  I don't see a point in storing those if that is the case.   We can just "shorten" them for table display.

PR #7064, opened 2017-07-21
feature: Add CiscoSB Port Suspended Status Info

Rosiak:
DO NOT DELETE THIS TEXT

#### Please note

> Please read this information carefully. You can run `./scripts/pre-commit.php` to check your code before submitting.

- [X] Have you followed our [code guidelines?](http://docs.librenms.org/Developing/Code-Guidelines/)

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`

PR #15248, opened 2023-08-24
Add discovery for ports errdisabled on Cisco NXOS

fbouynot:
It is not in the yaml file so it can access the interface name 

image

#### Please note

#### Testers


PipoCanaja:
Hi @fbouynot 
If I get it right, err-disabled ports are discovered during the discovery process. What happens when the issue (Err-disabled) is solved ? how does it look like in the WebUI ? And what happens at the following discovery ? 

It is not really the expected way of handling the err-disabled status.

fbouynot:
Hi @PipoCanaja 

> If I get it right, err-disabled ports are discovered during the discovery process.
You get it right.

I came with this because it was (I think) how it was done in https://github.com/librenms/librenms/pull/7064

> What happens when the issue (Err-disabled) is solved ? how does it look like in the WebUI ? And what happens at the following discovery ?

If it is only checked at discovery, either the OID exists, or it doesn't (which is why it is tricky to check).
So if the problem is solved, the sensors disappeared.

> It is not really the expected way of handling the err-disabled status.

Yes I do agree. I was trying to setup something based on existing code, even if suboptimal (can at least catch long time suspended ports).

Although now that I have something working and you ask me how it works, I can understand a little better how I could do.
Would it be OK if I : 
- discover sensor for each existing port instead, and apply a value discovered by snmpget on suspend oid OR 0 if oid doesn't exists
- add a script for polling doing the same except the discovery part

fbouynot:
I fixed the logic: sensors are persistent and can be polled.
image
image
I just need to regenerate test data, it should take two hours, so maybe I'll add it today, or maybe on monday.

murrant:
NACK on this.  It should either be a trap, or an extended port status on the port itself, not a sensor.

Issue #15240, opened 2023-08-22
 app/Models/Device.php


ssninnni:
### The problem

After updating to latest LibreNMS version API Request 
GET /api/v0/bills/19/graphs/bits?from=1690541332&to=1693519199

produced following error (full trace in Details)

`App\\Models\\Device::displayName(): Return value must be of type string, null returned`

Looks like the problem was caused by this one:

[https://github.com/librenms/librenms/commit/cbe2ba55cefe58b5372452bca32b8ecfc9b3fe7e](https://github.com/librenms/librenms/commit/cbe2ba55cefe58b5372452bca32b8ecfc9b3fe7e)

### Output of ./validate.php

```txt
===========================================
Component | Version
--------- | -------
LibreNMS  | 23.8.2 (2023-08-21T17:48:13+02:00)
DB Schema | 2023_08_02_120455_vendor_ouis_unique_index (255)
PHP       | 8.1.22
Python    | 3.6.8
Database  | MariaDB 10.11.5-MariaDB
RRDTool   | 1.7.0
SNMP      | 5.8
===========================================

[OK]    Composer Version: 2.5.8
[OK]    Dependencies up-to-date.
[OK]    Database connection successful
[OK]    Database Schema is current
[OK]    SQL Server meets minimum requirements
[OK]    lower_case_table_names is enabled
[OK]    MySQL engine is optimal
[OK]    Database and column collations are correct
[OK]    Database schema correct
[OK]    MySQl and PHP time match
[OK]    Active pollers found
[OK]    Dispatcher Service not detected
[OK]    Locks are functional
[OK]    Python poller wrapper is polling
[OK]    Redis is functional
[OK]    rrdtool version ok
[OK]    Connected to rrdcached
```

### Anything in the logs that might be useful for us?

```txt
{
   "message": "App\\Models\\Device::displayName(): Return value must be of type string, null returned",
   "exception": "TypeError",
   "file": "/opt/librenms/app/Models/Device.php",
   "line": 191,
   "trace": [
       {
           "file": "/opt/librenms/LibreNMS/Data/Graphing/GraphParameters.php",
           "line": 253,
           "function": "displayName",
           "class": "App\\Models\\Device",
           "type": "->"
       },
       {
           "file": "/opt/librenms/LibreNMS/Data/Graphing/GraphParameters.php",
           "line": 211,
           "function": "defaultTitle",
           "class": "LibreNMS\\Data\\Graphing\\GraphParameters",
           "type": "->"
       },
[...]

PR #15249, opened 2023-08-24
Revert "Added empty string check"

murrant:
Reverts librenms/librenms#15235

fixes: #15240

### DIFF
--- a/app/Models/Device.php
+++ b/app/Models/Device.php
@@ -180,15 +180,12 @@ class Device extends BaseModel
     {
         $hostname_is_ip = IP::isValid($this->hostname);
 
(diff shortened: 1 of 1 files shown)


MezentsevSPB:
Hello. This fix breaks the rules logic. The fact is that zero and null are neither the same thing. If nothing was specified in the parameter set, the value of null was written to the database. Now a zero is written there. This breaks the rules and imposes additional load on SQL queries. After editing and saving the parameters, null is overwritten with zero, which causes false positives of the rules.

murrant:
This has nothing to do with writing to the database.  If you have an issue, please open a ticket with a detailed explanation of the problem.

PR #15253, opened 2023-08-25
Update Install-LibreNMS.md

jult:
Script fails for users having installed python and pip using apt, which is also default for debian 12.

#### Please note

#### Testers

### DIFF
--- a/doc/Installation/Install-LibreNMS.md
+++ b/doc/Installation/Install-LibreNMS.md
@@ -101,6 +101,15 @@ su - librenms
 ./scripts/composer_wrapper.php install --no-dev
 exit
 ```
+
+Important: For people who have installed Python using apt, the script can fail trying to use pip installing python requirements.
+This is default behavior for debian version 11 testing (and newer). In order to work around this, you can safely run this:
+```
+su - librenms
+pip3 install --user --break-system-packages -r /opt/librenms/requirements.txt
+```
+which will finalize the install script for those users.
+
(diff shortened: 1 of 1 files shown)


jult:
Sorry, I accidentally deleted the systemctl restart php8.2-fpm line, which was not intended to be pushed ;)

murrant:
@jult push a new commit to your branch and it will update this.

murrant:
Maybe we just update the pip3 call in composer.json instead?

jult:
> Maybe we just update the pip3 call in composer.json instead?

Yes, but then users will wonder why you'd be 'breaking' system-packages for librenms.
A better option is probably to get rid of python altogether, I think..

murrant (changes requested):
I think this is the wrong approach.

ottorei:
> I think this is the wrong approach.

You mean the original or the venv? Running the Python inside venv would only need minimal changes to the update script and systemd unit. Running it in a venv could be a universal solution to avoid breaking any Python packages on all platforms.

jult:
A very similar discussion was going on for LetsEncrypt's certbot on their forums. They decided to drop pip/python and switch to use snap (snapcraft, snapd) altogether. And frankly, ever since, so many (mostly user) issues with certbot have disappeared..

murrant:
@jult yeah.  I still wish there was a clean fix for this.  I tried to convert the existing directory into a venv, but I don't know enough about python...

murrant:
Closing, this needs to be fixed another way.  Thanks!

PR #15265, opened 2023-08-29
add apps.no_app_id config item and add support for it to Logsize

VVelox:
Created this due to network issues with remote sensors occasionally leads to apps being removed and re-added, resulting in loss of old monitoring data.

This adds a optional work around for that via adding in a new config item to control if the app ID is used in the RRD file name.

Initial support has been added to Logsize.

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Applications.md
+++ b/doc/Extensions/Applications.md
@@ -155,6 +155,25 @@ LibreNMS during discovery and polling.
 
 ![Enable-applications](/img/Enable_applications.png)
 
+## Disabling using app ID with RRDs
+
+When a application is discovered it is assigned a ID. Then if
+discovery runs again and it has issues polling that app, that app
+will be removed from database table and it will loose association
+with that app ID, meaning next redicovery in which it is found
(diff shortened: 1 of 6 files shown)


murrant:
What happens when you have multiple of the same app per device, or is that not possible?

VVelox:
> What happens when you have multiple of the same app per device, or is that not possible?

It does not create multiple apps, but when a app is removed and re-added they end up with a different app ID and when that app ID is used for creating the RRDs this results in the old not being used. So lets say the old ID is 94 and the new one is 103, the result one be `app-logsize-94-suricata_flows.rrd` being used initially and then `app-logsize-103-suricata_flows.rrd`.

Enabling this causes `app-logsize-suricata_flows.rrd` to be used for it instead, meaning it even if the app ID changes the same RRD files are used.

VVelox:
The same thing happens if you disable a app for a machine and then re-enable it as well, regardless of discovery.

VVelox:
@murrant Closing this to look at updating this to do via soft-delete.

PR #15270, opened 2023-08-30
Application Soft Delete

VVelox:
Start of work of adding in more bits for Laravel DB stuff for Application handling. Apps when disabled now have `deleted_at` set and unset when enabled, perserving the app ID in case it is toggled or rediscovered with it gone and then reversed.

#### Please note

#### Testers


Jellyfrog:
What do we win with soft delete? (Except for id staying the same)

VVelox:
@Jellyfrog This is exactly the point for doing this.

The changing ID results in loosing data when it comes to apps.

VVelox:
This is also something you will notice a bit of an issue with when using snmpd heavily in proxy mode, which can best be described as something that is a bit buggy, but there is nothing else really out there. Basically a cron job to give it a regular kick as there is a damned good chance if left to run for to long it will start deciding it does not want to talk to talk to various hosts, especially if they are on a terrible network segment. Will occasionally decide it does not want to talk to a host, but with regular restarts it does is a whole hell of a lot less.

An app getting flipped off and then back on because of this is less like, but I've seen that happen as well.

includes/discovery/applications.inc.php
```diff
@@ -77,14 +78,26 @@
...
         } else {
-            dbInsert([
-                'device_id' => $device['device_id'],
-                'app_type' => $app,
-                'discovered' => 1,
-                'app_status' => '',
-                'app_instance' => '',
-            ], 'applications');
-
+            Application::upsert(
```
murrant, line 81: i still don't want us to use upsert for a single row.

VVelox: @murrant For some reason that works fine when used but breaks CI. That is why I revered it in [03c34ec](https://github.com/librenms/librenms/pull/15270/commits/03c34ec10d776c06365481d47f67a06f9c57b7b7).

VVelox: Actually I take that back. Was not setting discovered.

app/Observers/DeviceObserver.php
```diff
@@ -164,6 +164,7 @@ public function deleting(Device $device): void
         $device->vrfLites()->delete();
         $device->vServers()->delete();
         $device->wirelessSensors()->delete();
+        \DB::table('applications')->where('device_id', $device->device_id)->delete();
```
murrant, line 167: There is a line above that already deletes applications, just change that to:
`$device->applications()->forceDelete();`

includes/html/forms/application-update.inc.php
```diff
@@ -32,21 +34,24 @@
...
                 $status = ['status' => 0, 'message' => 'Application enabled'];
+            } else {
+                $status = ['status' => 1, 'message' => 'Database update for enabling the application failed'];
             }
         } else {
-            if (dbDelete('applications', '`device_id`=? AND `app_type`=?', [$device_id, $app])) {
-                log_event("Application disabled by user: $app", $device_id, 'application', 3);
+            $app->delete();
+            if ($app->save()) {
+                log_event("Application disablaed by user: $app", $device_id, 'application', 3);
```
murrant, line 51: Spelling here.  While we are making changes in this code, might as well change from log_event() to 
`Eventlog::log()`

includes/discovery/applications.inc.php
```diff
@@ -98,12 +98,10 @@
     echo str_repeat('-', $num);
     $vars = $apps_to_remove;
     array_unshift($vars, $device['device_id']);
-    dbDelete(
-        'applications',
-        '`device_id`=? AND `app_type` IN ' . dbGenPlaceholders($num),
-        $vars
-    );
     foreach ($apps_to_remove as $app) {
+        $app_obj = Application::withTrashed()->firstOrNew(['device_id' => $device['device_id'], 'app_type' => $app]);
```
murrant, line 102: You don't want withTrashed() in this query.

murrant: or firstOrNew.

Something like this:
```php
$appToDelete = Application::firstWhere(['device_id' => $device['device_id'], 'app_type' => $app]);
if ($appToDelete) {
   $appToDelete->delete();
}
```

or
```php
DeviceCache::getPrimary()->applicatins()->whereIn('app_type', $apps_to_remove)->get()->each->delete();
```

We could call `DeviceCache::getPrimary()->applicatins()->whereIn('app_type', $apps_to_remove)->delete();`, but that wouldn't fire events for the device observer.

includes/discovery/applications.inc.php
```diff
@@ -67,25 +69,23 @@
...
-                'discovered' => 1,
-                'app_status' => '',
-                'app_instance' => '',
-            ], 'applications');
-
-            echo '+';
+        if (! in_array($app, $enabled_apps)) {
+            $app_obj = Application::withTrashed()->firstOrNew(['device_id' => $device['device_id'], 'app_type' => $app]);
+            if ($app_obj->trashed()) {
+                $app_obj->restore();
```
murrant, line 85: ModuleModelObserver does not currently handle the restored event, so this won't output anything.

includes/discovery/applications.inc.php
VVelox: Yeah, went and changed this to...

```
        Application::where([
            ['device_id', $device['device_id']],
            ['app_type', $app],
        ])->delete()->save();
```

includes/discovery/applications.inc.php
VVelox: Ahh! Okay. You mean it does not hit on the restored bit.

 Added restored to ModuleModelObserver.

includes/discovery/applications.inc.php
```diff
@@ -98,12 +98,11 @@
...
-    dbDelete(
-        'applications',
-        '`device_id`=? AND `app_type` IN ' . dbGenPlaceholders($num),
-        $vars
-    );
     foreach ($apps_to_remove as $app) {
+        Application::where([
+            ['device_id', $device['device_id']],
+            ['app_type', $app],
+        ])->delete()->save();
```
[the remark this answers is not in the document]
murrant: This does not trigger the deleted event FYI.

murrant (approved):

PR #15286, opened 2023-09-06
TP-link EAP support

st4ro:
Hi, 
please check following commits - it extends TP-link EAP devices (wifi connected clients). 

Note there is BUG in FW of EAP 610 - they always respond with INT = 0 - see [ https://community.tp-link.com/en/business/forum/topic/618782]( https://community.tp-link.com/en/business/forum/topic/618782)

but EAP610-Outdoor(EU) v1.0, EAP653(EU) v1.0 are working 

#### Please note

#### Testers


st4ro:
> Why is this a new OS?
Hi,
EAP devices are WiFi AP - actual TP-link devices are CPE-R/switch. There are no information about CPU/MEM at EAP devices..

includes/definitions/tplink-eap.yaml
```diff
@@ -0,0 +1,12 @@
...
+type: wireless
+icon: tplink
+over:
+    - { graph: device_bits, text: 'Device Traffic'}
+    - { graph: device_wireless_clients, text: 'Number of Clients' }      
+poller_modules:
+    wifi: true
+discovery:
+    - sysObjectID:
+            - .1.3.6.1.4.1.11863.3.2.10
```
murrant, line 12: probably should just be .1.3.6.1.4.1.11863.3.

st4ro: Yes, definitely. My mistake

LibreNMS/OS/TplinkEap.php
```diff
@@ -0,0 +1,47 @@
...
+use LibreNMS\OS;
+
+class TplinkEap extends OS implements OSDiscovery,WirelessClientsDiscovery
+{
+   /**
+    * 
+    */ 
+    public function discoverOS(Device $device): void
+    {
+        $device->hardware = snmp_get($this->getDeviceArray(), '.1.3.6.1.2.1.1.5.0', '-Osqnv');
```
murrant, line 28: Use discovery yaml instead:

```yaml
modules:
    os:
        sysDescr_regex: '/(?<version>.*)/'
        hardware: SNMPv2-MIB::sysName.0
```

murrant:
Since you are defining sensors, you should probably include test data for them. (check link from bot)

LibreNMS/OS/TplinkEap.php
st4ro: Well, this is not clear to me - from existing YAML it looks like that I cannot combine existing yaml with your suggestinon. 

this is not working for me: 

```
os: tplink-eap
modules:
    os:
        hardware: SNMPv2-MIB::sysName.0
        sysDescr_regex: '/(?<version>.*)/'
type: wireless
icon: tplink
over:
    - { graph: device_bits, text: 'Device Traffic'}
    - { graph: device_wireless_clients, text: 'Number of Clients' }      
poller_modules:
    wifi: true
discovery:
    - sysObjectID:
             .1.3.6.1.4.1.11863.3.
```

am i missing something?

PipoCanaja:
Hi @st4ro 
Could you please generate the JSON file as well and add it to the PR.
Thanx

st4ro:
Hi @PipoCanaja  
I would like to but don't know which JSON file do you mean / how to create one. 
To be honest I already lost in documentation at [https://docs.librenms.org/Developing/os/Test-Units/](https://docs.librenms.org/Developing/os/Test-Units/)

Can you please provide way how to create this file?

PipoCanaja:
Hi
The snmprec file is the "collected" test data (the source). And the json file is the "computed" test data (the result after LibreNMS processed it. And the goal of the test data is to help for non regression tests: the same source data should provide the same computed data over time, and if it does change without notice, then we probably have a regression. 

So you collected correctly the source data (snmprec file). And you now need to run the "save-tests-data" script (chapter 2 of the doc you linked.)

st4ro:
Thank you. The PR now contains a JSON file and the correct SNMPREC file name. I also edited ifPhysAddress in JSON file manualy...

includes/definitions/tplink.yaml
```diff
@@ -7,7 +7,8 @@ over:
 discovery:
     -
         sysObjectID:
-            - .1.3.6.1.4.1.11863.
```
PipoCanaja, line 10: Any reason to be more restrictive than before ? Are you sure no device are now unselected for tplink OS

st4ro: Hi,
OID of EAP is 
             .1.3.6.1.4.1.11863.3

OID for SW and CPE-R is 
            - .1.3.6.1.4.1.11863.1.
            - .1.3.6.1.4.1.11863.5.

the effort was to separate these groups due to the different display (EAP has a sensor for the number of WiFi clients, but others such as memory or CPU are not there)

PipoCanaja: But right now, you restrict the discovery to 1.3.6.1.4.1.11863.1. and 1.3.6.1.4.1.11863.5. So 1.3.6.1.4.1.11863.3 devices are now "Generic Device". Probably not what we want to do. 
You can discover sensors for WIFI clients if present, and skip it if not. Same fo memory and cpu. No need to filter at the OS level (which can and should remain unique unless they are really different OS)

PipoCanaja (changes requested):

includes/definitions/tplink.yaml
st4ro: > You can discover sensors for WIFI clients if present, and skip it if not. Same fo memory and cpu. No need to filter at the OS level (which can and should remain unique unless they are really different OS)

that is what I was looking for, but does not found any of this in examples or ready devices. This is also reason why I create fresh OS.

Is there any example in existing code? This PR can be closed without merge then...

PipoCanaja: I don't know your devices so I can tell for sure which solution is best. I can only look at your PR and try help you find the best solution. 

For instance, the code you added in `LibreNMS/OS/TplinkEap.php` could be simply in the standard Tplink OS PHP file with a `if` statement matching the device that has wireless code. So this would cleanly handle both situations. 

For the processor and memory, the YAML may be able to discover both standard Tplink and TplinkEap and should only keep the one that has an SNMP reply, so once again it should behave correctly.

murrant:
This does not require a new OS, Please add detection of the new sensor to the existing OS.  Thank you.

PR #14207, opened 2022-08-15
Migrate xDSL code to module, and add support for VDSL2 MIB

PipoCanaja:
This PR adds an minimal support for VDSL2 MIB.
- New module for xDSL processing
- GUI changes from ADSL to xDSL
- Old ADSL code is integrated into the module and cleaned up a little bit
- New table is added for VDSL in DB cause ports_adsl table cannot be easily changed to something "generic"
- A few eloquent DB updates
- Config variable `lnms config:set enable_ports_adsl true` is now unsupported and must now be replaced by both:
    - `lnms config:set discovery_modules.xdsl true` 
    - `lnms config:set poller_modules.xdsl true`

DO NOT DELETE THE UNDERLYING TEXT

#### Please note

> Please read this information carefully. You can run `./lnms dev:check` to check your code before submitting.

- [X] Have you followed our [code guidelines?](https://docs.librenms.org/Developing/Code-Guidelines/)
- [ ] If my Pull Request does some changes/fixes/enhancements in the WebUI, I have inserted a screenshot of it.
- [X] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [test data](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers

If you would like to test this pull request then please run: `./scripts/github-apply <pr_id>`, i.e `./scripts/github-apply 5926`
After you are done testing, you can remove the changes with `./scripts/github-remove`.  If there are schema changes, you can ask on discord how to revert.

PR #15300, opened 2023-09-08
Feature port security

mradams33:
This is to add support for polling the maximum allowed MAC addresses per port and whether a port has sticky MACs enabled. This is only for ios/iosxe devices

#### Please note

- [x] If my Pull Request makes discovery/polling/yaml changes, I have added/updated [ios_ios-xe.snmprec, ios_ios.snmprec](https://docs.librenms.org/Developing/os/Test-Units/).

#### Testers


mradams33:
Keep getting this error on two of the DB tests. I haven't modified these files. Is there a way to get around this?

1) LibreNMS\Tests\OSDiscoveryTest::testOSDetection with data set "3com" ('3com')
OS 3com took longer than 10s to detect
Failed asserting that 180.41335892677307 is less than 10.

/home/runner/work/librenms/librenms/tests/OSDiscoveryTest.php:133
/home/runner/work/librenms/librenms/tests/OSDiscoveryTest.php:94
/home/runner/work/librenms/librenms/vendor/laravel/framework/src/Illuminate/Foundation/Testing/TestCase.php:174

murrant:
Some change you made slowed down polling a lot.  Check through things to see what it is.

mradams33:
I checked through my changes and made some adjustments. I enclosed my update section in polling/ports.inc in an if clause for only ios/ios-xe devices so not every device tries to execute it. I also changed my DB fields to nullable since the OIDS I’m polling are Cisco only. None of that changed anything. It’s still stuck on 3com. I have since commented out my changes to discovery/ports.inc and all but the array instantiation for the new OIDs in polling/ports.inc. Still hanging on 3com.

Michael Adams [A picture containing text, font, graphics, logo  Description automatically generated]
Network Engineer
Tech Solutions | Illinois State University
Phone: 309-438-1982
[Text  Description automatically generated]

From: Tony Murray ***@***.***>
Date: Monday, September 11, 2023 at 9:27 PM
To: librenms/librenms ***@***.***>
Cc: Adams, Michael ***@***.***>, Author ***@***.***>
Subject: Re: [librenms/librenms] Feature port security (PR #15300)
This message originated from outside of the Illinois State University email system. Learn why this is important<https://help.illinoisstate.edu/technology/page-88179827.html>

Some change you made slowed down polling a lot. Check through things to see what it is.

mradams33:
Removing the extra lines at the top and bottom of my snmprec files seems to have gotten the tests to go further than usual, but are still failing on 3com each time. Since I'm making a change to the DB, do I need to create a test for each OS?

PipoCanaja:
> Removing the extra lines at the top and bottom of my snmprec files seems to have gotten the tests to go further than usual, but are still failing on 3com each time. Since I'm making a change to the DB, do I need to create a test for each OS?

Yes indeed. If you change the DB, that means that all devices using that table (ports in that case) are impacted, which means 90% of them. 

Processing all snmprec->json is a long process. So may be before going further , it would be best to be sure that the overall design of your implementation is reviewed by maintainers.

mradams33:
I'm a little confused when walking through this doc. https://docs.librenms.org/Developing/os/Test-Units/

First off, is this what I need to be following? If so, I don't see how steps 1 & 2 allow me to generate test data for all OS's. It seems like they'd only work for the OS's I have on my network.

I know you said wait until a maintainer checks my code out, but I just wanted to try to get started with the tests.

murrant:
`./scripts/save-test-data.php -m ports` will start an snmpsim server with the saved snmprec data and go through all os that have port test data and re-generate the port data.

That said, I'm leery of cramming more stuff into the ports module.

mradams33:
Is there a better place for it? I understand if it might make polling times longer. Should I just keep this addition to my environment? Only issue with that is daily.sh. I could create some scripts that continually add everything back in so we can continue to get updates. Slight annoyance, but not the end of the world. Historical data for these OIDs isn't important.

Michael Adams



Michael Adams, DEVASC

Network Engineer

Tech Solutions | Illinois State University

Phone: 309-438-1982


mradams33:
> Is there a better place for it? I understand if it might make polling times longer. Should I just keep this addition to my environment? Only issue with that is daily.sh. I could create some scripts that continually add everything back in so we can continue to get updates. Slight annoyance, but not the end of the world. Historical data for these OIDs isn't important.

Just following up on if there is a better module to put these OIDs in, or if there is a better route for adding them before moving forward with testing.

murrant:
Sorry, I don't understand enough about this to give you that information.  It might help to research other vendors to see how they support similar data in SNMP.

I also noticed this data is never shown anywhere.  If we are collecting it, it is good to show in at least one place in the web ui.

murrant:
It would appropriate to make this a module so users can enable/disable too. For example: 
https://github.com/librenms/librenms/blob/master/LibreNMS/Modules/Vminfo.php

mradams33:
Can't disagree with that. If it's a separate module, should that also be saved to a separate table?

murrant:
Yeah probably.

PipoCanaja:
> I have new DB table created and a script written to poll for the OIDs in addition to using the Vminfo as a template for this module. Everything seems to be working when running poller.php -m on demand, but I can't figure out how to add the module to regular polling and be selectable on the global poller page. How do I do that?

You may have a look at this PR : 
https://github.com/librenms/librenms/pull/14207

mradams33:
Looks like everything is good except I'm using deprecated functions like dbInsert and dbFetchRow. After doing some digging, it looks like the new format is to define the poll function in the model then call it in the .inc.php file. Is that correct? That leads me to a couple questions:

1. Do I have to do it that way, or can I keep everything in .inc.php?
2. Do I have to pass this check, or is it more of a warning?

I'll admit I don't know PHP let alone have experience with Eloquent. Everything I've done has been based off of other scripts in Libre.  With that said, I'm not really sure how to get this done.

mradams33:
I did figure out how to do use Eloquent to do the DB functions. Now I'm failing on my Modules/PortSecurity.php class.

mradams33:
Realized current Module.php interface structure is different from my local Libre instance. I'll have to do some updates.

mradams33:
Passed all checks. Ready for review.

PipoCanaja:
Hi @mradams33 

As Port-Security is supported by multiple vendors, the name "cisco" should not be included in any Module file. En the Module itself should not contain "Cisco" in its name or Label. 
Only the specific Cisco impementation (in Cisco OS files or so) should contain Cisco.

That way, it will be possible to extend this module to other vendors.

mradams33:
> Hi @mradams33
> 
> As Port-Security is supported by multiple vendors, the name "cisco" should not be included in any Module file. En the Module itself should not contain "Cisco" in its name or Label. Only the specific Cisco impementation (in Cisco OS files or so) should contain Cisco.
> 
> That way, it will be possible to extend this module to other vendors.

Ok... The specific mib I used was made by Cisco so that's why I called it Cisco, similar to other modules. I was just basing my names off those. This module as is does not poll for port security oids for other vendors. I thought it would be misleading to not put Cisco in the name.
If that's the sticking point, I can remove Cisco from the names.

mradams33:
Removed Cisco references.

PipoCanaja:
> Removed Cisco references.

Perfect. The idea is to avoid locking a module to a vendor. LibreNMS code being class based, it makes more sense to have 1 port security module, and implement it for as many vendors as possible, each one with its own MIB and code of course, but from a user point of view, 1 module only.

mradams33:
I noticed polling time issues with this module. It was an issue with the amount of times it would run select queries to verify if the ports were in the ports table. I moved that part before the for loop and stored all of the data in an array. Cut down polling time for this module by ~13 seconds for the 4500 I was using to test against.

mradams33:
After optimizing the polling process, I was failing on the Fortigate test. Noticed there was a commit to mitigate that. Merged with the latest commit, and updated the default state for the port-security module (off). Testing should pass now, and be ready for review.

PipoCanaja:
Not an expert by far in Databases. But should'nt we have an id for port_security table, and port_id would be a foreign key, and device_id would be removed as it can be derived from port_id (which we'll always have around ) ? No change in the DB size but would be cleaner in my old database lessons memory ... @librenms/reviewers ?

mradams33:
Not a db expert either and I did consider that. I probably thought port_id was already unique in this table. I (selfishly) included device_id to make my queries easier in the script.


Michael Adams, DEVASC

Network Engineer

Tech Solutions | Illinois State University

Phone: 309-438-1982



mradams33:
Was there any official determination from the reviewers? Do I need to change the setup for the port_security table, or is it fine as is? What's my next step?

mradams33:
Bump

murrant:
Yes, you need to add `$table->id()` to your table's migration.  This is required for things like relationships and Percona.

murrant:
This looks pretty good overall, but I am curious.  I don't see any way to consume the data.  No API or webui.

mradams33:
I'm trying to add API calls for this. I created doc/API/PortSecurity.md, added in functions to includes/html/api_functions.inc.php, and added in routes to routes/api.php, but the new functions don't show up in the list when I do a curl to list all of the API's. Direct calling the functions doesn't work either. Is there something else I need to do? Does a service need restarted? I haven't committed these changes yet, just testing on my server.

mradams33:
I think I have everything ready. Not sure why the Web test is failing with connection refused. It failed on that before, but changed to succeeded without another commit.

murrant:
Upstream web test broke for awhile, close/open to re-run tests.

murrant:
@mradams33 You are getting there. Here are my notes:
I'd like to see an icon or something shown on ports to indicate port security is enabled.
Your poller module is written in hard mode copying the old code.
1. Instead of if OS checks, it should use inheritance 
2. polling should check for database entries before running snmp queries for performance reasons
3. use the SyncsModels trait to avoid having to manually sync the database

Look at this module for example: https://github.com/librenms/librenms/blob/master/LibreNMS/Modules/Vminfo.php

mradams33:
I think I'm reaching my ability to understand how things are working. I'm stuck at trying to understand where "vminfo" is coming from. I've tried simply replacing it with port_security (and other variations), but it just returns null. If I use vminfo, it returns as empty so there seems to be a built-in reference to something, but I can't figure out what it is. The other part I'm not sure about, but may be answered when the first is figured out is "instanceof VminfoPolling". Is there anything I need to do for it to return as PortSecurityPolling other than create the file (already done)?

$os->getDevice()->vminfo->isEmpty()
and
if ($os instanceof VminfoPolling) {
            $vms = $os->pollVminfo($os->getDevice()->vminfo);

Thank you guys for the help, it's been a learning experience. The only PHP I know is from trying to do this PR so thanks for bearing with me on trying to figure all of this out.

murrant:
That is a device relationship.  Defined here: https://github.com/librenms/librenms/blob/master/app/Models/Device.php#L1054

basically it translates to ```select * from `vminfo` where `vminfo`.`device_id` = ? and `vminfo`.`device_id` is not null``` and fills in the device_id from $device.  Then stores the results on $device->vmInfo, so the next time it is used it doesn't have to be fetched from the database.

More info: https://laravel.com/docs/11.x/eloquent-relationships

mradams33:
I'm a little confused on the STAN error 
Call to an undefined method                                                     
         Illuminate\Support\Collection&iterable<App\Models\PortSecurity>::makeHidden(). 
When I do a get_class(), it returns as Illuminate\Database\Eloquent\Collection which makes makeHidden a valid call. Plus the code works as intended on my Libre server.

Jellyfrog:
If you fix the other issues it sometimes resolves those weird ones automatically.

mradams33:
Finally got all checks to pass again for polling. I will try to figure out how to display some of the data in the web UI.

mradams33:
Are you thinking you want the data displayed on the device/port view (basic and/or detail), or having a separate tab like ARP/FDB tables and Neighbors?

laf:
@mradams33 I tried to rebase this PR but with nearly 2 years of updates it's gone a bit wrong!

I've created a new pull request with all the changes and updates I've done. Not trying to steal your thunder so if you're still around and want to submit this PR again, please grab my branch and submit it.

mradams33:
> @mradams33 I tried to rebase this PR but with nearly 2 years of updates it's gone a bit wrong!
> 
> I've created a new pull request with all the changes and updates I've done. Not trying to steal your thunder so if you're still around and want to submit this PR again, please grab my branch and submit it.

I'd rather not go through trying to figure things out again. To be clear, the port_security table and module will be added to the official release?

laf:
> I'd rather not go through trying to figure things out again. To be clear, the port_security table and module will be added to the official release?

Correct, it's now been merged in. However the table has changed so that it's no longer using column names based on the Cisco mib and is more generic so you'll need to drop the table you created.

laf:
This is now out in 25.10.0 @mradams33 and I've attributed it to you in the change log. Thanks.

PR #15306, opened 2023-09-12
Refactor discovery-protocols with SnmpQuery

PipoCanaja:
Rewrite discovery-protocols using SnmpQuery. 

#### Please note

#### Testers

### DIFF
--- a/includes/discovery/discovery-protocols.inc.php
+++ b/includes/discovery/discovery-protocols.inc.php
@@ -7,7 +7,7 @@ global $link_exists;
 
 if ($device['os'] == 'ironware') {
     echo ' Brocade FDP: ';
-    $fdp_array = snmpwalk_group($device, 'snFdpCacheEntry', 'FOUNDRY-SN-SWITCH-GROUP-MIB', 2);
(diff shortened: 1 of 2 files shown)


includes/discovery/discovery-protocols.inc.php
```diff
@@ -238,9 +237,9 @@
     echo PHP_EOL;
 } else {
     echo ' LLDP-MIB: ';
-    $lldp_array = snmpwalk_group($device, 'lldpRemTable', 'LLDP-MIB', 3);
+    $lldp_array = SnmpQuery::allowUnordered()->hideMib()->walk('LLDP-MIB::lldpRemTable')->table(3);
```
murrant, line 240: Can we limit allowUnordered to a specific OS?  It is preferred since that option can cause inifinite loops.
(there is a field to set specific OIDs to be allowed to be unordered per-os)

PipoCanaja: Makes sense. Just found an example of this.

PipoCanaja: Solved.

PipoCanaja:
reopened in #15324

PR #15309, opened 2023-09-13
Discovery: Allow filtering by device groups

mwobst:
The discovery supports filtering by stuff like module, os and type, but not by device group. This pull request enables filtering either by name or by ID.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/wrapper.py
+++ b/LibreNMS/wrapper.py
@@ -287,9 +287,7 @@ def poll_worker(
                 )
                 if exit_code not in [0, 6]:
                     thread_name = threading.current_thread().name
-                    logger.error(
-                        f"Thread {thread_name} exited with code {exit_code}"
-                    )
+                    logger.error(f"Thread {thread_name} exited with code {exit_code}")
                     ERRORS += 1
                     logger.error(output)
                 elif exit_code == 5:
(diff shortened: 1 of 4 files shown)


murrant (changes requested):
This is not the correct way to filter devices.  Use poller groups.

mwobst:
> This is not the correct way to filter devices. Use poller groups.

Okay, sorry, I should explain this one a bit more.

First: Would a feature request a better idea, for the sake of discussion?

Now the motivation:
We need to discover some modules quite regularly, but there is no distributed discovery (and we don’t want to use a single “behemoth” VM) and not all modules should be fired on all devices on a regular basis.

Also: We don’t want to mix up the distributed polling mode with the discovery. We use a single poller group id to let (right now) 6 pollers do the “loadbalancing”. Using those groups for some static filtering would be complicated. Using device-groups would make things a lot more flexible.

Some examples:
- several nexus device want regular bgp session discovery, but it would be pointless to run the module for every device (as the core module is also called)
- to retrieve the mac addresses at the ports, we want to call the fdb-table module every 5 minutes, using a pool of device groups to distribute all devices.

murrant:
I think this is the wrong approach to fix your problems.

Suggested options:
1. run this code in your local instances (might be a pain for you)
2. add polling for the modules you require (disabled by default)

Curious as to why you specifically need BGP discovery?

murrant:
Closing as this needs a bit more thought and is incompatible with some other work.  Thanks for the PR!

PR #15330, opened 2023-09-18
Add alert transport for Grafana Oncall

waddles:
Adds an alert transport for [Grafana Oncall](https://github.com/grafana/oncall) which uses a [Formatted Webhook](https://grafana.com/docs/oncall/latest/integrations/webhook/) payload.

Grafana Oncall only has the ability to list a single image url so it includes a graph for the device itself instead of a list of graphs for each interface that would be alerting.

Also there is a bug in the [rendering of the graph urls inside Grafana Oncall](https://github.com/grafana/oncall/issues/3031) but I have implemented a workaround that works for us.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Alert/Transport/Grafana.php
+++ b/LibreNMS/Alert/Transport/Grafana.php
@@ -28,7 +28,6 @@ class Grafana extends Transport
 
     public function deliverAlert(array $alert_data): bool
     {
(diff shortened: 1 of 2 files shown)


electrocret (approved):
LGTM

PR #15350, opened 2023-09-26
Update Windows version naming

I-FGSD:
I added Server 2022 and Windows 10/11 22H2 (and a few other popular but older windows builds that people might still have in use).

I also propose changing the naming of windows machines to better reflect the release and version of windows. This naming is derived from this page: https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions

For example, for Windows Server it was previously set to something like 'Windows Server 2019 1809 (NT 6.3)' which is incorrect as MS stopped using NT 6.3 since Windows 10 version 1507. Now it'll display as 'Windows Server 2019 (1809)'. The kernel/version is in parentheses and isn't also in the main part of the name as it's redundant/duplicated information.

#### Please note

#### Testers


murrant (approved):
I'm surprised there is no test data for this, but there are a lot of versions.

LGTM

PR #15354, opened 2023-09-27
Add unzip to all install docs dependencies

brointhemix:
For Debian 11 installation, Composer complains that neither unzip nor 7z is installed so it would unzip using PHP which may or may not corrupt the results file. This has always worked fine for me but I guess it might be a better idea to use the "real" unzipper.

I added 'unzip' to the list of installed software for Debian 11 steps in the documentation. I made no other changes.

```
librenms@librenms1:~$ ./scripts/composer_wrapper.php install --no-dev
Could not scan for classes inside "vendor/dapphp/radius" which does not appear to be a file nor a folder
> LibreNMS\ComposerHelper::preInstall
Installing dependencies from lock file
Verifying lock file contents can be installed on current platform.
Package operations: 118 installs, 0 updates, 0 removals
As there is no 'unzip' nor '7z' command installed zip files are being unpacked using the PHP zip extension.
This may cause invalid reports of corrupted archives. Besides, any UNIX permissions (e.g. executable) defined in the archives will be lost.
Installing 'unzip' or '7z' (21.01+) may remediate them.
```

#### Please note

#### Testers

### DIFF
--- a/doc/Installation/Install-LibreNMS.md
+++ b/doc/Installation/Install-LibreNMS.md
@@ -21,7 +21,7 @@ Connect to the server command line and follow the instructions below.
 === "Ubuntu 22.04"
     === "NGINX"
         ```
-        apt install acl curl fping git graphviz imagemagick mariadb-client mariadb-server mtr-tiny nginx-full nmap php-cli php-curl php-fpm php-gd php-gmp php-json php-mbstring php-mysql php-snmp php-xml php-zip rrdtool snmp snmpd whois unzip python3-pymysql python3-dotenv python3-redis python3-setuptools python3-systemd python3-pip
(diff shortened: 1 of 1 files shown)


murrant:
I don't understand why this is needed.  Did something fail?

Landrash:
@murrant It avoids the error/warning about unzip/7zip not being present. :)

brointhemix:
Hello!

I agree this is not a big update. But yes, with unzip installed you can avoid the warning from composer that it needs to decompress archives on its own rather than use a dedicated binary.

murrant:
If this is truly the case, shouldn't all install docs be updated? 

Please keep the packages alphabetical.

brointhemix (commented):
Sorted 'unzip' package alphabetically in installation steps for Ubuntu, Debian and CentOS.

brointhemix:
When I checked right now all other steps had unzip added except Ubuntu Apache section.

murrant (approved):
Thanks for looking and checking @brointhemix !

brointhemix:
Thanks for reviewing and accepting @murrant :)

Issue #14808, opened 2023-01-26
raisecom-ros is broken for Raisecom RAX721 switches


friek:
### The problem

The Raisecom RAX721 switches are relatively new models running major version 6 of Raisecom ROS. It turns out that some SNMP tables like all IF-MIB tables but also Q-BRIDGE-MIB tables don't have increasing OIDs in an snmpwalk. For example:
```bash
$ snmpwalk -v2c -M $HOME/.snmp/mibs:/usr/share/snmp/mibs:/usr/share/snmp/mibs/iana:/usr/share/snmp/mibs/ietf:$HOME/mibs/raisecom:. -c public hostname IF-MIB::ifIndex 2>&1 | tail -6
IF-MIB::ifIndex.272646331 = INTEGER: 272646331
IF-MIB::ifIndex.272646335 = INTEGER: 272646335
IF-MIB::ifIndex.268435456 = INTEGER: 268435456
Error: OID not increasing: IF-MIB::ifIndex.272646335
 >= IF-MIB::ifIndex.268435456
```

In my local installation, I've fixed this by adding the unordered OIDs to `includes/definitions/raisecom-ros.yaml` which leads to snmpwalk being called with `-Cc`:
```yaml
oids:
  unordered:
    - ifXEntry
    - ifDescr
    - ifAdminStatus
    - ifOperStatus
    - ifLastChange
    - ifType
    - ifPhysAddress
    - ifMtu
    - ifInErrors
    - ifOutErrors
    - ifInDiscards
    - ifOutDiscards
```

Besides the IF-MIB tables above I've also tried adding the `Q-BRIDGE-MIB::{dot1qVlanCurrentUntaggedPorts, dot1qVlanCurrentEgressPorts, dot1qVlanStaticUntaggedPorts, dot1qVlanStaticEgressPorts}` to fix vlan detection but that didn't work out alright. Those tables also don't have increasing OIDs in an snmpwalk but vlan detection is somehow broken anyway (which I consider a separate issue). Also note that in the unordered list above, the `IF-MIB::` prefix is excluded from the entries as they aren't used as such by LibreNMS when generating the snmpwalk commands.

Although this could be considered a bug in the SNMP implementation of ROS, they have no intention of fixing it. 

I can create a PR to merge in the locally made changes in `raisecom-ros.yaml` if that's desirable.

### Output of ./validate.php

```txt
% ./validate.php
[...]

PR #15365, opened 2023-09-30
Fix Raisecom devices having not increasing OIDs on snmpwalk

MrShunz:
Add useful IF-MIB entries as 'unordered oids'
to Raisecom ROS and ROAP definitions (raisecom.yaml and raisecom-ros.yaml)
because they don't have increasing OIDs in snmpwalk

#### Please note

#### Testers


murrant:
That is a horrifying SNMP implementation.  Unfortunately, the code that calls those OIDs does not use SnmpQuery yet, so I don't think this will work.

MrShunz:
Actually, taking hint from #14808 we tried on our installation and confirm `snmpwalk`gets called with `-Cc` flag and at least we get interface and some sensors data. Before it failed miserably after collecting a bunch of OIDs. And it was pretty useless if not for reachability checks.

murrant:
Oh nice, I did add support for that in the legacy code :)

murrant (approved):

PR #15366, opened 2023-10-02
Fix IPv6 address discovery for some junos devices (fixes #8770)

### LINKED ISSUE #8770: IPv6 addresses not being picked up on Juniper devices
Hello,

I'm having an issue where I have a couple of Juniper devices added and the IPv4 addresses are picked up no problem. However, the IPv6 addresses are not being picked up. The IPv6 BGP sessions are detected, however. But then the local IPv6 address is nowhere to be found.

I can confirm that the IPv6 addresses are available to be polled. I have added the same device to another Network Monitoring Platform and all the IPv6 IPs show up. Currently testing against an MX 104.

Has anyone else noticed this issue? Is there a workaround? Could it be due to missing MIBs?

gunkaaa:
Fix IPv6 address discovery for some Juniper devices. Fixes #8770 

Some Juniper devices return `SNMPv2-SMI::zeroDotZero` for `IP-MIB::ipAddressPrefix.ipv6.` ; we need to use `IPV6-MIB::ipv6AddrPfxLength.` instead. 

See:
* https://github.com/librenms/librenms/issues/8770
* https://community.librenms.org/t/issue-with-ipv6-address-discovery-on-juniper-platforms/8360

#### Please note

#### Testers


murrant:
This fix seems like it could break other devices.  If any item return zerodotzero, it will abort.

gunkaaa:
> This fix seems like it could break other devices. If any item return zerodotzero, it will abort.

Thanks @murrant, good catch.

I guess we'd want to check that there are no valid prefixes in `IP-MIB::ipAddressPrefix.ipv6.` at all? And would it be best to break this off into /includes/discovery/ipv6-addresses/junos.inc.php?

PR #15372, opened 2023-10-04
Added extra ASN-range comments, added Priv eBGP on host details, changed color to light-blue instead of red

tim427:

#### Please note

#### Testers

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1182,6 +1182,11 @@ function cache_peeringdb()
             sleep($rand);
             $peer_keep = [];
             $ix_keep = [];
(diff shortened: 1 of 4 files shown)


tim427:
> Maybe define them as one or more constants in the AutonomousSystem class that way it isn't all over the place. Only problem with that is you get into probably modifying those sql queries, which are legacy and need to be ported.

Yeah, would be perfect holidays project ;)

For now I've added the same method at the "Device -> Routing" view, as with "Routing"

(and the coloring changed, because I associated "red" with problems :) )

murrant (approved):

PR #15374, opened 2023-10-04
Specify radius attributes for auth roles

barryodonovan:
Add back to documentation the accepted radius attribute vales for each role.

I implemented this yesterday and could only find the appropriate values by reading back through the commit logs to their removal.

#### Please note

#### Testers

### DIFF
--- a/doc/Extensions/Authentication.md
+++ b/doc/Extensions/Authentication.md
@@ -304,13 +304,14 @@ The attribute `Filter-ID` is a standard Radius-Reply-Attribute (string) that
 can be assigned a specially formatted string to assign a single role to the user. 
 
 The string to send in `Filter-ID` reply attribute must start with `librenms_role_` followed by the role name.
-For example to set the admin role send `librenms_role_admin`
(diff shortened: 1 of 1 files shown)


murrant:
Hi, the reason they were removed is because there are no fixed roles.  It could be anything.   It might make sense to add them back with context that they are the built in roles and any role name can be set.

barryodonovan:
I've added a clarification that these are the built-in roles but that any role name can be passed. I'm on monthly so haven't seen full RBAC implementation, I will update docs again if required once I've used the feature.

murrant (approved):
Looks great, thanks!

murrant:
RBAC is implemented, but no management ui exists yet :)

barryodonovan:
I'll look at using what's there but don't think UI is in my skillset!

murrant:
The items:

1. figure out how collect a list of all permissions (probably just in a json file)
2. add a ui to create roles and assign permissions to roles
3. possibly allow permissions added directly to users, (unsure)

PR #15378, opened 2023-10-05
add dhcp-stats tests and update for v3 of the extend

VVelox:
Now displays lease and subnet info. :3

#### Please note

#### Testers


includes/html/functions.inc.php
```diff
@@ -1230,3 +1230,98 @@ function nfsen_live_dir($hostname)
...
+
+/**
+ * @params array
+ *
+ * @return string
+ *
+ * .headers :: A array of header items
+ * .rows :: A array of arrays. First level is row and second is column.
+ */
+function render_table($table_info)
```
murrant, line 1242: Any html templating should be done in blade. (https://laravel.com/docs/10.x/blade)

Also, this has a lot of security vulnerabilities.

VVelox: What vulnerabilities? That is a bit vague.

Looking this over and the only possibility I see is possible HTML injection via the returned data.

Jellyfrog (approved):

PR #15387, opened 2023-10-05
Smokeping menu and title use device displayname

fbouynot:
When devices are added with IP address but a displayname exists, use it in smokeping: 
image

#### Please note

#### Testers

### DIFF
--- a/app/Console/Commands/SmokepingGenerateCommand.php
+++ b/app/Console/Commands/SmokepingGenerateCommand.php
@@ -128,7 +128,7 @@ class SmokepingGenerateCommand extends LnmsCommand
         $smokelist = [];
         foreach ($devices as $device) {
             $smokelist[$device->type][$device->hostname] = ['transport' => $device->transport];
-            $smokelist[$device->type][$device->hostname]['displayname'] = $device->displayName() ?? $device->hostname;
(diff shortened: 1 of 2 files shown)


app/Console/Commands/SmokepingGenerateCommand.php
```diff
@@ -128,6 +128,7 @@ public function buildTargetsConfiguration($devices)
         $smokelist = [];
         foreach ($devices as $device) {
             $smokelist[$device->type][$device->hostname] = ['transport' => $device->transport];
```
murrant, line 130: Maybe set like this: 
```php
$smokelist[$device->type][$device->hostname] = [
    'transport' => $device->transport,
    'displayname' => $device->displayName(),
];

```

murrant: This entire loop could be replaced by `$devices->groupBy(['type', 'hostname'])`  Then you have full access to the device later instead of just selected fields.

fbouynot:
In SmokepingCliTest.php, testBuildTargets() calls buildTargets instead of buildTargetsConfiguration, bypassing my changes.
So I need to edit SmokepingCliTest.php to add displayname in $groups.

PR #15391, opened 2023-10-05
Throttle error reporting

murrant:
Sets how frequently errors can be reported (across all pollers) Also has the side effect of at most 1 error reported per run To disable, set reporting.throttle to 0 (for development and testing purposes)

With this setting, it should be safe for users to have error reporting enabled without destroying their performance if they had common or frequent errors.

#### Please note

#### Testers


Jellyfrog:
This should be using the native function, https://laravel.com/docs/10.x/errors#throttling-reported-exceptions

murrant:
Yes, but our Laravel isn't new enough and this is causing issues now.

PR #15421, opened 2023-10-09
Fix fping executable selection for IPv4 family

gdepeyrot:
Fix the fping executable selection for IPv4 family so that fping will be executed and fallback to fping6 -4 only if fping is not available. This change allows fping to ping hostnames that resolve as IPv6 and not fail when IPv4 does not answer to ping (some devices resolve both as IPv6 and IPv4 but only IPv6 answers ping requests because the IPv4 DMZ does not). It was a regression introduced in 23.8

#### Please note

#### Testers


murrant:
This still looks wrong.  fping6 is not an executable that exists on modern OS.

It seems there should be a third option:
IPv6 only
IPv4 only
OS Default (IPv6+IPv4 fallback or whatever it is)

murrant:
Perhaps, afaik centos is the only one that doesn't support -6.

murrant:
@gdepeyrot Any progress on this?

gdepeyrot:
Hi Tony, I pushed an update to my change proposal a few weeks ago. Is there anything I should change in it ?

murrant:
As I mentioned.  This change will not work for some users.

What we need is the user to explicitly communicate their intention.
We need a new device setting in LibreNMS: `Use IP family" with the options:
OS Default
IPv6 Only
IPv4 Only

gdepeyrot:
Ok, I'll look into that.

PR #15445, opened 2023-10-12
Add support for FiberStore "data center" switches

cmadamsgit:
FiberStore has two different model lines of switches, and this adds support for the "data center" line.  This was configured and tested against the N5860-48SC model.

The MIBs are from FiberStore's website, with the exception of the FS-FIBER-MIB (the website version is a little behind the software, so support sent me an updated version and should be getting it published soon).

This does not currently handle 4x10 or 4x25 QSFP+ light levels, as I don't currently have any to test against.

#### Please note

#### Testers


murrant:
Great!  It looks like you generated your test data on an older version of LibreNMS.  You may want to update and re-generate it.

includes/definitions/fs-switch-dc.yaml
```diff
@@ -0,0 +1,19 @@
...
+text: 'Fiberstore Datacenter Switch'
+type: network
+icon: fs
+group: fs
+over:
+    - { graph: device_bits, text: 'Device Traffic' }
+    - { graph: device_processor, text: 'Processor Usage' }
+    - { graph: device_mempool, text: 'Memory Usage' }
+mib_dir: fs
+poller_modules:
```
murrant, line 11: You don't need to set these modules to false.  Please remove them.

includes/definitions/fs-switch-dc.yaml
```diff
@@ -0,0 +1,19 @@
...
+    stp: false
+    wireless: false
+discovery_modules:
+    stp: false
+    wireless: false
+discovery:
+    - sysObjectID:
+        - .1.3.6.1.4.1.52642.1.1.10.1.516
```
murrant, line 19: This discovery is very specific and seems to overlap with fs-switch definition.

murrant: Perhaps this could be added to fs-switch?

murrant: If not it would be helpful to remove some of the oids from fs-switch and broaden the sysObjectID here.

cmadamsgit: I'll remove the poller_modules bit - I set that though because other definitions do it that way (disable the same modules for both discovery and poller).

Digging a little more, the sysObjectID should actually be backed up one level - 516 is a specific model switch (although FS isn't actually updating the MIB to document that).

However, I'm not sure of the best way to handle things... FiberStore has two switch lines, enterprise and data center, and they appear to run different OSes (their documentation refers to them as separate). That was why my approach was to create a whole new device type. Nothing under the existing fs-switch type exists on the data center OS, not even basic things like model or serial. Also, it looks like they put both lines under the same sysObjectID, so not easy to separate.

While I've used the enterprise line at a previous job, I don't have access to any at this time, so I don't have any way to dig in deeper and compare.

I'm open to suggestions on how to handle this.

cmadamsgit: Oh, and the test data was generated on an older version because that's the version I have running in production at this time, due to PHP/MySQL versions. We have plans to update that system, but it'll be a while probably, so I don't have a way to generate the data on a newer version right now.

murrant: Yeah, I'm aware they use multiple OS/vendors for their switches.  It is kind of hard to parse out.  It looks like the previous definition lumped some together, so we can either:
A. Lump them all together and have discovery sort out which sensors exist.
B. Separate them out better.

Generally which MIBs are used is a good guide if devices should be the same or separate OS, but that isn't a hard rule.

I suggest you install LibreNMS on a separate machine for development.  (to avoid breaking production among other things).

murrant (approved):

gebster:
Confirmed that it will work with the S5860-20SQ switch as well


Please update to add   - .1.3.6.1.4.1.52642.1.1.10.1.793   to the discovery


os: fs-switch-dc
text: 'Fiberstore Datacenter Switch'
type: network
icon: fs
group: fs
over:
    - { graph: device_processor, text: 'Processor Usage' }
mib_dir: fs
poller_modules:
    stp: false
    wireless: false
discovery_modules:
    stp: false
    wireless: false
discovery:
    - sysObjectID:
        - .1.3.6.1.4.1.52642.1.1.10.1.516
        - .1.3.6.1.4.1.52642.1.1.10.1.793

cmadamsgit:
I'm not really sure of the best approach... I was too specific with the sysObjectID match (I didn't catch right away that the value I used was one specific model), but I don't know how backing it up a level may interact with LibreNMS's discovery code for other FS products.

PR #15455, opened 2023-10-13
Minor fix and tweak for graylog integration

jepke:
Using graylog integration while having RFC5424 syslog messages `source` field is used to store process names, this causes a mismatch in most cases. So librenms will never display syslog messages.

Changing librenms to query graylog for `gl2_remote_ip` results in all messages recieved from this source ip.

#### Please note

#### Testers

### DIFF
--- a/app/ApiClients/GraylogApi.php
+++ b/app/ApiClients/GraylogApi.php
@@ -103,7 +103,7 @@ class GraylogApi
         }
 
         if ($device) {
-            $query[] = 'source: ("' . $this->getAddresses($device)->implode('" OR "') . '")';
+            $query[] = 'gl2_remote_ip: ("' . $this->getAddresses($device)->implode('" OR "') . '")';
         }
 
         if (empty($query)) {
--- a/app/ApiClients/GraylogApi.php
+++ b/app/ApiClients/GraylogApi.php
@@ -97,13 +97,14 @@ class GraylogApi
      */
     public function buildSimpleQuery(?string $search = null, ?Device $device = null): string
     {
+        $field = Config::get('graylog.query.field');
         $query = [];
         if ($search) {
             $query[] = 'message:"' . $search . '"';
         }
 
         if ($device) {
-            $query[] = 'gl2_remote_ip: ("' . $this->getAddresses($device)->implode('" OR "') . '")';
+            $query[] = $field . ': ("' . $this->getAddresses($device)->implode('" OR "') . '")';
         }
 
         if (empty($query)) {
--- a/LibreNMS/Alert/RunAlerts.php
+++ b/LibreNMS/Alert/RunAlerts.php
@@ -254,8 +254,9 @@ class RunAlerts
         $obj = $this->describeAlert($alert);
         if (is_array($obj)) {
             echo 'Issuing Alert-UID #' . $alert['id'] . '/' . $alert['state'] . ':' . PHP_EOL;
-            $this->extTransports($obj);
-
+            if ($alert['state'] != AlertState::ACKNOWLEDGED || Config::get('alert.acknowledged') === true) {
+                $this->extTransports($obj);
+            }
             echo "\r\n";
         }
 
--- a/app/Http/Controllers/Table/GraylogController.php
+++ b/app/Http/Controllers/Table/GraylogController.php
@@ -111,14 +111,16 @@ class GraylogController extends SimpleTableController
             $displayTime = $message['message']['timestamp'];
         }
 
-        $device = $this->deviceFromSource($message['message']['source']);
+        $device = $this->deviceFromSource($message['message']['gl2_remote_ip']);
+        $source = $this->deviceFromSource($message['message']['source']);
         $level = $message['message']['level'] ?? '';
         $facility = $message['message']['facility'] ?? '';
 
         return [
+            'device'    => $device ? Url::deviceLink($device) : htmlspecialchars($message['message']['gl2_remote_ip']),
             'severity'  => $this->severityLabel($level),
             'timestamp' => $displayTime,
-            'source'    => $device ? Url::deviceLink($device) : htmlspecialchars($message['message']['source']),
+            'source'    => htmlspecialchars($message['message']['source']),
             'message'   => htmlspecialchars($message['message']['message'] ?? ''),
             'facility'  => is_numeric($facility) ? "($facility) " . __("syslog.facility.$facility") : $facility,
             'level'     => (is_numeric($level) && $level >= 0) ? "($level) " . __("syslog.severity.$level") : $level,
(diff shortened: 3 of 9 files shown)


murrant:
I'm not sure this will work for all networks.  (aka it may break more than it fixes)  Do you think it is safe?

jepke:
As far as i've found going back to graylog 2, `gl2_remote_ip` field has the originating IP depending on what source is configured.
Cisco ASR9K uses source field for either process name's or interface ip depending on the location the message was created.
tested it on Huawei and Arista as well.

Tested both Raw as Syslog (rfc5424) graylog inputs, raw has source field set to gl2_remote_ip.

murrant:
@jepke what if the device sends logs from one IP, but LibreNMS polls a different IP?

jepke:
@murrant that will have to be corrected but that is a problem with using source too. Using Raw(tcp/udp) graylog inputs `source` has to match polling ip. Problem with the raw inputs is that severity levels are empty so all the filtering in Librenms isnt working. In order to have librenms properly work with the graylog api you have to use the Syslog(tcp/udp).

RFC5424 describes `eventSource` as application source. When a linux server is sending syslogs source field can contain things as eg. `apache` `nginx` etc. In this case to only way to match these to the correct device is `gl2_remote_ip` field so they actually show up with the correct device in Librenms.

All the networking equipment i have to my disposal has the ability to set source interface for sending syslogs.

Example output from graylog api (this is the Syslog RFC5424 input) 
```
{
  "query": "gl2_remote_ip=\"10.1.64.10\"",
  "built_query": "gl2_remote_ip=\"10.1.64.10\"",
  "used_indices": [],
  "messages": [
    {
      "highlight_ranges": {},
      "message": {
        "gl2_accounted_message_size": 559,
        "level": 6,
        "gl2_remote_ip": "10.1.64.10",
        "gl2_remote_port": 514,
        "streams": [
          "000000000000000000000001"
        ],
        "gl2_message_id": "01HCVZJK8R0001QYEBJVXVBVRX",
        "source": "config",
        "message": "- 33552: RP/0/RSP0/CPU0:config[65923]: %MGBL-CONFIG-6-DB_COMMIT : Configuration committed by user '{{SANITIZED}}'. Use 'show configuration commit changes 1000000984' to view the changes.",
        "gl2_source_input": "6527a33f55948c4213b618cf",
        "application_name": "65923",
        "full_message": "<190>1 Oct 16 12:12:47.652 CEST - config 65923 - - 33552: RP/0/RSP0/CPU0:config[65923]: %MGBL-CONFIG-6-DB_COMMIT : Configuration committed by user '{{SANITIZED}}'. Use 'show configuration commit changes 1000000984' to view the changes.",
        "facility_num": 23,
        "gl2_source_node": "c8b8c824-3dec-4db8-a116-46b59980d585",
        "_id": "8e396ec1-6c0c-11ee-b58a-0242989db672",
        "facility": "local7",
        "timestamp": "2023-10-16T10:12:47.000Z"
      },
      "index": "graylog_0",
      "decoration_stats": null
    }
  ],
  "fields": [],
  "time": 993,
  "total_results": 82,
  "from": "1970-01-01T00:00:00.000Z",
  "to": "2023-10-16T10:12:53.500Z",
  "decoration_stats": null
}
```

This is when in graylog Raw input (tcp/udp) is used where source is the actual IP but missing a few fields librenms does use )
```
{
  "query": "gl2_remote_ip=\"10.1.64.10\"",
  "built_query": "gl2_remote_ip=\"10.1.64.10\"",
  "used_indices": [],
  "messages": [
    {
      "highlight_ranges": {},
      "message": {
        "gl2_accounted_message_size": 314,
        "gl2_remote_ip": "10.1.64.10",
        "gl2_remote_port": 514,
        "streams": [
          "000000000000000000000001"
        ],
        "gl2_message_id": "01HCVZQFYY00015B7AGDKZ55MN",
        "gl2_source_node": "c8b8c824-3dec-4db8-a116-46b59980d585",
        "_id": "ed78ac20-6c0c-11ee-b58a-0242989db672",
        "source": "10.1.64.10",
        "message": "<190>1 Oct 16 12:15:27.446 CEST - config 65923 - - 33554: RP/0/RSP0/CPU0:config[65923]: %MGBL-CONFIG-6-DB_COMMIT : Configuration committed by user '{{SANITIZED}}'. Use 'show configuration commit changes 1000000985' to view the changes.",
        "gl2_source_input": "65254ec955948c4213b4f10d",
        "timestamp": "2023-10-16T10:15:27.454Z"
      },
      "index": "graylog_0",
      "decoration_stats": null
    }
  ],
  "fields": [],
  "time": 8,
  "total_results": 83,
  "from": "1970-01-01T00:00:00.000Z",
  "to": "2023-10-16T10:15:44.762Z",
  "decoration_stats": null
}
```

ottorei:
> @jepke what if the device sends logs from one IP, but LibreNMS polls a different IP?

If this is a concern, one should configure a pipeline in Graylog to add a custom field which uniquely identifies the device such as the "real" hostname or IP in NMS. This is probably only a problem when using raw inputs or if the device does not send it's hostname properly. Since the `gl2_remote_ip` is the originating packet, there might be problems if log forwarding is used ie. the device does not send directly to Graylog. If the originating address is the same that is used to poll the device, this should work just fine. But since someone could have log forwarding in place, I would suggest having a config option to set this to either hostname or IP mode in NMS side.

jepke:
i've been deploying graylog further into our network and found several more devices that send syslog in an even stranger format causing graylog to fill the source field with `month`.

I've also been thinking of making the source: which i changed hard coded into gl2_remote_ip a variable.
Just havent been able to figure out how to create the graylog config option as something like `$graylog['querySource'] = 'gl2_remote_ip'` and leave the default as `source`
That way one can also use graylog magic to create custom fields and use those as source.

jepke:
@ottorei 

Think latest commits give it the flexibility and default to `source` as how it currently is so it wont unintentional break others.

Renamed the added Device column to Origin as like you said when items are forwarded their origin is not the actual device but the forwarder IP. Both Origin and Source column will try to match a device on IP or hostname.

murrant:
@jepke fine with this, but I won't merge it until after this release.

app/ApiClients/GraylogApi.php
```diff
@@ -97,13 +97,14 @@ public function query(string $query = '*', int $range = 0, int $limit = 0, int $
      */
     public function buildSimpleQuery(?string $search = null, ?Device $device = null): string
     {
+        $field = Config::get('graylog.query.field');
```
ottorei, line 100: What if the user has not configured a default? Maybe this should use the configured option and fallback to the default if no configuration is present? Also, a small mention of this feature could be added to the documentation page ;)

murrant (approved):

PR #15511, opened 2023-10-27
Prefetch devices in group to avoid cost of subquery

slalomsk8er:
For widgets that filter by group the subquery is very costly and can lead to a timeout.
As proposed in https://community.librenms.org/t/librenms-getting-slow-and-eventlog-widget-gets-canceled-or-504/22270/4?u=slalomsk8er this prefetches the list of devices in the group to avoid the subquery and is orders of magnitude faster.

#### Please note

#### Testers

### DIFF
--- a/app/Models/DeviceRelatedModel.php
+++ b/app/Models/DeviceRelatedModel.php
@@ -39,13 +39,14 @@ class DeviceRelatedModel extends BaseModel
 
     public function scopeInDeviceGroup($query, $deviceGroup)
     {
-        // Build the comma-separated list of device IDs in SQL
+        // Build the list of device IDs in SQL
         $deviceIdsSubquery = \DB::table('device_group_device')
         ->where('device_group_id', $deviceGroup)
         ->pluck('device_id');
 
-        // Use the result in the whereIn clause
-        return $query->whereIn($query->qualifyColumn('device_id'), $deviceIdsSubquery);
+        // Use the result in the whereIn clause to avoid unoptimized subqueries
+        // use whereIntegerInRaw to avoid a the defautl limit of 1000 for whereIn
+        return $query->whereIntegerInRaw($query->qualifyColumn('device_id'), $deviceIdsSubquery);
     }
 
     // ---- Define Relationships ----
--- a/app/Models/DeviceRelatedModel.php
+++ b/app/Models/DeviceRelatedModel.php
@@ -39,11 +39,14 @@ class DeviceRelatedModel extends BaseModel
 
     public function scopeInDeviceGroup($query, $deviceGroup)
     {
-        return $query->whereIn($query->qualifyColumn('device_id'), function ($query) use ($deviceGroup) {
-            $query->select('device_id')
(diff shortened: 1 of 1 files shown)


app/Models/DeviceRelatedModel.php
```diff
@@ -39,11 +39,14 @@ public function scopeHasAccess($query, User $user)
...
-        return $query->whereIn($query->qualifyColumn('device_id'), function ($query) use ($deviceGroup) {
-            $query->select('device_id')
-                ->from('device_group_device')
-                ->where('device_group_id', $deviceGroup);
-        });
+        // Build the comma-separated list of device IDs in SQL
+        $deviceIdsSubquery = \DB::table('device_group_device')
+        ->where('device_group_id', $deviceGroup)
+        ->pluck('device_id')
+        ->implode(',');
```
murrant, line 46: No need to implode here, then explode down below.

Jellyfrog:
I don't understand how it can be faster to do two seperate queries and glue them togeheter i php. Feels like the original code doesn't quite do what I think in this case.

slalomsk8er:
> I don't understand how it can be faster to do two seperate queries and glue them togeheter i php. Feels like the original code doesn't quite do what I think in this case.

Well, as stated in the linked forum post, we observed, in case of the eventlog widget, a timeout.
This was the generated query:
```
use librenms;select * from `eventlog` where `eventlog`.`device_id` in (select `device_id` from `device_group_device` where `device_group_id` = 5) order by `datetime` desc limit 50 offset 0;
...
50 rows in set (2 min 53.973 sec)
```
So, 2 queries in this merge request versus 1+(rows in eventlog) queries in the original code, explains the performance difference.

murrant:
As Jellyfrog says the subquery should only be executed once.  I wonder if there is a bug in the database server code here.

slalomsk8er:
Maybe this still has some relevance https://stackoverflow.com/questions/16121305/how-to-run-sub-query-first-and-only-once

slalomsk8er:
Will check tomorrow if https://mariadb.com/kb/en/subquery-cache/ explains the behavior.

murrant:
@slalomsk8er If you find a better way, great!  Otherwise if you make the change I requested, I'm fine with this .

slalomsk8er:
10.5.21-MariaDB-0+deb11u1 Debian 11
I explicitly set `optimizer_switch='subquery_cache=on'` but still no cache was used as witnessed by the performance and confirmed with `EXPLAIN EXTENDED ...` and `SHOW WARNINGS;`.

Also applied the requested change.

murrant:
One downside with this approach is there is a limit in the number of placeholders provided for a whereIn this way, as opposed to the subquery where there is no practical limit.  Hopefully no one has that many devices in a device group.

EDIT: I just remembered we can do `whereIntegerInRaw` to mitigate that issue!  Please change the where in to that.

slalomsk8er:
Good catch - 1000 devices in a group could be possible!

PR #15516, opened 2023-10-28
Add support for Nokia 1830 PSS Optical gear

peelman:
Tested against 1830 PSS-8 Chassis, in single and multi-chassis configurations Supports Optical Power, temperature, current, power, and a variety of card states, alarms, and trail status.

Of note: With multi-chassis nodes, reading values from the non-master can be a "slow" process, slow in that the default timeout Libre ships with can cause the SNMP client to begin issuing retries.  This can result in the chassis daemon rate limiting and/or blocking the client.  The solution here is to raise the SNMP timeout within each device's configuration; unfortunately those values are not configurable on a per-OS basis (at least not at this time).

#### Please note

#### Testers


Jellyfrog:
There is a lot of standard mibs here that needs to be removed, like the IF-xxx ones.
Also, we only want to include mibs actually used

peelman:
@Jellyfrog removed extraneous mibs.

electrocret (approved):
LGTM

PR #15522, opened 2023-10-30
Added values from SFP DDM values for awplus switches

luttermann:
This pull request adds sensors for _current_, _dbm_, _voltage_ and _tempreture_ of SFP-modules for switches with _awplus_ os.

#### Please note

#### Testers


murrant:
Any reason you have done this in PHP instead of yaml?

luttermann:
> Any reason you have done this in PHP instead of yaml?

I don't know any other way to link the sensors to the actual interface name, but if it is possible to lookup values and split the $index in yaml it could be a way to go. But i really don't want a situation like on the xos where you have a bunch of sensors and no straight forward way to see what they actually are - see attached image from an Extreme XOS 31.1 switch

image

murrant (approved):
Thanks for your contribution. Please try to use yaml (there are examples of linking to ports in other yaml)

PR #15526, opened 2023-10-31
add auth_ldap_cacertfile and auth_ldap_ignorecert options

peejaychilds:

Add some options to
-  use a different CA cert for LDAPS than the system cert, and 
-  not require a valid cert when doing LDAPS

#### Please note

#### Testers


murrant:
Huh, I usually just do this via the system.

peejaychilds:
Yes I agree having a correctly configured system is much better - however we have some environments where for various reasons touching the system files is difficult and in some instances not appropriate (including some testing scenarios)

I'm going to have another crack at using an unpatched LdapAuthorizer.php again, but from memory we do something so strange with our LDAP we pretty much rewrote most of the functions

murrant (approved):

PR #15535, opened 2023-10-31
Fixed BGP Remote peer link and graph in Routing Overview, or just static text when not a device in LibreNMS

tim427:
Replaced "overlibLink" with "deviceLink" and using the RemotePeer-device, instead of LocalPeer-device.

When the device isn't monitored by LibreNMS, then just a static text with the address is shown.

Screenshot 2023-10-31 at 21 14 15

#### Please note

#### Testers

### DIFF
--- a/includes/html/pages/routing/bgp.inc.php
+++ b/includes/html/pages/routing/bgp.inc.php
@@ -1,7 +1,11 @@
 <?php
 
+use App\Models\Device;
 use LibreNMS\Exceptions\InvalidIpException;
 use LibreNMS\Util\IPv6;
+use LibreNMS\Util\Number;
+use LibreNMS\Util\Time;
+use LibreNMS\Util\Url;
 
 if (! Auth::user()->hasGlobalRead()) {
     include 'includes/html/error-no-perm.inc.php';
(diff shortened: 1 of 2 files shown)


PipoCanaja:
Hi @tim427 
As you jumped right into this page, would you try to move it to Laravel blades ?

PipoCanaja:
(that being said, I tested this PR and it looks better indeed) :)

includes/html/pages/routing/bgp.inc.php
```diff
@@ -272,22 +272,26 @@
...
 
         // Peer Address
-        $graph_array_zoom = $graph_array;
-        $graph_array_zoom['height'] = '150';
-        $graph_array_zoom['width'] = '500';
-        $overlib_link = 'device/device=' . $peer['device_id'] . '/tab=routing/proto=bgp/';
-        $peeraddresslink = '<span class=list-large>' . \LibreNMS\Util\Url::overlibLink($overlib_link, $peer_addr, \LibreNMS\Util\Url::graphTag($graph_array_zoom)) . '</span>';
+        $peer_device = \App\Models\Device::whereHas('bgppeers', fn ($q) => $q->where('bgpLocalAddr', $peer['bgpPeerIdentifier']))->first();
+        if (! empty($peer_device)) {
+            $peeraddresslink = '<span class=list-large>' . \LibreNMS\Util\Url::deviceLink($peer_device, $peer_addr, vars: ['tab' => 'routing', 'proto' => 'bgp']) . '</span>';
```
murrant, line 283: Hmm, looking at deviceLink, I think it should return the second argument if the first is not a device.
Also, you don't need the name (vars:) for the third argument here.  I only put it in last time so I could omit the second argument.

murrant:
If we are interested in redoing this whole page.  The SQL could be optimized a lot like this (basic query, and the peer relationship doesn't exist yet):
```php
$bgpPeers = \App\Models\BgpPeers::with(['device', 'peer'])->get();
```

Another option would be to turn it into an ajax table so it works well if there are a large amount of peers.

murrant (approved):

includes/html/pages/routing/bgp.inc.php
murrant: Fixed, hopefully I didn't break it.

murrant (approved):

includes/html/pages/routing/bgp.inc.php
tim427: Nice fix/addition :)!

PR #15537, opened 2023-10-31
bug - Fix missing device_id in NAC tab view, round 2

PipoCanaja:
fixes librenms/librenms#15534

Copy paste mistake ! I did it manually in the GitHub WebUI instead of using git CLI ...


electrocret (approved):
Derp, I'm surprised I missed that when I reviewed the original.

PR #15561, opened 2023-11-08
Jira transport update

LoveSkylark:
Rewriting Jira Transport to allow for "Webhooks" and "Custom Fields"

Custom fields will provide flexibility in defining which Jira fields should be automatically populated when a new ticket is opened.

The webhook implementation facilitates real-time updates directly in Jira. Two webhooks are used - one for opening tickets and another for closing them. When an alert is triggered, the "open" webhook creates a ticket with a unique ID. Then when the alert resolves, the "close" webhook references that ID to locate and close the associated ticket.

Additional logic runs within Jira as well. For example, if a ticket has been set to "In Progress" status in Jira, the "close" webhook can add a comment to the ticket, rather than closing the ticket directly.

Further information can be found in the updated "doc"

#### Please note

#### Testers


doc/Alerting/Transports.md
```diff
@@ -331,26 +331,49 @@ Configuration of the LibreNMS IRC-Bot is described [here](https://github.com/lib
...
-| URL | <https://myjira.mysite.com> |
 | Project Key | JIRAPROJECTKEY |
 | Issue Type | Myissuetype |
+| Open URL | <https://myjira.mysite.com> /  <https://webhook-open-url> |
+| Close URL | <https://webhook-close-url>  |
 | Jira Username | myjirauser |
 | Jira Password | myjirapass |
+| Enable webhook | ON/OFF |
+| Webhook ID | alert_id |
+| Custom Fileds | {"components":[{"id":"00001"}], "source": "LibrenNMS"} |
```
murrant, line 376: typo here

LibreNMS/Alert/Transport/Jira.php
```diff
@@ -29,54 +22,77 @@
...
                 'issuetype' => [
-                    'name' => $issuetype,
+                    'name' => $issue_type,
                 ],
             ],
         ];
 
+        // Add Custom Webhook ID to the payload
+        if ($webhook_on) {
+            if (! empty($this->config['webhook-id'])) {
```
murrant, line 68: You need to check that ` $data['fields'][$this->config['webhook-id']]` is set too.

murrant: Perhaps throwing an exception if it is not set (telling the user that the given field does not exist)

LoveSkylark: ok I've set it to default key to "alert_id" if value is not set, simpler if you don't need to specify your ID

LibreNMS/Alert/Transport/Jira.php
```diff
@@ -29,54 +22,77 @@
...
 class Jira extends Transport
 {
+    protected string $name = 'Jira';
+
     public function deliverAlert(array $alert_data): bool
     {
-        // Don't create tickets for resolutions
-        if ($alert_data['severity'] == 'recovery') {
-            return true;
+        $webhook_on = $this->config['enable-webhook'];
```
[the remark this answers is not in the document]
LoveSkylark: Good catch! 
set to "?? false" now

LibreNMS/Alert/Transport/Jira.php
```diff
@@ -1,23 +1,16 @@
 <?php
-/* Copyright (C) 2015 Aldemir Akpinar <aldemir.akpinar@gmail.com>
```
PipoCanaja, line 2: Not sure about removing credits on a file like this ... Any reason to do so ?

LibreNMS/Alert/Transport/Jira.php
```diff
@@ -1,23 +1,16 @@
...
+ * option) any later version.  Please see LICENSE.txt at the top level of
+ * the source code distribution for details. */
 
 /**
- * Jira API Transport
+ * Jira Webhook & API Transport
  *
- * @author  Aldemir Akpinar <aldemir.akpinar@gmail.com>
- * @copyright 2017 Aldemir Akpinar, LibreNMS
+ * @author Skylark <https://github.com/LoveSkylark>
```
PipoCanaja, line 12: Same here, you can add but not remove.

PipoCanaja (changes requested):

LibreNMS/Alert/Transport/Jira.php
LoveSkylark: I am not remove any lines or erasing credit from the old transporter, I thew it away and started fresh. 

Besides I don't think there is a line left from the original code that was written 7 years ago especially after murrant rewrote how the whole Transport section works.

LibreNMS/Alert/Transport/Jira.php
LoveSkylark: I am not removing any lines or erasing credit from the old transporter, I thew it away and started fresh. 

Code was originally written as a separate JiraWebhooks.php with webhooks only, then I rewrote it to allow for the original connections as well and submitted it as a replacement of for the old Jira transporter. If you look thorough the code only thing that remains are values needed to make it compatible with the old setup (and I only change that after murrant pointed out I would be breaking older Jira deployment).

But I'm happy to change this to anything that gets this approved, but I would suggest @murrant as actual author.

LibreNMS/Alert/Transport/Jira.php
PipoCanaja: Fine with me.

PipoCanaja:
Looks good to me, but no way to test on my side.

LoveSkylark (commented):
This looks good to me?

electrocret (approved):
LGTM

PR #15566, opened 2023-11-10
Linux MegaRAID SAS fixes

eskyuu:
We had a disk failure that was not alerted by LibreNMS even though the monitoring was working.  I worked out that alert only triggers when the polled value exceeds the threshold.

This patch updates a number of failure thresholds for the MegaSAS MIBs to alert when 1 disk fails instead of more than 1 disk.

I also found that the OID for the check labelled as the virtual disk failure was actually checking the predictive disk failure OID.  The predictive disk failure OID is being checked separately, so I have updated the virtual disk failure OID.

#### Please note

#### Testers


includes/definitions/discovery/linux.yaml
```diff
@@ -35,27 +35,27 @@ modules:
...
                 -
                     oid: adapterPropertiesTable
-                    value: pdDiskPredFailureCount
-                    num_oid: '.1.3.6.1.4.1.3582.4.1.4.1.2.1.19.{{ $index }}'
+                    value: vdOfflineCount
+                    num_oid: '.1.3.6.1.4.1.3582.4.1.4.1.2.1.20.{{ $index }}'
                     index: 'vdfail.{{ $index }}'
                     group: 'RAID Controller {{ $adapterID-APT }}'
-                    warn_limit: 1
+                    high_limit: 0.1
```
murrant, line 42: Why did you set these to 0.1?  It should probably be an integer.

eskyuu: It needs to be 0, but that evaluates to false, making it get set to null in the config and not alerting.  If it is set to 1 it won't trigger until more than 1 disk fails, which is bad.  The database value is a float, so 0.1 fixes the issue of not alerting.

A better fix would be to know where these options are translated from YAML to the database and differentiate between 0 and not set.  This would also require knowing where the limits are used for alerting and making sure that this code also differentiates between 0 and not set.  It may also require checking the config web page to do the same.

In this case, this patch would still be relevant, but all the 0.1 values would become 0.

includes/definitions/discovery/linux.yaml
```diff
@@ -35,27 +35,27 @@ modules:
                     descr: 'Virtual disks'
                 -
                     oid: adapterPropertiesTable
-                    value: pdDiskPredFailureCount
-                    num_oid: '.1.3.6.1.4.1.3582.4.1.4.1.2.1.19.{{ $index }}'
+                    value: vdOfflineCount
```
murrant, line 38: Hmm, the description of this sensor is "Virtual disk failure".
A  disk could be offline for multiple reasons.
Perhaps you should define a new sensor instead?

includes/definitions/discovery/linux.yaml
murrant: I can't reproduce that.  0 values are stored properly.

murrant: Also, wouldn't a value of 0 always be in alarm?

includes/definitions/discovery/linux.yaml
eskyuu: In this context what they refer to as a virtual disk is a RAID array.  This happens if a RAID array has failed (multi disk failure, or single disk failure for RAID0).  I'm not sure what other conditions can trigger the RAID array to go offline, but this can be changed if there's a better OID in the MIB to use for a disk failure.  I can say that the physical disk predictive failure is not the right OID.

eskyuu: Regarding 0 always alerting, I am quite sure that I originally tested using 0 when I originally found the issue, and have been running this as a manual patch.  I have just tested again and using 0 does come into the database correctly.  Either I am wrong for the current state of LibreNMS, and I will update these values to 0.

I can confirm that setting this to 0 does not alarm all the time.  It appears that the alerts only happen when you exceed the high value, and not reach it.  See below where the web UI shows the sensors as green when they are the same as the high value:
image

eskyuu: > Hmm, the description of this sensor is "Virtual disk failure". A disk could be offline for multiple reasons. Perhaps you should define a new sensor instead?

Any update on this?  I am very sure that virtual disk offline is indicative of a RAID array failure (e.g. 2 disk RAID5).  If there are other reasons why a virtual disk is offline, it would be an unusual deployment, and the admin of the system can adjust the threshold manually.

The reason I have modified the existing sensor instead of defining a new one is because the OID that is currently configured on the "Virtual disk failure" sensor in LibreNMS is the Phyiscal Disk predictive failure OID, which is clearly not correct.  The Phyiscal Disk predictive failure OID is also used correctly on the "Critical disks" sensor, so the Virtual disk failure" sensor is currently a duplicate of this.

eskyuu:
If there is a problem with the use of the virtual disk offline counter (which indicated that a RAID array is offline), should I just submit a request to remove the sensor, since the current OID is a duplicate of another sensor and clearly incorrect?

includes/definitions/discovery/linux.yaml
eskyuu: See screenshot below - the value of 0 with high limit of 0 does not result in an alarm.

includes/definitions/discovery/linux.yaml
eskyuu: I'm happy to just adjust all the alarm values to 0 so they alarm when 1 disk fails, and I can change the code to remove the "Virtual disk failure" sensor if there are still questions on the usefulness.

As it currently stands these sensors do not alert until 2 disks fail, and the "Virtual disk failure" sensor looks at predictive failures.  I am quite sure that the vdoffline counter is at worst case the best OID for detecting RAID array failures, and am certain that having 2 sensors looking at the same OID is pointless.

eskyuu: How about this - I've swapped back to the original sensor OID, but have updated the MIB reference to match and also corrected the description to match the OID (degraded array vs failed array).

If the above change is acceptable, what are the thoughts on adding another sensor for a RAID array failure?

eskyuu: Here is the location of code that determines the status shown on the web page, where sensors must be > the limit to change colour.  This is what I was seeing where a single failure still showed as green in the UI:
https://github.com/librenms/librenms/blob/a6809e4dcfaa303b82e970f131a4c489f062b07e/includes/html/functions.inc.php#L1085

murrant (approved):

PR #15575, opened 2023-11-14
WIP - Hide navbar when dashboard is full screen

eskyuu:
Hide the navbar and edit options when the dashboard page is set to full screen, giving more space for data.

#### Please note

#### Testers

### DIFF
--- a/resources/views/overview/default.blade.php
+++ b/resources/views/overview/default.blade.php
@@ -648,5 +648,19 @@
         $('#dashboard_name').val('Default');
         dashboard_add($('#add_form'));
     @endif
+
+     function checkFullScreen() {
+         if (window.innerHeight > (screen.height - 10)) {
+            $(".navbar-fixed-top").hide();
+            $("#dashboard-editor").hide();
+            $("body").css("padding-top", 0).css("padding-bottom", 0);
+         } else {
+            $(".navbar-fixed-top").show();
+            $("#dashboard-editor").show();
+            $("body").css("padding-top", "50px").css("padding-bottom", "50px");
+         };
+     }
+     window.addEventListener('resize', checkFullScreen(), false);
+     checkFullScreen();
 </script>
 @endpush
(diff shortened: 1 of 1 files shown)


eskyuu:
Thanks - I've updated to use your cleaner full screen detection.  The old code was taken from the geographical map page, so I have updated it in the work I'm doing there. 

Before screenshot:
image

After screenshot:
image

eskyuu:
Our TV dashboard doesn't like window.matchMedia('(display-mode: fullscreen)').  I may need to add in the old check as a fallback for older browsers.

murrant:
TV browsers are a little rough :)

eskyuu:
> TV browsers are a little rough :)

Do you have thoughts on merging in the current form?

eskyuu:
I found another issue with using the window.matchMedia method - Chrome on OSX reports fullscreen as true if the Chrome program is maximised (it's called full screen by the OS), even if the web page within chrome is not full screen.

I have reverted this change request back to the original code because this seems to work consistently.

PipoCanaja:
It fails on Chrome OSX currently : 
To reproduce : 
- Take a small browser window (let say 1 quarter of your screen)
- Move it down out of the screen
- Extend it up
   - The bar will soon disappear ...
image

eskyuu:
It will fail on any browser if you make the browser window larger than your screen height.  I'm not sure why anyone would do this under normal circumstances.

Having said this, I also got it to fail by setting my zoom to 80%.

The full screen map page (includes/html/pages/fullscreenmap.inc.php) currently implements the 10 pixel method, so whatever we do here should probably be implemented there too.

PipoCanaja:
> It will fail on any browser if you make the browser window larger than your screen height. I'm not sure why anyone would do this under normal circumstances.
> 
> Having said this, I also got it to fail by setting my zoom to 80%.
> 
> The full screen map page (includes/html/pages/fullscreenmap.inc.php) currently implements the 10 pixel method, so whatever we do here should probably be implemented there too.

Yep, I was not reaching height of screen, but you are right, my zoom is not 100% so that may be the reason it fails.

eskyuu:
There doesn't appear to be a consistently safe way to do this, so I will close this request.

If you add ?bare=yes to the overview URL, it achieves the same result.

PR #15578, opened 2023-11-14
Add port description API endpoints and documentation

nicklockhart-fullfibre:

This pull request adds a new API endpoint - `/api/v0/ports/:portid/description` - and associated functions - `update_port_description` and `get_port_description` - which allow a developer to both read and write the description (`ifAlias`) for a port. Previously, the description could be read as part of the data returned by `/api/v0/ports/:portid`, but there was no way to update the description/`ifAlias` via the API. This PR adds an endpoint to enable such functionality.

#### Please note

    - (not relevant)
    - (not relevant)

#### Testers

I have tested these endpoints on my own install and they appear to work as intended. However, I would appreciate a double-check from a security perspective - I believe that these endpoints are authenticated through `Port::hasAccess`, but if this is not the correct way to verify that a user is authorised to update a port in such a manner, please let me know.

### DIFF
--- a/includes/html/api_functions.inc.php
+++ b/includes/html/api_functions.inc.php
@@ -1084,6 +1084,45 @@ function get_port_info(Illuminate\Http\Request $request)
     });
 }
 
+function update_port_description(Illuminate\Http\Request $request)
+{
+    $port_id = $request->route('portid');
+    $port = Port::hasAccess(Auth::user())
+        ->where([
+            'port_id' => $port_id,
+        ])->first();
+    if (empty($port)) {
+        return api_error(400, 'Invalid port ID.');
+    }
+
+    $data = json_decode($request->getContent(), true);
+    $field = 'description';
+    $content = $data[$field];
+
+    if (empty($content)) {
+        return api_error(400, 'New port description has not been supplied.');
+    }
+
+    $port->ifAlias = $content;
+    $port->save();
+
+    return api_success_noresult(200, 'Port description updated.');
+}
+
+function get_port_description(Illuminate\Http\Request $request)
+{
+    $port_id = $request->route('portid');
+    $port = Port::hasAccess(Auth::user())
+        ->where([
+            'port_id' => $port_id,
+        ])->first();
+    if (empty($port)) {
+        return api_error(400, 'Invalid port ID.');
+    } else {
+        return api_success($port->ifAlias, 'port_description');
+    }
+}
+
 /**
  * @throws \LibreNMS\Exceptions\ApiException
  */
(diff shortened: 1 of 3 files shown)


includes/html/api_functions.inc.php
```diff
@@ -1084,6 +1084,45 @@ function get_port_info(Illuminate\Http\Request $request)
...
+
+    $data = json_decode($request->getContent(), true);
+    $field = 'description';
+    $content = $data[$field];
+
+    if (empty($content)) {
+        return api_error(400, 'New port description has not been supplied.');
+    }
+
+    $port->ifAlias = $content;
```
murrant, line 1106: This won't work as expected, it will be overwritten by the poller.  You would need to set a device attribute to ensure it is not overwritten.

nicklockhart-fullfibre: OK, I'll look into this. I wasn't sure if it would be overwritten by the poller, as you can set the description for a port through the web UI in Port Settings, but if this would be overwritten then I'll look into a better solution for making sure that doesn't happen.

murrant: Look at the code for overriding it in the webui, should show you how to do it.

nicklockhart-fullfibre: > Look at the code for overriding it in the webui, should show you how to do it.

so, just to make sure I'm understanding this right - I'm looking at `/includes/html/forms/update-ifalias.inc.php` - I set the `ifAlias` as I have been doing, but then I also set a flag in `devices_attribs`, and that flag tells the poller to not overwrite the port description?

I can also see that the webUI logs when an interface description is updated - I'll make sure to add that too :)

```php
set_dev_attrib($device, 'ifName:' . $ifName, 1);
```
```
+-----------+-----------+-------------------+--------------+---------------------+
| attrib_id | device_id | attrib_type       | attrib_value | updated             |
+-----------+-----------+-------------------+--------------+---------------------+
|       286 |       154 | ifName:lag-4/fp-3 | 1            | 2023-11-13 16:30:59 |
+-----------+-----------+-------------------+--------------+---------------------+
```

nicklockhart-fullfibre: I've changed `update_port_description` to mimic the behaviour of the web UI, so it shouldn't be overwritten by the poller now. If I've made any mistakes anywhere, let me know.

nicklockhart-fullfibre: @murrant Hi, would it be possible to request a review on this? I requested a review last Monday but haven't heard back from you and GitHub won't allow me to request again.

laf (approved):
LGTM

laf:
Thanks @nicklockhart-fullfibre and sorry for the delay in merging.

PR #15586, opened 2023-11-16
Change default graph image to SVG

electrocret:
All modern browsers support SVG. It is immensely more efficient than png, so it should be the default.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Data/Store/Rrd.php
+++ b/LibreNMS/Data/Store/Rrd.php
@@ -580,7 +580,7 @@ class Rrd extends BaseDatastore
         }
 
         // if valid image is returned with error, extract image and feedback
-        $image_type = Config::get('webui.graph_type', 'png');
+        $image_type = Config::get('webui.graph_type', 'svg');
         $search = $this->getImageEnd($image_type);
         if (($position = strrpos($process->getOutput(), $search)) !== false) {
             $position += strlen($search);
--- a/misc/config_definitions.json
+++ b/misc/config_definitions.json
@@ -5967,7 +5967,7 @@
             "type": "boolean"
         },
         "webui.graph_type": {
-            "default": "png",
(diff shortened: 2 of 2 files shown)


peejaychilds:
Whilst I don't have any particular issue with this change I do remember we did something similar for a cacti deployment about 100 years ago and found the end user performance was impacted as the size of all the graphs was much higher than the PNGs

Looking at my lab deployment just the processor graph on a host ... so this can add up ?

```
 Downloads$du -sh graph*
 20K	graph.png
1.7M	graph.svg

```

murrant:
Make sure you enable gzip (or better) on your webserver. And compare those sizes.
Note that it is enabled for the default install instructions.

electrocret:
I've been using SVG on all of my Libre instances for months, and it has noticeably improved graph lazy load times.

peejaychilds:
```
$ gzip graph.svg
$ du -sh graph*
 20K	graph.png
352K	graph.svg.gz
```

I guess I should go do some real wire comparisons toggling on/off etc with a bunch of graphs.

As I said I don't think this will cause any issues given it is 2023 ... but binary packed images are pretty compact.

Perhaps switching from png->svg just moves some of the processing to the client, or out of the rrdtool process into the web-server, or probably a bit of all of the above....

murrant:
I guess it depends on the graph:
rw-r--r-- 1 murrant murrant  85K Nov 17 07:40 graph.png
-rw-r--r-- 1 murrant murrant 309K Nov 17 07:27 graph.svg
-rw-r--r-- 1 murrant murrant 42K Nov 17 07:27 graph.svg.gz

laf:
I'd vote for this to be the default.

murrant (approved):
We can merge it and see

PR #15597, opened 2023-11-27
Syslog field limitations and tag field visiblity in the frontend

Bierchermuesli:
This PR extends the length of the syslog program, level, and tag fields beyond the previous 10-character limit. Additionally, it makes the tag field visible and filterable in all frontend tables. Previously, the tag field was only accessible in the alert rule manager and similar areas. This change addresses the issue of longer tags/programs being truncated, which previously led to mismatches any alert rules

image

Some test Data: 
```
INSERT INTO `syslog` (`device_id`, `facility`, `priority`, `level`, `tag`, `timestamp`, `program`, `msg`, `seq`) VALUES
(7, 'local7', '190', 'informational', 'FMANFP-6-IPACCESSLOGP', '2023-11-27 15:24:23', 'fman_fp_image:', 'list ACL-IPV4-INGRESS permitted udp 203.0.113.185(39077) -> 203.0.113.184(53), 1 packet', 180953528),
(7, 'local7', '189', 'notice', 'ROUTING-BGP-5-MAXPFX', '2023-11-25 13:32:27', 'BGP', 'No. of IPv4 Unicast prefixes received from 203.0.113.0 has reached 804950, max 1048576', 180559872),
(5, '23', '189', 'ERROR', 'LINEPROTO-5-UPDOWN', '2022-12-05 22:11:07', '', 'Line protocol on Interface GigabitEthernet1/0/41, changed state to up', 93),
(5, NULL, NULL, NULL, NULL, '2022-12-05 22:09:24', 'PROG', 'message', 90);

```

In a further toughs: Im unsure which kind of syslog standard we want to support....? program and tag field? level and prio field?

#### Please note

#### Testers


murrant:
I wonder how well this schema will apply when someone has millions of rows in the syslog table...

ottorei:
> I wonder how well this schema will apply when someone has millions of rows in the syslog table... Probably need to test that and maybe split up the into multiple migrations.

Should there be some kind of soft limit if someone actually pushed all syslogs into NMS? I think having millions of syslog lines would probably cause some kind of issues in any case...

Bierchermuesli:
i resized the fields to varchar 32 and added a index to the tag field. please advice

PipoCanaja:
Looks good to me but I have no LNMS instance I can use to validate the DB part.

Bierchermuesli:
Closed due no further need

PR #15612, opened 2023-12-02
Unix agent: Add Linux thermal sensor support

eliandoran:
The unix-agent returns information about some thermal sensors in the following format:

```
<<<lnx_thermal>>>
thermal_zone0 enabled acpitz 27800 106000 critical 85000 active 55000 active 0 active 0 active 0 active 
thermal_zone1 enabled acpitz 29800 106000 critical 106000 passive 
thermal_zone2 enabled x86_pkg_temp 62000 -274000 passive -274000 passive 
```

This information is not processed by `includes/polling/unix-agent.inc.php` and neither by any submodules in `includes/polling/unix-agent/*.inc.php`.

This uses the same approach as `hddtemp.inc.php`, although I'm not sure if it's the best approach since it seems to be doing quite a bit of DB operations.

image

image

image

#### Please note

#### Testers


eliandoran:
@murrant , @laf ,

This is my first PR on the project so please tell me if there's anything I could improve it with considering I am not familiar with it.

I see that the linter fails with:

> Call to deprecated function dbUpdate()

What would be the best approach to solve it? Maybe to try to use the `Sensor` class.

VVelox:
> @murrant , @laf ,
> 
> This is my first PR on the project so please tell me if there's anything I could improve it with considering I am not familiar with it.
> 
> I see that the linter fails with:
> 
> > Call to deprecated function dbUpdate()
> 
> What would be the best approach to solve it? Maybe to try to use the `Sensor` class.

Yup!

Looking at https://github.com/librenms/librenms/actions/runs/7070582633/job/19247409085?pr=15612 it is using depreciated calls. Will want to look at using the sensor class. :)

VVelox (changes requested):
Appears to be good outside of the depreciated functions. Will want to use the Sensor model.

Also some adding a snmpsim file and matching test results for that file would be great as well to verify it all is working as expected. :)

VVelox:
> > Appears to be good outside of the depreciated functions. Will want to use the Sensor model.
> > Also some adding a snmpsim file and matching test results for that file would be great as well to verify it all is working as expected. :)
> 
> @VVelox, @Jellyfrog ,
> 
> Regarding tests it's a bit more complicated. This is not an addition to the SNMP where there's a whole infrastructure implemented for tests. This one is for the Unix agent, which basically just interprets the output of https://docs.librenms.org/Extensions/Agent-Setup/. I'll see what I can do, but I don't see any tests for Unix agent.

hmmm, derp, yeah, forgot about the agent thing =^.^=

So looking at it, I actually don't think we have test support for that currently. @Jellyfrog @laf @murrant Can either of y'alls confirm that? Taking a look at tests I don't see anything there like that off hand.

That said if possible it would be great to have this using the Sensor model instead of the older DB stuff which we are slowly trying to get ride of. :)

eliandoran:
> > > Appears to be good outside of the depreciated functions. Will want to use the Sensor model.
> > > Also some adding a snmpsim file and matching test results for that file would be great as well to verify it all is working as expected. :)
> > 
> > 
> > @VVelox, @Jellyfrog ,
> > Regarding tests it's a bit more complicated. This is not an addition to the SNMP where there's a whole infrastructure implemented for tests. This one is for the Unix agent, which basically just interprets the output of https://docs.librenms.org/Extensions/Agent-Setup/. I'll see what I can do, but I don't see any tests for Unix agent.
> 
> hmmm, derp, yeah, forgot about the agent thing =^.^=
> 
> So looking at it, I actually don't think we have test support for that currently. @Jellyfrog @laf @murrant Can either of y'alls confirm that? Taking a look at tests I don't see anything there like that off hand.
> 
> That said if possible it would be great to have this using the Sensor model instead of the older DB stuff which we are slowly trying to get ride of. :)

I had a more detailed look at the `unix-agent`. It seems that it's a legacy module (it's an include file that is wrapped into a `LegacyModule`). I think we should refactor first the `unix-agent.inc.php` into a non-legacy module and then we can test it better.

The problems I see right now with testing the unix agent is that:

* When `unix-agent.inc.phg` is imported it already starts reading on a socket. It would be much easier if we just called a function to avoid having to mock sockets (which seems quite a challenge).
* It also interacts directly with the database. It would have been nicer to keep the Device and Sensors in memory.

In the meantime I'm still trying to wrap my head around how to use the new model system with `Sensor`.

laf:
Tests are failing on this and it's now 15 months since the last update. Closing for now but if you can fix the tests please comment and we can re-open

PR #15633, opened 2023-12-12
Add functionality for custom maps (weathermaps)

eskyuu:
This adds functionality for admins to create weather maps.

Overview:
 - Maps have been created using the vis.js library, but with all physics and interactions disabled
 - Editing is done using the vis.js editor
 - In order to allow for 2-way traffic reporting and multiple links between nodes, a mid-point is created for each line.  This can be dragged around in the editor, but is hidden in the viewer.
 - A background image can be added as long as it is under 2MB in size
 - Background images are stored in a separate table in the database so they are not fetched every time the map is viewed
 - Background images are cached to disk, and streamed from disk to the client to avoid fetching binary blobs from the database regularly
 - "Nodes" can added to the map can optionally be associated with a LibreNMS device
 - If either end of a link is associated with as device, the link can be associated with a port
 - If a link is associated with a port, it will show percentage utilisation (this can be turned off)
 - It will attempt to decode the SNMP speed if it is set
 - It will make the lines wider for faster links
 - Lines will change colour on a continuous scale from green->yellow->red based on link percentage utilisation (0->50->100)
 - Lines will change from red->purple if the link utilisation is calculated over 100%
 - Access to a given custom map either requires global read access, or access to ALL devices on the custom map.  If a custom map does not have any linked devices, it will only be visible to users with global read.
 - The viewer will refresh the data based on the polling interval, and update the map accordingly
 
I would appreciate any testing and/or comments on getting it working in its current form.

I do still have a couple of ideas to implement to improve the UI.

#### Please note

#### Testers


eskyuu:
Can I please get some assistance with the following test failures.
 - I can't see how to set the collation for the mariadb 10.5 errors related to utf8mb4 utf8mb4_bin
 - I don't understand why the static analysis is complaing about the background_image property

murrant:
@eskyuu Awesome, this has been on my personal todo.  Excited to take a look (although I'm not sure when I'll have time)

murrant:
I got a brief amount of time to look at this.
The edge/ link functionality seems weird.  There is a point in the middle?
I assumed I could set ports on both ends for edges.
Icon selection probably needs to be improved.  OS icon would be nice as an option too.
Would link be a better terminology for edge?
A widget would be nice :), but that can come in the future.

Code points:
You'll need to add documentation
It could use a simplification/refactor pass.  For example, mixing javascript and blade if/else not ideal.
Use the newer device and port select functionality (see init_select2()).
Create specific permissions, then map them to global read/admin.  This will make it easier when we have full RBAC.
All new Models should use timestamps, and likely you don't need to cast every field.
Model relationships node1/2 edges1/2 are not very clear and could have more descriptive names.

eskyuu:
> I got a brief amount of time to look at this.
> 
>     * The edge/ link functionality seems weird.  There is a point in the middle?

This point is covered in the original pull request text.  In short, the point in the middle is only visible in the editor so it can be moved.

The long version is that vis.js only allows you to have a single colour on an "edge", and also a single bit of text.  The only way to get 2 bits of text and 2 line colours is to have another point in the middle.  Adding the point in the middle allows percentages to be given for link utilisation in both directions, and also to have different colours based on utilisation in each direction.  The point in the middle is big enough to see in the editor so it can be dragged, but is small enough in the viewer to become invisible.

Another reason to have the point in the middle is if you have multiple links running between 2 devices, you can drag the second and third links to the side of the first one.  The php weathemap plugin has similar functionality where you can set a point for the links between devices to pass through.  We use this for some devices that have multiple links between them.

>     * I assumed I could set ports on both ends for edges.

No, you choose one port per link, and choose whether the reverse the traffic direction or not.

>     * The edit bar buttons look like they should have icons.

Yes, the vis.js package does have buttons that have not been included in librenms when the javascript package was included.  Should I download and add these as part of the pull request?

>     * Icon selection probably needs to be improved.  OS icon would be nice as an option too.

The icon selection comes from fontaweseome, which is supported natively by vis.js and both are included in librenms.  I chose the icons that seemed the most useful.  Additional fontawesome icon can be added to the list by code.

OS icon is already a separate option (Device Image) because it's not a fontawesome icon.  I'm happy to rename this option if you think it would help.

>     * Would link be a better terminology for edge?

Yes, I agree that link would be better terminology, however the editor built into vis.js calls it an edge.  I'm happy to implement something if you've got any suggestions on how to resolve this.

>     * A widget would be nice :), but that can come in the future.

I agree, and would probably work on this next.

> 
> Code points:
> 
>     * You'll need to add documentation

I'm happy to do this.  Where do I add documentation?

>     * It could use a simplification/refactor pass.  For example, mixing javascript and blade if/else not ideal.

Once again, I'm willing to do this but will need some more specific pointers.  For example, I have built a single page for both the editor and viewer to avoid duplication of parts of the page, but could split these into separate pages to cut out most of the blade if/else code.

>     * Use the newer device and port select functionality (see init_select2()).

I will do this.

>     * Create specific permissions, then map them to global read/admin.  This will make it easier when we have full RBAC.

I will need help with this (maybe an example of another page / controller that does what you're asking for).  The current access control is not as simple as global read / admin.  I would also be happy to work to implement better access control when RBAC is being developed.

>     * All new Models should use timestamps, and likely you don't need to cast every field.

Do you have an example of a model and corresponding database migration that does use timestamps?

I did notice that the links model that you created seemed to rely on auto detection of most things, so I will have a play with removing some more casts and see if it still works.

I also discovered that the Mariadb and MySQL errors in the filing tests are due to me using json fields, which require case sensitivity provided by utf8mb4_bin, and also show as longtext fields with a data validator.  Should I avoid the use of the json field type in MySQL, or do you want to adjust the tests to allow for them?

I do not understand why phpstan is giving the error on the background image field.

>     * Model relationships node1/2 edges1/2 are not very clear and could have more descriptive names.

I thought about this while making the code, but couldn't initially see how the labels made any difference.  They do influence the direction that traffic flow is shown on the viewer, but this can be reversed if needed (e.g. we have some wireless equipment that only gives throughput stats on the ethernet side of the device, so for the wireless link we need to choose the ethernet port and reverse the traffic direction).  I'm happy to rename them to anything you feel is appropriate.  The other option would be node_from and node_to (and edge_from and edge_to).

murrant:
The link center point makes sense, but I went in with know expectations as a user would, so it could be a point of confusion for users. One a drawn map, I would like to show the connecting ports so techs know which ones to check quickly.  I understand that it makes sense to only use one for usage data. Edge is fine if we can't easily change the nomenclature.

Yeah, download the icons.  I think font awesome has crappy options for network type icons.  Switch,router,firewall,ap are some of the icons I would want. I'm unsure about why I didn't see OS Image, maybe the list was too long and I didn't scan it very well...

Take over this page: https://docs.librenms.org/Extensions/Network-Map/

ports_fdb has timestamps.

I was going to mention this, but json is not supported very well yet, use text or whatever more basic field.
edgesEast edgesWest? or something like that, I ccan't give a better name, because I haven't internalized what they do yet :)

As for refactoring.

To reuse code, I think the best option is components.  We already have some in LibreNMS. https://laravel.com/docs/10.x/blade#components
 If those aren't the right option, you can use blade snippets and include one or the other in the view based on an if statement.

For javascript, personally, working with jquery gives me a headache.  In the future this would likely get ported to Alpine or Vue.js.  Jquery usage is fine here and I'm not going to be very picky about it as long as it works.

eskyuu:
As a counter-point to having 2 ports on the one link - you would actually need up to 3 ports on the one link.  If the actual link that connects the devices does not report traffic correctly, you would need to tag the connecting interfaces at each end, plus a third interface for graphing.  We have actual equipment on our network today that would need this workaround.  I can't see how information for all 3 interfaces could be incorporated into the weathermap unless you added interface names next to the percentages.

Having 2 ports would also make it possible to choose the wrong port at one end, causing confusion between the reported throughput on the weathermap and the data displayed on the graph when you hover over the link (I assume that this is where you would want to show the second interface).

The other consideration is that the existing php weathermap plugin only supports a single interface.  I'm only saying that the expectation for people migrating from php wearthemap is that only a single interface will be supported, and I agree that improving the existing solution is a good idea where possible.

Librenms already has other maps that show device links and dependencies, which this is designed to compliment and not replace.  I think that adding the ability to add explicit port links in addition to the LLDP/CDP would be a good idea to improve those maps.


I think there was some confusion about the vis.js icons.  There are missing images for the buttons on the editor, and not missing icons for nicer network maps.  Vis.js supports external images, as well as icon libraries like fontawesome.  I am happy to incorporate another font library into librenms and use this if someone else can point me at a library to use.


I'm happy to wait for better name suggestions and do a search and replace.  As far as I'm concerned there is nothing special about either end, so any name I come up with is an arbitrary label of the nodes.  As mentioned, for traffic reporting there is a "from" and "to" end, so that would be my suggestion if you're happy for me to change it to that.


Thanks for the remaining info.  I will work on implementing these.

eskyuu:
It turns out there wasn't really as much code re-use as I had imagined when I started.  I have just split it up into separate blades.

eskyuu:
And I believe that all the code points have been addresses.

I believe the outstanding original points are as follows:
- The edge/ link functionality is ok, but needs documenting.
- Documentation needs to be created (I will work on this next)

Future work and items still being discussed:
- Ports on both ends needs to be discussed further
- Widget will follow
- I will need help if a specific permission is required (or this can follow for the editor when RBAC is implemented)
- Node1/2 names may be changed
- Icon selection can be changed if someone can provide a good set of icons to use (either CSS like font awesome, or a set of images)

Work that has been completed/accepted is:
- The missing vis.js icons have been added, so the edit bar buttons look correct
- The edge terminology used by vis.js will stand
- blade if/else has been eliminated by moving to separate pages
- Newer device and port select functionality has bene implemented
- All new Models use timestamps, and field casting has been removed.

eskyuu:
I've just added/updated documentation.  I am now just waiting on feedback.

electrocret:
> I've just added/updated documentation.  I am now just waiting on feedback.

It looks like there are references to undefined properties ATM.

eskyuu:
@electrocret  - This is what I have asked for assistance with above because the test appears to be broken.  The property is defined, and the code works.  I'm guessing the issue is because it's a blob data field, and this is causing an issue with the test.

electrocret:
@eskyuu Sorry, I missed that. I'm sure someone familiar with the tests can help out with that. ( @murrant @Jellyfrog )

I tested this on my Dev instance. This is sweet! 

Thoughts on things it's lacking for it to replace PHPWeathermaps for me, or issues I had:
-Can show bps on links instead of just percentages.
-Dashboard widget - So the Weathermaps can be embedded in dashboards.
-Documentation how to add your own custom Icons. (I can see people complaining that "it's not the logical switch icon" etc, or branding.)
-Add ability to put Labels on Links. (Maybe putting it on the mid-point node?)
-Add ability to aggregate multiple ports data onto a link.
-Clicking on Edge lines links you to the port in question (Or aggregate if multiple ports are in edge).
-Add ability to add a key to the weathermap if desired, so people can see the scale.
-Add ability to customize the color scale. 
-Ability to adjust where on the link lines the usage is displayed (I ran into instances where it was hidden by a device.)
-Configure custom links on nodes. (For big networks I like to breakup weathermaps into smaller bite-sized networks, and link between them with abstracted clouds. Such as each Datacenter getting its own map, and the backbone getting a map)

murrant:
@eskyuu About the lint errors.  Dynamic properties are deprecated in PHP and will be dropped in future versions.

The linter takes database migrations into account, but because you used a statement to create background_image, it does not know it exists. use this instead:
```
$table->binary('background_image');
```

eskyuu:
The issue with blob and varbinary fields is that they can only store 64kiB, which isn't enough for a background image.  All the documentation I read says that mediumblob isn't supported by laravel.

I have come up with a solution where I define the column as a binary, then modify it to be mediumblob in the same migration.  Is this ok?

eskyuu:
@electrocret - Thanks for the feedback.  I would like to either get this merged as-is, or at least make sure everyone is happy with the current implementation before spending more time on adding more features.

The one exception would be to change the custom icons to be a set of images instead of using fontawesome.  This could also make it easier for people to add their own images if they are unhappy with the default set.

One thing that you mentioned that I don't think I can change is the location of the text on the line, because it's controlled by the vis.js library.  I did find that I could influence the text position by moving the center node, but I'm open to any other options if there is a way to influence where vis.js places the text.

eskyuu:
> The issue with blob and varbinary fields is that they can only store 64kiB, which isn't enough for a background image. All the documentation I read says that mediumblob isn't supported by laravel.
> 
> I have come up with a solution where I define the column as a binary, then modify it to be mediumblob in the same migration. Is this ok?

Scrap that - the tests fail because SQLite alter table command doesn't support modifying the column type.

So I'm stuck with how to create the binary data field that will support a megabyte size file.

eskyuu:
> > The issue with blob and varbinary fields is that they can only store 64kiB, which isn't enough for a background image. All the documentation I read says that mediumblob isn't supported by laravel.
> > I have come up with a solution where I define the column as a binary, then modify it to be mediumblob in the same migration. Is this ok?
> 
> Scrap that - the tests fail because SQLite alter table command doesn't support modifying the column type.
> 
> So I'm stuck with how to create the binary data field that will support a megabyte size file.

I've put a try/catch around the alter table command, since SQLite supports 2GB in a BLOB column.  Worst case, any database that doesn't support the alter table syntax may just have a smaller maximum image size.

electrocret:
> @electrocret - Thanks for the feedback. I would like to either get this merged as-is, or at least make sure everyone is happy with the current implementation before spending more time on adding more features.

That works with me. This is night and day better than the existing built in LibreNMS mapping tools.

murrant:
I suggest putting all the node image files in the same directory and adjust .gitignore to include the ones LibreNMS supplies.

eskyuu:
> @eskyuu do you mind if I add a couple of commits such as translation and some visual tweaks?

I'm very happy to incorporate other commits.

eskyuu:
> I suggest putting all the node image files in the same directory and adjust .gitignore to include the ones LibreNMS supplies.

I like this idea.  Let me know if there's a better way of doing this to what I've just committed.

murrant:
TODO:

- [ ] custom request with validation
- [ ] controller refactor
- [ ] permissions
- [ ] ui tweaks

murrant:
@eskyuu Now would be a good time to see how much I broke :D

eskyuu:
I've just had a quick click through the editor and it looks like all the buttons work without spitting out any errors.  

I especially like the change to the custom map manager page.

eskyuu:
OK, I found an issue when trying to upload the background images.  the $.ajax method doesn't like it when you have a background image selected.  From what I can read, you need to use POST and x-www-form-urlencoded content type in order to upload images, but this breaks the validation.

I'm not sure how to keep the validation as well as allow for the image upload.  The best option I can think of would be to make the image upload function separate from the other map settings.

eskyuu:
I have split the background image update into a separate modal so I can use a form data upload without affecting the other settings.  

I have also changed the delete button on the map page to return to the main list of maps, which I think makes it easier to navigate if you want to edit multiple maps.

murrant (approved):

santiag0z:
@murrant and @eskyuu, very happy to know that LIbreNMS has its own Weathermaps.
I was testing here before the monthly version came out, there seems to be a bug in the node style when selecting Device Image and Image.

Below is the bug:
https://github.com/librenms/librenms/assets/34222550/548346e8-c2b3-4181-88ea-370777b1bd06

santiag0z:
> @murrante@eskyuu, muito feliz em saber que o LIbreNMS possui seus próprios mapas meteorológicos. Eu estava testando aqui antes do lançamento da versão mensal, parece haver um bug no estilo do nó ao selecionar Imagem e Imagem do Dispositivo.
> 
> Abaixo está o bug: https://github.com/librenms/librenms/assets/34222550/548346e8-c2b3-4181-88ea-370777b1bd06

Fix by @eskyuu, thanks! #15695

PipoCanaja:
@eskyuu Well done ! I just tested it in my home lab, fantastic work !

antik60:
Nice work @eskyuu !!

It will be useful if maps can be not only nested but can inherrit device status. So at the root map you can have color change if a device is down in a leaf map. As like geographical maps do when you zoom out..

networkgorilla:
This is really nice work and is a much-needed improvement for the mapping functionality as hopefully it can be a solution towards phasing out Weathermapper that is no longer being developed. I do have a recommendation (request?) which could be interesting for distributions with a large number of access switches attached to it. And it would be instead of the image of that distribution being a switch or some other network device, it could be a long vertical line. So then the access switches could be attached to the long vertical line with enough spacing where it doesn't look like a packet switching spider, hehe. 

I've whipped up a quick visualization in Weathermapper. Though unsure of the actual functionality if implemented into this feature as this depends on being able to modify the links to a certain position on the node it is attached based on cardinal directions. I'll also take a look to see if I can't figure something out in that regard.

example

Cheers!

PR #15640, opened 2023-12-14
Fixing memory scale for datacom-dmos devices

lhwolfarth:
Fixing memory scale for datacom-dmos devices

#### Please note

#### Testers


PipoCanaja:
According to the tests, it seems that it is not working. Moreover, you fixed the index to '1'. Better to use the index provided by the device in the MIB (that would allow the device to provide more than one value).
Does this code work in your environnement ? If yes, then you probably need to upgrade the SNMPREC file (and match the JSON file to it of course)

lhwolfarth:
@PipoCanaja, in my tests it was working. I also tested now without using the index 1 and it changed from the scale of MiB to GiB but still consistent (right values).
I will update the yaml removing the "index 1" and the upload the new test data.
Thank you!

PipoCanaja:
Seems that the tests went somehow wrong (all empty). Fixing that for you

PipoCanaja:
LGTM, so let's control the tests and it will be ready for merge

PipoCanaja (approved):

PR #15670, opened 2023-12-24
Search caches more deeply for valid indexes on sensor discovery

peelman:
When there are multiple devices behind a controller, but they are named once; in order to reference that name within a sensor descr or group, the device ID needs to be extracted from a precache entry. 

Example (cut down for brevity):

```snmpwalk -m +ALL -M +/opt/librenms/mibs/poweralert/ -v2c -c public hostname.domain.tld tlpDeviceTable
TRIPPLITE-PRODUCTS::tlpDeviceIndex.1 = Gauge32: 1
TRIPPLITE-PRODUCTS::tlpDeviceIndex.2 = Gauge32: 2
TRIPPLITE-PRODUCTS::tlpDeviceType.1 = OID: TRIPPLITE-PRODUCTS::tlpPdu
TRIPPLITE-PRODUCTS::tlpDeviceType.2 = OID: TRIPPLITE-PRODUCTS::tlpEnvirosense
TRIPPLITE-PRODUCTS::tlpDeviceManufacturer.1 = STRING: TRIPP LITE
TRIPPLITE-PRODUCTS::tlpDeviceManufacturer.2 = STRING: TRIPP LITE
TRIPPLITE-PRODUCTS::tlpDeviceModel.1 = STRING: PDUMV20NETLX
TRIPPLITE-PRODUCTS::tlpDeviceModel.2 = STRING: E2MTHDI 
TRIPPLITE-PRODUCTS::tlpDeviceName.1 = STRING: DEVICENAME1
TRIPPLITE-PRODUCTS::tlpDeviceName.2 = STRING: DEVICENAME2
TRIPPLITE-PRODUCTS::tlpDeviceLocation.1 = STRING: Location Of Device1
TRIPPLITE-PRODUCTS::tlpDeviceLocation.2 = STRING: Location of Device2

snmpwalk -m +ALL -M +/opt/librnems/mibs/:/opt/librenms/mibs/poweralert/ -v2c -c public hostname.domain.tld tlpEnvDetail
Expected DESCRIPTION (Corrected spelling errors): At line 45 in /opt/librenms/mibs/poweralert//TRIPPLITE-12X
Bad operator (INTEGER): At line 73 in /usr/share/snmp/mibs/ietf/SNMPv2-PDU
TRIPPLITE-PRODUCTS::tlpEnvTemperatureC.2 = INTEGER: 238 0.1 degrees Celsius
TRIPPLITE-PRODUCTS::tlpEnvTemperatureF.2 = INTEGER: 749 0.1 degrees Fahrenheit
TRIPPLITE-PRODUCTS::tlpEnvTemperatureInAlarm.2 = INTEGER: false(2)
TRIPPLITE-PRODUCTS::tlpEnvHumidityHumidity.2 = INTEGER: 33 percent
TRIPPLITE-PRODUCTS::tlpEnvHumidityInAlarm.2 = INTEGER: false(2)


snmpwalk -m +ALL -M +/opt/librnems/mibs/:/opt/librenms/mibs/poweralert/ -v2c -c public hostname.domain.tld tlpPduInputPhase
TRIPPLITE-PRODUCTS::tlpPduInputPhaseIndex.1.1 = Gauge32: 1
TRIPPLITE-PRODUCTS::tlpPduInputPhaseFrequency.1.1 = Gauge32: 599 0.1 Hertz
TRIPPLITE-PRODUCTS::tlpPduInputPhaseVoltage.1.1 = Gauge32: 1173 0.1 Volts
TRIPPLITE-PRODUCTS::tlpPduInputPhaseVoltageMin.1.1 = Gauge32: 0 0.1 Volts
TRIPPLITE-PRODUCTS::tlpPduInputPhaseVoltageMax.1.1 = Gauge32: 1196 0.1 Volts
```

This was added specifically to deal with TrippLite's MIBs, but i could see it bearing fruit in other areas too, letting you use Groups more effectively, or otherwise allowing for more descriptive naming of sensors.

#### Please note

#### Testers

### DIFF
--- a/LibreNMS/Device/YamlDiscovery.php
+++ b/LibreNMS/Device/YamlDiscovery.php
@@ -270,6 +270,14 @@ class YamlDiscovery
                     return $pre_cache[$name][$index];
                 } elseif (count($pre_cache[$name]) === 1 && ! is_array(current($pre_cache[$name]))) {
                     return current($pre_cache[$name]);
+                } else {
+                    $index_countdown = explode('.', $index);
+                    foreach ($index_countdown as $sub_index) {
+                        array_pop($index_countdown);
+                        if (isset($pre_cache[$name][implode('.', $index_countdown)][$name])) {
+                            return $pre_cache[$name][implode('.', $index_countdown)][$name];
+                        }
+                    }
                 }
             } else {
                 return $pre_cache[$name];
--- a/LibreNMS/Device/YamlDiscovery.php
+++ b/LibreNMS/Device/YamlDiscovery.php
@@ -270,14 +270,6 @@ class YamlDiscovery
                     return $pre_cache[$name][$index];
                 } elseif (count($pre_cache[$name]) === 1 && ! is_array(current($pre_cache[$name]))) {
                     return current($pre_cache[$name]);
-                } else {
-                    $index_countdown = explode('.', $index);
-                    foreach ($index_countdown as $sub_index) {
-                        array_pop($index_countdown);
-                        if (isset($pre_cache[$name][implode('.', $index_countdown)][$name])) {
-                            return $pre_cache[$name][implode('.', $index_countdown)][$name];
-                        }
-                    }
                 }
             } else {
                 return $pre_cache[$name];


peelman:
Alrighty; I'm at a loss on those those testing failures...the code changed here wouldn't have affected that _one_ device in the manner prescribed by that output...

PipoCanaja:
It does indeed touch all YAML discovery so it may very well impact all devices. You need to be especially carefull, and run those tests (using "save-test-data" on this precise timos device) to understand how YAML discovery was impacted here. If the change is not a problem, then you'll need to update all impacted testfiles. If not you'll need to fix your code to avoid regressions.

PipoCanaja:
And after a quick look at this code, it does not seems to be safe at all to run it on random MIBs. It is very specific to TRIPPLITE, and maybe very specific to a PART of it, not all of it.

peelman:
1) I don't get this error when running the check locally (or i wouldn't have submitted it)
2) if it was broad-reaching, why did only a single unit test fail randomly from a Nokia IXR-s
3) eh...it is not specific to tripplite.  Similiar dumb happens all over where the Description of a thing isn't duplicated all over, but the values for a thing may be off in other MIB trees and only referenced via a subindex.

PipoCanaja:
The problem is that your code might return a subindex match before the default "return" statement, which indeed might break working anything that count on it. 
To check this device in particular, you can us save-test-data script and check the difference in the resulting file on your machine. You can also simulate it ( check lnms dev:simulate ) and run the discovery and the poller to understand how to adapt your code and not break existing tests.

murrant:
@peelman 2. CI stops on the first error to not waste compute.  If you run it locally, you can have it output all errors.

murrant:
@peelman for your usecase, I think {{ tlpDeviceName:1 }} will work.

peelman:
> @peelman for your usecase, I think {{ tlpDeviceName:1 }} will work.

So that didn't work.  But `{{ tlpDeviceName:0 }}` does.

But that goes against [this](https://github.com/librenms/librenms/blob/master/doc/Developing/os/Health-Information.md):

    When referencing an oid in another table the full index will be used to match the other table. If this is undesirable, you may use a single sub index by appending the sub index after a colon to the variable name. Example {{ $ifName:2 }}

Which indicates a `$` is needed (indicating it has to be a real variable (i.e. from the precache, which is what was messing me up (among several other things). when I was trying this before.

Given @murrant's chunk of code that exists to do this in a lot less invasive way, closing/nuking this PR.  Those docs need a twice-over.  Blah.

PipoCanaja:
Hi @peelman 
Could you please open a new PR on the documentation file, with the correct syntax and any explanation you feel necessary to help somebody else in the same situation, and I'll merge it. 
Thanx

peelman:
> Hi @peelman Could you please open a new PR on the documentation file, with the correct syntax and any explanation you feel necessary to help somebody else in the same situation, and I'll merge it. Thanx

I would, and actually started to, but I don't think I have a firm enough understanding for what @murrant was doing there to adequately correct it.  There's a whole thing about ranges that isn't in the docs at all.

It looks like you can do any of these (*assuming an index of `9.8.7.6`*):

* `oidName:0` # returns value at `oidName.9`
* `oidName:1` # returns value at `oidName.8`
* `oidName:1-3` # returns value at `oidName.8.7.6`
* `oidName:2-3` # returns value at `oidName.7.6`

Is that right?  Missing anything?

PipoCanaja:
In fact, both the variable (with $) and the name of the oid (without $) are accepted in the `{{ }}` if I remember well. But the scope is not exactly the same, so for you precached-data, it seems that only the OID was directly accepted. 

What is for sure, the doc has room for improvement here. 

I'll try to get some time to dig in the code and confirm or not what I remember :)

peelman:
n/m, Email saved the day.

    In fact, both the variable (with $) and the name of the oid (without $) are accepted in the {{ }} if I remember well. But the scope is not exactly the same, so for you precached-data, it seems that only the OID was directly accepted.

:)

peelman:
@PipoCanaja @murrant 

        // parse sub_index options name with trailing colon and index
        if (preg_match('/^(.+):(\d+)(?:-(\d+))?$/', $name, $matches)) {
            // subindex found
            $name = $matches[1];

            //create the sub-index values in order to try to match them with precache
            $sub_indexes = explode('.', $index);

            // if subindex is a range, get them all, otherwise just get the first
            $index = isset($matches[3])
                ? implode('.', array_slice($sub_indexes, (int) $matches[2], (int) $matches[3]))
                : $sub_indexes[(int) $matches[2]];
        }

(assuming an `$index` = `9.8.7.6`, `$variable` =  `?`)

* `oidName:0` # returns value at `oidName.9`
* `oidName:1` # returns value at `oidName.8`
* `oidName:1-2` # returns value at `oidName.8.7`
* `oidName:2-3` # returns value at `oidName.7.6`
* `$variable:1` # returns value at `$variable.8`

What I'm fuzzy on is the code is just searching for `$name`, so if you pass `'$name'` its going to look for that strict string, including the `$`. 

I just want to make sure I grok the behavior before trying to clarify the docs.

